PR #82, opened 2016-12-13
fix(menu): Cancel running animations on destroy.

sgomes:
@traviskaufman PTAL!

### DIFF
--- a/packages/mdc-menu/simple/foundation.js
+++ b/packages/mdc-menu/simple/foundation.js
@@ -83,6 +83,7 @@ export default class MDCSimpleMenuFoundation extends MDCFoundation {
     this.scaleY_ = 0;
     this.running_ = false;
     this.selectedTriggerTimerId_ = 0;
+    this.animationRequestId_ = 0;
   }
 
   init() {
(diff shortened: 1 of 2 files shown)


packages/mdc-menu/simple/foundation.js
```diff
@@ -83,6 +83,7 @@ export default class MDCSimpleMenuFoundation extends MDCFoundation {
     this.scaleY_ = 0;
     this.running_ = false;
     this.selectedTriggerTimerId_ = 0;
+    this.animationRequestId_ = null;
```
traviskaufman: nit: since this will always be a non-zero, non-negative integer, it's probably best just to set this to `0`. This also means you can get rid of the `if` statement within destroy and just call `cancelAnimationFrame`. It also keeps it consistent with `selectedTriggerTimerId_`.

packages/mdc-menu/simple/foundation.js
```diff
@@ -186,8 +191,9 @@ export default class MDCSimpleMenuFoundation extends MDCFoundation {
 
     // Stop animation when we've covered the entire 0 - 1 range of time.
     if (currentTime < 1) {
-      requestAnimationFrame(() => this.animationLoop_());
+      this.animationRequestId_ = requestAnimationFrame(() => this.animationLoop_());
     } else {
+      this.animationRequestId_ = null;
```
traviskaufman: same comment as above; we can just set this to `0`.

packages/mdc-menu/simple/foundation.js
sgomes: After running some tests on multiple browsers, it seems safe to call `cancelAnimationFrame(0)`, so making suggested changes.

sgomes:
@traviskaufman Changes made, PTAL!

PR #104, opened 2016-12-15
docs(getting-started): Add post-release amendments and updates

traviskaufman:
- Remove the need for NodeJS/NPM to be installed locally. Use
  [unpkg](https://unpkg.com/#/) instead.
- Remove outdated "coming soon" blurbs.
- Correct minor typos and grammatical errors.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,6 +1,3 @@
-> NOTE: We have yet to publish our initial release to npm. However, once we do, the installation
-instructions will work.
-
 # Getting started
 
 This guide will help you get started using MDC-Web on your own sites and within your own projects.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -56,7 +50,9 @@ the assets needed for MDC-Web. Put the following within `index.html` in the `gre
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <title>Greeting App</title>
-    <link rel="stylesheet" href="/node_modules/material-components-web/dist/material-components-web.css">
+    <link
+      rel="stylesheet"
+      href="https://unpkg.com/material-components-web@latest/dist/material-components-web.min.css">
```
sgomes: LGTM only if `unpkg.com` is reliable.

traviskaufman: It looks reliable to me: https://unpkg.com/#/stats

Thoughts?

sgomes: LGTM 👍 We can always change if we need to.

PR #114, opened 2016-12-16
Use PureComponent instead of mixin

Klaasvaak:
Use PureComponent instead of the PureRenderMixin.

From React docs:
> The PureRenderMixin mixin predates React.PureComponent. This reference doc is provided for legacy purposes, and you should consider using React.PureComponent instead.


traviskaufman:
Thanks for your help with this @Klaasvaak ! As soon as you sign our CLA, I'll be able to merge.

Klaasvaak:
@traviskaufman I signed it

PR #122, opened 2016-12-16
fix(select): Prevent overflow on smaller screens

### LINKED ISSUE #112: Select overflows bounds of viewport on iOS

Latest (0.1.1)

iOS (Chrome)

### What OS are you using?

iOS v10.1.1 (but affects all OSs)

1. Ensure 
1. Navigate to http://material-components-web.appspot.com/select.html
1. Observe the select elements

The entire select element fits within the bounds of the screen

The select overflows the bounds of the screen.

This seems to happen because we don't put any constraints on the width of the select when we calculate it. We should consider putting a `max-width: 100%` style on the element and then truncating the text with an ellipsis, or something similar.

traviskaufman:
- Set max-width of component to 100% - the size of the padding added to
  accomodate the drop-down arrow
- Truncate overflowing selected text, similar to a normal <select>
  element
- Adds "ignore: local" for scss/dollar-variable-pattern to stylelint
  config so local Sass vars don't have to be prefixed
- Disallows the legacy "md-" prefix that was allowed in our stylelint
  config

[Delivers #136281135]
Resolves #112

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -27,9 +27,11 @@
 
 // postcss-bem-linter: define select
 .mdc-select {
+  $dd-arrow-padding: 24px;
+
   @include mdc-typography(subheading2);
   @include mdc-theme-prop(color, text-primary-on-light);
-  @include mdc-rtl-reflexive-box(padding, right, 24px);
+  @include mdc-rtl-reflexive-box(padding, right, $dd-arrow-padding);
 
   // Resets for <select> element
   appearance: none;
@@ -41,6 +43,7 @@
   display: inline-flex;
   align-items: center;
   justify-content: flex-start;
+  max-width: calc(100% - #{$dd-arrow-padding});
   height: 32px;
   transition:
     mdc-animation-exit(border-bottom-color, 150ms),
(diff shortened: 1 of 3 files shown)


packages/mdc-select/mdc-select.scss
```diff
@@ -41,6 +43,7 @@
   display: inline-flex;
   align-items: center;
   justify-content: flex-start;
+  max-width: calc(100% - #{$dd-arrow-padding});
```
amsheehan: Noice

packages/mdc-select/mdc-select.scss
```diff
@@ -95,6 +98,8 @@
     transition:
       mdc-animation-exit(opacity, 125ms),
       mdc-animation-exit(transform, 125ms);
+    white-space: nowrap;
+    overflow: hidden;
```
amsheehan: Mayhap we want to also add `text-overflow: ellipsis`?

traviskaufman: I did that originally, but - at least on iOS - the native select doesn't use it, so I elided it

screen shot 2016-12-16 at 16 32 40

amsheehan: Hmmm. Looks like they shouldn't have spent all that time implementing ES2019. 😅

traviskaufman:
Thanks for the quick review @amsheehan! 🌪

Issue #16, opened 2016-12-05
remove implicit DOM querying within constants files


traviskaufman:
See #17 

Once the above story is done, we should go back and remove places where this is happening.

> Copied from https://github.com/google/material-design-lite/issues/4746

traviskaufman:
@amsheehan can we follow up here with all packages/foundations that will be affected by these changes?

amsheehan:
Existing packages that will need revision:

- mdc-checkbox
- mdc-drawer
- mdc-menu
- mdc-ripple
- mdc-snackbar

We would have to change the foundations classes for each package since the plan is to ask the user to pass the component an object with a DOM object as its child (or any server side flavor thereof).

PR #140, opened 2016-12-21
fix(checkbox): Make DOM access lazy to avoid SSR problems.

petejohanson:
So... I'm working on trying out getting SSR working with VueJS 2.0 and my (early) WIP VueJS binding, and the access of `document` in the `constants.js` of many packages for looking up things like `ANIM_END_EVENT_NAME` bombs as one would expect.

Here is a proposed change to the way those "constants" are looked up to be on-demand and lazy loaded once, with mdc-checkbox as the example.

Thoughts on the approach? If you like it, I'm happy to do the same changes for other components.

### DIFF
--- a/packages/mdc-checkbox/constants.js
+++ b/packages/mdc-checkbox/constants.js
@@ -29,15 +29,20 @@ export const cssClasses = {
   ANIM_INDETERMINATE_UNCHECKED: `${ANIM}-indeterminate-unchecked`,
 };
 
+export function dom() {
+  return {
+    ANIM_END_EVENT_NAME: (() => {
+      const el = document.createElement('div');
+      // NOTE: We can immediately assume that the prefix is 'webkit' in browsers that don't
+      // support unprefixed animations since the only browsers up to two major versions back that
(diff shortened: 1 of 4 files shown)


sgomes:
Woohoo, awesome! :D Thanks, @petejohanson!

@traviskaufman This looks like a fix for #16. Did we settle on an approach for handling this kind of thing, e.g. lazy eval vs utils file?

traviskaufman:
Thanks for taking a stab at this @petejohanson 

As it turns out, @amsheehan is working on a solution for this already as part of #17 , and IIRC is tackling #16 along with it. That should unblock SSR for you. Therefore, I'd hold off on merging this PR, but if the approach outlined in #17 doesn't work for you, please let us know and we can discuss it further.

petejohanson:
The approach proposed for mdc-animation looks great to me. Should work fine for the VueJS SSR case, AFAICT.

Closing this PR then, I'll revisit the SSR scenario once #17 is in.

traviskaufman:
@petejohanson the fix for this is out in our latest release. Let us know if you're still having problems!

PR #142, opened 2016-12-22
Mdc web/impl event name query

amsheehan:
#17 

Added index.js to mdc-animation to serve as a utility library for MDC Web. For now I don't know if the lone function in there should be part of a publicly documented API as it probably won't be used by developers. In the future we may want to implement something though if it becomes more like the Sass portion of mdc-animation.

### DIFF
--- a/packages/mdc-animation/README.md
+++ b/packages/mdc-animation/README.md
@@ -1,6 +1,6 @@
 # mdc-animation
 
-mdc-animation is a Sass / CSS library which provides variables, mixins, and classes for Material Design animation, based off of the [motion guidelines](https://material.google.com/motion/duration-easing.html#duration-easing-common-durations). Currently, it only covers easing curves, but will be expanded in the future. If you are interested in contributing to MDC Web's animation library, take a look at the [contributing page](https://github.com/material-components/material-components/blob/develop/CONTRIBUTING.md).
+mdc-animation is a Sass / CSS / JavaScript library which provides a toolbelt for Material Design animation, based off of the [motion guidelines](https://material.google.com/motion/duration-easing.html#duration-easing-common-durations). Currently, it only covers easing curves, but will be expanded in the future. If you are interested in contributing to MDC Web's animation library, take a look at the [contributing page](https://github.com/material-components/material-components/blob/develop/CONTRIBUTING.md).
 
 ## Installation
 
@@ -97,3 +97,21 @@ CSS Classes have the exact same name as their mixin counterparts.
 
 All animation variables are marked with `!default`, thus they can be overridden should the need
 arise.
+
+### JavaScript
+
+MDC Web ships with a set of utility functions designed to make animations easier to implement.
+
+### Using Utility Functions
+
+To use:
+```js
+import {getCorrectEventName} from '@material/mdc-animation';
+
+const eventToListenFor = getCorrectEventName('animationstart');
+```
+
+| Method Signature | Description |
+| --- | --- |
+| `getCorrectEventName(windowObj: Object, eventType: string)` | Returns a JavaScript event name. Prefixed if necessary. |
+| `getCorrectPropertyName(windowObj: Object, eventType: string)` | Returns a CSS property name. Prefixed if necessary. |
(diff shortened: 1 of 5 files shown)


packages/mdc-animation/README.md
```diff
@@ -1,6 +1,6 @@
 # mdc-animation
```
traviskaufman: I think we should add a section in the README which documents the JS utility functions

packages/mdc-animation/index.js
```diff
@@ -0,0 +1,101 @@
...
+    noPrefix: 'transform',
+    webkitPrefix: '-webkit-transform',
+  },
+  transition: {
+    noPrefix: 'transition',
+    webkitPrefix: '-webkit-transition',
+  },
+};
+
+function hasNode(windowObj) {
```
traviskaufman: maybe a better method name for this would be `hasProperShape`?

Also, we should probably be checking that `typeof windowObj.document.createElement === 'function'`, rather than just checking that it's defined, since we're calling it.

packages/mdc-animation/index.js
```diff
@@ -0,0 +1,101 @@
...
+//
+// Parameters:
+// windowObject: Object -- Contains Document with a `createElement()` method
+// eventType: string -- The type of animation
+//
+// returns the value of the event as a string, prefixed if necessary.
+// If proper arguments are not supplied, this function will return
+// the property or event type without webkit prefix.
+//
+export function getAnimationEventName(windowObj, eventType) {
```
traviskaufman: hmm, I'm wondering if people may think that this method implies that this is only for  _animation event_ names, and not css properties or things dealing with `transition`. Perhaps breaking this into two distinct functions would make it more semantic? e.g. `getCorrectEventName`, `getCorrectPropertyName` or something similar.

test/unit/mdc-animation/mdc-animation.test.js
```diff
@@ -0,0 +1,107 @@
...
+        style: {
+          'webkitTransform': {},
+        },
+      };
+    },
+  },
+});
+
+test('#getAnimationEventName does not prefix events and properties when not necessary', (t) => {
+  const windowObj = window;
```
traviskaufman: I'd honestly still mock out window here. Doing so will make these tests deterministic such that you'll always get the same result in any browser. For example, this test may fail on older versions of Safari.

traviskaufman:
Also FYI this is more like a `feat` in terms of the commit type, since we're adding a new feature and not doing any refactoring to `mdc-animation`.

packages/mdc-animation/index.js
amsheehan: Yes, I had been struggling emotionally with this while writing it. The fork to event name or property name is invisible to the user of this function and that seems wrong.

traviskaufman:
assigning @sgomes as additional reviewer in case I don't get to this in time

packages/mdc-animation/README.md
```diff
@@ -1,6 +1,6 @@
...
 
-mdc-animation is a sass / css library which provides variables, mixins, and classes for Material Design animation, based off of the [motion guidelines](https://material.google.com/motion/duration-easing.html#duration-easing-common-durations). Currently, it only covers easing curves.
+mdc-animation is a Sass / CSS / JavaScript library which provides a toolbelt for Material Design animation, based off of the [motion guidelines](https://material.google.com/motion/duration-easing.html#duration-easing-common-durations). Currently, it only covers easing curves, but will be expanded in the future. If you are interested in contributing to MDC Web's animation library, take a look at the [contributing page](https://github.com/material-components/material-components/blob/develop/CONTRIBUTING.md).
```
traviskaufman: What's the justification for this last part? How will it expand? May have been a conversation that I'm forgetting. Also, don't want to confuse people with things the material-motion org is doing.

test/unit/mdc-animation/mdc-animation.test.js
```diff
@@ -0,0 +1,120 @@
...
+        style: {
+          'webkitTransform': {},
+        },
+      };
+    },
+  },
+});
+
+test('#getCorrectEventName does not prefix events when not necessary', (t) => {
+  const windowObj = td.object(window);
```
traviskaufman: sorry, what I meant by this was to replace `window` altogether with a mocked out property. 


```js
const windowObj = td.object({
  document: {
    createElement: () => ({
      style: { transform: 'none' },
    }),
  },
});
```

sgomes:
@amsheehan @traviskaufman Let me know if my review is still needed; it looks like you guys have things covered.

sgomes:
Changes LGTM, thanks @amsheehan! 👍

PR #151, opened 2016-12-28
docs(checkbox): fix code sample

appsforartists:


sgomes:
Thanks, @appsforartists, nice catch!

I'm not entirely sure if this is the right fix, though, as looking at the demos it looks like the `div` with class `mdc-checkbox__mixedmark` should be inside the `div` with class `mdc-checkbox__background` (and if that's the case there are other examples to fix in this document). I'll leave it to @traviskaufman to review, as he's more familiar with checkbox code than me.

appsforartists:
I used the indentation to deduce where it was meant to go.  I trust TK knows better than I. :smiley:

appsforartists:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-form-field is nicely indented and has the right number of tags, so I've updated to match it.

packages/mdc-checkbox/README.md
```diff
@@ -52,6 +57,7 @@ easily position checkboxes and their labels.
         <path class="mdc-checkbox__checkmark__path" fill="none" stroke="white"
               d="M1.73,12.91 8.1,19.28 22.79,4.59"/>
       </svg>
+    </div>
```
traviskaufman: I think this part needs to be updated as well with what you did for the first example.

appsforartists:
Like that?

PR #163, opened 2017-01-07
Use default value instead of undefined check

AlexanderOtavka:
The constructor of component was using a check for undefined to specify a default value instead of the new ES2015 syntax.  Using the new syntax is clearer and more concise, since it shows the parameter is optional right where it is declared.


AlexanderOtavka:
Thanks!

PR #180, opened 2017-01-17
fix(textfield): Labels on single-line text fields support native browser autocomplete

### LINKED ISSUE #137: Single-line text fields do not work with auto-filled form fields
screen shot 2016-10-06 at 11 43 39 am

I tried detecting this via change events which didn't work. I also tried checking for `value` on init but that only worked for non-password inputs and was non-deterministic in terms of its timing.

Possible workarounds:
##### Style for `:autofill`

This is clearly the most elegant solution, but looks like it's [Chrome only](https://developer.mozilla.org/en-US/docs/Web/CSS/:-webkit-autofill)
##### Have elements poll to check if they're auto-filled

This will lead to the highest design fidelity, but is extremely inefficient and will incur a significant performance impact, especially on mobile browsers.
##### Tell users to set `autocomplete="off"`

This would in theory solve the problem. However, it is not only an inconvenience to the user, but it's [not guaranteed to work correctly](http://caniuse.com/#search=auto) for all types of input.
##### Tell users to only use JS-enhanced single-line inputs with form fields that won't be auto-completed

😢 

> Copied from https://github.com/google/material-design-lite/issues/4827

----

- [x] Summarize research on how most browsers handle auto-filling and present best solution
- [ ] Implement solution

amsheehan:
resolves #137

### DIFF
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -65,17 +65,11 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
   constructor(adapter = {}) {
     super(Object.assign(MDCTextfieldFoundation.defaultAdapter, adapter));
 
-    let receivedUserInput = false;
-
-    this.getKeyInteractionState = () => receivedUserInput;
-    this.setKeyInteractionState = (state) => {
-      receivedUserInput = state;
-    };
-
+    this.receivedUserInput_ = false;
     this.inputFocusHandler_ = () => this.activateFocus_();
     this.inputBlurHandler_ = () => this.deactivateFocus_();
     this.inputInputHandler_ = () => this.autoCompleteFocus_();
-    this.inputKeydownHandler_ = () => this.setKeyInteraction_();
+    this.inputKeydownHandler_ = () => this.receivedUserInput_ = true;
   }
 
   init() {
@@ -107,13 +101,9 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
   }
 
   autoCompleteFocus_() {
-    if (!this.getKeyInteractionState()) {
+    if (!this.receivedUserInput_) {
       this.activateFocus_();
     }
-  }
-
-  setKeyInteraction_() {
-    this.setKeyInteractionState(true);
   }
 
   showHelptext_() {
@@ -129,7 +119,7 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
     this.adapter_.removeClass(FOCUSED);
     if (!input.value) {
       this.adapter_.removeClassFromLabel(LABEL_FLOAT_ABOVE);
-      this.setKeyInteractionState(false);
+      this.receivedUserInput_ = false;
     }
     if (isValid) {
       this.adapter_.removeClass(INVALID);
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -172,5 +172,4 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
       });
     });
   }
-
 }
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -68,7 +68,18 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
     this.adapter_.addClass(MDCTextfieldFoundation.cssClasses.UPGRADED);
     this.adapter_.registerInputFocusHandler(this.inputFocusHandler_);
     this.adapter_.registerInputBlurHandler(this.inputBlurHandler_);
+
+    const input = this.getNativeInput_();
+
+    this.registerOnAutoComplete_(input)
+      .then((element) => {
+        console.log("element change: ", element);
+      });
+    // TODO: Remove this too, duh
+    this.timeoutThenFill()
+
   }
+  
 
   destroy() {
     this.adapter_.removeClass(MDCTextfieldFoundation.cssClasses.UPGRADED);
@@ -85,8 +96,7 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
 
   showHelptext_() {
     const {ARIA_HIDDEN} = MDCTextfieldFoundation.strings;
-    this.adapter_.removeHelptextAttr(ARIA_HIDDEN);
-  }
+    this.adapter_.removeHelptextAttr(ARIA_HIDDEN); }
 
   deactivateFocus_() {
     const {FOCUSED, INVALID, LABEL_FLOAT_ABOVE} = MDCTextfieldFoundation.cssClasses;
@@ -149,5 +159,31 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
       value: '',
       disabled: false,
     };
+  }
+
+  registerOnAutoComplete_(el) {
+    return new Promise((resolve, reject) => {
+      let hasKeyInteraction = false;
+
+      el.addEventListener('input', (e) => {
+        if (hasKeyInteraction === false) {
+          resolve(e)
+        }
+      });
+
+      el.addEventListener('keydown', (e) => {
+        hasKeyInteraction = true;
+      })
+    })
+  } 
+
+
+// TODO: remove this obvi
+  timeoutThenFill() {
+    const input = document.querySelector('input[name=email]');
+
+    setTimeout(()=>{
+      input.value = "hello"
+    }, 1000);
   }
 }
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -65,21 +65,16 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
   }
 
   init() {
+    const input = this.getNativeInput_();
+
     this.adapter_.addClass(MDCTextfieldFoundation.cssClasses.UPGRADED);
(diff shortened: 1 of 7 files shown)


packages/mdc-textfield/foundation.js
```diff
@@ -150,4 +156,20 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
       disabled: false,
     };
   }
+
+  registerAutoCompleteHandler_(input) {
+    return new Promise((resolve, reject) => {
```
traviskaufman: Did you happen to test this in IE11? Unfortunately, I don't think Promises are supported - http://caniuse.com/#feat=promises :( 

If that is the case, you could use a callback-based approach, e.g. `registerAutoCompleteHandler(onRegistered)` where `onRegistered` is your callback function.

packages/mdc-textfield/foundation.js
```diff
@@ -65,9 +65,15 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
   }
 
   init() {
+    const input = this.getNativeInput_();
+
     this.adapter_.addClass(MDCTextfieldFoundation.cssClasses.UPGRADED);
     this.adapter_.registerInputFocusHandler(this.inputFocusHandler_);
     this.adapter_.registerInputBlurHandler(this.inputBlurHandler_);
+    this.registerAutoCompleteHandler_(input)
+      .then((element) => {
```
traviskaufman: do you need to have `(element)` as an arg here? Or can you just do `(() => this.activateFocus_())` ?

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -116,6 +116,16 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
   }
 }
 
+// Move label when textfield gets autofilled in Chrome
+.mdc-textfield__input {
+  // stylelint-disable plugin/selector-bem-pattern
+  &:-webkit-autofill + .mdc-textfield__label {
```
traviskaufman: If you use `&:autofill` here, does autoprefixer output the correct CSS?

packages/mdc-textfield/foundation.js
```diff
@@ -150,4 +156,20 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
       disabled: false,
     };
   }
+
+  registerAutoCompleteHandler_(input) {
+    return new Promise((resolve, reject) => {
+      let hasKeyInteraction = false;
+
+      input.addEventListener('input', (e) => {
```
traviskaufman: Two things here:

1. We shouldn't be using `addEventListener` here directly on the input, since foundations should try to avoid using DOM API methods directly whenever possible. Instead, let's create a `registerInputKeyboardEventHandler(type, handler)` method similar to `registerInput{Focus,Blur}Handler` above that can accommodate for this, and then include it within the vanilla component. 

2. We need to clean up these handlers after they're used. I see this being done in two places: when the promise is resolved ( / callback is called, since we should change the method to be IE11-friendly), as well as in the foundation's `destroy()` method, in case the foundation is destroyed before the textfield is interacted with. In order to facilitate this let's make a complementary `deregisterInputKeyboardEventHandler(type, handler)` method, and then refactor the event handlers in this method as well as the `hasKeyInteraction` variable into instance properties. 

Since these are adapter changes, we also need to update the component to implement them, as well as the README to document them. Finally, since every adapter change is considered breaking, we have to remember to add a `BREAKING CHANGE:` footer to the final squash and merge commit.

packages/mdc-textfield/foundation.js
```diff
@@ -150,4 +156,20 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
...
+        if (!hasKeyInteraction) {
+          resolve(e);
```
traviskaufman: do we need to call resolve with the event object? It doesn't look like we use it.

packages/mdc-textfield/foundation.js
```diff
@@ -150,4 +156,20 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
...
+        }
+      });
+
+      input.addEventListener('keydown', function(e) {
```
traviskaufman: Do we need to have the event object passed into this function? It doesn't look like we use it.

Also, this should be a fat-arrow function.

```js
('keydown', () => {
  hasKeyInteraction = true;
});
```

packages/mdc-textfield/mdc-textfield.scss
Garbee: Nope, it doesn't prefix autofill pseudos. You can test in [an online playground](https://autoprefixer.github.io/).

packages/mdc-textfield/foundation.js
```diff
@@ -83,6 +100,16 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
     this.showHelptext_();
   }
 
+  autoCompleteFocus_(keyInteraction) {
```
traviskaufman: I feel like we don't need to pass in a `keyInteraction` argument here, since we already have `keyInteractionOccurred` as an instance property. So it can just be

```js
if (!this.keyInteractionOccurred) {
  this.activateFocus_();
}

File: packages/mdc-textfield/foundation.js [hunk 8]
```diff
@@ -83,6 +100,16 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
...
+    if (!keyInteraction) {
+      this.activateFocus_();
+    }
+  }
+
+  setKeyInteraction_() {
+    this.keyInteractionOccurred = true;
```
traviskaufman: What happens in the following scenario:

Say there are two text fields, and the following events happen:

1. User begins typing into first textfield
1. Browser auto-completes for both the first and second textfield
1. User clears second textfield
1. User goes back to first textfield and begins typing
1. Browser auto-completes both first and second textfield _again_

In this case, since `keyInteractionOccured` is already set to `true` from the first go-around, would the autocomplete functionality work as expected?

I think we might have to use some more detailed heuristics in order to get this to work correctly. For example, is the element currently focused might be a better indication of whether or not the input event was triggered by an auto-complete rather than a user interaction (since un-focused elements can only be `input` by means of an auto-complete, I would imagine).

packages/mdc-textfield/foundation.js
```diff
@@ -60,20 +64,33 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
 
   constructor(adapter = {}) {
     super(Object.assign(MDCTextfieldFoundation.defaultAdapter, adapter));
+
+    this.keyInteractionOccurred = false;
```
traviskaufman: This should probably be a private property

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -105,6 +107,19 @@ test('#init adds mdc-textfield--upgraded class', (t) => {
   t.end();
 });
 
+test('#setKeyInteraction sets keyInteractionOccurred to true', (t) => {
```
traviskaufman: We try not to test private methods, besides simply referencing `adapter_` in component tests. The reason is bc private methods are considered implementation details; a means to an end if you will. And what we really want is to be testing the ends. Specifically:

- Focused styles are applied when an `input` event is triggered on an element and no keydown was triggered beforehand (or, per the comment above,  some other means, such as if the element is not focused).

So it looks like that's just one single event test that should be implemented.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -126,6 +141,34 @@ test('#destroy deregisters blur handler', (t) => {
...
+});
+
+test('#destroy deregisters keydown handler', (t) => {
+  const {foundation, mockAdapter} = setupTest();
+  foundation.destroy();
+  t.doesNotThrow(() => td.verify(mockAdapter.deregisterInputKeydownHandler(td.matchers.isA(Function))));
+  t.end();
+});
+
+test('#destroyAutoCompleteHandlers deregisters input handler', (t) => {
```
traviskaufman: nit: The description here and below should be `#destroy deregisters ...`. We don't care about `#destroyAutoCompleteHandlers` since that's a private method and thus an implementation detail.

packages/mdc-textfield/foundation.js
amsheehan: I think you're right. What if we make `deactivateFocus()_` reset `keyInteractionOccurred` if the input value is an empty string. We already check for it in the method and I think it makes sense to add that to the list of concerns for `deactivateFocus()_`, and it solves for the above scenario.

test/unit/mdc-textfield/foundation.test.js
amsheehan: 👍

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -105,6 +107,37 @@ test('#init adds mdc-textfield--upgraded class', (t) => {
   t.end();
 });
 
+test('#on input focuses if input event occurs without any other events', (t) => {
```
traviskaufman: micronit: `'on'` instead of `'#on`, usually we reserve `#` for testing public methods

amsheehan: ahaaaa Thank you sir

ranjeet-choudhary:
@amsheehan is this pull not available in master i'm facing issue with auto complete

Issue #20, opened 2016-12-05
Add Ripples to components which need it


traviskaufman:
- [x] Regular buttons
- [x] FABs
- [ ] Checkboxes
- [x] Lists (List Items)

Pending #4467 

> Copied from https://github.com/google/material-design-lite/issues/4614

traviskaufman:
@sgomes @amsheehan perhaps we need some follow-up issues regarding incorporating ripples into components which use lists, such as drawers, selects, and menus

sgomes:
@traviskaufman Ripples in demos + readme entries pointing to the interactive lists section in the mdc-list readme?

traviskaufman:
closed via #182

PR #182, opened 2017-01-17
Feat: Add ripples to buttons and FABs

traviskaufman:
- Add ink ripple support to standard buttons
- Add ink ripple support to FABs

Note: This PR should be **rebased and merged**, not squashed and merged. This is because it contains two separate commits for FABs and Buttons.

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -189,5 +189,26 @@
       </section>
     </main>
     <script src="assets/material-components-web.js" charset="utf-8"></script>
+    <script>
+      // Because we load our CSS via webpack, we need to ensure that all of the correct styles
+      // are applied before ripples are attached. Otherwise, ripples may use the computed styles of
+      // elements before our CSS is applied, leading to improper UX.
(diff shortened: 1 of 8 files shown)


traviskaufman:
#20

demos/button.html
```diff
@@ -189,5 +189,26 @@
       </section>
     </main>
     <script src="assets/material-components-web.js" charset="utf-8"></script>
+    <script>
+      // Because we load our CSS via webpack, we need to ensure that all of the correct styles
```
sgomes: LGTM, but this is really unfortunate :-/ We may want to reconsider our demo infrastructure at some point, as the current one makes it hard to demonstrate server-side rendering / correct first renders, not to mention that it makes things much harder to understand for folks that prefer to look at demos rather than readmes.

traviskaufman: Agreed that I think we should reconsider our demo infrastructure to be more straightforward about how styles are loaded.

PR #206, opened 2017-01-24
feat(checkbox): Add ripples to checkboxes

### LINKED ISSUE #20: Add Ripples to components which need it

amsheehan:
Resolves #20

### DIFF
--- a/packages/mdc-checkbox/index.js
+++ b/packages/mdc-checkbox/index.js
@@ -40,7 +40,6 @@ export class MDCCheckbox extends MDCComponent {
   initRipple_() {
     const adapter = Object.assign(MDCRipple.createAdapter(this), {
       isUnbounded: () => true,
-      isSurfaceActive: () => false,
       registerInteractionHandler: (type, handler) => this.nativeCb_.addEventListener(type, handler),
       deregisterInteractionHandler: (type, handler) => this.nativeCb_.removeEventListener(type, handler),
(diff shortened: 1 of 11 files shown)


packages/mdc-checkbox/index.js
```diff
@@ -30,6 +32,34 @@ export class MDCCheckbox extends MDCComponent {
...
 
+  constructor(...args) {
+    super(...args);
+    this.ripple_ = this.initRipple_();
+  }
+
+  initRipple_() {
+    const adapter = Object.assign(MDCRipple.createAdapter(this), {
+      isUnbounded: () => true,
+      isSurfaceActive: () => false,
```
traviskaufman: how come? what happens when the checkbox is interacted with via keyboard?

traviskaufman:
Actually come to think of it, let's make a follow-up story for integrating the ripple into all of the framework examples. This way, we can correctly prioritize it and not block this feature.

packages/mdc-checkbox/index.js
amsheehan: Good call. The checkbox behaves as expected when tabbing through and pressing space and only space to actuate the checkbox. This can be removed.

Issue #170, opened 2017-01-11
Layout grid component


sgomes:
Spec reference: https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-grid

sgomes:
As an FYI, priority on this item has been bumped up significantly, and work on it will start soon.

sgomes:
This component has now landed in master. It will be made available on npm with the next scheduled release (probably early next week).

PR #228, opened 2017-01-30
feat(layout-grid): Add initial implementation of the layout grid.

sgomes:
Implements #170.

@traviskaufman PTAL, as you volunteered for the review :)

### DIFF
--- a/.stylelintrc.yaml
+++ b/.stylelintrc.yaml
@@ -245,6 +245,14 @@ rules:
     - severity: warning
   # Part of google's style guide
   number-leading-zero: never
+  # No unknown unit
+  unit-no-unknown:
+    - true
+    # Temporary rule to ignore CSS Grid `fr` unit while stylelint doesn't support it.
+    # See https://github.com/stylelint/stylelint/issues/2307
+    # TODO(sgomes): remove once fixed in stylelint
+    - ignoreUnits:
+        - fr
 
   # We use Harry Roberts' BEM dialect as our preferred way to format classes.
   # See: http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/
--- a/demos/layout-grid.html
+++ b/demos/layout-grid.html
@@ -1,0 +1,196 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+-->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>MDC Layout Grid Demo</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <script src="assets/material-components-web.css.js" charset="utf-8"></script>
+
+    <style>
+    .demo-grid {
+      background-color: #DDDDDD;
+      margin-bottom: 32px;
+    }
+
+    .demo-cell {
+      box-sizing: border-box;
+      background-color: #666666;
+      height: 200px;
+      padding: 8px;
+      color: white;
+
+      font-size: 1.5em;
+    }
+
+    .demo-grid.max-width {
+      max-width: 1600px;
+    }
+
+    .demo-grid.left-align {
+      margin-left: 0;
+      margin-right: auto;
+    }
+
+    .demo-grid-legend {
+      margin: 16px 0 8px 0;
+    }
+
+    .demo-ruler {
+      position: fixed;
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      bottom: 0;
+      left: 0;
+      height: 20px;
+      width: 100%;
+      margin: 0;
+      background: black;
+      color: white;
+    }
+
+    .demo-controls {
+      display: none;
+      margin-bottom: 8px;
+    }
+    .demo-warning::after {
+      content: "This browser does not support custom properties, so margins and gutters cannot be changed at runtime.";
+    }
+
+    @supports (--foo: green) {
+      .demo-controls {
+        display: block;
+      }
+
+      .demo-warning::after {
+        content: "";
+      }
+    }
+    </style>
+  </head>
+  <body>
+    <main class="mdc-typography">
+      <h1>MDC Layout Grid</h1>
+
+      <div class="demo-controls">
+        Margin:
+        <select id="margin">
+          <option value="8px">8px</option>
+          <option value="16px" selected>16px</option>
+          <option value="24px">24px</option>
+          <option value="40px">40px</option>
+        </select>
+
+        <br>
+
+        Gutter:
+        <select id="gutter">
+          <option value="8px">8px</option>
+          <option value="16px" selected>16px</option>
+          <option value="24px">24px</option>
+          <option value="40px">40px</option>
+        </select>
+      </div>
+      <div class="demo-warning"></div>
+
+
+      <div class="demo-grid-legend">Grid of 1 column wide items</div>
+      <div class="demo-grid mdc-layout-grid">
+        <div class="demo-cell mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+        <div class="demo-cell mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+        <div class="demo-cell mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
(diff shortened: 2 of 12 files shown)


demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
+<!DOCTYPE html>
```
traviskaufman: General comment: Changing margin / gutter widths don't work in IE or Edge since they don't support custom properties. We should probably either hide or disable them in those browsers and add a note saying why they're disabled, similar to what we do within the ripple demo.

Furthermore, the demo page was broken in Edge on Windows 10 when I tested it in my VM:

layout-grid-edge

.stylelintrc.yaml
```diff
@@ -245,6 +245,11 @@ rules:
     - severity: warning
   # Part of google's style guide
   number-leading-zero: never
+  # No unknown units
```
sgomes: Done.

demos/layout-grid.html
sgomes: It looks like support broke when I added the CSS Grid PE, because Edge supports the old style CSS Grid, and autoprefixer is converting to that. Added the `grid: false` option to autoprefixer to prevent that. There is a [bug in autoprefixer](https://github.com/postcss/autoprefixer/issues/778) that requires an extra workaround for now, though.

Also, added the check for custom properties to hide the controls and display a warning. Unfortunately, MS Edge misreports its support for custom properties, and I can't find a reliable way of feature-detecting it, so I added a warning that gets shown if the user's browser claims it supports custom properties.

demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>MDC Layout Grid Demo</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <script src="assets/material-components-web.css.js" charset="utf-8"></script>
+
+    <style>
+    .mdc-layout-grid {
```
sgomes: Done.

demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
...
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-1">1</div>
+      </div>
+
+      <div class="grid-legend">Grid of 4 column wide items</div>
+      <div class="mdc-layout-grid">
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-4 mdc-layout-grid__cell--order-1">5</div>
```
sgomes: Oops, that was some random testing that got committed. Not supposed to be there. I'll leave the ordering stuff for the readme.

demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
...
+
+      <div class="grid-legend">Grid with max width and left alignment</div>
+      <div class="mdc-layout-grid max-width left-align">
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-4"></div>
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-4"></div>
+        <div class="mdc-layout-grid__cell mdc-layout-grid__cell--span-4"></div>
+      </div>
+
+      <div class="ruler"><div id="current"></div></div>
+        </main>
```
sgomes: Fixed.

demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
...
+
+    <script>
+      (function(global) {
+        'use strict';
+        var marginRadio = document.querySelector('#margin');
```
sgomes: I'm assuming you meant other names, since those are the ones they have?

demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
...
+        marginRadio.addEventListener('change', function() {
+          document.documentElement.style.setProperty('--mdc-layout-grid-margin', marginRadio.value);
+        }.bind(this));
```
sgomes: Old code, dropped.

demos/layout-grid.html
```diff
@@ -0,0 +1,178 @@
...
+
+        var gutterRadio = document.querySelector('#gutter');
+        gutterRadio.addEventListener('change', function() {
+          document.documentElement.style.setProperty('--mdc-layout-grid-gutter', gutterRadio.value);
+        }.bind(this));
```
sgomes: Old code, dropped.

packages/mdc-layout-grid/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@material/layout-grid",
+  "version": "0.1.0",
+  "description": "The Material Components for the web layout grid component",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/material-components/material-components-web.git"
+  }
```
sgomes: Done.

packages/mdc-layout-grid/README.md
```diff
@@ -0,0 +1,178 @@
+# mdc-layout-grid
+
+MDC layout grid is a CSS-only component that implements the
+[Material Design layout grid guidelines](https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-grid),
+and makes them available to developers as CSS classes and Sass mixins.
```
sgomes: Added.

packages/mdc-layout-grid/README.md
```diff
@@ -0,0 +1,178 @@
...
+npm install --save @material/layout-grid
+```
+
+## The layout grid
+
+A grid consists of a group of cells, which get positioned in sequence according to a predefined number of columns.
+Cells specify how many columns to span (the default being 4), and get placed side by side while there is room. When
+there isn't enough room for a cell, it gets moved to the beginning of the next row, and placement continues as usual.
+
+The grid has 12 columns in desktop mode, 8 columns in tablet mode, and 4 columns in phone mode. Column widths are
```
sgomes: Added.

packages/mdc-layout-grid/README.md
```diff
@@ -0,0 +1,178 @@
...
+variable; it's the margins and gutters that are fixed, with columns taking up the remainder of the space.
```
sgomes: Fixed.

packages/mdc-layout-grid/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+The default size for both the margins and gutters in MDC layout grid is 16px.
+
+You can change the margins and gutters for a grid using the `--mdc-layout-grid-margin` and `--mdc-layout-grid-gutter`
+custom properties, respectively. This requires support for CSS custom properties on the end-user's browser.
+
+The Material Design spec recommends 8px, 16px, 24px or 40px as the sizes to choose from, but MDC layout grid doesn't impose any restrictions.
+
+```css
+.my-grid {
```
sgomes: I consider that very different usage, covered by the Sass mixins section. Unless you're talking about changing the global Sass variables we use internally, which I would consider unsupported usage.

packages/mdc-layout-grid/README.md
```diff
@@ -0,0 +1,178 @@
...
+  --mdc-layout-grid-margin: 40px;
+  --mdc-layout-grid-gutter: 16px;
+}
+```
+
+> Note: Due to the implementation of MDC layout grid, it's not possible to use a margin smaller than half of the size
+of the gutter, in most browsers. As support for [CSS Grid](https://drafts.csswg.org/css-grid/) lands in browsers, this
```
sgomes: Good point. Changed.

packages/mdc-layout-grid/_variables.scss
```diff
@@ -0,0 +1,30 @@
...
+  phone: 0px
+);
+
+$mdc-layout-grid-columns: (
+  desktop: 12,
+  tablet: 8,
+  phone: 4
+);
+
+$mdc-layout-grid-default-column-span: 4;
```
sgomes: I'll add the `!default`, but I prefer not to document this, as it seems like a very error-prone / hard-to-debug-later way of doing things.

sgomes:
@traviskaufman All comments addressed, PTAL!

demos/layout-grid.html
sgomes: Actually, removing the warning I mentioned, as the problem with MS Edge misreporting custom property support seems to be exclusive to the preview version of Edge.

sgomes: The autoprefixer bug has already been fixed and a new version released (!), so I've removed the workaround and bumped up the required autoprefixer version in `package.json`.

demos/layout-grid.html
traviskaufman: ah sorry -_- `marginSelect` and `gutterSelect`?

traviskaufman:
💯 looks great!

demos/layout-grid.html
sgomes: D'oh, of course! Fixed.

sgomes:
Thanks! Waiting for green build and then merge \o/

gurmukhp:
This looks awesome and would be super useful for an app we're building right now!

raisaveuc:
Thanks for merging the new layout grid! I'd like to use this in a project--could you please add this to npm so I can add the new package?

traviskaufman:
We'll be releasing it on Monday!

Issue #211, opened 2017-01-24
Add ripple integration to checkboxes in framework examples


traviskaufman:
Once we land #206, we need to update all of the framework examples which use checkbox to integrate ripples into it. This involves:

* building a ripple component for that framework
* integrating the ripple component into the checkboxes

- [x] React
- [ ] Angular2
- [ ] VueJS
- [ ] Aurelia

traviskaufman:
@petejohanson @Thanood Any chance you'd be interested in working on the VueJS and Angular components to this issue, respectively? :)

petejohanson:
@traviskaufman I would prefer not to have duplicate work between petejohanson/vue-mdc and the framework examples, at this point. It's on my list to implement for checkboxes/radio buttons in vue-mdc, once I complete that work, I will see if it can easily be brought in here for reference.

tranrate:
@traviskaufman I am working on tranrate/aurelia-mdc. I've got checkboxes working but not yet got #206 and ripples working with checkboxes. As soon as I've got an initial version of aurelia-mdc I can bring it here for reference if that helps. @Thanood used the "Simple Approach" wrapping components. Aurelia-mdc will use foundations and adapters. Do you want to stick with the "Simple Approach" for your examples?

traviskaufman:
@tranrate awesome! 

Since the Simple Approach method is the one used throughout our demos, if you do switch to the Foundations / Adapters approach I'd probably ask you to widen the scope of work to convert the demo entirely. Alternatively, you could remove the other components and just display the checkbox using that pattern. But it would indeed be strange to mix the two together, since users may not know what the best approach for Aurelia would be.

tranrate:
@traviskaufman Once I've got the existing components working, I am probably going to convert your demo entirely as that will be a good initial test that everything works. For your examples it would be best to stick with the Simple Approach and then put links to projects like aurelia-mdc for people who don't want to reinvent the wheel. I will try and make time to check @Thanood's examples and see if they need much amendment to work with ripples.

PR #233, opened 2017-01-31
feat(framework-examples): Add ripple support to React checkbox example

codesuki:
This will only work if the newest checkbox.css is available. So maybe should be merged after a new checkbox version is released (0.1.3?).

Also `getMatchesProperty` is not exported from Ripple. I might send another pull request since it looks very useful for implementing new components.


codesuki:
I signed it!

framework-examples/react/src/Checkbox.js
```diff
@@ -101,6 +110,45 @@ export default class Checkbox extends PureComponent {
     isAttachedToDOM: () => Boolean(this.refs.nativeCb),
   });
 
+  // For browser compatibility we extend the default adapter which checks for css variable support.
+  rippleFoundation = new MDCRippleFoundation(Object.assign(MDCRipple.createAdapter(this), {
+    isUnbounded: () => true,
+    isSurfaceActive: () => this.refs.root[MATCHES](':active'),
```
traviskaufman: what you actually want to check is that the _input_ element is active, rather than the root. See https://github.com/material-components/material-components-web/pull/241

codesuki:
Thanks for the feedback! I'll have a look in about 40 minutes. I'll also reply to the other pullreq seems I messed something up there.

framework-examples/react/src/Checkbox.js
codesuki: Makes sense! I was wondering why it's the root. I was using the native js version that you fixed as my guide. Thanks for clearing that up.

codesuki:
Fixed!

I have a question about ripple. Maybe you can clear something up for me.
In https://github.com/material-components/material-components-web/issues/211 you are proposing to build a ripple component for each framework.
In general I agree with that and since I started to work on a [React wrapper](https://github.com/codesuki/react-material-components/tree/initial) for this project I wanted to implement it correctly, but..

I figured out that the classes and CSS need to be set on the DOM element carrying the specific class like `mdc-button`, `mdc-checkbox`, or else it won't work. Which makes sense since they have support for the ripple effect in their CSS.
Maybe I just didn't have the right idea yet but that resulted in passing references which made the code really ugly.
```
<Ripple><Button/></Ripple>
```
Here ripple needs a ref of button and also needs to be able to set buttons `className`, but button is only available in `this.props.children` to ripple... 
Anyway I don't want to go deeper into this here.

I tried using `ripple-surface` which didn't work at all.

Question 1:
Do you have an idea for a clean implementation of a <Ripple> react component?

Question 2:
Am I right saying that using `ripple-surface` would result in some generic ripple effect and we should use the proper `mdc-button` etc. classes to get the real effect?

For now I found it best to just implement the ripple inside the specific class like here:
https://github.com/codesuki/react-material-components/blob/initial/src/button/button.jsx

What do you think?

traviskaufman:
@codesuki regarding your question,

You're totally right in that the ripple should be a _part_ of a component, not something completely different. E.g. with buttons, a button component would manage a ripple as part of the component itself. This is how we approach it within MDC-Web, although for our components, we simply instruct users to attach a ripple to buttons. The reason for this is that our vanilla buttons require no javascript, so we didn't want to introduce a wrapper component simply to facilitate a ripple

For frameworks like React, JS is _always_ required no matter what the component since it takes care of rendering the component DOM for the end user. In this case, you'll definitely want to manage the ripple within those components, as you've suggested. Perhaps using a higher-order function that creates a component with ripple capabilities would be a sound strategy?

codesuki:
Thanks for clearing that up! Glad I am on the same page.
I was thinking about a higher order function.. maybe if the need arises I'll go with that.

PR #234, opened 2017-01-31
fix(ripple): Export getMatchesProperty

codesuki:
This will be useful to keep browser compatibility for people implementing components using ripple.

### DIFF
--- a/packages/mdc-ripple/index.js
+++ b/packages/mdc-ripple/index.js
@@ -20,7 +20,7 @@ import {supportsCssVariables, getMatchesProperty} from './util';
 
 const MATCHES = getMatchesProperty(HTMLElement.prototype);
 
-export {MDCRippleFoundation};
+export {MDCRippleFoundation, getMatchesProperty};
 
 export class MDCRipple extends MDCComponent {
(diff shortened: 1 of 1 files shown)


traviskaufman:
Actually, it looks like we already _do_ export this within our `util.js` class (sorry, been a minute since I looked at that code specifically). Could you not just import `@material/ripple/util` and use it that way?

codesuki:
You are right. It is already exported. I got confused because in the react example I used
```import {MDCRipple, MDCRippleFoundation} from '@material/ripple/dist/mdc.ripple';```.

I tried it again and reproduced the error I had.

In line with the current sample I used the compiled `.js` files from `dist/` which do not contain `getMatchesProperty`. 
If I try to import from `@material/ripple` I get the following compile error:
```
Uncaught SyntaxError: Unexpected token import
```
ripple/index.js
```js
import {MDCComponent} from '@material/base'; <--- 
```

So maybe it's a webpack configuration issue.

In my private project I actually am using this successfully
```import { MDCRipple, MDCRippleFoundation } from '@material/ripple';
import { getMatchesProperty } from '@material/ripple/util';```

PR #236, opened 2017-01-31
docs(demos): Adds readme to demos folder

### LINKED ISSUE #131: Add README to demos folder
There's a demos folder, with lots of useful-looking html, but I'm unsure how we're supposed to view it? I've tried cloning and simply opening the html files, but am getting 

`assets/material-components-web.css.js Failed to load resource: net::ERR_FILE_NOT_FOUND`

There's no .md file in the demos folder, no `assets` folder anywhere I could find, and no package.json so running npm install doesn't do much. Is it explained anywhere how to actually view these demos?

amsheehan:
resolves #131

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -21,6 +21,6 @@ Then run the demo server:
 npm run dev
 ```
 
-This will spin up a server that can be accessed at 127.0.0.1:8080
+This will spin up a server that can be accessed at localhost:8080
 
 Pop open any major browser and enjoy!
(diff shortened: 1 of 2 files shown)


demos/README.md
```diff
@@ -0,0 +1,26 @@
...
+```
+npm i
+```
+
+Then run the demo server:
+```
+npm run dev
+```
+
+This will spin up a server that can be accessed at 127.0.0.1:8080
```
sgomes: Nit: I'd use `localhost` instead of `127.0.0.1`.

PR #237, opened 2017-02-01
feat(npm-keywords): Add keywords to components package.json files

### LINKED ISSUE #102: Add keywords to all npm packages
Currently, none of our npm packages have any keywords, which help discovery on the registry. We should look into having some standard keywords, such as `material`, `material design`, `material components`. Additionally, we could have component-specific keywords such as `button`, `checkbox`, etc.

amsheehan:
resolves #102


traviskaufman:
What if for all of them we just used:

- `material-design`
- `material-components`
- `<COMPONENT_NAME>`

E.g. for mdc-button it would be `material-design`, `material-components`, `buttons`, for mdc-animation it would be `material-design`, `material-components`, `animation`. 

We should also probably update the MDC-Web-specific part of the Authoring Components guide to reflect this

amsheehan:
I've seen some packages that add the technology they use in the package, but that might just add  noise. I like your approach.

traviskaufman:
@material-components/mdc-web anyone else want to weigh in here re npm keywords?

package.json
```diff
@@ -2,6 +2,10 @@
   "private": true,
   "description": "Material Components Web",
   "license": "Apache-2.0",
+  "keywords": [
```
traviskaufman: You probably don't need these keywords in our main repo package since it's private and won't ever be published to npm

packages/mdc-button/package.json
```diff
@@ -3,6 +3,11 @@
   "description": "The Material Components for the web button component",
   "version": "0.2.1",
   "license": "Apache 2.0",
+  "keywords": [
+    "material components",
+    "material design",
+    "buttons"
```
traviskaufman: "button"

sgomes:
This looks good. I just have two comments:
- For component types, we should probably choose singular vs plural and stick to it. I see `buttons` and `forms` but `menu` and `list`.
- For components that have different names in the Material Design spec and on the web platform we should probably use both.

amsheehan:
@sgomes do you have an example where the name in the spec doesn't match a component we have in the wild? I looks like we have pretty good parity for names so far.

sgomes:
@amsheehan Possibly none at the moment, since the one that comes to mind, `select`. isn't a component in the spec per se but rather a more specific implementation of the menu.

PR #249, opened 2017-02-03
feat(ripple): Implement improved origin point rules

### LINKED ISSUE #188: Implement improved origin point rules for ripple
- Ripple is centered for _“small”_ objects, as well as for non-pointer-based interactions (note that _small_ needs to be defined for a concrete value here).
- Ripple starts at pointer center of contact and “slides” to middle using pre-determined motion guidelines to do so.

cristobalchao:
Resolves #188


packages/mdc-ripple/_variables.scss
```diff
@@ -1,17 +1,17 @@
-//
```
traviskaufman: Do we need this file anymore?

PR #262, opened 2017-02-06
Fix missing whitespace in section demo

0xdec:
Missing space before the label attribute was causing the browser to complain


0xdec:
Sorry, changed my GitHub username since I last signed it, fixed now!

Issue #25, opened 2016-12-05
Implement Slider Component


traviskaufman:
- [x] [Continuous Slider](https://www.google.com/design/spec/components/sliders.html#sliders-continuous-slider)
- [ ] [Discrete Slider](https://www.google.com/design/spec/components/sliders.html#sliders-discrete-slider)

@sgomes already doing some work on this. 
### Notes

Step values need to be able to be quantized to decimal places, e.g. `min = 0, max = 5, step = .2`. Possible algorithm: Take the raw value, divide it by `step`, round that number, and then multiply the original step value by the rounded number. E.g.

``` js
function quantize(val, min, max, step) {
  const numSteps = Math.round(val / step);
  const quantizedVal = numSteps * step;
  return Math.max(min, Math.min(max, quantizedVal));
}

const min = 0, max = 5, step = .2;
quantize(3.56, min, max, step); // 3.6
quantize(2.148692, min, max, step) // 2.2
quantize(1.061733, min, max, step) // 1
```

> Copied from https://github.com/google/material-design-lite/issues/4494

chriscox:
Is this slider currently in the roadmap? Any ETA?

traviskaufman:
@chriscox take a look in our [Pivotal Tracker](https://www.pivotaltracker.com/n/projects/1664011)!

traviskaufman:
cc @pgbross this is where we're tracking work for sliders. 

If you're interested in migrating your work from #487 over to our implementation, we usually first want to see a visual prototype (which we actually already have), as well as an [eng outline doc](https://docs.google.com/document/d/1Kybm7XJDTy0KUcuMaw5bzirQNBsDqCPCae8U_ag_a1k/edit) detailing how you plan on building the component. Would this be something you're interested in looking into? 

If so, let us know _before_ starting any work on the eng outline and I'll port over my visual prototype to codepen :). I worked pretty closely alongside the design team on the visual prototype so it should take away a lot of the ambiguity around the subtle design interactions that sliders require. Thanks!

pgbross:
@traviskaufman thank you for your feedback on #487. 

I am interested in looking into sliders further and am willing to have a go based on your visual prototype and beginning with the engineering outline document. Are there any eng outline documents for existing MDC-Web components or a worked example that could be used to see the style and level of detail expected in the different sections?

I will wait until you have time to port your visual prototype to codepen and look forward to seeing it in due course.

traviskaufman:
[...]

PR #268, opened 2017-02-07
Added a range slider

MarcMcIntosh:
Had to use slightly different implementations for gecko, webkit and edge.

I've tested on chromium and firefox, firefox looks the best as chromium has a very limited selection of  pseudo elements for styling ranged inputs.  

Kind of working blind with edge / internet explorer,  
if any one could check that out it would be great :)


MarcMcIntosh:
who's robot?

packages/mdc-range-slider/mdc-range-slider.scss
```diff
@@ -0,0 +1,243 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */
+@import "@material/elevation/mixins";
+@import "@material/theme/mixins";
+@import "variables";
+@import "mixins";
+
+[type="range"].mdc-range-slider {
```
Garbee: We shouldn't really need the type here. Not only does it break BEM (which you also should add a linting marker for) but it isn't really necessary. The docs will instruct people to use this against a range type only and that should be sufficient. This is increasing specificity which makes it more difficult for developers to modify styling if they wish to.

Garbee:
As I recall from MDL 1.x feedback we were planning on not relying on the range input type with MCW due to some issues getting the MD styling just right. 

@sgomes Do you recall what those issues were? If not I can go search the issue tracker.

packages/mdc-range-slider/mdc-range-slider.scss
MarcMcIntosh: good point,

MarcMcIntosh:
@Garbee is it an issue with webkit having only one pseudo element  for the whole track?

sgomes:
Thanks, @MarcMcIntosh, this is a lot of work!

As @Garbee was mentioning though, I fear we'll quickly run into issues if we try to implement things on top of native controls. Here are some issues I remember off the top of my head:
- Some aspects of the controls aren't animatable in some browsers, so you get sudden jumps in size when scaling the thumb, for example
- The scaling thumb causes touch issues in some browsers, presumably because they get confused about the touch area size changing
- There are differences in behaviour between browsers with the same slider implementation (e.g., Chrome and Safari, which both use `webkit` prefixes
- There are differences in behaviour between browser versions
- Things get very difficult or impossible when going beyond the simple continuous slider

As such, we had decided against using the native control in MDC-Web

MarcMcIntosh:
I was making one figured may as well add it :) didn't even think about touch issues... , ran into a few of those inconsistencies. It probally have been easier to just to write inline an svg with min and max values applied to the view box.

traviskaufman:
Hey @MarcMcIntosh 

First off, thanks for taking the time to put all of this together, I really appreciate your willingness to contribute to our codebase!

That being said, all of @sgomes' points are valid. Besides just that, MD sliders require some very intricate treatments that are not present in the spec, and that most (if not all?) existing MD implementations haven't nailed down properly. Thus, I'm probably going to have to close this PR for now, since I don't think we'll be able to land it in our codebase.

That being said, we're tracking the slider issue at #25. If this is a component you're interested in undertaking, we'd be happy to help you out but ask that you follow the processes outlined within our [authoring components doc](../tree/master/docs/authoring-components.md).

Finally, if you'd rather release this as a standalone component built on top of MDC-Web, that's totally doable as well! You can simply depend on `@material/typography` and `@material/theme` and publish the component outside of MDC-Web proper, as a very lightweight MD wrapper around a native range slider.

MarcMcIntosh:
@traviskaufman I was using the SCSS code base anyway and need a slider for a small project and figured may as-well make one. I've got a couple of projects I'd like to finish before making additional commitments.

Shouldn't take to long though so I've taken a look at the react Checkbox.js in the framework-examples to get an idea of how javascript components are intended to be used.... Might take a while to get my head around guessing  user provides the elements id's and class names... mdc attaches listeners and classname modifiers to them?

PR #269, opened 2017-02-07
chore(stylelint): Remove `fr` unit workaround

### LINKED ISSUE #256: Remove `fr` unit workaround from stylelint config
stylelint 7.8.0 has been released, which adds the `fr` unit to the list of known units, removing the need for [our current workaround](https://github.com/material-components/material-components-web/blob/master/.stylelintrc.yaml#L251).

- [ ] update `package.json` to require 7.8.0
- [ ] remove the workaround from `.stylelintrc.yaml`
- [ ] ensure we don't run into any of the deprecations announced in the [7.8.0 release](https://github.com/stylelint/stylelint/releases/tag/7.8.0), refactoring if we do

BBosman:
- Updated `stylelint` to `7.8.0`.
- Removed the `fr` unit workaround from the stylelint config.
- Updated `stylelint-config-standard` to `16.0.0` as the old version referenced now deprecated options (specifically: `rule-non-nested-empty-line-before`, `declaration-block-no-ignored-properties`, `media-feature-no-missing-punctuation` and the `blockless-group` option of `at-rule-empty-line-before`)
- Introduced the `stylelint-order` plugin, to mimic the old `declaration-block-properties-order` behavior which was deprecated.
- Dropped `selector-root-no-composition` from the stylelint config as it was deprecated and there is no replacement. They suggest that rule could be done by a plugin, but nobody made one so far afaik.
- Fixed a few linting issues which where flagged as a result of the updated `stylelint-config-standard`.

Fixes #256.


sgomes:
Thanks so much for taking care of this, @BBosman ! 🎉  

@traviskaufman I think you worked on the original version of the rules, so poking you in case there's something in particular you'd like a look at. LGTM from my end.

traviskaufman:
LGTM! Ship it! 🚢

PR #274, opened 2017-02-07
fix(drawer): Fix Temporary Drawer on IE11

### LINKED ISSUE #202: Temporary Drawer not working on IE11

Tested on 0.2.0 and current master.

IE11

### What OS are you using?

Windows 10

1. Run the demo server
2. Go to http://127.0.0.1:8080/drawer/temporary-drawer.html
3. Observe the drawer should be hidden by default, but it is not

The temporary drawer should be hidden by default.

The temporary drawer is visible by default, and not even clickable. Not working at all.

Only happening on IE11, it works on Edge. Unfortunately I don't have a Windows Phone to test on.

BBosman:
Fixes #202.

- Verified the change fixes the temporary drawer on IE11.
- Verified the change doesn't affect behavior in other browsers (tested on Chrome 55 only though).
- Arbitrarily chose to add 7% (which is derived from 20px rounded up on a 320px viewport).

### DIFF
--- a/packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
+++ b/packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
@@ -67,6 +67,7 @@
     width: calc(100% - 56px);
     max-width: 280px;
     height: 100%;
+    transform: translateX(-107%);
     transform: translateX(calc(-100% - 20px));
     will-change: transform;
     box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


sgomes:
Woohoo! :D This is exactly the approach I was thinking of taking.

Thanks for taking the time to implement it, test it out, and submit it as a PR. I tested it up in Safari, Firefox and Edge as well, and there doesn't seem to be any breakage.

PR #281, opened 2017-02-09
feat(ripple): Add programmatic ripple activation/deactivation.

sgomes:
First stage in fixing #9

### DIFF
--- a/packages/mdc-ripple/README.md
+++ b/packages/mdc-ripple/README.md
@@ -182,6 +182,23 @@ ripple.
 MDCRipple.attachTo(document.querySelector('.surface'));
 ```
 
+### Ripple API
+
+The component allows for programmatic activation / deactivation of the ripple, for interdependent interaction between
+components. This is used for making form field labels trigger the ripples in their corresponding input elements, for
+example.
+
+#### MDCRipple.activate()
+
+Triggers an activation of the ripple (the first stage, which happens when the ripple surface is engaged via interaction,
+such as a `mousedown` or a `pointerdown` event). It expands from the center.
+
+#### MDCRipple.deactivate()
+
+Triggers a deactivation of the ripple (the second stage, which happens when the ripple surface is engaged via
+interaction, such as a `mouseup` or a `pointerup` event). It expands from the center.
+
+
 ### Unbounded Ripples
 
 If you'd like to use _unbounded_ ripples, such as those used for checkboxes and radio buttons, you
--- a/packages/mdc-ripple/foundation.js
+++ b/packages/mdc-ripple/foundation.js
@@ -145,8 +145,9 @@ export default class MDCRippleFoundation extends MDCFoundation {
     }
 
     activationState.isActivated = true;
+    activationState.isProgrammatic = e === null;
     activationState.activationEvent = e;
-    activationState.wasActivatedByPointer = (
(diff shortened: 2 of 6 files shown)


packages/mdc-ripple/foundation.js
```diff
@@ -145,8 +145,9 @@ export default class MDCRippleFoundation extends MDCFoundation {
     }
 
     activationState.isActivated = true;
+    // If e is null, this is a programmatic activation.
     activationState.activationEvent = e;
-    activationState.wasActivatedByPointer = (
+    activationState.wasActivatedByPointer = e === null ? false : (
```
traviskaufman: What if instead of checking for `e === null` we added a boolean property `isProgrammatic` to the activationState object, and used that instead? e.g.

```js
activationState.isProgrammatic = e === null;
// ...
const isPointerEvent = !this.activationState_.isProgrammatic && /* ... */;
```

packages/mdc-ripple/README.md
```diff
@@ -182,6 +182,23 @@ ripple.
 MDCRipple.attachTo(document.querySelector('.surface'));
 ```
 
+### Ripple API
+
+The component allows for programmatic activation / deactivation of the ripple, for complex interaction between deeply
```
traviskaufman: could we say "for interdependent interaction between components" instead of "for complex interaction between deeply tied components"?

packages/mdc-ripple/README.md
```diff
@@ -182,6 +182,23 @@ ripple.
...
+tied components. This is used for making form field labels trigger the ripples in their corresponding input elements,
+for example.
+
+#### MDCRipple.activate()
+
+Triggers an activation of the ripple (the first stage, which happens on a down event). It is represented visually as if
```
traviskaufman: could we say "which happens when the ripple surface is engaged via interaction, such as a `mousedown` or a `pointerdown` event instead of "which happens on a down event"? Also I'm not sure we need to talk about the visual representation as coming from a keyboard event; we can just say it "expands from the center" or something a bit less implementation-specific instead.

packages/mdc-ripple/README.md
```diff
@@ -182,6 +182,23 @@ ripple.
...
+coming from a keyboard event.
+
+#### MDCRipple.deactivate()
+
+Triggers a deactivation of the ripple (the second stage, which happens on an up event). It is represented visually as if
```
traviskaufman: could we say "which happens when the ripple surface is engaged via interaction, such as a `mouseup` or a `pointerup` event instead of "which happens on an up event"?

packages/mdc-ripple/foundation.js
sgomes: Good idea. Done.

packages/mdc-ripple/README.md
sgomes: Done.

packages/mdc-ripple/README.md
sgomes: Done.

packages/mdc-ripple/README.md
sgomes: Done.

sgomes:
Thanks, @traviskaufman! Will merge when build goes green.

Issue #194, opened 2017-01-19
Implement hover state for ripple


traviskaufman:
Design has new motion patterns for ink ripples on hover for flat surfaces. Thus,

- [ ] Ripples need to be configurable such that they can either be shown or not shown when a surface is hovered over (both for css-only and JS ripple)
- [ ] The hover UX needs to be implemented
- [ ] Components with elevation that use ripples need to not show an ink wash on hover, but change elevation instead

traviskaufman:
Q's needing answered:

- Which components need do _not_ receive ink wash on hover
- For components that need an elevation change, is that duration / curve always constant or does it depend on the shape and dimension of the surface?

kfranqueiro:
Current design direction no longer calls for ripple on hover under any circumstances, so I'm closing this as obsolete.

PR #282, opened 2017-02-09
Feat/implement improved graceful degradation ripple

### LINKED ISSUE #189: Implement improved graceful degradation for ripple
Implement new UX for gracefully degraded CSS-only interaction state effects

found in motion docs under `exploration->touchresponse->exploration-touchresponse-motionstudies-02-page31`

* Implement graceful degradation as part of `mdc-ripple-base` mixin
* Change docs to say that components no longer need to implement graceful degradation themselves
* Update docs for ripple talk about how graceful degradation is done.
* _(Separate commit/PR)_ Update components to remove custom graceful degradation styling

cristobalchao:
Resolves #189

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -308,11 +308,9 @@
           }
         }, 250);
         function init() {
-          var btns = document.querySelectorAll('.mdc-button');
+          var btns = document.querySelectorAll('.mdc-button:not([data-demo-no-js]');
           for (var i = 0, btn; btn = btns[i]; i++) {
-            if (!btn.hasAttribute('data-demo-no-js')) {
-              mdc.ripple.MDCRipple.attachTo(btn);
-            }
+            mdc.ripple.MDCRipple.attachTo(btn);
           }
         }
       })();
--- a/demos/button.html
+++ b/demos/button.html
@@ -75,38 +75,38 @@
           </div>
         </fieldset>
         <fieldset>
-          <legend>Graceful degraded buttons</legend>
-          <button class="mdc-button" no-js>
+          <legend>Buttons - CSS Only</legend>
+          <button class="mdc-button" data-demo-no-js>
             Default
           </button>
-          <button class="mdc-button mdc-button--raised" no-js>
+          <button class="mdc-button mdc-button--raised" data-demo-no-js>
             Raised
           </button>
-          <button class="mdc-button mdc-button--dense" no-js>
+          <button class="mdc-button mdc-button--dense" data-demo-no-js>
             Dense Default
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--dense" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--dense" data-demo-no-js>
             Dense Raised
           </button>
-          <button class="mdc-button mdc-button--compact" no-js>
+          <button class="mdc-button mdc-button--compact" data-demo-no-js>
             Compact
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--compact" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--compact" data-demo-no-js>
             Compact Raised
           </button>
-          <button class="mdc-button mdc-button--primary" no-js>
+          <button class="mdc-button mdc-button--primary" data-demo-no-js>
             Default with Primary
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--primary" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--primary" data-demo-no-js>
             Raised with Primary
           </button>
-          <button class="mdc-button mdc-button--accent" no-js>
+          <button class="mdc-button mdc-button--accent" data-demo-no-js>
             Default with Accent
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--accent" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--accent" data-demo-no-js>
             Raised with Accent
           </button>
-          <div class="mdc-button mdc-button--raised" tabindex="0" role="button" no-js>
+          <div class="mdc-button mdc-button--raised" tabindex="0" role="button" data-demo-no-js>
             Div
           </div>
         </fieldset>
@@ -220,38 +220,38 @@
           </div>
         </fieldset>
         <fieldset>
-          <legend>Graceful degraded buttons</legend>
-          <button class="mdc-button" no-js>
+          <legend>Buttons - CSS Only</legend>
+          <button class="mdc-button" data-demo-no-js>
             Default
           </button>
-          <button class="mdc-button mdc-button--raised" no-js>
+          <button class="mdc-button mdc-button--raised" data-demo-no-js>
             Raised
           </button>
-          <button class="mdc-button mdc-button--dense" no-js>
+          <button class="mdc-button mdc-button--dense" data-demo-no-js>
             Dense Default
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--dense" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--dense" data-demo-no-js>
             Dense Raised
           </button>
-          <button class="mdc-button mdc-button--compact" no-js>
+          <button class="mdc-button mdc-button--compact" data-demo-no-js>
             Compact
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--compact" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--compact" data-demo-no-js>
             Compact Raised
           </button>
-          <button class="mdc-button mdc-button--primary" no-js>
+          <button class="mdc-button mdc-button--primary" data-demo-no-js>
             Default with Primary
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--primary" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--primary" data-demo-no-js>
             Raised with Primary
           </button>
-          <button class="mdc-button mdc-button--accent" no-js>
+          <button class="mdc-button mdc-button--accent" data-demo-no-js>
             Default with Accent
           </button>
-          <button class="mdc-button mdc-button--raised mdc-button--accent" no-js>
+          <button class="mdc-button mdc-button--raised mdc-button--accent" data-demo-no-js>
             Raised with Accent
           </button>
-          <div class="mdc-button mdc-button--raised" tabindex="0" role="button" no-js>
+          <div class="mdc-button mdc-button--raised" tabindex="0" role="button" data-demo-no-js>
             Div
           </div>
         </fieldset>
@@ -310,7 +310,7 @@
         function init() {
           var btns = document.querySelectorAll('.mdc-button');
           for (var i = 0, btn; btn = btns[i]; i++) {
-            if (!btn.hasAttribute('no-js')) {
+            if (!btn.hasAttribute('data-demo-no-js')) {
               mdc.ripple.MDCRipple.attachTo(btn);
             }
           }
--- a/demos/button.html
+++ b/demos/button.html
@@ -75,6 +75,42 @@
           </div>
         </fieldset>
         <fieldset>
+          <legend>Graceful degraded buttons</legend>
+          <button class="mdc-button" no-js>
+            Default
+          </button>
+          <button class="mdc-button mdc-button--raised" no-js>
+            Raised
+          </button>
+          <button class="mdc-button mdc-button--dense" no-js>
+            Dense Default
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--dense" no-js>
+            Dense Raised
+          </button>
+          <button class="mdc-button mdc-button--compact" no-js>
+            Compact
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--compact" no-js>
+            Compact Raised
+          </button>
+          <button class="mdc-button mdc-button--primary" no-js>
+            Default with Primary
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--primary" no-js>
+            Raised with Primary
+          </button>
+          <button class="mdc-button mdc-button--accent" no-js>
+            Default with Accent
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--accent" no-js>
+            Raised with Accent
+          </button>
+          <div class="mdc-button mdc-button--raised" tabindex="0" role="button" no-js>
+            Div
+          </div>
+        </fieldset>
+        <fieldset>
           <legend>Links with Button Style</legend>
           <a href="/button.html" class="mdc-button">
             Default
@@ -183,6 +219,42 @@
             Div
           </div>
         </fieldset>
+        <fieldset>
+          <legend>Graceful degraded buttons</legend>
+          <button class="mdc-button" no-js>
+            Default
+          </button>
+          <button class="mdc-button mdc-button--raised" no-js>
+            Raised
+          </button>
+          <button class="mdc-button mdc-button--dense" no-js>
+            Dense Default
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--dense" no-js>
+            Dense Raised
+          </button>
+          <button class="mdc-button mdc-button--compact" no-js>
+            Compact
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--compact" no-js>
+            Compact Raised
+          </button>
+          <button class="mdc-button mdc-button--primary" no-js>
+            Default with Primary
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--primary" no-js>
+            Raised with Primary
+          </button>
+          <button class="mdc-button mdc-button--accent" no-js>
+            Default with Accent
+          </button>
+          <button class="mdc-button mdc-button--raised mdc-button--accent" no-js>
+            Raised with Accent
+          </button>
+          <div class="mdc-button mdc-button--raised" tabindex="0" role="button" no-js>
+            Div
+          </div>
+        </fieldset>
         <fieldset disabled>
           <legend>Disabled Buttons</legend>
           <button class="mdc-button">
@@ -238,7 +310,9 @@
         function init() {
           var btns = document.querySelectorAll('.mdc-button');
           for (var i = 0, btn; btn = btns[i]; i++) {
-            mdc.ripple.MDCRipple.attachTo(btn);
+            if (!btn.hasAttribute('no-js')) {
+              mdc.ripple.MDCRipple.attachTo(btn);
+            }
           }
         }
       })();
--- a/demos/ripple.html
+++ b/demos/ripple.html
@@ -47,8 +47,8 @@
         border-radius: 50%;
       }
 
-      .mdc-ripple-surface[data-mdc-ripple-is-unbounded]::before,
-      .mdc-ripple-surface[data-mdc-ripple-is-unbounded]::after {
+      .mdc-ripple-surface[data-mdc-ripple-is-unbounded][data-demo-no-js]::before,
+      .mdc-ripple-surface[data-mdc-ripple-is-unbounded][data-demo-no-js]::after {
         height: 100%;
         left: 0;
         top: 0;
--- a/demos/ripple.html
+++ b/demos/ripple.html
@@ -59,7 +59,7 @@
       section {
         align-items: center;
         display: flex;
-
+        margin-left: 10px;
       }
 
       section > div {
@@ -89,6 +89,13 @@
         user-select: none;
         -webkit-user-select: none;
       }
+
+      @media (max-width: 565px) {
+        section {
+          flex-direction: column;
+          width: 100%;
+        }
+      }
     </style>
   </head>
   <body>
@@ -106,8 +113,8 @@
           </div>
         </div>
         <div>
-          <h2>Bounded - Graceful degraded</h2>
-          <div class="mdc-ripple-surface demo-surface mdc-elevation--z2" no-js tabindex="0">
+          <h2>Bounded - CSS Only</h2>
+          <div class="mdc-ripple-surface demo-surface mdc-elevation--z2" data-demo-no-js tabindex="0">
             Interact with me!
           </div>
         </div>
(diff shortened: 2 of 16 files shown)


Garbee:
How exactly should the "enhanced" states look when tabbing through a document? Currently there is no indication of what button has focus which is extremely inaccessible.

demos/button.html
```diff
@@ -75,6 +75,42 @@
           </div>
         </fieldset>
         <fieldset>
+          <legend>Graceful degraded buttons</legend>
+          <button class="mdc-button" no-js>
```
traviskaufman: can we make those attributes `data-demo-no-js` instead of `no-js` so that they are valid HTML attrs? I also think having the `demo` in there is useful since it signals to users that the data attribute is not part of the component's API and exists for demo purposes only

demos/ripple.html
```diff
@@ -92,41 +99,83 @@
...
-        <div class="mdc-ripple-surface demo-surface mdc-elevation--z2" tabindex="0">
-          Interact with me!
+        <div>
+          <h2>Bounded</h2>
+          <div class="mdc-ripple-surface demo-surface mdc-elevation--z2" tabindex="0">
+            Interact with me!
+          </div>
+        </div>
+        <div>
+          <h2>Bounded - Graceful degraded</h2>
```
traviskaufman: nit: How about "CSS Only" instead of "Gracefully degraded"?

demos/button.html
```diff
@@ -75,6 +75,42 @@
           </div>
         </fieldset>
         <fieldset>
+          <legend>Graceful degraded buttons</legend>
```
traviskaufman: How about "CSS Only" instead of "Gracefully degraded"?

demos/ripple.html
```diff
@@ -57,7 +57,14 @@
...
       section {
-        margin: 48px;
+        align-items: center;
+        display: flex;
+
+      }
+
+      section > div {
+        margin: 48px 0;
+        width: 400px;
```
traviskaufman: should we add a media query here that sets it to `100%` width if the viewport is smaller than `400px`? I'm thinking of how this would appear on v. small screens

demos/ripple.html
```diff
@@ -92,41 +99,83 @@
...
-          Interact with me!
+        <div>
+          <h2>Bounded</h2>
+          <div class="mdc-ripple-surface demo-surface mdc-elevation--z2" tabindex="0">
+            Interact with me!
+          </div>
+        </div>
+        <div>
+          <h2>Bounded - Graceful degraded</h2>
+          <div class="mdc-ripple-surface demo-surface mdc-elevation--z2" no-js tabindex="0">
```
traviskaufman: same comment regarding `no-js` --> `data-demo-no-js`

traviskaufman:
@cristobalchao not sure that this branch was cleanly rebased from master since it looks like there's a bunch of extemporaneous commits prior to 64b9db3. You may want to try and `git pull --rebase origin master` on this branch, checkout a fresh branch and cherry-pick 64b9db3 and 7a4df35 over, or use some other strategy.

packages/mdc-button/mdc-button.scss
```diff
@@ -22,6 +22,39 @@
...
 // postcss-bem-linter: define button
+
+.mdc-button {
+  @include mdc-ripple-base;
+  @include mdc-ripple-bg((pseudo: "::before"));
+  @include mdc-ripple-fg((pseudo: "::after"));
+
+  overflow: hidden;
+
+  @include mdc-theme-dark(".mdc-button", true) {
```
traviskaufman: I think this can just be `@include mdc-theme-dark {` here. We use the first argument in the case where we need to style an _BEM element_ class  to adhere to a dark theme treatment

```scss
.mdc-component__element {
  @include mdc-theme-dark(".mdc-component") {
    // ...
  } 
}
```

And we usually use the second argument when we're styling a modifier class for dark theme treatment, e.g.

```scss
.mdc-component--mod {
  @include mdc-theme-dark(".mdc-component", true) {
    // ...
  }
}
```

The [mdc-theme README](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#mdc-theme-dark-mixin) has more info on this.

packages/mdc-button/mdc-button.scss
```diff
@@ -22,6 +22,39 @@
...
+      @include mdc-ripple-bg((pseudo: "::before", theme-style: $theme-style, opacity: .12));
+      @include mdc-ripple-fg((pseudo: "::after", theme-style: $theme-style, opacity: .12));
+    }
+  }
+}
+
+.mdc-button--raised {
+  @each $theme-style in (primary, accent) {
+    &.mdc-button--#{$theme-style} {
+      @include mdc-ripple-base;
```
traviskaufman: Should we put a `TODO` here and in other components containing elevation that we want to disable ink wash on hover and alter elevation instead as part of https://github.com/material-components/material-components-web/issues/194?

packages/mdc-ripple/README.md
```diff
@@ -30,6 +30,9 @@ to provide components (or any element at all) with a
 material "ink ripple" interaction effect. It is designed to be efficient, un-invasive, and usable
 without adding any extra DOM to your elements.
 
+and includes a gracefully degraded version that can be used
```
traviskaufman: This paragraph looks a bit broken...

Maybe we should say:

MDC Ripple also works without javascript, where it gracefully degrades to a simpler CSS-Only implementation.

packages/mdc-ripple/README.md
```diff
@@ -74,41 +75,17 @@ use [mdc-elevation](https://github.com/material-components/material-components-w
...
-  &:focus::before {
-    opacity: .06;
-  }
-}
 ```
 
 #### Adding the ripple Sass
 
-When a ripple is successfully initialized on an element, it dynamically adds a `mdc-ripple-upgraded` class to that element. Therefore, to add a ripple to our surface, first we include the proper Sass mixins within our surface's styles when it contains this class. We also add a few additional properties that ensure the ripple's UX is correct.
+To add a ripple to our surface, first we include the proper Sass mixins within our surface's styles when it contains this class. We also add a few additional properties that ensure the ripple's UX is correct.
```
traviskaufman: I think we can remove the portion of text that says "when it contains this class", since I'm pretty sure _"this class"_ refers to `mdc-ripple-upgraded`, which we now take care of automatically.

packages/mdc-ripple/README.md
```diff
@@ -121,6 +98,8 @@ When a ripple is successfully initialized on an element, it dynamically adds a `
 
 This code sets up `.surface` with the correct css variables as well as `will-change` properties to support the ripple. It then dynamically generates the correct selectors such that the surface's `::before` element functions as a background ripple, and the surface's `::after` element functions as a foreground ripple.
 
+When a ripple is successfully initialized on an element, it dynamically adds a `mdc-ripple-upgraded` class to that element. If ripple is not initialized but Sass misins are included within our surface, the ripple would be on a graceful degraded state.
```
traviskaufman: Change "the ripple would be on a graceful degraded state" to "the ripple will still work, but it would use a simpler, CSS-Only implementation which relies on `:hover`, `:active`, and `:focus`".

demos/button.html
cristobalchao: SGTM

demos/button.html
cristobalchao: SGTM

demos/ripple.html
cristobalchao: SGTM

demos/ripple.html
cristobalchao: SGTM

demos/ripple.html
cristobalchao: SGTM

packages/mdc-button/mdc-button.scss
cristobalchao: SGTM

packages/mdc-button/mdc-button.scss
cristobalchao: SGTM

packages/mdc-ripple/README.md
cristobalchao: Much better!

packages/mdc-ripple/README.md
cristobalchao: SGTM

packages/mdc-ripple/README.md
cristobalchao: SGTM

demos/ripple.html
Garbee: Or just `max-width: 400px;` so on small screens it just gets smaller as-needed. But doesn't grow above that.

cristobalchao:
All issues addressed PTAL

packages/mdc-button/mdc-button.scss
```diff
@@ -22,6 +22,40 @@
...
 // postcss-bem-linter: define button
+
+.mdc-button {
+  @include mdc-ripple-base;
+  @include mdc-ripple-bg((pseudo: "::before"));
+  @include mdc-ripple-fg((pseudo: "::after"));
+
+  overflow: hidden;
+
+  @include mdc-theme-dark(".mdc-button") {
```
traviskaufman: this should just be `@include mdc-theme-dark {` with no argument

packages/mdc-button/mdc-button.scss
```diff
@@ -22,6 +22,40 @@
...
+  @each $theme-style in (primary, accent) {
+    &.mdc-button--#{$theme-style} {
+      @include mdc-ripple-base;
+      @include mdc-ripple-bg((pseudo: "::before", theme-style: $theme-style, opacity: .12));
+      @include mdc-ripple-fg((pseudo: "::after", theme-style: $theme-style, opacity: .12));
+    }
+  }
+}
+
+/* TODO(cristobalchao): Disable ink wash on hover and alter elevation instead for raised surfaces. */
```
traviskaufman: Can we switch this to a `// Sass style comment`? This way it's not output in the built CSS

packages/mdc-ripple/README.md
```diff
@@ -121,6 +95,8 @@ When a ripple is successfully initialized on an element, it dynamically adds a `
 
 This code sets up `.surface` with the correct css variables as well as `will-change` properties to support the ripple. It then dynamically generates the correct selectors such that the surface's `::before` element functions as a background ripple, and the surface's `::after` element functions as a foreground ripple.
 
+When a ripple is successfully initialized on an element, it dynamically adds a `mdc-ripple-upgraded` class to that element. If ripple is not initialized but Sass misins are included within our surface, the ripple will still work, but it would use a simpler, CSS-Only implementation which relies on `::hover`, `::active`, and `::focus`.
```
traviskaufman: nit: `:hover`, `:active`, and `:focus`, not double-colon. Double-colons are for pseudo-_elements_, whereas single-colons are for pseudo-_classes_

packages/mdc-ripple/_mixins.scss
```diff
@@ -101,6 +109,26 @@
   }
   // stylelint-enable at-rule-empty-line-before, block-closing-brace-newline-after
 
+  &.mdc-ripple-upgraded {
+    &#{$pseudo} {
+      transition-duration: 0ms;
```
traviskaufman: Should this just be `transition: none;` instead?

packages/mdc-ripple/_mixins.scss
```diff
@@ -42,6 +42,13 @@
 
   will-change: transition, opacity;
```
traviskaufman: Eek! This should be `transform`, not `transition` 😝 by bad! Mind fixing this?

packages/mdc-ripple/README.md
cristobalchao: Ups, my bad!

packages/mdc-ripple/_mixins.scss
cristobalchao: Much better!

packages/mdc-ripple/_mixins.scss
cristobalchao: Lol! I totally missed that too. Fixed!

packages/mdc-button/mdc-button.scss
cristobalchao: sgtm

packages/mdc-button/mdc-button.scss
cristobalchao: Yup!

demos/ripple.html
```diff
@@ -92,41 +106,83 @@
...
-          favorite
+        <div>
+          <h2>Unbounded</h2>
+          <div class="mdc-ripple-surface demo-surface material-icons" data-mdc-ripple-is-unbounded
+               aria-label="Favorite" tabindex="0">
+            favorite
+          </div>
+        </div>
+        <div>
+          <h2>Unbounded - CSS Only</h2>
```
sgomes: The fallback ripple seems larger than the JS-enabled one. Is this intentional?

packages/mdc-ripple/README.md
```diff
@@ -25,10 +25,10 @@
   - [Caveat: Safari](#caveat-safari)
   - [Caveat: Theme Custom Variables](#caveat-theme-custom-variables)
 
-MDC Ripple provides the Javascript and CSS required
-to provide components (or any element at all) with a
-material "ink ripple" interaction effect. It is designed to be efficient, un-invasive, and usable
-without adding any extra DOM to your elements.
+MDC Ripple provides the Javascript and CSS required to provide components (or any element at all) with a
+material "ink ripple" interaction effect. It is designed to be efficient, un-invasive, and usable without adding any extra DOM to your elements.
```
sgomes: Nit: no need for hyphen in "un-invasive".

packages/mdc-ripple/_mixins.scss
```diff
@@ -126,15 +154,17 @@
...
-  left: 0;
-  width: var(--mdc-ripple-fg-size);
-  height: var(--mdc-ripple-fg-size);
-  transform: scale(0);
-  transform-origin: center center;
+  top: calc(50% - #{$radius});
+  left: calc(50% - #{$radius});
+  width: $radius * 2;
+  height: $radius * 2;
+  transform: scale(var(--mdc-ripple-fg-scale));
```
sgomes: In order to make `var()` as bulletproof as possible, consider following the pattern:

```css
transform: scale(0);
transform: scale(var(--mdc-ripple-fg-scale, 0));
```
(replace 0 with whatever default value you deem fit)

The first line gives you a fallback for browsers that don't understand custom properties, whereas the second value in the `var()` function gives you a fallback in case `--mdc-ripple-fg-scale` isn't defined.

Another option is to stick a big a big `@supports` around everything (which to be fair, you may already be doing in JavaScript), but you should at least keep the second parameter to `var()`.

packages/mdc-ripple/_mixins.scss
```diff
@@ -156,6 +186,22 @@
   }
   // stylelint-enable at-rule-empty-line-before, block-closing-brace-newline-after
 
+  &.mdc-ripple-upgraded {
+    &#{$pseudo} {
+      top: 0;
+      left: 0;
+      width: var(--mdc-ripple-fg-size);
```
sgomes: Same comment as above, and everywhere else `var()` is used in the file.

traviskaufman: Do we need to use the dual-property method for `.mdc-ripple-upgraded`? I think we left that off initially b/c we don't even upgrade the ripple if Custom Vars aren't supported by the browser. Or, should we just do it for consistency?

sgomes: Yeah, if we're checking in JS, there's no need for the first fallback. Still might be useful to do `var(--mdc-ripple-fg-size, 16px)` (or whatever sensible default), in case `--mdc-ripple-fg-size` isn't defined.

traviskaufman: True, the second arg fallback is always a good call

cristobalchao:
All issues addressed PTAL

packages/mdc-button/mdc-button.scss
```diff
@@ -22,6 +22,40 @@
 @import "@material/typography/mixins";
 
 // postcss-bem-linter: define button
+
+.mdc-button {
```
traviskaufman: These style rules should be put within the `.mdc-button` declaration below, which is the top-level declaration for `mdc-button`. Having an extra one seems redundant, no?

packages/mdc-radio/mdc-radio.scss
```diff
@@ -58,8 +62,8 @@ $mdc-radio-transition-duration: 120ms;
       left: 0;
       width: 100%;
       height: 100%;
-      transform: scale(0, 0);
-      transition: mdc-radio-exit(opacity), mdc-radio-exit(transform);
+      // transform: scale(0, 0);
```
traviskaufman: forgot to remove?

packages/mdc-ripple/_mixins.scss
```diff
@@ -118,23 +152,29 @@
   }
 
   &.mdc-ripple-upgraded--unbounded#{$pseudo} {
-    top: var(--mdc-ripple-top);
-    left: var(--mdc-ripple-left);
-    width: var(--mdc-ripple-fg-size);
-    height: var(--mdc-ripple-fg-size);
+    top: calc(50% - #{$radius / 2}); // fallback for browsers that don't support custom properties
```
traviskaufman: I don't think we need the fallback here, since this gets styled by using `mdc-ripple-upgraded--unbounded`. Since the JS checks for CSS Var support before attaching any of the `mdc-ripple-upgraded` classes, we can guarantee here that CSS variables are supported and thus you can remove this.

packages/mdc-ripple/_mixins.scss
```diff
@@ -101,6 +109,32 @@
   }
   // stylelint-enable at-rule-empty-line-before, block-closing-brace-newline-after
 
+  &.mdc-ripple-upgraded {
+    &#{$pseudo} {
+      top: calc(50% - #{$radius});
+      left: calc(50% - #{$radius});
+      width: $radius * 2;
+      height: $radius * 2;
+      transform: scale(0); // fallback for browsers that don't support custom properties
```
traviskaufman: You can remove the inline comment; this is a convention in our codebase which we talk about in our mdc-theme package.

packages/mdc-ripple/_mixins.scss
```diff
@@ -156,13 +197,35 @@
   }
   // stylelint-enable at-rule-empty-line-before, block-closing-brace-newline-after
 
+  &.mdc-ripple-upgraded {
+    &#{$pseudo} {
+      top: 0;
+      left: 0;
+      width: $radius; // fallback for browsers that don't support custom properties
```
traviskaufman: Same comment here regarding not needing fallback for `mdc-ripple-upgraded`

packages/mdc-ripple/foundation.js
```diff
@@ -272,7 +273,8 @@ export default class MDCRippleFoundation extends MDCFoundation {
     let approxCurScale = 0;
     if (msElapsed > FG_TRANSFORM_DELAY_MS) {
       const percentComplete = Math.min((msElapsed - FG_TRANSFORM_DELAY_MS) / this.xfDuration_, 1);
-      approxCurScale = percentComplete * this.fgScale_;
+      // approxCurScale = percentComplete * this.fgScale_;
```
traviskaufman: forgot to remove?

cristobalchao: My bad!

packages/mdc-radio/mdc-radio.scss
cristobalchao: Ups...my bad!

cristobalchao:
All issues addressed PTAL

packages/mdc-ripple/_mixins.scss
```diff
@@ -115,7 +115,7 @@
       left: calc(50% - #{$radius});
       width: $radius * 2;
       height: $radius * 2;
-      transform: scale(0); // fallback for browsers that don't support custom properties
+      transform: scale(0);
```
traviskaufman: this fallback is not needed, since the property is under `.mdc-ripple-upgraded`

traviskaufman:
Couple of issues I noticed in the demo:

- JS Checkboxes and JS Radios no longer seem to use the correct ripple functionality
- I think for upgraded ripples, there's a jump-cut when hovering over the elements. I think we need to remove the hover styles attached by the graceful degredation fallback when the ripple is upgraded. This is especially apparent for unbounded ripples.

cristobalchao:
All issues addressed PTAL

traviskaufman:
Excellent work @cristobalchao!

omtinez:
Does this address elements retaining the "ripple" on focus? E.g. the radio buttons keep a "halo" after clicking them: http://i.imgur.com/NoP7SsU.png

I though I would verify since this seemed like it addresses that before opening a new issue. Also, related: are there any plans to allow users from disabling the ripple effects altogether?

Garbee:
@omtinez That seems like intended functionality. That is focus state for accessibility purposes and is highly required.

Ripple effects are added via JavaScript instantiation. Otherwise there should be a CSS-only effect for these states that looks like a ripple but doesn't function like the actual ripples. If you use the JS though for these form controls, Ripples are basically bundled-in since they are a specification detail.

cristobalchao:
@omtinez That's the intended behavior. As @Garbee mentioned, the focus state is required for all components for accessibility reasons. About disabling the ripple effects altogether, there are no plans in the short term future.

omtinez:
I commend your focus on accessibility, but I think that the current look is a bit over the top (in my opinion, hideous) by making a radio button appear twice as big. Specially considering that the Javascript-enabled behavior is a ripple for a fraction of a second; I'm not sure that the Javascript-enabled behavior is much more accessible e.g. to someone visually impaired. I'm happy to open a separate issue if you are willing to discuss the idea there (or want it just for tracking purposes, even if it's going to be immediately closed).

For the sake of documenting, even though it breaks the accessibility argument, this is the CSS workaround that I'm using and disables the enormous halo around radio buttons:

.mdc-radio__background::before {
    display: none;
}

Garbee:
These states [are explicitly defined](https://material.io/guidelines/components/selection-controls.html#selection-controls-radio-button) in the spec. So it adheres to MD guidance as-is and is accessible. Disabling this is discouraged unless you replace it with some other focus indicator for people who need this. However unless the MD team provides us with alternative measures to implement for a11y in this case it shouldn't be changed.

I believe the focus state is shown initially then removed with JS since that is a part of the ripple effect specification somewhere else. IMO it isn't as useful for a11y but that is purely subjective conjecture and I have no data to back it up to influence any changes.

PR #288, opened 2017-02-12
fix:MDCCheckbox add box-sizing: content-box to checkbox component

senadir:
consider adding the default value of box-sizing to the element
most websites today are using a variation of the global `box-sizing: border-box` https://css-tricks.com/box-sizing/
checkbox fails to work if the page box-sizing is set to `border-box`, adding `content-box` ensure that the element will be rendered as intended regardless of the page global box-sizing value

### DIFF
--- a/packages/mdc-checkbox/mdc-checkbox.scss
+++ b/packages/mdc-checkbox/mdc-checkbox.scss
@@ -54,6 +54,7 @@
 .mdc-checkbox {
   display: inline-block;
   position: relative;
+  box-sizing: content-box;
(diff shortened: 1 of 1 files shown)


senadir:
I signed it!

traviskaufman:
I tested this locally and it doesn't seem to break any pre-existing functionality, so I'd be okay with merging it in. @sgomes WDYT?

sgomes:
Merged as a7f62213406d5c4e6c4626aeccdfa0081fe58c0f

Issue #130, opened 2016-12-19
Migration Guide from MDL to MDC


SabineMa:
This is not a Feature Request and not a bug but the question: will you provide a migration guide for people using MDL and want to migrate to MDC?

sgomes:
Thanks for the suggestion, @SabineMa! There's a big difference between the two, but it could prove useful as a complement to our Getting Started guide. @traviskaufman What do you think?

igilham:
This would definitely be useful to anyone who uses MDL and wants to migrate.

sgomes:
We've discussed this question a bit andthink there's a few things that could be useful in a migration guide. Let me try to summarize them:

- List of components in MDC-Web and MDL, and how they map to each other.
- Discussion of the different ways of depending on / using MDL, and the high-level changes you have to make to move from MDL to MDC-Web with each one. Plus discuss the new options only available in MDC-Web.
- Discussion of whether or not to use `mdc-auto-init` (auto initialization was mandatory in MDL, but not in MDC-Web).

All of this peppered with plenty of links into the existing MDC-Web documentation, for finding out more. How does this sound, @SabineMa @traviskaufman @amsheehan?

ohdoukyoung:
I wonder if the scope of the browser support is also included in the migration guide.
Are there any plans to support IE9-10 like MDL?

sgomes:
@ohdoukyoung No plans for IE9-10 in MDC-Web, sorry; it's missing too much, like flexbox and calc. We'll add that info to the guide.

SabineMa:
Only for the sake of completeness 

- List of NEW components in MDC-Web

sgomes:
@SabineMa Yes, sounds good!

@igilham Yep, good idea. It might also be useful have the indication of whether those missing components are planned for later, and which milestone (beta or stable) they're planned for.

koutsenko:
Hello, will you implement DataTable component?
I've used it in react-mdl (MDL v.1 hasn't framework bindings, so additional react-mdl layer was necessary). But I don't see DataTable in the beta milestone of MDC-Web.

p.s. If you'll create a such task, please add the feature: data scrolling with fixed (visible)  header also.
Thanks a lot.

sgomes:
This is now [merged](/material-components/material-components-web/blob/master/docs/migrating-from-mdl.md).

PR #295, opened 2017-02-16
docs: Add MDL migration guide.

sgomes:
Implements #130.
[skip ci]

### DIFF
--- a/docs/migrating-from-mdl.md
+++ b/docs/migrating-from-mdl.md
@@ -1,0 +1,345 @@
+# Migrating from Material Design Lite
+
+Material Components for the web (MDC-Web) is the successor to the Material Design Lite (MDL) project.
+
+While the philosophy behind the two projects is quite similar, migrating to MDC-Web requires a number of changes, from
+class names to different DOM structures. In addition, there are several choices to be made regarding component
+initialization, how to depend on MDC-Web, and theming/styling mechanisms.
+
+This document attempts to summarize and guide you through the work involved. Let’s get started!
+
+> **Note:** If you’re thinking of migrating your application to MDC-Web, please bear in mind that it’s still in an alpha
+state and thus APIs and certain UX features are subject to change.
+
+
+## Depending on MDC-Web
+
+MDL is distributed on NPM, Bower, and through its own CDN. MDC-Web is currently only available on NPM.
+
+While MDL is a singular, universal library consisting of all components and styles, MDC-Web has been designed to be
+modular, allowing you to make the choice of either pulling in everything, or just the packages you want.
+
+### NPM
+
+MDC-Web is made available on [NPM](https://www.npmjs.com/), with packages living under the `@material` namespace.
+
+In order to install e.g. the button component, you can run:
+
+```
+npm install --save @material/button
+```
+
+Some packages serve as dependencies for others, so don’t be surprised if you end up with multiple packages in your
+`node_modules` folder! There are no runtime dependencies outside of MDC-Web, so the dependency tree should remain small.
+
+If you want all of MDC-Web, you can pull in the meta package:
+
+```
+npm install --save material-components-web
+```
+
+### Content Distribution Network (CDN)
+
+While MDL is available over a CDN, there’s currently no equivalent in MDC-Web (it’s in the plans for the future,
+though!).
+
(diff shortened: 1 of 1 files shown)


docs/migrating-from-mdl.md
```diff
@@ -0,0 +1,318 @@
...
+Material Components for the web (MDC-Web) is the successor to the Material Design Lite (MDL) project.
+
+While the philosophy behind the two projects is quite similar, migrating to MDC-Web requires a number of changes, from
+class names to different DOM structures. In addition, there are several choices to be made regarding component
+initialization, how to depend on MDC-Web, and theming/styling mechanisms.
+
+This document attempts to summarize and guide you through the work involved. Let’s get started!
+
+> **Note:** If you’re thinking of migrating your application to MDC-Web, please bear in mind that it’s still in an alpha
+state and thus not recommended for production.
```
traviskaufman: I don't know if we want to say `not recommended for production`. We may want to say something like `APIs and certain UX features are subject to change`.

docs/migrating-from-mdl.md
```diff
@@ -0,0 +1,318 @@
...
+
+
+## Depending on MDC-Web
+
+MDL is distributed on NPM, Bower, and through its own CDN. MDC-Web is currently only available on NPM.
+
+While MDL is an all-in-one solution, MDC-Web has been designed to be modular, allowing you to make the choice of either
```
traviskaufman: Is "all-in-one" solution the right term here? Do we want to say something like "While MDL is a singular, universal library consisting of all components and styles", etc.

docs/migrating-from-mdl.md
```diff
@@ -0,0 +1,318 @@
...
+And JS:
+
+```
+https://unpkg.com/material-components-web@latest/dist/material-components-web.js
+```
+
+Check the [unpkg CDN main page](https://unpkg.com/) for more information on how to request particular version ranges.
+
+There’s also the option of getting individual packages, rather than the `material-components-web` meta-package. However,
+since unpkg only serves up package contents without resolving dependencies, you’d need to maintain the dependency tree
```
traviskaufman: technically they could rely on the files in `dist`, which are bundled, so I'm not sure this is entirely true.

docs/migrating-from-mdl.md
```diff
@@ -0,0 +1,318 @@
...
+yourself, which is not recommended.
+
+
+## Initializing components
```
traviskaufman: We should probably write something about writing your own DOM here, no?

docs/migrating-from-mdl.md
```diff
@@ -0,0 +1,318 @@
...
+```css
+.my-grid {
+  --mdc-layout-grid-margin: 40px;
+  --mdc-layout-grid-gutter: 16px;
+}
+```
+
+Be sure to read the README for each component to get all the details on their usage.
+
+> **Note:** CSS Custom Property support in components is a progressive enhancement feature that doesn’t work in every supported browser. If
```
traviskaufman: Should we talk about how we also usually provide SCSS variables / plain CSS styling guidance in order to help with this?

docs/migrating-from-mdl.md
sgomes: Done.

docs/migrating-from-mdl.md
sgomes: Done.

docs/migrating-from-mdl.md
sgomes: Got it. I'll remove that bit.

docs/migrating-from-mdl.md
sgomes: That's the same as in MDL, but I'll add a bit about how MDC-Web doesn't write any DOM for you, unlike MDL.

docs/migrating-from-mdl.md
sgomes: Done.

sgomes:
Thanks, @traviskaufman! Unfortunately, I don't think there are a lot of opportunities for the "before" and "after" type stuff you mention, unless we get pretty deep into how to port each individual component. If we really want to do that, I'd suggest adding a "migrating from MDL" section to each components' README instead, in which case let me know and I'll create an issue for it!

Issue #143, opened 2016-12-22
Investigate future test infrastructure


traviskaufman:
Continued from #5 

- [x] Determine what type of mocha interface we want to use. The Qunit-style interface looks most similar to what we already have.
- [x] Decide on setup for running mocha locally 
- [x] Investigate what it would take to refactor all tests. Investigate potentially using something like recast.
- [ ] Ensure that Cross-Browser testing via Sauce Labs works and is less flaky
- [x] Ensure that debugging still works in development, and coverage still works on single runs / CI.
- [x] Ensure that all docs regarding testing are updated

PR #296, opened 2017-02-16
refactor: Move tests from tape to mocha

traviskaufman:
Implements all of #143  except for cross-browser testing.

This commit migrates all of our testing infrastructure from tape over to
mocha. We were hitting tape's limits in terms of usability for MDC-Web,
and this gives us a clean break.

Note that most of the changes here were done using an automated script
that can be found at
https://gist.github.com/traviskaufman/59476a12b65925d06b048b61b8ef085c

Note that there are some bugs in the code but I double-checked and it
looked like everything transpiled correctly.

 ## Improvements over the old infrastructure

- _Massive_ speedup in test runtime. Our full suite executes in ~6s
  (locally), which is about 3x faster than before.
- No pollution from TAP console output, only the results from the karma
  reporter are logged to the command line.
- A way better debugging experience. You can now exclude one or more
  tests, or include one or more tests, a big step up from test.only()
  that existed before. Furthermore, the Karma Debug UI now uses mocha's
  html reporter, which produces beautiful output and allows you to drill
  down into individual tests and only run them, view the test code in
  the browser, and more.
- All of the ease and usability of mocha (returning promises for async
  tests, a rich set of configuration options, top-level exception
  handling within tests, etc.)
- A more robust assertion library using chai, including better error
  reporting when assertions fail.

 ## What Changes

 ### Spec Descriptions

Instead of

```js
// math.test.js

import test from 'tape';

test('1 + 1 equals 2', (t) => {
  t.equal(1 + 1, 2);
  t.end();
});
```

You write

```js
// math.test.js

import {assert} from 'chai';

suite('MDCMath');

test('1 + 1 equals 2', () => {
  assert.equal(1 + 1, 2);
});
```

There is one `suite()` function per file, which is a top-level
description of the tests this file contains. Suites should always start
with `MDC`, which is enforced by eslint.

 ### Testdouble verifications

Instead of writing

```js
t.doesNotThrow(() => td.verify(mockAdapter.addClass('foo')));
```

You write

```js
td.verify(mockAdapter.addClass('foo');
```

 ### tape assertion methods vs. chai's assertion methods

- `t.true` -> `assert.isOk`
- `t.false` -> `assert.isNotOk`
- everything other assertion method pretty much stays the same. Plus, we
  have even more robust methods to work with thanks to chai.

 ## Next steps

- Re-enabling SauceLabs cross-browser testing, which is the final part
  of #143 .


traviskaufman:
Rebased master and update mdc-form-field tests. Will merge as soon as TravisCI goes green.

PR #299, opened 2017-02-17
chore(coverage): Move away from isparta-loader

BBosman:
When running `npm install` the following warning is displayed:

```bash
npm WARN deprecated isparta-loader@2.0.0: Package is deprecated, use https://github.com/deepsweet/istanbul-instrumenter-loader
```

This PR makes that switch.

These are the numbers before and after the switch:

Before
> Statements   : 99.03% ( 2456/2480 ), 164 ignored
> Branches     : 96.68% ( 1224/1266 ), 262 ignored
> Functions    : 99.01% ( 801/809 ), 17 ignored
> Lines        : 98.37% ( 1449/1473 )

After
> Statements   : 98.26% ( 1354/1378 )
> Branches     : 94.84% ( 496/523 )
> Functions    : 98.26% ( 620/631 )
> Lines        : 98.23% ( 1331/1355 )

As you can see the numbers dropped a bit. For _branches_ even below the allowed threshold of 95%. 

Looking into those numbers it appears the new numbers are more accurate than the old ones. For example, if you look at the mdc-radio/index.js.html coverage report you get these results:

Before
before

After
after

In the _before_ you can see that the _constructor_ has been (partly) marked as _ignored_, leading to a higher coverage percentage. There are also small issues with the highlighting. In the _after_ those highlighting issues are gone and the constructor is no longer ignored.

As a result of these changes a couple of new tests need to be written to reach the 95% again though. 😋


BBosman:
Travis CI failed as expected due to the drop below 95%.

traviskaufman:
Nice! And those coverage drops do look like true positives, as we currently aren't fully testing our Radio Adapter methods. I think I'll have time to write the rest of the tests today, and push it up on this branch if that works. Otherwise, if you're interested in taking a crack at writing in the tests I can point you to exactly what would need to be done.

traviskaufman:
Turns out that merging latest master fixes the coverage issue, probably due to the new unrelated tests we added 😅 

We should definitely make a note to fill in any coverage gaps caught, but I'm okay with merging this for now.

PR #301, opened 2017-02-17
feat(typography): Add !default to variables

### LINKED ISSUE #280: Mark mdc-typography variables as !default to allow overriding them
Is there a reason why mdc-typography [variables](https://github.com/material-components/material-components-web/blob/master/packages/mdc-typography/_variables.scss) (for example `$mdc-typography-styles`) don't have the !default flag set?

In my current project I would need to override few font size and weight settings from Material Design typography guidelines and it would be nice to do it the same way you can override mdl-layout-grid [variables](https://github.com/material-components/material-components-web/blob/master/packages/mdc-layout-grid/_variables.scss) as those have the !default flag.

BBosman:
Fixes #280.

### DIFF
--- a/packages/mdc-typography/_variables.scss
+++ b/packages/mdc-typography/_variables.scss
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-$mdc-typography-font-family: Roboto, sans-serif;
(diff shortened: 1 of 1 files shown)


traviskaufman:
Thank you @BBosman! @sgomes mind taking a look at this?

PR #323, opened 2017-02-21
docs(drawer): Update demo to use mdc-toolbar

### LINKED ISSUE #308: Update drawer demo to use mdc-toolbar
The drawer demo can be updated to use the `.mdc-toolbar` instead of `.demo-toolbar`.

peluprvi:
Closes #308


traviskaufman:
CI is failing because TravisCI doesn't start Sauce Labs, even though we're using the JWT addon as described, so I wouldn't worry about it given that this is a docs change.

traviskaufman:
btw if you add `[ci skip]` as part of your commit message, it will skip TravisCI. This is okay since this PR only handles demos.

peluprvi:
It's my pleasure to help!

Changing the temporary drawer to use `mdc-toolbar--fixed` creates an issue on h4 margin. Is there any class that already exists for fix that?

traviskaufman:
@peluprvi I think you need to change `mdc-toolbar__fixed-adjust` to `mdc-toolbar-fixed-adjust` (`-`, not `__`). That should correct the issue.

peluprvi:
@traviskaufman I still getting the `h1` too close to the [temporary drawer](https://github.com/peluprvi/material-components-web/blob/update-drawer-eg/demos/drawer/temporary-drawer.html) toolbar. Maybe I'm using something wrong.

traviskaufman:
@peluprvi it looks fine to me

Local server (with your changes):

drawer-local

Demo Site:

screen shot 2017-02-23 at 12 27 56

peluprvi:
@traviskaufman Permanent looks fine. The **Temporary** has the issue. I can't do a local server test and dig in it now. I can take a look tonight.

Current http://material-components-web.appspot.com/drawer/temporary-drawer.html (testing on Chrome):

image


Changing the toolbar and adding the `mdc-toolbar-fixed-adjust` class to `demo-main`:

image

image

image

traviskaufman:
@peluprvi hmm, looks like that's due to [margin-collapsing](http://stackoverflow.com/questions/1394724/css-margin-top-when-parents-got-no-border). I think the right solution is to put `overflow: auto` on `.demo-main`. I'm wondering if we should document this somewhere cc @yeelan0319 @sgomes

peluprvi:
Can be a `display: flex` on `body` too, like the Permanent Drawer demos.

traviskaufman:
(holding off on merging until last issue with temp drawer body is fixed)

@peluprvi I'd prefer if we did `overflow: auto` rather than changing the layout structure of the body just to prevent margin collapsing

traviskaufman:
Thank you @peluprvi !

PR #346, opened 2017-02-22
feat(menu): Move current time retrieval to adapter.

sgomes:
BREAKING CHANGE: adapters have to implement the new `getAccurateTime`
method.

Removes flakiness from tests as a side-effect.


packages/mdc-menu/README.md
```diff
@@ -245,6 +245,7 @@ The adapter for temporary drawers must provide the following functions, with cor
 | `isRtl() => boolean` | Returns boolean indicating whether the current environment is RTL. |
 | `setTransformOrigin(value: string) => void` | Sets the transform origin for the menu element. |
 | `setPosition(position: { top: string, right: string, bottom: string, left: string }) => void` | Sets the position of the menu element. |
+| `getAccurateTime() => number` | Returns an accurate representation of the current time, fit for animation purposes. |
```
traviskaufman: Can we suggest using `window.performance.now()` here? Either way, I think we want to be more specific about what "current time, fit for animation purposes" means in this description.

sgomes: Yup, will change. Thanks!

sgomes:
Waiting for CI green light and then merging.

Issue #47, opened 2016-12-05
Implement Grid List Component


traviskaufman:
Spec reference: https://www.google.com/design/spec/components/grid-lists.html#grid-lists-specs

> Copied from https://github.com/google/material-design-lite/issues/4498

yeelan0319:
Implemented the first prototype, see it in [CodePen](http://codepen.io/yeelan0319/pen/NdEPEz). Supports:
- Single line footer
- Single line footer with icon on left
- Single line footer with icon on right
- Two line footer
- Two line footer with icon on left and no icon on support text
- Two line footer with icon on right and icon on support text

yeelan0319:
Next step we need to figure out how grid list item size is defined and if it may have any relationship with grid layout with our designers.

yeelan0319:
Based on our conversation with the designer this morning, the grid list sounds like
- `Grid list cell` doesn't necessarily have things to do with `Layout grid columns`. On the other end, decouple the relationship between this 2 is desirable.
- `Grid list container` contains `Grid list items`. The former interact with overall layout (e.g.`Layout grid`).
- Two patterns are possible here: `Grid list container` defines how many cells per row it want to have and `grid cell` size responsive to the screen width changes v.s. `grid cell` size is defined and occupy the row space provided by `grid list container`. Both are possible and seems like the former is preferred.

Following up conversation would includes:
- If the `grid cell` size is responsive, do we have any constraints(e.g: max/min width) to prevent extreme cases (e.g: to preserve text readability)? If we put extra constraints, what shall we sacrifice instead: gutter width? number of cells per row?
- Another following up question is this also makes define different cell per row rule necessary for different screen width, so we need to have some **breakpoints**. For example it can changes at the same time when layout grid changes from 12 columns to 8 or 4 columns if that make sense. In this case, do we provide default value if user not define their rule for that screen width? 
- Do we preserve aspect ratio for the `grid list items`? If so, is that fixed 1:1/16:9? A set of predefined values? Customizable?

yeelan0319:
Had another conversation with our designer, there is definitely a lot to figure out, so instead, we will work together to define how the grid list should behave on web.
We will start from the most rigid example - 12 columns with gutter of 8px and I updated the prototype on codepen correspondingly. 

Please feel free to left your comments and thought over grid list.

yeelan0319:
[...]

PR #352, opened 2017-02-23
fix(switch): Remove checkbox references from SCSS

### LINKED ISSUE #322: Incorrect use of mdc-checkbox-* references within mdc-switch SCSS
Within mdc-switch code, there are implicit references made to `mdc-checkbox`, in two places:

- `@include mdc-theme-dark(".mdc-checkbox")` is used instead of `@include mdc-theme-dark(".mdc-switch")` ([example](https://github.com/material-components/material-components-web/blob/e0298d92e184150aa6372c3345765f12918c890b/packages/mdc-switch/mdc-switch.scss#L50)).

- `@include mdc-checkbox-transition` is used instead of a `mdc-switch-transition` mixin that should be defined ([example](https://github.com/material-components/material-components-web/blob/e0298d92e184150aa6372c3345765f12918c890b/packages/mdc-switch/mdc-switch.scss#L62)).

This hasn't bitten us I think, because mdc-checkbox et. al. is included within the sass compilation context when we bundle our assets via webpack. However, if someone were to try and bundle our sass manually, it would fail. Furthermore, I don't think `mdc-switch--theme-dark` would work properly if attached to the `mdc-switch` element.

I personally think we should maybe prioritize fixing this for this sprint cc @amsheehan

traviskaufman:
- Remove all references to mdc-checkbox in mdc-theme-dark mixin
- Add `mdc-switch-transition` function and replace calls to
  `mdc-checkbox-transition` with that function
- (techdebt) Update cross-env and cz-conventional-changelog
- (techdebt) Add allowed scopes to validate-commit-msg config, and docs
  on how to add new scopes when a component is added.

Fixes #322

### DIFF
--- a/docs/authoring-components.md
+++ b/docs/authoring-components.md
@@ -725,6 +725,10 @@ Concretely:
   contains javascript, ensure that its component namespace is exported within
   `material-components-web`, and it is registered with `mdc-auto-init`. Lastly, remember to add it
   to `package.json` of `material-components-web`.
+- Ensure that the correct **commit subject** for the package is added to the
+  `config.validate-commit-msg.scope.allowed` array within the top-level `package.json` at the root
+  of the repo. The commit subject is the _name the component, without the `mdc-`/`@material/`_.
+  E.g., for `mdc-icon-toggle`, the correct subject is `icon-toggle`.
 
 ### Testing
 
--- a/package.json
+++ b/package.json
@@ -33,9 +33,9 @@
     "chai": "^3.5.0",
     "cli-table": "^0.3.1",
     "cp-file": "^4.1.0",
-    "cross-env": "^2.0.0",
+    "cross-env": "^3.1.4",
     "css-loader": "0.14.5",
-    "cz-conventional-changelog": "^1.1.6",
+    "cz-conventional-changelog": "^2.0.0",
     "del-cli": "^0.2.0",
     "dom-events": "^0.1.1",
     "eslint": "^3.6.1",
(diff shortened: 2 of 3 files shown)


docs/authoring-components.md
```diff
@@ -725,6 +725,10 @@ Concretely:
   contains javascript, ensure that its component namespace is exported within
   `material-components-web`, and it is registered with `mdc-auto-init`. Lastly, remember to add it
   to `package.json` of `material-components-web`.
+- Ensure that the correct **commit subject** for the package is added to the
```
yeelan0319: Do you mean only if this package is added to `config.validate-commit-msg.scope.allowed` then its commit message will be checked? I tried to make a wrong commit message "feat(mdc-grid-list): Implement mdc-grid-list (#47)" but it still went through.

sgomes:
Approved!

traviskaufman:
SGTM I included it in here since it doesn't really break anything. Also I'm going to update the list to include `grid-list` so that merging in this change doesn't break #359

docs/authoring-components.md
traviskaufman: > Do you mean if this package is added to config.validate-commit-msg.scope.allowed then its commit message will be checked and if I wrote wrong commit message like "feat(mdc-grid-list): Implement mdc-grid-list (#47)" it won't go through?
That's correct!

> What about cases if I forget to add this package to that config file, will that commit message go through?
It won't. So this ensures that either you:

- Use a pre-defined scope
- Add the new component scope to the config

PR #353, opened 2017-02-23
feat(select): Add value retrieval mechanisms to JS API

### LINKED ISSUE #232: Add value property to mdc-select JS implementation
Our `mdc-select` JS component is missing the `value` getter ~/setter~ property, which is [part of the HTMLSelectElement API](https://html.spec.whatwg.org/multipage/forms.html#dom-select-value). We should implement this in order to have better parity with the native DOM counterpart.

traviskaufman:
- Add `value` property to JS component to retrieve the value of the
  currently selected option
- Add `getValue()` method to foundation in order to drive `value`
  property in component.
- Add `getValueForOptionAtIndex` foundation method.
- Enhance `verifyDefaultAdapter` foundation test helper by making it so
  that when checking for correct adapter methods, the method names don't
  have to be in the same order as they are defined on the object.

BREAKING CHANGE: **New adapter method:** `getValueForOptionAtIndex(index: string) => string`
should return the "value" of the option at the given index. Please add
this method to your adapter implementations.

Resolves #232

### DIFF
--- a/packages/mdc-select/README.md
+++ b/packages/mdc-select/README.md
@@ -49,7 +49,8 @@ import {MDCSelect} from 'mdc-select';
 
 const select = new MDCSelect(document.querySelector('.mdc-select'));
 select.listen('MDCSelect:change', () => {
-  alert(`Selected "${select.selectedOptions[0].textContent}" at index ${select.selectedIndex}"`);
+  alert(`Selected "${select.selectedOptions[0].textContent}" at index ${select.selectedIndex} ` +
+        `with value "${select.value}"`);
 });
 ```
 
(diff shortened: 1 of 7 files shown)


packages/mdc-select/README.md
```diff
@@ -293,6 +297,10 @@ within `componentDidUpdate`.
 
 ### The full foundation API
 
+#### MDCSelectFoundation.getValue() => string
+
+Returns the value of the currently selected option.
```
yeelan0319: Maybe also document when invalid what happens if user haven't made any choice.

"Returns the value of the currently selected option. If none of options is selected yet, return empty string."

traviskaufman:
@yeelan0319 changes made PTAL 👀

PR #354, opened 2017-02-23
fix(textfield): Adjust labels when initializing pre-filled textfields

### LINKED ISSUE #300: Floating label overlap when browser pre-fills text

0.2.0

Chrome

### What OS are you using?

OSX

1. Go to the [mdc-textfield demo](http://material-components-web.appspot.com/textfield.html)
1. Type any text into the `Multi-line Textfields` example such that the label moves into position
1. Click outside of the multi-line textfield to blur it
1. Press the browser back button
1. Press the browser forward button
1. Observe that the browser has filled the `Multi-line Textfield` automatically with the text you typed before. Because of this the label is now overlapping the textfield content.

1. The field should detect any browser pre-filled values and move the floating label

1. The floating label overlaps the textfield content

There is already a section in the readme on [prefilled text](https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#pre-filled-text-fields) which suggests adding the `mdc-textfield__label--float-above` class, but I feel this should be automatically handled by the component as it is a common case because the browser will pre-fill forms when navigating.

traviskaufman:
Also change cssClasses getter to use static strings rather than
templates, which will be required for future internal interop.

Fixes #300

### DIFF
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -16,19 +16,17 @@
 
 import {MDCFoundation} from '@material/base';
 
-const ROOT = 'mdc-textfield';
-
 export default class MDCTextfieldFoundation extends MDCFoundation {
   static get cssClasses() {
     return {
-      ROOT,
-      UPGRADED: `${ROOT}--upgraded`,
-      DISABLED: `${ROOT}--disabled`,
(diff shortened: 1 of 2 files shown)


traviskaufman:
@yeelan0319 rebased and split the PR into two separate commits. *This PR should be rebased and merged*

PR #356, opened 2017-02-24
feat(tabs): add tabs component

codesuki:
I am working on adding a `tabs` component. Still work in progress. 

The biggest problem I have currently is that after adding the sliding animation to the content panes right after load all the panes are sliding out of the screen except the active one (see GIF 2). 
Don't have a good idea how to fix that yet. Maybe someone has a hint? Forcing some special class on load only might fix it but seems dirty.

If I could get some feedback I would improve things and when it turns into something mergeable I would write tests. 

TODO (maybe incomplete):
- [ ] Support different screen sizes
- [ ] Scrollable tabs
- [ ] Content resizing
- [ ] Follow Material specs more closely (especially animations)

good

bug


traviskaufman:
Hey @codesuki,

So again I want to say thank you for taking the initiative on this and we really appreciate your willingness to contribute.

However, I've spent a lot of time over the past few weeks working on the Material Design tabs prototype as discussed on the issue, and unfortunately the scope and complexity is such that I would like for the core team to tackle implementing this component. For example, here is the visual prototype: http://codepen.io/traviskaufman/pen/PpwwpG?editors=0010 as you can see there is a ton of different permutations of tabs, as well as configurations that must be accommodated. Not to mention items such as RTL and dark mode support that have not been covered in the prototype. 

Furthermore, we'd like to land tabs as part of our current two-week iteration which ends next week, and I'd hate to put that pressure on you. 

That being said, if there are other components you may be interested in taking on, please let me know and I can assign myself to the issue and denote that you'll be working on them. We can then work together and I can walk you through all of the requirements of submitting new components to our codebase. We have icon buttons and expansion panels on our roadmap if you're at all interested in those. Let us know and thanks so much again!

codesuki:
No problem! I'll have a look at other tasks when I find the time.

PR #358, opened 2017-02-26
Fix SCSS import by alias

### LINKED ISSUE #351: Cannot @import scss
I really want to use your library, especially with Sass, however I can’t `@include` the `.scss` styles with my current configuration.  That is using Webpack 2 and `sass-loader`.  Here is an [example project](https://github.com/davibe/next.js-css-global-style-test).  Here is [a comment describing the issue](https://github.com/davibe/next.js-css-global-style-test/issues/3#issuecomment-281793280) - basically I can’t import the `.scss` (of button in this case) neither with relative paths nor with the `~` alias to `node_modules` that `node-sass` / `sass-loader` supports.  The problem in this case is `@import "@material/animation/variables”;`.  I expect there would be many other similar imports in the library…  Do you have / know of a working Webpack 2 example?

I thought Eyeglass would help, but I can’t set that up with Webpack 2.  Here is [an eyeglass issue](https://github.com/sass-eyeglass/eyeglass/issues/153) and [my comment on it](https://github.com/sass-eyeglass/eyeglass/issues/153#issuecomment-281724903) that has a broken version of my config trying to tell `sass-loader` to use `eyeglass` for importing.

So I'm trying this with `material-components-web@0.4.0` and more specifically `@material/button@0.2.1` importing from its dependency `@material/animation@0.1.3`.  It's also worth noting that I'm using `yarn` which keeps a flat directory structure of my dependencies.  I wonder if you assume a regular `npm install` and thus hierarchical dependencies?

Here is the example error:

```
 error  in ./pages/index.js.scss

Module build failed:
^
      File to import not found or unreadable: @material/animation/variables.
Parent style sheet: /Users/om/Dev/mel/drmelgill.com/node_modules/@material/button/mdc-button.scss
      in /Users/om/Dev/mel/drmelgill.com/node_modules/@material/button/mdc-button.scss (line 17, column 1)

 @ ./pages?entry 43:15-41
 @ multi ./~/next/dist/client/webpack-hot-middleware-client ./pages?entry
```

Either of these two would cause it, and `mdc-button.scss` is found in both cases:

- `@import '../node_modules/@material/button/mdc-button.scss’;` // valid relative path
- `@import  '~@material/button/mdc-button.scss’;` // ~ = node_modules

So I'm not really sure if this is about configuration help, documentation example, bug report, or a feature request...

KorsaR-ZN:
Fix #351

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,13 +108,12 @@ Each component also comes with a Sass source file that can be included in your a
 @import 'material-components-web';
 
 // Using individual components / mixins
-@import '@material/checkbox';
-@import '@material/typography';
-@import '@material/elevation/mixins'; // Mixins for elevation.
+@import '~@material/checkbox';
+@import '~@material/typography';
+@import '~@material/elevation/mixins'; // Mixins for elevation.
 ```
 
-> NOTE: The components' Sass files expect that the `node_modules` directory containing the
-`@material` scope folder is present on the Sass include path.
+> NOTE: "~" this advanced mechanism to resolve files by webpack documentation in [here](https://github.com/webpack-contrib/sass-loader#imports)
 
 ## Running the demos
 
--- a/packages/material-components-web/material-components-web.scss
+++ b/packages/material-components-web/material-components-web.scss
@@ -14,24 +14,24 @@
  * limitations under the License.
  */
 
-@import "@material/animation/mdc-animation";
-@import "@material/button/mdc-button";
-@import "@material/card/mdc-card";
-@import "@material/checkbox/mdc-checkbox";
-@import "@material/drawer/mdc-drawer";
-@import "@material/elevation/mdc-elevation";
-@import "@material/fab/mdc-fab";
-@import "@material/form-field/mdc-form-field";
-@import "@material/icon-toggle/mdc-icon-toggle";
-@import "@material/layout-grid/mdc-layout-grid";
-@import "@material/list/mdc-list";
-@import "@material/menu/mdc-menu";
-@import "@material/radio/mdc-radio";
-@import "@material/ripple/mdc-ripple";
-@import "@material/select/mdc-select";
-@import "@material/snackbar/mdc-snackbar";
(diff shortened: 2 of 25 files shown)


traviskaufman:
Thanks for your contribution, but unfortunately, `~` is only webpack-specific, and we want our sass to be able to be used by our standalone packages. Thus, we need to keep our import paths the way they currently are, so we cannot accept this PR.

Elemecca:
For the benefit of search engines, the description this PR should have had:

This project currently treats the namespace of Sass modules the same as Node.js' require, where module names are assumed to be absolute unless preceded by ./ or ../, so imports usually look like this:

```scss
@import "../mixins";
```

In order to do that, this was added to the Webpack sass-loader config:

```javascript
 includePaths: glob.sync('packages/*/node_modules').map((d) => path.join(__dirname, d)),
```

However, the [README for sass-loader](https://github.com/webpack-contrib/sass-loader#imports) says:

> The sass-loader uses node-sass' custom importer feature to pass all queries to the Webpack resolving engine. Thus you can import your Sass modules from `node_modules`. Just prepend them with a `~` to tell Webpack that this is not a relative import. \[...] Webpack needs to distinguish between `bootstrap` and `~bootstrap` because CSS and Sass files have no special syntax for importing relative files. Writing `@import "file";` is the same as `@import "./file";`.

Since material-components-web isn't using the standard sass-loader resolution mechanism I need to add a similar `includePaths` hack to my own Webpack config in order to avoid resolution errors for transitive dependencies when I `@import` its Sass files from my components' Sass files:

```javascript
includePaths: [path.join(__dirname, 'node_modules')],
```

I wondered why this project isn't using sass-loader's standard mechanism for Node module imports, and Travis' comment above is the answer:

> \[...] unfortunately, `~` is only Webpack-specific, and we want our Sass to be able to be used by our standalone packages. Thus, we need to keep our import paths the way they currently are \[...]

PR #365, opened 2017-02-28
fix template syntax error

guillaumevincent:
In vuejs 2.2 parsing raise template error if html element has no matching end tag


guillaumevincent:
I signed it!

PR #366, opened 2017-02-28
Fix mdc-toolbar__fixed-adjust in README

eejdoowad:
Change mdc-toolbar-fixed-adjust to mdc-toolbar__fixed-adjust in README.

### DIFF
--- a/packages/mdc-toolbar/README.md
+++ b/packages/mdc-toolbar/README.md
@@ -34,7 +34,7 @@ the top of the screen, add an `mdc-toolbar--fixed` class to the toolbar element.
 
 **Adjusting sibling elements of fixed toolbars**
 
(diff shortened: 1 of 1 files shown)


eejdoowad:
I signed it!

traviskaufman:
Looks like you're using an outdated version of mdc-toolbar. We hotfix-released an `0.1.1` version with the correct class name of `mdc-toolbar-fixed-adjust`. You can see it at http://material-components-web.appspot.com/toolbar/fixed-top-toolbar.html

AdFad666:
I also wondered why this wasn't working with `https://unpkg.com/material-components-web@latest/dist/material-components-web.min.css` thanks for pointing it out!

Issue #319, opened 2017-02-21
Improve branch coverage


sgomes:
Our branch coverage is getting close to the 95% lower limit in some browsers: 

```
=============================== Coverage summary ===============================
Statements   : 98.22% ( 1379/1404 )
Branches     : 95.29% ( 506/531 )
Functions    : 98.02% ( 642/655 )
Lines        : 98.19% ( 1354/1379 )
================================================================================
```

We should proactively add some tests to existing code.

amsheehan:
Seems like this would be a nice to have, but might be worth looking at: https://coveralls.io

traviskaufman:
@amsheehan looks like [CodeCov is recommended over coveralls](https://www.slant.co/versus/7928/7929/~coveralls_vs_codecov) ? but open to either

amsheehan:
heh nice! I have actually never seen that site before.

PR #368, opened 2017-02-28
test(infrastructure): More test coverage

BBosman:
A few tweaks to tests to increase coverage a bit.

Related to #319.

### DIFF
--- a/test/unit/mdc-base/component.test.js
+++ b/test/unit/mdc-base/component.test.js
@@ -153,6 +153,32 @@ test('#emit dispatches a custom event with the supplied data', () => {
   assert.deepEqual(evt.detail, data);
 });
 
+test('#emit dispatches a custom event with the supplied data where custom events aren\'t available', () => {
+  const root = document.createElement('div');
+  const f = new FakeComponent(root);
+  const handler = td.func('eventHandler');
+  let evt = null;
+  td.when(handler(td.matchers.isA(Object))).thenDo((evt_) => {
+    evt = evt_;
+  });
+  const data = {evtData: true};
+  const type = 'customeventtype';
+
+  root.addEventListener(type, handler);
+
+  const {CustomEvent} = window;
+  window.CustomEvent = undefined;
+  try {
+    f.emit(type, data);
+  } finally {
+    window.CustomEvent = CustomEvent;
+  }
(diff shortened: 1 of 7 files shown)


packages/mdc-base/component.js
```diff
@@ -77,6 +77,7 @@ export default class MDCComponent {
   // with the given data.
   emit(evtType, evtData) {
     let evt;
+    /* istanbul ignore else: Will never fire in Chrome */
```
traviskaufman: Meh...if it doesn't fire in certain browsers, we should hack it in our tests so it will, e.g.

```js
test('handles situation where custom events aren\'t available', () => {
  const {CustomEvent} = window;
  window.CustomEvent = undefined;
  // test code...
  window.CustomEvent = CustomEvent;
});
```

packages/mdc-drawer/util.js
```diff
@@ -54,6 +54,8 @@ export function supportsCssCustomProperties(globalObj = window) {
   if ('CSS' in globalObj) {
     return globalObj.CSS.supports('(--color: red)');
   }
+
+  /* istanbul ignore next: Will never fire in Chrome */
```
traviskaufman: similarly to above we should write a test case to simulate this condition

BBosman:
Made a few tweaks based on your comments.

New coverage:
```
=============================== Coverage summary ===============================
Statements   : 98.72% ( 1390/1408 )
Branches     : 95.9% ( 515/537 )
Functions    : 98.49% ( 651/661 )
Lines        : 98.7% ( 1365/1383 )
================================================================================
```
The numbers can't really be compared though, as I rebased the PR after making my changes.

test/unit/mdc-base/component.test.js
```diff
@@ -153,6 +153,32 @@ test('#emit dispatches a custom event with the supplied data', () => {
...
+
+  const {CustomEvent} = window;
+  window.CustomEvent = undefined;
+  try {
+    f.emit(type, data);
+  } finally {
+    window.CustomEvent = CustomEvent;
+  }
+
+  assert.isOk(evt !== null);
```
traviskaufman: nit: you can use `assert.isNotNull` here

PR #372, opened 2017-03-02
docs(select): Fix 'for' attribute of a label in demo

anton-kachurin:
Made "Dark Theme" checkbox selectable by clicking it's label in "select.html" demo


anton-kachurin:
I signed it!

PR #374, opened 2017-03-02
docs(checkbox): Fix indeterminate button in JS checkbox demo

traviskaufman:
The "Make Indeterminate" button in the JS checkbox demo is currently not
working because the "indeterminate" property is being set on the
checkbox _root_ element, rather than the native instance. This commit
fixes that.

Skipping TravisCI as this is a change to the demo only.

[ci skip]

### DIFF
--- a/demos/checkbox.html
+++ b/demos/checkbox.html
@@ -89,7 +89,7 @@
         <div class="mdc-form-field">
           <div id="mdc-js-checkbox" class="mdc-checkbox">
             <input type="checkbox"
-                   id="my-checkbox"
+                   id="native-js-checkbox"
                    class="mdc-checkbox__native-control"/>
             <div class="mdc-checkbox__background">
               <svg version="1.1"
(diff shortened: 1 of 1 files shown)


anton-kachurin:
@traviskaufman Shouldn't be MDCCheckbox instance 'indeterminate' setter used instead?

        document.getElementById('make-ind').addEventListener('click', function() {
          checkboxInstance.indeterminate = true;
        });

traviskaufman:
@anton-kachurin definitely :) Obviously I rushed this PR out. Nice catch!

demos/checkbox.html
```diff
@@ -144,15 +144,15 @@
...
 
         var formField = checkbox.parentElement;
         var formFieldInstance = new MDCFormField(formField);
-
         formFieldInstance.input = checkboxInstance;
 
+        var nativeCheckbox = checkbox.querySelector('.mdc-checkbox__native-control');
         document.getElementById('make-ind').addEventListener('click', function() {
-          checkbox.indeterminate = true;
+          nativeCheckbox.indeterminate = true;
```
amsheehan: Should use setter for instance of MDCheckbox.

Other than that LGTM

PR #410, opened 2017-03-20
fix(framework-examples): Fix VueJS snackbar example

### LINKED ISSUE #405: Vue framework snackbar example references non-existent TRANS_END_EVENT_NAME string.

> 0.6.0

> Version 58.0.3029.19 beta (64-bit)

### What OS are you using?

> Fedora 25 (Linux p5q 4.9.13-201.fc25.x86_64)

> 1. run the vue framework demo server
> 2. go to localhost:8080
> 3. Press the "show snackbar" button twice
> 4. Observe only the first snackbar message is shown

> The v-mdc-snackbar component should use the same method that the packages/snackbar/index.js class uses to determine the correct transitionend event name.

> The custom v-mdc-snackbar component is referencing a string removed when fixing issue #160, TRANS_END_EVENT_NAME. It should instead be using the method from the mdc-animation package thus:
```
      registerTransitionEndHandler(handler) {
        vm.$refs.root.addEventListener(getCorrectEventName(window, 'transitionend'), handler);
      },
      deregisterTransitionEndHandler(handler) {
        vm.$refs.root.removeEventListener(getCorrectEventName(window, 'transitionend'), handler);
      }
```

snackbar/index.js was updated in commit 4206d84b512078186485caf4abe4f47d4ce974fc  Mdc web/remove implicit dom querying (#160) 2017-01-06

pgbross:
Resolves #405

### DIFF
--- a/framework-examples/vue/src/v-mdc-snackbar/Snackbar.vue
+++ b/framework-examples/vue/src/v-mdc-snackbar/Snackbar.vue
@@ -10,8 +10,7 @@
 
 <script lang="babel">
 import { MDCSnackbarFoundation } from '@material/snackbar';
-
-const { TRANS_END_EVENT_NAME } = MDCSnackbarFoundation.strings;
+import { getCorrectEventName } from '@material/animation';
 
 export default {
   props: {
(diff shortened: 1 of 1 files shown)


pgbross:
I signed it!

pgbross:
I signed it.

PR #415, opened 2017-03-21
fix(dialog): Center dialog on screen in all supported browsers (#413)

### LINKED ISSUE #413: Dialog is not centered on screen
As per the spec:
"Simple dialogs are displayed centered vertically and horizontally on the screen."

The dialog is not currently displayed vertically centered on screen:

image

This seems to be because the `mdc-dialog__surface` is set to `bottom: 50vh`, which positions the bottom of the surface at the vertical centre of the screen, rather than center the entire surface on the screen.

**Possible fix strategies**: Use `bottom: 50vh` with `transform: translateY(50%)`, or use `bottom` with a `calc()`. Removing `bottom: 50vh` also seems to work, but may need cross-browser testing.

vivekmarakana:
Fixes #413


amsheehan:
It looks like the positioning problems are manifesting themselves in more than just vertical space. In Safari and FF I am seeing some issues with horizontal positioning as well.

The culprit appears to be `position: absolute` on `.mdc-dialog__surface`

If you can make that change as well this looks good to me and we can get it merged! Thank you so much for picking this up so quickly 🥇

vivekmarakana:
Working on this. Will update the pull request.

In the same PE, I'll also change 

```javascript
dialog.listen('MDCDialog:accept', () => {
  console.log('accepted');
})

dialog.listen('MDCDialog:cancel', () => {
  console.log('canceled');
})
```

to 

```javascript
dialog.listen('MDCDialog:accept', function() {
  console.log('accepted');
})

dialog.listen('MDCDialog:cancel', function() {
  console.log('canceled');
})
```

in `README.md` and `dialog.html` as older browser even Safari 9 does not understand arrow functions and all other function in the same file are in traditional syntax. So changing everything to traditional for the sake of consistency.

Let me know if there was any particular reason to keep these function as arrow functions.

vivekmarakana:
Made suggested changes and updated PR.

amsheehan:
@vivekmarakana these changes LGTM. We are going to have a chat with someone from design to get the okay on the 2px border radius since the spec doesn't give explicit guidance on it.

Once we get confirmation on that this looks good to merge. Thanks!

sgomes:
@amsheehan FYI, per #421 it looks like the chat already happened and the 2px border radius should be added.

amsheehan:
@sgomes Thank you sir!

PR #418, opened 2017-03-21
chore(codecov): Adds codecov coverage reports to MDC Web

### LINKED ISSUE #380: Integrate with CodeCov / Coveralls
Along the lines of #319, it would be nice to have more visibility into how every PR affects our coverage. [CodeCov](https://codecov.io/) and [Coveralls](https://coveralls.io/) are two popular options for this, and should be able to be integrated with our codebase relatively painlessly. I personally like CodeCov because it comments on PRs for you, but would be open to Coveralls as well if a compelling case can be made for it.

amsheehan:
resolves #380


package.json
```diff
@@ -20,7 +20,8 @@
     "pretest": "npm run lint",
     "test": "karma start --single-run",
     "posttest": "istanbul report --root coverage text-summary && istanbul check-coverage --lines 95 --statements 95 --branches 95 --functions 95",
-    "test:watch": "karma start --auto-watch"
+    "test:watch": "karma start --auto-watch",
+    "report-coverage": "codecov"
```
traviskaufman: Do we need to do this if we already do it within `after_success`?

amsheehan: I suppose not. If we want to generate a report from dev we can just use the `codecov` command

Issue #361, opened 2017-02-27
Ensure all custom property usage has fallbacks


sgomes:
In general, we should strive to use the following pattern as much as possible, when using custom property values in CSS:

```css
.foo {
  /* For browsers that don't understand custom properties */
  color: black;
  /* `black` used as fallback if `--color` isn't defined */
  color: var(--color, black);
}
```

Acceptable alternatives include wrapping custom property usage inside `@supports` blocks:

```css
@supports(--css-vars: 'yes') {
  /* ... */
}
```

or otherwise making sure (through JS, for example), that the browser supports custom properties before using them.

That said, it can generally be considered a code smell to have a `var()` lookup without a second parameter (the fallback).

Tasks:
- [ ] Ensure all `var()` lookups have a second parameter
- [ ] Ensure all `var()` lookups have a fallback of some sort for browsers that don't support custom properties
- [ ] Run `material-components-web` generated CSS through `postcss-custom-properties`. Remove any errors and attempt to reduce warnings as much as possible, while taking into account the limitations of the tool.

This is a follow-up to #360.

traviskaufman:
👍 👍 👍 

Perhaps we could write / contribute to a stylelint rule which enforces that all custom props have a fallback?

sgomes:
Done in #367.

PR #419, opened 2017-03-21
feat(ripple): Implement subset of improved interaction response guidelines

### LINKED ISSUE #190: Implement subset of improved interaction response guidelines for ripple
-  _Tap_ - This could be a quick mousedown/mouseup, quick keydown/keyup, quick pointerstart/pointerend, trackpad tap, etc.
- _Tap and hold_ - Mousedown/keydown/pointerstart + hold + mouseup/keyup/pointerend

The rest of the improved interaction response guidelines will be covered in separate stories.

traviskaufman:
Background
===

When the ripple was originally designed from a motion perspective,
bounded and unbounded activation/deactivation used two different
animation styles, distinct from one another.

The UX updates to the ripple remove this distinction of bounded vs.
unbounded solely with regard to how the ripple animates in and out.
There is still a distinction between bounded vs. unbounded, but we no
longer have to account for this when animating the ripple. Thus, there
are no API changes; only UX changes.

Changes
===

- Implement UX changes for "tap" + "tap and hold" interactions
- Remove all references of "bounded" vs. "unbounded" from
  activation/deactivation code

Resolves #190

### DIFF
--- a/packages/mdc-ripple/_keyframes.scss
+++ b/packages/mdc-ripple/_keyframes.scss
@@ -20,14 +20,27 @@
 
 @keyframes mdc-ripple-fg-radius-in {
   from {
-    transform: translate(0) scale(1);
+    // NOTE: For these keyframes, we do not need custom property fallbacks because they are only
+    // used in conjunction with `.mdc-ripple-upgraded`. Since MDCRippleFoundation checks to ensure
+    // that custom properties are supported within the browser before adding this class, we can
+    // safely use them without a fallback.
     transform: translate(var(--mdc-ripple-fg-translate-start, 0)) scale(1);
     animation-timing-function: $mdc-animation-fast-out-slow-in-timing-function;
   }
 
   to {
-    transform: translate(0) scale(0);
-    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 0));
+    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
+  }
+}
+
+@keyframes mdc-ripple-fg-opacity-in {
+  from {
+    opacity: 0;
+    animation-timing-function: linear;
+  }
+
+  to {
+    opacity: 1;
   }
 }
 
@@ -39,27 +52,5 @@
 
   to {
     opacity: 0;
-  }
-}
-
-@keyframes mdc-ripple-fg-unbounded-opacity-deactivate {
-  from {
-    opacity: 1;
-  }
-
-  to {
-    opacity: 0;
-  }
-}
-
-@keyframes mdc-ripple-fg-unbounded-transform-deactivate {
-  from {
-    transform: 0;
-    transform: var(--mdc-ripple-fg-approx-xf, 0);
-  }
-
-  to {
-    transform: scale(0);
-    transform: scale(var(--mdc-ripple-fg-scale, 0));
   }
 }
--- a/packages/mdc-ripple/_mixins.scss
+++ b/packages/mdc-ripple/_mixins.scss
@@ -8,7 +8,8 @@
 //      http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS, // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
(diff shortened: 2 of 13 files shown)


traviskaufman:
@cristobalchao if you could do a quick pass on the UX here I'd appreciate it!

packages/mdc-ripple/_keyframes.scss
```diff
@@ -20,46 +20,33 @@
 
 @keyframes mdc-ripple-fg-radius-in {
   from {
-    transform: translate(0) scale(1);
```
BBosman: Why is this removed? Don't we need a fallback for browsers that don't support var?

packages/mdc-ripple/_keyframes.scss
```diff
@@ -20,46 +20,33 @@
...
     transform: translate(var(--mdc-ripple-fg-translate-start, 0)) scale(1);
     animation-timing-function: $mdc-animation-fast-out-slow-in-timing-function;
   }
 
   to {
```
BBosman: Same here.

packages/mdc-ripple/_mixins.scss
```diff
@@ -215,29 +205,13 @@
...
-  &.mdc-ripple-upgraded--unbounded.mdc-ripple-upgraded--foreground-unbounded-deactivation#{$pseudo} {
-    animation:
-      mdc-ripple-fg-unbounded-opacity-deactivate 0 linear,
-      mdc-ripple-fg-unbounded-transform-deactivate 0 $mdc-animation-fast-out-slow-in-timing-function;
-    animation:
-      mdc-ripple-fg-unbounded-opacity-deactivate var(--mdc-ripple-fg-unbounded-opacity-duration, 0) linear,
-      mdc-ripple-fg-unbounded-transform-deactivate var(--mdc-ripple-fg-unbounded-transform-duration, 0) $mdc-animation-fast-out-slow-in-timing-function;
+  &.mdc-ripple-upgraded--foreground-deactivation#{$pseudo} {
+    // Retain transform from mdc-ripple-fg-radius-in activation
+    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
```
BBosman: Same here.

packages/mdc-ripple/_keyframes.scss
BBosman: See #361.

traviskaufman: @BBosman good question. Typically, yes. However, this keyframe is only attached when `mdc-ripple-upgraded` is added to the ripple surface. Before we add that class, we check - within JS - to make sure custom properties are supported. Thus, fallbacks here aren't necessary.

...I'm going to add this explanation as a comment 😃

packages/mdc-ripple/_keyframes.scss
traviskaufman: [ack](https://github.com/material-components/material-components-web/pull/419#discussion_r107239584)

packages/mdc-ripple/_mixins.scss
traviskaufman: [ack](https://github.com/material-components/material-components-web/pull/419#discussion_r107239584)

packages/mdc-ripple/_keyframes.scss
BBosman: 👍

packages/mdc-ripple/constants.js
```diff
@@ -15,42 +15,33 @@
  */
 
 export const ROOT = 'mdc-ripple';
-export const UPGRADED = `${ROOT}-upgraded`;
```
yeelan0319: nit: It seems like the exported const `ROOT` and `UPGRADED` is not being used anywhere.

traviskaufman:
@yeelan0319 Changes made PTAL!

packages/mdc-ripple/_keyframes.scss
```diff
@@ -20,46 +20,37 @@
...
+    // that custom properties are supported within the browser before adding this class, we can
+    // safely use them without a fallback.
     transform: translate(var(--mdc-ripple-fg-translate-start, 0)) scale(1);
     animation-timing-function: $mdc-animation-fast-out-slow-in-timing-function;
   }
 
   to {
-    transform: translate(0) scale(0);
-    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 0));
+    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
```
yeelan0319: Is there any reason why the default value is different from `--mdc-ripple-fg-scale` default value (which is 0 and default to 0 else where in `_mixins.scss`). Is this a spec change?

packages/mdc-ripple/_mixins.scss
yeelan0319: Same thing here.

yeelan0319:
@traviskaufman 
Brought this up in our stand up today as a parking lot since it seems like there is no further information about why `bounded` & `unbounded` is removed and I feel I am lacking of quite a few context information to give useful feedbacks on those implementation.

Moreover since so many changes have been made, I also want to make sure there is no changing to our API and we don't need to make any updates to `README.md`. 


Given we might need to review code across different components and git doesn't always clever enough to show diff correctly, I feel it might be helpful to provide slightly more high level description about the commit, I found a [guideline](https://github.com/blog/1943-how-to-write-the-perfect-pull-request) on Github that we might found useful.

traviskaufman:
@yeelan0319 

When the ripple was originally designed from a motion perspective, `bounded` and `unbounded` activation/deactivation used two different animation styles, distinct from one another. 

The UX updates to the ripple _remove_ this distinction of `bounded` vs. `unbounded` _solely with regard to how the ripple animates in and out_. There is still a distinction between bounded vs. unbounded, but we no longer have to account for this when animating the ripple. Thus, there are no API changes; only UX changes. 

I'll go ahead and add this description above to the commit message.

packages/mdc-ripple/_keyframes.scss
traviskaufman: Looks like an oversight on my part for the `--mdc-ripple-fg-scale` default value, which should be set to `1`. Will change.

packages/mdc-ripple/_mixins.scss
traviskaufman: [ack](https://github.com/material-components/material-components-web/pull/419/files#r108443126)

traviskaufman:
@yeelan0319 changes made PTAL

yeelan0319:
Cool. Thanks Travis. This feedback is useful 👍 🌟

PR #423, opened 2017-03-22
fix(drawer): Export util methods in index.js

### LINKED ISSUE #422: Package mdc-drawer missing export 'util'

> 0.7.0

>Version 58.0.3029.19 beta (64-bit)

### What OS are you using?

> Fedora 25 (Linux p5q 4.9.13-201.fc25.x86_64)

> The mdc-drawer util methods should be available for users of the @material/drawer package (from the dist/mdc.drawer.js code) without having to import util from the node_modules folder separately. Note this change is also needed when no module system is used under the global `mdc` namespace where there is no workaround.

> The util methods are unavailable and need to be imported separately when writing, say, a framework integration example.

The mdc-dialog package does this correctly by including:
```
export{util};
```
in its index.js file, and the same should be done in mdc-drawer/index.js.

pgbross:
Resolves #422


amsheehan:
This looks good to me. Please rebase with a fresh pull on master so we can try and get codecov to pass. Thanks for the PR!

amsheehan:
Please add documentation to the README explaining that users can now use `mdc-drawer/util`. Thanks!

PR #427, opened 2017-03-23
fix(layout-grid): Prevent overflowing components within cells

### LINKED ISSUE #406: Image overflowing the cell div

material-components-web@0.6.0

>This bug is affecting Firefox 52.
Mozilla/5.0 (Windows NT 10.0; WOW64; rv:52.0) Gecko/20100101 Firefox/52.0

### What OS are you using?

>Windows 10 Version 1607

http://codepen.io/anon/pen/EWojwo

>The image should not overflow the cell div 

>The image overflow the cell div so it is superposing iwith the other cell div

>There is no problem on Chrome 56 and Edge 14. 
It seems to be due to a width: auto;.

vivekmarakana:
- Adding `max-height: 100%` and `max-width: 100%` to prevent overflowing components
- Also adding `overflow:hidden` in case it is still overflowing

Fixes #406

### DIFF
--- a/packages/mdc-layout-grid/_mixins.scss
+++ b/packages/mdc-layout-grid/_mixins.scss
@@ -48,6 +48,9 @@
 
   @supports (display: grid) {
     width: auto;
+    max-width: 100%;
+    max-height: 100%;
+    overflow: hidden;
     grid-column-end: span min($span, map-get($mdc-layout-grid-columns, $size));
   }
 }


packages/mdc-layout-grid/_mixins.scss
```diff
@@ -48,6 +48,9 @@
 
   @supports (display: grid) {
     width: auto;
+    max-width: 100%;
+    max-height: 100%;
+    overflow: hidden;
```
amsheehan: @yeelan0319 Is this clipping the correct behavior? Do we want the image to be contained or covering the parent?

vivekmarakana: I ran into an issue while fixing this. If there's any other component inside the cell, then the child cell will still overflow as it has 100% max-height.

Check out this pen: http://codepen.io/vivekmarakana/pen/VpxqJa

amsheehan: That makes sense. What worries me about using `max-` width/height for this as a fix is that I can't see anyone being happy with their image being stretched. What if we use `min-` width/height? That way the aspect ratio of the image is preserved, even if the crop and resolution is unfavorable.

Would like to get @traviskaufman and @yeelan0319 to weigh in. 

Looks good so far though, thanks!

amsheehan: We'd also want `overflow: hidden` on `mdc-layout-grid__cell`, no?

traviskaufman: okay, so, backing _way_ up here:

What's the reason for attaching `mdc-layout-grid__cell` to the `` element within an additional `mdc-layout-grid__cell` element? Cells aren't supposed to be nested within other cells; that behavior is actually undefined.

traviskaufman:
Closing as issue was fixed by a different solution.

PR #429, opened 2017-03-23
docs(dialog): Fix broken link

jscissr:


amsheehan:
@jscissr Please rebase on a fresh version of master so we can get CodeCov to pass before merging. We are aware that CodeCov is a little bit flakey on two tests in particular and have a plan in motion to remedy that. For now, rebasing should work to get a green light from CodeCov.

Thanks!

jscissr:
There are no new commits on master, rebasing would be a no-op, so I'm not sure I understand your request. I see there is #431, but this has not been merged yet.

PR #430, opened 2017-03-23
refactor: Omit extraneous svg attributes

jscissr:
Optimized using https://github.com/svg/svgo
The `xmlns` attribute can be removed for inline svg, see https://github.com/svg/svgo/pull/530


traviskaufman:
Have you tested this in IE11 and Edge? We've had some weirdness with SVG in the past with those browsers specifically, which is why we included those exact attributes.

jscissr:
I have now tested the demo in IE11 and seen no issues. CI is also ok. Maybe you can test it in Edge?

traviskaufman:
NOTE: _Not_ marking this as a breaking change, even though _technically_ our HTML structure has changed, due to the fact that removing this attribute does not change anything about the DOM.

PR #439, opened 2017-03-26
docs(list): fix error in two line list item example in Readme

srinathh:
As far as I can make out from the CSS, there is no `mdc-list-item__text__primary` defined and I assume this should simply be `mdc-list-item__text`. The second line should have the class `mdc-list-item__text__secondary`

### DIFF
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -71,8 +71,8 @@ While in theory you can add any number of "lines" to a list item, you can use th
 <ul class="mdc-list mdc-list--two-line">
   <li class="mdc-list-item">
     <span class="mdc-list-item__text">
-      <span class="mdc-list-item__text__primary">Two-line item</span>
-      <span class="mdc-list-item__text__primary">Secondary text</span>
+      <span class="mdc-list-item__text">Two-line item</span>
+      <span class="mdc-list-item__text__secondary">Secondary text</span>
     </span>
   </li>
 </ul>
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -333,7 +333,7 @@ The html for this can be easily added
 <ul class="mdc-list mdc-list--two-line msgs-list">
   <li class="mdc-list-item">
     <span class="mdc-list-item__text">
-      <span class="mdc-list-item__text">Ali Connors</span>
+      Ali Connors
       <span class="mdc-list-item__text_secondary">Lunch this afternoon? I was...</span>
     </span>
 
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -71,7 +71,7 @@ While in theory you can add any number of "lines" to a list item, you can use th
 <ul class="mdc-list mdc-list--two-line">
   <li class="mdc-list-item">
     <span class="mdc-list-item__text">
-      <span class="mdc-list-item__text">Two-line item</span>
+      Two-line item
       <span class="mdc-list-item__text__secondary">Secondary text</span>
     </span>
   </li>
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -333,7 +333,7 @@ The html for this can be easily added
 <ul class="mdc-list mdc-list--two-line msgs-list">
   <li class="mdc-list-item">
     <span class="mdc-list-item__text">
-      <span class="mdc-list-item__text__primary">Ali Connors</span>
+      <span class="mdc-list-item__text">Ali Connors</span>
       <span class="mdc-list-item__text_secondary">Lunch this afternoon? I was...</span>
     </span>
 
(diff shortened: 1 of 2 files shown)


amsheehan:
@srinathh It looks like the demo for lists uses `mdc-list-item__text__primary` pretty heavily. As part of this pull request, can you also change the demo to the correct classnames? Thanks!

srinathh:
Hmmm... but in the compiled CSS file I didn't find a definition for __
primary. Are you certain it's defined?

srinathh:
@traviskaufman i have just rechecked the [material-components-web.css](https://unpkg.com/material-components-web@0.7.0/dist/material-components-web.css) and there is no `mdc-list-item__text__primary` defined in the style sheet. There's only 'mdc-list-item__text` and `mdc-list-item__text__secondary` defined. The relevant section of the file is pasted below.

If the intention was to have a `mdc-list-item__text__primary` defined then this should probably be an issue to be filed. If the intention was to have just `mdc-list-item__text` as the default text size and `mdc-list-item__text__secondary` as the smaller text size (which does work), then the documentation fix above would be correct and I can fix the other pieces that @amsheehan has mentioned before re-submitting.

```
  .mdc-list-item__text {
    display: -webkit-inline-box;
    display: -ms-inline-flexbox;
    display: inline-flex;
    -webkit-box-orient: vertical;
    -webkit-box-direction: normal;
        -ms-flex-direction: column;
            flex-direction: column; }
    .mdc-list-item__text__secondary {
      font-family: Roboto, sans-serif;
      -moz-osx-font-smoothing: grayscale;
      -webkit-font-smoothing: antialiased;
      font-size: 0.875rem;
      font-weight: 400;
      letter-spacing: 0.04em;
      line-height: 1.25rem;
      color: rgba(0, 0, 0, 0.54);
      color: var(--mdc-theme-text-secondary-on-background, rgba(0, 0, 0, 0.54)); }
      .mdc-list-item__text__secondary--theme-dark,
      .mdc-theme--dark .mdc-list-item__text__secondary {
        color: rgba(255, 255, 255, 0.7);
        color: var(--mdc-theme-text-secondary-on-dark, rgba(255, 255, 255, 0.7)); }
      .mdc-list--dense .mdc-list-item__text__secondary {
        font-size: inherit; }
```

traviskaufman:
@srinathh You're totally right. Sorry about this confusion and thanks for bearing with us 😅 

The _correct_ structure for a two-line list would be as follows:

```html
<ul class="mdc-list mdc-list--two-line">
  <li class="mdc-list-item">
    <span class="mdc-list-item__text">
      Two-line item
      <span class="mdc-list-item__text__secondary">Secondary text</span>
    </span>
  </li>
</ul>
```

I believe that when I authored this demo I wanted to draw a distinction between "primary" and "secondary" text, so I added a sentinel `__text__primary` element to disambiguate this. However, I clearly did not wind up using that class 😄. Thus, this is the correct DOM structure.

Would you mind copying and pasting over `demos/list.html` with [this gist](https://gist.github.com/traviskaufman/838abc08652c654f6e985b2a6465a8a2), which is an updated version of the list demo page that uses the correct DOM structure (so that you don't have to go in and do it manually). Once that's done I'll approve this PR.

Thanks again for your patience!

packages/mdc-list/README.md
```diff
@@ -71,8 +71,8 @@ While in theory you can add any number of "lines" to a list item, you can use th
 <ul class="mdc-list mdc-list--two-line">
   <li class="mdc-list-item">
     <span class="mdc-list-item__text">
-      <span class="mdc-list-item__text__primary">Two-line item</span>
-      <span class="mdc-list-item__text__primary">Secondary text</span>
+      <span class="mdc-list-item__text">Two-line item</span>
```
traviskaufman: Remove the `<span>` elements wrapping `Two-line item` here and below (this matches what's in the demo)

packages/mdc-list/README.md
```diff
@@ -333,7 +333,7 @@ The html for this can be easily added
 <ul class="mdc-list mdc-list--two-line msgs-list">
   <li class="mdc-list-item">
     <span class="mdc-list-item__text">
-      <span class="mdc-list-item__text__primary">Ali Connors</span>
+      <span class="mdc-list-item__text">Ali Connors</span>
```
traviskaufman: ...could we remove the `<span>` here as well? 😅

srinathh:
Oh right there was the example at the end of the readme as well. Will do
:-)

So if I understand right, the recommendation is to have secondary text as a
child of primary?

traviskaufman:
@srinathh there is no "primary" css class; secondary text is a child of `mdc-list__item__text`. You can simply copy the DOM structure outlined [here](https://github.com/material-components/material-components-web/pull/439#issuecomment-291546258)

PR #440, opened 2017-03-26
Fix Server Side Rendering for Ripple

posva:
Currently, the Ripple tries to use HTMLElement upon inclusion. However, this breaks SSR because the variable doesn't exist in Node. I moved the computation of MATCHES inside of the function using it (since there's only one) but it's also possible to check if in browser when populating the MATCHES variable

### DIFF
--- a/packages/mdc-ripple/index.js
+++ b/packages/mdc-ripple/index.js
@@ -18,7 +18,7 @@ import {MDCComponent} from '@material/base';
 import MDCRippleFoundation from './foundation';
 import {supportsCssVariables, getMatchesProperty} from './util';
 
-const MATCHES = getMatchesProperty(HTMLElement.prototype);
+let MATCHES;
 
 export {MDCRippleFoundation};
 
@@ -33,6 +33,9 @@ export class MDCRipple extends MDCComponent {
   }
 
   static createAdapter(instance) {
+    // Allow server side rendering rendering by calling HTMLElement, only when
+    // it's needed
+    MATCHES = MATCHES || getMatchesProperty(HTMLElement.prototype);
     return {
       browserSupportsCssVars: () => supportsCssVariables(window),
       isUnbounded: () => instance.unbounded,


traviskaufman:
Hey @posva,

While MDC-Web _[foundations](https://github.com/material-components/material-components-web/blob/master/docs/architecture.md#components-foundations-adapters)_ support being loaded and rendered via SSR, _components_ are not. They expect a DOM environment in order to be correctly used. I'm skeptical to merge this PR, because I don't want to make the assumption that our vanilla components support SSR. Thus, I'm closing this for now.

If this is something you feel is desirable, please open up an issue detailing why you think our vanilla components should be able to be rendered on the server, as well as why simply using our foundation classes for this don't suit your needs. Thanks!

posva:
I'll test if using the foundation only is enough to have one single package that can be required on node and browser (with a bundler) as well

posva:
Hey @traviskaufman, trying to use the foundation, it's a bit unclear to me. I understood I need to create an adapter as the one created in the MDCRipple class:

```js
// extracted from ripple/index.js
static createAdapter(instance) {
    return {
      browserSupportsCssVars: () => supportsCssVariables(window),
      isUnbounded: () => instance.unbounded,
      isSurfaceActive: () => instance.root_[MATCHES](':active'),
      addClass: (className) => instance.root_.classList.add(className),
      removeClass: (className) => instance.root_.classList.remove(className),
      registerInteractionHandler: (evtType, handler) => instance.root_.addEventListener(evtType, handler),
      deregisterInteractionHandler: (evtType, handler) => instance.root_.removeEventListener(evtType, handler),
      registerResizeHandler: (handler) => window.addEventListener('resize', handler),
      deregisterResizeHandler: (handler) => window.removeEventListener('resize', handler),
      updateCssVariable: (varName, value) => instance.root_.style.setProperty(varName, value),
      computeBoundingRect: () => instance.root_.getBoundingClientRect(),
      getWindowPageOffset: () => ({x: window.pageXOffset, y: window.pageYOffset}),
    };
  }
```

So, I'll just be copying the source code of that file, which feels wrong since I'll have to follow any update of it. Or maybe that file won't change in the future so it's ok to copy that part?

traviskaufman:
Hey @posva,

It's totally appropriate to copy adapter code for your implementation. However, some adapter code may be different (e.g. how you reference an element), which is why we have that architecture in the first place. All of our adapter APIs are strictly versioned, so if there is a change to an adapter method that's needed it will be clearly outlined by a version bump and a Breaking change line item in our changelog.

Also you can always reach out to us with questions either on StackOverflow (using the tags `material-components` and `web` together) or on our Discord channel https://discord.gg/pRKaJB9

posva:
Hey @traviskaufman Thanks a lot for the answer. Now I understand how good the adapters are, I didn't realise how flexible they were when I saw them for the first time! I joined the discord 😄 
Keep up the good work!

PR #448, opened 2017-03-29
feat(toolbar): Improve toolbar to support multiple row.

yeelan0319:
Background
===

When toolbar was originally implemented, it was design to be a single
row toolbar. However, it turns out that we need one higher level
wrapper to group the current toolbar and future tabs. We carefully
balanced between introducing a new "appbar" concept v.s. extending the
functionality of existing toolbar to support multiple row, we decided
to go for the latter without introducing a new component.

Changes
===
- Implement multiple row toolbar
- Updated README and demo

BREAKING CHANGE: All existing toolbar need to add `<div class="mdc-toolbar__row">`
to properly align its contents.

### DIFF
--- a/packages/mdc-toolbar/README.md
+++ b/packages/mdc-toolbar/README.md
@@ -1,10 +1,9 @@
 # MDC Toolbar
 
-MDC Toolbar provides a RTL-aware Material Design toolbar component adhering to the
-[Material Design toolbar spec](https://www.google.com/design/spec/components/toolbars.html)
-Toolbars scroll with content by default, but supports fixed on
-top as well. Currently, this component does not yet support the "Waterfall" or
-"Flexible Header" patterns.
+MDC Toolbar acts as a container for multiple rows containing items such as
+application title, navigation menu, and tabs, among other things. Toolbars
+scroll with content by default, but supports fixed on top as well. Currently,
+this component does not yet support the "Waterfall" or "Flexible Header" patterns.
 
 
 ## Installation
@@ -20,12 +19,31 @@ Wrap the items with `mdc-toolbar` class in following way:
 
 ```html
 <header class="mdc-toolbar">
-  <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
-    <a class="material-icons">menu</a>
-    <span class="mdc-toolbar__title">Title</span>
-  </section>
+  <div class="mdc-toolbar__row">
+    <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
+      <a class="material-icons">menu</a>
+      <span class="mdc-toolbar__title">Title</span>
+    </section>
+  </div>
 </header>
 ```
+
+MDC Toolbars can accommodate multiple rows using the wrapper `mdc-toolbar__row`:
+
+```html
+<header class="mdc-toolbar">
+  <div class="mdc-toolbar__row">
+    <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
+      <a class="material-icons">menu</a>
+      <span class="mdc-toolbar__title">Title</span>
+    </section>
+  </div>
+  <div class="mdc-toolbar__row">
+    ...
+  </div>
+</header>
+```
+
 
 ### Fixed toolbars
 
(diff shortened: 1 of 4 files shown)


packages/mdc-toolbar/README.md
```diff
@@ -20,13 +19,32 @@ Wrap the items with `mdc-toolbar` class in following way:
...
+  <div class="mdc-toolbar__row">
+    <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
+      <a class="material-icons">menu</a>
+      <span class="mdc-toolbar__title">Title</span>
+    </section>
+  </div>
 </header>
 ```
 
+You can wrap second row such as mdc-tabs within mdc-toolbar:
```
amsheehan: Perhaps something like:

MDC Toolbars can accommodate multiple rows using the wrapper `mdc-toolbar__row`

packages/mdc-toolbar/mdc-toolbar.scss
```diff
@@ -61,12 +69,13 @@ $mdc-toolbar-mobile-breakpoint: 599px;
...
 
-    .mdc-toolbar__title {
-      @include mdc-typography(title);
+  &__title {
+    @include mdc-typography(title);
 
-      margin: 0;
-    }
+    margin: 0;
+    line-height: 1.5rem;
```
amsheehan: Is there something in `mdc-typography` that sets `line-height`?

packages/mdc-toolbar/README.md
```diff
@@ -1,8 +1,7 @@
 # MDC Toolbar
 
-MDC Toolbar provides a RTL-aware Material Design toolbar component adhering to the
-[Material Design toolbar spec](https://www.google.com/design/spec/components/toolbars.html)
-Toolbars scroll with content by default, but supports fixed on
+MDC Toolbar provides a multiple row container for application navigation, title
```
traviskaufman: How about:

"MDC Toolbar allows the use of multiple row containers for items such as application title, navigation menu, and tabs, among other things."

packages/mdc-toolbar/README.md
```diff
@@ -20,13 +19,32 @@ Wrap the items with `mdc-toolbar` class in following way:
...
+```html
+<header class="mdc-toolbar">
+  <div class="mdc-toolbar__row">
+    <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
+      <a class="material-icons">menu</a>
+      <span class="mdc-toolbar__title">Title</span>
+    </section>
+  </div>
+  <div class="mdc-toolbar__row">
+    <nav class="mdc-tabs">...</nav>
```
traviskaufman: Skeptical to put this in here just yet, since `mdc-tabs` is not yet complete.

packages/mdc-toolbar/mdc-toolbar.scss
```diff
@@ -21,7 +21,7 @@
 
 $mdc-toolbar-height: 64px;
```
traviskaufman: Should we rename these two variables to `$mdc-toolbar-row-height` and `$mdc-toolbar-mobile-row-height` respectively?

packages/mdc-toolbar/mdc-toolbar.scss
yeelan0319: Yes. But that line-height don't fits this case. The default line height was 2rem.

packages/mdc-toolbar/README.md
yeelan0319: Or how about 

"MDC Toolbar is a multiple row containers for items such as application title, navigation menu, and tabs, among other things."

Since `mdc-toolbar` itself is a container which contains `mdc-toolbar__row`.

packages/mdc-toolbar/README.md
yeelan0319: Is it likely that `mdc-tabs` be included in next release? If not, I will take it down and we can modify it later @amsheehan

packages/mdc-toolbar/README.md
traviskaufman: SG, but one grammar nit:

"MDC Toolbar acts as a container for multiple rows containing items such as..."

packages/mdc-toolbar/README.md
traviskaufman: Hopefully! But I'd rather not include this in here until it's landed, given that it may confuse users coming to this documentation for the first time when they don't immediately see `mdc-tabs` in our codebase.

yeelan0319: That's definitely valid consideration. Take it down now!

packages/mdc-toolbar/mdc-toolbar.scss
amsheehan: Got it

PR #449, opened 2017-03-29
docs(getting-started): remove link to issue #20

TomTasche:
the issue is closed now and merged to master, so the note should not be necessary anymore

please correct me if i'm wrong!


TomTasche:
I signed it!

PR #457, opened 2017-03-30
docs: Add closure walkthrough + conventions guide

### LINKED ISSUE #326: Determine and Document Type System and conventions for writing closure code
This is the "first" story within the Closure Compiler milestone. The goal of this story is to provide a foundation so that ourselves and contributors can build out closure annotations with a solid set of conventions.

- [x] Evaluate general approach taken in https://github.com/material-components/material-components-web/tree/experimental/closure-research (see `mdc-icon-toggle` and all dependents)
- [x] Debate and solidify all parts of the approach including:
  - Type heirarchy
  - File structure
  - Naming Conventions
- [x] Verify our approach by running ad-hoc tests on closurized code in `experimental/closure-research` branch: build the mdc-icon-toggle JS, build simple js entry point for closure to put on icon-toggle demo page, swap out mdc-web with that closure entry point. _Ensure that within that entry point, icon toggle API is being programatically accessed_.
- [ ] Create `docs/closure-compiler.md` detailing all of these conventions

---

## Preliminary Brainstorm on Conventions

- Attempt to make MDCComponent `@abstract` ~_(RESEARCH)_~
  - Follow-up: Can be done by removing the default `attachTo()` method, which we could document in code, in writing, and potentially write linter rules for.
- Investigate making MDCFoundation `@abstract` ~_(RESEARCH)_~
  - Follow-up: Can be done.
- Preserve type heirarchy from experimental branch
  - Adapters are `{!Object}` types
  - `MDCFoundation` parameterized by its adapter
    
  ```js
       /** @final @extends MDCFoundation<!MDCFooAdapter> */
       class MDCFooFoundation extends MDCFoundation { /* ... */ }
   ```
  -   `MDCComponent` parameterized by its foundation
    
  ```js
       /** @final @extends MDCComponent<!MDCFooFoundation> */
       class MDCFoo extends MDCComponent { /* ... */ }
   ```

- `@typedefs` are always `let` declarations, always pascal case, and always end in `Type`.
  ```js
  /** 
   * @typedef {{foo: string, bar: number}}
   */
  let EventDataType;
  ```

- All foundation/component classes that extend from `MDCBase` and are not meant to be extended _must be marked `@final`_. This will prevent unintended subclassing which would lead to easily-breakable client code.

- Objects that use event names or other external symbols as keys _must_ be declared as `@dict`

  ```js
  /** @dict */
  const activationDeactivationPairs = {
    'mousedown': 'mouseup',
    'touchstart': 'touchend',
  };
  ```
- Adapters always declared as `@record` objects
- File structure:
    - adapter.js for adapter
    - foundation.js for foundation 
    - index.js for component
- Eslint-disable for unused var warnings / unused import warnings when it comes to @typedefs, @record imports, etc.
    - Crude
    - Use ignore patterns instead
        - http://eslint.org/docs/rules/no-unused-vars 
        - Anything matching “MDC\<PascalCase\>Adapter” _(CONVENTION)_
        - Anything matching “\<PascalCase\>Type” _(CONVENTION)_
- All getter/setter properties need to be specified in the constructor
  - _EDIT_: This may not be the case anymore, see [below comment](https://github.com/material-components/material-components-web/issues/326#issuecomment-283769462)
- `--jscomp_off accessControls` :(
  - https://github.com/google/closure-compiler/issues/2261 
  - Short-term we disable
  - We should work with closure team and fix
- No using re-exported modules
  - BAD: `import {MDCFoundation} from ‘@material/base’`;
  - GOOD: `import MDCFoundation from ‘@material/base/foundation’`;

- `cssClasses` / `strings` / `numbers` are defined in `constants.js` (and transitively in `foundation.js`) as follows:
  - `cssClasses` => `/** @enum {string} */`
  - `strings` => `/** @enum {string} */`
  - `numbers` => `/** @enum {number} */`

traviskaufman:
- Add `docs/closure-compiler.md`, detailing our motivation for using
  closure compiler, our conventions, a full example skeleton of a
  closurized MDC-Web component, and common closure idioms that may not
  be as familiar to those who haven't worked with closure before.
- Update `authoring-components.md` to include information regarding
  closure compatibility.
- Amend `authoring-components.md` to specify that
  cssClasses/strings/numbers constants must be specified in a
  `constants.js` file. While not directly related to closure, we
  reference this a lot in our documentation. Thus, it was deemed
  appropriate to add as part of this commit.
- Add `closureWhitelist` config key to `package.json`, since it's
  referenced in these docs. This will be used for eventual build testing
  as we annotate all of our components for closure.

Resolves #326
[ci skip]

---

**NOTE**: Looking for a 👍 from all Reviewers here, as this is going to lay the groundwork for our closure code moving forward. 

**NOTE**: You'll notice I specifically did _not_ include `adapter.js` as part of the files to be specified for a component within `authoring-components.md`. This is because those files do not yet exist and I don't want to confuse people. Once they exist, I can update the authoring-components doc. If you feel it _should_ be added at this point, I can do so and just make a note as to why it's not in any files yet.

### DIFF
--- a/docs/authoring-components.md
+++ b/docs/authoring-components.md
@@ -38,6 +38,7 @@ may be subject to change. They will stabilize as we near towards a full release.
      * [Define all exported CSS classes, strings, and numbers as foundation constants.](#define-all-exported-css-classes-strings-and-numbers-as-foundation-constants)
      * [Extend components and foundations from mdc-base classes.](#extend-components-and-foundations-from-mdc-base-classes)
      * [Packages must be registered with our build infrastructure, and with material-components-web pkg](#packages-must-be-registered-with-our-build-infrastructure-and-with-material-components-web-pkg)
+     * [Closure Compatibility](#closure-compatibility)
   * [Testing](#testing)
      * [Verify foundation's adapters](#verify-foundations-adapters)
      * [Use helper methods](#use-helper-methods)
@@ -469,9 +470,10 @@ A typical component within our codebase looks like so:
 packages
   ├── mdc-component
       ├── README.md # The component's README
+      ├── constants.js # The component's cssClasses/strings/numbers constants
       ├── foundation.js # The component's foundation class
       ├── index.js # The file that contains the vanilla component class, as well as exports the foundation
-      ├── mdc-textfield.scss # The main source file for the component's CSS
+      ├── mdc-component.scss # The main source file for the component's CSS
       └── package.json # The components package file
 test/unit
   ├── mdc-component
@@ -675,33 +677,47 @@ the aforementioned rule, we would also like to provide lint rules to enforce the
   getter.
 - All strings that are used outside the context of the foundation class (CSS selectors, text that
   could potentially be localized, etc.) must be referenced by a `strings` static getter.
-- All semantic numbers leveraged by the adapter (timeout lengths, transition durations, etc.) must
+- All semantic numbers leveraged by the foundation (timeout lengths, transition durations, etc.) must
   be referenced by a `numbers` static getter.
+- These constants should be defined within a `constants.js` file, and then proxied through the
+  foundation.
 
 ```js
+// constants.js
+
+export const cssClasses = {
+  ROOT: 'mdc-new-component',
+  ACTIVE: 'mdc-new-component--active',
+  DISABLED: 'mdc-new-component--disabled',
+};
+
+export const strings = {
+  CHILD_SELECTOR: '.mdc-new-component__child',
+};
+
+export const numbers = {
+  DEFAULT_THROTTLE_DELAY_MS: 300,
+};
+
+// foundation.js
+
+import {cssClasses, strings, numbers} from './constants';
+
 class MDCNewComponentFoundation extends MDCFoundation {
   static get defaultAdapter() {
     // ...
   }
 
   static get cssClasses() {
-    return {
-      ROOT: 'mdc-new-component',
-      ACTIVE: 'mdc-new-component--active',
-      DISABLED: 'mdc-new-component--disabled'
-    };
+    return cssClasses;
   }
 
   static get strings() {
-    return {
-      CHILD_SELECTOR: '.mdc-new-component__child'
-    };
+    return strings;
   }
 
   static get numbers() {
-    return {
-      DEFAULT_THROTTLE_DELAY_MS: 300
-    };
+    return numbers;
   }
 }
 ```
@@ -729,6 +745,17 @@ Concretely:
   `config.validate-commit-msg.scope.allowed` array within the top-level `package.json` at the root
   of the repo. The commit subject is the _name the component, without the `mdc-`/`@material/`_.
   E.g., for `mdc-icon-toggle`, the correct subject is `icon-toggle`.
+- Ensure that the package name is added to the `closureWhitelist` array within the top-level
+  `package.json`.
+
+#### Closure Compatibility
+
+> NOTE: This section was introduced as part of our [closure compatibility milestone](https://github.com/material-components/material-components-web/milestone/4). Our
+currently existing components are in the process of being made compatible with closure.
+
+All core MDC-Web components must be fully compatible with the Google Closure Compiler using its
+advanced compilation mechanisms. We've provided a thorough explanation of this, as well as
+conventions, examples, and common closure patterns you may not be used to, in our [closure compiler documentation](./closure-compiler.md).
 
 ### Testing
 
@@ -753,7 +780,7 @@ be an easy and successful way to bootstrap fixtures without having to worry abou
 files or write unwieldy DOM API code.
 
 #### Always clean up the DOM after every test
-This is important. _Before calling `t.end()`, ensure that any elements attached to the DOM have been
+This is important. _Before a test ends, ensure that any elements attached to the DOM have been
 removed_.
 
 #### Verify adapters via testdouble.
--- a/docs/closure-compiler.md
+++ b/docs/closure-compiler.md
@@ -1,0 +1,697 @@
+# Annotating MDC-Web for the Closure Compiler
+
+> TL;DR read the section on our [type system](#mdc-web-type-system) and our [closure compiler conventions](#mdc-web-closure-conventions).
+
+## Who this document is for
+
+This document is for _core contributors to MDC-Web, as well as contributors who wish to author
+new components, or make non-trivial changes to existing components._ It assumes you're familiar with
+our codebase, and have read through most of our [Authoring Components guide](./authoring-components.md).
+
+## Why this is needed
+
+MDC-Web - and Material Design in general - was created by Google. Therefore, it is not only a top
+priority that MDC-Web works seamlessly for our external community, but also that _MDC-Web works
+seamlessly for all Google applications_.
+
+At Google, all Javascript is processed and minified by the
+[Closure Compiler](https://github.com/google/closure-compiler) (which will be referred to as
+**closure**, the **compiler**, or any combination of those terms). Thus, _in order for every Google
+application to deem MDC-Web viable for use within it, the library must be compilable using
+closure's [advanced compilation](https://developers.google.com/closure/compiler/docs/api-tutorial3)
+mechanisms_.
+
+### What about [externs](https://github.com/google/closure-compiler/wiki/FAQ#how-do-i-write-an-externs-file)?
+
+Simply put: They will not cut it. When closure uses externs, it omits compiling those libraries;
+instead it simply makes the compiled code aware they exist. This is unacceptable for many
+applications at Google, whose build infrastructures require that all JS source code be compilable,
+so as to maximize payload optimization and site speed performance.
+
+## Closure Overview
+
+If you've never worked on closure code before, we suggest you start by reading these pages in order:
+
+1. https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler
+1. https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System
+
+You can also check out Google Developers's [Getting Started with the UI](https://developers.google.com/closure/compiler/docs/gettingstarted_ui) tutorial which will
+introduce you to the [compiler appspot service](https://closure-compiler.appspot.com/home). The
+service is extremely useful for testing and debugging closure code in isolation.
+
+### Pro-tips for testing/debugging using the compiler appspot service
+
+If/when using the appspot service, make sure you have `ADVANCED_OPTIMIZATIONS`
+checked under the `optimization` option.
+
+Furthermore, you may want to check the `Pretty Print` option if you're debugging the output code.
+
+Additionally, if you use getter/setter properties within your code, add the following in between
+the `// ==ClosureCompiler==` / `// ==/ClosureCompiler==` comment block:
+
+```js
+// @language_out ECMASCRIPT5
+```
+
+This will tell closure to compile your code to ES5 (the default is ES3 which doesn't support
+accessor properties).
+
+Finally, you can make the permalinks to closure code a bit nicer by using a link shortener, such as
+https://goo.gl.
+
+You can use [this starter template](https://goo.gl/YSQkDi) to help debug your closure code, which
+has all of the above settings pre-configured.
+
+## MDC-Web Type System
+
+The following UML-like diagram shows a conceptual overview of the basic type system for MDC-Web. The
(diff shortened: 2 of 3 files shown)


docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+service is extremely useful for testing and debugging closure code in isolation.
+
+### Pro-tips for testing/debugging using the compiler appspot service
+
+If/when using the appspot service, make sure you have `ADVANCED_OPTIMIZATIONS`
+checked under the `optimization` option.
+
+Furthermore, you may want to check the `Pretty Print` option if you're debugging the output code.
+
+Additionally, if you use getter/setter properties within your code, add the following in between
```
shyndman: It would be good to add a link with language_out and advanced optimizations already configured.

https://goo.gl/c1syGx

docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+  // ...
+};
+
+/** @enum {number} */
+export const numbers = {
+  // ...
+};
+```
+
+Defining these as enums allows the use of their properties without worrying about them being renamed
```
shyndman: I'm not sure what this means. Doesn't Closure usually inline enum values?

docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+
+// GOOD
+/** @dict */
+window.settings = {
+  'windowObject': window,
+  'domReadyEvent': 'onready',
+  'scriptExecutionTimeoutMs': 3000,
+}
+
+// BAD
```
shyndman: It wasn't immediately obvious to me what was wrong with this. Then I realized that the keys weren't quoted, which is described below. Should the description come first? Or maybe the BAD comment should say why it's bad?

docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+#### foundation.js
+
+```js
+import MDCFoundation from '@material/base/foundation';
+import MDCExampleAdapter from './adapter';
+import {cssClasses, strings, numbers} from './constants';
+
+/** @final @extends {MDCFoundation<!MDCExampleAdapter>} */
+export default class MDCExampleFoundation extends MDCFoundation {
+  /** @return enum{cssClasses} */
```
shyndman: I've never seen kind of type annotation before. Good to know. 👍

docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+  }
+
+  /** @return enum{numbers} */
+  static get numbers() {
+    return numbers;
+  }
+
+  /** @return {!MDCExampleAdapter} */
+  static get defaultAdapter() {
+    return /** @type {!MDCExampleAdapter} */ ({
```
shyndman: No need for a cast. It can be inferred. Example: https://goo.gl/lklmPN

docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+seamlessly for all Google applications_. This is an absolute requirement for us.
+
+At Google, all Javascript is processed and minified by the
+[Closure Compiler](https://github.com/google/closure-compiler) (which will be referred to as
+**closure**, the **compiler**, or any combination of those terms). Thus, _in order for every Google
+application to deem MDC-Web viable for use within it, the library must be compilable using
+closure's [advanced compilation](https://developers.google.com/closure/compiler/docs/api-tutorial3)
+mechanisms_.
+
+### What about externs?
```
traviskaufman: Definitely, especially from Googlers. We've been asked about this internally a few times which is why I included it. I can make "externs" a link so that anyone unfamiliar can find out more info.

docs/closure-compiler.md
traviskaufman: Great suggestion

docs/closure-compiler.md
traviskaufman: Nice! I definitely did get an error for this...maybe it was changed upstream.

docs/closure-compiler.md
```diff
@@ -0,0 +1,699 @@
...
+applications at Google, whose build infrastructures require that all JS source code be compilable,
+so as to maximize payload optimization and site speed performance.
+
+## Closure Overview
+
+If you've never worked on closure code before, we suggest you start by reading these pages in order:
+
+1. https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler
+1. https://github.com/google/closure-compiler/wiki/Annotating-Types
+1. https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System
```
traviskaufman: Couldn't agree with you more.

traviskaufman:
@shyndman made changes based on your comments PTAL

docs/authoring-components.md
```diff
@@ -677,31 +679,45 @@ the aforementioned rule, we would also like to provide lint rules to enforce the
   could potentially be localized, etc.) must be referenced by a `strings` static getter.
 - All semantic numbers leveraged by the adapter (timeout lengths, transition durations, etc.) must
```
lynnjepsen: Why do numbers need to be leveraged by the adapter? I understand why you need the class names for CSS renaming, but why do we need DEFAULT_THROTTLE_DELAY_MS exposed?

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+   */
+  isActive() {
+    return this.active_;
+  }
+
+  /**
+   * @param {boolean} active
+   */
+  setActive(active) {
+    const {FADE_IN, FADE_OUT, IMPORTANT_MSG_FLASH} = cssClasses;
```
lynnjepsen: This seems brittle. Do we protect against constants.js and foundation.js from getting out of sync by just testing a lot?

docs/authoring-components.md
traviskaufman: I think that's a typo. I most likely meant _foundation_ there. Looks like it's unrelated to this PR but I'll definitely make that change. 

To answer: the reason we expose numbers is 1) to minimize the amount of magic numbers floating around our codebase 2) A lot of those numbers have something to do with what's happening in CSS, e.g. the amount of time a transition should take. This helps centralize that information and expose it in one place.

docs/closure-compiler.md
traviskaufman: I'm not sure they'll go out of sync, since `MDCExampleFoundation.{cssClasses,strings,numbers}` simply proxy to what's defined within `constants.js`. Where exactly do you see them going out of sync?

lynnjepsen: It is one of those "if you update this then update that". Aka if you update constants.js cssClasses, then you should update the implementation of foundation.js setActive method. It could be unintuitive to the developer to remember to update both places? Its a nit though

traviskaufman: (as discussed offline) testing, as well as (eventually) closure should prevent that from slipping through the cracks.

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+- All references to object keys _must be done using bracket notation_
+
+#### `Object<string, T> vs. @dict`
+
+- Use `Object<string, T>` for objects where _the type for every must be the same_.
+- Use `@dict` for objects where _the type for every value can vary_.
+
+### An example component skeleton
+
+The following shows a set of skeleton files for an example component:
```
bwobrien: Have we considered checking in these example files as an easier jumping off point?

traviskaufman: Hmm I haven't. Would separating these into different files make it easier such that users could simply copy/paste the file over? Or copy the files into a new directory like a template? I'd argue that having them in one markdown file actually might be better for documentation purposes since it keeps all the code local to one specific file. Therefore rather than have to jump around to multiple different files users could use this one page as a reference. WDYT?

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+  is no such thing as a "base adapter", they are simply considered to be plain objects.
+- **Foundations are parameterized by their adapters**. For example, an `MDCRippleFoundation` would be
+  parameterized by an `MDCRippleAdapter`. Thus, when declaring the `MDCRippleFoundation` class, the
+  proper JSDoc to specify this would be included: `@extends MDCFoundation<!MDCRippleAdapter>`.
+- **Components are parameterized by their foundations**. For example, an `MDCRipple` would be
+  parameterized by `MDCRippleFoundation`, which - as shown above - is itself parameterized by
+  `MDCRippleAdapter`. Thus, when declaring the `MDCRipple` class, the proper JSDoc to specify this
+  would be included: `@extends MDCComponent<!MDCRippleFoundation>`.
+
+## MDC-Web Closure Conventions
```
amsheehan: This may be the single most useful section for ~~external~~ contributors. Consider making this more prominent somehow, or putting the link in the TL;DR

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+example of this might be a map of event types to their respective listeners, as shown above.
+
+When closure compiles javascript using advanced optimizations, it [obfuscates property values](https://developers.google.com/closure/compiler/docs/api-tutorial3#better) in
+objects, classes, etc. While this leads to smaller code, it also leads to issues when those
+properties are used externally. In order to prevent this behavior, object keys need to be quoted,
+and those keys need to be referenced using bracket notation.
+
+### Member declarations via sentinel property reference expressions.
+
+**Example:**
```
amsheehan: This should maybe come after line 689

docs/closure-compiler.md
amsheehan: I don't mind having it here. The section that references the skeleton links to this part of the page, which I think is sufficient.

docs/closure-compiler.md
traviskaufman: It's actually already in there, but I tried to make it more prominent.

docs/closure-compiler.md
traviskaufman: I tried to keep all the examples at the top for consistency, but I see your point here. Moving it.

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+```js
+/** @final @extends {MDCComponent<!MyComponentFoundation>} */
+class MyComponent extends MDCComponent {
+  /**
+   * @param {...?} args
+   */
+  constructor(...args) {
+    super(...args);
+    /** @private {!Element} */
+    this.innerEl_;
```
yeelan0319: nit: add comment to make it more explicit this is the sentinel expression statement we are adding.

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+example of this might be a map of event types to their respective listeners, as shown above.
+
+When closure compiles javascript using advanced optimizations, it [obfuscates property values](https://developers.google.com/closure/compiler/docs/api-tutorial3#better) in
+objects, classes, etc. While this leads to smaller code, it also leads to issues when those
+properties are used externally. In order to prevent this behavior, object keys need to be quoted,
+and those keys need to be referenced using bracket notation.
+
+### Member declarations via sentinel property reference expressions.
+
+Closure enforces that properties cannot be added to object instances unless those properties are
```
yeelan0319: IMHO, a better way for describing it should be:
1. Closure requires property to be specified in their constructor
2. However, mdc allows property to be added via `initialize`
3. In that case, we are adding a sentinel statement (re-define) the property in components constructor.

The reason why we cannot do 
```
constructor(...args) {
    /** @private {!Element} */
    this.innerEl_ = DEFAULT_VALUE;
    super(...args);
  }
```
is because closure requires `super()` be the first expression within a method.

Basically, talk about the solution first and then discuss why alternative is not feasible, but not mixing those things together.

docs/closure-compiler.md
```diff
@@ -0,0 +1,697 @@
...
+specified in their constructors. This is problematic in cases where a base constructor calls a
+"setup" function in which properties are added to the instance. This is a pattern we use in our
+codebase via `MDCComponent#initialize()`, so that the component has an opportunity to perform any
+instantiation logic without losing all of the base constructor logic of assigning a root element,
+instantiating an adapter, etc. Because `initialize()` is called as part of the `super()` call within
```
yeelan0319: Also this might be a obvious question but I didn't quite get it.

What does closure mean property cannot be added outside constructor? It seems to me that `initialize` is called as part of `constructor`, so the property is added to `MDCComponent` at that time. So I wonder if that has to do with how closure identify Javascript is adding property? Does it mean literally that the property need to be defined within `constructor`?

Issue #432, opened 2017-03-23
Menu not covering other controls


tranrate:
## What MDC-Web Version are you using?

0.7.0

## What browser(s) is this bug affecting?

Chrome Version 57.0.2987.110 (32-bit)
Firefox 52.0.1 (64-bit)
## What OS are you using?

Windows 10
## What are the steps to reproduce the bug?

Place a set of MDC Radio buttons immediately below a Menu, so that the Menu will cover them when opened. Open the menu.
Code to cut and paste into Codepen to do this is after the screen shot below.
## What is the expected behavior?

The Radio buttons should be covered by the Menu
## What is the actual behavior?

The "Radio" buttons show through the Menu.

screen shot 2017-03-23 at 20:37
![mdc simple menu](https://cloud.githubusercontent.com/assets/23192430/24269218/61f70ad2-1008-11e7-8ed0-0cf98e7ad241.jpg)

### HTML used

    <main>
      <h1>MDC Simple Menu</h1>
      <p>
        Button position:
        <label>
          <input type="radio" name="position" value="top left" checked>
          Top left
        </label>
        <label>
          <input type="radio" name="position" value="top right">
          Top right
        </label>
        <label>
          <input type="radio" name="position" value="bottom left">
          Bottom left
        </label>
        <label>
          <input type="radio" name="position" value="bottom right">
          Bottom right
        </label>
      </p>
      <p>
        <label>
          <input type="checkbox" name="dark">
          Dark mode
        </label>
      </p>
      <div>
        <span>Last Selected item: <em id="last-selected">&lt;none selected&gt;</em></span>
      </div>
      <div class="demo-content">
        <div class="mdc-menu-anchor">
          <button class="toggle">Toggle</button>

          <div class="mdc-simple-menu" style="position: absolute;" tabindex="-1">
            <ul class="mdc-simple-menu__items mdc-list" role="menu" aria-hidden="true">
              <li class="mdc-list-item" role="menuitem" tabindex="0">Back</li>
              <li class="mdc-list-item" role="menuitem" tabindex="0">Forward</li>
              <li class="mdc-list-item" role="menuitem" tabindex="0">Reload</li>
              <li class="mdc-list-divider" role="separator"></li>

              <li class="mdc-list-item" role="menuitem" tabindex="0">Save As...</li>
              <li class="mdc-list-item" role="menuitem" tabindex="0">Print...</li>
              <li class="mdc-list-item" role="menuitem" tabindex="0">Cast...</li>
[...]

PR #460, opened 2017-03-31
fix(select): Fixing bug with select menu z-index

lukebergen:
Some elements were showing up over the select menu's pop out. e.g.
radio-buttons. To demonstrate the problem, I copied demos/radio.html:72-94 into
demos/select.html just after line 74. This puts a radio button right
where the select menu pops up. When running the demo without the fix,
you can see the radio button over the select menu. With the z-index fix,
the menu appropriately pops up over the radio-button.

Fixes [Issue 432](https://github.com/material-components/material-components-web/issues/432)


lukebergen:
I signed it!

PR #462, opened 2017-04-01
docs: Update sass whole package includes

Roang-zero1:
The whole package includes currently recommended in some docs require additional Sass include paths.
This change fixes this and aligns the includes with the other includes present in the package.


Roang-zero1:
I signed it!

PR #465, opened 2017-04-03
fix(dialog): Ensure isOpen() returns false when dialog is closed

traviskaufman:


packages/mdc-dialog/foundation.js
```diff
@@ -82,6 +82,7 @@ export default class MDCDialogFoundation extends MDCFoundation {
   }
 
   open() {
+    this.isOpen_ = true;
```
amsheehan: Is this move just for consistency with `close()`?

traviskaufman: Pretty much; I'm a fan of setting internal state as the first thing in methods. This way you can use the updated internal state in the rest of the method body.

amsheehan: That's a good idea. My thinking was that the dialog wouldn't open until all of the logic had been run successfully. I think this is a good approach though.

PR #466, opened 2017-04-03
docs(package): Improve language up front in terms of the goals

cristobalchao:
Improve language up front in terms of the goals for MDC-Web while onboarding

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Material Components for the web is the successor to [Material Design Lite](https
 
 MDC-Web strives to seamlessly incorporate into a wider range of usage contexts, from simple static websites to complex, JavaScript-heavy applications to hybrid client/server rendering systems. In short, whether you're already heavily invested in another framework or not, it should be easy to incorporate Material Components into your site in a lightweight, idiomatic fashion.
 
+> **Note:** MDC-Web is intended to be used for other developers to execute Material Design properly. MDC-Web delegates higher level capabilities to other JS frameworks and libraries. An example is that MDC-Web neither use web components nor renders the DOM of its components. By doing low-level operations, MDC-Web its able to accomplish its goals.
+
 **[Demos](https://material-components-web.appspot.com/)** (updated with every release)
 
 ## Quick start


README.md
```diff
@@ -11,6 +11,8 @@ Material Components for the web is the successor to [Material Design Lite](https
 
 MDC-Web strives to seamlessly incorporate into a wider range of usage contexts, from simple static websites to complex, JavaScript-heavy applications to hybrid client/server rendering systems. In short, whether you're already heavily invested in another framework or not, it should be easy to incorporate Material Components into your site in a lightweight, idiomatic fashion.
 
+> **Note:** MDC-Web is intended to be used for other developers to execute Material Design properly. MDC-Web delegates higher level capabilities to other JS frameworks and libraries. An example is that MDC-Web does not offer the latest of Web components capabilities, e.g. custom elements, Shadow DOM, HTML imports/templates. We expect other JS frameworks and libraries to deal with these configurations.
```
traviskaufman: I think this is a good start but I'd like to rephrase this section a bit. 

I think we could frame this better by saying that MDC-Web explicitly doesn't do things like use web components, render its own DOM, etc. so that it can be as interoperable as possible with other frameworks and libraries. It's not that we "expect other JS frameworks and libraries to deal with these configurations", but more like "we try very hard not to get in the way of what other JS frameworks and libraries do".

cristobalchao:
Changes made PTAL

README.md
```diff
@@ -11,6 +11,8 @@ Material Components for the web is the successor to [Material Design Lite](https
 
 MDC-Web strives to seamlessly incorporate into a wider range of usage contexts, from simple static websites to complex, JavaScript-heavy applications to hybrid client/server rendering systems. In short, whether you're already heavily invested in another framework or not, it should be easy to incorporate Material Components into your site in a lightweight, idiomatic fashion.
 
+> **Note:** MDC-Web is intended to be used for other developers to execute Material Design properly. MDC-Web delegates higher level capabilities to other JS frameworks and libraries. An example is that MDC-Web neither use web components nor renders the DOM of its components. By doing low-level operations, MDC-Web its able to accomplish its goals.
```
traviskaufman: More feedback:
- I don't think this has to be quoted, nor does it need `**Note:**`. It can just be a normal paragraph
- "higher-level capabilities" is a bit ambiguous; how about instead we say something like "MDC-Web does not handle things like DOM rendering, and automatic component instantiation". I don't want to use the word "delegate" because I don't want it to sound like people who want to use the library are expected to do a whole bunch of work just to get it working properly. 
- We can remove the last sentence. We don't do "low-level" operations, we just avoid doing certain things that we feel would step on the toes of non-UI-library code.

traviskaufman:
🏓 ping!

README.md
cristobalchao: I still have concerns about the message that we are trying to communicate. Before I iterate with your feedback, I would like to make sure we are on the same page. Initially, I thought that the main goal for this PR is to clarify that material components for the web does not offer "Web components" but a "bootstrap-y component" solution. The UX of these solutions are drastically different from each other, e.g. one offers encapsulation and the other one doesn't. Another analogy is the difference between a black box (Web components) vs a white box (bootstrap). Implicitly this adds new different audiences to the table:

1. Users that want to use these components by just thinking of black boxes. This is a higher level operation from a user's perspective, just plug and play. More suited to prototypers, first time learners, etc.

2. Users that want to have total control of components by working with white boxes (Low-level operations). This is more suited to developers that want to customize and/or build wrappers on their own frameworks.

Thoughts?

amsheehan:
🏓  pong!

cristobalchao:
Left comments inline a week ago. :-)

README.md
traviskaufman: We want to encourage _both_:

- People who just want to easily give a website a Material Design look and feel can drop MDC-Web onto their page and use our convenience mechanisms to instantiate and manage the components

- People who need more control over components can use them a-la-carte and integrate them into their frameworks, etc.

traviskaufman:
I think we should probably get more concensus with @omerziv about what it is we're exactly trying to communicate here. So, I'm going to close this PR for now and we can re-open it when we have a better idea of what that is 😄

PR #467, opened 2017-04-03
fix(grid-list): Updates the grid-list's foundation to NOT center tile…

### LINKED ISSUE #454: Grid-list does not handle empty list
v0.7.0
Chrome
Windows10

1. run the demo server
2. go to localhost:8080/grid-list.html
3. no errors
4. remove all li from one of the ul
5. error: Uncaught TypeError: Cannot read property 'offsetWidth' of undefined
    at Object.getOffsetWidthForTileAtIndex (index.js:32)

It should handle empty ul's

Throws error when list is empty

This is the failing line: `return this.root_.querySelectorAll(MDCGridListFoundation.strings.TILE_SELECTOR)[index].offsetWidth;
      },`

Here's an working version of getOffsetWitdthForTileAtIndex:

`getOffsetWidthForTileAtIndex: (index) => {
        const tile = this.root_.querySelectorAll(MDCGridListFoundation.strings.TILE_SELECTOR)[index];
        return tile ? tile.offsetWidth : 0;
      },`

lynnjepsen:
…s when there are no tiles

Resolves #454, Updates the grid-list's foundation to NOT center tiles when there are no tiles. Also updates grid-list's index.js to check if a tile exists before calling .offsetWidth on it.


packages/mdc-grid-list/foundation.js
```diff
@@ -25,6 +25,7 @@ export default class MDCGridListFoundation extends MDCFoundation {
   static get defaultAdapter() {
     return {
       getOffsetWidth: () => /* number */ 0,
+      isEmpty: () => false,
```
traviskaufman: I feel like this may be a _bit_ too abstract for an adapter method. Instead, how do you feel about a `getNumberOfTiles()` method?

```js
getNumberOfTiles: () => /* number */ 0,
```

To me this narrows the scope of what the adapter needs to do down to just returning the number of elements the list contains, which needs to be done fully within the host environment. Whereas checking that that number is equal to `0` is something we could do in the foundation.

packages/mdc-grid-list/index.js
```diff
@@ -28,8 +28,12 @@ export class MDCGridList extends MDCComponent {
   getDefaultFoundation() {
     return new MDCGridListFoundation({
       getOffsetWidth: () => this.root_.offsetWidth,
+      isEmpty: () => {
+        return this.root_.querySelectorAll(MDCGridListFoundation.strings.TILE_SELECTOR).length == 0;
+      },
       getOffsetWidthForTileAtIndex: (index) => {
-        return this.root_.querySelectorAll(MDCGridListFoundation.strings.TILE_SELECTOR)[index].offsetWidth;
+        const tile = this.root_.querySelectorAll(MDCGridListFoundation.strings.TILE_SELECTOR)[index];
+        return tile ? tile.offsetWidth : 0;
```
traviskaufman: I don't think you need this ternary statement here, since `getOffsetWidthForTileAtIndex` is only called when the number of tiles is greater than 0 (thanks to the short-circuit added at the top of `alignCenter`).

packages/mdc-grid-list/README.md
```diff
@@ -258,6 +258,7 @@ functions, with correct signatures:
 | Method Signature | Description |
 | --- | --- |
 | `getOffsetWidth() => number` | Get root element `mdc-grid-list` offsetWidth. |
+| `getNumberOfTiles() => number` | Get the number of `mdc-grid-tile`. |
```
traviskaufman: supernit: Just to make sure we're being crystal clear about what we need from implementors, could we change the description to "Get the number of `mdc-grid-tile` elements contained within the grid list"?

PR #474, opened 2017-04-04
docs: Add eng outline doc info to CONTRIBUTING, Issue Template

traviskaufman:
The MDC-Web Core Team has recently introduced the concept of
"Engineering Outlines" into our process of developing component. These
function as more lightweight versions of "design documents", where we
can get a good sense of how an engineer plans on approaching a component
from an implementation perspective. We've found this practice extremely
useful for us, and have decided to require it as part of developing new
components. This PR adds information on eng outlines being a requirement, as well as where to find the template for them, in our CONTRIBUTING
and issue template docs.


amsheehan:
The only thing I'd add is _why_ we're enforcing this. Something along the lines of "We have found that enforcing a design doc has allowed us to speed up development, offer more informed feedback on component implementations, and has resulted in components that take into account all of the concepts MDC Web components should account for (RTL, aria roles, etc...)"

traviskaufman:
@amsheehan changes made!

Issue #435, opened 2017-03-23
Select Menu not anchored to the control


tranrate:
## What MDC-Web Version are you using?

0.7.0
## What browser(s) is this bug affecting?

Chrome Version 57.0.2987.110 (32-bit)
Firefox 52.0.1 (64-bit)
## What OS are you using?

Windows 10
## What are the steps to reproduce the bug?

On your Demo site, go to the Select demo. Reduce the size of your browser window so that you've got a scroll bar for the page. Open the Fully Featured Component, then use the scroll bar to scroll the page. 
## What is the expected behavior?

The Select Menu should scroll with the rest of the page so that it stays above its Input Control.
## What is the actual behavior?

The Menu doesn't move.
## Any other information you believe would be useful?

This only seems to happen to the Menu which is part of the Select. When the Simple Menu Control is used on its own, it works fine.

tranrate:
mdc-select.scss line 87 change fixed to absolute. Seems to solve but needs fully testing.

amsheehan:
Looks like you may be right on the proposed fix. Would you mind making a pull request with that change?

lukebergen:
Running into a different but semi-related bug. Warning about checking that position from fixed to absolute: I think it'll mess up the position of the menu if the page has scrolled. e.g.
1) On the demo page http://material-components-web.appspot.com/select.html
2) Edit the style on the `.mdc-select__menu` to set the position to `absolute`
3) Edit the style on the H1 at the top of the page and set `margin-bottom: 1200px;` or so to force the select menu down below the fold.
4) Now scroll down to the menu and open it.

Note that the menu opens well above the element that it's supposed to be anchored to.

traviskaufman:
err...are we sure that we want the intended behavior to be to have modal menus scroll with the page?

The JS menu was modeled after the browser's native contextual menu which:

- Has the highest possible z-index and true modality within the viewport regardless of it's stacking context
- Overflows any contained element to expand to the height of the viewport in order to show all possible options.

For this reason, we made the menu fixed position. If it were absolute, and the user scrolled past the menu, how would she then know that clicking on part of a page is actually canceling that menu.

Perhaps I'm missing something here? @amsheehan ?

traviskaufman:
[...]

PR #478, opened 2017-04-05
Fix for #435 Select Menu not anchored to Control 

tranrate:
Change .scss from fixed to absolute. Modify positioning code in foundation to correctly position menu when window scrolled (as mentioned in comment by Luke Bergen). Requires an extra Adapter Interface for pageYOffset. PageYOffset is not supported on IE before v9 according to MDN but we only support IE11 anyway.

Tested by modifying Demo to move control nearer bottom of page. Do you need any further testing by me?

Martin


tranrate:
I signed it.

Martin

tranrate:
Obviously need to sort the tests

PR #482, opened 2017-04-05
Select Menu not anchored to the control #435

tranrate:
mdc-select.scss line 87 change fixed to absolute. Positioning of Menu in Foundation corrected to handle window scrolling and conditions when Menu would extend beyond Window (see Luke Bergen's comment). New Adapter interface getPageYOffset required to handle scrolling.

This probably requires some additional tests setting up to cover the new cases?

There is one case that this does not cover. That is if the Select is inside a scrollable element e.g. a Scrollable Dialog. The problem is how to detect this without doing a traversal from the control to the window and checking each parent element, which is obviously not very efficient. Perhaps we need an extra option in the api to allow the programmer to indicate when this is necessary?

Martin


tranrate:
I have altered this slightly to put window.innerHeight and pageYOffset into a single adapter interface called computeBoundingContainer. This can then be replaced in a similar fashion to the BoundingContainer for Ripples. This seems the best way to resolve the problem of scrollable elements other than the main window and any other similar problems.

Martin

traviskaufman:
Thanks for this work @tranrate but it looks like we still need to discuss this before accepting a PR. Adding a `do-not-merge` label for now. I apologize for the miscommunication here.

amsheehan:
Closing this PR. Please reference https://github.com/material-components/material-components-web/issues/435 for details.

PR #484, opened 2017-04-06
fix(demos): Fix switch demo header typo

0xdec:


traviskaufman:
Looks like this is a dupe of what's in #485 ?

0xdec:
Oh yes, sorry. I meant to separate this into its own PR as it is strictly a typo fix, but it seems I accidentally included it in the other branch as well. I'll just fix up #485 and close this one if you think that's a good idea.

traviskaufman:
Yeah, let's do that instead. Closing.

PR #485, opened 2017-04-06
feat(demos): Update demos styling

0xdec:
Didn't know if I should split this into multiple PRs or not.


demos/icon-toggle.html
```diff
@@ -24,11 +24,6 @@
         background: #303030;
       }
 
-      .mdc-theme--dark:not(.demo-color-combo) {
```
traviskaufman: If you're going to remove this then I'd make sure the non-disabled dark theme example still has enough padding such that the entire ripple can be contained within the dark background.

demos/simple-menu.html
```diff
@@ -35,7 +35,11 @@
         position: relative;
         flex: 1;
         margin-top: 16px;
-        background-color: antiquewhite;
```
traviskaufman: Let's leave this as is. The gray being used here has a bit too low a contrast IMHO.

0xdec: Okay will do. Are you fine with it changing to "#333" when dark mode is on?

demos/icon-toggle.html
0xdec: Will do. Would adding left padding to all icons so that they are still aligned be an acceptable solution?

demos/icon-toggle.html
```diff
@@ -24,9 +24,13 @@
...
-        padding: 1rem;
-        padding-top: 0;
+      section.mdc-theme--dark:not(.demo-color-combo) {
+        padding-bottom: 1rem;
+      }
+
+      .mdc-icon-toggle {
```
traviskaufman: I'd maybe just put a `demo-wrapper` class around the icon toggles that give it `1rem` margins rather than mutate the icon toggle itself.

0xdec: Done!

0xdec:
Added a couple more fixes that I noticed while updating this PR

0xdec:
Thank you, happy to help in any way I can!

PR #490, opened 2017-04-07
feat(radio): Adds a value property to the component

shyndman:


shyndman:
Ah, thanks. Taking a look.

shyndman:
Value test added.

PR #491, opened 2017-04-07
fix(dialog): Handle focus trapping correctly

### LINKED ISSUE #409: Dialog put focus on both button in demo example

>0.7.0

> Chrome Version 57.0.2987.98 (64-bit)
> Chrome Canary Version 59.0.3046.0 (Official Build) canary (64-bit)

### What OS are you using?

> macOS 10.12.3 (16D32)

> 1. run the demo server
> 2. go to localhost:8080/dialog.html
> 3. Click `show dialog` button on the left
> 4. Hit tab (more than 1 time) and then close the dialog (either by click one of the button or escape)
> 5. Click `show scrolling dialog` button on the right
>

> Just the `accept` button should trapped focus

> Both `decline` and `accept` trapped focus

A couple of known fact that may help debug:
- It only happens when the page first load, once you interact with `scrolling dialog`, I am not able to reproduce it.
- It only happens when you interact with `default dialog` then `scrolling dialog`, not visa versa.

### LINKED ISSUE #424: Initial dialog focus seems to trigger side-effects on underlying page

0.7.0

Chrome Version 56.0.2924.87 (64-bit)

### What OS are you using?

macOS Sierra Version 10.12.3

1. Click somewhere blank on the page to ensure that focusing will be within the context of the document
1. Press the `tab` key twice so that the "Show Scrolling Dialog" button is focused
1. Press the spacebar to display the dialog
1. Press the `tab` key to focus on the "Decline" button

The decline button should be the only element that gains focus, or changes at all

The "Show Dialog" button also appears to highlight as if it were being focused.

screen shot 2017-03-22 at 13 14 55

When you hit the `esc` key to cancel the dialog, both buttons appear to have gained focus:

screen shot 2017-03-22 at 13 15 46

### LINKED ISSUE #426: Opening Dialog Adds aria-hidden to Entire Page
This is applicable for all operating system, browser, and screen reader combinations.
When a dialog is opened, the other elements on the page are supposed to have aria-hidden. However, in the mdc dialog, the document body has aria-hidden applied to it, which also hides the dialog itself since the dialog is a child of the document body.

traviskaufman:
- Remove all focus trapping logic within the dialog foundation,
  replacing it with adapter methods set up to handle focus trapping.
- Add [focus-trap](https://npmjs.com/focus-trap) as a dependency and use
  it for implementing the adapter methods at the component level
- Add a util method for easily creating properly configured focus-trap
  instances that framework components can use when implementing their
  adapters.
- Remove use of aria-hidden and replace with visibility styles so that
  tab-focusing works correctly when dialog is hidden.
- [techdebt] Remove all unnecessary adapter API methods and dead code
- [techdebt] Make grammatical corrections in README
- [techdebt] Remove unneeded passive event listener logic for dialog handlers
- [demos] Fix ripples incorrectly appearing in buttons within the demos.

Resolves #424
Resolves #409
Resolves #426

NOTE: The additional issues resolved are by-products from switching to
focus-trap.

BREAKING CHANGE: There are a few changes that need to be taken into
account for this commit:

- Dialogs no longer require an `aria-hidden="true"` attribute.
- Dialogs _do_ require a `style="visibility:hidden"` attribute for
  correct first render.
- `trapFocusOnSurface` and `untrapFocusOnSurface` methods must be
  implemented for the adapter
- `hasClass`, `setAttr`, `registerFocusTrappingHandler`,
  `deregisterFocusTrappingHandler`, `numFocusableTargets`,
  `setDialogFocusFirstTarget`, `setInitialFocus`,
  `getFocusableElements`, `saveElementTabState`,
  `restoreElementTabState`, `makeElementUntabbable`, `setBodyAttr`,
  `rmBodyAttr`, `getFocusedTarget`, and `setFocusedTarget` have all been
  removed from the adapter.
- `applyPassive`, `saveElementTabState`, and `restoreElementTabState`
  have all been removed from `mdcDialog.util`.

### DIFF
--- a/packages/mdc-dialog/README.md
+++ b/packages/mdc-dialog/README.md
@@ -18,9 +18,9 @@ Dialogs inform users about a specific task and may contain critical information 
 
 ```html
 <aside id="my-mdc-dialog"
+  style="visibility:hidden"
   class="mdc-dialog"
   role="alertdialog"
-  aria-hidden="true"
   aria-labelledby="my-mdc-dialog-label"
   aria-describedby="my-mdc-dialog-description">
   <div class="mdc-dialog__surface">
@@ -41,20 +41,17 @@ Dialogs inform users about a specific task and may contain critical information 
 </aside>
 ```
 
-> **NOTE**: The `.mdc-dialog__footer__button--accept` element _must_ be the _final focusable element_ within the dialog
-in order for this component to function properly.
-
-In the example above, we've created a dialog box in an `aside` element. Note that you can place content inside 
+In the example above, we've created a dialog box in an `aside` element. Note that you can place content inside
 the dialog. There are two types: dialog & dialogs with scrollable content. These are declared using CSS classes.
 
-Some dialogs will not be tall enough to accomodate everything you would like to display in them. For this there is a 
+Some dialogs will not be tall enough to accomodate everything you would like to display in them. For this there is a
 `mdc-dialog__body--scrollable` modifier to allow scrolling in the dialog.
 
 ```html
   <aside id="mdc-dialog-with-list"
+    style="visibility:hidden"
     class="mdc-dialog"
     role="alertdialog"
-    aria-hidden="true"
     aria-labelledby="mdc-dialog-with-list-label"
     aria-describedby="mdc-dialog-with-list-description">
     <div class="mdc-dialog__surface">
@@ -76,14 +73,14 @@ Some dialogs will not be tall enough to accomodate everything you would like to 
           <li class="mdc-list-item">Luna</li>
           <li class="mdc-list-item">Marimba</li>
           <li class="mdc-list-item">Schwifty</li>
-        </ul> 
+        </ul>
       </section>
       <footer class="mdc-dialog__footer">
         <button type="button" class="mdc-button mdc-dialog__footer__button mdc-dialog__footer__button--cancel">Decline</button>
         <button type="button" class="mdc-button mdc-dialog__footer__button mdc-dialog__footer__button--accept">Accept</button>
       </footer>
     </div>
-    <div class="mdc-dialog__backdrop"></div> 
+    <div class="mdc-dialog__backdrop"></div>
   </aside>
 ```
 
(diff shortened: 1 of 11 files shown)


packages/mdc-dialog/README.md
```diff
@@ -228,45 +213,77 @@ do so. We provide instructions on how to add ripples to buttons within the [mdc-
...
-| `setFocusedTarget(target: EventTarget) => void` | Sets focus on the given target, e.g. by calling `focus()` |
 | `notifyAccept() => {}` | Broadcasts an event denoting that the user has accepted the dialog. |
 | `notifyCancel() => {}` | Broadcasts an event denoting that the user has cancelled the dialog. |
+| `trapFocusOnSurface() => {}` | Sets up the DOM which the dialog is contained in such that focusability is restricted to the elements on the dialog surface (see [Handling Focus Trapping](#handling-focus-trapping) below for more details). |
+| `untrapFocusOnSurface() => {}` | Removes any affects of focus trapping on the dialog surface from the DOM (see [Handling Focus Trapping](#handling-focus-trapping) below for more details). |
+
+#### Handling Focus Trapping
 
+In order for dialogs to be fully accessible, they must conform to the guidelines outlined in
+https://www.w3.org/TR/wai-aria-practices/#dialog_modal. The main implication of these guidelines is
```
amsheehan: "The main implication of these guidelines is that such that the only focusable elements..."

_Should read:_

"The main implication of these guidelines is that the only focusable elements..."

test/unit/mdc-dialog/foundation.test.js
```diff
@@ -83,92 +82,76 @@ test('#open registers all events registered within open()', () => {
...
 
-  foundation.init();
   foundation.open();
-  td.verify(mockAdapter.restoreElementTabState('foo'));
-  td.verify(mockAdapter.restoreElementTabState('bar'));
+  td.verify(mockAdapter.addBodyClass(cssClasses.SCROLL_LOCK));
 });
 
-test('#close makes elements untabbable', () => {
+test('#close removes the scroll lock class from dialog background', () => {
```
amsheehan: This is on me for coming up with the naming convention for the component, but potentially we should rephrase "background" to "content underneath dialog" for the test description. Aria calls things behind the dialog "background" which is why I called it this to begin with, but I feel this could too easily be confused with the concept of a dialog backdrop, which we also have.

traviskaufman: np! Change `background` to `content underneath the dialog`

traviskaufman: Actually just changed it to `the body` to match the spec description above

PR #497, opened 2017-04-10
fix(checkbox): Prevent checkboxes and radios from changing size with multiline label

### LINKED ISSUE #495: Checkboxes and radios render incorrectly when label has multiple lines

0.8.0

Chrome 57.0.2987.133 (64-bit)

### What OS are you using?

Windows 10 

1. Navigate to https://material-components-web.appspot.com/checkbox.html
2. Add some text to label, make the label be splitted into multiple lines.

Checkbox height and width should be equal.

As the rows increase, the checkbox width becomes narrower.

Radios have the same issue.

mikeyzm:
Resolves #495


traviskaufman:
@amsheehan PTAL when you get the chance

Issue #23, opened 2016-12-05
Investigate how to be consistent across components with responsive breakpoints


traviskaufman:
Spec reference: https://www.google.com/design/spec/layout/responsive-ui.html#responsive-ui-breakpoints.

> Copied from https://github.com/google/material-design-lite/issues/4496

20zinnm:
Any update on this? Very important for material-components-web to be a viable library.

Garbee:
What's the core team up to in the [contributing guide](https://github.com/material-components/material-components-web/blob/master/CONTRIBUTING.md#whats-the-core-team-up-to) explains how to see what is being worked on in more detail. In this case, the [pivotal issue](https://www.pivotaltracker.com/story/show/128619139) is assigned to the early-January 2018 sprint. So it isn't even being looked at until around the time this sprint will start unless it gets moved up.

If someone from the community wants to jump in and work on this sooner then they are more then welcome to. All you need to do is follow the contributing guide on [building components](https://github.com/material-components/material-components-web/blob/master/CONTRIBUTING.md#building-components).

Thank you.

berets76:
What do you exactly mean with "SASS/style helpers for responsive breakpoints" ? 
Why a global library instead a suggested `mdc-layout-grid__cell--phone-hide` or like the bootstrap `hidden-xs` ?

Garbee:
Direct classes to hide things aren't exactly the style for MCW. Perhaps the grid will have them, but there are still other places where these helpers are useful. Which is why we want a component to provide these internally via Sass to be easily re-used everywhere in a MCW project.

Providing `hidden-xs` type classes "like Bootstrap" is not a focus point for the library. MCW is not trying to be a framework to do everything for developers. That's a "utility only" type class which doesn't fit with what is trying to be provided.

The goal isn't to be a framework and do a lot of little functions for developers. It's simply to provide the Material Design Specification's components to the web in a re-usable way. That's it. Anything that is beyond that goes beyond the scope of what the project is trying to achieve.

berets76:
I well know MCW isn't trying to be a framework, if I would a framework I would know where to look, but, since in guidelines there isn't the concept of hide, for example, grid cells on breakpoints, so what kind of SASS/style library are you thinking ?

I don't understand in what other places you could get benefit hiding things: any other component should be, by guidelines, into a (layout) grid.

[...]

PR #504, opened 2017-04-11
fix(dialog): Closing Animations not running #433

### LINKED ISSUE #433: Dialog Closing Animations not running
## What MDC-Web Version are you using?

0.7.0
## What browser(s) is this bug affecting?

Chrome Version 57.0.2987.110 (32-bit)
Firefox 52.0.1 (64-bit)
## What OS are you using?

Windows 10
## What are the steps to reproduce the bug?

On your Demo site open a Dialog and then Close it.
## What is the expected behavior?

The Dialog should run its "Closing" Animation.
## What is the actual behavior?

The Close method "hides" the Dialog without waiting for the animation to run.

## Any other information you believe would be useful?

The foundation needs to wait for a TransitionEnd event before hiding the the Dialog and closing everything down. At the moment the adapter doesn't allow for a TransitionEnd event.

tranrate:
-  Refactor scss to include Animating class.
-  Modify foundation to use Animating class and transitionEnd event for opening and closing Dialog. Alter open close and destroy methods appropriately.
-  Move style="visibility:hidden" attribute into scss so that it is not required in html.

Closes #433

BREAKING CHANGE: 
-  Dialogs do not require a style="visibility:hidden" attribute in html.
-  registerTransitionEndHandler, deregisterTransitionEndHandler, and isDialog methods must be implemented by the adapter


packages/mdc-dialog/foundation.js
```diff
@@ -118,6 +128,18 @@ export default class MDCDialogFoundation extends MDCFoundation {
     }
   }
 
+  handleTransitionEnd_(ev) {
```
amsheehan: Ultra-nit: We usually call this `evt`

packages/mdc-dialog/mdc-dialog.scss
```diff
@@ -30,10 +30,29 @@ $mdc-dialog-dark-theme-bg-color: #303030 !default;
...
+  top: 0;
+  left: 0;
   align-items: center;
   justify-content: center;
-  opacity: 0;
-  z-index: -1;
+  width: 100%;
+  height: 100%;
+  visibility: hidden;
+  z-index: 2;
```
amsheehan: `z-index` is set on line 54 as well. I believe this one is extraneous.

packages/mdc-dialog/mdc-dialog.scss
```diff
@@ -131,7 +133,6 @@ $mdc-dialog-dark-theme-bg-color: #303030 !default;
     padding: 8px;
   }
 
-  // TODO: Replace with breakpoint variable
```
amsheehan: We would like to keep this comment in for now. `//` style comments are stripped out in the compiled CSS, so this only really exists for us in dev. These will all be removed once #23 lands.

tranrate:
Hi @amsheehan 

1.  Demo updated - conflicts with Revamp Demo/Catalog #594 resolved.
2.  Foundation.js line 131 - ev changed to evt.
3.  Mdc-dialog.scss line 134 - Comment added back in.
4.  Mdc-dialog.scss line 40 - z-index left as it is, the second one is for the opaque backdrop which has to be behind the dialog box but in front of the main screen. Used a relative index for the backdrop so that only the main one needs changing if anyone needs to adjust the z-index in the future.

Martin

traviskaufman:
@amsheehan could we get another review here?

Also note that in the commit message body we need to change `BREAKING CHANGES:` to `BREAKING CHANGE:`

tranrate:
@amsheehan 

Commit message changed.

Martin

PR #510, opened 2017-04-12
fix(infrastructure): Print the entire invalid git commit message.

shyndman:
Previously only the invalidate subject line would print, so if you'd written a big meaty comment that failed to validate you'd have to rewrite it.


shyndman:
This happened to me. 😠

PR #511, opened 2017-04-13
docs: Preparing component markdown for documentation site.

shyndman:
* Front matter introduced. We'll still need `excerpts` fields across the board for SEO.
* Cleaned up headlines and component names in intro sections. The convention is
  now to refer the components as "Upper Case Component". Prefixing all components
  with MDC in headlines seemed a little strange in the doc site context, although I'm happy to discuss if you think otherwise.
* Adds icon-lists where applicable. We now link out to spec in intro paragraphs as well as the link lists, but we'll iron this out before release.
* URLs to spec have been updated across the board to point to material.io.

### DIFF
--- a/packages/mdc-dialog/README.md
+++ b/packages/mdc-dialog/README.md
@@ -20,6 +20,9 @@ a dialog should be.
   <li class="icon-list-item icon-list-item--spec">
     <a href="https://material.io/guidelines/components/dialogs.html">Dialogs</a>
   </li>
+  <li class="icon-list-item icon-list-item--link">
+    <a href="https://material-components-web.appspot.com/dialog.html">Demo</a>
+  </li>
 </ul>
 
 ## Installation
--- a/packages/mdc-dialog/README.md
+++ b/packages/mdc-dialog/README.md
@@ -1,10 +1,25 @@
-# MDC Dialog
+<!--docs:
+title: "Dialogs"
+layout: detail
+section: components
+iconId: dialog
+-->
+
+# Dialogs
 
 The MDC Dialog component is a spec-aligned dialog component adhering to the
 [Material Design dialog pattern](https://material.io/guidelines/components/dialogs.html).
 It implements a modal dialog window. You may notice that full screen components outlined in the dialog spec
 do not appear in MDC Dialog. This is because they have been deemed to be outside of the scope of what
 a dialog should be.
+
+## Design & API Documentation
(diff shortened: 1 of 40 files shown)


packages/mdc-dialog/README.md
```diff
@@ -1,11 +1,26 @@
...
 [Material Design dialog pattern](https://material.io/guidelines/components/dialogs.html).
 It implements a modal dialog window. You may notice that full screen components outlined in the dialog spec
 do not appear in MDC Dialog. This is because they have been deemed to be outside of the scope of what
 a dialog should be.
 
+## Design & API Documentation
+
+<ul class="icon-list">
+  <li class="icon-list-item icon-list-item--spec">
+    <a href="https://material.io/guidelines/components/dialogs.html">Checkbox</a>
```
traviskaufman: Text here should say "Dialog"

packages/mdc-select/README.md
```diff
@@ -1,9 +1,24 @@
...
+
+MDC Select provides Material Design single-option and multi-option select menus. It functions analogously to the
 browser's native `<select>` element, and includes a gracefully degraded version that can be used
 in conjunction with the browser's native element. Both are fully accessible, and fully RTL-aware.
 
+## Design & API Documentation
+
+<ul class="icon-list">
+  <li class="icon-list-item icon-list-item--spec">
+    <a href="https://material.io/guidelines/components/menus.html">Menus</a>
```
traviskaufman: FYI we took this from the "Text Field Drop-Down" section within textfields, not menus.

shyndman: Honestly, I find both articles so bizarrely organized. It doesn't look like I can link directly to anything that's about select controls. Should I include the link at all?

traviskaufman: yeah...I don't know...you could try linking to both? Maybe https://material.io/guidelines/components/text-fields.html#text-fields-input-types and then say `(see "menus and pickers")`. 😭

shyndman:
Totally reasonable request, but I have a few things I have to tackle before docs. If you don't mind I'm going to keep the PR open and add additional commits. We can split up the review any way you'd like once I have everything together.

traviskaufman:
SGTM!

traviskaufman:
Putting a `do-not-merge` label on here until you're ready for a review

PR #515, opened 2017-04-15
Scrollbar below toolbar

Touficbatache:
Hi,
In MDC-Web when the toolbar is fixed, the scrollbar comes over it. This pull request allows the users to change that and make the scrollbar appear below the toolbar without affecting it.
I'll add some screenshots to show the difference between before and after my changes.
Thanks
Screenshot

### DIFF
--- a/packages/mdc-toolbar/README.md
+++ b/packages/mdc-toolbar/README.md
@@ -96,7 +96,7 @@ of the toolbar (respectively).
 ```
 
 Toolbar sections are laid out using flexbox. Each section will take up an equal
-amount of space within the toolbar.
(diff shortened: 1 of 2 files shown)


traviskaufman:
Thanks for the PR, but we won't be accepting this as it hasn't been vetted by design or discussed with our engineering team. You'll most likely just want to add those as custom classes outside of MDC yourself if this is your intended behavior, as we usually just opt for what the browser does automatically with its Chrome (scrollbars, etc.) rather than force it to behave a certain way. 

In the future, feel free to open an issue outlining what you would like to change about components so we can discuss it there first. Thanks again!

PR #516, opened 2017-04-16
fix(list): Move list-divider style from select

posva:
I found out select scss file was affecting list style so I just moved the relevant CSS to the stylesheet of the list. If that's not how it should be, I'd have at least raised the point of that extra css rules on the select stylesheet 🙂


traviskaufman:
Hey @posva,

we purposefully alter the list styles within mdc-select to accommodate them being placed on `<option>` elements, as the comment above that code hints at. Thus, this actually is the expected behavior. If you'd like to add a some additional context within that comment as to why this is needed (for example, what I just said above 😛 ), we'll gladly accept that PR.

posva:
Oh, that's ok. But I really think that the css selector should be different (eg nested) because right now that CSS affects all dividers wherever they are. For instance the divider in the Temporary Drawer is affected by that: http://material-components-web.appspot.com/drawer/temporary-drawer.html
I'll update the PR for that hoping it's what is needed 😄

posva:
It's absolutely Ok 🙂

PR #517, opened 2017-04-17
fix: Change scss file to use slash for comment.

yeelan0319:
# Background

The scss pre-processor will strip out `//` commented contents but not `/*` ones. So based on this fact, I am cleaning up the existing comments in `.scss` file.

Meanwhile, not quite sure what scope this commit should belong to, feel free to let me know what scope I should use for it.


yeelan0319:
My pleasure improving code quality :)

PR #522, opened 2017-04-17
docs: improve closure guide.

yeelan0319:
I missed the deadline for #457.

So I create this PR to incorporate some of the suggestions I had there.


docs/closure-compiler.md
```diff
@@ -46,21 +46,20 @@ checked under the `optimization` option.
...
+// ==ClosureCompiler==
 // @language_out ECMASCRIPT5
+// ...Other configurations
+// ==/ClosureCompiler==
 ```
 
 This will tell closure to compile your code to ES5 (the default is ES3 which doesn't support
 accessor properties).
 
-Finally, you can make the permalinks to closure code a bit nicer by using a link shortener, such as
```
yeelan0319: I removed this because I remember in the tutorial it said:

> If you want the Closure Compiler service to serve the compressed file, choose a name for the output file using the @output_file_name parameter at the top of the input field. Note that the default for the output file name is default.js but you should change it to a more meaningful name for your project. Also note that the Closure Compiler service hosts the output file on its servers for **an hour**.

Also feel we are selling a lot at the same time but this is relatively less important...

yeelan0319:
How could a `.md` update break TravisCI? 😕

PR #528, opened 2017-04-18
fix(auto-init): Register MDCPersistentDrawer

### LINKED ISSUE #527: (mdc-auto-init) Could not find constructor in registry for MDCPersistentDrawer
MDCPersistentDrawer constructor seems to not be registered in MDCAutoInit: when using it, the error message in the title is displayed.

traviskaufman:
Fixes #527


traviskaufman:
codecov failure is false negative :/

PR #530, opened 2017-04-19
feat(drawer): Emit open/close events on slidable drawers

### LINKED ISSUE #357: Emit events for open/close drawer actions
I noticed that the event pattern is implemented in `MDCComponent`.
Is there a plan for components to emit events?

For example,
`MDCTemporaryDrawer` could emit an open/close event - this would allow quick integrations without implementing the foundation adapters, and wiring up event handlers is certainly easier.

I was implementing mdc bindings for Knockout - and I have to subclass the `MDCTemporaryDrawerFoundation` just to get access to `open`/`close` in a non-hackish way.

anton-kachurin:
Fixes #357


anton-kachurin:
Do I need to confirm? I confirm then

amsheehan:
@anton-kachurin when I update the branch it adds a commit on my behalf. Thanks again!

PR #532, opened 2017-04-19
fix(ripple): Ripple should not activate on disabled label click

### LINKED ISSUE #377: Disabled radio fires ripple when corresponding label is clicked

0.5.0

Chrome v56

### What OS are you using?

OS X v10.12.3

1. Navigate to http://material-components-web.appspot.com/radio.html
1. Under the "Disabled" section, click the label that reads "Disabled Radio 1"

Nothing happens

A ripple is fired on the radio

Most likely culprit is that _mdc-form-field's [handleClick method](https://github.com/material-components/material-components-web/blob/master/packages/mdc-form-field/foundation.js#L51-L54) is not taking whether or not the element is disabled into account_. In order to fix this, handle click needs to know whether or not the input is disabled, and if it is, should not activate the ripple. 

~Concretely, here's what I believe needs to be done:~  EDIT: See @anton-kachurin's [proposed solution below](https://github.com/material-components/material-components-web/issues/377#issuecomment-284131455).

- ~A method `isInputDisabled() => boolean` needs to be added to the mdc-form-field adapter. This method should return `true` if the form-field's input is disabled, false otherwise.~
- ~The mdc-form-field's foundation class should check `isInputDisabled()` within `handleClick`, and not activate the ripple if `isInputDisabled()` returns `true.`~
- ~The component's API should be modified such that a user can declaratively specify _how_ the _vanilla JS component_ checks if it's input is disabled. This could, by default, by checking a `disabled` property on the `input` class, but it should be allowed to be specified. _Before submitting a PR for this issue, please propose an API for this as an issue comment and await sign-off from one of the core contributors_.~

amsheehan:
resolves #377


packages/mdc-checkbox/index.js
```diff
@@ -64,6 +64,7 @@ export class MDCCheckbox extends MDCComponent {
 
   getDefaultFoundation() {
     return new MDCCheckboxFoundation({
+      shouldIgnoreRippleActivation: () => this.disabled,
```
traviskaufman: Hmm....I'm thinking that `isSurfaceDisabled` or `isSurfaceInteractive` might be a better method name here. 

With `shouldIngoreRippleActivation`, that kind of leaves the heuristics of when a ripple should not be triggered up to the _implementor_. I'm unsure if this is what we should be going for. Instead, we should encode the heuristics of when the ripple should be activated into the foundation itself (e.g. when `isSurfaceActive` is false and we encounter a keyboard event, we don't enable the ripple).

packages/mdc-radio/README.md
```diff
@@ -137,6 +137,8 @@ Since MDC Radio is primarily driven by its native control, the adapter API is ex
 | `addClass(className: string) => void` | Adds a class to the root element. |
 | `removeClass(className: string) => void` | Removes a class from the root element. |
 
+> **NOTE**: There are some cases where we do not want ripples to activate. Most of this logic is handled by native control attributes such as `disabled` on input fields. One caveat is that ripples will activate on disabled elements if their corresponding `label` element is clicked. For this, we override the ripple's `shouldIgnoreRippleActivation()` adapter method in `mdc-radio`'s `getDefaultFoundation()`, and check if the control is disabled before activating the ripple.
```
traviskaufman: What's the reason for including this note here? Seems out of context with the rest of the README. If this is advice for how framework implementors can deal with configuring the ripple properly, we may want to give more background on that and also mention this within the `mdc-checkbox` README.

amsheehan: This actually is no longer applicable. The latest code keeps all of this logic in `mdc-ripple`, whose README documents the behavior.

PR #540, opened 2017-04-21
fix(toolbar): Prevent cut-offs of the toolbar section content

### LINKED ISSUE #524: Ripples get cut-off when they are nested in mdc-toolbar__section

> Master

> All browser

> Here is an [example](https://material.uplabs.com/posts/google-material-design-components-demo) the ripples get cut-off

> The ripples should complete

> The ripples get cut off to the shape of mdc-toolbar__section 

We added `overflow:hidden` for #508 to accommodate very long section. It seems removing that from `mdc-toolbar__section` is necessary but I haven't yet thought a way to solve these 2 problems together without introducing extra structure. Thoughts?

anton-kachurin:
Fixes #524

### DIFF
--- a/packages/mdc-toolbar/mdc-toolbar.scss
+++ b/packages/mdc-toolbar/mdc-toolbar.scss
@@ -59,7 +59,7 @@ $mdc-toolbar-mobile-breakpoint: 599px;
     align-items: flex-start;
     justify-content: center;
     z-index: 1;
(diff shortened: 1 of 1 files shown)


amsheehan:
@anton-kachurin Don't worry about CLA bot. It has a hard time with change 😄 . I will merge this once CI passes

PR #554, opened 2017-04-25
refactor(animation): Refactor switch statement to be simpler

lynnjepsen:
Also improves the branch coverage of our animation tests


amsheehan:
@lynnjepsen looks like a flakey test is the culprit here.

Issue #559, opened 2017-04-25
Replace stylelint `order/declaration-block-properties-order` rule with `order/properties-order`


traviskaufman:
<img width="1005" alt="screen shot 2017-04-25 at 16 38 20" src="https://cloud.githubusercontent.com/assets/1185269/25406739/c45d3fb6-29d5-11e7-88a6-1305bf8009de.png">

https://github.com/hudochenkov/stylelint-order

- [ ] Replace the rule with the above plugin, using the `properties-order` array to preserve the order of our current properties.
- [ ] Fix any files which cause the new lint rule to fail

trimox:
Got it.

trimox:
@traviskaufman, all set. 
Merged on the 26th.

traviskaufman:
Thank you @trimox!

PR #561, opened 2017-04-26
style: Replace stylelint `order/declaration-block-properties-order` rule with `order/properties-order` and update scss to match

trimox:
* Updated stylelint-order to 0.4.3 latest
* Fixed any files which cause the new lint rule to fail
* Replaced the rule with the above plugin, using the properties-order array to preserve the order of our current properties.

Resolves:
https://github.com/material-components/material-components-web/issues/559


amsheehan:
@trimox Don't worry about CLA-bot, it's a little fussy when we update branches. I'll merge once CI passes

PR #581, opened 2017-05-03
feat(tabs): Implement a tab component

amsheehan:
resolves #581


test/unit/mdc-tabs/mdc-tab-bar-foundation.test.js
```diff
@@ -0,0 +1,140 @@
...
+test('#init registers listeners', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const {isA} = td.matchers;
+
+  foundation.init();
+  td.verify(mockAdapter.bindOnMDCTabSelectedEvent());
+  td.verify(mockAdapter.registerResizeHandler(isA(Function)));
+});
+
+if (supportsCssVariables(window)) {
```
traviskaufman: I don't think this needs to be wrapped in this if statement because this logic doesn't use custom variables...?

test/unit/mdc-tabs/mdc-tab-bar-foundation.test.js
```diff
@@ -0,0 +1,140 @@
...
+
+test('#switchToTabAtIndex throws if index is out of bounds', () => {
+  const {foundation, mockAdapter} = setupTest();
+
+  td.when(mockAdapter.getNumberOfTabs()).thenReturn(2);
+
+  assert.throws(() => foundation.switchToTabAtIndex(2, true));
+});
+
+if (supportsCssVariables(window)) {
```
traviskaufman: Same comment above regarding this being wrapped.

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,245 @@
...
+test('attachTo returns a component instance', () => {
+  assert.isOk(MDCTabBar.attachTo(setupTest().root) instanceof MDCTabBar);
+});
+
+test('#get tabs returns tabs', () => {
+  const {component} = setupTest();
+
+  assert.isArray(component.tabs);
+
+  for (let i = 0; i < component.tabs.length; i++) {
```
traviskaufman: nit: `component.tabs.forEach`?

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,245 @@
...
+    component.tabs[1].getDefaultFoundation().adapter_.notifySelected();
+
+    raf.flush();
+
+    assert.isTrue(component.tabs_[1].root_.classList.contains(cssClasses.ACTIVE));
+    raf.restore();
+  });
+}
+
+test('adapter#unbindOnMDCTabSelectedEvent removes listener from component', () => {
```
traviskaufman: I think this test would be more robust if it first added a listener via `bind` and then removed it via `unbind`.

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,245 @@
...
+  const targetIndex = 1;
+
+  component.tabs[targetIndex].isActive = true;
+
+  assert.isTrue(component.getDefaultFoundation().adapter_.isTabActiveAtIndex(targetIndex));
+});
+
+test('adapter#setTabActiveAtIndex sets tab at target index to true or false', () => {
+  const {component} = setupTest();
+  const targetIndex = 0;
```
traviskaufman: I might use `1` here since `0` is active by default

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,245 @@
...
+        <span class="mdc-tab-bar__indicator"></span>
+      </nav>
+    </div>`;
+}
+
+function setupTest() {
+  const fixture = getFixture();
+  const root = fixture.querySelector('.mdc-tab-bar');
+  const indicator = fixture.querySelector('.mdc-tab-bar__indicator');
+  const component = new MDCTabBar(root);
```
traviskaufman: It doesn't look like you're mocking out tabs here. This means that tab bar is not being tested under _isolation_. Bugs in `MDCTab` may cause `MDCTabBar` to fail. I'd suggest using `tabFactory` to create Mock `MDCTab` instances that are guaranteed to work. See https://github.com/material-components/material-components-web/blob/master/test/unit/mdc-select/mdc-select.test.js#L24-L63 for an example

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,245 @@
...
+  assert.isTrue(component.tabs[targetIndex].preventDefaultOnClick);
+});
+
+test('adapter#measureTabAtIndex calls measureSelf on the tab at a given index', () => {
+  const {component} = setupTest();
+  const targetIndex = 1;
+
+  component.getDefaultFoundation().adapter_.measureTabAtIndex(targetIndex);
+
+  assert.equal(component.tabs[targetIndex].computedWidth, component.tabs[targetIndex].root_.offsetWidth);
```
traviskaufman: here's an example of where mocks would be useful. If you had a `MockTab` component used within the tab factory, you could make `measureSelf()` a stub and just verify that it was invoked. This is perfectly acceptable and preferred because:

- `measureSelf()` is part of `MDCTab`'s public API
- `measureSelf()` is tested within `mdc-tab.test.js`
- If `measureSelf()` has bugs, it will not affect this test (nor should it)

test/unit/mdc-tabs/mdc-tab-foundation.test.js
```diff
@@ -0,0 +1,134 @@
...
+  assert.isTrue(foundation.preventsDefaultOnClick());
+});
+
+test('#setPreventDefaultOnClick sets preventDefaultOnClick_' +
+ 'to the value of preventDefaultOnClick argument', () => {
+   const {foundation} = setupTest();
+
+   foundation.setPreventDefaultOnClick(false);
+
+   assert.isFalse(foundation.preventDefaultOnClick_);
```
traviskaufman: 🚫 Please don't reference private properties outside of the classes in which they're defined. 

What this really should be testing is:

- When this is set to `true`, `preventDefault()` is called within the click handler
- When this is set to `false`, `preventDefault()` is _not_ called within the click handler

test/unit/mdc-tabs/mdc-tab-foundation.test.js
```diff
@@ -0,0 +1,134 @@
...
+
+test('#measureSelf sets computedWidth_ and computedLeft_ for tab', () => {
+  const {foundation, mockAdapter} = setupTest();
+
+  td.when(mockAdapter.getOffsetWidth()).thenReturn(200);
+  td.when(mockAdapter.getOffsetLeft()).thenReturn(100);
+
+  foundation.measureSelf();
+
+  assert.equal(foundation.computedWidth_, 200);
```
traviskaufman: 🚫 Same comment as above. Really what you'll want to test is that `getOffsetWidth()` and `getOffsetLeft()` _get called_ as a result of `measureSelf()` getting called.

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+import {createMockRaf} from '../helpers/raf';
+import {supportsCssVariables} from '../../../packages/mdc-ripple/util';
+import {MDCTab} from '../../../packages/mdc-tabs/tab';
+import {MDCTabFoundation} from '../../../packages/mdc-tabs/tab';
+import {cssClasses} from '../../../packages/mdc-tabs/tab/constants';
+
+function getFixture() {
+  return bel`
+    <div>
+      <a class="mdc-tab mdc-tab--active" href="#one">Item One</a>
```
traviskaufman: How come `mdc-tab--active` is added here?

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+}
+
+suite('MDCTab');
+
+test('attachTo returns a component instance', () => {
+  assert.isOk(MDCTab.attachTo(getFixture()) instanceof MDCTab);
+});
+
+if (supportsCssVariables(window)) {
+  test('#destroy cleans up ripple on tab', () => {
```
traviskaufman: we may also want to add a test verifying that the ripple is instantiated

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+  const {root, component} = setupTest();
+
+  assert.equal(component.computedLeft, root.offsetLeft);
+});
+
+test('#get/set isActive', () => {
+  const {component} = setupTest();
+
+  component.isActive = false;
+  assert.isFalse(component.root_.classList.contains(cssClasses.ACTIVE));
```
traviskaufman: use `root` from `setupTest()` instead of `component.root_`

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+import td from 'testdouble';
+import {createMockRaf} from '../helpers/raf';
+import {supportsCssVariables} from '../../../packages/mdc-ripple/util';
+import {MDCTab} from '../../../packages/mdc-tabs/tab';
+import {MDCTabFoundation} from '../../../packages/mdc-tabs/tab';
+import {cssClasses} from '../../../packages/mdc-tabs/tab/constants';
+
+function getFixture() {
+  return bel`
+    <div>
```
traviskaufman: why the wrapper div?

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+  const {component} = setupTest();
+
+  component.isActive = false;
+  assert.isFalse(component.root_.classList.contains(cssClasses.ACTIVE));
+
+  component.isActive = true;
+  assert.isTrue(component.root_.classList.contains(cssClasses.ACTIVE));
+});
+
+test('#get/set preventDefaultOnClick', () => {
```
traviskaufman: this should also test that the foundation logic from `preventDefaultOnClick` kicks in

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+  component.getDefaultFoundation().adapter_.addClass('foo');
+  assert.isTrue(root.classList.contains('foo'));
+});
+
+test('adapter#removeClass removes a class from the root element', () => {
+  const {root, component} = setupTest();
+
+  root.classList.add('foo');
+  assert.isTrue(root.classList.contains('foo'));
```
traviskaufman: I don't think is necessary to test that [DOMTokenList.prototype.add](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList) is working correctly 😉

test/unit/mdc-tabs/mdc-tab.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+test('adapter#deregisterInteractionHandler removes an event listener from the root element', () => {
+  const {root, component} = setupTest();
+  const type = 'click';
+  const handler = td.func('eventHandler');
+
+  root.addEventListener(type, handler);
+
+  component.getDefaultFoundation().adapter_.deregisterInteractionHandler(type, handler);
+  domEvents.emit(root, 'click');
```
traviskaufman: `type`

packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -0,0 +1,177 @@
...
+}
+
+.mdc-tab--active {
+  @include mdc-theme-prop(color, text-primary-on-light);
+
+  @include mdc-theme-dark(".mdc-tab-bar") {
+    @include mdc-theme-prop(color, text-primary-on-dark);
+  }
+
+  &:before {
```
amsheehan: The way IE11 lays out absolutely positioned `table-cell` elements. This was the only way I could get the indicator to line up correctly. 😢

test/unit/mdc-tabs/mdc-tab.test.js
amsheehan: D'oh!
👍

packages/material-components-web/material-components-web.scss
```diff
@@ -33,6 +33,7 @@
 @import "@material/select/mdc-select";
 @import "@material/snackbar/mdc-snackbar";
 @import "@material/switch/mdc-switch";
+@import "@material/tabs/mdc-tabs";
```
amsheehan: Yes. It didn't clear this comment because the comment is in `material-components-web.scss`

test/unit/mdc-tabs/mdc-tab.test.js
amsheehan: 👍

packages/mdc-tabs/tab-bar/index.js
```diff
@@ -0,0 +1,106 @@
...
+  static attachTo(root) {
+    return new MDCTabBar(root);
+  }
+
+  get tabs() {
+    return this.tabs_;
+  }
+
+  get activeTab() {
+    return this.tabs[this.foundation_.activeTabIndex_];
```
traviskaufman: How about `this.tabs[this.activeTabIndex]`?

packages/mdc-tabs/tab-bar/index.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  set activeTab(tab) {
+    this.foundation_.switchToTabAtIndex(this.tabs.indexOf(tab), true);
```
traviskaufman: How about `this.setActiveTab_(tab, false)`?

packages/mdc-tabs/README.md
```diff
@@ -0,0 +1,414 @@
...
+```html
+<nav class="mdc-tab-bar">
+  <a class="mdc-tab mdc-tab--active" href="#one">Item One</a>
+  <a class="mdc-tab" href="#two">Item Two</a>
+  <a class="mdc-tab" href="#three">Three</a>
+  <span class="mdc-tab-bar__indicator"></span>
+</nav>
+```
+
+### Using the JavaScript Component
```
traviskaufman: We should, at some point in this section, talk about `tabFactory` as a way of controlling how `MDCTabBar` instantiates its `MDCTab` components.

packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  // TODO: Replace with breakpoint variable
+  @media screen and (max-width: 600px) {
+    min-width: 72px;
+    padding: 0 12px;
+  }
+
+  // postcss-bem-linter: ignore
```
traviskaufman: why does this need to be ignored?

test/unit/mdc-tabs/mdc-tab-bar-foundation.test.js
```diff
@@ -0,0 +1,149 @@
...
+});
+
+test('#init sets -webkit-transform and transform for indicator', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const raf = createMockRaf();
+
+  foundation.init();
+  raf.flush();
+
+  td.verify(mockAdapter.setStyleForIndicator(td.matchers.anything(), td.matchers.anything()));
```
traviskaufman: this doesn't really verify that `-webkit-transform` and `transform` are set. It just verifies that `setStyleForIndicator` is called for _some_ set of two arguments

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,320 @@
...
+});
+
+test('#set activeTab proxies to foundation_.switchToTabAtIndex', () => {
+  const {root, foundation} = setupTest();
+  const component = new MDCTabBar(root, foundation);
+  const targetIndex = 1;
+  const tab = component.tabs[targetIndex];
+
+  component.activeTab = tab;
+  td.verify(foundation.switchToTabAtIndex(targetIndex, td.matchers.isA(Boolean)));
```
traviskaufman: I'd rather the foundation _not_ be mocked here and we check that the correct classes are added to the mock tab element.

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,320 @@
...
+  tab.isActive = true;
+  assert.equal(component.tabs.indexOf(tab), component.activeTabIndex);
+});
+
+test('#set activeTabIndex proxies to foundation_.switchToTabAtIndex', () => {
+  const {root, foundation} = setupTest();
+  const component = new MDCTabBar(root, foundation);
+
+  component.activeTabIndex = 1;
+  td.verify(foundation.switchToTabAtIndex(1, td.matchers.isA(Boolean)));
```
traviskaufman: same comment as above

demos/tabs.html
```diff
@@ -0,0 +1,549 @@
...
+          </a>
+          <a class="mdc-tab" href="#nearby">
+            <i class="material-icons mdc-tab__icon" aria-label="nearby">person_pin</i>
+          </a>
+          <span class="mdc-tab-bar__indicator"></span>
+        </nav>
+      </section>
+
+      <section>
+        <h2>CSS-Only Icon Labels</h2>
```
amsheehan: That sounds fine. What about keeping CSS-Only? I think it implies that the best experience will be with the fully upgraded JavaScript component, and that sits well with me.

"CSS-Only Icon Tab Labels" ?

packages/mdc-tabs/README.md
amsheehan: 👍

packages/mdc-tabs/README.md
```diff
@@ -0,0 +1,414 @@
...
+import {MDCTabBar, MDCTabBarFoundation} from '@material/tabs';
+
+const tabBar = new MDCTabBar(document.querySelector('#my-mdc-tab-bar'));
+```
+
+## Tab Section
+
+### Tab component API
+
+#### MDCTab.destroy() => void
```
amsheehan: I added a `setter/getter` section to the component API docs to address this. Let me know if there is a more preferred notation. I couldn't really find any examples of documenting instances of public getters AND setters. This seems logical to me

packages/mdc-tabs/README.md
```diff
@@ -0,0 +1,414 @@
...
+#### MDCTabFoundation.measureSelf() => void
+
+Sets `computedWidth_` and `computedLeft_` for a tab.
+
+
+## Tab Bar Section
+
+### Tab Bar component API
+
+#### MDCTabBar.initialize() => void
```
amsheehan: 👍

packages/mdc-tabs/tab-bar/index.js
amsheehan: I think that will result in a circular reference with `get activeTabIndex`. What might be required is a public method like `getActiveTabIndex` in the foundation which calls `activeTabIndex_`.

amsheehan: Implemented, tested, and documented.

packages/mdc-tabs/tab-bar/index.js
amsheehan: 👍

packages/mdc-tabs/tab/mdc-tab.scss
amsheehan: It doesn't

test/unit/mdc-tabs/mdc-tab-bar-foundation.test.js
amsheehan: Correct. I added two additional verifications as well for all of the styles `init()` sets on the indicator.

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,328 @@
...
+test('#get activeTab returns active tab', () => {
+  const {component} = setupTest();
+  const targetIndex = 1;
+  const tab = component.tabs[targetIndex];
+
+  tab.isActive = true;
+  assert.equal(tab, component.activeTab);
+});
+
+test('#set activeTab proxies to foundation_.switchToTabAtIndex', () => {
```
traviskaufman: (here and below) rather than saying "proxies to foundation_.", which describes an implementation detail not seen in this test, it would be better to describe what the implementation is trying to accomplish

test/unit/mdc-tabs/mdc-tab-bar.test.js
```diff
@@ -0,0 +1,328 @@
...
+
+  // NOTE: Because of this issue: https://bugs.chromium.org/p/chromium/issues/detail?id=239648#
+  // Chai cannot observe an error thrown as a result of an event being dispatched.
+  const evtObj = {
+    detail: {tab},
+  };
+  assert.throws(() => component.tabSelectedHandler_(evtObj));
+});
+
+test('adapter#unbindOnMDCTabSelectedEvent removes listener from component', () => {
```
traviskaufman: I don't think this is testing what you think.

You're listening for `MDCTab:selected` on the component but firing it on `tab.root`. Are you sure the event is propagating up to the component level? Because if so that handler should be called, right?

You may instead want to use a similar approach above, and check that `detail.tab` is _not_ active after the event is emitted on the component's root element.

packages/mdc-toolbar/mdc-toolbar.scss
```diff
@@ -44,13 +44,14 @@ $mdc-toolbar-mobile-breakpoint: 599px;
     display: flex;
```
amsheehan: Yes. AFAIK they are only necessary for toolbars with tabs inside of them, so this wouldn't have been fixed anywhere else.

demos/tabs.html
```diff
@@ -0,0 +1,549 @@
+<!DOCTYPE html>
```
amsheehan: Not a problem

test/unit/mdc-tabs/mdc-tab-bar.test.js
amsheehan: You are right. Verified that the handler was getting called. Switched to the approached used in `bindOn...`

PR #585, opened 2017-05-04
fix(textfield): Make bar respect invalid styling

### LINKED ISSUE #584: The textfield component's bottom bar doesn't respect the "invalid styling"
## Bugs

v0.10.0

Chrome

### What OS are you using?

Linux

1. Visit https://material-components-web.appspot.com/textfield.html
2. Scroll down to "Password field with validation"
3. Look at the "Choose password*" textfield

The bottom bar should be red.

The bottom bar is gray.

image

should be

image

The styling is set, but the invalid styling is not specific enough so it gets overridden by the default color.

AustP:
Fixes #584


AustP:
I signed it!

Garbee:
We shouldn't be using important here. Check the specificity and see if the rules conflicting are equal. If they are then the invalid styling simply needs to be moved further down. Otherwise, we need to increase the specificity of the invalid styling.

Important should only be used as an absolute last resort or out of complete necessity for a task.

AustP:
@Garbee 

The specificity for the default is ([on line 148](https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/mdc-textfield.scss#L148))

    .mdc-textfield--upgraded:not(.mdc-textfield--fullwidth):not(.mdc-textfield--multiline):after

The specificity for the invalid styling is ([on line 205](https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/mdc-textfield.scss#L205))

    .mdc-textfield--invalid:not(.mdc-textfield--focused):after

The invalid styling is already below the default styling and I don't see any way to add specificity. I agree that using `!important` should be a last resort, but in this case I think that we are there.

Garbee:
`.mdc-textfield--upgraded.mdc-textfield--invalid:not(.mdc-textfield--focused):after` seems to me like it would do the trick. Do this rule in addition to the existing one (for non-JS environments.) The problem is stemming from the upgraded color scheme for when textfields use JS becoming more specific than the styling for non-JS fields. So with one minor addition for an extra target for upgraded *and* invalid fields, we have the right rule specificity without resorting to `!important` which hurts developers trying to modify things in their applications.

* `.mdc-textfield--upgraded:not(.mdc-textfield--fullwidth):not(.mdc-textfield--multiline):after` - 31 specificity
* `.mdc-textfield--invalid:not(.mdc-textfield--focused):after` - 21 specificity
* `.mdc-textfield--upgraded.mdc-textfield--invalid:not(.mdc-textfield--focused):after` - 31 specificity

AustP:
@Garbee 

Ah good catch. I didn't even think about using `mdc-textfield--upgraded`. I also applied those changes to the dark theme since it was suffering from the same thing.

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -213,8 +214,10 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
 
   @include mdc-theme-dark(".mdc-textfield", true) {
     &:not(.mdc-textfield--focused) {
-      &::after {
-        background-color: $mdc-textfield-error-on-dark;
+      &.mdc-textfield--upgraded {
+        &::after {
```
Garbee: We need to target *both* upgraded and non-upgraded after pseudo elements. If I'm reading this correctly, it will only target the upgraded form.

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -212,7 +212,7 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
 
   @include mdc-theme-dark(".mdc-textfield", true) {
     &:not(.mdc-textfield--focused) {
-      &::after {
+      &.mdc-textfield--upgraded::after {
```
Garbee: we need both non-upgraded and upgraded after pseudo elements here. Both of these should read like so:

```
&::after,
&.mdc-textfield--upgraded::after {
```

AustP:
Thanks for working on this with me @Garbee. I appreciate your help even though you could've just done it yourself and would've gotten it right a long time ago. =D

AustP:
@amsheehan I'm sorry, I haven't really participated in pull requests before. Do I need to do anything? I saw that you put some commits onto my fork, do I need to approve them somehow?

amsheehan:
@AustP Any time we update a branch to catch up with the latest master, Github makes a commit on our behalf. CLA-bot is none too happy about it, but such is life.

TL;DR: No, you don't have to do anything further. 👍

PR #588, opened 2017-05-06
fix(ripple): Re-layout component on activation

### LINKED ISSUE #464: Ripples incorrectly laid out for dialog buttons

0.8.0

Tested on Chrome Version 57.0.2987.133 (64-bit)

### What OS are you using?

macOS v10.12.3

1. Click the "Show Dialog" button
1. Press and hold the spacebar key, observing the ripple on the accept button as you do so

The ripple should cover the entirety of the button

The ripple is sized incorrectly and placed incorrectly within the dialog.

If you open a dialog and then resize the screen, even a little bit, the ripple becomes positioned and sized correctly. This leads me to believe that _this is an issue with when the ripple is being laid out_. 

I think that a solution might be to call `layout()` on all of the ripple instances after the dialog opens. However, this presents an issue since the `layout()` method is not exposed at the `MDCRipple` component level. Perhaps this is something we should look into in order to accomplish this.

anton-kachurin:
Fixes #464

### DIFF
--- a/packages/mdc-ripple/foundation.js
+++ b/packages/mdc-ripple/foundation.js
@@ -201,7 +201,7 @@ export default class MDCRippleFoundation extends MDCFoundation {
     this.adapter_.removeClass(FG_DEACTIVATION);
 
     // Force layout in order to re-trigger the animation.
-    this.adapter_.computeBoundingRect();
+    this.layoutInternal_();
     this.adapter_.addClass(BG_ACTIVE_FILL);
     this.adapter_.addClass(FG_ACTIVATION);
     this.activationTimer_ = setTimeout(() => this.activationTimerCallback_(), DEACTIVATION_TIMEOUT_MS);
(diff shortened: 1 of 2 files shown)


traviskaufman:
So just want to discuss this change a bit.

The reason why we have `layout/layoutInternal` is so we _don't_ have to perform these expensive computations every time the user interacts with the component. This has the very likely chance of causing [layout thrashing](wilsonpage.co.uk/preventing-layout-thrashing/) and delivering sub-par performance, especially on UI-heavy web pages, and/or on resource constrained devices such as phones. 

Re that issue, I think it might be better just to have components manage this manually? Components like dialogs (re the issue) know when they're going to appear and be laid out, so they can intelligently manage laying out the ripple. Perhaps a PR that alters the behavior of dialogs to do this would be better?

anton-kachurin:
Let's keep this PR open until we find a suitable solution and move the discussion to the issue thread.

traviskaufman:
I'm actually going to go ahead and close this for now since we're trying to clear out our ever-growing PR backlog. However I've linked this PR to the open issue and that issue is in our tracker, so we can pick up the convo there if that works. Thanks as always!

PR #592, opened 2017-05-07
fix(drawer): fix RTL closed position of temporary drawer

### LINKED ISSUE #551: Navigation Drawer incorrectly positioned in RTL context
## Bugs

0.9.1

Chrome on Arch Linux (Version 58.0.3029.81 (64-bit))

### What OS are you using?

Arch Linux (rolling release)

Please view this codepen:

When in RTL mode and in the closed state, the drawer is incorrectly positioned inside the page: http://codepen.io/nulldev/pen/LyRBZZ

~~With the drawer is in RTL mode and the header and text in LTR mode (hence creating a LTR right-hand drawer), the margins on the icons are set incorrectly:
http://codepen.io/nulldev/pen/rmMrwN~~

Elemecca:
This fixes #551, in which closed temporary drawers appear in the middle of the screen when RTL because they still use the `transform` value for LTR.

That was happening because `.mdc-temporary-drawer__drawer` was passed as `$root-selector` to the `mdc-rtl` mixin wrapping the RTL drawer styles, which resulted in this style snippet:

```css
[dir="rtl"] .mdc-temporary-drawer__drawer .mdc-temporary-drawer__drawer,
.mdc-temporary-drawer__drawer[dir="rtl"] .mdc-temporary-drawer__drawer {
  -webkit-transform: translateX(107%);
          transform: translateX(107%);
  -webkit-transform: translateX(calc(100% + 20px));
          transform: translateX(calc(100% + 20px));
}
```

That obviously never matches as it's looking for `.mdc-temporary-drawer__drawer` nested inside itself. This drops the `$root-selector` parameter, resulting in the correct style:

```css
[dir="rtl"] .mdc-temporary-drawer__drawer,
.mdc-temporary-drawer__drawer[dir="rtl"] {
  -webkit-transform: translateX(107%);
          transform: translateX(107%);
  -webkit-transform: translateX(calc(100% + 20px));
          transform: translateX(calc(100% + 20px));
}
```

### DIFF
--- a/packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
+++ b/packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
@@ -65,7 +65,7 @@
 
     @include mdc-slideable-drawer;
 
-    @include mdc-rtl(".mdc-temporary-drawer__drawer") {
+    @include mdc-rtl(".mdc-temporary-drawer") {
       @include mdc-slideable-drawer-rtl;
     }
 


packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
```diff
@@ -65,7 +65,7 @@
 
     @include mdc-slideable-drawer;
 
-    @include mdc-rtl(".mdc-temporary-drawer__drawer") {
+    @include mdc-rtl {
```
amsheehan: It looks like this actually just needs the correct root selector instead of the `__drawer` element.

This would be more specific in its output:
```css
@include mdc-rtl(".mdc-temporary-drawer") {
  ...
}
```

Once that's in I think were good to merge this.

Thanks!

PR #594, opened 2017-05-08
Revamp Demo/Catalog

liquidx:
All demo pages use mdc-toolbar.
All demo pages use black/green theme from Material Components.
All demo pages have a hero section that shows the representative example of component.
All demo pages use Roboto Mono for the Toolbar font.

Adds demos.css to be included in all demos to share style.
Adds a CSS reset in demos.css to clear up styles in the demo.


liquidx:
It's pretty much ready for review. The CSS is not the best because there's not a lot of consistency between each component demo, but this consolidates the styling some what.

Let me know if there's a demo template that I should update with the new structure.

Staging version of this is here:
https://staging-dot-material-components-web.appspot.com/

appsforartists:
Seeing a nasty [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) on every click on my Chromebook.

liquidx:
1. I'm not sure about what is going on with the Ripple. Since it may mean touching the component, maybe that should be done in a separate CL. I've been trying to only touch the demo/catalog in this PR. I can file an issue for that.

2. Good point about the "vanilla" theme being hidden. On the flip side, this does ensure we actually support the theming on all our components. I think the consistency with the look of our docs site is important, but I would support adding example where possible of what the default component would look like.

**Card**
Fixed both issues.

**Checkbox**
Fixed and changed highlight to green.

**Dialog**
Fixed.

** Grid List **
Fixed, also replaced the Toggle RTL button with a checkbox.

** Lists **
Ah yes, the list groups look weird with the borders. I'll fix them.

** Select **
Fixed fully featured demo. Oops, used the same id in two places.

**Simple Menu**
Fixed issues with flexbox.

**Snackbar**
Sorry, I think my reordering of the CSS rules exposed that we didn't override the snackbar transform offscreen during loading. Fixed now.

**Toolbar**
Fixed back button issue.
I deliberately left the last two examples as is since it was demonstrating some color changes.

**Typography**
Fixed by adding an override to this.

liquidx:
As for the FOUC - this was already happening before I started styling the demos.

I suggest doing a sweep over all the demos to ensure we hide some of the visibility until page loaded. I had to do some of that in the snackbar demo.

traviskaufman:
Thanks for filing those issue! Agreed that we can hold off on demo ripple issues but we may want to prioritize that for our next iteration cc @omerziv. Also yeah don't worry about the FOUC, that's our fault, not @liquidx's 😉 We will definitely be addressing this when it comes time to refactor our demo site.

demos/demos.css
```diff
@@ -0,0 +1,111 @@
+
+/* http://meyerweb.com/eric/tools/css/reset/
```
traviskaufman: This is actually an incorrect copyright header. Instead we should say something like:

`Adapted from ... <link to ...>`.

Also, I'd prefer it if we used [normalize.css](http://necolas.github.io/normalize.css/) instead for [reasons listed here](http://stackoverflow.com/questions/6887336/what-is-the-difference-between-normalize-css-and-reset-css) and imported it using [cdnjs](https://cdnjs.com/libraries/normalize) via `@import`.

demos/demos.css
```diff
@@ -0,0 +1,111 @@
...
+
+/* Base style */
+
+:root {
+  --mdc-theme-primary: #212121;
+  --mdc-theme-accent: #64DD17;
+}
+
+body {
+  font-family: Roboto, -apple-system, Arial, sans-serif;
```
traviskaufman: could we remove `-apple-system` here? Seems pretty non-standard.

demos/demos.css
```diff
@@ -0,0 +1,111 @@
...
+}
+
+p {
```
traviskaufman: I think by using `normalize.css` you'll be able to get rid of these `p`, `h1`, and `h2` treatments

demos/demos.css
```diff
@@ -0,0 +1,111 @@
+
```
traviskaufman: General comment for this file is to please use consistent indentation (which for us is two spaces). Thanks!

demos/fab.html
```diff
@@ -39,12 +43,57 @@
...
+
+      .hero button {
+        margin-left: 32px;
+        margin-right: 32px;
+
+      }
+
+      /* Override colors to match catalog theme. */
+
+      .mdc-fab {
```
traviskaufman: I don't recommend doing this to `.mdc-fab`, as it's misleading as to how the FAB leverage's MDC-Theme (it is supposed to use it's accent color, not it's primary).

If this is the desired style for the hero you can use the [mdc-theme--primary-bg](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#theme-color-classes) modifier class to achieve it.

demos/fab.html
```diff
@@ -39,12 +43,57 @@
...
+        background-color: #212121;
+      }
+
+      .mdc-fab--plain {
```
traviskaufman: I don't think this is needed as plain fabs have `white` backgrounds by default

demos/fab.html
```diff
@@ -39,12 +43,57 @@
...
+        background-color: #ffffff;
+      }
+
+
+      /* Override mini-FABs with smaller icon size. */
+      .mdc-fab--mini.material-icons .mdc-fab__icon {
+        font-size: 18px;
```
traviskaufman: I'm not so crazy about this, because again it gives users who our looking at canonical demos for our components the incorrect idea about what the sizes of our icons will be. I'd much rather incorporate this into the FAB scss itself in a later PR as long as the design team 🆗 's it

demos/list.html
```diff
@@ -60,32 +62,139 @@
...
+            </a>
+          </li>
+        </ul>
+      </section>
+
+      <section class="preamble mdc-typography--body1">
+        <aside>
+          <p>
+          <em>NOTE:</em> For the purposes of this demo, we've set a max-width of 600px on all
+          <code>mdc-list</code> elements surrounded by a 1px border. This is
```
traviskaufman: nit: can we change "surrounded by a 1px border" to ", and surrounded them by a 1px border". Otherwise, it reads as if we've _only_ set a max-width of 600px on bordered lists, when in reality we set it on all lists.

demos/select.html
```diff
@@ -122,13 +173,15 @@
         </select>
       </section>
 
-      <section>
+      <!--
```
traviskaufman: Can we remove this instead of commenting it out?

demos/simple-menu.html
```diff
@@ -17,11 +17,13 @@
 <html>
```
traviskaufman: This demo still doesn't technically show the menu opening from the top-left and top-right. It positions the buttons there, but it's so far down on the page that the menu opens upward rather than downward. My suggestion would be to add more height to the button's container such that when positioned in the top-left or top-right positions, the menu opens from top to bottom.

demos/card.html
```diff
@@ -134,123 +121,207 @@ <h2 class="mdc-card__subtitle">Subtitle here</h2>
...
-        <section class="mdc-card__actions">
-          <button class="mdc-button mdc-button--compact mdc-card__action">Action 1</button>
-          <button class="mdc-button mdc-button--compact mdc-card__action">Action 2</button>
-        </section>
       </div>
 
-      <div class="mdc-card demo-card">
-        <div class="mdc-card__horizontal-block">
+      <div>
+        <style>
```
liquidx: Done. Not sure why they were here in the first place.

demos/card.html
```diff
@@ -134,123 +121,207 @@ <h2 class="mdc-card__subtitle">Subtitle here</h2>
...
-      <div class="mdc-card demo-card">
-        <div class="mdc-card__horizontal-block">
-          <img class="mdc-card__media-item mdc-card__media-item--3x" src="images/1-1.jpg"></img>
-          <section class="mdc-card__actions mdc-card__actions--vertical">
-            <!-- TODO(sgomes): Replace with icon buttons when we have those. -->
-            <button class="mdc-button mdc-button--compact mdc-card__action">A1</button>
-            <button class="mdc-button mdc-button--compact mdc-card__action">A2</button>
+      <div>
+        <style>
+          .demo-card--small .mdc-card__media {
```
liquidx: Done, as above, not sure why they were here in the first place.

demos/demos.css
liquidx: Done

liquidx: Fixed

demos/demos.css
liquidx: Done

demos/demos.css
liquidx: -apple-system will pull in San Francisco. But agreed, let's pull that out for now.

demos/demos.css
liquidx: Indeed, removed.

demos/fab.html
liquidx: Good point, removed.

demos/fab.html
liquidx: Removed, may had been something I had to do when I had the CSS imports in the wrong order.

demos/fab.html
liquidx: The current icon is too big though for the mini fab. I'll take it up with the design team and remove it.

demos/list.html
liquidx: Done

demos/select.html
liquidx: Done

demos/simple-menu.html
liquidx: I made the menu shorter and the box taller so it accomodated it in more screen sizes.

appsforartists:
Just took a quick skim - looks rad Alastair!

I wonder how much of the list should be clickable.  If it was mobile, the whole row would be a link.  Is that too wide for a larger screen?

At the very least, I think the icons should be clickable.

demos/grid-list.html
```diff
@@ -27,637 +29,720 @@
...
+      </section>
+
+    </main>
     <script src="assets/material-components-web.js" charset="utf-8"></script>
     <script type="text/javascript">
-      function toggleRTL(elem) {
-        var gridListEl = elem.parentElement.querySelector('#example');
-        gridListEl.getAttribute('dir') === 'rtl' ? gridListEl.setAttribute('dir', 'ltr') : gridListEl.setAttribute('dir', 'rtl');
-      }
+      // function toggleRTL(elem) {
```
traviskaufman: could you remove this commented-out code?

demos/grid-list.html
```diff
@@ -27,637 +29,720 @@
...
+      //   var gridListEl = elem.parentElement.querySelector('');
+      //   gridListEl.getAttribute('dir') === 'rtl' ? gridListEl.setAttribute('dir', 'ltr') : gridListEl.setAttribute('dir', 'rtl');
+      // }
       function initGridList() {
         var nodes = document.querySelectorAll('.mdc-grid-list');
         for (var i = 0, node; node = nodes[i]; i++) {
           mdc.gridList.MDCGridList.attachTo(node);
         }
       }
+      // Hook up RTL
```
traviskaufman: nit: comment here isn't needed. The code is self-describing enough :)

demos/simple-menu.html
```diff
@@ -32,10 +34,23 @@
       }
 
       .demo-content {
-        position: relative;
```
traviskaufman: `position: relative` needs to be present here otherwise the `mdc-menu-anchor` won't be positioned correctly.

demos/grid-list.html
liquidx: Done

demos/grid-list.html
liquidx: Done

demos/simple-menu.html
liquidx: Done

liquidx:
Right on! My first commit :)

PR #600, opened 2017-05-09
refactor: Move string literals for events to constants

### LINKED ISSUE #556: code gardening - For events instead of string literals use constants
#### What MDC-Web Version are you using?
0.9.1

#### Purpose
As the api for framework integration is interpreted, the framework will add listeners for events. In order easier to get en overview of the event api collect the events as constants. In current version the events is defined by string literals in index.js.

#### Example
For `mdc-menu` expose event api with `MDCSimpleMenu:selected` and `MDCSimpleMenu:cancel` in e.g.
https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/simple/constants.js

Current string literal `MDCSimpleMenu:selected`
https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/simple/index.js#L95

Current string literal `MDCSimpleMenu:cancel`
https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/simple/index.js#L99

anton-kachurin:
Resolves #556


traviskaufman:
I think we have to do this with tabs but we can just do that as part of #668 cc @amsheehan

anton-kachurin:
@traviskaufman tabs are the last two commits ;)

traviskaufman:
@anton-kachurin oh nice! Thanks again for keeping this in sync 🙏

PR #602, opened 2017-05-09
chore(package): update postcss-loader to version 2.0.3

amsheehan:
Closes #595

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -56,6 +56,7 @@ const CSS_LOADER_CONFIG = [
   {
     loader: 'postcss-loader',
     options: {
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -56,6 +56,7 @@ const CSS_LOADER_CONFIG = [
   {
     loader: 'postcss-loader',
     options: {
+      sourceMap: true,
```
traviskaufman: should this be `process.env.MDC_ENV === 'development'`?

traviskaufman: basically, this should _not_ produce sourcemaps for `npm run build` or `npm run build:min`

PR #615, opened 2017-05-09
fix(menu): Add z-index to MDC Menu to correct stacking related issue

### LINKED ISSUE #614: Menu stacking issue
0.10.0

Chrome Version 58.0.3029.110 (Official Build) (64-bit)

Place a MDC component (e.g. button, textfield) immediately below a Menu, so that the Menu will cover them when opened. Open the menu.

Elements should be behind MDC Simple Menu.

Elements show through the Menu when opened.

trimox:
Fixes #614


traviskaufman:
@amsheehan can you verify that this change does not break any existing demos or components which use menus (e.g. `mdc-select`)? Once that's verified I'm happy to merge!

PR #648, opened 2017-05-15
fix(demos): Add favicon to all demo pages.

liquidx:


robzenn92:
@liquidx would be better if we extend it with this [https://github.com/audreyr/favicon-cheat-sheet](https://github.com/audreyr/favicon-cheat-sheet)?

traviskaufman:
@robzenn92 I think for the purposes of just having a favicon at all for these demos, this works fine for now. Thanks for the suggestion though!

robzenn92:
@traviskaufman cool! thanks

PR #651, opened 2017-05-16
fix(tab-bar): Use proper import for animation

### LINKED ISSUE #650: Version 0.11.0 cannot be imported

0.11.0

1. Install MDC-Web:
```sh
npm i material-components-web
```

2. Import the library:
```javascript
import * as mdc from 'material-components-web';
```

Should produce no errors.

```
Error: Cannot find module "../../../packages/mdc-animation" at webpack:///~/@material/tabs/tab-bar/foundation.js:18:0
```

It is caused by improper import of the `@material/animation` module. This type of errors seriously affects the ability of the third-party developers to update MDC-Web to the latest version.

Can we think of a way how to incorporate checking of this behavior before the release?

anton-kachurin:
Fixes #650

### DIFF
--- a/packages/mdc-animation/README.md
+++ b/packages/mdc-animation/README.md
@@ -123,7 +123,7 @@ MDC Web ships with a set of utility functions designed to make animations easier
 
 To use:
 ```js
-import {getCorrectEventName} from '@material/mdc-animation';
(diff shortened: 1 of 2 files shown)


traviskaufman:
Damn. Merging and submitting a patch now. Thanks for catching this.

Issue #647, opened 2017-05-15
Add border-radius to card component


sgarrity:
It appears that the Material Design Guidelines call for a 2dp radius rounded corner on card components. However, the card component in the material-components-web library doesn't appear to have a rounded corner at all.

If a border radius is added to the outer card div, it may also require `overflow: hidden;` to ensure the corners aren't bled-through by contained images or other elements. This overflow change could have other side-effects to watch out for.

yeelan0319:
Hi @sgarrity ,

Thanks for filing it! Definitely a bug that we overlooked.

I think you are right on the track, basically it should be 
- [] Add `border-radius: 2px;` to `.mdc-card` in `packages/mdc-card/mdc-card.scss`
- [] Add `overflow: hidden` to `.mdc-card` in `packages/mdc-card/mdc-card.scss`

I also double checked in the guideline. Basically the guideline describe card as the entry point and should be simple. There isn't any example that things flow outside. Furthermore, when you use overflow menu on cards as they suggested, the menu will also live in the region of card. So I think it is fine we put `overflow:hidden` there.

Would you want to file a PR for it?

sgarrity:
PR filed. This is my first PR to a Google project, so let me know if there are any procedures or hoops I need to jump through (I tried to follow the commit message guidelines and signed the CLA so far).

Garbee:
FTR hidden overflow on cards was a major pain-point for developers with MDL. Everyone seemed to intuitively think that overflow (especially of menus) should be allowed. That may be something to bring up with the design team to see if it is **really** the absolute behavior or just something that wasn't thought through that deeply before.

sgarrity:
I did wonder about hidden overflow being a problem.

We could assume that the contents of a card need to be aware of the border radius and handle their own border radius. For example, if there is a full-bleed image at the top of a card, that image could also implement a top-left and top-right border radius.

Maybe this discussion should move over to the PR? I'm not what the preferred organization of discussion is.

Garbee:
Contents being aware of the parent is pretty difficult in CSS. Integrating that level of detail would lead to coupled component layers and not be very BEM-friendly to boot.

I really think we either need confirmation from the design team that card contents shouldn't overflow the card boundaries or in some cases (like menus) it can be allowed. That answer itself from the designers can make the call on whether hidden overflow should or should not pass.

Touficbatache:
This issue should be closed.

PR #656, opened 2017-05-16
fix(card): Add rounded corners to card component

sgarrity:
This PR adds a 2px border radius to the card component and sets overflow: hidden; to prevent a contained image from bleeding out over the border.

I did wonder if the 2px border value should be a SCSS variable.

This fixes issue #647


sgarrity:
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

I signed it!

Issue #659, opened 2017-05-17
Toolbar ratio calculation not updated on resize


mhuebert:
The Toolbar foundation sets calculations.toolbarRatio once, on init: 

https://github.com/material-components/material-components-web/blob/master/packages/mdc-toolbar/foundation.js#L167

But this ratio depends on the toolbarRowHeight, which changes on resize: 

https://github.com/material-components/material-components-web/blob/master/packages/mdc-toolbar/foundation.js#L124

The ratio is not updated, however. This means that if you resize the window and cross a threshold such that the toolbarRowHeight changes, the ratio is no longer correct, and an incorrect margin is applied to the sibling element.

I believe this would be a simple fix, by adding 
```js
this.calculations_.toolbarRatio = this.adapter_.getOffsetHeight() / newToolbarRowHeight;
```
after line 124.

traviskaufman:
Thanks for the PR @mhuebert! Left some comments there 😄

mhuebert:
@traviskaufman thanks for the comments! I just submitted PR #697, which includes a test and also conforms to expected commit message format.

PR #660, opened 2017-05-17
Update toolbarRatio when toolbarRowHeight changes

mhuebert:
For #659


mhuebert:
I have tested this locally and for me the behavior is correct. I can resize width across threshholds, and the margin applied to the sibling element remains correct.

traviskaufman:
Thanks for the PR! 

I think it may make more sense here just to call `this.initKeyRatio_()` since it also sets other properties that are dependent on `toolbarRatio` being set correctly.

Furthermore, could we add a unit test here that would fail without the changes being made in this PR?

PR #662, opened 2017-05-17
chore(infrastructure): Add check imports script as a pretest

traviskaufman:
Will hopefully prevent issues like #650 from occurring in the future.
This doesn't cover SCSS, but at least it narrows the surface area
where mistakes can be made. This also in no way replaces vigilance
while reviewing PRs.

---

_Note to reviewers_: I tested this manually be creating true negatives for all `import` cases (absolute imports, module imports not specified as dep, relative imports outside package root)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "lint:css": "stylelint packages/**/*.scss",
     "lint": "npm-run-all --parallel lint:*",
     "postinstall": "lerna bootstrap",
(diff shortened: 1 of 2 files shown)


traviskaufman:
@lynnjepsen sure did! Here's a screenshot of the tail of `npm test` with the incorrect import

screen shot 2017-05-19 at 12 14 29

PR #689, opened 2017-05-19
feat(tabs): Implement a tab bar scroller component

### LINKED ISSUE #642: Implement Tabs Scroller
Now that MDC Tabs has landed, implement a spec aligned scroller component. This includes:

- [ ] Actionable buttons to scroll the tabs horizontally
- [ ] RTL compatibility
- [ ] Accessible through keyboard interactions

https://material.io/guidelines/components/tabs.html#tabs-types-of-tabs

amsheehan:
resolves #642


demos/tabs.html
```diff
@@ -587,8 +641,12 @@ <h1 class="mdc-toolbar__title">Title</h1>
...
         (function() {
           document.getElementById('toggle-rtl').addEventListener('click', function(evt) {
-            evt.target.parentElement.hasAttribute('dir') ? evt.target.parentElement.removeAttribute('dir') : evt.target.parentElement.setAttribute('dir', 'rtl');
-            window.basicTabBar.layout()
+            var mainElement = document.querySelector('#tabs-demo-main-section');
+            mainElement.hasAttribute('dir') ? mainElement.removeAttribute('dir') : mainElement.setAttribute('dir', 'rtl');
+
+            window.basicTabBar.layout();
+            window.tabBarScroller.layout();
+            tabBarScroller.tabBar.layout();
```
traviskaufman: nit: why not reference tabBarScroller using `window` here since it's referenced using `window` everywhere else?

packages/mdc-tabs/README.md
```diff
@@ -5,6 +5,7 @@ The MDC Tabs component contains components which are used to create spec-aligned
 
 - **mdc-tab**: The individual tab elements
 - **mdc-tab-bar**: The main component which is composed of `mdc-tab` elements
+- **mdc-tab-bar-scroller**: The component which controls the horizontal scrolling behavior of `mdc-tab-bars` that overflow their containers
```
traviskaufman: super, super nit: `mdc-tab-bar`s (the "s" goes outside the backtick) 😅

packages/mdc-tabs/README.md
```diff
@@ -284,6 +285,118 @@ import {MDCTabBar, MDCTabBarFoundation} from '@material/tabs';
 const tabBar = new MDCTabBar(document.querySelector('#my-mdc-tab-bar'));
 ```
 
+
+### Using the JavaScript Tab Bar Scroller Component
+
+`mdc-tab-bar-scroller` ships with a Component/Foundation combo for ingesting instances of `mdc-tab-bar`. `mdc-tab-bar-scroller` uses its `initialize()` method call a factory function which gathers and instantiates any tab bar elements that are children of the `mdc-tab-bar-scroller` root element.
```
traviskaufman: nit: How about "which wraps" instead of "for ingesting", which sounds a bit graphic...

packages/mdc-tabs/README.md
```diff
@@ -284,6 +285,118 @@ import {MDCTabBar, MDCTabBarFoundation} from '@material/tabs';
...
+
+The anatomy of `mdc-tab-bar-scroller` includes an instance of `mdc-tab-bar`, forward and back indicators which, when actioned on, move the tab bar left and right, and a scroll frame. The scroll frame is the parent element of the tab bar, and serves to mask the tabs in the tab bar when they overflow the available width.
```
traviskaufman: Should we mention that the tab bar is RTL-aware?

packages/mdc-tabs/README.md
```diff
@@ -284,6 +285,118 @@ import {MDCTabBar, MDCTabBarFoundation} from '@material/tabs';
...
+
+
+#### Including in code
```
traviskaufman: Do you think we need to repeat this entire section or could we just show import `MDCTabBarScroller[Foundation]` instead?

packages/mdc-tabs/README.md
```diff
@@ -284,6 +285,118 @@ import {MDCTabBar, MDCTabBarFoundation} from '@material/tabs';
...
+call `attachTo()` and pass it a DOM element.
+
+```javascript
+mdc.tabs.MDCTabBarScroller.attachTo(document.querySelector('#my-mdc-tab-bar-scroller'));
+```
+
+#### Manual Instantiation
+
+Tab Bar Scrollers can easily be initialized using their default constructors as well, similar
+to `attachTo`. This process involves a factory to create an instance of `MDCTabBar`
```
traviskaufman: We may want to show an example of how to manually pass a `TabBarFactory` function to the component, e.g.

```js
const scrollerEl = document.querySelector('.mdc-tab-bar-scroller');
const tabBarScroller = new MDCTabBarScroller(scrollerEl, /* foundation */ undefined, (el) => new MDCTabBar(el));
```

packages/mdc-tabs/README.md
```diff
@@ -426,3 +539,75 @@ Updates the active tab to be the tab at the given index, emits `MDCTabBar:change
...
+| `isRTL() => boolean` | Returns true if in RTL context. False otherwise. |
+| `registerBackIndicatorClickHandler(handler: EventListener) => void` | Registers an event handler to be called when a `click` event happens on the back indicator |
+| `deregisterBackIndicatorClickHandler(handler: EventHandler) => void` | Deregisters an event handler from a `click` event happening on the back indicator |
+| `registerForwardIndicatorClickHandler(handler: EventHandler) => void` | Registers an event handler to be called when a `click` event happens on the forward indicator |
+| `deregisterForwardIndicatorClickHandler(handler: EventHandler) => void` | Deregisters an event handler from a `click` event happening on the forward indicator. |
+| `registerTabInteractionHandler(evt: string, handler: EventHandler) => void` | Registers an event handler to be called when a `focus`, `touchstart`, or `mousedown` event happens on the root of the component. These events gets dispatched to the listener during the capture phase. They also govern the scrolling behavior when tabs are tabbed to or actioned on. |
```
traviskaufman: maybe instead of `[de]registerTabInteractionHandler` we call it `[de]registerCapturedInteractionHandler`?

packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -0,0 +1,261 @@
...
+        tabIsNotOccluded = tabBarWidthLessTabOffsetLeft > this.currentTranslateOffset_;
+      }
+
+      if (tabIsNotOccluded) {
+        continue;
+      }
+
+      tabWidthAccumulator += this.adapter_.getComputedWidthForTabAtIndex(i);
+
+      const setScrollTarget = tabWidthAccumulator > this.adapter_.getOffsetWidthForScrollFrame();
```
traviskaufman: nit but how about `scrollTargetDetermined` for this var name? It sounds more like a bool than `setScrollTarget` which almost sounds like a method name

packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -0,0 +1,261 @@
...
+      if (setScrollTarget) {
+        scrollTargetIndex = this.adapter_.isRTL() ? i + 1 : i;
+        break;
+      }
+    }
+
+    this.scrollToTabAtIndex_(scrollTargetIndex);
+  }
+
+  scrollForward(evt) {
```
traviskaufman: to make clear that passing an event object here is optional, could we set `evt = null` for the param?

packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -0,0 +1,261 @@
...
+      evt.preventDefault();
+    }
+
+    const scrollFrameOffsetWidth = this.adapter_.getOffsetWidthForScrollFrame() + this.currentTranslateOffset_;
+    let scrollTargetIndex = 0;
+
+    for (let i = 0; i < this.adapter_.getNumberOfTabs(); i++) {
+      const tabOffsetLeftAndWidth =
+        this.adapter_.getComputedLeftForTabAtIndex(i) + this.adapter_.getComputedWidthForTabAtIndex(i);
+      let setScrollTarget = tabOffsetLeftAndWidth > scrollFrameOffsetWidth;
```
traviskaufman: same comment here as above; maybe `scrollTargetDetermined` is a better name?

packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -0,0 +1,261 @@
...
+    this.scrollToTabAtIndex_(scrollTargetIndex);
+  }
+
+  layout() {
+    cancelAnimationFrame(this.layoutFrame_);
+    this.scrollFrameScrollLeft_ = this.adapter_.getScrollLeftForScrollFrame();
+    this.layoutFrame_ = requestAnimationFrame(() => this.layout_());
+  }
+
+  isRTL() {
```
traviskaufman: does this necessarily have to be a public method?

packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -0,0 +1,261 @@
...
+    return this.adapter_.isRTL();
+  }
+
+  handleTabInteraction_(evt) {
```
traviskaufman: maybe `handlePossibleTabKeyboardFocus_` would be more semantic (albeit longer)?

packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    const remainingTabBarWidth = this.adapter_.getOffsetWidthForTabBar() - this.currentTranslateOffset_;
+    if (remainingTabBarWidth > this.adapter_.getOffsetWidthForScrollFrame()) {
+      this.adapter_.addClassToForwardIndicator(INDICATOR_ENABLED);
+    } else {
+      this.adapter_.removeClassFromForwardIndicator(INDICATOR_ENABLED);
+    }
+
+    this.pointerDownRecognized_ = false;
```
traviskaufman: not sure if this is the best place for handling this. Maybe it should go in `handleTabInteraction_` instead?

test/unit/mdc-tabs/mdc-tab-bar-scroller-foundation.test.js
```diff
@@ -0,0 +1,520 @@
...
+
+import {setupFoundationTest} from '../helpers/setup';
+import {verifyDefaultAdapter} from '../helpers/foundation';
+import {createMockRaf} from '../helpers/raf';
+
+import MDCTabBarScrollerFoundation from '../../../packages/mdc-tabs/tab-bar-scroller/foundation';
+
+suite('MDCTabBarScrollerFoundation');
+
+test('exports cssClasses and strings', () => {
```
traviskaufman: _ideally_ these would be two different tests, but that's just me being a stickler for "test one thing only per test case".

packages/mdc-tabs/README.md
amsheehan: 🤢 🤢 🤢

packages/mdc-tabs/README.md
amsheehan: We certainly don't _need_ to repeat it, but I think it should stay for two reasons:

1. It doesn't hurt people who already have a tab bar and want to enhance it with scroller now that its available
1. People coming to tab bar from scratch may want to just scroll down to this, then cut and paste for their flavor of tab bar scroller inclusion.

packages/mdc-tabs/README.md
amsheehan: I think it makes sense to keep some mention of tab in the name. To us, since we anticipate this may have something to do with the act of pressing the tab _key_, there may be some cognitive baggage. For everyone else I think they'll assume the semantics here are describing a tab component, and any interaction associated with it.

packages/mdc-tabs/tab-bar-scroller/foundation.js
amsheehan: Yep. Totally agree.

packages/mdc-tabs/README.md
traviskaufman: ack; the note regarding pasting in the code makes sense.

packages/mdc-tabs/README.md
traviskaufman: I just thought it was confusing because we're actually adding an event listener on the _root_ component, and not the _tab_ component. Reading this, I would've thought we were adding an event listener onto a tab itself.

test/unit/mdc-tabs/mdc-tab-bar-scroller-foundation.test.js
amsheehan: Yeah, that makes sense. Bummer that it adds so much overhead to our test files. 😛

packages/mdc-tabs/README.md
amsheehan: ahh I see. That makes sense. Through that lens I think you're right that captured interaction handler makes more sense.

PR #702, opened 2017-05-22
fix link for Sauce Connect proxy

arijitlaik:
was going thorough this file, the saucelab wiki page seems to have been deleted and replaced


arijitlaik:
I signed it!

PR #711, opened 2017-05-24
docs(checkbox): third example has a checkbox that can be disabled with javascript

### LINKED ISSUE #710: docs(checkbox): would be great to have an example of a disabled checkbox via javascript
With #684 we removed the possibility of having an example of a disabled checkbox via javascript.

I would like to add a button in order to allow the checkbox within the **With Javascript** example to be disabled via MDCCheckbox API. As @anton-kachurin wrote: it was there but then got lost during refactoring.

Please, let me know if any concern.

robzenn92:
Should resolve #710 .


demos/checkbox.html
```diff
@@ -201,6 +202,10 @@
         document.getElementById('make-ind').addEventListener('click', function() {
           checkboxInstance.indeterminate = true;
         });
+
+        document.getElementById('toggle-disabled').addEventListener('click', function() {
```
amsheehan: This should also add the correct `mdc-checkbox--disabled` modifier class to the example `mdc-checkbox` so the hover state behavior is correct.

robzenn92: @amsheehan Hmm I expected it would be added by the API.. I wonder if there is a setter that adds the `mdc-checkbox--disabled` class to the root element without adding it explicitly.

Otherwise, of course, I will address it by adding the following:
`checkboxInstance .addClass('mdc-checkbox--disabled');`

amsheehan: Aha! You are correct. Would you actually mind adding that logic to the `setDisabled` method in the foundation? It would look something like:

1. Add `mdc-checkbox--disabled` to `cssClasses` in `constants.js`
1. Then in the `foundation` add this line to the the `setDisabled()` method

```js
  setDisabled(disabled) {
    this.getNativeControl_().disabled = disabled;
    if (disabled) {
      this.adapter_.addClass(cssClasses.DISABLED);
    } else {
      this.adapter_.removeClass(cssClasses.DISABLED);
    }
  }
```
^ Something like that


I believe that should do it, and then your demo code should work 👍

robzenn92: That is exactly what I was expecting! 😃 I think that every class should have its setters that should be in charge of these changes to the root element. 🎉 🎉

robzenn92:
@amsheehan: comment addressed.
btw, should we add the same behavior other setters or components? I thinks so..

demos/checkbox.html
amsheehan: We will make those decisions on a component by component basis, but it has come up more than just here. Toolbars are another example where we are considering something like this.

If you wouldn't mind adding these to the checkbox I'll give it another review! Thanks

amsheehan:
ping 🏓

robzenn92:
@amsheehan here you go! Sorry for the delay, I am having an hard week! 😄 Let me know. Thanks

PR #716, opened 2017-05-24
fix(linear-progress): Fix version number of linear progress indicator

lynnjepsen:


yeelan0319:
@lynnjepsen 

Please go over the [authoring component doc](https://github.com/material-components/material-components-web/blob/master/docs/authoring-components.md#authoring-components-for-mdc-web) for `package.json` section.

For new component, we set version number to `0.0.0` and put 
```
"publishConfig": {
  "access": "public"
}
``` 
So that lerna will be able to pick up this new scoped component and bump its version correspondingly.

lynnjepsen:
Fixed 

"publishConfig": {
  "access": "public"
} 

and the README. PTAL

PR #741, opened 2017-05-27
[Docs] Adding Discord invitation to #web room

willlarche:


willlarche:
Unnecessary with badge on top.

PR #745, opened 2017-05-30
feat(drawer): Implement dynamic drawer

Yizhachok:
Implement dynamic drawer by taking codebase from temporary drawer and combining it with persistent
drawer. Helping to easy change behaviour on fly without replacing one drawer by another

### DIFF
--- a/demos/drawer/dynamic-drawer-fixed.html
+++ b/demos/drawer/dynamic-drawer-fixed.html
@@ -1,0 +1,184 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+-->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Drawer (Dynamic fixed) - Material Components Catalog</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="icon" type="image/png" href="/images/logo_components_color_2x_web_48dp.png"/>
+    <script src="../assets/material-components-web.css.js" charset="utf-8"></script>
+    <link rel="stylesheet" href="../demos.css">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <style>
+      /* Ensure layout covers the entire screen. */
+      html {
+        height: 100%;
+      }
+
+      /* Place drawer and content side by side. */
+      .demo-body {
+        margin: 0;
+        padding: 0;
+        width: 100%;
+        height: 100%;
+        display: flex;
+        flex-direction: row;
+        box-sizing: border-box;
+      }
+
+      /* Holder for navigation*/
+      .demo-navigation {
+        top: 0;
+        left: 0;
+        width: 100%;
+        height: 100%;
+        display: flex;
+        position: fixed;
+        pointer-events: none;
+        align-items: flex-start;
+      }
+
+      .demo-navigation .mdc-dynamic-drawer__persistent,
+      .demo-navigation .mdc-toolbar {
+        pointer-events: auto;
+      }
+      .demo-navigation .mdc-dynamic-drawer__temporary {
+        position: absolute;
+      }
+
+      /* A simple menu button. */
+      .demo-menu {
+        background: none;
+        border: none;
+        width: 24px;
+        height: 24px;
+        padding: 0;
+        margin: 0 24px 0 0;
+        color: #FFF;
+        box-sizing: border-box;
+      }
+
+      .demo-main {
+        height: 110%;
+        flex-grow: 1;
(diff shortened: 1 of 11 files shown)


Yizhachok:
I signed it!

Yizhachok:
@lynnjepsen please take a look at examples and commit description. If you init temporary or persistent drawer you can't easy change behavior from temporary to persistent or back because there styles that hard-coded for each drawer behavior with special prefix like `mdc-temporary-drawer__drawer`. I suggest to define  `mdc-dynamic-drawer` class with basic styles and add modificator class to root element like `mdc-dynamic-drawer__temporary` or `mdc-dynamic-drawer__persistent` to provide exact behavior. Then when screen size change we can decide witch drawer better to use and change it without removing old drawer and generating new with moving inner content & drawer initialization. On my code we can just write `drawer.state = 'temporary';` to change behavior to another.
Also this commit can be partial solution for https://github.com/material-components/material-components-web/tree/master/packages/mdc-drawer#permanent-drawer-usage

> TODO(sgomes): Give advice on how to hide permanent drawer in mobile.

in mobile we can show temporary drawer by one line of code.

Garbee:
Is there a related issue discussing this implementation already? I'm having trouble finding one.

My initial thought is that this kind of extra component shouldn't really be necessary in the core. It is making a minimum three split drawer with more on the way for other modes. Thinking ahead with the future complexity of drawer components, I'm not sure why this specifically is **necessary** to include in core.

traviskaufman:
Hey @Yizhachok 

Thanks for taking the time to build this 😃  However, as outlined in our [Components Request Policy](https://github.com/material-components/material-components/blob/develop/COMPONENTS_REQUEST_POLICY.md), we don't accept contributions for new or modified components unless they are explicitly detailed within the spec.

Within the spec, temporary and persistent drawers are two different entities. Therefore, something like this would have to be proposed in an issue, discussed by the core team, validated by our UX designers, and then signed off on before we'd begin to review a PR. Thus, I'm going to close this PR.

I definitely see how one may want to switch between these two behaviors, especially in the context of accommodating different form factors. Have you considered writing this component as a standalone component outside of MDC-Web? We have information on how to build MDC-Web compatible components in our [authoring components guide](https://github.com/material-components/material-components-web/blob/master/docs/authoring-components.md). Otherwise, if you'd like to see this land in the core MDC-Web repo, please read our Components Request Policy linked to above and follow the procedure outlined in that doc. 

If you'd like to continue this discussion, please open an issue and we can talk about it there. Thanks again!

piotr-cz:
@traviskaufman 
The problem is that [guidelines](https://material.io/guidelines/patterns/navigation-drawer.html) define Permament drawer as recommended for Desktop and Temporary drawer as required for Mobile and recommended for Tablet.

What would you suggest for a responsive website?

Considering quickly other alternatives (rendering one type based on browser sniffing or switching between two drawers), feature like a dynamic drawer seems to me like best option.

I've tried out the Persistent type, but the content adapts to remaining width which may be 125px and is way too narrow to show anything.

@Yizhachok is you code available as a separate 3p component?

Garbee:
> switching between two drawers

This is exactly the recommended solution. Pick the drawer types you want and where and implement them. Desktop for example can use one of 3 or 4 different drawer types. Supporting a dynamic drawer to handle converting a mobile drawer to any of those would be quite complex. That is why a dynamic drawer is not on the table currently. It is pretty trivial to handle rendering and working with multiple drawers without much drawback.

piotr-cz:
Just to be sure: recommended solution is to render both drawers (Permanent for desktop and Temporary for Mobile/ Tablet) and toggle their visibility based on viewport change?

Yizhachok:
@piotr-cz Sorry for such delayed answer.
Seems like maintainers of this repo didn't understand specific things of web development and maybe trying to move standards from Android development to web that is not so good.
By creating this pull request I suggest to reduce codebase (maybe make other drawers as deprecated) and use single codebase with small CSS changes to let developers to change behavior without redundant HTML code end exclude code duplication. As I understand they interested in support several drawers and bigger codebase (mostly similar between each other) and do things harder and less flexible to use.

In my project I have written "drawer" from zero with automatic hiding on mobile phones via CSS and that's works much better for me.

If you whant you can copy code from my fork but I will not support it in future. Unfortunately
there is no separated 3p module to use.

Garbee:
> Seems like maintainers of this repo didn't understand specific things of web development

Being passive-aggressive about things isn't a good way to conduct conversation. I understand it is frustrating that you put all this effort into a PR that was rejected. That however doesn't mean the maintainers get to be talked down to, even passively.

There is a Contributing guide in the root of the repository with a [Building Components](https://github.com/material-components/material-components-web/blob/master/CONTRIBUTING.md#building-components) section. This outlines what is required to add a component to the system. This outline was not followed with this request. These steps exist to ensure that the components included are modular enough to solve the needs of developers and keep long-term maintenance as minimal as possible. They also exist to make sure the components are well thought out before any code is written. This saves a lot of time and effort programming things that won't pass muster.

I had the exact same discussion with the team member who originally built the drawer system. I wanted a singular drawer that shifted as-needed. But after having a very thoughtful discussion with them about the MD spec's needs and the engineering requirements of MCW I came around to the current multi-drawer method. Even if it adds a little extra markup and CSS parsing, it's well worth it internally to the project to have it this way. In the end, gzip makes the repeated markup overhead quite minimal in *most cases*. That's a trade-off worth it in the MCW long-term goals. If your projects can't work that way, there are ways to work around the issue by adding some custom CSS and JS in your app to either of the drawers. If those are too "hacky", then writing your own drawer component is completely doable as well.

The entire foundation of MCW was build by web engineers. So the web is at the heart of the project and everything it does.

Garbee:
Also, the feedback here that a dual-purpose drawer is desired is useful. That can be taken into account later once release is done to see if something should be done internally about that. There are ways however of having this discussion without writing any code. That way no one's put out too much time into an implementation when it isn't even known if it is wanted yet.

We appreciate all the feedback that developers are willing to provide. This is certainly useful information, it simply isn't being acted upon at the current time.

Issue #253, opened 2017-02-05
Expose mdc-ripple's util.js for ES5


Hardtack:
I'm implementing custom MDCRippleFoundation adapter.
As you mentioned in document, `supportsCssVariables ` function in `util.js` would be useful to implement `browserSupportsCssVars() => boolean`.
But, `@material/ripple/dist/mdc.ripple` does not exposes the `utils` module. So I cannot use the module without transpiler configured for MDC.

``` javascript
/*
 * In `some_library/src/foo.js`
 * Will be transpiled to ES5 code in `some_library/lib/foo.js`
 */
import {supportsCssVariables} from '@material/ripple/util';

...
```

If i use transpiler to use that module, users of some_library also have to use same transpiler. But It does not seem to be good constraints.

``` javascript
/* In application that does not use any transpiler  */

// Raises syntax error when importing `@material/ripple/util`
var RippleWrapper = require('some_library/lib/foo.js'); 
```

codesuki:
I also had that problem.

Anyhow for now you can circumvent it by doing the following:
```
rippleFoundation = new MDCRippleFoundation(Object.assign(MDCRipple.createAdapter(this), {
// your funcs
}
```

This will give you a default foundation that already uses `supportsCssVariables` so you don't have to implement it yourself.

traviskaufman:
Hey @Hardtack,

We try to keep exposition of `util` modules in our ES5 sources to a minimum, and instead encourage people who want to make use of those `util` functions _to_ use a transpiler to transpile our ES2015 sources. And added benefit of transpiling our modules is that you also don't get all of the webpack/UMD-specific artifacts that are present within our ES5 sources. Is there something within your build system that makes it difficult to include MDC-Web sources as a transpilation target?

> If i use transpiler to use that module, users of some_library also have to use same transpiler. But It does not seem to be good constraints.

I'm not sure I understand why users would also have to use a transpiler. If you're transpiling our source files as part of transpiling `some_library`, wouldn't that mean that all an end user would have to do is pull in the entire transpiled bundle?

As @codesuki we made `MDCRipple.createAdapter()` for this purpose of smoothing over some of the more rough edges in the ripple's adapter API. If this solve is inadequate for you, let us know and if there is a strong enough argument for exposing `util` as part of an export on the main ripple component I'm sure we'd consider it.

Hardtack:
@traviskaufman

When I transpile my code to ES5 without webpack just with babel, it does not copy required module into output file. Like this:
[...]

PR #751, opened 2017-05-30
feat(ripple): export util from @material/ripple

Kerrick:
[The docs recommend](https://material.io/components/web/catalog/ripples/#using-the-foundation) using certain functions from util when implementing an adapter for the ripple foundation. It is not currently possible to `import { util } from '@material/ripple';`, even though this pattern is [established](https://github.com/material-components/material-components-web/blob/92aab2b2dffaa9d31ce99d68322dcce90dcffc0a/packages/mdc-toolbar/index.js#L23) in `mdc-toolbar`.

### DIFF
--- a/packages/mdc-ripple/README.md
+++ b/packages/mdc-ripple/README.md
@@ -33,6 +33,7 @@ path: /catalog/ripples/
     - [Specifying known element dimensions](#specifying-known-element-dimensions)
   - [Caveat: Safari](#caveat-safari)
   - [Caveat: Theme Custom Variables](#caveat-theme-custom-variables)
+  - [The util API](#the-util-api)
 
 MDC Ripple provides the Javascript and CSS required to provide components (or any element at all) with a material "ink ripple" interaction effect. It is designed to be efficient, uninvasive, and usable without adding any extra DOM to your elements.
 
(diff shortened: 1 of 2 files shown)


anton-kachurin:
Related issue is #253. It also mentions that docs for ripple need to be expanded.

lynnjepsen:
Actually, I approved a little too quickly...could you also add documentation to the README? And put Resolves #253: in the body of the commit.

Kerrick:
The docs already say the following under **Using the foundation**:

> We provide a `supportsCssVariables` function within the `util.js` which we recommend using, as it handles this for you.

What more should I document about it?

Also, should I add an example to the ES2015 etc. imports (example below)? Was wondering about that since it's under **Adding the ripple JS** rather than **Using the foundation**.

```js
import {MDCRipple, MDCRippleFoundation, util} from '@material/ripple';
```

I'll be glad to make the requested changes once I'm a bit clearer on what's best. :+1:

lynnjepsen:
Mdc-drawer has a good example of documenting **The util API**

https://github.com/material-components/material-components-web/tree/master/packages/mdc-drawer

Mdc-ripple README should document 
- supportsCssVariables
- applyPassive
- getMatchesProperty
- getNormalizedEventCoords

And, again using mdc-drawer as an example, looks like you should add 

`import {MDCRipple, MDCRippleFoundation, util} from '@material/ripple';`

Thank you!

yeelan0319:
Hi @Kerrick, trying to loop back to see if there is any updates :)

Let us know if there is anything you need us help for clarification.

Kerrick:
Whoops, sorry about that! Got caught up in a big product launch at work and lost track. Should be able to get these changes back to you shortly.

Kerrick:
I've pushed the requested changes :+1:

PR #753, opened 2017-05-31
fix link in changelog

hebbet:


robzenn92:
@hebbet I think that CHANGELOG.md still has a lot of broken links due to parenthesis (which I also would like to fix). However, it is generated by `npm run changelog` in [line 26](https://github.com/material-components/material-components-web/blob/master/scripts/post-release.sh#L26). Hence, I think the fix will be lost the next release and it does not worth to fix it. But I am not sure about it. Lets see.

PR #756, opened 2017-05-31
Add Noto Sans as a fallback

yguedidi:
See the last point of [Typeface](https://material.io/guidelines/style/typography.html#typography-typeface): Font stack

### DIFF
--- a/packages/mdc-typography/_variables.scss
+++ b/packages/mdc-typography/_variables.scss
@@ -14,7 +14,7 @@
 // limitations under the License.
 //
 
-$mdc-typography-font-family: Roboto, sans-serif !default;
+$mdc-typography-font-family: Roboto, "Noto Sans", sans-serif !default;
 
 $mdc-typography-font-weight-values: (
   thin: 100,


yguedidi:
I signed it!

shyndman:
Hey @yguedidi,

While this is conceptually correct, I'm not sure it would work well in practice. In reality, Noto is spread across a [number of different fonts](https://www.google.com/get/noto/) would likely be named differently depending on the scripts they support (ie. "Noto Sans Arabic", "Noto Sans Bengali", ...).

That said, I'm not the expert here. @davelab6, can you confirm?

yeelan0319:
@shyndman ,

I think you are right. The Noto font provided by [Google Fonts EA](https://fonts.google.com/earlyaccess) are all named differently. If users do wish to use `Noto Sans` as a `font-family`, they need to write their own `@font-face` rule.

shyndman:
What about on platforms that come with Noto pre-installed (like Android)?

yeelan0319:
Based on the answer [here](https://android.stackexchange.com/questions/85555/how-to-install-noto-font-on-android) and [Noto web site](https://www.google.com/get/noto/#/), I think there is a huge file contains all the Noto Sans called Noto Sans. However, if you download individual language one, they are named by the language...

@davelab6 could you confirm this?

davelab6:
I'm not aware of a single huge file. @marekjez86 do you have any suggestions? :)

marekjez86:
@yguedidi @shyndman @yeelan0319 : Noto family of consists of single script fonts named Noto Sans (or Serif) followed by a script name. Exception is the font that support Latin Greek Cyrillic (LGC) fonts where the script name is omitted. Why it is this way I'm not certain - I wasn't around when this happened.

We have no plans to provide as single file for all fonts nor a single combining script X with LGC (we also get this type of requests).

davelab6:
> Why it is this way I'm not certain

My guess is that in a cost/benefit analysis of various approaches, this one was chosen. 

The benefit is that by having each script in 1 font there is a kind of engineering "modularity" at play which it potentially powerful and useful: Each script can have its own hinting, vertical metrics or other metadata values, rather than 'generic' and 'compromise' values which are "ok" for several scripts at once but may not be ideal. The LGC exception is an example of this, where say the Vietnamese is unusually tall compared to the rest of the Latin and so splitting it into 4+ fonts could enable finer control of font metadata values for each of the scripts within those fonts. 

The cost to such great 'modularity' power, is that it requires great responsibility; an application or operating system must have a 'font fallback stack' system that can reliably weave all these fonts together, when rendering text. 

It is a reasonable position to say, "we have no plans to provide as single file for all fonts nor a single combining script X with LGC" and that implies, "you should plan to have a fall back system that works with what we already provide." 

And here we are! :)

yeelan0319:
Thanks @davelab6 @marekjez86 for the clarification.

I am going to close this PR since the fallback doesn't really have practical meaning in the foreseeable future. Let's left this up to the developer to overwrite for the script they are using instead.

PR #779, opened 2017-06-02
docs(select) Replace mdl-list class with mdc-list in documentation (#778)

### LINKED ISSUE #778: Multi Select - Documentation Fix
The documentation example for a multi-select labels the following:

`<select multiple size="6" class="mdc-multi-select mdl-list">`

When it should be:

`<select multiple size="6" class="mdc-multi-select mdc-list"` (switch mdl-list with mdc-list)

phillycheeze:
fixes #778

The wrong class name was in the documentation example for Multi Select.


phillycheeze:
I signed it!

PR #780, opened 2017-06-02
fix(menu): Add disabled list items to menu

### LINKED ISSUE #706: Select does not respect disabled property on items
## Bugs

MDC 0.3.6

Chromium 58.0.3029.110

### What OS are you using?

Arch Linux

[CodePen demo](http://codepen.io/Holi0317/pen/XRowGQ)
1. The upper one is select with JS enabled, while the other one is not
2. An option, `Milk, Yogurt, and Cheese` is disabled in HTML markup
3. Select it

When an item is disabled in HTML markup property, it should be disabled in both JS or pure CSS version of select component.

 - In JS enabled version of select, we can select that option.
 - In pure CSS version, disabled works correctly.

amsheehan:
resolves #706


packages/mdc-menu/simple/foundation.js
```diff
@@ -37,6 +37,7 @@ export default class MDCSimpleMenuFoundation extends MDCFoundation {
       removeClass: (/* className: string */) => {},
       hasClass: (/* className: string */) => {},
       hasNecessaryDom: () => /* boolean */ false,
+      eventTargetHasClass: (/* target: EventTarget, className: string */) => false,
```
yeelan0319: IMHO, we should make the method name more generic and straightforward about its intention. I feel `elementHasClass: (/*element: DOMElement, className: string*/` sounds more clear when I implement adapter method because the method is opinion-less about whether it is a event target. And all event target are DOMElement based on [MDN](https://dom.spec.whatwg.org/#dom-event-target). WDYT?

packages/mdc-select/README.md
```diff
@@ -154,6 +154,40 @@ style dependencies for both the mdc-list and mdc-menu for this component to func
...
+```html
+<div class="mdc-select" role="listbox">
+  <span class="mdc-select__selected-text">Pick a food group</span>
+  <div class="mdc-simple-menu mdc-select__menu">
+    <ul class="mdc-list mdc-simple-menu__items">
+      <li class="mdc-list-item" role="option" id="grains" tabindex="0">
+        Bread, Cereal, Rice, and Pasta
+      </li>
+      <li class="mdc-list-item mdc-list-item--disabled" role="option" id="vegetables" tabindex="-1" aria-disabled="true">
+        Vegetables
```
yeelan0319: Super nit: Vegetables (Disabled)?

packages/mdc-select/mdc-select.scss
```diff
@@ -153,6 +155,16 @@
         @include mdc-theme-prop(color, text-primary-on-dark);
       }
     }
+
+    &--disabled {
```
yeelan0319: What do you think if we extract this part and make it a mixin class in either `mdc-simple-menu` or `mdc-list`? Then we may remove the duplicated definition here.

packages/mdc-menu/README.md
```diff
@@ -136,6 +136,25 @@ classes:
 | `mdc-simple-menu--open-from-bottom-right` | Open the menu from the bottom right. |
 
 
+#### Disabled list items
```
yeelan0319: Shall we provide a method in menu to set menu item to disabled programmatically? 
I think we could potentially use the ones that defined in `mdc-select`, couldn't we?

amsheehan:
@traviskaufman the only thing that sticks in my craw with using `mdc-list-item[aria-hidden="true"]` in `mdc-list.scss` is that we'll have to use `!important` on every rule because of specificity and the order of the css in `mdc-simple-menu.scss` (it overrides a few `mdc-list-item` styles).

traviskaufman:
@amsheehan That wasn't what I was suggesting. I was suggesting that within `mdc-simple-menu.scss` we add some new style rules to `.mdc-list-item` when used _within mdc-menu_ that styles them differently based on their aria-*disabled* (_not_ `aria-hidden`) property. Wdyt?

amsheehan:
@traviskaufman Aha! Yes. I agree with that entirely. 👍

packages/mdc-menu/README.md
```diff
@@ -136,6 +136,25 @@ classes:
...
+and set `aria-hidden` to true.
+
+```html
+<div class="mdc-simple-menu" tabindex="-1">
+  <ul class="mdc-simple-menu__items mdc-list" role="menu" aria-hidden="true">
+    <li class="mdc-list-item" role="menuitem" tabindex="0">
+      A Menu Item
+    </li>
+    <li class="mdc-list-item mdc-list-item--disabled" role="menuitem" tabindex="-1" aria-hidden="true">
+      Another Menu Item
```
amsheehan: 👍

packages/mdc-menu/simple/foundation.js
amsheehan: I don't think we take this approach anywhere else. We refer to elements as eventTargets pretty much whenever an adapter method will interact with an event.

packages/mdc-select/README.md
amsheehan: Works for me.

packages/mdc-select/mdc-select.scss
amsheehan: We no longer use this modifier class, nor do we set any style in `mdc-list` for a disabled list item. 👍

packages/mdc-menu/simple/foundation.js
traviskaufman: +1 to @amsheehan's comment. The method is opinionless by design, because we [don't want to make reference's to anything DOM-related within an foundation code](https://github.com/material-components/material-components-web/blob/master/docs/authoring-components.md#do-not-reference-host-objects-within-foundation-code), the primary place where an adapter is used.

packages/mdc-menu/README.md
amsheehan: I'm not fundamentally opposed to it. Especially since there is some language around a mechanism like that in the [spec](https://material.io/guidelines/components/menus.html#menus-usage). 
 >Menu items which are relevant but need certain conditions to be met may be disabled. For example, the Copy menu option becomes enabled when text is selected.

I don't know if that's what this story is concerned with though. @traviskaufman WDYT?

packages/mdc-menu/simple/mdc-simple-menu.scss
```diff
@@ -189,6 +190,23 @@ $mdc-simple-menu-item-fade-duration: .3s;
...
+    cursor: default;
+
+    @include mdc-theme-prop(color, text-disabled-on-light);
+
+    @include mdc-theme-dark(".mdc-select") {
+      @include mdc-theme-prop(color, text-disabled-on-dark);
+    }
+  }
+
+  .mdc-list-item[aria-disabled="true"]:focus::before {
```
traviskaufman: nit: you could nest these

```scss
.mdc-list-item[aria-disabled="true"] {
  &:focus::before,
  &:active::before {
    opacity: 0;
  }
}
```

packages/mdc-menu/simple/foundation.js
```diff
@@ -300,6 +301,14 @@ export default class MDCSimpleMenuFoundation extends MDCFoundation {
   }
 
   handlePossibleSelected_(evt) {
+    const targetIsDisabled =
+      this.adapter_.getAttributeForEventTarget(
+        evt.target, MDCSimpleMenuFoundation.strings.ARIA_DISABLED_ATTR
+      ) === 'true';
+    if (targetIsDisabled) {
+      evt.stopPropagation();
```
traviskaufman: yeah, we probably just want to `return` here. We've had lots of trouble with `stopPropagation` as it is...

packages/mdc-menu/simple/mdc-simple-menu.scss
```diff
@@ -189,6 +190,23 @@ $mdc-simple-menu-item-fade-duration: .3s;
     opacity: .18;
   }
 
+  .mdc-list-item[aria-disabled="true"] {
+    cursor: default;
```
traviskaufman: have we considered putting `pointer-events: none` on here? That may also take care of not having to check in event handling logic whether the component is disabled.

packages/mdc-select/mdc-select.scss
```diff
@@ -139,6 +139,8 @@
 
 .mdc-select__menu {
   .mdc-list-item {
+    user-select: none;
```
traviskaufman: dunno if you need this because IIRC `mdc-select__menu` is a `mdc-simple-menu`

demos/select.html
```diff
@@ -69,13 +69,13 @@
...
             <ul class="mdc-list mdc-simple-menu__items">
-              <li class="mdc-list-item" role="option"  aria-disabled="true">
+              <li class="mdc-list-item" role="option" aria-disabled="true">
                 Pick a food group
               </li>
               <li class="mdc-list-item" role="option" tabindex="0">
                 Bread, Cereal, Rice, and Pasta
               </li>
-              <li class="mdc-list-item" role="option"  tabindex="0">
+              <li class="mdc-list-item" role="option" aria-disabled="true" tabindex="-1">
```
amsheehan: It just depends if we want users to be able to tab to disabled options/menu items

packages/mdc-menu/simple/mdc-simple-menu.scss
amsheehan: Yes I did consider it. The problem is the event still triggers handlers on the item's parent which causes selects to close when items with `pointer-events: none` are clicked.

packages/mdc-select/mdc-select.scss
amsheehan: 👍

packages/mdc-menu/simple/mdc-simple-menu.scss
amsheehan: 👍

demos/select.html
traviskaufman: I believe that disabled items should not be focusable. Also, `-1` for a tabindex doesn't allow users to tab to it; I believe it means that JS can `focus()` it if need be. So yeah I feel like removing it would be the better option.

packages/mdc-menu/simple/mdc-simple-menu.scss
traviskaufman: Ahh of course. That makes sense.

packages/mdc-menu/simple/foundation.js
amsheehan: The other option is to pass the event to the close method and check if the target is disabled there. The problem is that there is a document click handler that closes the select that gets triggered if we let the event propagate.

amsheehan: You know what, I just implemented it that way and I think it's better. That way the user can still listen for events on a disabled item and do whatever custom thing they may want.

demos/select.html
amsheehan: 👍  I'm convinced

amsheehan:
Alright, the biggest change in the latest commit is just that now we check for `aria-disabled` in both `handlePossibleSelected_()` and `close()` in the `foundation`.

1. We check it in `handlePossibleSelected_()` to prevent the custom event from firing which eventually sets `aria-selected` after a delay (resulting in a _disabled_ option with _selected_ styles)
1. We now have an optional parameter `evt` in `close()`'s signature. We also check for `aria-disabled`, and if it exists we return right away to prevent the select menu from closing.

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -271,6 +271,39 @@ testFoundation('#open anchors the menu on the top left in LTR when not close to
...
+
+    td.when(mockAdapter.getAttributeForEventTarget(td.matchers.anything(), strings.ARIA_DISABLED_ATTR))
+      .thenReturn('true');
+
+    foundation.close(mockEvt);
+
+    td.verify(mockAdapter.deregisterDocumentClickHandler(td.matchers.anything()), {times: 0});
+  });
+
+// const {foundation, mockAdapter} = setupTest();
```
yeelan0319: Super nit: May be remove it if we no longer use them?

amsheehan: Ha. Not a super nit, super mistake on my part.

packages/mdc-menu/README.md
```diff
@@ -256,8 +275,8 @@ The adapter for temporary drawers must provide the following functions, with cor
 | `getNumberOfItems() => numbers` | Returns the number of _item_ elements inside the items container. In our vanilla component, we determine this by counting the number of list items whose `role` attribute corresponds to the correct child role of the role present on the menu list element. For example, if the list element has a role of `menu` this queries for all elements that have a role of `menuitem`. |
 | `registerInteractionHandler(type: string, handler: EventListener) => void` | Adds an event listener `handler` for event type `type`. |
 | `deregisterInteractionHandler(type: string, handler: EventListener) => void` | Removes an event listener `handler` for event type `type`. |
-| `registerDocumentClickHandler(handler: EventListener) => void` | Adds an event listener `handler` for event type 'click'. |
-| `deregisterDocumentClickHandler(handler: EventListener) => void` | Removes an event listener `handler` for event type 'click'. |
+| `registerBodyClickHandler(handler: EventListener) => void` | Adds an event listener `handler` for event type 'click'. |
```
traviskaufman: **N.B.** This is now a breaking change due to the adapter API methods having changed, so we can't forget to include that info in the final commit message

amsheehan: 👍

PR #781, opened 2017-06-02
feat(slider): Implement continuous slider component

traviskaufman:
- Add continuous slider component
- Check in `package-lock.json`, part of npm5/node8
- Disable stylelint for `demos.css`

Part of #25


packages/mdc-slider/README.md
```diff
@@ -0,0 +1,274 @@
...
+
+const slider = new MDCSlider(document.querySelector('.mdc-slider'));
+slider.listen('MDCSlider:change', () => console.log(`Value changed to ${slider.value}`));
+```
+
+You can also include MDCSlider via its UMD version located at `dist/mdc.slider[.min].js`
+
+```js
+// CommonJS
+const {MDCSlider} = require('@material/slider/dist/mdc.slider');
```
yeelan0319: Correct me if I am wrong, but I think `const MDCSlider = require('mdc-slider')` works?

yeelan0319:
Also there is a noticeable flash in thumb when user set value of slider to 0 (or they provide no value), which I don't feel it is a good UX experience.

The reason causing that is because we always add `mdc-slider--off` in the `constructor` if there are needed and that may be executed after the slider is fully rendered. 

I think there are potentially 2 strategies here. 
1. Thumbs are set to `display:none` before the `constructor` is completed and we only show it after we fully understand how it would be rendered.
2. In the document, we suggest the users of `mdc-slider` to manually add `mdc-slider--off` in their mark up if they know the slider have `value==0` in the first place. 

IMHO, I prefer we handle the logic if that is not too much of overhead. WDYT?

test/unit/mdc-slider/foundation.test.js
```diff
@@ -0,0 +1,728 @@
...
+  const {foundation} = setupTest();
+  assert.equal(foundation.getMin(), 0);
+});
+
+test('#constructor sets the default slider step to 0 (no step)', () => {
+  const {foundation} = setupTest();
+  assert.equal(foundation.getStep(), 0);
+});
+
+test('#constructor sets the disabled state to enabled', () => {
```
yeelan0319: super nit: the default disabled state

test/unit/mdc-slider/foundation.test.js
```diff
@@ -0,0 +1,728 @@
...
+
+  td.when(mockAdapter.computeBoundingRect()).thenDo(() => {
+    numComputations++;
+    return /* dummy value */ {};
+  });
+
+  foundation.layout();
+  foundation.layout();
+
+  assert.equal(numComputations, 2);
```
yeelan0319: Maybe use `td.verify(mockAdapter.computeBoundingRect(), {times: 2});` and remove the counter instead?

test/unit/mdc-slider/foundation.test.js
```diff
@@ -0,0 +1,728 @@
...
+
+  foundation.setValue(20);
+
+  td.verify(mockAdapter.notifyInput(), {times: 0});
+  td.verify(mockAdapter.notifyChange(), {times: 0});
+
+  raf.restore();
+});
+
+test('#getMax/#setMax basic functionality', () => {
```
yeelan0319: super nit: feel the test description should be more descriptive.

How about "Slider Max getter and setter work correctly"?

test/unit/mdc-slider/foundation.test.js
```diff
@@ -0,0 +1,728 @@
...
+  raf.flush();
+
+  foundation.setMax(50);
+
+  td.verify(mockAdapter.setAttribute('aria-valuemax', '50'));
+
+  raf.restore();
+});
+
+test('#getMin/#setMin basic functionality', () => {
```
yeelan0319: Super nit: same thing here.

test/unit/mdc-slider/foundation.test.js
```diff
@@ -0,0 +1,728 @@
...
+  raf.flush();
+
+  foundation.setMin(10);
+
+  td.verify(mockAdapter.setAttribute('aria-valuemin', '10'));
+
+  raf.restore();
+});
+
+test('#getStep/#setStep basic functionality', () => {
```
yeelan0319: Super nit: and here

test/unit/mdc-slider/foundation.test.js
```diff
@@ -0,0 +1,728 @@
...
+  foundation.setStep(1);
+  raf.flush();
+
+  td.verify(mockAdapter.setThumbContainerStyleProperty(TRANSFORM_PROP, 'translateX(10px) translateX(-50%)'));
+  td.verify(mockAdapter.setTrackStyleProperty(TRANSFORM_PROP, 'scaleX(0.1)'));
+
+  raf.restore();
+});
+
+test('#isDisabled/#setDisabled basic functionality', () => {
```
yeelan0319: Super nit: and here.

test/unit/mdc-slider/foundation-keyboard-events.test.js
```diff
@@ -0,0 +1,661 @@
...
+ */
+
+import {assert} from 'chai';
+import td from 'testdouble';
+
+import {TRANSFORM_PROP, setupEventTest as setupTest} from './helpers';
+
+suite('MDCSliderFoundation - keyboard events');
+
+function createFakeEvent(info = {}) {
```
amsheehan: This would be useful in a lot of places. Consider pulling it out into one of the helper files?

test/unit/mdc-slider/foundation-keyboard-events.test.js
```diff
@@ -0,0 +1,661 @@
...
+  raf.flush();
+
+  assert.equal(foundation.getValue(), 46);
+  td.verify(mockAdapter.setThumbContainerStyleProperty(TRANSFORM_PROP, 'translateX(46px) translateX(-50%)'));
+  td.verify(mockAdapter.setTrackStyleProperty(TRANSFORM_PROP, 'scaleX(0.46)'));
+
+  raf.restore();
+});
+
+test('on any other keydown does nothing', () => {
```
amsheehan: nit: I mean... kind of. This really just tests if any other key is the Enter key.

packages/mdc-slider/foundation.js
```diff
@@ -0,0 +1,420 @@
...
+    this.adapter_.deregisterInteractionHandler('keydown', this.keydownHandler_);
+    this.adapter_.deregisterInteractionHandler('focus', this.focusHandler_);
+    this.adapter_.deregisterInteractionHandler('blur', this.blurHandler_);
+    ['mousedown', 'pointerdown', 'touchstart'].forEach((evtName) => {
+      this.adapter_.deregisterThumbContainerInteractionHandler(evtName, this.thumbContainerPointerHandler_);
+    });
+    this.adapter_.deregisterResizeHandler(this.resizeHandler_);
+  }
+
+  layout() {
```
amsheehan: This updates UI, but does nothing to alert the user of the value change. e.g.: If a user programmatically changes the step value, we see the slider move to the closest step. However, we don't get an event for the value change.

demos/slider.html
```diff
@@ -0,0 +1,223 @@
...
+        padding: 0 16px;
+      }
+    </style>
+  </head>
+  <body class="mdc-typography">
+    <header class="mdc-toolbar mdc-toolbar--fixed">
+      <div class="mdc-toolbar__row">
+        <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
+          <span class="catalog-back">
+            <a href="/"><i class="material-icons">&#xE5C4;</i></a>
```
traviskaufman: Ah I see. I believe other demo code needs to be updated as well.

demos/slider.html
```diff
@@ -0,0 +1,223 @@
...
+      #hero-slider-wrapper {
+        margin: 0 auto;
+        width: 100%;
+        max-width: 600px;
+        /* Blended equivalent of rgba(0, 0, 0, .05); */
+        --mdc-slider-bg-color-behind-component: #f2f2f2;
+      }
+
+      .mdc-theme--dark {
+        background-color: #333;
```
traviskaufman: Agreed and noted. Should I file an issue? Or should we do this within demos.scss and make a general note somewhere?

demos/slider.html
traviskaufman: ...nvm!

demos/slider.html
amsheehan: I think an issue is appropriate. That way we can toss up a `help-wanted` label and see if anyone bites.

packages/mdc-slider/README.md
traviskaufman: It actually doesn't, since in `package.json` we specify the package name as `@material/slider`

packages/mdc-slider/README.md
```diff
@@ -0,0 +1,274 @@
...
+| `getMin() => number` | Returns the min value the slider can have |
+| `setMin(min: number) => number` | Sets the min value the slider can have |
+| `getStep() => number` | Returns the step value of the slider |
+| `setStep(step: number) => void` | Sets the step value of the slider |
+| `isDisabled() => boolean` | Returns whether or not the slider is disabled |
+| `setDisabled(disabled: boolean) => void` | Disables the slider when given true, enables it otherwise. |
+
+### Theming
+
+All thematic elements of sliders make use of the **primary theme color**, including when the
```
traviskaufman: Yes, but we document that directly within the demo itself. Do you think it's not called out enough?

packages/mdc-slider/index.js
```diff
@@ -0,0 +1,145 @@
...
+      deregisterInteractionHandler: (type, handler) => {
+        this.root_.removeEventListener(type, handler);
+      },
+      registerThumbContainerInteractionHandler: (type, handler) => {
+        this.thumbContainer_.addEventListener(type, handler);
+      },
+      deregisterThumbContainerInteractionHandler: (type, handler) => {
+        this.thumbContainer_.removeEventListener(type, handler);
+      },
+      registerBodyInteractionHandler: (type, handler) => {
```
traviskaufman: 👍 techdebt issue?

packages/mdc-slider/README.md
```diff
@@ -0,0 +1,274 @@
...
+#### Setting the correct background color for disabled/off slider thumbs
+
+One tricky issue with sliders is how the thumb is supposed to look when in the disabled or
+"off" states. In these cases, certain portions of the slider's thumb and track are supposed to
+become "transparent" and reveal the background color behind it. However, this presents a problem as
+there is no elegant way to derive what the background color behind the slider should be. We could
+theoretically walk up the DOM until we found an ancestor with a set background, but that would break
+the component's encapsulation model.
+
+To solve this, you can supply a css custom property `--mdc-slider-bg-color-behind-component`. When
```
traviskaufman: Ahh interesting. I interpret your question as what happens if there's some non-solid background behind the slider component (e.g. a background image) where the slider element is overlayed on top of it. I'm not sure how this would be handled, but I feel like it's also a bit of an edge case?

packages/mdc-slider/index.js
```diff
@@ -0,0 +1,145 @@
...
+
+  getDefaultFoundation() {
+    return new MDCSliderFoundation({
+      addClass: (className) => this.root_.classList.add(className),
+      removeClass: (className) => this.root_.classList.remove(className),
+      getAttribute: (name) => this.root_.getAttribute(name),
+      setAttribute: (name, value) => this.root_.setAttribute(name, value),
+      removeAttribute: (name) => this.root_.removeAttribute(name),
+      computeBoundingRect: () => this.root_.getBoundingClientRect(),
+      getTabIndex: () => this.root_.tabIndex,
```
traviskaufman: I'd rather use `tabIndex` because it's guaranteed to be a number, and can't be corrupted by invalid attribute data, e.g. `<div tabindex="nope">`

packages/mdc-slider/mdc-slider.scss
```diff
@@ -0,0 +1,226 @@
...
+    width: 21px;
+    height: 21px;
+    transition: transform 266.67ms ease-out, opacity 266.67ms ease-out, background-color 266.67ms ease-out;
+    border-radius: 50%;
+    opacity: 0;
+  }
+}
+
+.mdc-slider--active {
+  &:not(.mdc-slider--disabled) .mdc-slider__thumb {
```
traviskaufman: yeah dunno why I did that...removing `:not(.mdc-slider--disabled)` since the slider can't ever be active if also disabled

demos/slider.html
```diff
@@ -0,0 +1,223 @@
...
+          var committedValue = demoRoot.querySelector('.committed-value');
+          var sliderEl = demoRoot.querySelector('.mdc-slider');
+          var slider = new mdc.slider.MDCSlider(sliderEl);
+
+          slider.listen('MDCSlider:input', function() {
+            value.textContent = slider.value;
+          });
+
+          slider.listen('MDCSlider:change', function() {
+            committedValue.textContent = slider.value;
```
traviskaufman: Sorry but I'm not sure what you mean here...

test/unit/mdc-slider/foundation.test.js
traviskaufman: I think "correctly" is equally as vague, if not more. What is "correctly"? How many aspects of the functionality does "correctly" cover? How about something like `#setMax/#getMax sets/retrieves the maximum value, respectively`. It's more verbose but also more descriptive.

test/unit/mdc-slider/foundation.test.js
traviskaufman: If you try and `td.verify` a method that's been mocked out, testdouble will emit a warning ([rightfully so](https://github.com/testdouble/testdouble.js/blob/master/docs/B-frequently-asked-questions.md#why-shouldnt-i-call-both-tdwhen-and-tdverify-for-a-single-interaction-with-a-test-double)). Added an explanation as to why we need to stub and verify , thus use this method, in a comment.

test/unit/mdc-slider/foundation-keyboard-events.test.js
traviskaufman: Are there places you can identify currently where this would be useful? Otherwise I'm skeptical hesitant to prematurely refactor it.

traviskaufman:
> Also there is a noticeable flash in thumb when user set value of slider to 0 (or they provide no value), which I don't feel it is a good UX experience...IMHO, I prefer we handle the logic if that is not too much of overhead. WDYT?

I'm not sure we can handle this overhead automatically. Because the slider programmatically sets all of its values, we have to wait for the script to execute for those styles to automatically be applied, thus there's always a potential for FOUC. I think your suggestion regarding manually adding the class for the off state is a good one; I'll go ahead and add that to the README.

packages/mdc-slider/foundation.js
traviskaufman: I think that's fine. When you update an `HTMLInputElement`'s `value` prop programmatically you don't get an event either, so this keeps our behavior consistent with the way the DOM API behaves.

traviskaufman:
All comments addressed. PTAL!

test/unit/mdc-slider/foundation.test.js
yeelan0319: Works for me. I didn't noticed that warning message, good testdouble!

packages/mdc-slider/index.js
yeelan0319: +1 to @traviskaufman .I think it is also how [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/tabIndex) recommend for accessing tabIndex.

yeelan0319:
Agree to your concern about the racing condition in initialization, also noticed you changed the initial value of the demo. 👍  

PR looks all good to me.

traviskaufman:
Spoke with @amsheehan offline who said LGTM.

PR #764, opened 2017-05-31
fix(toolbar): Improve layout and scrolling logic of items in toolbars

amsheehan:
resolves #607

Issue #598, opened 2017-05-08
Unstyled icons in toolbar


joshas:
If I take HTML code from example https://material-components-web.appspot.com/drawer/temporary-drawer.html I need additional style to remove default button styles and set button size. Shouldn't menu button styling be defined in material-components-web CSS?

Garbee:
Do you mean buttons as list items within the drawer? If so see #505.

Otherwise, could you be more specific about what you're expecting? I'm not quite sure what it is you're asking about exactly.

joshas:
I had issue with "hamburger" menu icon, but now implemented it from toolbar example, where it has other issues.

donmccurdy:
Was this resolved? I'm copying one of the toolbar examples from the documentation:

```html
<header class="mdc-toolbar">
  <div class="mdc-toolbar__row">
    <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
      <a href="#" class="material-icons">menu</a>
      <span class="mdc-toolbar__title">Title</span>
    </section>
  </div>
</header>
```

And getting this:

<img width="916" alt="screen shot 2017-05-22 at 4 26 09 pm" src="https://cloud.githubusercontent.com/assets/1848368/26332424/795f5cce-3f0b-11e7-9be2-91ec1e57c6b1.png">

I assume I need to do something to get icons working, but can't find what that is.

joshas:
@donmccurdy To get icons working, you need to load Material Icons stylesheet
```html
<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```

But you'll get unstyled icons. To fix it demo uses a bit of custom css:
```css
.material-icons {
    text-decoration: none;
    cursor: pointer;
    padding-right: 8px;
    color: white;
}
```
For use in production you'll probably want to make icon styles specific to toolbar only and maybe set `color: inherit` so icons would get correct color when changing theme.

I'd expect that icon colors should be handled by framework.

donmccurdy:
Thanks @joshas — your snippet works nicely in my case. 

I was expecting icon colors to match the theme automatically; if that's out of scope for some reason, a section in the docs discussing how to get icons working and properly colored would be great. Glad to send a PR for that.

yeelan0319:
We agree with having this feature out of box is totally a valid ask. To address this problem, we would like to breakdown it into 2 sub-problems:

1. Correct wrong positioning for icon in toolbar. This is already captured by #607 file by @joshas 
[...]

PR #785, opened 2017-06-04
docs(toolbar): Add warning that the icons used in the toolbar demos aren't loaded in by default.

danielbolan:
I was going to try to help with with #598, but it seems that @amsheehan already took care of all the work with #764 :)

This just closes out the last action item instead, adding some docs for the demo page to let people know why their icons aren't showing up when they copy the example code.

### DIFF
--- a/demos/toolbar/index.html
+++ b/demos/toolbar/index.html
@@ -66,7 +66,7 @@
         margin-bottom: 16px;
       }
 
-      p.intro {
+      div.intro {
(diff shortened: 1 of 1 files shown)


danielbolan:
Oops, I think when I added my work email to this account I forgot to make it public. Ah well, I signed it anyways.

amsheehan:
Aha! I just saw it was in Yiran's comment to add this. Thanks for the PR!

PR #792, opened 2017-06-05
fix(demos): Fix non-unique ids in radio demo

jscissr:
Previously, the first set of radios would change when clicking the labels of the disabled radios.

https://material-components-web.appspot.com/radio.html

### DIFF
--- a/demos/radio.html
+++ b/demos/radio.html
@@ -90,23 +90,23 @@
         <h2>With Javascript</h2>
         <div class="mdc-form-field">
           <div class="mdc-radio">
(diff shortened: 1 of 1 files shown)


jscissr:
Just found another one: In https://material-components-web.appspot.com/switch.html the first 'off/on' label changes the centered toggle

traviskaufman:
@jscissr Thanks for helping us track these down. Feel free to submit a separate PR for that one, or just roll it into here.

traviskaufman:
@jscissr going to merge this in. Feel free to open another PR for switches and thanks again!

jscissr:
I see now the other bug was already fixed in #747, the demo page just hasn't been updated yet.

Issue #675, opened 2017-05-18
Implement updated UX states for text fields


traviskaufman:
- [x] Move `mdc-textfield__bottom-line` out of `mdc-textfield--box` scope and put within the general `mdc-textfield` scope, replacing any `::after` style rules with `.mdc-textfield__bottom-line` style rules
- [x] Idle
- [x] Hover
- [x] Focus ("Press" in spec)
- [x] Invalid ("Error" in spec)
- [x] Ensure textfield can gracefully degrade to CSS-Only
- [x] Ensure all states work for dark mode

_Breaking change_: The underline for `mdc-textfield`s will need to be graduated to a DOM element

PR #801, opened 2017-06-08
Text field boxes + Ripple Improvements

### LINKED ISSUE #673: Implement Text Field Boxes
Note: This does _not_ cover the new UX interaction states for text fields in general. That is covered by a separate issue.

- [x] Add a `mdc-textfield--box` class which provides the boxed background for a textfield, and alters any baselines that need to be alter
- [x] Implement the ripple treatments for text field boxes.
- [x] Ensure text field box dark mode treatments are in place
- [x] Ensure dense mode treatments are in place
- [x] Ensure text field boxes work within an RTL context
- [x] Ensure text field box can gracefully degrade

traviskaufman:
This PR constitutes a number of commits which resolve #673 and should be **rebased and merged**.

- fix(ripple): Remove fg deactivation class when animation finishes
- feat(ripple): Add layout() method to component
- feat(ripple): Reduce the fade out time for foreground ripple effect
- feat(textfield): Implement text field boxes


packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -123,10 +124,197 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
...
+  @include mdc-theme-dark(".mdc-textfield", true) {
+    background-color: rgba(white, .1);
+  }
+
+  &::before,
+  &::after {
+    position: absolute;
+    top: 0;
+    left: 0;
+    // TODO ripple positioning logic?
```
yeelan0319: Is this associate with other issues? Feel better to leave a reference here

test/unit/mdc-textfield/mdc-textfield.test.js
```diff
@@ -48,13 +49,54 @@ test('#constructor assigns helptextElement to the id specified in the input aria
...
+  const component = new MDCTextfield(root, undefined, (el) => new FakeRipple(el));
+  assert.equal(component.ripple.root, root);
+});
+
+test('#constructor sets the ripple property to `null` when given a non `mdc-textfield--box` element', () => {
+  const component = new MDCTextfield(getFixture());
+  assert.isNull(component.ripple);
+});
+
+test('#constructor initializes a default ripple when no ripple factory given when given a ' +
```
yeelan0319: How about "constructor initializes a default ripple when ripple factory is not given in `mdc-textfield--box` element?"

test/unit/mdc-textfield/mdc-textfield.test.js
```diff
@@ -48,13 +49,54 @@ test('#constructor assigns helptextElement to the id specified in the input aria
...
+});
+
 function setupTest() {
   const root = getFixture();
   const component = new MDCTextfield(root);
   return {root, component};
 }
 
-test('#adapter.initialSyncWithDom', () => {
+test('#initialSyncWithDom', () => {
```
yeelan0319: How about "#initialSyncWithDom set disable to false"?

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -123,10 +124,197 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
...
+    }
+
+    &--float-above {
+      transform: translateY(-50%) scale(.75, .75);
+    }
+  }
+
+  &.mdc-ripple-upgraded {
+    @include mdc-ripple-base;
+    // TODO: Add ::before as ripple bg once #194 lands
```
amsheehan: It looks like the background fade out on the hover ripple doesn't exist, making it seem like it disappears very abruptly on interaction end events (moving the pointer out of the target). Do you have any sense of whether or not this component will receive ink wash on hover as a result of #194 ? 

If it does not I _think_ it will solve this problem as the ripple on click interactions fades out as expected.

amsheehan:
Can I make a request that we make the docs changes to `full width text fields` part of this PR? Essentially we need to note that `full width text fields` do not support labels. Eh?

packages/mdc-textfield/mdc-textfield.scss
traviskaufman: Old TODO :p removing

test/unit/mdc-textfield/mdc-textfield.test.js
traviskaufman: That to me sounds like there's some "ripple factory" _inside_ of an `mdc-textfield--box` element. How about: "#constructor when given a `mdc-textfield--box` element, initializes a default ripple when no ripple factory given"

packages/mdc-textfield/mdc-textfield.scss
traviskaufman: @amsheehan yeah it will. There is a fade out, but it is very quick (<100ms). You can see it in devtools if you slow down the animations using the "Animations" tab. I worked with motion design on it and they signed off on it, so I'm assuming it's correct?

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -123,10 +124,197 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
...
+  }
+
+  // TODO: Move into general .mdc-textfield decl and replace ::after textfield styles with this
+  // element as part of #675
+  .mdc-textfield__bottom-line {
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    width: 100%;
+    height: 1px;
```
traviskaufman: Yeah so this is a change that hasn't yet made it into spec. When we implement #675 it'll elevate to 2dp on hover, and stay at 2dp on focus.

demos/textfield.html
```diff
@@ -129,7 +136,67 @@
...
+
+            document.getElementById('box-dark-theme').addEventListener('change', function(evt) {
+              wrapper.classList[evt.target.checked ? 'add' : 'remove']('mdc-theme--dark');
+            });
+
+            document.getElementById('box-dense').addEventListener('change', function(evt) {
+              tfEl.classList[evt.target.checked ? 'add' : 'remove']('mdc-textfield--dense');
+              tf.ripple.layout();
+            });
+          }, 80);
```
traviskaufman: Yeah agreed that 0 is better; I didn't realize we could use 0. Hopefully sometime soon we'll have a more robust solution for loading these demo styles.

traviskaufman:
@amsheehan I can add the note about full width textfields as part of this PR for sure.

traviskaufman:
@yeelan0319 @amsheehan PTAL!

@yeelan0319 That bug looks like it's for the ripple. I will file an issue for it.

packages/mdc-textfield/mdc-textfield.scss
amsheehan: 👍

packages/mdc-textfield/mdc-textfield.scss
amsheehan: Curious. I had initially slowed it down but still couldn't see it animating out. I'm thinking the darker background of the text box background is the same color as the ink wash, which makes it appear as though there is no fade out. All the user can see is the foreground ripple (which doesn't animate out afaik)

Either way, if motion signed off, that's good enough for me. 👍

yeelan0319:
@traviskaufman SGTM. LGTMed this PR.

PR #812, opened 2017-06-10
fix(toolbar): Little changes in MDC Toolbar

### LINKED ISSUE #790: Little changes in MDC Toolbar
Hey,
In MDC Toolbar, the title has a left-margin of 16px. 

**The problems** with that are:

1. If you remove the left menu icon, the title will have a wrong padding on the left.
2. According to Material Design Guidelines, the margin shouldn't be `16px`. The margin from the left of the screen is 16, the width of the icon is 24, and the title is 72 away from the screen so we get 72 - 24 - 16 = `32px` which is the distance between the icon and the title.
image

**The solution** is to add a margin-right to the left menu icon of `32px`.

I am ready to make  a PR to fix this.

touficbatache:
Fixes #790

### DIFF
--- a/packages/mdc-toolbar/mdc-toolbar.scss
+++ b/packages/mdc-toolbar/mdc-toolbar.scss
@@ -79,7 +79,6 @@
 
     align-self: center;
     margin: 0;
-    margin-left: 16px;
     padding: $mdc-toobar-element-vertical-padding 0;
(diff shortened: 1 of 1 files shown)


touficbatache:
After updating MDC-Web, everything is correct and nothing needs change.

PR #820, opened 2017-06-13
Expose the rest of SASS variables

Kikobeats:


Garbee:
What is the justification for the rest of these variables being exposed?

Kikobeats:
Actually interested in secondary height: If it's possible create a big card, why not possible create increment the size of the secondary content to be aligned?

yeelan0319:
@Kikobeats, thanks for making this request!
MDC grid list is made to follow the requirements in [Material Guideline](https://material.io/guidelines/components/grid-lists.html#grid-lists-specs), so we are not going to make these variable as part of API.

As that being said, however, we will definitely reflect this request to our designer. If you currently have a specific design that doesn't fit what we provided in MDC, we strongly encourage you to override it.

PR #825, opened 2017-06-14
fix(toolbar): Toolbar for Desktop should be 24px per guidelines

### LINKED ISSUE #786: Toolbar for Desktop should be 24px per guidelines

0.11.1

Chrome

### What OS are you using?

MacOS Sierra 10.12.5 (16F73)

No step, just initial styles.

Padding left and right for DESKTOP should be 24px per Material Design Guidelines:
https://material.io/guidelines/layout/structure.html#structure-app-bar

Padding left and right is 28px.

touficbatache:
Hey,
This should fix #786. To explain a little what I did: added a left and right padding of `8px` because the icon takes already `16px` from `padding-left` or `padding-right` so **8 + 16 = 24 px**
@amsheehan

### DIFF
--- a/packages/mdc-toolbar/mdc-toolbar.scss
+++ b/packages/mdc-toolbar/mdc-toolbar.scss
@@ -41,6 +41,7 @@
     height: auto;
     min-height: $mdc-toolbar-row-height;
     box-sizing: border-box;
+    padding: 0 8px;
 
     // TODO: refactor this out when #23 is implemented
     @media (max-width: $mdc-toolbar-mobile-landscape-width-breakpoint)
@@ -51,6 +52,7 @@
     // TODO: refactor this out when #23 is implemented
     @media (max-width: $mdc-toolbar-mobile-breakpoint) {
       min-height: $mdc-toolbar-mobile-row-height;
+      padding: 0;
     }
   }
 


touficbatache:
Sure!

touficbatache:
The problem is that if the user doesn't have a menu icon with just a title, the padding won't be applied to the title...

yeelan0319:
Good catch, @touficbatache 

Then I think we need to rely on [sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_selectors) to see if the menu exists in that case:

```
.mdc-toolbar__icon--menu {
  padding: 16px 24px;
}
.mdc-toolbar__title {
  margin-left: 24px;
}
.mdc-toolbar__icon--menu + .mdc-toolbar__title {
  margin-left: 8px;
}

@media (max-width: 839px) {
  .mdc-toolbar__icon--menu {
    padding: 16px;
  }
  // put important here because media query will be override by more granular sibling rule 
  .mdc-toolbar__title {
    margin-left: 16px !important;
  }
}
``` 

WDYT?

touficbatache:
Seems good. What about the other icons? The ones on the right?

yeelan0319:
Oh yeah, definitely that should the case :) and you are right `mdc-toolbar__icon` should also be adjusted correspondingly, though they don't need to worry about title.

Let me know when you finishes the update, I will go over it again, thanks!

touficbatache:
@yeelan0319 I made the necessary changes but they are not showing here... https://github.com/touficbatache/material-components-web/blob/master/packages/mdc-toolbar/mdc-toolbar.scss

yeelan0319:
Interesting. Code loooks good at glance, but I am not even sure if Github is capturing the correct change since it shows it is from `unknown repository`...

@touficbatache Could you close this PR and then reopen a new one? Let's make sure we merge the correct thing into master 👍

touficbatache:
Yeah, no problem ;)

PR #826, opened 2017-06-14
docs(elevation): Simplify README to follow proposed best practices

lynnjepsen:


packages/mdc-elevation/README.md
```diff
@@ -50,124 +44,35 @@ npm install --save @material/elevation
...
-  &:active {
-    @include mdc-elevation(8);
-  }
-}
-```
-
-If you need more configurability over your transitions, you can easily accomplish this by using
-the `mdc-elevation-transition-rule` function in conjunctions with the exported sass variables that
-mdc-elevation exposes for purposes of transitioning.
+If you need more configurability over your transitions, use the `mdc-elevation-transition-rule` function in conjunctions with the exported sass variables.
```
amsheehan: remove the "s" from "conjunction"

PR #831, opened 2017-06-15
feat(infrastructure): Create script for that rewrites .scss imports

lynnjepsen:


scripts/sass-closure-rewriter.sh
```diff
@@ -0,0 +1,44 @@
...
+#  limitations under the License.
+#
+
+set -e
+
+function log() {
+  echo -e "\033[36m[closure-rewriter]\033[0m" "$@"
+}
+
+function join_by { local IFS="$1"; shift; echo "$*"; }
```
traviskaufman: Doesn't look like this is used anywhere?

scripts/sass-closure-rewriter.sh
```diff
@@ -0,0 +1,44 @@
...
+for pkg in $PACKAGE_NAMES ; do
+  if [[ $pkg != *"mdc-"* ]]; then
+    continue
+  fi
+  cp -r "packages/$pkg" $CLOSURE_PKGDIR
+done
+rm -fr $CLOSURE_PKGDIR/**/{node_modules,dist}
+
+log "Rewriting all import statements to be closure compatible"
+node scripts/rewrite-sass-import-statements-for-closure.js $CLOSURE_PKGDIR
```
traviskaufman: nit: newline at EOF

scripts/sass-closure-rewriter.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
traviskaufman: General comment here is that some people may not understand why we need to rewrite our _sass_ for closure. Should we maybe mention that this is to help make it closure-stylesheets compatible for internal infrastructure?

lbjones:
Can you explain how to use this script?

Looks like the files are written to the CLOSURE_TMP folder.  Do you import your files from there?

lynnjepsen:
Sorry for the long delay in a reply. sass-closure-rewriter.sh rewrites files to CLOSURE_TMP folder. From there you can copy them wherever you like. 

I wrote it to help us migrate our Sass files into Google's internal build system, which is based off directory structure and does not use node. So, the script might not be suited for any other use case. But, you can always use the script as an example, and write your own script to suit your needs.

Issue #819, opened 2017-06-13
Raised button on dark theme uses primary-on-dark text even though the background is primary, not dark


unascribed:
### What MDC-Web Version are you using?

0.13.0

### What browser(s) is this bug affecting?

Chromium 58.0.3029.81
> Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.81 Safari/537.36

Firefox Developer Edition 54.0b12
> Mozilla/5.0 (X11; Linux x86_64; rv:54.0) Gecko/20100101 Firefox/54.0

### What OS are you using?

Arch Linux x86_64 (last updated May 3rd)

### What are the steps to reproduce the bug?

1. Create a raised button
2. Add mdc-theme--dark to a parent element
3. Set the primary color to something bright, like Yellow 500
4. Observe the fact the button is completely illegible

https://codepen.io/anon/pen/EXyBGe

### What is the expected behavior?

The button text color should change to black, as if it was using --mdc-theme-text-primary-on-primary

### What is the actual behavior?

The button text color stays white, as it's using --mdc-theme-text-primary-on-dark

touficbatache:
That seems like a bug!

makotosalon:
Please unsubscribe.


2017/06/14 5:06、William Thompson <notifications@github.com> のメッセージ:

What MDC-Web Version are you using?

0.13.0

What browser(s) is this bug affecting?

Chromium 58.0.3029.81

Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.81 Safari/537.36

Firefox Developer Edition 54.0b12

Mozilla/5.0 (X11; Linux x86_64; rv:54.0) Gecko/20100101 Firefox/54.0

What OS are you using?

Arch Linux x86_64 (last updated May 3rd)

What are the steps to reproduce the bug?

Create a raised button
Add mdc-theme--dark to a parent element
Set the primary color to something bright, like Yellow 500
Observe the fact the button is completely illegible
https://codepen.io/anon/pen/EXyBGe

What is the expected behavior?

The button text color should change to black, as if it was using --mdc-theme-text-primary-on-primary

What is the actual behavior?

The button text color stays white, as it's using --mdc-theme-text-primary-on-dark

―

touficbatache:
[...]

PR #833, opened 2017-06-15
fix(button): Remove background-color from raised buttons on dark theme

unascribed:
See: #819

This runs contrary to the guidelines, but considering MDC gives full control of button color through class names (and CSS is very flexible), it's likely a non-issue. This also fixes the text color being wrong when a light primary color is used.

**Fair warning**: I tested this by deleting the directive with Chrome devtools. I'm 90% sure my change here has the desired effect, but I haven't yet had a chance to test a fresh compile.

### DIFF
--- a/packages/mdc-button/mdc-button.scss
+++ b/packages/mdc-button/mdc-button.scss
@@ -114,7 +114,6 @@
     }
 
     @include mdc-theme-dark(".mdc-button") {
-      @include mdc-theme-prop(background-color, primary);
 
       // postcss-bem-linter: ignore
       &::before {


unascribed:
Signed.

Garbee:
> This runs contrary to the guidelines,

We should get input from the MD team on this then. As, what is currently provided is exactly what is specified here. The default's shouldn't go against that.

unascribed:
I'd agree, but I was (somewhat indirectly) asked to PR this by @traviskaufman on #819.

Garbee:
I am not asking for a merge block. Simply a delay to get feedback from the design team on if this is appropriate. As what change is made by this PR is specifically against my entire understanding of the button component from when I made the original base for MCW.

If I were to comment on the issue with that, it could easily get lost/looked over. Best keep the discussions targeted on the issue/PR in question rather than spreading conversation out.

amsheehan:
I won’t be available until Monday. @yiranm or @lynnjepsen, could you facilitate this?

lynnjepsen:
I apologize for not replying sooner, but I've found some time to look into this issue. I've written a longer explanation on the [original issue](https://github.com/material-components/material-components-web/issues/819) and created a [new PR](https://github.com/material-components/material-components-web/pull/853) to solve the original issue.

amsheehan:
Closing, as @lynnjepsen seems to have confirmed what @Garbee and the spec says with design. Thanks again @unascribed, we really appreciate the effort, but in this case I think we're seeing a case that should be handled on an implementation level as regards `color` for `mdc-button`'s on a dark theme.

PR #835, opened 2017-06-15
fix(infrastructure): Harden closure declaration source rewriting

traviskaufman:
- Ensure that all JS src files are recursively picked up when sending
  `--js` flag to closure compiler
- Ensure `export ... from ...` statements are covered when rewriting
  declaration sources.


traviskaufman:
cc @bogusred

PR #842, opened 2017-06-20
feat(slider): Implement discrete slider and discrete slider with marker

### LINKED ISSUE #25: Implement Slider Component

yeelan0319:
Closes #25


amsheehan:
Looks like we need some more branch coverage. Once that passes I'll start the review. 👍

yeelan0319:
Test improved, please take a look :)

test/unit/mdc-slider/foundation.test.js
```diff
@@ -600,6 +724,25 @@ test('#setMin updates "aria-valuemin" to the new minimum', () => {
   raf.restore();
 });
 
+test('#setMin re-place track markers if slider is discrete and displays markers', () => {
```
amsheehan: "replace"

test/unit/mdc-slider/foundation.test.js
```diff
@@ -668,6 +827,25 @@ test('#setStep updates the slider\'s UI when altering the step value', () => {
   raf.restore();
 });
 
+test('#setStep re-place track markers if slider is discrete and displays markers', () => {
```
amsheehan: "replace"

test/unit/mdc-slider/mdc-slider.test.js
```diff
@@ -29,11 +29,15 @@ suite('MDCSlider');
 
 function getFixture() {
   return bel`
-    <div class="mdc-slider" tabindex="0" role="slider" aria-valuemin="0" aria-valuemax="100" aria-valuenow="0">
+    <div class="mdc-slider foo" tabindex="0" role="slider" aria-valuemin="0" aria-valuemax="100" aria-valuenow="0">
```
amsheehan: Add any extraneous class names in the individual test using `root.classList.add('foo')` when you're setting up the test. The fixture should be as generic as possible, and any test that needs `foo` present should exist in isolation.

test/unit/mdc-slider/mdc-slider.test.js
```diff
@@ -232,6 +236,13 @@ test('#stepDown decrements the slider by the step value if no value given and a
   assert.equal(component.value, 10);
 });
 
+test('adapter#hasClass checks if a class exists on root element', () => {
+  const {component} = setupTest();
+  const hasFoo = component.getDefaultFoundation().adapter_.hasClass('foo');
```
amsheehan: See comment above. Remove `foo` from the fixture, and refactor the test to looks something like:

```js

root.classList.add('foo');

assert.isTrue(component.getDefaultFoundation().adapter_.hasClass('foo');
```

demos/slider.html
```diff
@@ -109,56 +109,122 @@
...
       </section>
     </main>
     <script src="assets/material-components-web.js" charset="utf-8"></script>
     <script>
       (function() {
         setTimeout(function() {
           mdc.slider.MDCSlider.attachTo(document.getElementById('hero-slider'));
-          initDemo(document.getElementById('continuous-slider-example'));
+          initDemo(document.getElementById('slider-example'));
         }, 80);
```
yeelan0319: We need to keep this one as the style loader we use sometime will not be fully loaded when the page runs to the end of this script. It will unstably have some funny weird effect if they are not initialize correctly. We have similar setup in `toolbar` and `dialog`.

demos/slider.html
```diff
@@ -169,55 +235,90 @@
           var disabled = demoRoot.querySelector('[name="disabled"]');
           var useCustomColor = demoRoot.querySelector('[name="use-custom-color"]');
           var rtl = demoRoot.querySelector('[name="rtl"]');
-          var value = demoRoot.querySelector('.value');
-          var committedValue = demoRoot.querySelector('.committed-value');
-          var sliderEl = demoRoot.querySelector('.mdc-slider');
-          var slider = new mdc.slider.MDCSlider(sliderEl);
```
yeelan0319: Currently the demo is set up to have 3 type of slider configured by the same control panel, since the control panel is really long (has `max`, `min`, `rtl`, etc...) 
In this case, I do want to select a `demoRoot` that differentiate contents from hero part, but also since there are multiple sliders within that wrapper, I removed the `continuous` in the class and name the variables to identify which one is which.

Does that answer your question?

packages/mdc-slider/README.md
```diff
@@ -125,6 +144,31 @@ for the minimum and maximum values, which can always be set. This is to ensure c
 The step value can be any positive floating-point number, or `0`. When the step value is `0`, the
 slider is considered to not have any step.
 
+For discrete slider, it requires to have a step value other than `0`. So if a step is not provided,
```
yeelan0319: If you set a step to a negative value, it will throw you an error on your face :p
I will document that part~

packages/mdc-slider/index.js
```diff
@@ -116,6 +119,28 @@ export class MDCSlider extends MDCComponent {
       setTrackStyleProperty: (propertyName, value) => {
         this.track_.style.setProperty(propertyName, value);
       },
+      setMarkerValue: (value) => {
+        this.pinValueMarker_.innerText = value;
+      },
+      appendTrackMarkers: (numMarkers) => {
```
yeelan0319: The reason why I have it as an adapter method is because mdc-slider need to able to dynamically change the markers on track when users calls `setMax`, `setMin`, etc...

Also this cannot be done in `initialize` because `initialize` run before `foundation.init`, but we are actually doing this in `initialSyncWithDOM`.

packages/mdc-slider/index.js
```diff
@@ -116,6 +119,28 @@ export class MDCSlider extends MDCComponent {
...
+          frag.appendChild(marker);
+        }
+        this.trackMarkerContainer_.appendChild(frag);
+      },
+      removeTrackMarkers: () => {
+        while (this.trackMarkerContainer_.firstChild) {
+          this.trackMarkerContainer_.removeChild(this.trackMarkerContainer_.firstChild);
+        }
+      },
+      setLastTrackMarkersStyleProperty: (propertyName, value) => {
```
yeelan0319: I did quite get the question, I think `setLastTrackMarkersStyleProperty` is already an adapter method, do you mean you want it to be more generic so that it is not specific to last track marker?

packages/mdc-slider/mdc-slider.scss
```diff
@@ -77,6 +78,51 @@ $mdc-slider-dark-theme-grey: #5c5c5c;
     }
   }
 
+  &__track-marker-container {
+    display: flex;
+    flex-flow: row nowrap;
```
yeelan0319: Good catch!

packages/mdc-slider/mdc-slider.scss
```diff
@@ -109,6 +155,37 @@ $mdc-slider-dark-theme-grey: #5c5c5c;
...
+    top: 0;
+    left: 0;
+    align-items: center;
+    justify-content: center;
+    width: 26px;
+    height: 26px;
+    margin-top: -2px;
+    margin-left: -2px;
+    transform: rotate(-45deg) scale(0) translate(0, 0);
+    transition: transform 100ms ease-out;
```
yeelan0319: Good question, I don't think it is specified on spec. I will ask our motion design for some suggestion!

test/unit/mdc-slider/foundation.test.js
```diff
@@ -135,6 +151,94 @@ test('#destroy deregisters all component event handlers registered during init()
...
+  td.when(mockAdapter.computeBoundingRect()).thenReturn({left: 0, width: 100});
+  td.when(mockAdapter.hasClass(cssClasses.IS_DISCRETE)).thenReturn(true);
+  td.when(mockAdapter.hasClass(cssClasses.HAS_TRACK_MARKER)).thenReturn(false);
+  foundation.init();
+  raf.flush();
+
+  foundation.setMax(100);
+  foundation.setMin(0);
+  foundation.setStep(10);
+  foundation.setupTrackMarker();
```
yeelan0319: Yes sir, 4 tests below.

`#setupTrackMarker does execute if discrete slider does not display markers`

test/unit/mdc-slider/foundation.test.js
yeelan0319: Is redo placement equivalent to replace? May be I should totally change how I say it... Any suggestions?

test/unit/mdc-slider/foundation.test.js
yeelan0319: Same above.

test/unit/mdc-slider/mdc-slider.test.js
yeelan0319: Great suggestion!

packages/mdc-slider/index.js
amsheehan: Yes. This could be more generic.

test/unit/mdc-slider/foundation.test.js
amsheehan: Maybe re-renders?

test/unit/mdc-slider/foundation.test.js
amsheehan: hahaha far from perfect!

packages/mdc-slider/index.js
yeelan0319: IMHO, the rule of thumb for mdc is to make the adapter as less ambiguous as possible, which it can almost read like a human sentence. The other components I feel have been taking this approach like `setStyleForTilesElement`, `setStyleForFlexibleRowElement`, `setStyleForTitleElement`. 

However, that being said, I totally can see value in this approach when we refactor to have global adapter. I think we should make changes that breaks our convention as a more holistic decision. WDYT?

PR #846, opened 2017-06-22
fix(toolbar): Adjusting sibling elements on mobile landscape

mikegleasonjr:
Without the fix there is a gap when hitting `@media (max-width: 959px) and (orientation: landscape)`:

2017-06-22 at 12 51 am

Because the row height is changed on that breakpoint: 
https://github.com/material-components/material-components-web/blob/master/packages/mdc-toolbar/mdc-toolbar.scss#L48


yeelan0319:
Thanks @MichaelTamm for fixing the bug :)

mikegleasonjr:
My pleasure. Will try to contribute more since it is pretty easy to do with you guys.

PR #852, opened 2017-06-22
feat(snackbar): Implement full-featured Snackbar component

### LINKED ISSUE #28: Implement Snackbar Component
Spec Reference: https://www.google.com/design/spec/components/snackbars-toasts.html 

It's been called a "Toast" in the past, but the thing that's not Android-specific is a **snackbar**.
- [x] Initial Implementation
- [x] Dark Theme (background: `#fafafa`, text: text-primary-on-dark, action: theme primary color)
- [x] Keyboard Focus Handling
- [x] In demo, ensure that "Action on Bottom" is unchecked and disabled when "multiline" is unchecked
- [x] End-to-end usage example in README (instantiation, using it in response to an event)
- [x] _Start_-aligned snackbars (left for LTR, right for RTL)
- [x] Add `dismissOnAction` option, which defaults to `true`. When `true`, snackbar dismisses as soon as a user interacts with its action button (see #28).
- [x] Address #564 

> Copied from https://github.com/google/material-design-lite/issues/4490

amsheehan:
resolves #28


packages/mdc-snackbar/README.md
```diff
@@ -125,6 +142,38 @@ properties and their usage.
 | multiline | Whether to show the snackbar with space for multiple lines of text | Optional |  Boolean |
 | actionOnBottom | Whether to show the action below the multiple lines of text | Optional, applies when multiline is true |  Boolean |
 
+### Responding to a Snackbar Action
+
+To respond to a snackbar action, assign a function to the optional `actionHandler` property in the object that gets passed to the `show` method. If you choose to set this property, you *must _also_* set the `actionText` property.
+
```
yeelan0319: nit: Could we also document it will throw an error in that case?

packages/mdc-snackbar/README.md
```diff
@@ -149,10 +198,18 @@ The adapter for snackbars must provide the following functions, with correct sig
 | `removeClass(className: string) => void` | Removes a class from the root element. |
 | `setAriaHidden() => void` | Sets `aria-hidden="true"` on the root element. |
 | `unsetAriaHidden() => void` | Removes the `aria-hidden` attribute from the root element. |
-| `setMessageText(message: string) => void` | Set the text content of the message element. |
-| `setActionText(actionText: string) => void` | Set the text content of the action element. |
 | `setActionAriaHidden() => void` | Sets `aria-hidden="true"` on the action element. |
 | `unsetActionAriaHidden() => void` | Removes the `aria-hidden` attribute from the action element. |
+| `setActionText(actionText: string) => void` | Set the text content of the action element. |
```
yeelan0319: super nit: Set => Sets

packages/mdc-snackbar/README.md
```diff
@@ -149,10 +198,18 @@ The adapter for snackbars must provide the following functions, with correct sig
...
+| `setMessageText(message: string) => void` | Set the text content of the message element. |
```
yeelan0319: same thing here :)

packages/mdc-snackbar/README.md
```diff
@@ -149,10 +198,18 @@ The adapter for snackbars must provide the following functions, with correct sig
...
+| `setFocus() => void` | Sets focus on the action button. |
+| `visibilityIsHidden() => boolean` | Returns document.hidden property. |
+| `registerBlurHandler(handler: EventListener) => void` | Registers an event handler to be called when a `blur` event is triggered on the action button |
+| `deregisterBlurHandler(handler: EventListener) => void` | Deregisters a `blur` event handler from the actionButton |
+| `registerVisibilityChangeHandler(handler: EventListener) => void` | Registers an event handler to be called when a 'visibilitychange' event occurs |
+| `deregisterVisibilityChangeHandler(handler: EventListener) => void` | Deregisters an event handler to be called when a 'visibilitychange' event occurs |
+| `registerCapturedInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event handler to be called when the given event type is triggered on the `body` |
```
amsheehan: I don't think so. What this name suggests is that we are doing something with an interaction that has previously been captured (albeit immediately after).

I agree we should be acknowledging that we are listening during the capture phase. :+1:

packages/mdc-snackbar/index.js
```diff
@@ -47,6 +47,16 @@ export class MDCSnackbar extends MDCComponent {
       unsetActionAriaHidden: () => getActionButton().removeAttribute('aria-hidden'),
       setActionText: (text) => { getActionButton().textContent = text; },
       setMessageText: (text) => { getText().textContent = text; },
+      setFocus: () => getActionButton().focus(),
+      visibilityIsHidden: () => document.hidden,
+      registerBlurHandler: (handler) => getActionButton().addEventListener('blur', handler, true),
```
amsheehan: Again, I don't think so. What matters is that we're reacting to a `blur` event. I'll document that it happens during the capture phase

test/unit/mdc-snackbar/foundation.test.js
```diff
@@ -74,6 +77,30 @@ test('#destroy calls adapter.deregisterActionClickHandler() with a registerActio
   td.verify(mockAdapter.deregisterActionClickHandler(changeHandler));
 });
 
+test('#destroy calls adapter.deregisterVisibilityChangeHandler() with a function', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const {isA} = td.matchers;
+
+  foundation.destroy();
+  td.verify(mockAdapter.deregisterVisibilityChangeHandler(isA(Function)));
```
amsheehan: Whoooooops! :smile:

amsheehan: Oh, no I disagree. We already do that when we test the adapter method.

test/unit/mdc-snackbar/foundation.test.js
```diff
@@ -74,6 +77,30 @@ test('#destroy calls adapter.deregisterActionClickHandler() with a registerActio
...
+});
+
+test('#destroy calls adapter.deregisterBlurHandler() with a function', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const {isA} = td.matchers;
+
+  foundation.destroy();
+  td.verify(mockAdapter.deregisterBlurHandler(isA(Function)));
```
amsheehan: Same as above :)

test/unit/mdc-snackbar/foundation.test.js
```diff
@@ -74,6 +77,30 @@ test('#destroy calls adapter.deregisterActionClickHandler() with a registerActio
...
+});
+
+test('#destroy calls adapter.deregisterCapturedInteractionHandler() with an event type and function 3 times', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const {isA} = td.matchers;
+
+  foundation.destroy();
+  td.verify(mockAdapter.deregisterCapturedInteractionHandler(isA(String), isA(Function)), {times: 3});
```
amsheehan: Agreed with specify which event its deregistering.

packages/mdc-snackbar/index.js
yeelan0319: Document it definitely helps, but IMHO give hint in naming also helps. Imagine if I am the person implementing those adapters in another framework (instead of using it directly provided by our vanilla adapter), I probably won't know I need to have a capturing eventListener just by looking at this name and confused why the component is not behavior correctly.

test/unit/mdc-snackbar/foundation.test.js
yeelan0319: However, this is `destroy()` but `init()` 😃 . I can see totally why we won't test it in `init()`

packages/mdc-snackbar/index.js
amsheehan: Okay. That makes sense. I'd rather have it be explicitly clear, rather than rely on the signature for `addEventListener`

test/unit/mdc-snackbar/foundation.test.js
amsheehan: Deregistering the event handlers is not the concern of `destroy()` though. It is the concern of `adapter.deregisterVisibilityChangeHandler()`. Therefore, we should test that destroy calls that, and then in the test for `deregisterVisibilityChangeHandler()`, test that the event handlers are removed properly.

touficbatache:
What's the state of this PR? Should it be merged?

piotr-cz:
@amsheehan 
What's the reason for listening for `focus` event on `document.body` and not on action button?

IMO this is a bug which may cause snackbar to lock itself in opened state (for which I've prepared a fix in #2183) but I may not fully understand the intentions or implementation as there are no comments in code

PR #862, opened 2017-06-26
fix(grid): use correct selector for size specific column span

sivaraam:


sivaraam:
:+1:

sivaraam:
@yeelan0319 You're welcome!

PR #866, opened 2017-06-26
feat(auto-init): Annotate mdc-auto-init for closure

### LINKED ISSUE #333: Annotate mdc-auto-init for closure

nckh:
Fixes #333 

My guess is we are going to need to refer to `attachTo()` with array syntax (`const component = Ctor['attachTo'](node)`) to work around obfuscation. The annoying thing is that each child class of MDCComponent,  similarly to what `goog.exportProperty()` does to expose obfuscated members, would have to define something like this:

MDCSimpleMenu['attachTo'] = MDCSimpleMenu.attachTo;

If `attachTo()` was not a static method we could have used an interface.

Any thoughts? Did you have another way in mind originally?
Cheers!

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,7 @@
   },
   "closureWhitelist": [
     "mdc-animation",
+    "mdc-auto-init",
     "mdc-base",
     "mdc-menu"
   ]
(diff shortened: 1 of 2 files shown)


lynnjepsen:
I'm actually not even sure we need to annotate mdc-auto-init with closure.

> If attachTo() was not a static method we could have used an interface.

And this needs further discussion. I'm going to close this PR now, and @nckh can open an if they really want mdc-auto-init annotated with closure.

PR #867, opened 2017-06-27
feat(checkbox): Annotate mdc-checkbox for closure

### LINKED ISSUE #334: Annotate mdc-checkbox for closure

nckh:
Fixes #334

### DIFF
--- a/packages/mdc-checkbox/constants.js
+++ b/packages/mdc-checkbox/constants.js
@@ -19,7 +19,6 @@ const ROOT = 'mdc-checkbox';
 
 /** @enum {string} */
 export const cssClasses = {
-  ROOT: 'mdc-checkbox',
   UPGRADED: 'mdc-checkbox--upgraded',
   CHECKED: 'mdc-checkbox--checked',
   INDETERMINATE: 'mdc-checkbox--indeterminate',
--- a/packages/mdc-checkbox/constants.js
+++ b/packages/mdc-checkbox/constants.js
@@ -14,8 +14,10 @@
  * limitations under the License.
  */
 
+/** @const {string} */
 const ROOT = 'mdc-checkbox';
 
+/** @enum {string} */
 export const cssClasses = {
   ROOT: 'mdc-checkbox',
   UPGRADED: 'mdc-checkbox--upgraded',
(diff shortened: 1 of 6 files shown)


packages/mdc-ripple/index.js
```diff
@@ -114,3 +114,12 @@ export class MDCRipple extends MDCComponent {
...
 }
+
+/**
+ * @typedef {!{
+ *   root_: !Element,
+ *   unbounded: (boolean|undefined),
+ *   disabled: (boolean|undefined)
+ * }}
+ */
+let RippleAbleType;
```
lynnjepsen: I talked to our MD team a bit, and I think "RippleCapableSurface" is more clear. Can you add a link to our guidelines on when to use ripples:
https://material.io/guidelines/motion/choreography.html#choreography-radial-reaction

Can you also add a note about what "unbounded" and "disabled" means from a ripple's point of view.

packages/mdc-checkbox/constants.js
```diff
@@ -14,8 +14,10 @@
  * limitations under the License.
  */
 
+/** @const {string} */
 const ROOT = 'mdc-checkbox';
```
nckh: Indeed it's not. Removed!

packages/mdc-checkbox/foundation.js
```diff
@@ -220,6 +257,10 @@ export default class MDCCheckboxFoundation extends MDCFoundation {
     }
   }
 
+  /**
+   * @return {!HTMLInputElement|InputPropsType}
```
nckh: Removed `HTMLInputElement` and moved `InputPropsType` def to the adapter so that the result of `querySelector()` can be casted to this type in the component class.

Returning that fake element seems to be required by some unit tests verifying the getters still work when no native control is returned.

packages/mdc-ripple/index.js
nckh: Done. I changed from `@typedef` to `@record` so that those comments can be inlined.

nckh:
Thanks for the review Lynn. 

I've also used the `EventListener` type every where callbacks were typed to `Function` to align with the original annotations in `MDCCheckboxFoundation.defaultAdapter`.

lynnjepsen:
Sorry for the long delay, I was on vacation.

This generally looks good, so I'm going to update the branch and make sure all tests still pass. If they do, I will merge!

Issue #225, opened 2017-01-29
Temporary drawer stops propagation of all child elements' click events.


Hardtack:
## Bugs

Follow the template below to ensure the quickest and most accurate response to your issue.

### What MDC-Web Version are you using?

0.1.1

### What browser(s) is this bug affecting?

Safari, Chrome

### What OS are you using?

MacOS

### What are the steps to reproduce the bug?

1. Add temporary drawer using foundation.

2. Add `click` event to child element of drawer (`mdc-temporary-drawer__drawer`).

3. Child element's `click` event handler doesn't fire because of interaction handler added by `MDCTemporaryDrawerFoundation`
   https://github.com/material-components/material-components-web/blob/7fa157cb4f4c801283e799c10097af55d21cd341/packages/mdc-drawer/temporary/foundation.js#L102

### What is the expected behavior?

It should fire click event of elements in drawer.

### What is the actual behavior?

It doesn't. because of `evt.stopPropagation();`

### Any other information you believe would be useful?

I'm writing [react wrapper of material-components-web](https://github.com/Hardtack/react-mdcw) and I'm writing example with drawer.

Since I'm using hash-history with react-router. Page can be routed without re-rendering drawer. So we should close drawer manually when menu in drawer is clicked like this.

``` javascript
<Drawer>
  ...
  <Content>
    <Link to="/button" onClick={() => this.setState(drawerOpen: false)}>
      Button Examples
    </Link>
  </Content>
</Drawer>
```

Since `evt.stopPropagation()` is applied in `mdc-temporary-drawer__drawer`, `onClick={...}` will never be fired.

traviskaufman:
Are you sure that `evt.stopPropagation` is the reason why the click handler doesn't fire?

We call `stopPropgatation` within the bubbling phase of the event, so unless the `<Link>` component is a parent of the `mdc-temporary-drawer__drawer` element, it should activate that event handler before the propagation is stopped.

Could you link us to a codepen / jsbin / plunker with a reproduction of the issue? Feel free to use your implementation in the repro.

Hardtack:
Oh I misunderstood mechanisms of stopPropagation
I thought that stopPropagation is the reason because it works when I remove interaction handler from temporary-drawer__drawer
I will check again and give you some code that reproduce the issue

Hardtack:
It seems that react is not working well with native events. See this [code](http://codepen.io/Hardtack/pen/rjJwyL)
[...]

PR #869, opened 2017-06-27
fix(dialog): allow click events to propagate

### LINKED ISSUE #794: mdc-dialog with list items ; how to listen for clicks properly?
I'd like to use the dialog with list items. Of course the purpose is for a user to select an item. See example on official site "choose a ringtone": https://material-components-web.appspot.com/dialog.html

However, I noticed that it's not possible by default for my list items to listen for clicks. It seems that the framework attaches an event handler on the dialog and stops propagation, not letting me any chance to handle the selection of an item.

```
$(document).ready(function() {
  const MDCDialog = mdc.dialog.MDCDialog;
  const dialog = new MDCDialog(document.getElementById('mdc-dialog-with-list'));
  dialog.show();
  
  $(document).on('click', '.mdc-list-item', function() {
    alert('hello');
  });
});
```
https://jsfiddle.net/ftmn974x/

I guess I need to write a new adapter for the foundation to let a click happen in a dialog? But that seems way too heavy.

What I don't get as well is that in the official example and guideline this kind of list is clearly a list of items that you can select. However,I have to implement the css when hovering an item myself as well as the event handlers (and as mentioned above maybe even rewrite an adapter for this simple issue?).

Thank you for any advice

jonathanzong:
This PR fixes #794. I implemented the short fix proposed by @amsheehan in that thread and confirmed that listeners on the list items receive click events. cc @jverkoey


amsheehan:
Hey @jonathanzong, Thanks for the PR!

It looks like some of the tests are failing. As soon as that gets fixed up I'll review. From the output in Travis CI, it looks like there are two tests failing:

`foundation.test.js:199`
&
`foundation.test.js:256`

The first one we can remove since we're scrapping `evt.stopPropagation()`. The other will just have to be modified a little to accommodate the new (less greedy) event logic. If you could also fix the spelling mistake in the description of the second test it would be greatly appreciated!

`on click closese the dialog and notifies cancellation` should read: `on click _closes_ the dialog and notifies cancellation`

Thanks again

jonathanzong:
Upon further inspection I'm realizing that the intent of the removed test is to make sure click events don't propagate to items behind / unrelated to the dialog (see your comment [here](https://github.com/material-components/material-components-web/issues/225#issuecomment-296836508) in #225). My solution is to stop propagation when the `evt.currentTarget` (as opposed to `evt.target`) reaches the dialog root. I think this is the approach that lets events bubble, but not too far.

However, I'm not sure how to write a test for this behavior (how to mock events both inside and outside the dialog root). Do you have any advice or resources that might point me in the right direction?

jonathanzong:
Hi @amsheehan, I just updated the tests and the CI checks have passed, would appreciate your feedback!

Garbee:
> My solution is to stop propagation when the `evt.currentTarget` (as opposed to `evt.target`) reaches the dialog root. I think this is the approach that lets events bubble, but not too far.

This isn't emulating the natural dialog element though. Natural HTML `<dialog>`'s allow for the event to propagate all the way to the document itself. The component should reflect that functionality to align with how the web works.

packages/mdc-dialog/foundation.js
```diff
@@ -52,7 +52,14 @@ export default class MDCDialogFoundation extends MDCFoundation {
   constructor(adapter) {
     super(Object.assign(MDCDialogFoundation.defaultAdapter, adapter));
     this.isOpen_ = false;
-    this.componentClickHandler_ = () => this.cancel(true);
+    this.componentClickHandler_ = (evt) => {
+      if (this.adapter_.eventTargetHasClass(evt.currentTarget, cssClasses.ROOT)) {
```
amsheehan: As Garbee mentioned, events should be allowed to propagate down to the document.

test/unit/mdc-dialog/foundation.test.js
```diff
@@ -185,14 +185,17 @@ test('#cancel calls notifyCancel when shouldNotify is set to true', () => {
   td.verify(mockAdapter.notifyCancel());
 });
 
-test('on dialog surface click calls evt.stopPropagation() to prevent click from propagating to background el', () => {
+test('on dialog surface click calls evt.stopPropagation() to prevent click from propagating past the root', () => {
```
amsheehan: Once the event is allowed to reach the document, this test should be removed.

test/unit/mdc-dialog/foundation.test.js
```diff
@@ -246,17 +249,40 @@ test('on dialog surface click does not close or notify if the event target is no
...
 
   foundation.open();
-  handlers.click();
+  handlers.click(evt);
 
   td.verify(mockAdapter.removeClass(cssClasses.OPEN));
   td.verify(mockAdapter.notifyCancel());
 });
 
+test('on click does not close or notify cancellation if event target is not the backdrop', () => {
```
amsheehan: nit: might be worth mentioning for future devs that this effectively means a click on the surface does not close the dialog

test/unit/mdc-dialog/foundation.test.js
```diff
@@ -246,17 +249,40 @@ test('on dialog surface click does not close or notify if the event target is no
...
+    target: {},
+  };
+
+  td.when(mockAdapter.eventTargetHasClass(evt.target, cssClasses.BACKDROP)).thenReturn(false);
+
+  foundation.open();
+  handlers.click(evt);
+
+  td.verify(mockAdapter.removeClass(cssClasses.OPEN), {times: 0});
+  td.verify(mockAdapter.notifyCancel(), {times: 0});
```
amsheehan: It will be enough to test that the `removeClass()` adapter method isn't called.

jonathanzong:
Thanks @amsheehan, yes I did sanity check this on Chrome/Safari/Firefox and events do propagate.

PR #900, opened 2017-07-04
fix(select): Selection handler parameter

xxRockOnxx:
Should not destructure.

### DIFF
--- a/packages/mdc-select/foundation.js
+++ b/packages/mdc-select/foundation.js
@@ -83,7 +83,7 @@ export default class MDCSelectFoundation extends MDCFoundation {
       }
     };
     this.displayViaKeyboardHandler_ = (evt) => this.handleDisplayViaKeyboard_(evt);
-    this.selectionHandler_ = ({detail}) => {
+    this.selectionHandler_ = (detail) => {
       const {index} = detail;
       this.close_();
       if (index !== this.selectedIndex_) {


xxRockOnxx:
I signed it!

yeelan0319:
Hi @xxRockOnxx ,

I believe the deconstruct is needed. In mdc-simple-menu [README.md](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/README.md#handling-selection-events), the event data is wrapped inside `detail`. The implementation of the wrapping is in mdc-base component.js's `emit` method.

I am going to close this one, but feel free to leave further question if you have.

xxRockOnxx:
Pretty sure there's no `detail`.

See https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/simple/index.js#L119

It only emit the `index` and `item` directly NOT

````
{
  detail: {
    index: //
   item: //
  }
}

xxRockOnxx:
I rechecked up to base. Yes, you are right. It uses `emit` that emits `Custom Event` that has details. Maybe I'll just copy that as well instead.

yeelan0319:
You are right the argument passed in to `this.emit` is without not wrapped by `detail`, but that is not the evt object being passed to the handler yet. It got wrapped later in the `mdc-base` defined method. 

screen shot 2017-07-07 at 16 05 26

PR #908, opened 2017-07-06
fix(textfield): Full-width textfields not displaying accurate font

### LINKED ISSUE #903: Textarea not displaying accurate placeholder font on Android
I have a textarea with the following markup:

```
<div class="mdc-textfield mdc-textfield--fullwidth mdc-textfield--upgraded">
      <input class="mdc-textfield__input" name="title" id="title" type="text" aria-label="Title" placeholder="Title">
    </div>

<div class="mdc-textfield mdc-textfield--multiline mdc-textfield--fullwidth mdc-textfield--upgraded" id="descriptionContainer">
      <textarea class="mdc-textfield__input" id="description" name="description" rows="8" cols="40" aria-label="Description" placeholder="Description"></textarea>
    </div>
```

It looks right everywhere but for some reason is a little off on Android. I can't figure it out as I've looked at the markup / css about 30x now. The fact it works right on Desktop Chrome and iOS is the most baffling. Bug?

##### Android

browser-chrometabbedactivity-07042017162649

##### Desktop

screen shot 2017-07-04 at 4 25 45 pm

touficbatache:
Fixes and closes #903 @amsheehan


packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -48,6 +48,7 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
 
   &__input {
     @include mdc-theme-prop(color, text-primary-on-light);
+    @include mdc-typography-base;
```
amsheehan: This should work. What I'm thinking though is we only include `mdc-typography-base` on this selector pattern and not in `.mdc-textfield` as well.

So the work here would be removing the following from `.mdc-textfield`:
```css
@include mdc-typography-base;
@each $prop in (font-size, letter-spacing) {
  #{prop}: map-get(map-get($mdc-typography-styles, subheading2), $prop);
}
```

and putting it in `&__input`'s block instead

amsheehan: This way we will reduce our CSS footprint once compiled

touficbatache: Okay! Done. Would you mind taking a look at #922?

PR #909, opened 2017-07-06
fix(list): List end detail has a wrong padding

### LINKED ISSUE #904: List end detail has a wrong padding

> 0.14.0

> All browsers

### What OS are you using?

> All OS

> 1. Go to http://material-components-web.appspot.com/list.html
> 2. Scroll down to the **End Detail** section
> 3. Observe the end detail item's location

> The end detail item should be 16px away from the end of the list

> The end detail item should be 32px away from the end of the list

> This could easily be fixed by removing the `margin-right: 16px` from the `mdc-list-item__end-detail` class as the `mdc-list` class already has a padding of 16px on the right. Would you like a PR?

touficbatache:
Fixes and closes #904 @amsheehan

### DIFF
--- a/packages/mdc-list/mdc-list.scss
+++ b/packages/mdc-list/mdc-list.scss
@@ -78,7 +78,7 @@ $mdc-list-side-padding: 16px;
 
   &__end-detail {
     @include mdc-list-detail-size_(24px);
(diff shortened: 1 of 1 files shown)


touficbatache:
No problem. Happy to help 😃 @yeelan0319

PR #917, opened 2017-07-07
docs(rtl): Simplify README to follow best practices

lynnjepsen:


packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,20 @@ npm install --save @material/rtl
...
-  left: 0;
 
-  @include mdc-rtl {
-    left: auto;
-    right: 0;
-  }
+| Mixin | Description |
+| ----------------------------------------------- | - |
+| `mdc-rtl($root-selector)` | Creates a rule that is applied when the root element is within an RTL context |
+| `mdc-rtl-reflexive-box($base-property, $default-direction, $value, $root-selector)` | Applies the value to the `#{$base-property}-#{$default-direction}` property  in a LTR context, and flips the direction in an RTL context. This mixin zeros out the original value in an RTL context. |
```
yeelan0319: find a typo in comments here in `mdc-rtl/mixin.scss` line 128:
```
* You can also pass a 4th optional $root-selector argument which will be forwarded to `mdc-rtl`,
* e.g. `@include mdc-rtl-reflexive-box-property(margin, left, 8px, ".mdc-component")`.
```
It should be `@include mdc-rtl-reflexive-box`, we don't have a mixin called `mdc-rtl-reflexive-box-property`

yeelan0319: I suggest we emphasize the size effect this mixin have, it *zeros out the counter-party value*, which might surprise our user in some case.

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,20 @@ npm install --save @material/rtl
...
+| `mdc-rtl-reflexive-property($base-property, $left-value, $right-value, $root-selector)` | Emits rules that assign `#{$base-property}`-left to `#{left-value}` and `#{base-property}`-right to `#{right-value}` in a LTR context, and vice versa in a RTL context. |
```
yeelan0319: Put a line when we should use `mdc-rtl-reflexive-property` instead of `mdc-rtl-reflexive-box` by a quick comparison between its difference?

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,20 @@ npm install --save @material/rtl
...
+| `mdc-rtl-reflexive-position($pos, $value, $root-selector)` | Applies the value to the specified position in a LTR context, and flips the direction in an RTL context. |
```
yeelan0319: IMHO, having a short example will help me understand the behavior of the mixin classes better than have only text. If you feel it is too long, I suggest we change the verbose example to a simpler one.

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,20 @@ npm install --save @material/rtl
 
 ## Usage
 
-Simply `@import "@material/rtl/mixins";` and start using the mixins. Each mixin is described below.
+### Sass Mixins
 
-### mdc-rtl
+`$base-property` is a base box-model property. e.g. margin / border / padding.
```
yeelan0319: Listing the args out here looks a bit unnatural to me. Can we put it as part of the following table and use a more explanatory name, such as `position-property`?

packages/mdc-rtl/README.md
lynnjepsen: I want developers to read about what each mixin accomplishes. This lets developers decide where they can use the mixins in their own code. Once a developer is using a mixin, we can upgrade the implementation of the mixin **so long as it still accomplishes the same original goal**. 

When we document code examples of the actual CSS output of the mixin, we risk developers using a mixin simply because they want that exact output...NOT because they want to accomplish the same goal the mixin aims to solve.

So we can document examples of _where_ to use this mixins, but we should avoid documenting the actual CSS output. The CSS output is an implementation detail.

...I looked at our code and I couldn't actually find great examples of us using these properties...Theres one in checkbox, but it's more confusing to me than helpful.

lynnjepsen:
Addressed comments, PTAL

packages/mdc-rtl/README.md
yeelan0319: Pointed well taken, I agree that we shouldn't include transpiled css here. However, the single line document confuse me when to use which mixin class. Could we consider having a paragraph like:

In case for symmetric RTL properties, you can use `mdc-rtl-reflexive-box`. For example, when button have `margin-left: 8px; margin-right: 0px;` to left edge and `margin-left: 0px; margin-right: 8px` to right edge in RTL mode.
```
@include mdc-rtl-reflexive-box('margin', 'left', 8px, '.mdc-foo')
```

packages/mdc-rtl/README.md
yeelan0319: Try to emphasize this point :) WDYT?

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,20 @@ npm install --save @material/rtl
...
+`$pos` is a horizontal position property, either "left" or "right".
```
yeelan0319: We should also update here

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,20 @@ npm install --save @material/rtl
...
-  @include mdc-rtl {
-    left: auto;
-    right: 0;
-  }
+| Mixin | Description |
+| ----------------------------------------------- | - |
+| `mdc-rtl($root-selector)` | Creates a rule that is applied when the root element is within an RTL context |
+| `mdc-rtl-reflexive-box($base-property, $default-direction, $value, $root-selector)` | Applies the value to the `#{$base-property}-#{$default-direction}` property  in a LTR context, and flips the direction in an RTL context. This mixin zeros out the original value in an RTL context. |
+| `mdc-rtl-reflexive-property($base-property, $left-value, $right-value, $root-selector)` | Emits rules that assign `#{$base-property}`-left to `#{left-value}` and `#{base-property}`-right to `#{right-value}` in a LTR context, and vice versa in a RTL context. |
+| `mdc-rtl-reflexive-position($pos, $value, $root-selector)` | Applies the value to the specified position in a LTR context, and flips the direction in an RTL context. |
```
yeelan0319: and here.

lynnjepsen:
Added another paragraph about how each RTL mixin kind of "builds" on the previous one.

packages/mdc-rtl/README.md
lynnjepsen: Sorry...but I think maybe you're look at an old commit. I changed this in the latest commit.

Or maybe I screwed something up in my commits...

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,22 @@ npm install --save @material/rtl
...
-
-  @include mdc-rtl {
-    padding-left: auto;
-    padding-right: 4px;
-  }
-}
-```
+| Mixin | Description |
+| ----------------------------------------------- | - |
+| `mdc-rtl($root-selector)` | Creates a rule that is applied when the root element is within an RTL context |
```
yeelan0319: nit: Create rules applied within `$root-selector` scope when the root element is contained within an RTL context

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,22 @@ npm install --save @material/rtl
 
 ## Usage
 
-Simply `@import "@material/rtl/mixins";` and start using the mixins. Each mixin is described below.
+### Sass Mixins
 
-### mdc-rtl
+`mdc-rtl` is the most flexible mixin, because it can work with multiple CSS properties. All other RTL mixins logic could be engineered by only using `mdc-rtl`, but we provide these mixins for convenience.
```
yeelan0319: nit and opnionated: 
> but we recommend using a more specific mixin below for better code readability

packages/mdc-rtl/README.md
```diff
@@ -28,76 +26,22 @@ npm install --save @material/rtl
...
-.mdc-foo {
-  position: absolute;
-  left: 0;
+Both `mdc-rtl-reflexive-property` and `mdc-rtl-reflexive-box` work with one base box-model property, e.g. margin, border, padding. But `mdc-rtl-reflexive-property` is more flexible because it accepts different left and right values. `mdc-rtl-reflexive-box` assumes the left and right values are the same, and therefore that the box-model is symmetrical.
 
-  @include mdc-rtl {
-    left: auto;
-    right: 0;
-  }
+`mdc-rtl-reflexive-position` is the least flexible mixin. It only works with one horizontal position property, "left" or "right". It also assumes the left and right values are the same.
```
yeelan0319: less flexible -> most specific mixin about position?

Issue #755, opened 2017-05-31
Issue with customizing colors


mdecaffmeyer:
By default, the color used are: 
```
$mdc-theme-primary: #3f51b5 !default; /* Indigo 500 */
$mdc-theme-accent: #ff4081 !default; /* Pink A200 */
$mdc-theme-background: #fff !default; /* White */
```

By looking into developer tools, I see that the text on the primary is color is white:
`:root { --mdc-theme-text-primary-on-primary: white; }`

However when I change the primary color to cyan:
```
$mdc-theme-primary: #00BCD4;
@import "material-components-web/material-components-web";
```

Suddenly the text on primary color becomes black:
`:root { --mdc-theme-text-primary-on-primary: rgba(0, 0, 0, 0.87); }`

Why did this property change as I didn't change it manually?

Garbee:
The text color is automatically selected for the best contrast. This is done not only because the specification requires it but simply for accessibility for users to be most capable to read the content.

The calculations for the contrast are handled in the [theme functions](https://github.com/material-components/material-components-web/blob/master/packages/mdc-theme/_functions.scss) sass file.

mdecaffmeyer:
Indeed when I choose Cyan in the color tool, I have black text by default on the cyan background:

https://material.io/color/#!/?view.left=0&view.right=0&primary.color=00BCD4

However I've seen on another palette that it shows white text on cyan bg:

https://www.materialpalette.com/cyan/red

I also think that the latter is much more beautiful (white text on cyan), but then I violate material's recommendations, which leaves me pretty confused... ;-(

What do you think?

Garbee:
Material.io is the official resource, so I'd err towards that. MaterialPalette is a completely un-related community site doing its own thing.

To check the contrast on your own, you can use one of many [color contrast tools](http://webaim.org/resources/contrastchecker/) and see if they pass the a11y recommendations or not. In this case, `#ffffff` on `#00bcd4` fails while `#000000` on it passes. Hence, why we chose black instead of white.

mdecaffmeyer:
Thank you, I didn't know the WCAG 2.0 spec:-) it's really nice that the library computes automatically the required text color according to the spec.

[...]

Issue #608, opened 2017-05-09
Fix mdc demo site bugs for color specification


liquidx:
When the accent color is set to green, it seems like the ripples are purple instead of green.

I can see this in the Ripple example (at the bottom). I expect it to be green.

![screen shot 2017-05-09 at 1 56 17 pm](https://cloud.githubusercontent.com/assets/72062/25864970/543e7714-34bf-11e7-8390-9d2621ae51fe.png)

amsheehan:
We use custom theme variables to style ripples. One thing we can do is update the demos so that they use SASS. 

Here is more info on why you're getting a very easter-y ripple color: https://github.com/material-components/material-components-web/tree/master/packages/mdc-ripple#caveat-theme-custom-variables

traviskaufman:
To clarify, the work to be done here is:

- [ ] Convert `demos/demos.css` to `demos/demos.scss`
- [ ] Make `demos.scss` buildable by sass/webpack such that the demos change the scss theme vars, import `material-components-web`, and then add whatever custom CSS is already present

PR #918, opened 2017-07-07
docs(theme): Simplify README to follow best practices

lynnjepsen:


packages/mdc-theme/README.md
```diff
@@ -46,13 +46,13 @@ npm install --save @material/theme
 
 ## Usage
 
-### Sass
+### Themeing
```
yeelan0319: Theming. IMHO, I think `Change them colors` feels less ambiguous

packages/mdc-theme/README.md
```diff
@@ -46,13 +46,13 @@ npm install --save @material/theme
...
 
-#### Changing the theme colors
+MDC Theme mixins use CSS custom properties for browsers that support it, with a fallback to a pre-processed 
+static color if they don't. This enables runtime theming if CSS properties are available.
 
-MDC Theme makes it quite easy to change the theme colors for your application, and have the changes apply to all MDC-Web
-components. Simply define the three theme color variables before importing `mdc-theme` or any MDC-Web components that rely
-on it:
+To set the theme colors for your application, simply define the three theme color variables before importing
```
yeelan0319: Define or override? I feel the latter is more appropriate as we do provide default theme color. How about:

> MDC Theme makes it easy to develop your brand colors. You can simply override the default theme color by through Sass variable or CSS custom properties. In the latter case, it enables runtime theming.
> 
> **A note for Sass variables**, you need you define the three theme color variables before importing `mdc-theme` or any MDC-Web components that rely on it, like following:

packages/mdc-theme/README.md
```diff
@@ -36,8 +38,6 @@ In addition, it's important to consider the style of text:
 to establish a visual identity and color many parts of your application. The latter refers to the style of text
 that is most prominent (low opacity, high contrast), and used to display most content.
 
-The text contrast colors are automatically calculated at the Sass level and exposed as part of this module.
```
yeelan0319: We have several issue regarding this, worth to provide a paragraph explaining this.
#608 #755

packages/mdc-theme/README.md
```diff
@@ -64,205 +64,96 @@ $mdc-theme-background: #fff;
...
+| `--mdc-theme-text-icon-on-dark` | Icons on top of a dark-colored background. |
+
+### Sass Mixins, Variables, and Functions
+
+| Mixin                                           | Description |
+| ----------------------------------------------- | - |
+| `mdc-theme-prop($property, $style, $important)` | Applies a theme color to a property |
+| `mdc-theme-dark($root-selector, $compound)` | Creates a rule that is applied when the root element is within an Dark Theme context. If using the mixin on anything other than the base selector, you need to specify the base selector as a parameter. If using the mixin with a modifier class, pass `true` for the second argument. |
+
+#### Properties
```
yeelan0319: What are these properties and how can I use them? I don't think this clearly explained here.

lynnjepsen:
Updated README based on comments, PTAL.

packages/mdc-theme/README.md
```diff
@@ -16,28 +16,25 @@ path: /catalog/theme/
   </a>
 </div>-->
```
yeelan0319: How about:

> MDC Theme is a foundational module that theming MDC-Web components. 
> 
> The colors in this module are derived from three theme colors:
> - Primary: the primary color used in your application. This applies to a number of UI elements, such as app bars.
> - Accent: the accent color used in your application. This applies to UI elements such as FABs.
> - Background: the background color for your application. This is the color on top of which your UI is drawn.
> 
> and five text colors:
> - Primary, used for most text.
> - Secondary, used for text which is lower in the visual hierarchy.
> - Hint, used for text hints (such as those in text fields and labels).
> - Disabled, used for text in disabled components and content.
> - Icon, used for icons.
> 
> MDC-web themes the component using the combination of `mdc-theme--<theme>` for color and `mdc-theme-text-<text>-on-<theme>` for text on it. Users can apply the css modifiers to change the theme or the components.
>
> **Note: Dark theme**
> All components by defaults assume it is used on a light colored background (by default #FFF), but some of them are made aware of a Dark Theme context, aka placed on top of a dark background and provide a corresponding looking. There are two ways to make the components theme themselves in their dark mode: The first is to add `mdc-theme--dark` to a *container* element, which hold the component. The second way is to add `<component_name>--theme-dark` modifier class to the actual component element.

lynnjepsen:
I went with

Intro paragraph. More about how we define theme color and what we provided.
How to change theme color, using css custom properties and sass
CSS modifiers available out of box
How to define your theming rules using sass mixins

I think almost all users will want to know how to change theme color, so it seemed more important than CSS modifiers.

And I rewrote the intro paragraph with your suggestions.

packages/mdc-theme/README.md
```diff
@@ -16,27 +16,33 @@ path: /catalog/theme/
...
 
-MDC Theme is a foundational module that provides theming to MDC-Web components, and also makes it available to
-developers as Sass functions and mixins, as CSS custom properties, and as a set of CSS classes.
+This color palette comprises primary and accent colors that can be used for illustration or to develop your brand colors.
```
yeelan0319: super nit: feel if this sentence is repeating itself with the next one. Feel free to neglect if you disagree.

lynnjepsen: I took it straight from the Material Design guidelines page, so I'd like to leave it. I think it provides a link between the guidelines and are code.

And then I realized we didn't have a link to the guidelines! So I added that.

PR #922, opened 2017-07-09
fix(typography): Add missing typography color contrast / text colors

### LINKED ISSUE #883: Add missing typography color contrast  / text colors
Would it be possible to add the missing typography color contrast / text colors outlined on https://material.io/guidelines/style/typography.html#typography-other-typographic-guidelines. For example, caption is 54%, body1 is 87%, etc.

I couldn't find documentation that these exist on https://material-components-web.appspot.com/typography.html.

Looks like color contrast in MDL was implemented in https://github.com/google/material-design-lite/blob/mdl-1.x/src/_mixins.scss using `opacity`

touficbatache:
Fixes and closes #883 @amsheehan

### DIFF
--- a/packages/mdc-typography/_variables.scss
+++ b/packages/mdc-typography/_variables.scss
@@ -25,6 +25,11 @@ $mdc-typography-font-weight-values: (
   black: 900
 ) !default;
 
+$mdc-typography-font-contrast-values: (
+  light: .54%,
+  dark: .87%
+) !default;
+
 /* TODO(sgomes): Figure out what to do about desktop font sizes. */
 
 /* TODO(sgomes): Figure out what to do about i18n and i18n font sizes. */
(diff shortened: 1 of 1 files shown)


yeelan0319:
Is this pattern we should apply to all text? @amsheehan @lynnjepsen 
If we are not sure, maybe it worth asking the design team?

lynnjepsen:
There is design spec for these 87% and 54% values:

https://material.io/guidelines/style/typography.html#typography-other-typographic-guidelines

But I think we're already handling this in MDC Theme

https://github.com/material-components/material-components-web/blob/c509e7da32e8c14e29526b5e20bea68d98b2fe84/packages/mdc-theme/_variables.scss#L35

I'm going to comment on the original issue, and see if there is a better solution.

touficbatache:
It's been nearly a month. Should this be merged or what?!

yeelan0319:
@touficbatache ,we had a discussion with our designer for the text contrast as there is many ambiguity on design guideline for it. 

TL;DR: Text contrast (color) is about text hierarchy but not simply correlate to text size. We cannot simply associate certain typography size with contrast but need to look at larger context. So here I am closing this PR.

A little bit longer version about the discussion:
Here is a paragraph on design guideline - color section:  

> The most important text has an opacity of 87%.
> Secondary text, which is lower in the visual hierarchy, has an opacity of 54%.

However, importance is a ambiguous idea. It is highly correlated with the actual environment and context that the typography is being used for. So we cant make a hard and fast rule, based on the typescale and sizing, what should always inherit 87% and what inherits 54%. And for our pre-defined components, it do have a clear hierarchy within its scope, which enabled us to assign a color contrast to them. A good example could be `mdc-card`.

We are very grateful that you helped us surface this ambiguity from our design guideline even though we cannot accept this PR. It is a very good find out and we definitely will put further thoughts into our future guidelines.

touficbatache:
I understand this, but if you don't want to directly change the typographic styles, could we create classes that will do this?

yeelan0319:
If you look at theming `packages/mdc-theme/variables.scss` that we already provide the text contrast and corresponding css modifier. Though we don't have a quick answer to the combination, you should be able to use these text color modifier with `mdc-typography`.

PR #925, opened 2017-07-10
fix(drawer): Temporary drawer is below toolbar

touficbatache:
@amsheehan


touficbatache:
@lynnjepsen Could you please close #935 as this was resolved?

Shyam-Chen:
@touficbatache Has not yet released.

PR #941, opened 2017-07-13
fix(dialog): Dialog buttons should use primary color

### LINKED ISSUE #940: Dialog buttons should use primary color

> 0.15.0

> All browsers

### What OS are you using?

> Windows 7 / Android 7.0

> 1. Go to http://material-components-web.appspot.com/dialog.html
> 2. Observe the component's behavior

> The button text should use the primary color

> The button text is black

> All the photos show a dialog with colored buttons: https://material.io/guidelines/components/dialogs.html

touficbatache:
Fixes and closes #940


Garbee:
This issue smells to me like the actions for dialogs need a rethink in the way they are done over a simple color patch. Why are all these extra classes included when they are just regular MDC buttons? We could re-use all the same MDC button classes giving developers control over the views from there and simply add a single `mdc-dialog_action` node to do the dialog-specific stylings of a button (such as placement and margin to other nodes.)

yeelan0319:
I agree with what @Garbee mentioned here.

If we take a close look at [dialog spec](https://material.io/guidelines/components/dialogs.html#dialogs-specs), in the *color* section, it does mentioned that default system color should be used but user can specify to have accent color applied to distinguish action from text. 

In that case, I believe the best case is as what Garbee suggested having a `mdc-dialog__action` to specify color but not add it in general to `&__button`.

Could you make those changes? @touficbatache

touficbatache:
@yeelan0319 Yes sure!

touficbatache:
Yes, but what and where should I add it?

yeelan0319:
Let's add a paragraph at the bottom of ## Dialog usage, which should be a header3 ###

> ### Dialog Action Color
> Dialog actions use system colors by default, but you can use a contrasting color, such as the palette’s accent color, to distinguish dialog actions from dialog content. To emphasize an action from other contents, add `mdc-dialog_action` to `mdc-button` to apply accent color.

```
<aside class="mdc-dialog">
  <div class="mdc-dialog__surface">
    <footer class="mdc-dialog__footer">
      <button type="button" class="mdc-button mdc-dialog__footer__button mdc-dialog__footer__button--cancel">Decline</button>
      <button type="button" class="mdc-button mdc-dialog__footer__button mdc-dialog__footer__button--accept mdc-dialog__action">Accept</button>
    </footer>
</aside>
```

In general, the way we explain in our documents is including a general explanation what the class is about and a short example code to help people visualize how things could be used.

touficbatache:
Okay, thanks for helping me. I'll do that the next time I want to update the README, which I'll do right now :smiley:

touficbatache:
When will this be merged?

touficbatache:
Can anyone accept and merge this? @yeelan0319 @amsheehan @lynnjepsen @Garbee

packages/mdc-dialog/README.md
```diff
@@ -112,6 +112,20 @@ Some dialogs will not be tall enough to accomodate everything you would like to
 Note that unlike the css classnames, the specific ID names used do not have to be _exactly_ the same as listed above.
 They only need to match the values set for their corresponding aria attributes.
 
+### Dialog Action Color ###
+
+Dialog actions use system colors by default, but you can use a contrasting color, such as the palette’s accent color, to distinguish dialog actions from dialog content. To emphasize an action from other contents, add `mdc-dialog_action` to `mdc-button` to apply accent color.
```
yeelan0319: super nit: Should be `mdc-dialog__action` instead of `mdc-dialog_action`

touficbatache:
Done! :smiley:

piotr-cz:
This doesn't seem to have any effect on v0.23.0 as the color value is being overwritten by `.mdc-button:not(:disabled)` selector, which [forces primary color](https://github.com/material-components/material-components-web/blob/v0.23.0/packages/mdc-button/_mixins.scss#L42)

touficbatache:
Hmm, maybe we should add a `!important` to the __action CSS? What do you think @yeelan0319?

touficbatache:
Yeah, it seems pointless as of 0.23.0... Could you submit a PR to remove the `mdc-dialog__action` from the CSS and the README? I think you should also point out that it uses the primary color by default in the README.md file.

touficbatache:
Oh, after looking back into the code, I found out that the `mdc-dialog__action` class makes the button use [the secondary color](https://github.com/material-components/material-components-web/blob/master/packages/mdc-dialog/mdc-dialog.scss#L142) and not the primary.

PR #946, opened 2017-07-14
chore(animation): Rename `exit` to `exit-permanent` to match spec

acdvorak:
BREAKING CHANGE: We currently don't distinguish between "temporary" and "permanent" exits in mdc-animation. However, the timing function we're using is actually that of "permanent", so we should rename the function to reflect that. This will also allow us to add "temporary" exit in the future.

https://material.io/guidelines/motion/movement.html#movement-movement-in-out-of-screen-bounds


acdvorak:
Verified all affected demos (checkbox, drawer, linear-progress, radio, select, snackbar).

PR #948, opened 2017-07-14
chore: Add @alternate annotation

lynnjepsen:
This lets us compile our Sass files into Closure Stylesheets.

### DIFF
--- a/packages/mdc-ripple/_mixins.scss
+++ b/packages/mdc-ripple/_mixins.scss
@@ -59,6 +59,8 @@
   // stylelint-disable at-rule-empty-line-before, block-closing-brace-newline-after
   @if ($theme-style) {
     $theme-value: map-get($mdc-theme-property-values, $theme-style);
+
+    /* @alternate */
     $css-var: $theme-value;
     $css-var: var(--mdc-theme-#{$theme-style}, $theme-value);
 
@@ -115,6 +117,8 @@
     left: calc(50% - #{$radius});
     width: $radius * 2;
     height: $radius * 2;
+
+    /* @alternate */
     transform: scale(0);
     transform: scale(var(--mdc-ripple-fg-scale, 0));
   }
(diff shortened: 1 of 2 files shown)


packages/mdc-ripple/_mixins.scss
```diff
@@ -179,8 +192,12 @@
...
       left: 0;
+
+      /* @alternate */
       width: $radius;
       width: var(--mdc-ripple-fg-size, $radius);
+
+      /* @alternate */
       height: $radius;
       height: var(--mdc-ripple-fg-size, $radius);
       transform: scale(0);
```
acdvorak: Insert blank line before `transform`

packages/mdc-ripple/_mixins.scss
```diff
@@ -195,12 +212,20 @@
 
   &.mdc-ripple-upgraded--unbounded#{$pseudo} {
     top: 0;
+
+    /* @alternate */
```
acdvorak: For readability, I would suggest using the same style you have above:

```css
/* @alternate */
top: 0;
top: var(--mdc-ripple-top, 0);

/* @alternate */
left: 0;
left: var(--mdc-ripple-left, 0);

...
```

Or you could do something like this:

```css
/* @alternate */ top: 0;
/* @alternate */ top: var(--mdc-ripple-top, 0);

/* @alternate */ left: 0;
/* @alternate */ left: var(--mdc-ripple-left, 0);

...
```

Personally, I tend to prefer the first one. Since we only have two alternate values, we can get away with only having a single annotation.

Ditto for the other `@alternate` annotations in this file.

packages/mdc-theme/_mixins.scss
```diff
@@ -28,11 +28,15 @@
 
   @if $important {
     #{$property}: map-get($mdc-theme-property-values, $style) !important;
+
+    /* @alternate */
```
acdvorak: Ditto my comment above

packages/mdc-ripple/_mixins.scss
lynnjepsen: Our linter doesn't like that...

"Unexpected empty line before declaration "

amsheehan: feel free to disable the linter for this block

acdvorak: Interesting. I guess it just feels weird to have a blank line before `/* alternate */` (which has the effect of grouping duplicate properties together) but _not_ a blank line before `transform`.

PR #963, opened 2017-07-17
chore: Update stylelint modules to the latest

lynnjepsen:
* Updated .stylelintrc.yaml to no longer include deprecated rules
* Updated .stylelintrc.yaml to include at-rule-no-unknown, so that SCSS
 @-rules are linted properly


.stylelintrc.yaml
```diff
@@ -206,22 +220,8 @@ rules:
     - true
     - ignoreProperties:
         - contain
-  # There is no reason that a specific ID would be needed for UI components
-  selector-no-id: true
```
amsheehan: I think we still want to enforce these things, not just remove them.

lynnjepsen:
Thanks for the review. I've re-enforced the rules with some new config! PTAL.

Issue #944, opened 2017-07-13
.mdc-grid-tile__support-text does not cut off the text overflow.


cgfarmer4:
### What MDC-Web Version are you using?

> 0.15.0

### What browser(s) is this bug affecting?

> All

### What OS are you using?

>MacOS 10.12.1, Android WebView

### What are the steps to reproduce the bug?

1. Add support text longer than expected to the `.mdc-grid-tile__support-text` element.

### What is the expected behavior?

`.mdc-grid-tile__support-text` element concatenates with ellipsis similar to the parent.

### What is the actual behavior?

Text overflows and looks janky.

![screen shot 2017-07-13 at 2 37 39 pm](https://user-images.githubusercontent.com/336449/28189048-2f6fad02-67d9-11e7-96d0-edc5286d311b.png)

lynnjepsen:
Seems like we should add these three properties to `mdc-grid-tile__support-text`

```
text-overflow: ellipsis;
white-space: nowrap;
overflow: hidden;
```

Happy to accept a PR to fix this!

cgfarmer4:
Had old email in my git history but hopefully that new one has all the right minutiae.

PR #965, opened 2017-07-18
fix(grid): Supoort title overflow not concatenating(#944)

cgfarmer4:
https://github.com/material-components/material-components-web/issues/944#issuecomment-315866959


cgfarmer4:
I signed it!

PR #974, opened 2017-07-19
feat(toolbar): add `pointer` for `icon` element

Shyam-Chen:


Shyam-Chen:
done.

Garbee:
This seems icky to me. The *icons* shouldn't trigger anything. In fact they should be `pointer-events: none` so they don't trigger any events in JS or anything in-accurately. It is the underlying buttons that should have the hover triggers and event listeners if any are to exist.

yeelan0319:
@Garbee Curious about your suggestion here "icon shouldn't trigger anything. It is the underlying buttons that should have the hover triggers". Does this means a button need to be nested within `&__icon` if the icon is intended to have an action? Such as:
```
    <div class="mdc-toolbar__icon" aria-label="Download" alt="Download">
      <button><i class="material-icons">file_download</i></button>
    </div>
  </div>
</header>
```

IMHO, this makes toolbar structure bulkier than it is but the gain here is pretty limited, since based on the [guideline](https://material.io/guidelines/layout/structure.html#structure-app-bar) that all the icons within toolbar should be actionable. Correct me if I get your meaning wrong.

Garbee:
Ah, I see now looking at the code more. The class naming changed some from what we were initially demoing for the component early on. This LGTM knowing that.

yeelan0319:
Yeah, there were some other issues related to those icons so we renamed it.
I saw you filed a couple of issues regarding the naming, let's triage them later on.

PR #1000, opened 2017-07-24
Edge CSS Custom Properties Support

arjunyel:
Edge now supports [CSS Custom Properties](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/csscustompropertiesakacssvariables/)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -235,7 +235,7 @@ following to the `<head>` tag of `index.html`:
 </style>
 ```
 
-If you are using any modern browser you'll see that the button background as well as the focused underline and label on text
+If you're using any supported browser, besides IE 11, you'll see that the button background as well as the focused underline and label on text
 fields are now a nice, dark shade of blue.
 
 > Note that using CSS Variables is just one way of theming using MDC-Web. Check out our
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -235,7 +235,7 @@ following to the `<head>` tag of `index.html`:
 </style>
 ```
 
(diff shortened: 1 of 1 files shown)


arjunyel:
@kfranqueiro good catch, I updated the pull request

docs/getting-started.md
```diff
@@ -235,7 +235,7 @@ following to the `<head>` tag of `index.html`:
 </style>
 ```
 
-If you are using any modern browser besides Edge (which is currently [working on it](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/csscustompropertiesakacssvariables/)), you'll see that the button background as well as the focused underline and label on text
+If you're using any supported browser, besides IE 11, you'll see that the button background as well as the focused underline and label on text
```
kfranqueiro: Nit: remove extra comma after "browser"; IMO the sentence flows better with just the one comma after "IE 11".

Garbee: The extra comma here is perfectly correct grammatically. "besides IE11" in this case is an appositive, which are denoted with *surrounding* commas.

kfranqueiro: Ah, point taken. Disregard that nit then :)

kfranqueiro:
I'm going to go ahead and merge this documentation update. However, we're aware of some issues related to Edge, some of which may be due to issues with its current implementation of CSS custom properties (see known issues listed on [caniuse](http://caniuse.com/#search=variables)). We are tracking these in Pivotal, but don't have a timeline on fixing them.

PR #1003, opened 2017-07-24
docs(elevation): Moving CSS Classes ahead of Sass mixins, 

lynnjepsen:
since they have heavier usage

### DIFF
--- a/packages/mdc-elevation/README.md
+++ b/packages/mdc-elevation/README.md
@@ -48,7 +48,7 @@ npm install --save @material/elevation
 
 Some components have a set elevation. For example, a raised MDC Button has elevation 2.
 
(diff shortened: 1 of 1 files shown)


packages/mdc-elevation/README.md
```diff
@@ -44,6 +44,18 @@ npm install --save @material/elevation
 
 ## Usage
 
+### CSS Classes
+
+Some components have a set elevation. For example, a raised MDC Button has elevation 2.
+
+However, if you want to set the elevation of an element, which is not an Material Design component, then you
```
amsheehan: "~However~ If you want... ...not _a_ Material Design component, ~then~ you can apply..."

PR #1008, opened 2017-07-25
fix(ripple): Fallback to use CSS only ripple in Edge.

### LINKED ISSUE #663: Edge: JS Button problem
When using/testing mdc-web on the demo site, there seems to be an issue with the js button.

0.11.1 via this page:
https://material-components-web.appspot.com/

Microsoft Edge 41.16193.1001.0

### What OS are you using?

Windows 10 Build 16193

Open Edge to https://material-components-web.appspot.com/button.html and click on a button that is not CSS only. 

It behaves normally or  degrades gracefully to a CSS only button (which appears to be what happens on IE 11)

js-button-edge

yeelan0319:
Closes #663.

### DIFF
--- a/packages/mdc-ripple/util.js
+++ b/packages/mdc-ripple/util.js
@@ -34,7 +34,16 @@ export function supportsCssVariables(windowObj) {
     windowObj.CSS.supports('(--css-vars: yes)') &&
     windowObj.CSS.supports('color', '#00000000')
   );
-  return explicitlySupportsCssVars || weAreFeatureDetectingSafari10plus;
+  // See: https://caniuse.com/#search=custom%20properties
+  // See: README section on Edge
+  const weAreFeatureDetectingEdge = (
+    typeof CSS !== 'undefined' &&
+    CSS.supports('(-ms-ime-align:auto)')
+  );
+  return (
+    (explicitlySupportsCssVars && !weAreFeatureDetectingEdge) ||
+    weAreFeatureDetectingSafari10plus
+  );
 }
 
 //
--- a/packages/mdc-ripple/README.md
+++ b/packages/mdc-ripple/README.md
@@ -32,6 +32,7 @@ path: /catalog/ripples/
     - [Keyboard interaction for custom UI components](#keyboard-interaction-for-custom-ui-components)
     - [Specifying known element dimensions](#specifying-known-element-dimensions)
   - [Caveat: Safari](#caveat-safari)
+  - [Caveat: Edge](#caveat-edge)
   - [Caveat: Theme Custom Variables](#caveat-theme-custom-variables)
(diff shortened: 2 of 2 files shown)


packages/mdc-ripple/util.js
```diff
@@ -34,7 +34,16 @@ export function supportsCssVariables(windowObj) {
     windowObj.CSS.supports('(--css-vars: yes)') &&
     windowObj.CSS.supports('color', '#00000000')
   );
-  return explicitlySupportsCssVars || weAreFeatureDetectingSafari10plus;
+  // See: https://caniuse.com/#search=custom%20properties
+  // See: README section on Edge
+  const weAreFeatureDetectingEdge = (
```
amsheehan: This should be named something less human. It's full on uncanny valley.

`supportsMSIMEAlignAuto` maybe

yeelan0319: But won't it make it less clear that our goal is FD Edge? So when we change the FD feature, we are not always changing the function name?

yeelan0319:
@kfranqueiro I think you raised an very interesting solution and I like it!

Another worry I have with this approach is a more specific FD definitely will lower the "false positive" rate (imagine there is a Edge browser that have CSS Variables correctly implemented a day and in my method it raises the false alarm) but at the same time increase the "false negative" rate (imagine another day that a different Edge version comes out with CSS Variables not correct but not applied to its parent as well, in this case it will think it is right). So there are pros and cons to both method.

I think if we want to be more specific, we need further to valid:
- Animation is somehow tested as well
- The behavior of applying style to its parent is consistent across all Edge version

WDYT?

yeelan0319:
@kfranqueiro @amsheehan I am not very familiar with the cons of UA sniffing besides it sounds to me a little bit unreliable?

kfranqueiro:
I think typically the encouragement towards feature detection over UA sniffing is because user agents can be easily spoofed and even vendors themselves have been known to do funny things to them to defeat developers' workarounds, so it becomes a never-ending battle.  In this particular case though, the efficacy of the feature detect is also in question.

I'm currently downloading the preview release VM to see if its version of Edge behaves the same, and I'll see what versions I have on my own machines tonight.

kfranqueiro:
I was able to test with Edge 15.15063 and 15.16215 (via modern.ie VMs) which both exhibit the same behavior which my POC detects.  I was able to test Edge 14.14393 on my personal machine which does not report support of custom properties at all, so this problem is specific to Edge 15 (14 already won't upgrade ripples).

yeelan0319:
Scoped FD sounds like a plan! For the moment have the good willing that Edge will make it behave as other browsers do.

@kfranqueiro I will close this PR and let you file a new one with scoped FD then 👍

PR #1012, opened 2017-07-25
feat(auto-init): Fire event on mdcAutoInit complete

### LINKED ISSUE #954: Event to detect all components init is complete
Is there a way to know when `mdc` components are completely loaded after `mdc.autoInit()` is called? I cannot find anything like this in the docs.

yshterev:
FIx #954 

@yeelan0319 The dialog component is extending `MDCComponent` so it has direct access to `emit` method. I am not sure how to access the emit method properly in `mdcAutoInit` function. I tried to import `MDCComponent`, but I guess it is not good. 

Also I am not sure what the test should be.
Any help appreciated.

### DIFF
--- a/packages/mdc-auto-init/README.md
+++ b/packages/mdc-auto-init/README.md
@@ -160,3 +160,10 @@ warning, you could simply pass in a nop.
 ```
 
 This will suppress any warnings about already initialized elements.
+
+### Events
+
+#### MDCAutoInit:End
+Triggered when initialization of all components is complete.
+
+`document.addEventListener("MDCAutoInit:End", () => {...});`
--- a/packages/mdc-auto-init/index.js
+++ b/packages/mdc-auto-init/index.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+// import {MDCComponent} from '@material/base';
+
 const registry = Object.create(null);
 
 const CONSOLE_WARN = console.warn.bind(console);
@@ -49,6 +51,8 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
       configurable: true,
     });
   }
+
+  // MDCComponent.emit('MDCAutoInit:End');
 }
 
 mdcAutoInit.register = function(componentName, Ctor, warn = CONSOLE_WARN) {
--- a/packages/mdc-auto-init/index.js
+++ b/packages/mdc-auto-init/index.js
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-// import {MDCComponent} from '@material/base';
-
 const registry = Object.create(null);
 
 const CONSOLE_WARN = console.warn.bind(console);
@@ -25,6 +23,7 @@ const CONSOLE_WARN = console.warn.bind(console);
  */
 export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
   const nodes = root.querySelectorAll('[data-mdc-auto-init]');
+  let component = null;
   for (let i = 0, node; (node = nodes[i]); i++) {
     const ctorName = node.dataset.mdcAutoInit;
     if (!ctorName) {
@@ -43,7 +42,7 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
     }
 
     // TODO: Should we make an eslint rule for an attachTo() static method?
-    const component = Ctor.attachTo(node);
+    component = Ctor.attachTo(node);
     Object.defineProperty(node, ctorName, {
       value: component,
       writable: false,
@@ -52,7 +51,9 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
     });
   }
 
-  // MDCComponent.emit('MDCAutoInit:End');
+  if (component !== null) {
+    component.emit('MDCAutoInit:End', {}, true);
+  }
 }
 
 mdcAutoInit.register = function(componentName, Ctor, warn = CONSOLE_WARN) {
--- a/packages/mdc-auto-init/index.js
+++ b/packages/mdc-auto-init/index.js
@@ -26,7 +26,7 @@ const CONSOLE_WARN = console.warn.bind(console);
  * @param {!Object} evtData
  * @param {boolean=} shouldBubble
  */
-function emit(element, evtType, evtData, shouldBubble = false) {
+export function emit(element, evtType, evtData, shouldBubble = false) {
   let evt;
   if (typeof CustomEvent === 'function') {
     evt = new CustomEvent(evtType, {
--- a/packages/mdc-auto-init/index.js
+++ b/packages/mdc-auto-init/index.js
@@ -24,7 +24,7 @@ const CONSOLE_WARN = console.warn.bind(console);
  * @param {!Object} evtData
  * @param {boolean=} shouldBubble
  */
-export function emit(evtType, evtData, shouldBubble = false) {
+function _emit(evtType, evtData, shouldBubble = false) {
   let evt;
   if (typeof CustomEvent === 'function') {
     evt = new CustomEvent(evtType, {
@@ -71,7 +71,7 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
     });
   }
 
-  emit('MDCAutoInit:End', {});
+  _emit('MDCAutoInit:End', {});
 }
 
 mdcAutoInit.register = function(componentName, Ctor, warn = CONSOLE_WARN) {
--- a/packages/mdc-auto-init/index.js
+++ b/packages/mdc-auto-init/index.js
@@ -19,11 +19,33 @@ const registry = Object.create(null);
 const CONSOLE_WARN = console.warn.bind(console);
 
 /**
+ * Fires a cross-browser-compatible custom event from HTML element or HTML document,
+ * with the given data.
+ * @param {HTMLElement, HTMLDocument} element
+ * @param {string} evtType
+ * @param {!Object} evtData
+ * @param {boolean=} shouldBubble
+ */
+function emit(element, evtType, evtData, shouldBubble = false) {
+  let evt;
+  if (typeof CustomEvent === 'function') {
+    evt = new CustomEvent(evtType, {
+      detail: evtData,
+      bubbles: shouldBubble,
+    });
+  } else {
+    evt = document.createEvent('CustomEvent');
+    evt.initCustomEvent(evtType, shouldBubble, false, evtData);
+  }
+
+  element.dispatchEvent(evt);
+}
+
+/**
  * Auto-initializes all mdc components on a page.
  */
 export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
   const nodes = root.querySelectorAll('[data-mdc-auto-init]');
-  let component = null;
   for (let i = 0, node; (node = nodes[i]); i++) {
     const ctorName = node.dataset.mdcAutoInit;
     if (!ctorName) {
@@ -42,7 +64,7 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
     }
 
     // TODO: Should we make an eslint rule for an attachTo() static method?
-    component = Ctor.attachTo(node);
+    const component = Ctor.attachTo(node);
     Object.defineProperty(node, ctorName, {
       value: component,
       writable: false,
@@ -51,9 +73,7 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
     });
   }
 
-  if (component !== null) {
-    component.emit('MDCAutoInit:End', {}, true);
-  }
+  emit(document, 'MDCAutoInit:End', {}, true);
 }
 
 mdcAutoInit.register = function(componentName, Ctor, warn = CONSOLE_WARN) {
--- a/packages/mdc-auto-init/index.js
+++ b/packages/mdc-auto-init/index.js
@@ -18,12 +18,6 @@ const registry = Object.create(null);
 
 const CONSOLE_WARN = console.warn.bind(console);
 
-/**
- * Fires a cross-browser-compatible custom event with the given data
- * @param {string} evtType
- * @param {!Object} evtData
- * @param {boolean=} shouldBubble
- */
 function _emit(evtType, evtData, shouldBubble = false) {
   let evt;
   if (typeof CustomEvent === 'function') {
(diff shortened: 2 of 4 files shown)


yshterev:
I signed it!

yeelan0319:
@yshterev Thanks for filing the PR.

So here in the `auto-init` you don't really need to import `MDCComponent` - all the components themselves are instances of an derived `MDCComponent` and have this `emit` method.

**For firing event**
Take a look at `mdc-auto-init` index.js line 45 where an component instance is created here. 
You can add `component.emit('MDCAutoInit:End', {}, true);` after line 51. It will make the component emit an `MDCAutoInit:End` event after it inits. 
You can find a detailed explanation of the argument explanation in `mdc-base` document.

**Test firing event**
Take a look at unit test for `mdc-tab-bar` line 234. It need some modification but basically you can see a good reference how to test whether a event if fired.

**Listen to fired event (FYI)**
I also created a [pen](https://codepen.io/yeelan0319/pen/EvjyrL) illustrate how the custom event could be listened.

Feel free to ping me back for further question and thanks again for being on it!

yshterev:
@yeelan0319 The FakeComponent in `mdc-auto-init` test was too fake. It did not have `emit` method, so I decided to use add the real `mdc-base` component. I think this way the test will be more accurate. 

I am not sure my test logic is very good. It relies on the first argument passed to the `handler` func being the event object.  Please review.

packages/mdc-auto-init/index.js
```diff
@@ -41,14 +42,18 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
...
+    component = Ctor.attachTo(node);
     Object.defineProperty(node, ctorName, {
       value: component,
       writable: false,
       enumerable: false,
       configurable: true,
     });
   }
+
+  if (component !== null) {
```
yeelan0319: There might be multiple components initialize with `mdc-auto-init` method and my original thought is each of them should have their own `init:end` event. 

If you are interested in just having one custom event about the whole auto-init end, maybe we shouldn't do that from a component level.

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -18,14 +18,29 @@ import bel from 'bel';
...
-    this.node = node;
+    super();
+    this.root_ = node;
+  }
+
+  get root() {
+    return this.root_;
+  }
+
+  getDefaultFoundation() {
```
yeelan0319: Why we are adding Foundation method here? It would inherent from MDCComponent so you don't need to reimplement it.

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +110,13 @@ test('#register warns when registered key is being overridden', () => {
...
+
+test('#emit event from MDCComponent on init complete', () => {
+  const root = setupTest();
+  const handler = td.func('init event handler');
+
+  root.addEventListener('MDCAutoInit:End', handler);
+  mdcAutoInit(root);
+
+  td.verify(handler(td.matchers.isA(Object)));
+});
```
yeelan0319: Test looks good!

packages/mdc-auto-init/README.md
```diff
@@ -160,3 +160,10 @@ warning, you could simply pass in a nop.
 ```
 
 This will suppress any warnings about already initialized elements.
+
+### Events
+
+#### MDCAutoInit:End
+Triggered when initialization of all components is complete.
```
yeelan0319: If that's the case, we probably shouldn't rely on component to emit event.

See my comment below

packages/mdc-auto-init/index.js
yshterev: Probably I got this wrong. 

If each component emits its own `init:end` event when done, how would you know when all of components are initialzied? Or maybe we can use combination of both? Each component can emit own event when initialized, like `ComponentNameInit:End` and then 'MDCAutoInit:End' when all of them are initialized?

kfranqueiro: If I understand correctly the original issue was regarding having a way to tell when _all_ components initialize, so I agree that should be the main goal of this PR.

Having one event fire per component init seems potentially excessive and is perhaps a separate bridge we should cross if a convincing use case arises?

Also, I'm not sure whether it makes sense to have the "all initialized" event emit from a particular component, since which component emits it could change just based on how an app changes during development.  I'm open to thoughts/arguments otherwise here though.

yshterev: @kfranqueiro When I submitted the issue I was not sure. The discussion in the issue makes it clear it is about having a way to tell when all components initialize, so I changed its name to be more clear.

I guess using the component to trigger the event was suggested by @yeelan0319 to keep it DRY and reuse the [emit](https://github.com/material-components/material-components-web/blob/master/packages/mdc-base/component.js#L108) method.

I also think the PR should resolve the issue, meaning "all initialized" only until other use case arises.

yeelan0319: @yshterev I looked at the comments on the issue and discussion here. I think it was slightly different though when we discuss it internally as we thought it would be done at component level. 
So as now we understand the landscape of the question better, I think it should be proper to dispatch a custom event directly instead of using `emit` method from a single component. Could you update the PR accordingly? Thanks!

lynnjepsen:
Hey @yshterev, is there anything unclear about the changes requested here?

test/unit/mdc-auto-init/mdc-auto-init.test.js
yshterev: Some tests don't pass. The error is `Error: Subclasses must override getDefaultFoundation to return a properly configured foundation class`

yshterev:
@lynnjepsen Sorry for the delay. PR updated.

yshterev:
What should I do to pass the codecov test? Maybe rebase from master?

test/unit/mdc-auto-init/mdc-auto-init.test.js
yeelan0319: Thanks for updating the PR. 
I think once we directly implemented emit function, we no longer need the `fake` element here to inherit from `MDCComponent`.
I think we can safely change it back to what it was and simply adding two new test cases.

yeelan0319:
@yshterev Any updates to the changes or let me know if you have more questions!

yshterev:
FakeComponent in test reverted. 2 new tests added for the emit function.

packages/mdc-auto-init/index.js
```diff
@@ -19,6 +19,29 @@ const registry = Object.create(null);
...
 
 /**
+ * Fires a cross-browser-compatible custom event from HTML element or HTML document,
+ * with the given data.
+ * @param {HTMLElement, HTMLDocument} element
+ * @param {string} evtType
+ * @param {!Object} evtData
+ * @param {boolean=} shouldBubble
+ */
+export function emit(element, evtType, evtData, shouldBubble = false) {
```
yeelan0319: IMHO, I might not consider making emit a **private** helper function of `mdc-auto-init` instead of public. So far, I cannot see a strong argument why this method need to be public.

If testing is your biggest concern, how about cover both branch using the following methods?
The logic behind this is, we don't need explicitly export `emit` to test it but we mock the run-time environment (as you did in testing `emit` directly) and see if `mdcAutoInit` behave uniformly in both case.
```
test('#emit event from MDCComponent on init complete', () => {
  const root = setupTest();
  const handler = td.func('init event handler');
  let evt = null;

  td.when(handler(td.matchers.isA(Object))).thenDo((evt_) => {
    evt = evt_;
  });

  const type = 'MDCAutoInit:End';

  document.addEventListener(type, handler);
  mdcAutoInit(root);

  assert.isOk(evt !== null);
  assert.equal(evt.type, type);
});

test('#emit event from MDCComponent on init complete - custom events not supported', () => {
  const root = setupTest();
  const handler = td.func('init event handler');
  let evt = null;

  td.when(handler(td.matchers.isA(Object))).thenDo((evt_) => {
    evt = evt_;
  });

  const type = 'MDCAutoInit:End';

  document.addEventListener(type, handler);
  const {CustomEvent} = window;
  window.CustomEvent = undefined;
  try {
    mdcAutoInit(root);
  } finally {
    window.CustomEvent = CustomEvent;
  }

  assert.isOk(evt !== null);
  assert.equal(evt.type, type);
});
```

Let me know if this makes sense or your consideration for making `emit` public. Thanks!

yshterev: Sure. Your proposal is better. I exported `emit` just because of the test.
Do you think I should remove the `element` param from the emit function and just use `document`? I don't think `element` param makes sense anymore because tests always attach the listener to document.

yeelan0319: @yshterev Sorry for the tardiness. Yes. I think that is totally a valid suggestion, right now we are only emitting event at document level, which means the `element` is not really being used. Feel free to proceed with this idea and let me know when it is ready to review.

yshterev:
@yeelan0319 Updated

packages/mdc-auto-init/index.js
```diff
@@ -19,6 +19,27 @@ const registry = Object.create(null);
 const CONSOLE_WARN = console.warn.bind(console);
 
 /**
+ * Fires a cross-browser-compatible custom event with the given data
+ * @param {string} evtType
+ * @param {!Object} evtData
+ * @param {boolean=} shouldBubble
+ */
+export function emit(evtType, evtData, shouldBubble = false) {
```
yeelan0319: Could we consider making this a **private** method?
My concern is there might not be a use case where this function will be called by external user (now). 

We could do `function emit_` to identify this is a private method and remove `export`. I also commented about revoke and test below.

What do you think?

packages/mdc-auto-init/index.js
```diff
@@ -49,6 +70,8 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
       configurable: true,
     });
   }
+
+  emit('MDCAutoInit:End', {});
```
yeelan0319: `_emit('MDCAutoInit:End', {});` instead.

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -18,6 +18,7 @@ import bel from 'bel';
 import td from 'testdouble';
 import {assert} from 'chai';
 import mdcAutoInit from '../../../packages/mdc-auto-init';
+import {emit} from '../../../packages/mdc-auto-init';
```
yeelan0319: We can proxy over `mdcAutoInit ` to test private method `emit` is called, so we could remove this line.

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
 
   td.verify(warn(contains('(mdc-auto-init) Overriding registration')));
 });
+
+test('#emit dispatches a custom event with supplied data', () => {
```
yeelan0319: Consider rewording '#emit dispatches a custom event with supplied data' > '#mdcAutoInit dispatches a *MDCAutoInit:End* event when initialization of all components is complete'

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+  const handler = td.func('eventHandler');
+  let evt = null;
+
+  td.when(handler(td.matchers.isA(Object))).thenDo((evt_) => {
+    evt = evt_;
+  });
+
+  const data = {evtData: true};
```
yeelan0319: I think we could remove this line as the event doesn't come with data

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+  const type = 'MDCAutoInit:End';
+
+  document.addEventListener(type, handler);
+  emit(type, data);
```
yeelan0319: And call `mdcAutoInit()` instead.

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+
+  assert.isOk(evt !== null);
+  assert.equal(evt.type, type);
+  assert.deepEqual(evt.detail, data);
```
yeelan0319: If we remove data above, also here.

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+});
+
+test('#emit dispatches an event with supplied data - custom events not supported', () => {
```
yeelan0319: Same thing here. Consider rewording '#mdcAutoInit dispatches a *MDCAutoInit:End* event when initialization of all components is complete - custom events not supported'?

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+  const handler = td.func('eventHandler');
+  let evt = null;
+
+  td.when(handler(td.matchers.isA(Object))).thenDo((evt_) => {
+    evt = evt_;
+  });
+
+  const data = {evtData: true};
```
yeelan0319: Here

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+  const type = 'MDCAutoInit:End';
+
+  document.addEventListener(type, handler);
+  const {CustomEvent} = window;
+  window.CustomEvent = undefined;
+  try {
+    emit(type, data);
```
yeelan0319: And here

test/unit/mdc-auto-init/mdc-auto-init.test.js
```diff
@@ -95,3 +96,47 @@ test('#register warns when registered key is being overridden', () => {
...
+  } finally {
+    window.CustomEvent = CustomEvent;
+  }
+
+  assert.isOk(evt !== null);
+  assert.equal(evt.type, type);
+  assert.deepEqual(evt.detail, data);
```
yeelan0319: And here

yshterev:
@yeelan0319 Sorry .. I was sleepy I guess. Updated.

yshterev:
@yeelan0319 My pleasure 👍

PR #1022, opened 2017-07-26
feat(button): Modify button HTML structure and corresponding styles

### LINKED ISSUE #985: Make button stateful
- [ ] Make necessary changes to button style to have proper default state
- [ ] Make `mdc-ripple` be stateful

yeelan0319:
Close #985 

BREAKING CHANGE: Modify `mdc-button` HTML to have a wrapper `mdc-button__inner`. All existing implementations need to add this extra layer to apply correct styles.

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -23,20 +23,25 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
     <style>
-      fieldset {
-        margin: 24px;
-        margin-top: 0;
-        margin-bottom: 16px;
+      .demo-wrapper {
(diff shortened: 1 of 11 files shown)


Garbee:
What functionality being added/changed requires the addition of an inner element?

yeelan0319:
@Garbee We have an [milestone](https://github.com/material-components/material-components-web/milestone/7) for button improvements
TL;DR, there are some baseline styles changes but mostly it is because the newly added shade effects which goes parallel with ripple effect.

amsheehan:
Did anyone have a practical conversation with design on these changes? My perspective is that we should be trying to keep these components as true to native elements as we can. Putting inline elements inside of a button, while technically valid HTML, seems to overcomplicate `<button>`'s unnecessarily.

yeelan0319:
Closing it, we had a discussion for not making the breaking changes, see #985

PR #1028, opened 2017-07-27
feat(typography): Add button style to typography

yeelan0319:
#Background

This feature is added for allowing button style being used in multiple components

### DIFF
--- a/packages/mdc-typography/_mixins.scss
+++ b/packages/mdc-typography/_mixins.scss
@@ -35,6 +35,8 @@
   font-weight: #{map-get($style-props, font-weight)};
   letter-spacing: map-get($style-props, letter-spacing);
   line-height: map-get($style-props, line-height);
+  text-decoration: map-get($style-props, text-decoration);
+  text-transform: map-get($style-props, text-transform);
 }
 
 @mixin mdc-typography-adjust-margin($style) {
--- a/packages/mdc-typography/_variables.scss
+++ b/packages/mdc-typography/_variables.scss
@@ -35,76 +35,107 @@ $mdc-typography-styles: (
     line-height: 7rem, /* 112sp */
     font-weight: map-get($mdc-typography-font-weight-values, light),
     letter-spacing: -.04em,
-    margin: -1rem 0 3.5rem -.085em /* -16sp 0 56sp -.085em */
+    margin: -1rem 0 3.5rem -.085em /* -16sp 0 56sp -.085em */,
+    text-decoration: inherit,
+    text-transform: inherit
   ),
   display3: (
     font-size: 3.5rem, /* 56px */
     line-height: 3.5rem, /* 56px */
     font-weight: map-get($mdc-typography-font-weight-values, regular),
     letter-spacing: -.02em,
-    margin: -8px 0 64px -.07em
+    margin: -8px 0 64px -.07em,
+    text-decoration: inherit,
+    text-transform: inherit
   ),
   display2: (
     font-size: 2.813rem, /* 45px */
     line-height: 3rem, /* 48px */
     font-weight: map-get($mdc-typography-font-weight-values, regular),
     letter-spacing: normal,
-    margin: -.5rem 0 4rem -.07em /* -8sp 0 64sp -.07em */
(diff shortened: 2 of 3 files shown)


packages/mdc-typography/_mixins.scss
```diff
@@ -47,3 +51,11 @@
   white-space: nowrap;
   overflow: hidden;
 }
+
+@mixin mdc-typography-button {
```
lynnjepsen: I think this mixin should be in MDC Button package, and maybe renamed to `mdc-button-text`

And you should replace line 46(ish) in mdc-button.scss with `@include mdc-button-text`

Does `user-select: none;` and `-webkit-appearance: none` really belong in this mixin?

packages/mdc-typography/_mixins.scss
```diff
@@ -35,6 +35,10 @@
   font-weight: #{map-get($style-props, font-weight)};
   letter-spacing: map-get($style-props, letter-spacing);
   line-height: map-get($style-props, line-height);
+
+  @if $style == "button" {
```
lynnjepsen: I wouldn't *force* all our users to include this mixin all the time. I would leave *recommendation* that they include both mixins. Its a little more flexible. 

Theres nothing on the [typography spec](https://material.io/guidelines/style/typography.html#) about buttons text being centered, though there is some information about it being "all caps". But there is ALSO information about *not* using caps for certain languages. So again, not forcing the user to include the other mixin gives us flexibility to handle these situations.

packages/mdc-typography/_variables.scss
```diff
@@ -106,5 +106,12 @@ $mdc-typography-styles: (
     font-weight: map-get($mdc-typography-font-weight-values, regular),
     letter-spacing: .08em,
     margin: -.5rem 0 1rem -.04em /* -8sp 0 16sp -.04em */
+  ),
+  button: (
+    font-size: .875rem, /* 14sp */
```
lynnjepsen: MDC Button actually uses 14px font-size and 36px line-height explicitly.

Are you changing this on purpose?

yeelan0319: .0875rem == 14px and 2.25rem == 36px, which I think it is exactly the same number.

packages/mdc-typography/_mixins.scss
yeelan0319: Make sense, I will move them to other components respectively.

packages/mdc-typography/_mixins.scss
yeelan0319: There is a line on spec said:

> “Button style (Medium 14sp, all caps) is used for all buttons. Button text should be all caps in languages that have capitalization. ”

So I think we should enforce capitalization but as you suggested not `text-align` center, etc.., SG?There is a line on spec said:

> “Button style (Medium 14sp, all caps) is used for all buttons. Button text should be all caps in languages that have capitalization. ”

So I think we should enforce capitalization but as you suggested not `text-align` center, etc.., SG?

lynnjepsen: can we just set text-decoration: none on all of the typography styles? We never really want decoration, button isn't special in this way.

And can we put text-transform: none for the other typography styles? Instead of having this if-statement just for button.

packages/mdc-typography/_variables.scss
lynnjepsen: https://github.com/material-components/material-components-web/blob/master/packages/mdc-button/mdc-button.scss#L42

font-size was purposely set to 14px...but I cannot figure out why it wasn't .0975rem in the first place, so maybe this "fixes" it.

packages/mdc-typography/_mixins.scss
yeelan0319: I am not sure if `text-decoration: none` will work for every other typography, for example links, I don't think we are opinionated that links cannot use `mdc-typography`, so I am a bit hesitate to do that.

So it seems like what you want to achieve here is just make button less special, how about use `text-decoration:  inherit` and `text-transform: inherit` then?

yeelan0319:
#1064 Reopened

PR #1041, opened 2017-07-28
fix(ripple): Feature-detect buggy Edge behavior for custom properties

### LINKED ISSUE #663: Edge: JS Button problem

kfranqueiro:
This adds feature detection for buggy var() behavior in Edge WRT pseudo selectors, and will not upgrade ripples if the buggy behavior is found.

Because this adds quite a bit of logic (including DOM operations) to the function, this PR also changes the function to cache its result once a true or false result is obtained, much like we do for various other util functions.  This requires adding an optional force parameter to the function for tests to use.

This PR also updates tests and adds one to test the new case, and adds a section to the README.

Fixes #663.

### DIFF
--- a/packages/mdc-ripple/README.md
+++ b/packages/mdc-ripple/README.md
@@ -428,11 +428,11 @@ External frameworks and libraries can use the following utility methods when int
 
 #### util.supportsCssVariables(windowObj, forceRefresh = false) => Boolean
 
-Determine whether the current browser supports CSS variables (custom properties).
+Determine whether the current browser supports CSS variables (custom properties). This function caches its result; `forceRefresh` will force recomputation, but is used mainly for testing and should not be necessary in normal use.
 
 #### util.applyPassive(globalObj = window, forceRefresh = false) => object
 
-Determine whether the current browser supports passive event listeners, and if so, use them.
+Determine whether the current browser supports passive event listeners, and if so, use them. This function caches its result; `forceRefresh` will force recomputation, but is used mainly for testing and should not be necessary in normal use.
 
 #### getMatchesProperty(HTMLElementPrototype) => Function
 
--- a/packages/mdc-ripple/README.md
+++ b/packages/mdc-ripple/README.md
@@ -31,6 +31,7 @@ path: /catalog/ripples/
     - [Using a sentinel element for a ripple](#using-a-sentinel-element-for-a-ripple)
     - [Keyboard interaction for custom UI components](#keyboard-interaction-for-custom-ui-components)
     - [Specifying known element dimensions](#specifying-known-element-dimensions)
+  - [Caveat: Edge](#caveat-edge)
   - [Caveat: Safari](#caveat-safari)
   - [Caveat: Theme Custom Variables](#caveat-theme-custom-variables)
   - [The util API](#the-util-api)
@@ -45,8 +46,8 @@ In order to function correctly, MDC Ripple requires a _browser_ implementation o
 
 Because we rely on scoped, dynamic CSS variables, static pre-processors such as [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) will not work as an adequate polyfill ([...yet?](https://github.com/postcss/postcss-custom-properties/issues/32)).
 
-[Most modern browsers](http://caniuse.com/#feat=css-variables) support CSS variables, so MDC ripple will work just fine. In other cases, MDC ripple will _still work_ if you include it in your codebase. It will simply check if CSS variables are supported upon initialization and if they aren't, gracefully exit. The only exception to this rule is Safari, which does support CSS variables
-but unfortunately ripples are disabled for (see [below](#caveat-safari) for an explanation).
+[Most modern browsers](http://caniuse.com/#feat=css-variables) support CSS variables, so MDC ripple will work just fine. In other cases, MDC ripple will _still work_ if you include it in your codebase. It will simply check if CSS variables are supported upon initialization and if they aren't, gracefully exit. The only exceptions to this rule are Edge and Safari, which do support CSS variables
+but unfortunately ripples are disabled for (see [below](#caveat-edge) for an explanation).
 
 
 ## Installation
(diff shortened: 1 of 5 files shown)


packages/mdc-ripple/README.md
```diff
@@ -45,8 +46,8 @@ In order to function correctly, MDC Ripple requires a _browser_ implementation o
 
 Because we rely on scoped, dynamic CSS variables, static pre-processors such as [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) will not work as an adequate polyfill ([...yet?](https://github.com/postcss/postcss-custom-properties/issues/32)).
 
-[Most modern browsers](http://caniuse.com/#feat=css-variables) support CSS variables, so MDC ripple will work just fine. In other cases, MDC ripple will _still work_ if you include it in your codebase. It will simply check if CSS variables are supported upon initialization and if they aren't, gracefully exit. The only exception to this rule is Safari, which does support CSS variables
-but unfortunately ripples are disabled for (see [below](#caveat-safari) for an explanation).
+[Most modern browsers](http://caniuse.com/#feat=css-variables) support CSS variables, so MDC ripple will work just fine. In other cases, MDC ripple will _still work_ if you include it in your codebase. It will simply check if CSS variables are supported upon initialization and if they aren't, gracefully exit. The only exceptions to this rule are Edge and Safari, which do support CSS variables
```
lynnjepsen: I feel like this whole section is trying to remind users how many browsers we support "just fine". Users assume we support things just fine. They want documentation for what we do NOT support. So I would put just this.

Edge and Safari, even though they do [support CSS variables](http://caniuse.com/#feat=css-variables), do not support MDC Ripple. See [Safari Caveat](#caveat-safari) and [Safari Edge](#caveat-edge) for an explanation.

packages/mdc-ripple/README.md
```diff
@@ -380,6 +381,17 @@ this.ripple_ = new MDCRippleFoundation({
...
 ```
 
+## Caveat: Edge
+
+> TL;DR ripples are disabled in Edge because of issues with its support of CSS variables in pseudo elements.
+
+Edge introduced CSS variables in version 15.  Unfortunately, there are
+[issues](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11495448/)
+involving its implementation for `::before` and `::after` pseudo-element selectors which cause ripples to
+behave incorrectly.  We feature-detect one of these bugs and do not upgrade ripples if the bug is observed.
```
lynnjepsen: "one of these bugs"....which one? Or is there only one "issue", in which case update the "issues" word in the second sentence.

lynnjepsen: I continue to not quite understand what "upgrade ripple" is. Are there two types of ripples? Or are we upgrading from no ripple to a ripple?

packages/mdc-ripple/util.js
```diff
@@ -15,13 +15,21 @@
  */
 
 /** @private {boolean|undefined} */
+let supportsCssVariables_;
```
lynnjepsen: document what this variable is.

packages/mdc-ripple/util.js
```diff
@@ -15,13 +15,21 @@
...
+
+/** @private {boolean|undefined} */
 let supportsPassive_;
 
 /**
  * @param {!Window} windowObj
+ * @param {boolean=} forceRefresh
```
lynnjepsen: document what this argument is

packages/mdc-ripple/util.js
```diff
@@ -34,7 +42,32 @@ export function supportsCssVariables(windowObj) {
...
     windowObj.CSS.supports('color', '#00000000')
   );
-  return explicitlySupportsCssVars || weAreFeatureDetectingSafari10plus;
+
+  if (!explicitlySupportsCssVars && !weAreFeatureDetectingSafari10plus) {
+    supportsCssVariables_ = false;
+    return supportsCssVariables_;
+  }
+
+  // Detect versions of Edge with buggy var() support
```
lynnjepsen: Put this whole section below in a separate function, which will make it easier to document the various arguments

lynnjepsen:
I updated the branch with master to kick off Travis CI again. I don't think the failures have to do with your PR, but lets double check with this run.

packages/mdc-ripple/README.md
kfranqueiro: "issues" was based on the wording in the linked report. I'm updating the wording to hopefully resolve the ambiguity though.

kfranqueiro: I re-read the README and realized that "initialize" is the word I probably wanted here; `upgraded` is the suffix on the class that's applied when initialization occurs on a browser that supports Ripple.

packages/mdc-ripple/util.js
kfranqueiro: We have a very similar argument in the util function below this which isn't documented, and I was under the impression it was normal for us to not document these within the jsdoc comments added for Closure?

I did realize that I hadn't added this argument to the README so I added it there.  We don't document it there either though, and I wonder if we should - forceRefresh is only really needed for testing purposes, it shouldn't be needed for normal use.

packages/mdc-ripple/util.js
kfranqueiro: I was under the impression we don't include docs in our comments for the closure compiler, and this variable is very similar in function to the one immediately below it which is also not documented.  Should I be documenting both?

amsheehan:
@kfranqueiro if you spin on ie11 in VirtualBox you can run the tests locally in ie11 and debug

kfranqueiro:
I'm unable to reproduce the `this.skip` errors in IE11 via modern.ie or on my Win10 laptop using `npm run test:watch`, so I'm not sure what's going on in the CI environment. The most I've noticed in mocha's releases is there was a fix for `skip` and `only` in IE8/9 but I haven't found anything mentioning IE11.

kfranqueiro:
Okay, I was able to repro (used faulty grep before) and realized the IE11 issue wasn't due to IE11, it was due to us using arrow functions for all of our tests, which means we don't get the object mocha normally runs tests in the context of which provides `this.skip`.  Pushed a commit to fix that; now we'll see if CI passes.

packages/mdc-ripple/util.js
lynnjepsen: Yes document both. These were never documented, and I still don't understand what they are trying to accomplish. We added annotation comments for the compiler...but it wasn't done by a core member of the team...he was just trying to get the compiler to pass.

"supportsCssVariables" -> "whether or not the current runtime execution supports custom css properties"? or "whether or not the current component supports the var syntax"? or something else?

"supportsPassive" -> I mean, passive what? Passive ripple? Passive browser?

packages/mdc-ripple/util.js
lynnjepsen: >  forceRefresh is only really needed for testing purposes

OMG yes, we should def document that its only for testing purpose then! Why is it only for testing purposes??

packages/mdc-ripple/util.js
```diff
@@ -15,13 +15,47 @@
  */
 
 /** @private {boolean|undefined} */
+let supportsCssVariables_;
+
+/** @private {boolean|undefined} */
 let supportsPassive_;
 
 /**
  * @param {!Window} windowObj
```
lynnjepsen: nit: move those two inline commments into the jsdoc style compiler comments

packages/mdc-ripple/util.js
kfranqueiro: The unit tests all run in a single environment and need to re-test that function several times with different window object mocks to exercise the different code paths, so they need a way to force it to re-evaluate even if there's a cached result.  Meanwhile for typical use the cached result is correct because the page is always in the same browser, and it doesn't make sense to re-evaluate every time you instantiate a ripple (there could be a lot on a page).

PR #1064, opened 2017-08-02
feat(typography): Add button style to typography

yeelan0319:
Reopen #1028


piotr-cz:
There should be an update for [demo](https://material-components-web.appspot.com/typography.html), too

PR #1068, opened 2017-08-03
fix(toolbar): Wrong placement of last icon when there is a menu

### LINKED ISSUE #1026: Toolbar: Wrong placement of last icon when there is a menu

> 0.16.0

> All browsers

### What OS are you using?

> Windows 7

> 1. Have a toolbar [like here](http://material-components-web.appspot.com/toolbar/default-toolbar.html)
> 2. Add a simple menu to the most right icon
> 3. Observe the right padding (placement) of the last icon

> It should have a `padding-right: 16px;`

> It has a `padding-right: 8px;` (like normal icons)

> This is caused by `mdc-toolbar__icon:last-child`. The toolbar icon isn't the last child because there is a simple menu after it. A simple fix would be to use `last-of-type` instead of `last-child`.
Would you like a PR to fix this?

touficbatache:
Fixes and closes #1026 @yeelan0319


kfranqueiro:
Should we have an example of this added to the toolbar demos?  Maybe addressable separately but IIUC there's nothing OOTB that demonstrates the issue/fix right now.

touficbatache:
I can do that, so from what I unrderstood I should add an example with the menu, correct?

kfranqueiro:
Yes, specifically what I had in mind was add another example to the toolbar demos that contains a menu on the last icon.  It should then be easy to test with/without the CSS change to see the fix.

touficbatache:
I made the necessary changes! @yeelan0319 @kfranqueiro

PR #1076, opened 2017-08-04
fix: Make CSS custom properties compatible with sass-spec 3.5

### LINKED ISSUE #1075: theme 0.1.6 CSS custom properties not compatible with sass 3.5

mdc-theme 0.1.6

Use sass (ruby sass) latest release (3.5) to compile `mdc-theme.scss`

```
$ gem install sass -v 3.5
Successfully installed sass-3.5.0
$ sass _3.5_ packages/mdc-theme/mdc-theme.scss
```

That it outputs normally theme CSS custom properties, i.e.

```
$ sass _3.5_ packages/mdc-theme/mdc-theme.scss | grep "\--mdc-theme-primary:"
  --mdc-theme-primary: #3f51b5;
```

Leaves theme CSS custom properties unprocessed

```
$ sass _3.5_ packages/mdc-theme/mdc-theme.scss | grep "\--mdc-theme-primary:"
  --mdc-theme-primary: map-get($mdc-theme-property-values, $style);
```

The reason why this issue isn't affecting MDC demos yet is because `node-sass` (via `libsass`) is still in the process of catching up with latest changes in `sass-spec`.

But it seems that libsass will soon release `3.5` with the same behavior for CSS custom properties. See https://github.com/sass/libsass/issues/2076.

Once the issue linked above is addressed, `libsass` released and `node-sass` updated, the demos will suffer from the same issue and the theme properties won't be processed.

grzuy:
Fixes #1075

### DIFF
--- a/packages/mdc-dialog/mdc-dialog.scss
+++ b/packages/mdc-dialog/mdc-dialog.scss
@@ -24,7 +24,7 @@
 $mdc-dialog-dark-theme-bg-color: #303030 !default;
 
 :root {
-  --mdc-dialog-dark-theme-bg-color: $mdc-dialog-dark-theme-bg-color;
+  --mdc-dialog-dark-theme-bg-color: #{$mdc-dialog-dark-theme-bg-color};
 }
 
 // postcss-bem-linter: define dialog
--- a/packages/mdc-drawer/permanent/mdc-permanent-drawer.scss
+++ b/packages/mdc-drawer/permanent/mdc-permanent-drawer.scss
@@ -22,7 +22,7 @@ $mdc-permanent-drawer-width: 240px;
 $mdc-permanent-drawer-dark-theme-bg-color: #212121 !default;
 
 :root {
-  --mdc-permanent-drawer-dark-theme-bg-color: $mdc-permanent-drawer-dark-theme-bg-color;
+  --mdc-permanent-drawer-dark-theme-bg-color: #{$mdc-permanent-drawer-dark-theme-bg-color};
 }
 
 .mdc-permanent-drawer {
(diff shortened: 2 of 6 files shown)


Garbee:
FTR here if I recall correctly when we started MCW we specifically opted to only support `nodesass` builds. If it worked with regular ruby Sass, great. Otherwise it wasn't a concern that we were too focused on. In an attempt to keep the support focus as tight as possible.

@traviskaufman if he has time may remember what the exact decision was on that. I could be incorrectly remembering all the details.

kfranqueiro:
I don't think this PR is out of line with that goal?  IIUC the point of this is to future-proof our code so that when libsass and thus node-sass inevitably catch up to 3.5, we won't run into this issue.  See the libsass issue linked in the issue this addresses.

Garbee:
Ah, Ok. I see now. I got mixed up in what Sass was doing upstream. So this is making it *future-proof* vs fixing a bug running it with older Sass versions.

SGTM.

grzuy:
@acdvorak Good catch :-)

Updated PR to also future-proof `dialog`, `drawer`, `layout-grid` and `toolbar`.
Also updated PR title given it's no longer specific to theme package.

grzuy:
Thank you!

kevinarrieta:
Does anybody know when the npm package would be updated with this commit? Thanks!

grzuy:
@kevinarrieta I am not a maintainer so can't say for sure, but based on previous releases i think they do release every two weeks on Mondays. So i suspect the next one could be Monday 21st.

kevinarrieta:
Thank you @grzuy. I'm having the same issue using Jekyll, so thank you for taking the time on fixing this!

PR #1077, opened 2017-08-05
chore(drawer): Update copyright notice in SCSS files

acdvorak:
Update the year from 2016 to 2017 and use the same comment style found in all other SCSS files (`// ...` instead of `/** ... */`).


Garbee:
Copyright year should not be arbitrarily changed. Unless a *significant* portion of the code was modified within a given year, the copyright should be left alone. Even if it were updated it should be a comma separated list of years. For example, `2016,2017` and/or a range where needed to cover a small span of them.

+1 to the comment style change to keep it consistent. But IMO the dates should be left alone as the majority of these files predate the MCW repo and that work was done in 2016. While it has seen a few updates and tweaks this year I'm personally not seeing enough to justify a copyright year modification.

acdvorak:
I totally agree that copyright year should not be changed arbitrarily 🙂

My intention was to correct what I saw as a typo: Both files were added to the MCW repo in 2017, and the analogous MDL styles ([`_layout.scss:84`](https://github.com/google/material-design-lite/blob/245f19d74fe86eb3ca64bad6819c9784c748fe1c/src/layout/_layout.scss#L84) and [`_variables.scss:189`](https://github.com/google/material-design-lite/blob/3b749f1455f477927225723bd662ae9b51a8ca66/src/_variables.scss#L189)) bear little resemblance to the MCW code, AFAICT.

Granted, both versions use `transform: translateX()` and a `cubic-bezier()` `transition-timing-function`, but A) they have different values, B) the code is organized very differently, and C) there's literally no other way to implement a sliding drawer if you care about performance (which we surely do 🙂). So it's not clear to me that these files are _derivative works_ of MDL; they feel more like separate implementations that were _informed_ by MDL.

However, being new to the team, I'm definitely not as familiar with the MCW or MDL codebases (or their histories) as you are, so I **sincerely appreciate** your deep experience, sharp eyes, and insightful comments on stuff like this 😀

I will follow up with some folks on our side to get a broader perspective, but in the mean time, does the above persuade you, or do you still feel that 2016 is the most accurate year of first publication for these files?

Garbee:
This is a Google project. I simply throw up a yellow card when I see something fishy with the licensing going on. For a complete solid answer on whether or not it should be modified, that would be a question to ask the lawyers.

Given the changes though (even though some foundation is still there) a `2016,2017` update seems reasonable to me.

acdvorak:
Very good points, @Garbee - thanks! And keep those yellow cards coming 😀

PR #1087, opened 2017-08-07
fix(dialog): Layout footer buttons' ripples after open transition ends

### LINKED ISSUE #464: Ripples incorrectly laid out for dialog buttons

kfranqueiro:
BREAKING CHANGE: Adds a new adapter method, `layoutFooterRipples`, to allow the foundation to
communicate with ripples when the dialog's opening transition ends.

Fixes #464.


kfranqueiro:
Alex realized that the reason codecov detected decreased coverage is because I'm testing the foundation but not specifically the new calls to MDCRipple#layout.  I'll have to see if I can make a component test to do that.

touficbatache:
Is this being worked on? @amsheehan @kfranqueiro

kfranqueiro:
Hey @touficbatache, thanks for checking in; sorry I hadn't gotten back to this as quickly as I'd hoped, but I'm still hoping to get this sorted out this week.

kfranqueiro:
I added a commit which adds a component test that stubs `MDCRipple#layout` in order to verify the default adapter calls it. However, I had to disable the `mocha/no-hooks` rule in order to do this, since I have to stub a prototype-level method, which means I need to be sure to tear it down properly lest the stubbed method hang around and affect other tests. I'm willing to discuss alternate approaches, but I'm also inclined to think that sometimes a proper `after` or `afterEach` is necessary in tests.

PR #1088, opened 2017-08-07
chore(package): Update stylelint-scss to the latest version

lynnjepsen:
Fixed deprecated rule, and updated mdc-textfield.scss to pass linter


package.json
```diff
@@ -86,7 +86,7 @@
     "stylelint": "^8.0.0",
     "stylelint-config-standard": "^17.0.0",
     "stylelint-order": "^0.6.0",
-    "stylelint-scss": "^1.5.1",
+    "stylelint-scss": "^2.0.0",
```
kfranqueiro: Do we want to reference `^2.0.1` instead as greenkeeper pointed out? https://github.com/kristerkari/stylelint-scss/releases/tag/2.0.1

PR #1089, opened 2017-08-07
refactor(button): Remove CSS that override styles with the same styles

lynnjepsen:


packages/mdc-button/mdc-button.scss
```diff
@@ -130,10 +127,6 @@
   &--primary {
     @include mdc-theme-prop(color, primary);
 
-    @include mdc-theme-dark(".mdc-button") {
```
acdvorak: Removing this changes the color of the `default with primary` button in dark mode.

### Before (current prod behavior):
mdc-web-pr-1089-before

### After (this PR):
mdc-web-pr-1089-after

packages/mdc-button/mdc-button.scss
```diff
@@ -149,10 +142,6 @@
   &--accent {
     @include mdc-theme-prop(color, accent);
 
-    @include mdc-theme-dark(".mdc-button") {
```
acdvorak: Removing this changes the color of the `default with accent` button in dark mode.

### Before (current prod behavior):
mdc-web-pr-1089-before

### After (this PR):
mdc-web-pr-1089-after

Garbee:
as @acdvorak found, the color overrides were an inherit design decision when the component was built. They are what allowed for the most seamless API for developers to work with without multiplying the number of classes that needed to exist.

yeelan0319:
Yep. @lynnjepsen 
`@include mdc-ripple-base;` will incorporate proper ripple opacity behavior and couldn't be removed at each style level as it will result in incorrect override if you take a closer look. I did a little bit clean up in #1074 , but still you cannot remove those lines

lynnjepsen:
Hmm I see the issue with coloring and dark mode. I put that coloring CSS back in place for primary and accent buttons.

But I don't understand why we need `@include mdc-ripple-base` 4 times in mdc-button.scss. I checked the demos and I can't see any change,

PR #1053, opened 2017-08-01
Addresses 3 issues

Teoxoy:
https://github.com/material-components/material-components-web/issues/999 - Partially completed (I need answers)
https://github.com/material-components/material-components-web/issues/1018 - Added new feature to MDCTextfield
https://github.com/material-components/material-components-web/issues/1049 - Updated READMEs to fix es2015 imports
Sorry for referencing 3 issues in 1 PR.

Issue #1018, opened 2017-07-26
Custom validity on MDCTextfield


DBartsch:
So, I'm having a MDCTextfield that is initialized automatically.

I want to validate it by myself and this can't be done with any pattern or max-/minlength attribute.

As far as I understood the documentation all that's to do is calling `setCustomValidity` with any string as a parameter, but after doing so the field isn't marked as invalid. getting the validation properties of the HTMLInputElement also doesn't show anything.

Am I missing something out or is something catching and reverting the invalid status?

Just giving the `.mdc-textfield` the class `mdc-textfield--invalid` is also not working because the field may get and lose it's focus afterwards again (i tested it without any focus changes, too).

Teoxoy:
I would also like this feature to be implemented. Looking at the foundation class the validity is being set on blur. A better implementation would be to have a property 'setValidity' that you can set yourself and another variable 'useNativeValidityChecking' that will be true by default and when setValidity will be called it will become false. So this will not break current implementations of the textfield.
I will do a commit in my fork. And I would like to hear the opinion of a member of the project.

Teoxoy:
The valid variable is just a setter. The only way to implement a getter in the foundation would require the adapter to have a function hasClass and check for 'this.adapter_.hasClass(INVALID);'.

Teoxoy:
Should I submit a PR ?

yeelan0319:
Hi @DBartsch! Could you provide a little bit description about what you are tring to do here? It would be great if you can provide a codepen to reproduce the error so that we can nail it down. Thanks!

PR #1092, opened 2017-08-07
Feature/textfield custom validity

Teoxoy:
https://github.com/material-components/material-components-web/issues/1018 - Added new feature to MDCTextfield

#1053 - moved here


packages/mdc-textfield/README.md
```diff
@@ -153,9 +153,9 @@ by HTML5's form validation API.
 </div>
 ```
 
-An input's validity is checked via `checkValidity()` on blur, and the styles are updated
-accordingly. When using the `required` attribute, an asterisk will be automatically appended to the
-label text, as per the spec.
+By default an input's validity is checked via `checkValidity()` on blur, and the styles are updated
+accordingly. The validity can also be set via the MDCTextfield.valid variable. When using the `required`
```
lynnjepsen: Try this:

Set the MDCTextfield.valid variable to set the input's validity explicitly. MDC Textfield automatically appends an asterisk to the label text if the `required` attribute is set.

packages/mdc-textfield/foundation.js
```diff
@@ -99,7 +100,7 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
   }
 
   deactivateFocus_() {
-    const {FOCUSED, INVALID, LABEL_FLOAT_ABOVE} = MDCTextfieldFoundation.cssClasses;
+    const {FOCUSED, LABEL_FLOAT_ABOVE} = MDCTextfieldFoundation.cssClasses;
     const input = this.getNativeInput_();
     const isValid = input.checkValidity();
```
lynnjepsen: can you change the implementation of the `checkValidity` method to check `useNativeValidityChecking_`?

It's odd that `deactivateFocus_` would both `checkValidity` and `changeValidity`

packages/mdc-textfield/foundation.js
```diff
@@ -167,4 +175,9 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
       badInput: false,
     };
   }
+
+  setValid(isValid) {
+    this.useNativeValidityChecking_ = false;
```
lynnjepsen: I think there should also be a way to go back to using native validity checking

packages/mdc-textfield/foundation.js
```diff
@@ -57,6 +57,7 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
     this.inputBlurHandler_ = () => this.deactivateFocus_();
     this.inputInputHandler_ = () => this.autoCompleteFocus_();
     this.inputKeydownHandler_ = () => this.receivedUserInput_ = true;
+    this.useNativeValidityChecking_ = true;
```
lynnjepsen: the word "native" here is a little misleading. Makes it sound like its happening in an Android/iOS app, not web.

I think what you mean is that MDC Textfield has some default validity checking, and this variable overrides that.

why not `useCustomValidityChecking_` and flip true to false?

kfranqueiro: "Native" here refers to depending on the browser's native HTML and DOM features for form validation; I think that's reasonable, but flipping the variable name is also reasonable.

Teoxoy: Yes, that's why I used "Native". Should I leave it as is or modify it to "Custom" ?

kfranqueiro: IMO "Custom" and defaulting to `false` might be easier to understand anyway. i.e. the default behavior is no custom validity (use native HTML validity checking).  The non-default is adding something moreso than removing something.

packages/mdc-textfield/foundation.js
Teoxoy: `checkValidity` is a native HTML/JS function for validating the input.

Teoxoy: I will move the input check in the if block just for performance reasons like so:
`if (!this.useCustomValidityChecking_) {
  this.changeValidity_(input.checkValidity());
}`

packages/mdc-textfield/foundation.js
Teoxoy: I'm not sure if someone would ever use that feature. Is it worth having another public variable to change that ?

packages/mdc-textfield/foundation.js
kfranqueiro: I may be missing something, but shouldn't we not even compute `isValid` if not using native validity checking? Shouldn't that be moved inside the `if` as well?

packages/mdc-textfield/foundation.js
kfranqueiro: I'm curious of a use-case for that as well. I would think it's not typical (and potentially confusing) to change validation logic of a particular component instance on-the-fly, so you're either going to use this method directly to control valid state yourself, or you're going to rely on native validity checking.

Teoxoy:
I stated this on the issue #1018 : 

> The valid variable is just a setter. The only way to implement a getter in the foundation would require the adapter to have a function hasClass and check for this.adapter_.hasClass(INVALID); or an additional variable that keeps track of the validity.

Would it be worth to implement a getter ?
IMO if the user sets the variable, he will not need to check it.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -246,19 +258,34 @@ test('on blur does not remove mdc-textfield__label--float-above if input has a v
   td.verify(mockAdapter.removeClassFromLabel(cssClasses.LABEL_FLOAT_ABOVE), {times: 0});
 });
 
-test('on blur removes mdc-textfield--invalid if input.checkValidity() returns true', () => {
+test('on blur removes mdc-textfield--invalid if setValid was not set and input.checkValidity() returns true', () => {
```
lynnjepsen: if custom validity is false

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -246,19 +258,34 @@ test('on blur does not remove mdc-textfield__label--float-above if input has a v
...
   const {mockAdapter, blur} = setupBlurTest();
   blur();
   td.verify(mockAdapter.removeClass(cssClasses.INVALID));
 });
 
-test('on blur adds mdc-textfied--invalid if input.checkValidity() returns false', () => {
+test('on blur adds mdc-textfied--invalid if setValid was not set and input.checkValidity() returns false', () => {
```
lynnjepsen: if custom validity is false

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -246,19 +258,34 @@ test('on blur does not remove mdc-textfield__label--float-above if input has a v
...
   const {mockAdapter, blur, nativeInput} = setupBlurTest();
   nativeInput.checkValidity = () => false;
   blur();
   td.verify(mockAdapter.addClass(cssClasses.INVALID));
 });
 
+test('on blur does not remove mdc-textfield--invalid if setValid was set and input.checkValidity() returns true', () => {
```
lynnjepsen: if custom validity is true

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -246,19 +258,34 @@ test('on blur does not remove mdc-textfield__label--float-above if input has a v
...
+  const {foundation, mockAdapter, blur} = setupBlurTest();
+  foundation.setValid(false);
+  blur();
+  td.verify(mockAdapter.removeClass(cssClasses.INVALID), {times: 0});
+});
+
+test('on blur does not add mdc-textfied--invalid if setValid was set and input.checkValidity() returns false', () => {
```
lynnjepsen: if custom validity is true

lynnjepsen:
> Would it be worth to implement a getter ? IMO if the user sets the variable, he will not need to check it.

I think you're right. No need to check the variable if the user sets it.

lborgman:
I just wanted to read the value of ".valid". It took me some time to find out why it didn't work. 😐

I think it's quite confusing to have a setter and a non-functional getter. (I still have not read the spec, but I found this )

My usecase is just something similar to a form validation before submit.

Issue #1025, opened 2017-07-27
MDCSelect object doesn't accept select element as root


Teoxoy:
At the end of the README file for the select package there is an example of how you can switch between selects for better cross-device UX. In the example the select element is being attached to an MDCSelect obj, that would not work because the initialize function tries to create a menu with a non existent root element.
Possible fixes: Update README to show a working example or allow and implement select element as root for MDCSelect

kfranqueiro:
Sorry for being late to reply to this. We've made updates to MDC Select's DOM structure more recently and updated the README accordingly.

PR #1097, opened 2017-08-08
feat(select): Add new UX styles and behavior to select

### LINKED ISSUE #674: Implement text field boxes for mdc-select
Similar to #673 but for `mdc-select`.

amsheehan:
resolves #674


amsheehan:
BREAKING CHANGE: Change to Adapter methods, adding methods to add/remove classnames and attributes from mdc-select__label and mdc-select__bottom-line

packages/mdc-select/mdc-select.scss
```diff
@@ -93,12 +88,165 @@
...
+      transform: translateY(-40%) scale(.75, .75);
+    }
+  }
+
   &__selected-text {
+    display: flex;
+    align-items: flex-end;
+    margin-bottom: 6px;
+    padding-right: 10px;
+    padding-left: 16px;
```
acdvorak: It this supposed to be indented relative to the `__label`?

image

It looks a little strange to me, but I'm not sure where the relevant spec is.

demos/select.html
```diff
@@ -179,37 +180,42 @@ <h2 class="mdc-typography--title">CSS Only</h2>
     <script>
       (function() {
         var MDCSelect = mdc.select.MDCSelect;
-        var root = document.getElementById('js-select');
-        var currentlySelected = document.getElementById('currently-selected');
+        var heroSelect = document.getElementById('hero-js-select');
```
kfranqueiro: Not in demo pages, which don't get transpiled and still need to work in all browsers we support.

packages/mdc-select/README.md
```diff
@@ -194,19 +210,18 @@ seamless, un-invasive experience in browsers where a native select may be more a
...
-  <optgroup label="Fruits">
-    <option value="apple">Apple</option>
-    <option value="oranges">Orange</option>
-    <option value="banana">Banana</option>
-  </optgroup>
-  <option value="dairy">Milk, Yogurt, and Cheese</option>
-  <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
-  <option value="fats">Fats, Oils, and Sweets</option>
-</select>
+<div class="mdc-select">
```
kfranqueiro: You might want to update the wording in the preceding paragraph, since I guess the `mdc-select` class does not work directly on the native `select` element, but rather wrapped around it.

packages/mdc-select/README.md
```diff
@@ -331,9 +346,16 @@ within `componentDidUpdate`.
...
+| `addClassToLabel(className: string) => void` | Adds a class to the label |
+| `removeClassFromLabel(className: string) => void` | Removes a class from the label |
+| `addClassToBottomLine(className: string) => void` | Adds a class to the bottom line |
+| `removeClassFromBottomLine(className: string) => void` | Removes a class from the bottom line |
+| `setBottomLineAttr(attr: string, value: string) => void` | Adds an attribute to the bottom line |
 | `setAttr(attr: string, value: string) => void` | Sets attribute `attr` to value `value` on the root element. |
 | `rmAttr(attr: string) => void` | Removes attribute `attr` from the root element. |
 | `computeBoundingRect() => {left: number, top: number}` | Returns an object with a shape similar to a `ClientRect` object, with a `left` and `top` property specifying the element's position on the page relative to the viewport. The easiest way to achieve this is by calling `getBoundingClientRect()` on the root element. |
+| `registerPointerDownHandler(evtType: string, handler: EventListener) => void`| Adds an event listener for pointer down interactions to the root element |
```
kfranqueiro: Maybe I lack historical context, but what's the reason for increasing the adapter API surface for this? It looks like we could just reuse `registerInteractionHandler` to register `touchstart` and `pointerdown` handlers and likewise for deregistering.

packages/mdc-select/foundation.js
```diff
@@ -174,10 +196,12 @@ export default class MDCSelectFoundation extends MDCFoundation {
 
     let maxTextLength = 0;
     for (let i = 0, l = this.adapter_.getNumberOfOptions(); i < l; i++) {
+      const selectBoxAddedPadding = 26;
```
kfranqueiro: Also, given that we're using `getComputedStyle` in this function already, is it feasible to use that here rather than risk skew between SCSS and JS?

packages/mdc-select/foundation.js
```diff
@@ -216,9 +253,14 @@ export default class MDCSelectFoundation extends MDCFoundation {
     this.adapter_.setMenuElStyle('transform-origin', `center ${itemOffsetTop}px`);
   }
 
-  close_() {
+  close_(removeFloatingLabelClass) {
```
kfranqueiro: Nit: rename param to `shouldRemoveFloatingLabelClass`, more indicative of being a boolean

packages/mdc-select/foundation.js
```diff
@@ -85,14 +94,17 @@ export default class MDCSelectFoundation extends MDCFoundation {
     this.displayViaKeyboardHandler_ = (evt) => this.handleDisplayViaKeyboard_(evt);
     this.selectionHandler_ = ({detail}) => {
       const {index} = detail;
-      this.close_();
+      const removeFloatingLabelClass = false;
```
kfranqueiro: Does this const exist purely for code style? It seems unnecessary since it's used only once in close proximity in each case.  If we do keep it, I'd suggest renaming to `shouldRemoveFloatingLabelClass`.

packages/mdc-select/index.js
```diff
@@ -69,29 +71,40 @@ export class MDCSelect extends MDCComponent {
...
+      addClassToLabel: (className) => this.label_.classList.add(className),
+      removeClassFromLabel: (className) => this.label_.classList.remove(className),
+      addClassToBottomLine: (className) => this.bottomLine_.classList.add(className),
+      removeClassFromBottomLine: (className) => this.bottomLine_.classList.remove(className),
+      setBottomLineAttr: (attr, value) => this.bottomLine_.setAttribute(attr, value),
       setAttr: (attr, value) => this.root_.setAttribute(attr, value),
       rmAttr: (attr, value) => this.root_.removeAttribute(attr, value),
-      computeBoundingRect: () => this.root_.getBoundingClientRect(),
+      computeBoundingRect: () => this.surface_.getBoundingClientRect(),
+      registerPointerDownHandler: (evtType, handler) => this.root_.addEventListener(evtType, handler),
```
kfranqueiro: `evtType` is inconsistent with `type` used for existing `[de]registerInteractionHandler` immediately below.  (Also see question I raised in README about why these need to exist.)

packages/mdc-select/index.js
```diff
@@ -69,29 +71,40 @@ export class MDCSelect extends MDCComponent {
...
-        this.root_.tabIndex = 0;
+        this.menuEl_.tabIndex = 0;
       },
       makeUntabbable: () => {
-        this.root_.tabIndex = -1;
+        this.menuEl_.tabIndex = -1;
       },
-      getComputedStyleValue: (prop) => window.getComputedStyle(this.root_).getPropertyValue(prop),
-      setStyle: (propertyName, value) => this.root_.style.setProperty(propertyName, value),
+      getComputedStyleValue: (prop) => window.getComputedStyle(this.surface_).getPropertyValue(prop),
```
kfranqueiro: If we're changing what element `getComputedStyleValue` and `setStyle` operate on, we should update their descriptions in the README which say they operate on the root element.

Also, does changing the target of `getComputedStyleValue` have any adverse effects on existing `resize` code, or is that code precisely the point of changing this?

test/unit/mdc-select/foundation.test.js
```diff
@@ -185,8 +186,8 @@ test('#resize resizes the element to the longest-length option', () => {
   foundation.init();
   foundation.resize();
   assert.equal(ctx.font, '16px Roboto');
-  // ceil(letter-spacing * 'longest'.length + longest measured width)
-  const expectedWidth = Math.ceil((2.5 * 7) + Math.max(...widths));
+  // ceil(letter-spacing * 'longest'.length + longest measured width + extra padding)
+  const expectedWidth = Math.ceil((2.5 * 7) + Math.max(...widths) + 26);
```
kfranqueiro: If we expose 26 as a constant as Andy suggested, could we reuse it here?

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -214,17 +278,17 @@ test('adapter#focus focuses on the root element', () => {
...
-  const {component, fixture} = setupTest();
-  fixture.tabIndex = 0;
+  const {component, menuEl} = setupTest();
+  menuEl.tabIndex = 0;
   component.getDefaultFoundation().adapter_.makeUntabbable();
-  assert.equal(fixture.tabIndex, -1);
+  assert.equal(menuEl.tabIndex, -1);
 });
 
 test('adapter#getComputedStyleValue gets the computed style value of the prop from the root element', () => {
```
kfranqueiro: I'm confused; shouldn't this test fail and need to be changed, since `getComputedStyleValue` was changed to look at the surface rather than root element?

packages/mdc-select/README.md
```diff
@@ -403,12 +425,16 @@ First, wrap both a custom select and a native select within a wrapper element, l
 <div class="select-manager">
```
kfranqueiro: Can we test that this example still actually works?  (See #1025)

demos/select.html
```diff
@@ -97,37 +99,30 @@
...
                 </li>
-                <li class="mdc-list-item" role="option" id="grains" tabindex="0">
-                  Bread, Cereal, Rice, and Pasta
+                <li class="mdc-list-item" role="option" id="cotton-candy" tabindex="0">
+                  Candy (cotton)
                 </li>
                 <li class="mdc-list-item" role="option" id="vegetables" aria-disabled="true" tabindex="0">
                   Vegetables
                 </li>
-                <li class="mdc-list-item" role="option" id="fruit" tabindex="0">
```
kfranqueiro: Why were these (and not Vegetables) removed? More specifically, removing these and leaving a disabled option (with the intent of being a group header) as the last item seems like bad/confusing UX.

demos/select.html
```diff
@@ -141,19 +136,25 @@ <h2 class="mdc-typography--title">Fully-Featured Component</h2>
           <label for="disabled">Disabled</label>
         </div>
       </section>
+
+
```
amsheehan: It is not. Assume all extraneous white space is a d'oh on my part :)

demos/select.html
```diff
@@ -179,37 +180,42 @@ <h2 class="mdc-typography--title">CSS Only</h2>
     <script>
       (function() {
         var MDCSelect = mdc.select.MDCSelect;
-        var root = document.getElementById('js-select');
-        var currentlySelected = document.getElementById('currently-selected');
+        var heroSelect = document.getElementById('hero-js-select');
+        mdc.select.MDCSelect.attachTo(heroSelect);
+      })();
+    </script>
```
amsheehan: Just to make it easier to copy and paste for people who want to do that.

packages/mdc-select/foundation.js
amsheehan: You both have good points. Ken, I think that seems feasible

demos/select.html
amsheehan: The intention is not about making a group header. It is about a disabled option. The others were removed for the sake of brevity.

packages/mdc-menu/simple/foundation.js
```diff
@@ -91,8 +91,10 @@ export default class MDCSimpleMenuFoundation extends MDCFoundation {
     this.keyupHandler_ = (evt) => this.handleKeyboardUp_(evt);
     /** @private {function(!Event)} */
     this.documentClickHandler_ = (evt) => {
-      this.adapter_.notifyCancel();
-      this.close(evt);
+      if (!evt.target.classList.contains('mdc-list-item')) {
```
amsheehan: possibly, but how far down the rabbit hole should we support?

packages/mdc-select/README.md
```diff
@@ -160,26 +169,33 @@ When used in components such as MDC Select, `mdc-list-item`'s can be disabled.
...
+<div class="mdc-select" role="listbox" tabindex="0">
+  <div class="mdc-select__surface">
+    <div class="mdc-select__label">Pick a Food Group</div>
+    <div class="mdc-select__selected-text"></div>
+    <div class="mdc-select__bottom-line"></div>
+  </div>
   <div class="mdc-simple-menu mdc-select__menu">
     <ul class="mdc-list mdc-simple-menu__items">
-      <li class="mdc-list-item" role="option" id="grains" tabindex="0">
+      <li class="mdc-list-item" role="option" aria-disabled="true">
```
amsheehan: Sure is. Good eye!

packages/mdc-select/README.md
```diff
@@ -160,26 +169,33 @@ When used in components such as MDC Select, `mdc-list-item`'s can be disabled.
...
+        Pick a food group
+      </li>
+      <li class="mdc-list-item" role="option" tabindex="0">
         Bread, Cereal, Rice, and Pasta
       </li>
-      <li class="mdc-list-item" role="option" id="vegetables" tabindex="-1" aria-disabled="true">
-        Vegetables (Disabled)
+      <li class="mdc-list-item" role="option" aria-disabled="true" tabindex="0">
```
amsheehan: tabindex="-1" is correct for disabled items.

packages/mdc-select/README.md
amsheehan: Ahh yep. Thanks

packages/mdc-select/README.md
amsheehan: Yeah I'll try it

packages/mdc-select/README.md
amsheehan: Mayhap we can, Ken. Mayhap we can.

packages/mdc-select/foundation.js
amsheehan: It is entirely for code style. Saves a future dev from having to traverse the file to see the effect passing `true/false` would have on the `close_()` method. I will make the name change.

packages/mdc-select/index.js
amsheehan: That code is precisely the reason for changing this. The extra DOM node is needed so the visible surface of the `select` element is different from its container. We need this change because the new styles require `overflow: hidden` on the visible surface, which would be problematic for showing a menu. We need the new surface however for the new ripple and bottom line behaviors, which are what this method is concerned with

test/unit/mdc-select/mdc-select.test.js
amsheehan: It was changed. This test now looks at the `surface` and not the `fixture`

kfranqueiro: I'm talking about the `getComputedStyleValue` test, which doesn't show any changes in the diff and still references `fixture`?  I know you changed the `setStyle` test, but you changed both of these adapter methods.

demos/select.html
acdvorak: IE 11 (mostly) supports `const`/`let`, so which browsers are we concerned about? Safari 9.x?

packages/mdc-menu/simple/foundation.js
acdvorak: I typically walk all the way up the tree until I get to the `body` or `documentElement`. It shouldn't really impact performance at all, and it ensures correct behavior in all cases.

demos/select.html
amsheehan: Even if everything fully supported es2015 syntax overnight, I don't think it would be a very good use of time to re-implement the demos just for the new syntactic sugars available to us. As delicious as they may be. :)

packages/mdc-menu/simple/foundation.js
amsheehan: Are you suggesting we do something like this:

```
function checkAncestryForClassname(el, classname) {
  if (el.classList.contains(classname)) {
    return true;
  }

  return el.parentNode && checkAncestryForClassname(el.parentNode, classname);
}
```

kfranqueiro: Basically, but it might be better done as a loop to avoid recursive calls?  (Also your code would either return `true` or `null` as written.)

```js
while (el) {
  if (el.classList.contains(className)) {
    return true;
  }
  el = el.parentNode;
}
return false;
```

packages/mdc-select/foundation.js
amsheehan: Upon further review, it's going to be problematic using the adapter method since it returns the value as a string with the unit attached, e.g.: `"16px"`. Going to assign it to a constant and make a note that it has to stay in sync with a `scss` variable.

packages/mdc-select/README.md
amsheehan: We can indeed. They are, in fact, the same. :)

test/unit/mdc-select/mdc-select.test.js
amsheehan: Interesting. I think it's because the `surface` is set to be the same width as the `root`. I'll update the test.

amsheehan:
@kfranqueiro These look like there is a story for the ripple improvements: https://github.com/material-components/material-components-web/issues/194

I'm inclined to leave this as is until that gets played, as this isn't _really_ a bug in `mdc-select` or `mdc-textfield`

packages/mdc-menu/simple/foundation.js
acdvorak: Yep, what @kfranqueiro said 😀

kfranqueiro:
RE the ripple on hover... it sounds like #194 is now obsolete according to current design direction, so we should probably just remove hover ripple from this PR (otherwise it'll just make more work to steamroll it later anyway).

test/unit/mdc-select/foundation.test.js
```diff
@@ -34,14 +35,16 @@ test('exports strings', () => {
 
 test('default adapter returns a complete adapter implementation', () => {
   verifyDefaultAdapter(MDCSelectFoundation, [
-    'addClass', 'removeClass', 'setAttr', 'rmAttr', 'computeBoundingRect',
+    'addClass', 'removeClass', 'addClassToLabel', 'removeClassFromLabel', 'addClassToBottomLine',
+    'removeClassFromBottomLine', 'setBottomLineAttr', 'setAttr', 'rmAttr', 'computeBoundingRect',
+    'registerPointerDownHandler', 'deregisterPointerDownHandler',
```
kfranqueiro: Remove [de]registerPointerDownHandler

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -214,34 +261,34 @@ test('adapter#focus focuses on the root element', () => {
 });
 
 test('adapter#makeTabbable sets the root element\'s tabindex to 0', () => {
-  const {component, fixture} = setupTest();
-  fixture.tabIndex = -1;
+  const {component, menuEl} = setupTest();
```
kfranqueiro: Update the names of these tests since it no longer affects the root element?

packages/mdc-select/README.md
amsheehan: We may need to update the MDC Select example shown here, but what makes you think this wouldn't work anymore?

packages/mdc-menu/simple/foundation.js
```diff
@@ -276,6 +274,25 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+   * Handle clicks and cancel the menu if not a list item
+   * @param {!Event} evt
+   * @private
+   */
+  handleDocumentClick_(evt) {
+    let el = evt.target;
+
+    while (el && el !== document.body) {
+      if (this.adapter_.eventTargetHasClass(el, cssClasses.LIST_ITEM)) {
+        return true;
```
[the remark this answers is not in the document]
amsheehan: True that

packages/mdc-select/foundation.js
```diff
@@ -216,9 +255,14 @@ export default class MDCSelectFoundation extends MDCFoundation {
     this.adapter_.setMenuElStyle('transform-origin', `center ${itemOffsetTop}px`);
   }
 
-  close_() {
+  close_(shouldRemoveFloatingLabelClass) {
     const {OPEN} = MDCSelectFoundation.cssClasses;
     this.adapter_.removeClass(OPEN);
+
+    if (shouldRemoveFloatingLabelClass && this.getSelectedIndex() === -1) {
```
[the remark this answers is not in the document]
amsheehan: That's a good question. We certainly don't need it anymore.

packages/mdc-select/mdc-select.scss
```diff
@@ -15,68 +15,65 @@
...
 
 @mixin mdc-select-dd-arrow-svg-bg_($fill-hex-number: 000000, $opacity: .54) {
   background-image: url(data:image/svg+xml,%3Csvg%20width%3D%2210px%22%20height%3D%225px%22%20viewBox%3D%227%2010%2010%205%22%20version%3D%221.1%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20xmlns%3Axlink%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%22%3E%0A%20%20%20%20%3Cpolygon%20id%3D%22Shape%22%20stroke%3D%22none%22%20fill%3D%22%23#{$fill-hex-number}%22%20fill-rule%3D%22evenodd%22%20opacity%3D%22#{$opacity}%22%20points%3D%227%2010%2012%2015%2017%2010%22%3E%3C%2Fpolygon%3E%0A%3C%2Fsvg%3E);
 }
 
+@function mdc-select-transition($property) {
+  @return #{$property} 180ms $mdc-animation-standard-curve-timing-function;
+}
+
+$mdc-select-arrow-padding: 24px;
```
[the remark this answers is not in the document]
amsheehan: Yep. Good catch.

packages/mdc-select/mdc-select.scss
```diff
@@ -88,12 +85,127 @@
...
+    padding-left: 16px;
+    border: none;
+    border-radius: 4px 4px 0 0;
+    outline: none;
+    background-color: rgba(black, .04);
+    // Resets for <select> element
+    appearance: none;
+    overflow: hidden;
+
+    &::before,
```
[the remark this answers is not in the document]
amsheehan: Fixed

packages/mdc-select/mdc-select.scss
```diff
@@ -88,12 +85,127 @@
...
+      border-radius: 50%;
+      background-color: inherit;
+      opacity: 0;
+      pointer-events: none;
+      content: "";
+    }
+
+    // NOTE: For some reasons, stylelint complains that the two patterns below don't follow BEM.
+    // However, it seems to emit the wrong selector for that pattern. Thus, we disable it above where
+    // we would normally disable it (&.mdc-textfield--disabled) as a workaround.
```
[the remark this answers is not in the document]
amsheehan: Ohhh whoops! Yeah, this doesn't need to be this broad, it is needed in the multi select block though. Updated accordingly.

demos/select.html
```diff
@@ -140,17 +134,21 @@ <h2 class="mdc-typography--title">Fully-Featured Component</h2>
...
-          <option value="grains">Bread, Cereal, Rice, and Pasta</option>
-          <option value="vegetables" disabled>Vegetables</option>
-          <option value="fruit">Fruit</option>
-          <option value="dairy">Milk, Yogurt, and Cheese</option>
-          <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
-          <option value="fats">Fats, Oils, and Sweets</option>
-        </select>
+        <div class="mdc-select">
+          <select class="mdc-select__surface">
+            <option value="" default selected>Pick a food group</option>
```
acdvorak: I don't think `default` is a valid attribute for `<option>` ([MDN attribute reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes))

demos/select.html
```diff
@@ -176,40 +174,47 @@ <h2>Select Multiple - CSS Only</h2>
...
+        var boxCurrentlySelected = document.getElementById('box-currently-selected');
+        var select = new MDCSelect(root);
+        select.ripple.layout();
+
         root.addEventListener('MDCSelect:change', function() {
           var item = select.selectedOptions[0];
           var index = select.selectedIndex;
-          currentlySelected.textContent = '"' + item.textContent + '" at index ' + index +
+          boxCurrentlySelected.textContent = '"' + item.textContent + '" at index ' + index +
            ' with value "' + select.value + '"';
```
acdvorak: I know we don't lint our demo code, but could we possibly reindent this line while we're here?

packages/mdc-select/README.md
```diff
@@ -189,24 +202,25 @@ To disable a list item, set `aria-disabled` to `"true"`, and set `tabindex` to `
 
 ### Using the Pure CSS Select
 
-The `mdc-select` CSS class also works on browser's native `<select>` elements, allowing for a
+The `mdc-select` CSS classes also work with the browser's native `<select>` element, allowing for a
 seamless, un-invasive experience in browsers where a native select may be more appropriate, such as
 on a mobile device. It does not require any javascript, nor any CSS for `mdc-menu` or `mdc-list`.
+Simply wrap the `<select>` element in one with class `mdc-select`, add `mdc-select__surface` to `<select>`,
```
acdvorak: This sentence was a little confusing to me; maybe reword this line and the next line to so something like:

```markdown
Simply:

1. Wrap the `<select>` with a block element that has an `mdc-select` class
2. Add the `mdc-select__surface` class to the `<select>`
3. Append `<div class="mdc-select__bottom-line"></div>` immediately after the `<select>`
```

demos/select.html
amsheehan: Whoa. Yep.

demos/select.html
amsheehan: Sure can

packages/mdc-select/README.md
amsheehan: I'm going use `E.g.:` instead of `Simply:` since that's a pretty common pattern in our READMEs.

packages/mdc-select/foundation.js
```diff
@@ -190,8 +215,21 @@ export default class MDCSelectFoundation extends MDCFoundation {
...
     this.setMenuStylesForOpenAtIndex_(focusIndex);
+    this.adapter_.addClassToLabel(cssClasses.LABEL_FLOAT_ABOVE);
+    this.adapter_.addClassToBottomLine(cssClasses.BOTTOM_LINE_ACTIVE);
     this.adapter_.addClass(OPEN);
     this.adapter_.openMenu(focusIndex);
+    this.isFocused_ = true;
+  }
+
+  setBottomLineOrigin_(evt) {
+    const targetClientRect = evt.target.getBoundingClientRect();
```
acdvorak: If the client decides to put child elements inside of the option, wouldn't `evt.target` potentially point to the wrong element? E.g.:

```html
<li class="mdc-list-item" role="option" tabindex="0">
  <div class="foo">
    <i class="material-icons">date_range</i>
    Date Range
  </div>
</li>
```

packages/mdc-select/index.js
```diff
@@ -54,6 +56,12 @@ export class MDCSelect extends MDCComponent {
     this.foundation_.setDisabled(disabled);
   }
 
+  constructor(...args) {
+    super(...args);
+
+    this.ripple;
```
acdvorak: What's this statement for?

packages/mdc-menu/simple/foundation.js
```diff
@@ -275,6 +273,27 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
 
+  /**
+   * Handle clicks and cancel the menu if not a list item
+   * @param {!Event} evt
+   * @private
+   */
+  handleDocumentClick_(evt) {
+    let el = evt.target;
+
+    while (el && el !== document.body) {
```
acdvorak: Would it make sense to refactor this logic out of `handleDocumentClick_` and into `eventTargetHasClass`? Maybe renaming it to something like `eventTargetOrAncestorHasClass`.

That would make the logic easier to reuse and simpler to test.

kfranqueiro: It would also make the adapter method harder to write and I think we're all for simple adapter methods... so it's up to which is higher priority.

If we ever end up tackling #1104 this might be a candidate for that...

amsheehan: Yeah, I think we've always tried to keep the adapter methods as close to one line as possible.

packages/mdc-select/foundation.js
```diff
@@ -176,10 +197,14 @@ export default class MDCSelectFoundation extends MDCFoundation {
 
     let maxTextLength = 0;
     for (let i = 0, l = this.adapter_.getNumberOfOptions(); i < l; i++) {
+      // SURFACE_RIGHT_PADDING and SURFACE_LEFT_PADDING correspond to variables set in ./mdc-select.scss
```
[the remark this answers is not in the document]
amsheehan: I think `getComputedStyle()` is more aligned with what the rest of the library does in that regard.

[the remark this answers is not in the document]
amsheehan: I guess we could utilize `this.adapter_.getComputedStyleVlaue(style)` since it inspects the surface for typographical styles with a given style rule.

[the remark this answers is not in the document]
amsheehan: And then do the same in the tests.

packages/mdc-select/foundation.js
amsheehan: I...actually don't think there would be a consequence to this. The method is setting the transform origin in the x direction for the bottom line to be the location of the pointer along the x-axis. This would (I think) have the same result no matter what children are actioned on. I will test this, but I suspect we're fine as is with this.

packages/mdc-select/index.js
amsheehan: Oooh, that's some trash left over from a previous commit. Removing.

packages/mdc-select/foundation.js
```diff
@@ -219,9 +257,13 @@ export default class MDCSelectFoundation extends MDCFoundation {
     this.adapter_.setMenuElStyle('transform-origin', `center ${itemOffsetTop}px`);
   }
 
-  close_() {
+  close_(shouldRemoveFloatingLabelClass) {
```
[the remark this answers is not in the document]
amsheehan: Well, I know what my next Medium post is going to be. hahaha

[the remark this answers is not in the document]
amsheehan: So, the only argument I have is that this example doesn't really fall into what's described in that blog post. Since I define a variable and pass that, it's entirely readable and easy to assume what the boolean is governing in the `close_()` method. Thoughts?

packages/mdc-select/foundation.js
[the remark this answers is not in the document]
acdvorak: Could we mock the `getComputedStyle` method so that it returns
`require('./mdc-select.scss')['$mdc-select-arrow-padding']` (or w/e)?

packages/mdc-select/foundation.js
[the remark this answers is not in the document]
acdvorak: That's a very good point - and thanks for using such explicit variable names 😄  🎉 

Personally, I would still prefer an enum because it prevents an entire class of human errors by forcing the caller to use statically typed values with descriptive names. It also looks a little cleaner IMO.

E.g.:

```js
const shouldRemoveFloatingLabelClass = true;
this.close_(shouldRemoveFloatingLabelClass);
```

vs.

```js
this.close_(FloatingLabelBehavior.REMOVE_ON_CLOSE);
```

In this particular case I don't feel _too_ strongly either way, though, since it's a private method with only one param. So it's totally up to you 🙂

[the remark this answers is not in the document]
kfranqueiro: So, I'd prefer avoiding the enum, but before I even weigh in on that conversation... do we even need this parameter?

We're calling this method in 2 places: one place always passes `true`, while the other always passes `false`. Can't we just move the remaining condition and `removeClassFromLabel` call to the location where we're currently calling this with `true`?

packages/mdc-menu/simple/foundation.js
```diff
@@ -275,6 +273,27 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+      if (this.adapter_.eventTargetHasClass(el, cssClasses.LIST_ITEM)) {
+        return;
+      }
+      el = el.parentNode;
+    }
+
+    if (this.adapter_.hasClass(MDCSimpleMenuFoundation.cssClasses.OPEN)) {
```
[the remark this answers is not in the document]
amsheehan: We actually don't need the first step if we move `this.adapter_.registerBodyClickHandler()` back into the raf callback.

packages/mdc-select/foundation.js
```diff
@@ -107,7 +118,14 @@ export default class MDCSelectFoundation extends MDCFoundation {
       MDCSimpleMenuFoundation.strings.SELECTED_EVENT, this.selectionHandler_);
     this.adapter_.registerMenuInteractionHandler(
       MDCSimpleMenuFoundation.strings.CANCEL_EVENT, this.cancelHandler_);
+    ['mousedown', 'touchstart'].forEach((evtType) => {
+      this.adapter_.registerInteractionHandler(evtType, this.setPointerXOffset_);
+    });
     this.resize();
+
+    if (this.selectedIndex_ === -1) {
+      this.adapter_.removeClassFromLabel(cssClasses.LABEL_FLOAT_ABOVE);
```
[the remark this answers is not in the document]
amsheehan: It isn't, removed.

packages/mdc-select/mdc-select.scss
```diff
@@ -89,12 +83,110 @@
     z-index: 4; // Should pop up above everything else. temporary-drawer is next highest at 3.
   }
 
+  &__surface {
+    @include mdc-typography(subheading2);
+    @include mdc-theme-prop(color, text-primary-on-light);
+    @include mdc-rtl-reflexive-box(padding, right, $mdc-select-arrow-padding);
+    @include mdc-rtl-reflexive-box(padding, left, $mdc-select-label-padding);
+    @include mdc-ripple-surface;
+    @include mdc-ripple-color;
```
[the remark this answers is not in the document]
amsheehan: Yeah, if anything I think they'd be folded into ripple or theme or something, I definitely don't think select should depend on text field. For now I'll just make select match text field.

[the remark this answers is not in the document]
amsheehan: > As discussed offline, we are going to leave these ripple color values alone because they match the states work that Ken is doing. This means we can either make Text Field reflect this, or leave Text Field alone until states work commences. -amsheehan

packages/mdc-select/mdc-select.scss
```diff
@@ -89,12 +83,110 @@
...
+      transform-origin: right top;
+    }
+
+    &--float-above {
+      transform: translateY(-40%) scale(.75, .75);
+    }
+  }
+
   &__selected-text {
+    display: flex;
```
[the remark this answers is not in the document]
amsheehan: It allows us to align the text properly.

`align-items: flex-end;`

packages/mdc-select/mdc-select.scss
```diff
@@ -89,12 +83,110 @@
...
+    align-items: flex-end;
+    margin-bottom: 6px;
+    padding-right: 0;
```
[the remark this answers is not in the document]
amsheehan: Hmmm they don't appear to be doing much of anything.

packages/mdc-select/foundation.js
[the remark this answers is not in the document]
amsheehan: Shooooot. Ken wins. The only caveat being we have to check if an item is selected in the cancel handler before removing that label.

packages/mdc-select/mdc-select.scss
```diff
@@ -15,69 +15,60 @@
...
+  height: 56px;
   border: none;
-  border-bottom: 1px solid rgba(black, .12);
-  border-radius: 0;
-  background-color: transparent;
+  border-radius: 4px 4px 0 0;
+  outline: none;
   background-repeat: no-repeat;
-  background-position: right center;
-  font-family: Roboto, sans-serif;
```
kfranqueiro: FYI this was also removed in a PR that was just merged. Not sure if this will cause a merge conflict next time you merge from master.

packages/mdc-select/mdc-select.scss
```diff
@@ -211,8 +313,8 @@
     .mdc-list-divider {
       margin-left: -16px;
     }
+    // stylelint-enable
```
kfranqueiro: Not sure why `plugin/selector-bem-pattern` was omitted here? We still include it in all other stylelint-enable comments.

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -41,10 +41,16 @@ class FakeMenu {
 function getFixture() {
   return bel`
     <div class="mdc-select" role="listbox" tabindex="0">
-      <span class="mdc-select__selected-text">Pick a food group</span>
-      <div class="mdc-select__menu mdc-simple-menu">
-        <ul class="mdc-simple-menu__items">
-          <li class="mdc-list-item" role="option" tabindex="0">An option</li>
+      <div class="mdc-select__surface">
```
kfranqueiro: I thought you were talking about needing to nest this fixture within another div to make this work, but I don't see an additional parent div from the last commit, I just see added indentation. What happened?

packages/mdc-select/README.md
```diff
@@ -331,18 +348,23 @@ within `componentDidUpdate`.
...
 | `deregisterInteractionHandler(type: string, handler: EventListener) => void` | Removes an event listener `handler` for event type `type` on the root element. |
 | `focus() => void` | Focuses the root element |
-| `makeTabbable() => void` | Allows the root element to be tab-focused via keyboard. We achieve this by setting the root element's `tabIndex` property to `0`. |
-| `makeUntabbable() => void` | Disallows the root element to be tab-focused via keyboard. We achieve this by setting the root element's `tabIndex` property to `-1`. |
-| `getComputedStyleValue(propertyName: string) => string` | Get the root element's computed style value of the given dasherized css property `propertyName`. We achieve this via `getComputedStyle(...).getPropertyValue(propertyName). `|
-| `setStyle(propertyName: string, value: string) => void` | Sets a dasherized css property `propertyName` to the value `value` on the root element. We achieve this via `root.style.setProperty(propertyName, value)`. |
+| `makeTabbable() => void` | Allows the select options to be tab-focused via keyboard. We achieve this by setting the root element's `tabIndex` property to `0`. |
```
kfranqueiro: I'd vote for changing "select options" back to "root element" on this and the next line, and then these lines end up unchanged, much like the implementation itself.

test/unit/mdc-select/mdc-select.test.js
amsheehan: I read the docs... hahaha

You can operate on the fixture as a whole as if it were any other dom node, so instead of trying to select on `mdc-select` in the test I can just do `fixture.setAttribute('style', 'whatever')`

packages/mdc-select/README.md
amsheehan: Mmmkay

amsheehan:
We'd have to do a bit of work to get the position to change in RTL. It would require a new adapter method and some work in the foundation. If that's something we want to do I can add it.

acdvorak:
Can we darken the border color of the CSS-only select on focus? Currently there's no way to tell that the element has focus if there are no items selected.

amsheehan:
@acdvorak Oowee can do!

PR #1100, opened 2017-08-09
feat(theme): step 01: accent color defaults to primary color

### LINKED ISSUE #1099: Theme: accent color should default to primary color
After talking to our designers, we felt that it doesn't really make sense to provide a specific default value for accent color, because it has a low likelihood of looking good with your brand's custom primary color.

acdvorak:
Designers felt that it doesn't make sense to provide a specific default value for accent color, because it has a low likelihood of looking good with your brand's custom primary color.

Instead, accent color should default to primary color.

Fixes #1099

### DIFF
--- a/packages/mdc-theme/_variables.scss
+++ b/packages/mdc-theme/_variables.scss
@@ -21,7 +21,7 @@
   If you're a user customizing your color scheme in SASS, these are probably the only variables you need to change.
 */
 $mdc-theme-primary: #3f51b5 !default; /* Indigo 500 */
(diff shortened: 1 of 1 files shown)


acdvorak:
Closing this issue after discussing with the team.

If the default secondary color clashes with a client's custom primary color, the client can simply change secondary to something else (including making it the same as primary).

Having a default secondary color that's different from the primary color is valuable because it makes it obvious to clients that there are two different theme slots which can be customized.

PR #1112, opened 2017-08-10
Change theme's SCSS comment style from `/**` to `//`

acdvorak:
`//` comments don't appear in the CSS output.

### DIFF
--- a/packages/mdc-theme/_variables.scss
+++ b/packages/mdc-theme/_variables.scss
@@ -16,28 +16,18 @@
 
 @import "./functions";
 
-//
 // Main theme colors.
-//
 // If you're a user customizing your color scheme in SASS, these are probably the only variables you need to change.
-//
-
 $mdc-theme-primary: #3f51b5 !default; // Indigo 500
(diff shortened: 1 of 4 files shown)


packages/mdc-theme/_variables.scss
```diff
@@ -16,20 +16,28 @@
...
 @import "./functions";
 
-/*
-  Main theme colors.
-  If you're a user customizing your color scheme in SASS, these are probably the only variables you need to change.
-*/
-$mdc-theme-primary: #3f51b5 !default; /* Indigo 500 */
-$mdc-theme-accent: #ff4081 !default; /* Pink A200 */
-$mdc-theme-background: #fff !default; /* White */
+//
```
yeelan0319: Super nit: if we introduce no empty lines in other file, could we do consistently in this file?

acdvorak: Done!

The reason I added the blank lines is because they become necessary in PR #1116. But I'll remove them from this PR (#1112) and add them in #1116 instead.

PR #1113, opened 2017-08-10
chore(demos): Remove redundant CSS ref in HTML

acdvorak:
`assets/demo-styles.css.js` already includes the styles present in `assets/material-components-web.css.js`, so remove the latter.

### DIFF
--- a/demos/linear-progress.html
+++ b/demos/linear-progress.html
@@ -16,7 +16,6 @@
     <meta charset="utf-8">
     <title>Linear Progress - Material Compoonents Catalog</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <script src="assets/material-components-web.css.js" charset="utf-8"></script>
     <link href="demos.css" rel="stylesheet">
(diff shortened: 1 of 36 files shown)


acdvorak:
Great catch, @kfranqueiro! I removed the rest, PTAL 🙂

demos/linear-progress.html
```diff
@@ -16,7 +16,6 @@
     <meta charset="utf-8">
     <title>Linear Progress - Material Compoonents Catalog</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <script src="assets/material-components-web.css.js" charset="utf-8"></script>
```
kfranqueiro: Removing this currently breaks this demo's styles because the next line is broken - `demos.css` should be `assets/demo-styles.css.js`

acdvorak: Good catch. I fixed that in PR #1119, but forgot to backport it here.

Fixed it here - PTAL. Thanks!

PR #1117, opened 2017-08-11
feat(theme): Add SCSS variables for Material Design color palette

acdvorak:


packages/mdc-theme/_color_palette.scss
```diff
@@ -0,0 +1,290 @@
...
+$material-color-orange-400: #ffa726;
+$material-color-orange-500: #ff9800;
+$material-color-orange-600: #fb8c00;
+$material-color-orange-700: #f57c00;
+$material-color-orange-800: #ef6c00;
+$material-color-orange-900: #e65100;
+$material-color-orange-a100: #ffd180;
+$material-color-orange-a200: #ffab40;
+$material-color-orange-a400: #ff9100;
+$material-color-orange-a700: #ff6500;
```
kfranqueiro: Is this correct? The color swatches on material.io/guidelines say this is `#ff6d00`.

acdvorak: Wow, great catch! Fixed.

I wrote a little script to verify that all of the colors match the material.io site, and this was the only one that didn't, so I'm fairly confident the rest are correct.

PTAL!

amsheehan: Bless both of your hearts to DEATH

PR #1128, opened 2017-08-16
perf(infrastructure): Cut build time in half with opt-in env var

acdvorak:
A new `GENERATE_SOURCE_MAPS=false` env var disables source map generation in dev mode, which cuts build and recompile times in half.

Invoked like so:

```bash
GENERATE_SOURCE_MAPS=false npm run dev
```

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -27,8 +27,8 @@ const PUBLIC_PATH = '/assets/';
 const IS_DEV = process.env.MDC_ENV === 'development';
 const IS_PROD = process.env.MDC_ENV === 'production';
 const IS_FAST_BUILD = process.env.MDC_BUILD === 'fast';
-const GENERATE_SOURCE_MAPS = IS_DEV && !IS_FAST_BUILD;
-const DEVTOOL = GENERATE_SOURCE_MAPS ? 'source-map' : false;
+const GENERATE_SOURCE_MAPS = IS_DEV && !IS_FAST_BUILD && !process.env.MDC_GENERATE_SOURCE_MAPS;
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -26,6 +26,9 @@ const OUT_PATH = path.resolve('./build');
 const PUBLIC_PATH = '/assets/';
 const IS_DEV = process.env.MDC_ENV === 'development';
 const IS_PROD = process.env.MDC_ENV === 'production';
+const IS_FAST_BUILD = process.env.MDC_BUILD === 'fast';
+const GENERATE_SOURCE_MAPS = IS_DEV && !IS_FAST_BUILD && process.env.MDC_GENERATE_SOURCE_MAPS !== 'false';
```
kfranqueiro: Is there a reason we're introducing multiple env vars here (`MDC_BUILD` and `MDC_GENERATE_SOURCE_MAPS` here, and `MDC_DEVTOOL` below)? Do you have more plans for `MDC_BUILD=fast`?

Do we document the env vars we look at here, anywhere? Should we (even if it's just in this file above the constants we define)?

acdvorak: Good point. I removed `MDC_BUILD=fast` and `MDC_DEVTOOL`, so now there's only `GENERATE_SOURCE_MAPS=false`, which seems simpler and more straightforward.

AFAIK, the only env var we currently have is `MDC_ENV=[development|production]`. It's not explicitly documented anywhere, and I'm not sure it needs to be, because I can't think of a reason that a developer would want to change its value.

I'm still experimenting with making our build faster, so I'd prefer not to document `MDC_GENERATE_SOURCE_MAPS` until I'm finished making changes and the API is stable.

PR #1129, opened 2017-08-16
feat(theme): `mdc-theme-prop` accepts literal color values

acdvorak:
The `mdc-theme-prop` SCSS mixin now accepts literal color values (e.g., `green`, `#fff`) for `$value`, in addition to property names from the `$mdc-theme-property-values` map.

This gives components the flexibility to do either `mdc-theme-prop(color, primary)` or `mdc-theme-prop(color, #fff)`.


packages/mdc-theme/_mixins.scss
```diff
@@ -18,22 +18,38 @@
...
 // Applies the correct theme color style to the specified property.
 // $property is typically color or background-color, but can be any CSS property that accepts color values.
-// $style should be one of the map keys in $mdc-theme-property-values (_variables.scss).
+// $style should be one of the map keys in $mdc-theme-property-values (_variables.scss), or a literal color value.
 @mixin mdc-theme-prop($property, $style, $important: false) {
-  @if not map-has-key($mdc-theme-property-values, $style) {
-    @error "Invalid style: '#{$style}'. Choose one of: #{map-keys($mdc-theme-property-values)}";
-  }
+  @if type-of($style) == "color" {
+    $value: $style;
```
kfranqueiro: I guess `$value` is here for consistency with the `@else` branch below, but it seems kind of unnecessary here? Could easily just reference `$style` directly.

amsheehan:
My only concern is that this isn't entirely all that helpful IMO.

First and foremost, MDC Web should serve Material Design and its guidelines. While this doesn't explicitly run counter to that, it allows developers a little bit _too_ much arbitrary control over the theming subsystem, without making anything easier for the user. To me, having the mixing accept literal color values is the same as saying, "just add the color ala carte in your CSS using normal style rules". It does nothing to express our guidance over theming, nor does it enforce better code (like having the user supply a map of color values for our mixin to fall back on).

acdvorak:
After discussing offline with the team, we decided that this PR is indeed the best option. It gives clients the most flexibility to customize component colors.

PR #1130, opened 2017-08-16
refactor(fab): Move all styles into new mdc-fab-{base,theme} mixins

acdvorak:
# DIFFBASE = PR #1129

This refactor will make it easier to override FAB's baseline theme with custom colors, by simply calling a single mixin.


packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,154 @@
...
+      fg-color: rgba(0, 0, 0, .26),
+      ripple: null
+    ),
+  );
+}
+
+@function mdc-map-get-child_($parent-map, $keys...) {
+  $value: $parent-map;
+
+  @each $key in join($keys...) {
```
kfranqueiro: Can you add a comment as to what this function does? The name and how it's used do kind of explain it, but when I first looked at the loop I expected that the reassignment and re-referencing of `$value` was a mistake.

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,154 @@
...
+
+@import "@material/animation/functions";
+@import "@material/animation/variables";
+@import "@material/elevation/mixins";
+@import "@material/ripple/mixins";
+@import "@material/theme/mixins";
+
+@function mdc-fab-theme-default-config_() {
+  @return (
+    standard: (
```
kfranqueiro: Related to #1129, I'm concerned that this will become confusing for our users if we suddenly allow mixing and matching literal colors and names from our theme map.  Will it be possible to draw a clear line between where this works and where it doesn't?

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,154 @@
...
+
+  display: inline-flex;
+  position: relative;
+  justify-content: center;
+  width: 56px;
+  height: 56px;
+  padding: 0;
+  transition: box-shadow 280ms $mdc-animation-standard-curve-timing-function;
+  border: none;
+  border-radius: 50%;
```
kfranqueiro: I believe border-radius (and possibly also border itself) may also end up being customizable. Maybe not necessary for a first pass but possibly worth noting.

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,154 @@
...
+
+  @include mdc-theme-prop(background-color, mdc-map-get-child_($config, $type, bg-color));
+  @include mdc-theme-prop(color, mdc-map-get-child_($config, $type, fg-color));
+}
+
+@mixin mdc-fab-theme($config: ()) {
+  $config: map-merge(mdc-fab-theme-default-config_(), $config);
+
+  &:not(.mdc-ripple-upgraded) {
+    -webkit-tap-highlight-color: rgba(0, 0, 0, .18);
```
kfranqueiro: Should this actually be in the base mixin if it won't change? (Maybe it _should_ change based on the FAB's foreground color, but that's probably a separate design question...)

packages/mdc-fab/mdc-fab.scss
```diff
@@ -12,98 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-@import "@material/animation/functions";
-@import "@material/animation/variables";
-@import "@material/elevation/mixins";
-@import "@material/ripple/mixins";
-@import "@material/theme/mixins";
+@import "mixins";
```
kfranqueiro: Seems like we are typically explicit and reference `./mixins` for imports in the same folder.

acdvorak:
Superseded by PR #1131.

PR #1131, opened 2017-08-16
feat(theme): Luminance-aware light/dark variants

acdvorak:
If the given theme color is already very light, the "light" variant will be darkened instead of lightened.

Likewise, if the given theme color is already very dark, the "dark" variant will be lightened instead of darkened.

This ensures that the light and dark variants are always different than their corresponding main color.

See the theme demo page for an example.


packages/mdc-theme/_functions.scss
```diff
@@ -51,3 +56,66 @@
...
 }
+
+// Return the given number if $value >= $min and $value <= $max. If $value falls
+// outside those bounds, return the nearest boundary value ($max or $min).
+@function mdc-theme-clamp-number_($value, $min, $max) {
+  @return max($min, min($max, $value));
+}
+
+// lighten() and darken() require values to be between 0% and 100%.
+@function mdc-theme-clamp-lighten-amount_($percentage) {
```
kfranqueiro: If this is used for both lightening and darkening, then this function name is somewhat confusing.  What about `mdc-theme-clamp-percentage_`?

Also, do you foresee other use cases for `mdc-theme-clamp-number_`, or should we just combine these two into a single function?

packages/mdc-theme/_functions.scss
```diff
@@ -51,3 +56,66 @@
...
+// within its tonal palette.
+//
+// If the color is already very dark, it will be lightened instead of darkened
+// to ensure that the returned value is visually distinct from the input color.
+//
+// If the color is very light, it will be darkened twice as much as usual to
+// ensure that the returned value is visually distinct from the light variant
+// (which will actually end up being a dark shade).
+@function mdc-theme-dark-variant($color, $num-indexes: 2) {
+  $offset: $_mdc-theme-tonal-offset;
```
kfranqueiro: Is this a style guide convention? Why not just reference `$_mdc-theme-tonal-offset` on the next line?

packages/mdc-theme/_functions.scss
```diff
@@ -51,3 +56,66 @@
...
+// within its tonal palette.
+//
+// If the color is already very light, it will be darkened instead of lightened
+// to ensure that the returned value is visually distinct from the input color.
+//
+// If the color is very dark, it will be lightened twice as much as usual to
+// ensure that the returned value is visually distinct from the dark variant
+// (which will actually end up being a light tint).
+@function mdc-theme-light-variant($color, $num-indexes: 2) {
+  $offset: $_mdc-theme-tonal-offset;
```
kfranqueiro: Is this a style guide convention? Why not just reference `$_mdc-theme-tonal-offset` on the next line?

packages/mdc-theme/_functions.scss
```diff
@@ -51,3 +56,66 @@
...
+//
+// If the color is very light, it will be darkened twice as much as usual to
+// ensure that the returned value is visually distinct from the light variant
+// (which will actually end up being a dark shade).
+@function mdc-theme-dark-variant($color, $num-indexes: 2) {
+  $offset: $_mdc-theme-tonal-offset;
+  $amount: $offset * $num-indexes;
+
+  @if mdc-theme-luminance($color) < .2 {
+    @return lighten($color, mdc-theme-clamp-lighten-amount_($amount));
```
kfranqueiro: Won't this mean that a color's dark variant will be exactly the same as its light variant? Wouldn't that create confusion?

packages/mdc-theme/_functions.scss
```diff
@@ -51,3 +56,66 @@
...
+@function mdc-theme-light-variant($color, $num-indexes: 2) {
+  $offset: $_mdc-theme-tonal-offset;
+  $amount: $offset * $num-indexes;
+
+  @if mdc-theme-luminance($color) < .2 {
+    @return lighten($color, mdc-theme-clamp-lighten-amount_($amount * 2));
+  }
+
+  @else if mdc-theme-luminance($color) > .8 {
+    @return darken($color, mdc-theme-clamp-lighten-amount_($amount));
```
kfranqueiro: Won't this mean that a color's dark variant will be exactly the same as its light variant? Wouldn't that create confusion?

packages/mdc-theme/_functions.scss
acdvorak: I split this into two functions to make it more readable, but I think you're right that they can probably just be combined.

packages/mdc-theme/_functions.scss
acdvorak: Done.

packages/mdc-theme/_functions.scss
acdvorak: Done.

packages/mdc-theme/_functions.scss
acdvorak: No, in this case the light and dark variants are calculated differently, but the code definitely doesn't make that clear. I refactored it and put both calculations in a single function, so hopefully it's easier to see what's going on.

packages/mdc-theme/_functions.scss
acdvorak: No, in this case the light and dark variants are calculated differently, but the code definitely doesn't make that clear. I refactored it and put both calculations in a single function, so hopefully it's easier to see what's going on.

acdvorak:
Addressed comments; PTAL!

packages/mdc-theme/_functions.scss
```diff
@@ -51,3 +56,63 @@
...
+}
+
+// Generate light and dark variants of the given color, offset by approximately
+// the specified number of indexes within the color's tonal palette.
+@function mdc-theme-tonal-variants_($color, $num-indexes: 2) {
+  $luminance: mdc-theme-luminance($color);
+  $amount-1x: mdc-theme-clamp-percentage_($_mdc-theme-tonal-offset * $num-indexes);
+  $amount-2x: mdc-theme-clamp-percentage_($_mdc-theme-tonal-offset * $num-indexes * 2);
+
+  @if $luminance < .2 {
```
kfranqueiro: Yeah, I totally misinterpreted this before and this seems like a clearer way to organize it.

Random thought (maybe a design question): is there a reason these thresholds are absolute rather than relative to the tonal offset?

acdvorak: Nope! Updated 🙂

acdvorak:
* Added a note to the README explaining how the functions handle extreme luminance values.
* The theme demo page already has an example of a very dark color (primary = `#212121` (grey 900)), which is probably sufficient.

packages/mdc-theme/_functions.scss
```diff
@@ -16,6 +16,11 @@
 
 @import "./constants";
 
+// Used by the functions below to shift a color's lightness by approximately
```
kfranqueiro: Would "luminance" or "shade" possibly make more sense here than "lightness"?

acdvorak: I'm not sure which term is technically correct (the best kind of correct!).

[libsass's implementation of `lighten` and `darken`](https://github.com/sass/libsass/blob/0dea5f2d6b6b13e39a0d66b5ee45e29ed362c029/src/functions.cpp#L566) converts the input color to HSL, increases the 'L', and returns the result.

According to [Wikipedia's article on HSL](https://en.wikipedia.org/wiki/HSL_and_HSV), the 'L' stands for "luminance" _or sometimes_ "lightness", which the [article later defines as being different things](https://en.wikipedia.org/wiki/HSL_and_HSV#Formal_derivation).

In addition, we don't use Sass's built-in `lightness` function - instead, we implement our own `mdc-theme-luminance` function based on a [W3 spec](https://www.w3.org/TR/WCAG20-TECHS/G17.html#G17-tests), which makes me think that there are subtly different and competing usages of the words "lightness" and "luminance" floating around the Web and in our codebase.

In short, I have no idea which word to use 😝

PR #1133, opened 2017-08-16
feat(infrastructure): Add env var to output CSS files directly instead of wrapping them in JS

acdvorak:
This will allow us to eventually move the demo pages away from webpack's horrifying JavaScript-wrapped CSS to a simple hot-swap of the extracted CSS files. Doing so will eliminate the FOUC and make it possible to break apart our monolithic recompiles into smaller, faster, more modular chunks.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -86,6 +86,7 @@ module.exports = [{
     radio: [path.resolve('./packages/mdc-radio/index.js')],
     ripple: [path.resolve('./packages/mdc-ripple/index.js')],
     select: [path.resolve('./packages/mdc-select/index.js')],
+    selectionControl: [path.resolve('./packages/mdc-selection-control/index.js')],
     slider: [path.resolve('./packages/mdc-slider/index.js')],
     snackbar: [path.resolve('./packages/mdc-snackbar/index.js')],
     tabs: [path.resolve('./packages/mdc-tabs/index.js')],
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -75,7 +75,10 @@ const CSS_LOADER_CONFIG = [
   },
 ];
 
-const CSS_JS_FILENAME_OUTPUT_PATTERN = `[name]${IS_PROD ? '.min' : ''}.css.js`;
+// In development, these are emitted as js files to facilitate hot module replacement. In
+// all other cases, ExtractTextPlugin is used to generate the final css, so this is given a
+// dummy ".css-entry" extension.
+const CSS_JS_FILENAME_OUTPUT_PATTERN = `[name]${IS_PROD ? '.min' : ''}.css${IS_DEV ? '.js' : '-entry'}`;
 const CSS_FILENAME_OUTPUT_PATTERN = `[name]${IS_PROD ? '.min' : ''}.css`;
 
 const createCssLoaderConfig = () =>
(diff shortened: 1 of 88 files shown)


webpack.config.js
```diff
@@ -46,29 +50,43 @@ if (LIFECYCLE_EVENT == 'test' || LIFECYCLE_EVENT == 'test:watch') {
...
     loader: 'sass-loader',
     options: {
-      sourceMap: true,
+      sourceMap: GENERATE_SOURCE_MAPS,
       includePaths: glob.sync('packages/*/node_modules').map((d) => path.join(__dirname, d)),
     },
   },
 ];
 
+const CSS_WRAPPED_IN_JS_CONFIG = [{loader: 'style-loader'}].concat(SASS_LOADER_CONFIG);
```
kfranqueiro: How come this was made into a constant when it's only used in one place, and the other side of the ternary where this is used doesn't also have a constant?  (Personally I'd vote for neither being a constant since it's just one place, but both being constants could also be valid.)

webpack.config.js
```diff
@@ -26,6 +26,10 @@ const OUT_PATH = path.resolve('./build');
 const PUBLIC_PATH = '/assets/';
 const IS_DEV = process.env.MDC_ENV === 'development';
 const IS_PROD = process.env.MDC_ENV === 'production';
+const IS_FAST_BUILD = process.env.MDC_BUILD === 'fast';
+const GENERATE_SOURCE_MAPS = IS_DEV && !IS_FAST_BUILD && process.env.MDC_GENERATE_SOURCE_MAPS !== 'false';
+const WRAP_CSS_IN_JS = process.env.MDC_WRAP_CSS_IN_JS !== 'false';
```
kfranqueiro: Is there a reason this doesn't get flipped by IS_FAST_BUILD?  As with my comment in #1128, I'm confused why we're adding this many separate environment variables.  I had assumed the point of adding IS_FAST_BUILD was to flip *all* the switches to go as fast as possible.

acdvorak: Removed `IS_FAST_BUILD` (both here and in PR #1128).

webpack.config.js
acdvorak: Inlined. Thanks!

acdvorak:
`MDC_WRAP_CSS_IN_JS=false` now disables source map generation unless you forcibly enable it with `MDC_GENERATE_SOURCE_MAPS=true`.

I've been thinking of these env vars as "experiment flags". They allow us to break up the work of improving build times into small, bite-sized PRs that are easier to review, and the changes they introduce are hidden from users by default.

IMHO, the env var is indeed still useful to me (though probably not to anyone else), because I want to continue iterating on this stuff without disrupting other developers.

webpack.config.js
```diff
@@ -189,7 +204,7 @@ module.exports = [{
     // In development, these are emitted as js files to facilitate hot module replacement. In
     // all other cases, ExtractTextPlugin is used to generate the final css, so this is given a
     // dummy ".css-entry" extension.
-    filename: '[name].' + (IS_PROD ? 'min.' : '') + 'css' + (IS_DEV ? '.js' : '-entry'),
+    filename: CSS_JS_FILENAME_OUTPUT_PATTERN,
```
kfranqueiro: The comment preceding this line mentions a dummy `.css-entry` extension but that doesn't seem to actually be employed anymore. Does it still need to be? Otherwise the comment needs to be updated.

acdvorak: Done. Good catch! 😄

kfranqueiro:
LGTMing this after discussing with Andy and understanding where this is a stepping stone towards.

PR #1135, opened 2017-08-16
Feat/textarea impl

amsheehan:


amsheehan:
DO NOT MERGE. This pull request only exists because the diff will be easier for the reviewer to parse than the one from `feat/textarea-impl` -> `master`

packages/mdc-textfield/README.md
```diff
@@ -340,19 +350,26 @@ complicated.
...
-| deregisterInputKeydownHandler(handler: EventListener) => void | Un-registers an event listener on the native input element for a "keydown" event |
+| deregisterInputKeydownHandler(handler: EventListener) => void | Deregisters an event listener on the native input element for a "keydown" event |
+| registerInputPointerDownHandler(handler: EventListener) => void | Registers an event listener on the native input element for a pointer event |
+| deregisterInputPointerDownHandler(handler: EventListener) => void | Deregisters an event listener on the native input element for a "pointer" event |
+| registerTransitionEndHandler(handler: EventListener) => void | Registers an event listener on the native input element for a "transitionend" event |
+| deregisterTransitionEndHandler(handler: EventListener) => void | Deregisters an event listener on the native input element for a "transitionend" event |
+| setBottomLineAttr(attr: string, value: string) => void | Sets an attribute with a given value on the bottom line element |
+| setHelptextAttr(name: string, value: string) => void | Sets an attribute on the help text element |
```
kfranqueiro: To be consistent with the previous line, "Sets an attribute with a given value on the help text element"

packages/mdc-textfield/README.md
```diff
@@ -340,19 +350,26 @@ complicated.
...
+| removeHelptextAttr(name: string) => void | Removes an attribute on the help text element |
```
kfranqueiro: on -> from

packages/mdc-textfield/foundation.js
```diff
@@ -43,6 +45,11 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
       deregisterInputInputHandler: (/* handler: EventListener */) => {},
       registerInputKeydownHandler: (/* handler: EventListener */) => {},
       deregisterInputKeydownHandler: (/* handler: EventListener */) => {},
+      registerInputPointerDownHandler: (/* evtType: string, handler: EventListener */) => {},
```
kfranqueiro: Why is this specifically named `registerInputPointerDownHandler` if it accepts any event type?  If we're going to do this, we might as well remove all of the specific `registerInputFooHandler` methods and just have `registerInputHandler(evtType, handler)`.  That would be an even bigger adapter change (but maybe a welcome one to reduce 6 methods to 2).

Also, your documentation in the README is inconsistent with this; the README doesn't mention it takes `evtType`.

packages/mdc-textfield/foundation.js
```diff
@@ -98,12 +127,24 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
     this.adapter_.removeHelptextAttr(ARIA_HIDDEN);
   }
 
+  transitionEnd_(evt) {
+    const {BOTTOM_LINE_ACTIVE} = MDCTextfieldFoundation.cssClasses;
+
+    // We need to wait for the bottom line to entirely transparent
+    // before removing the class. If we do not, we see the line start to
+    // scale down before disappearing - sheehana
```
kfranqueiro: Is it normal for us to put specific author names in comments? Not sure I've seen us do it before in this codebase.

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -28,6 +28,10 @@ $mdc-textfield-divider-on-light: rgba(black, .12);
 $mdc-textfield-divider-on-dark: rgba(white, .12);
 $mdc-textfield-disabled-border-on-light: rgba(#231f20, .26);
 $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
+$mdc-textfield-underline-on-light-idle: rgba(black, .5);
```
kfranqueiro: You list `light-idle` before `light`, but `dark` before `dark-idle`. Pick one plz?

packages/mdc-theme/_variables.scss
```diff
@@ -36,14 +36,20 @@ $mdc-theme-text-colors: (
     secondary: rgba(black, .54),
     hint: rgba(black, .38),
     disabled: rgba(black, .38),
-    icon: rgba(black, .38)
+    icon: rgba(black, .38),
+    textfield-background: rgba(black, .02),
```
kfranqueiro: Adding component-specific values into mdc-theme maps seems really confusing; prior to these textfield changes, these maps had *only* purely generic theme values.

I suspect that we would rather parameterize these via a mixin within textfield itself, given the newer direction that color theming is taking in general?  (cc @acdvorak)

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -242,57 +287,6 @@ $mdc-textfield-disabled-border-on-dark: rgba(white, .3);
...
-    height: 1px;
-    transform: scaleY(1);
-    transform-origin: center bottom;
-    transition:
-      background-color 33ms $mdc-animation-standard-curve-timing-function,
-      mdc-textfield-transition(transform);
-    background-color: rgba(black, .42);
-    pointer-events: none;
-
-    &::after {
```
kfranqueiro: I'm a little confused what happened to these `::after` styles?

packages/mdc-textfield/README.md
```diff
@@ -179,7 +188,7 @@ UX for client-side form field validation.
 ### Multi-line - With Javascript
```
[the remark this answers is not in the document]
amsheehan: There will be both multi-line text fields as well as `textarea`s

packages/mdc-textfield/foundation.js
amsheehan: In practice all it does take are pointer events—the name is there to reinforce that. I think we want to keep one method per registration. A user may need to override some subset of these and not all of them.

I agree with the documentation oversight and fixed it.

packages/mdc-textfield/foundation.js
amsheehan: I don't know if we've every had a rule one way or the other. I can take it out.

packages/mdc-textfield/mdc-textfield.scss
amsheehan: It became a real boy

packages/mdc-theme/_variables.scss
amsheehan: For the sake of merge sanity, since there was a ton of flux in theming at the time, lets merge this and circle back when we have a complete implementation and methodology for theming.

kfranqueiro: `mdc-theme-prop` can now accept color literals, so there shouldn't be a need to do this anymore?

PR #1157, opened 2017-08-18
feat(button): implement unelevated button

### LINKED ISSUE #986: Create new button style: unelevated button
- [ ] Create new unelevated button modifier
- [ ] Update README.md
- [ ] Update demo page

See prototype [here](https://codepen.io/yeelan0319/pen/jwQaxw)

yeelan0319:
Closes #986 
Also fixed theme color `accent` -> `secondary`

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -67,6 +67,15 @@
       .mdc-form-field {
         margin: 50px 0 -20px 30px;
       }
+      .hero {
+        position: relative;
+      }
+      .hero .note {
+        position: absolute;
+        bottom: 10px;
+        left: 36px;
+        color: rgba(33, 33, 33, .38);
+      }
 
     </style>
   </head>
@@ -90,6 +99,7 @@
         <button class="mdc-button mdc-button--raised">
           Raised
         </button>
+        <small class="note">Note: "secondary" was previously called "accent" in the Material spec.</small>
       </section>
 
       <section class="demo-wrapper">
@@ -136,7 +146,7 @@
               Primary
             </button>
             <button class="mdc-button mdc-button--accent">
-              Accent
+              Secondary
             </button>
             <a href="javascript:void(0)" class="mdc-button">
               Link
@@ -160,9 +170,33 @@
               Primary
             </button>
             <button class="mdc-button mdc-button--raised mdc-button--accent">
-              Accent
+              Secondary
             </button>
             <a href="javascript:void(0)" class="mdc-button mdc-button--raised">
+              Link
+            </a>
+          </div>
+        </fieldset>
+
+        <fieldset>
+          <legend class="mdc-typography--title">Unelevated Button</legend>
+          <div>
+            <button class="mdc-button mdc-button--unelevated">
+              Baseline
+            </button>
+            <button class="mdc-button mdc-button--unelevated mdc-button--compact">
+              Compact
+            </button>
+            <button class="mdc-button mdc-button--unelevated mdc-button--dense">
+              Dense
+            </button>
+            <button class="mdc-button mdc-button--unelevated mdc-button--primary">
+              Primary
+            </button>
+            <button class="mdc-button mdc-button--unelevated mdc-button--accent">
+              Secondary
+            </button>
+            <a href="javascript:void(0)" class="mdc-button mdc-button--unelevated">
               Link
             </a>
           </div>
@@ -185,7 +219,7 @@
               Primary
             </button>
             <button class="mdc-button mdc-button--accent" data-demo-no-js>
-              Accent
+              Secondary
             </button>
             <a href="javascript:void(0)" class="mdc-button" data-demo-no-js>
               Link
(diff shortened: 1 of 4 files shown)


packages/mdc-button/mdc-button.scss
```diff
@@ -86,14 +87,9 @@
 }
 
 .mdc-button--raised {
+  @include mdc-button-contained();
```
kfranqueiro: Including `mdc-button-contained` separately here and below leads to a lot of duplicated CSS in the output that I think we can avoid. What about the following:

```scss
.mdc-button--raised,
.mdc-button--unelevated {
  @include mdc-button-contained();
}

.mdc-button--raised {
  @include mdc-elevation(2);
  ...
}
```

packages/mdc-button/README.md
```diff
@@ -63,6 +63,15 @@ semantically correct.
...
 
+#### Unelevated Button
+
+```html
+<button class="mdc-button mdc-button--unelevated">
+  Unelevated button
+</button>
+```
+
+
```
kfranqueiro: Is this double-newline intentional?

packages/mdc-button/README.md
```diff
@@ -124,8 +133,9 @@ The provided modifiers are:
 
 | Class                 | Description                                             |
 | --------------------- | ------------------------------------------------------- |
+| `mdc-button--raised`  | A contained button that is elevated upon the surface.   |
+| `mdc-button--unelevated`  | A contained button that sticks to the surface.      |
```
kfranqueiro: Maybe "is flush with the surface" instead of "sticks to the surface"?

packages/mdc-button/README.md
yeelan0319: Mistake, nice catch!

packages/mdc-button/mdc-button.scss
yeelan0319: Just for record keeping: I was doing this because I was worried about parameterizing them separately. But what @kfranqueiro pointed out is this `scss` file will be reflecting only baseline and theme style anyway, so it should be less of a concern within this context.

yeelan0319:
I reverted the modifier class to accent but I still keep this PR referring to secondary color as it is how we call it now, but let me know if you have a different argument :)

packages/mdc-button/_mixins.scss
```diff
@@ -0,0 +1,54 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "@material/ripple/mixins";
+@import "@material/theme/mixins";
+@import "@material/theme/variables";
+
+@mixin mdc-button-with-colored-background() {
```
yeelan0319: I like the name of `mdc-button-filled`!
However, we cannot name the other one `mdc-button-text` as this is a type of button, so this name will be used as text button mixin together with `mdc-button-stroked`, `mdc-button-raised` and `mdc-button-unelevated`. Do you think `mdc-button-hollow` make sense?

yeelan0319:
@acdvorak I would like to hear your opinion how you think about referring `secondary`/`accent` color during this period. I don't hold a strong opinion either way, so I would like to make it go with theming doc and schedule so we keep it consistent thus less confusing?

packages/mdc-button/_mixins.scss
kfranqueiro: (For the record, previously discussed) I'm tentatively thinking `unfilled` for the opposite, but I'll think about it more between now and the next PR where it'll matter 😃

acdvorak:
My take is that we should be as consistent with the spec as possible, except in cases where doing so would break backward compatibility for a lot of people.

Since the spec now says "secondary" instead of "accent", I think it's reasonable to reword "accent" to "secondary" in most places.

However, we should keep the `--accent` modifier as-is for now (i.e., do not rename it), because many people rely on it.

It might be helpful to add a line of text to the readme and/or demo that says something like:

> "Secondary" was previously called "accent" in the Material spec.

yeelan0319:
@acdvorak Thanks for getting on this and I agree with the direction you take. I updated the README and demo page for a note and also pointed people in the README to mdc-theme so they can find most information there.

demos/button.html
```diff
@@ -90,6 +99,7 @@
         <button class="mdc-button mdc-button--raised">
           Raised
         </button>
+        <small class="note">Note: "secondary" was previously called "accent" in the Material spec</small>
```
kfranqueiro: Add period at end of sentence.

demos/button.html
```diff
@@ -156,7 +166,34 @@ <h1 class="mdc-typography--display2">Ripple Enabled</h1>
...
+          </div>
+        </fieldset>
+
+        <fieldset>
+          <legend class="mdc-typography--title">Unelevated Button</legend>
+          <div>
+            <button class="mdc-button mdc-button--unelevated">
+              Baseline
+            </button>
+            <a href="javascript:void(0)" class="mdc-button mdc-button--unelevated">
```
kfranqueiro: Note: "Link" buttons were moved to the end of these lines in #1176, and "Disabled" buttons were removed (replaced by a checkbox at the top of the demo).

packages/mdc-button/README.md
```diff
@@ -80,9 +88,13 @@ with and have no visual interaction effect.
 ### Colored
 
 MDC Buttons have a default baseline color, but it is also possible to adopt the
-application's primary or accent color by adding the `mdc-button--primary` or
+application's primary or secondary color by adding the `mdc-button--primary` or
 `mdc-button--accent` modifier.
 
+> Note: "Secondary" was previously called "accent" in the Material spec. See
+[mdc-theme](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme)
+for detail.
```
kfranqueiro: detail. -> details.

demos/button.html
```diff
@@ -67,6 +67,15 @@
       .mdc-form-field {
         margin: 50px 0 -20px 30px;
       }
+      .hero {
+        position: relative;
+      }
+      .hero .note {
+        position: absolute;
+        bottom: 10px;
+        left:36px;
```
acdvorak: Add a space after `:`

packages/mdc-button/_mixins.scss
```diff
@@ -0,0 +1,54 @@
...
+
+@mixin mdc-button-filled() {
+  @include mdc-ripple-base;
+  @include mdc-ripple-bg((pseudo: "::before", base-color: white, opacity: .32));
+  @include mdc-ripple-fg((pseudo: "::after", base-color: white, opacity: .32));
+  @include mdc-theme-prop(color, text-primary-on-dark);
+
+  background-color: black;
+
+  fieldset:disabled &,
```
kfranqueiro: Please test the demo after merging master into this branch again, and ensure this doesn't re-regress #1167.

kfranqueiro:
This LGTM now, but I think a manual rebase/resolution might be necessary due to the cherry-pick from #1176.

yeelan0319:
@kfranqueiro Thanks for helping out with the rebase 👍

PR #1166, opened 2017-08-22
refactor(fab): Move all FAB styles to mixins

acdvorak:
Uses config objects for theming mixins. This makes it easy to extend somebody else's theme values, overriding only specific properties or sub-properties.


yeelan0319:
@acdvorak I am not sure if the plan is get all theme changes into one branch and then squash merge that into master? The annoying thing is once diff against a branch, it seems you also set merge target to that branch, in this example to `refactor/theme/00-theme-prop-value`.

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,147 @@
...
+// @param {color|ThemeProp} $fab-theme-config.fg-color
+// @param {RippleConfig} $fab-theme-config.ripple-config @see mdc-ripple-default-config_ in mdc-ripple/_mixins.scss.
+// @param {color|ThemeProp} $fab-theme-config.tap-highlight-color
+// @param {int} $fab-theme-config.resting-elevation
+// @param {int} $fab-theme-config.active-elevation
+//*
+@mixin mdc-fab-theme($fab-theme-config) {
+  $bg-color: map-get($fab-theme-config, bg-color);
+  $fg-color: map-get($fab-theme-config, fg-color);
+  $ripple-config: map-get($fab-theme-config, ripple-config);
```
yeelan0319: Based on the state discussion last week we had, the ripple color(ink color) will always be the same as text color, which in this case I think it should be `fg-color`?

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,147 @@
...
+  $tap-highlight-color: map-get($fab-theme-config, tap-highlight-color);
```
yeelan0319: And this one I also imagine it should be same as ripple

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,147 @@
...
+  $resting-elevation: map-get($fab-theme-config, resting-elevation);
```
yeelan0319: I think this are part of base-style but not a thing that user can customize with

packages/mdc-fab/mdc-fab.scss
```diff
@@ -12,98 +12,42 @@
...
+@include mdc-fab-selector-disabled(#{$mdc-fab}) {
+  @include mdc-fab-theme--disabled(map-get($mdc-fab-theme-config-defaults, disabled));
 }
 
-.mdc-fab__icon {
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  width: 100%;
+#{$mdc-fab__icon} {
```
yeelan0319: Hmmmm, it seems an old style problem we have. I have two questions here
1. Does this needed to be a mixin? I feel this is a bit overkill and I wonder if it might be better if it is part of the base style?
2. Could we nest this icon into fab instead of having it as standalone? To follow: `modifier` leaves outside but `element` lives nested.

packages/mdc-theme/_mixins.scss
```diff
@@ -18,22 +18,38 @@
 
 // Applies the correct theme color style to the specified property.
 // $property is typically color or background-color, but can be any CSS property that accepts color values.
-// $style should be one of the map keys in $mdc-theme-property-values (_variables.scss).
+// $style should be one of the map keys in $mdc-theme-property-values (_variables.scss), or a literal color value.
```
yeelan0319: I think @kfranqueiro briefly touched base on this, is the plan to change this into custom map?

packages/mdc-fab/_mixins.scss
acdvorak: I'd prefer to only do refactoring in this CL, and leave those kinds of changes to Ken as part of his FAB improvements.

packages/mdc-fab/_mixins.scss
acdvorak: Done!

packages/mdc-fab/_mixins.scss
acdvorak: I'd prefer to only do refactoring in this CL, and leave those kinds of changes to Ken as part of his FAB improvements.

packages/mdc-theme/_mixins.scss
acdvorak: Done!

packages/mdc-fab/mdc-fab.scss
```diff
@@ -12,98 +12,58 @@
...
-    @include mdc-theme-prop(color, text-primary-on-light);
-    @include mdc-ripple-bg((pseudo: "::before"));
-    @include mdc-ripple-fg((pseudo: "::after"));
-  }
-
-  &:active,
-  &:focus {
-    outline: none;
-  }
+$mdc-theme-property-values-custom: map-merge(
```
kfranqueiro: I don't think we should be adding keys to this map for each component, but I realize this is here because we only want to look up parameterized colors via the map...

I realize this is currently here primarily for plain and disabled variants, both of which we're planning to remove for FAB (plain should be easily achievable via the theme mixin, and disabled FABs shouldn't really be a thing), though we will have to answer this question eventually as it relates to disabled buttons.

As for the tap highlight color, can we compute that based on the foreground color plus this opacity setting, so it doesn't need to be a parameter?

packages/mdc-fab/README.md
```diff
@@ -128,21 +128,25 @@ You can also do this declaratively when using the [material-components-web](../m
...
-### Modifier
+### CSS Classes
+
+CSS Class | Description
+--- | ---
+`mdc-fab--mini` | Makes the fab smaller (40 x 40 pixels)
+`mdc-fab--plain` | Makes the FAB have a white background
 
-The provided modifiers are:
+### Sass Functions
```
acdvorak: Done!

packages/mdc-fab/README.md
```diff
@@ -128,21 +128,25 @@ You can also do this declaratively when using the [material-components-web](../m
...
 
-| Class             | Description                             |
-| ------------------| --------------------------------------- |
-| `mdc-fab--mini`   | Makes the fab smaller (40 x 40 pixels). |
-| `mdc-fab--plain`  | Makes the FAB have a white background.  |
+Function | Description
```
acdvorak: Done!

packages/mdc-fab/mdc-fab.scss
acdvorak: Removed property map after discussing with the team offline. We will now allow directly passing arbitrary color values instead.

Tap highlight color should probably be computed, but I'd prefer to do that in a separate PR to keep this one small and focused.

packages/mdc-fab/mdc-fab.scss
```diff
@@ -12,98 +12,58 @@
...
+.mdc-fab {
+  @include mdc-fab-base;
+  @include mdc-fab-theme((
+    bg-color: secondary,
+    fg-color: text-primary-on-secondary,
+    ripple-config: (base-color: white, opacity: .16),
+    tap-highlight-color: fab-default-tap-highlight-color
+  ));
+
+  :disabled &,
```
acdvorak: I think we probably want users to be able to customize disabled colors. If their theme happens to look similar-ish to our default disabled colors, they would probably want to change disabled to differentiate it from enabled.

packages/mdc-fab/mdc-fab.scss
```diff
@@ -12,98 +12,58 @@
...
 
-  // This allows for using SVGs within them to align properly in all browsers.
-  // Can remove once: https://bugzilla.mozilla.org/show_bug.cgi?id=1294515 is resolved.
-
-  // stylelint-disable selector-max-type
-  // postcss-bem-linter: ignore
-  > svg {
-    width: 100%;
+  &__icon {
+    @include mdc-fab-icon;
```
acdvorak: As we discussed offline, we won't use `&__icon`, but will instead write out the full class name below (non-nested).

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,129 @@
...
+
+  // stylelint-disable selector-max-type
+  // postcss-bem-linter: ignore
+  > svg {
+    width: 100%;
+  }
+  // stylelint-enable selector-max-type
+}
+
+@mixin mdc-fab-mini {
```
acdvorak: As we discussed offline, we will keep the mixin, but rename it using BEM. We can revisit this decision later.

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,129 @@
...
+  @if $bg-color {
+    @include mdc-theme-prop(background-color, $bg-color);
+  }
+
+  @if $fg-color {
+    @include mdc-theme-prop(color, $fg-color);
+  }
+}
+
+@mixin mdc-fab-icon {
```
acdvorak: As we discussed offline, we will keep the mixin, but rename it using BEM. We can revisit this decision later.

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,129 @@
...
+  }
+  // stylelint-enable selector-max-type
+}
+
+@mixin mdc-fab-mini {
+  width: 40px;
+  height: 40px;
+}
+
+@mixin mdc-fab-disabled {
```
acdvorak: As we discussed offline, we will keep the mixin, but rename it using BEM. We can revisit this decision later.

packages/mdc-fab/README.md
```diff
@@ -128,21 +128,25 @@ You can also do this declaratively when using the [material-components-web](../m
...
-The provided modifiers are:
+### Sass Mixins
 
-| Class             | Description                             |
-| ------------------| --------------------------------------- |
-| `mdc-fab--mini`   | Makes the fab smaller (40 x 40 pixels). |
-| `mdc-fab--plain`  | Makes the FAB have a white background.  |
+Mixin | Description
+--- | ---
+`mdc-fab-base` | Generates CSS properties and element/modifier classes for a basic FAB without any explicit color or theming
```
yeelan0319: Correct me if I misunderstand the conversation yesterday but I think our plan is not to implement `mdc-fab-base` and include all base styles in `mdc-fab.scss` for the moments?

packages/mdc-fab/_mixins.scss
```diff
@@ -0,0 +1,129 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@import "@material/animation/functions";
+@import "@material/animation/variables";
+@import "@material/elevation/mixins";
+@import "@material/ripple/mixins";
+@import "@material/theme/mixins";
+
+@mixin mdc-fab-base_ {
```
yeelan0319: Yeah, this is the part I am confused.

packages/mdc-fab/README.md
acdvorak: After discussing offline: This mixin is now private, and has been removed from the README. We will still keep the mixin because it is useful for organizing our code, but we won't export it to clients.

packages/mdc-fab/_mixins.scss
acdvorak: After discussing offline: This mixin is now private, and has been removed from the README. We will still keep the mixin because it is useful for organizing our code, but we won't export it to clients.

packages/mdc-fab/README.md
```diff
@@ -128,21 +128,40 @@ You can also do this declaratively when using the [material-components-web](../m
...
+
+### Sass Mixins
+
+Mixin | Description
+--- | ---
+`mdc-fab-theme($config)` | Generates theme-related CSS properties from the given config map
+
+#### `mdc-fab-theme($config)`
+
+Generates theme-related CSS properties from the given config map.
```
kfranqueiro: If we're going to repeat its name/description here, maybe it doesn't make sense to bother with the table above?

acdvorak: Done!

PR #1182, opened 2017-08-24
fix(slider): Change how step value is initialized

### LINKED ISSUE #1173: Slider: Discrete slider not initializing current value

v. 0.0.18

1. Set aria-valuenow to a certain value
2. Initialize the Slider with autoinit or js
3. Check the initialized sliders value

The initial state of the slider is set to the value specified with aria-valuenow.

The initial value of the slider is shown (and initialized as) zero.

### Other information

The behavior can be observed in the demos as well. 
Thanks for your work on these great components!

archnode:
resolves #1173 

Micro-change, but the "setStep"-function actually triggers a "setValue" which sets aria-valuenow to null since the value is still uninitialized at this time.


archnode:
I signed, I signed ..

lynnjepsen:
This generally looks good, but please add a test to cover this case. 

And @archnode, our Googlebot still wants you to sign the CLA

PR #1098, opened 2017-08-08
fix: Correct unbounded ripple diameter

touficbatache:
Fixes and closes #1067

Issue #1067, opened 2017-08-03
Correct unbounded ripple diameter


touficbatache:
### What MDC-Web Version are you using?

> 0.16.0

### What browser(s) is this bug affecting?

> All browsers (Chrome newest)

### What OS are you using?

> All OS (Windows 7, Android 7)

### What are the steps to reproduce the bug?

> 1. Add a ripple to the toolbar icons
> 2. Observe the component's behavior
>
> https://codepen.io/anon/pen/YxGLQE

### What is the expected behavior?

> The ripple should be close to the size of the icon.

### What is the actual behavior?

> The ripple is a lot bigger than what it should be.

Garbee:
You can't just put the icon into a toolbar. There is a toolbar icon structure that is necessary for them to work right. Also, this is going to change as per #992 a little bit along with #993.

yeelan0319:
@Garbee is right that you should use `__icon` provided by toolbar instead of `icon-toggle` here. However, that being said, it seems like the `unbounded ripple` do have an incorrect radius, so we need go ahead and correct that implementation.

What need to be changed here:

- [ ] Correct the padding of right most `__icon` in toolbar, so it have the same padding with other icons and use margin to make the last icon is positioned correctly.
- [ ] Make the diameter of unbounded ripple from half diagonal of the bounding box to half length of the bounding box (see the image below)

<img width="119" alt="screen shot 2017-08-03 at 15 21 25" src="https://user-images.githubusercontent.com/1894234/28940008-36a378d6-7861-11e7-81a5-03df548a2f0a.png">
<img width="117" alt="screen shot 2017-08-03 at 15 30 07" src="https://user-images.githubusercontent.com/1894234/28940009-3810eae6-7861-11e7-9aaf-3a9bdae3d31a.png">

touficbatache:
I'm not using `icon-toggle` in the demo @yeelan0319

touficbatache:
I would be happy to make those changes. Would you want to guide me through it @yeelan0319?

touficbatache:
@yeelan0319 What do you mean by:
> "Make the diameter of unbounded ripple from half diagonal of the bounding box to half length of the bounding box"
Should I change the ripples size from the `foundation.js` file?

kfranqueiro:
Unfortunately, we're going to need to revert the PR, because it caused visible regressions on *bounded* ripples.  See the animated gif in #1183.

yeelan0319:
[...]

PR #1183, opened 2017-08-24
fix(ripple): Revert #1098 to fix bounded ripples

### LINKED ISSUE #1184: Ripple circle on `mdc-button` is oddly-shaped

0.18.1

All browsers that support ripple. Specifically, I tested in:

* Chrome 60.0.3112.90
* Safari 10.1.2
* Firefox 54.0.1

### What OS are you using?

macOS 10.12.6 (Sierra)

1. Edit `demos/button.html` to make the button text content wider. E.g.:
    ```html
    <button class="mdc-button">
      Baseline (very long text)
    </button>
    ```
2. Run the demo server
3. Go to localhost:8080/button.html
4. Click on the button whose text you made longer
5. Observe the ripple width

[CodePen](https://codepen.io/anon/pen/OjwjJJ)

I'm not sure exactly what the ripple should look like, but it's definitely not correct right now.

The `::before` pseudo-element is super wide:

image

kfranqueiro:
This reverts commit 0f1ca3598248c6df7fe456e198ce6519397b0247 and fixes #1184.

Unfortunately this commit caused a major regression for bounded ripples, such as those seen in the button and dialog demos.

dialogripples

### DIFF
--- a/packages/mdc-ripple/foundation.js
+++ b/packages/mdc-ripple/foundation.js
@@ -471,7 +471,7 @@ export default class MDCRippleFoundation extends MDCFoundation {
     this.frame_ = this.adapter_.computeBoundingRect();
 
     const maxDim = Math.max(this.frame_.height, this.frame_.width);
-    const surfaceDiameter = Math.min(this.frame_.width, this.frame_.height);
+    const surfaceDiameter = Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2));
 
     // 60% of the largest dimension of the surface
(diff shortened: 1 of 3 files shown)


touficbatache:
I don't suggest reverting, but I do suggest fixing this.

lynnjepsen:
@touficbatache I'm sorry we are reverting this. In the name of efficiency, our team favors quick rollbacks over "fix forward". But I would love for you could create a follow up PR that re-implements the fix to https://github.com/material-components/material-components-web/issues/1067, PLUS a fix for the errors seen here:

dialogripples

PR #1187, opened 2017-08-25
Fix toolbar title margin in rtl

mojighahar:
margin-left: 24px to margin-right: 24px


mojighahar:
I signed it!

packages/mdc-toolbar/mdc-toolbar.scss
```diff
@@ -84,6 +84,11 @@
     padding: $mdc-toobar-element-vertical-padding 0;
     line-height: 1.5rem;
     z-index: 1;
+
+    @include mdc-rtl {
```
lynnjepsen: Can you use `mdc-rtl-reflexive-box` mixin to handle all the margins for `.mdc-toolbar__title`?

PR #1191, opened 2017-08-26
docs: remove invalid 'default' attribute from examples

sivaraam:
The 'option' element doesn't have a 'default' attribute.

So, remove the attribute from examples to avoid promoting
non-conforming usage.


sivaraam:
Just to give the context, I found this while trying to validate a HTML document that used the MDC 'select' component.

PR #1197, opened 2017-08-28
fix(textfield): Update label position on value changed via JS

### LINKED ISSUE #822: Update textfield label position when value is changed from JavaScript

Not sure. Neither material-components-web.min.js nor material-components-web.css seem to contain a version number. But the issue looks to be the same on the master branch of this repository.

Tested on Firefox 54 and Chrome 59.

### What OS are you using?

Windows 10

I am using MDC-Web together with Angular (v1) following this guide:
https://material.io/components/web/docs/framework-integration/
And specifically this link from the guide:
https://plnkr.co/edit/b4v160c186ErrPG5vNza?p=preview

I have a MDC Textfield as described in the guide above.
When the value of the text field is updated from JavaScript via Angular, the position of the label does not change to reflect if the text field is empty or not.

1. open https://plnkr.co/edit/EMfKHSByb2Z7VaxBOS3K?p=preview
2. Type something in the second text box (the one that is not MDC styled)
3. Observe that the MDC styled input is updated, but the label and the value overlap each other.

When I look at https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/foundation.js it seems to set LABEL_FLOAT_ABOVE when the component is loaded and when focus is lost.
But there does not seem to be a way to tell it that the value has been changed from JavaScript, which should update LABEL_FLOAT_ABOVE.

There should be some API or other way for the framework to tell the component that the text field value has changed, and it should update its label position to match the updated value.

The field value and the field label are written on top of each other, making them unreadable.

anton-kachurin:
Fixes #822

### DIFF
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -357,7 +357,7 @@ class MDCTextFieldFoundation extends MDCFoundation {
     // We have to check for this descriptor, since some browsers (Safari) don't support its return.
     // See: https://bugs.webkit.org/show_bug.cgi?id=49739
     if (validDescriptor(desc)) {
-      const nativeInputElementDesc = ({
+      const nativeInputElementDesc = /** @type {!ObjectPropertyDescriptor} */ ({
         get: desc.get,
         set: (value) => {
           desc.set.call(nativeInputElement, value);
@@ -375,8 +375,8 @@ class MDCTextFieldFoundation extends MDCFoundation {
     const {INPUT_PROTO_PROP} = MDCTextfieldFoundation.strings;
     const nativeInputElement = this.getNativeInput_();
     const inputElementProto = Object.getPrototypeOf(nativeInputElement);
-    const desc = Object.getOwnPropertyDescriptor(inputElementProto, INPUT_PROTO_PROP);
-
+    const desc = /** @type {!ObjectPropertyDescriptor} */ (
+      Object.getOwnPropertyDescriptor(inputElementProto, INPUT_PROTO_PROP));
     if (validDescriptor(desc)) {
       Object.defineProperty(nativeInputElement, INPUT_PROTO_PROP, desc);
     }
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -239,16 +239,16 @@ class MDCTextFieldFoundation extends MDCFoundation {
     const {FOCUSED, LABEL_SHAKE} = MDCTextFieldFoundation.cssClasses;
     const input = this.getNativeInput_();
 
+    this.adapter_.removeClassFromLabel(LABEL_SHAKE);
+    this.updateDefaultValidity_();
+
     this.isFocused_ = false;
     this.adapter_.removeClass(FOCUSED);
-    this.adapter_.removeClassFromLabel(LABEL_SHAKE);
 
     if (!input.value && !this.isBadInput_()) {
       this.updateLabelFloat_();
       this.receivedUserInput_ = false;
     }
-
-    this.updateDefaultValidity_();
   }
 
   updateDefaultValidity_() {
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -251,6 +251,9 @@ class MDCTextFieldFoundation extends MDCFoundation {
     }
   }
 
+  /**
+   * @private
+   */
   updateDefaultValidity_() {
     if (!this.useCustomValidityChecking_) {
       const input = this.getNativeInput_();
@@ -281,6 +284,7 @@ class MDCTextFieldFoundation extends MDCFoundation {
    * Updates the state of the Text Field's helper text based on validity and
    * the Text Field's options.
    * @param {boolean} isValid
+   * @private
    */
   updateHelperText_(isValid) {
     const {HELPER_TEXT_PERSISTENT, HELPER_TEXT_VALIDATION_MSG} = MDCTextFieldFoundation.cssClasses;
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -196,9 +196,6 @@ class MDCTextFieldFoundation extends MDCFoundation {
       const {LABEL_FLOAT_ABOVE, LABEL_SHAKE} = MDCTextfieldFoundation.cssClasses;
       const input = this.getNativeInput_();
 
-      this.adapter_.removeClassFromLabel(LABEL_SHAKE);
-      this.showHelptext_();
-
       if (input.value) {
         this.adapter_.addClassToLabel(LABEL_FLOAT_ABOVE);
       } else {
@@ -206,6 +203,8 @@ class MDCTextFieldFoundation extends MDCFoundation {
       }
 
       if (!this.useCustomValidityChecking_) {
+        this.adapter_.removeClassFromLabel(LABEL_SHAKE);
+        this.showHelptext_();
         this.changeValidity_(input.checkValidity());
       }
     }
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -79,6 +79,10 @@ class MDCTextFieldFoundation extends MDCFoundation {
     this.receivedUserInput_ = false;
     /** @private {boolean} */
     this.useCustomValidityChecking_ = false;
+    /** @private {boolean} */
+    this.labelFloatIsActive_ = false;
+    /** @private {boolean} */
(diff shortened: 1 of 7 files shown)


lynnjepsen:
Sorry @anton-kachurin, we did a lot of text field updates in a big change, and it has caused merge conflicts in this PR. Can you resolve them? Or it might be better to just create a whole new PR.

touficbatache:
@anton-kachurin What's up with this? Would love to see it implemented in MDC

anton-kachurin:
@touficbatache I've been waiting for a review for longer than a month and the activity log clearly shows that this PR isn't abandoned. At this very moment I'm working on resolving merge conflicts.

I don't consider your comment appropriate. This is an open source project, and if you personally would love to see something in it, go ahead and implement it. You always have an option to fork from my fork and take the branch to the point where it can be merged to master. Don't press on volunteers, we're doing it a spare time.

As a part of the community, please be respectful to others.

touficbatache:
I'm sorry, but I'm not being disrespectful. I'm just telling that your work is great by saying that I'd love seeing it implemented, I didn't say that to put pressure on anybody. Have a great day.

gjdev:
AFAIK this pull request only updates the label-float style after a programmatic change of the input value. However, a change of the input value may affect other styles when the new value introduces or fixes a previous input value error. Hence value changes may also affect the error styles, helptext, and errortext.

packages/mdc-textfield/foundation.js
```diff
@@ -189,6 +191,30 @@ class MDCTextfieldFoundation extends MDCFoundation {
...
+      if (input.value) {
+        this.adapter_.addClassToLabel(LABEL_FLOAT_ABOVE);
+      } else {
+        this.adapter_.removeClassFromLabel(LABEL_FLOAT_ABOVE);
+      }
+
+      if (!this.useCustomValidityChecking_) {
+        this.adapter_.removeClassFromLabel(LABEL_SHAKE);
+        this.showHelptext_();
+        this.changeValidity_(input.checkValidity());
```
anton-kachurin: @gjdev Validation is handled in this PR

gjdev: Sorry I missed that somehow.

lynnjepsen:
Hey @anton-kachurin, I wrote a proposal for a different way to solve #822 (in the #822 thread). Can you check it out and comment on that thread. I'd like to hash out some code architecture details before I can review any code changes to solve #822.

packages/mdc-textfield/foundation.js
```diff
@@ -190,6 +194,27 @@ class MDCTextfieldFoundation extends MDCFoundation {
...
+   * Changes the floating label's state
+   * @private
+   */
+  updateLabelFloat_() {
+    const input = this.getNativeInput_();
+    const {LABEL_FLOAT_ABOVE} = MDCTextfieldFoundation.cssClasses;
+
+    if (this.isFocused_ || input.value) {
+      if (!this.labelFloatIsActive_) {
+        this.labelFloatIsActive_ = true;
```
lynnjepsen: instead of you using this.labelFloatIsActive as a internal state variable to keep track of "if the LABEL_FLOAT_ABOVE class has been added". Why not add a method to the adapter to check `adapter_.labelHasClass(LABEL_FLOAT_ABOVE)`?

packages/mdc-textfield/foundation.js
```diff
@@ -190,6 +194,27 @@ class MDCTextfieldFoundation extends MDCFoundation {
...
+        this.adapter_.addClassToLabel(LABEL_FLOAT_ABOVE);
+      }
+    } else {
+      if (this.labelFloatIsActive_) {
```
lynnjepsen: you actually don't even need this check. It doesnt matter what the internal state of this.labelFloatIsActive is....if neither `this.isFocused_` or `input.value` is true, then you should remove LABEL_FLOAT_ABOVE from the label element.

packages/mdc-textfield/foundation.js
```diff
@@ -220,19 +245,27 @@ class MDCTextfieldFoundation extends MDCFoundation {
...
     const input = this.getNativeInput_();
 
+    this.adapter_.removeClassFromLabel(LABEL_SHAKE);
+    this.updateDefaultValidity_();
+
     this.isFocused_ = false;
     this.adapter_.removeClass(FOCUSED);
-    this.adapter_.removeClassFromLabel(LABEL_SHAKE);
 
     if (!input.value && !this.isBadInput_()) {
```
lynnjepsen: should this.isBadInput be checked within the logic of updateLabelFloat?

anton-kachurin:
@lynnjepsen Regarding your questions above:

1) Why these conditions were added:
```
      if (!this.labelFloatIsActive_) {
        this.labelFloatIsActive_ = true;
        this.adapter_.addClassToLabel(LABEL_FLOAT_ABOVE);
      }
```
```
      if (this.labelFloatIsActive_) {
        this.labelFloatIsActive_ = false;
        this.adapter_.removeClassFromLabel(LABEL_FLOAT_ABOVE);
      }
```

Before, the `LABEL_FLOAT_ABOVE` class was only added/removed on focus/blur events. These event always come in pairs, so we could assume that if 'onfocus` is processing, the label *is not* floating at the moment. And vice versa. 

Now we extend the foundation with the `setValue` method, and this method may be called at any arbitrary moment. That means the following situations should be assumed:
- The user focuses the input, then blurs it without entering any value, then `setValue('')` is called. In pseudocode:
```
addClass(FLOAT);
removeClass(FLOAT);
removeClass(FLOAT);
```
- The user focuses the input, then `setValue('not empty') is called:
```
addClass(FLOAT);
addClass(FLOAT);
```
The flag `this.labelFloatIsActive_` is here for filtering those duplicate calls for adding/removing the class.

The one main reason why I think it's better to double check before calling `adapter_.(add/remove)ClassFromLabel` is the **custom adapters**. While the default adapter implementation will discard duplicate `addClass`/`removeClass` calls with no side effects, we can't be sure that the custom ones will do the same. That's why in both cases the condition preventing duplicate calls was added. Please let me know if it should stay or not.

2) Why the boolean flag (`this.labelFloatIsActive_`) to maintain the internal state and not `adapter_.labelHasClass(LABEL_FLOAT_ABOVE)`:

My argument was that the `LABEL_FLOAT_ABOVE` is a completely internal state but then I realized that it is not:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#pre-filled-text-fields

With this in mind, we have to add `adapter_.labelHasClass` method.

3) `isBadInput`:
https://codepen.io/anon/pen/xPqPKO?editors=1010

`this.updateLabelFloat` is called in two situations:
- The user manipulates with the input (onfocus/onblur)
- The value is set programmatically (setValue)

The pen above shows that setting the `input.value = 'invalid value'` programmatically would never cause `input.validity.badInput` be false. That's why we only need to check `isBadInput` on blur

packages/mdc-textfield/foundation.js
```diff
@@ -110,6 +112,7 @@ class MDCTextfieldFoundation extends MDCFoundation {
       this.adapter_.registerTextFieldInteractionHandler(evtType, this.textFieldInteractionHandler_);
     });
     this.adapter_.registerTransitionEndHandler(this.transitionEndHandler_);
+    this.installPropertyChangeHooks_();
```
lynnjepsen: Add these `installPropertyChangeHooks` in the *next* PR. Its a little too hard to review otherwise. And I think that means you don't need the `this.setValueIsProcessing_` field.

packages/mdc-textfield/foundation.js
```diff
@@ -247,8 +280,11 @@ class MDCTextfieldFoundation extends MDCFoundation {
     if (isValid) {
       this.adapter_.removeClass(INVALID);
     } else {
-      this.adapter_.addClassToLabel(LABEL_SHAKE);
       this.adapter_.addClass(INVALID);
+
+      if (this.isFocused_) {
```
lynnjepsen: is there a test for this case? A test for when the text fields value is changed programmatically, but the user is not focused on that text field. You should verify that `this.adapter_.addClassToLabel(LABEL_SHAKE)` is not called.

anton-kachurin:
Unfortunately, multiple merge conflict do not allow me to keep working on this PR. Besides, the main reason for me to work on it was to implement a specific feature. Instead, I was asked to refactor the code and not to include the feature, and now all this work is gone.

anton-kachurin:
The work is continued in #1615

jordyboytjuh:
Is this issue also related to the new labels of the mdc-select (MDC 0.27.0)? When I set the selectedIndex over there on init, the label is written over the selected value.

PR #1203, opened 2017-08-28
docs(layout-grid): Update the margin recommendation.

### LINKED ISSUE #1036: Update README for device margin recommendation
- [ ] Update layout grid margin recommendation table.

yeelan0319:
Closes #1036

### DIFF
--- a/packages/mdc-layout-grid/README.md
+++ b/packages/mdc-layout-grid/README.md
@@ -72,7 +72,7 @@ By default, `mdc-layout-grid` behaves like a fluid container, which takes up its
 
 Layout grids set default margins and gutters to 24px on desktop, 16px on tablet and phone.
 
-The Material Design spec recommends 8px/16px/24px/40px as the padding sizes to choose from and recommended up to 160px as the margin sizes. However, MDC layout grid doesn't impose any restrictions.
+The Material Design spec recommends 8px/16px/24px/40px as the padding sizes to choose from and recommends up to 160px as the margin sizes. However, MDC layout grid doesn't impose any restrictions.
 
 
 #### CSS custom properties
--- a/packages/mdc-layout-grid/README.md
+++ b/packages/mdc-layout-grid/README.md
@@ -72,7 +72,7 @@ By default, `mdc-layout-grid` behaves like a fluid container, which takes up its
 
 Layout grids set default margins and gutters to 24px on desktop, 16px on tablet and phone.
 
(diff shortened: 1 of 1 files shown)


packages/mdc-layout-grid/README.md
```diff
@@ -72,7 +72,7 @@ By default, `mdc-layout-grid` behaves like a fluid container, which takes up its
 
 Layout grids set default margins and gutters to 24px on desktop, 16px on tablet and phone.
 
-The Material Design spec recommends 8px, 16px, 24px or 40px as the sizes to choose from, we set those as choices in our demo catalog. However, MDC layout grid doesn't impose any restrictions.
+The Material Design spec recommends 8px/16px/24px/40px as the padding sizes to choose from and recommended up to 160px as the margin sizes. However, MDC layout grid doesn't impose any restrictions.
```
kfranqueiro: "recommended" -> "recommends"

packages/mdc-layout-grid/README.md
```diff
@@ -72,7 +72,7 @@ By default, `mdc-layout-grid` behaves like a fluid container, which takes up its
 
 Layout grids set default margins and gutters to 24px on desktop, 16px on tablet and phone.
 
-The Material Design spec recommends 8px, 16px, 24px or 40px as the sizes to choose from, we set those as choices in our demo catalog. However, MDC layout grid doesn't impose any restrictions.
+The Material Design spec recommends 8px/16px/24px/40px as the padding sizes to choose from and recommends up to 160px as the margin sizes. However, MDC layout grid doesn't impose any restrictions.
```
kfranqueiro: Should this say gutter sizes instead of padding sizes? Using the term padding here is confusing, since that has a meaning in CSS and AFAIK the gutters aren't padding in the CSS sense.

Also, is "up to 160px" correct for margin sizes? That seems very large and is a significant change from the currently-published [guidelines](https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-grid) (and our [demo page](http://material-components-web.appspot.com/layout-grid.html)) which suggest the same set of valid sizes for both margins and gutters (8, 16, 24, 40).

yeelan0319: Yeah. I get this feedback from designer in the final review last month:
> nice to have more options for the margin (e.g. up to 160dp, but not discrete values)

I think 160dp is the largest we can go, but that might only be applicable to large screen application with sparse information density. In case you are interested in, check out the meeting note for layout I sent you on slack channel.

PR #1211, opened 2017-08-29
refactor(fab): Break theme mixin into three mixins, to decouple code

lynnjepsen:
There's no good explanation for what `config` was other than 'a collection of background, foreground, and ripple colors. From an API perspective, there is no reason to group the three properties together. The logic has no dependency on each other, as you can see in the old `mdc-fab-theme` implementation.

Andy mentioned that `config`... 'gives clients the ability to create extensible themes that can be "inherited" by other themes (similar to Android)'. But when I talked to designers, it was difficult to see a use case where `config` would be used by any other component. If a client is already customizing the color of FABs container, ink, and ripple, they will also probably customize the color of a Menu's container, ink, and ripple with _slightly different_ colors.

I also updated the names, a.k.a. background -> container and foreground -> ink, after talking to the designers. This matches the spec better.

### DIFF
--- a/packages/mdc-fab/mdc-fab.scss
+++ b/packages/mdc-fab/mdc-fab.scss
@@ -19,11 +19,9 @@
 
 .mdc-fab {
   @include mdc-fab-base_;
-  @include mdc-fab-theme((
-    bg-color: secondary,
-    fg-color: text-primary-on-secondary,
-    ripple-config: (base-color: white, opacity: .16)
-  ));
+  @include mdc-fab-container-color(secondary);
+  @include mdc-fab-ink-color(text-primary-on-secondary);
+  @include mdc-fab-ripple((base-color: white, opacity: .16));
 
   &:not(.mdc-ripple-upgraded) {
     -webkit-tap-highlight-color: rgba(0, 0, 0, .18);
@@ -31,11 +29,9 @@
 }
 
 .mdc-fab--plain {
-  @include mdc-fab-theme((
-    bg-color: white,
-    fg-color: text-primary-on-light,
-    ripple-config: ()
-  ));
+  @include mdc-fab-container-color(white);
+  @include mdc-fab-ink-color(text-primary-on-light);
+  @include mdc-fab-ripple(()); //default dark ripple
 }
 
 .mdc-fab--mini {
--- a/packages/mdc-fab/mdc-fab.scss
+++ b/packages/mdc-fab/mdc-fab.scss
@@ -31,7 +31,7 @@
 .mdc-fab--plain {
   @include mdc-fab-container-color(white);
   @include mdc-fab-ink-color(text-primary-on-light);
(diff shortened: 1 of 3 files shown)


amsheehan:
I agree with Ken. These, lets call them higher order mixins™, don't make a ton of sense to me. They make less sense since all they're doing is wrapping a single `mdc-theme` mixin in most of these cases. Why wouldn't our advanced users just use the `mdc-theme` mixin? I guess for users that want to change one of these properties for _all_ of their fabs at once?

lynnjepsen:
> Properties that are co-dependent on each other - with one theme mixin taking a config object, it has all the theming information together to potentially make informed decisions regarding how properties should influence or "inherit" from each other

These mixins are not co-dependent on each other so that each aspect of the FAB can be tweaked. They do not make any informed decisions on how properties might influence each other. It is up to an advanced user to "pick container, ink, and ripple colors that work together, and meet accessibility standards".

_Other_ mixins can make informed decisions, like changing the ink color to meet accessibility standards (see [this example](https://github.com/material-components/material-components-web/commit/caac30993c094d8309e22de913b448bd975cf7f7) or changing the ripple color to match the ink color.

I worry that if one mixin tries to serve all the use cases, it will stop being maintainable code. We've all heard it before...code should do One Thing.

> Mixins for different types of the same component (e.g. button with text/stroked/raised/unelevated) where some properties don't make sense in some contexts (e.g. stroke should never be combined with raised/unelevated)

Will take a look at Button here in a sec.

> Perhaps less importantly, if we have components that already possess mixins, this may make it harder to see at a glance which mixins are theming-related and which ones serve a different purpose

I actually think the names `mdc-fab-container-color` is more descriptive than `mdc-fab-theme`.

> I was under the impression we wanted a consistent approach to theme-related mixins across components, in which case I don't want us to lose sight of more complex concerns given that FAB is pretty simple.

Let me try this on Button too...

> They make less sense since all they're doing is wrapping a single mdc-theme mixin in most of these cases. Why wouldn't our advanced users just use the mdc-theme mixin

`mdc-fab-container-color` only requires the user to specify color. The `mdc-theme-prop` mixin requires the user to specify the property and the color. `mdc-fab-container-color` is providing a layer of abstraction, so the user only needs to provide the color. It allows us to change the CSS implementation of a FAB, without breaking any clients.

packages/mdc-fab/mdc-fab.scss
```diff
@@ -19,23 +19,19 @@
...
 
 .mdc-fab--plain {
-  @include mdc-fab-theme((
-    bg-color: white,
-    fg-color: text-primary-on-light,
-    ripple-config: ()
-  ));
+  @include mdc-fab-container-color(white);
+  @include mdc-fab-ink-color(text-primary-on-light);
+  @include mdc-fab-ripple(()); //default dark ripple
```
kfranqueiro: Nit: space after `//`

Issue #1192, opened 2017-08-26
Slider keeps following touches while scrolling on mobile devices


daniel-mf:
After sliding, the sliders keep actively tracking the touches until you give the document a single tap.

Tested in Chrome for Android and OSX
[slider demo](https://material-components-web.appspot.com/slider.html)

kfranqueiro:
Good catch; I was able to reproduce this in Chrome on Android, and on desktop using device emulation.

Worth noting, this only seems to happen when you actually drag to slide the control; if you simply tap the control, it doesn't seem to happen.

kfranqueiro:
The PR filed against this (#1214) seems to suggest that we may be leaking event handlers, or hooking up too many to begin with. This is further corroborated by adding logging statements to downHandler, moveHandler, and upHandler.

On Chrome desktop with device emulation, I get the following behaviors...

When tapping (no bug):

* downHandler (pointerdown)
* downHandler (touchstart)
* upHandler (pointerup)
* upHandler (touchend)
* downHandler (mousedown)
* upHandler (mouseup)

When dragging (bug):

* downHandler pointerdown
* downHandler touchstart
* moveHandler pointermove
* ...
* moveHandler touchmove
* ...
* upHandler touchend

For some reason, no pointerup event ever gets handled, so we end up with a leaked handler. This is odd, since in Chrome on desktop without emulating mobile, both mouse and pointer events occur consistently.

Moreover, while I can reproduce this on Chrome on Android, I couldn't reproduce it on Firefox on Android, and Alex wasn't able to reproduce it on Safari on iOS either.

I'm wondering at this point if this is a bug specific to Chrome on Android. However, I think we might be able to avoid the issue if we store an instance property to manage these handlers when active is first set, and avoid hooking up more than one set of handlers at a time? I'm open to thoughts on this.

daniel-mf:
@kfranqueiro thanks for the help with the PR!

I looked deeper and I noticed that `downHandler` is only deregistered in `destroy` event. So whenever body receives a `touchstart`, `downHandler` gets called, which registers `moveHandler` and `upHandler` again. So I don't see a reason `this.setActive_(false);` would stop `moveHandler` from firing. Am I missing something?

kfranqueiro:
[...]

PR #1214, opened 2017-08-30
Fix for #1192

daniel-mf:


daniel-mf:
I signed it!

kfranqueiro:
Thanks for taking a whack at this. Upon taking a closer look, however, I think this PR hints at a deeper issue - theoretically, we shouldn't need to check `active_` within these handlers because that suggests that `setActive_(false)` has already been called, which means these event handlers shouldn't be hooked up anymore in the first place. I think we're actually leaking event handlers (or hooking up duplicates to begin with).

I'm going to put more info into the original issue, so we can continue discussion there and come up with a fix for the root problem. Thanks again for the PR, because it provided a lead towards what's really going on.

PR #1223, opened 2017-08-31
chore(switch): Remove unused and undocumented mixins

lynnjepsen:


yeelan0319:
@lynnjepsen Nice catch 👍 

As side note related to this, I looked up `@function` in our codebase, it seems like the following module also put `@function` into `_mixin.scss`:
- mdc-elevation
- mdc-layout-grid
- mdc-ripple

Does it worth put a **tech-debt** story in tracker to correct this?

PR #1226, opened 2017-08-31
refactor(switch): Reference animation timing function directly

lynnjepsen:
This is the only place where we actually use mdc-animation mixins. I think referencing the animation variable directly is more clear.

### DIFF
--- a/packages/mdc-checkbox/_keyframes.scss
+++ b/packages/mdc-checkbox/_keyframes.scss
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 
+@import "@material/animation/variables";
 @import "@material/theme/mixins";
 @import "./variables";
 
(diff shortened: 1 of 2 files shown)


kfranqueiro:
I was wondering that too. Just remember, if we nuke it we need to mark it a BREAKING CHANGE.

kfranqueiro:
Oh. Also, we do reference the mixins from [mdc-animation.scss](https://github.com/material-components/material-components-web/blob/master/packages/mdc-animation/mdc-animation.scss) so we'd have to make the same changes there if we were to remove the mixins completely.

yeelan0319:
The mixin is also referred in `mdc-checkbox` although it is not being used, I suggested we remove it in PR [#1225](https://github.com/material-components/material-components-web/pull/1225#pullrequestreview-60245351). 

IMHO, this mixin created in the way is not very helpful in implementing other components, so it might worth consider removing/refactoring it. Most component I saw either implemented their own animation function, see [linear progress](https://github.com/material-components/material-components-web/blob/master/packages/mdc-linear-progress/_keyframes.scss) or instead using `_function.scss` directly.

PR #1232, opened 2017-09-02
fix(layout-grid): import the variables in the mixin

sivaraam:
While at it, improve the grammar of a sentence.


packages/mdc-layout-grid/README.md
```diff
@@ -215,7 +215,7 @@ When your contents need extra structure that cannot be supported by single layou
 
 The nested layout grid behaves exactly like when they are not nested, e.g, they have 12 columns on desktop, 8 columns on tablet and 4 columns on phone. They also use the **same gutter size** as their parents, but margins are not re-introduced since they are living within another cell.
 
-However, Material guideline do not recommend have a deeply nested grid since it might means a over complicated UX.
+However, Material guideline does not recommend to have a deeply nested grid as it might mean a over complicated UX.
```
touficbatache: That might be better :smiley:: "However, Material guideline does not recommend having a deeply nested grid as it might mean an over complicated UX."

sivaraam: Yeah seems better.

packages/mdc-layout-grid/_mixins.scss
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+@import "./variables"
+
 // returns the lower grid boundary or null if the smallest grid is selected
```
amsheehan: Looks like the linter is complaining here

`17:2  ✖  Expected indentation of 2 spaces   indentation`

sivaraam: I don't think I get that. What should be fixed?

touficbatache: You should indent the line 17 by 2 spaces. Not more, not less.

sivaraam: That's surprising I thought it was correctly indented after seeing it. Moreover, I didn't touch that line at all in the first place.

touficbatache: Maybe someone edited it in the past but it got accepted even with the error

sivaraam: I can't take this. This is so odd. Why doesn't the linter warn about a missing indentation at [line 27](https://github.com/sivaraam/material-components-web/blob/15be7c54edf3d0a6d8b7e70cd18045997d92064a/packages/mdc-layout-grid/_mixins.scss#L27) or [line 39](https://github.com/sivaraam/material-components-web/blob/15be7c54edf3d0a6d8b7e70cd18045997d92064a/packages/mdc-layout-grid/_mixins.scss#L39) which has a similar comment to this line?

amsheehan: Your text editor may have parsed and saved it incorrectly. You can always run `npm run lint` to see what will run on TravisCI.

sivaraam: Ok, I found the issue. I missed a semicolon on line 15.

PR #1240, opened 2017-09-05
style: Specify where box-sizing is in the order

lynnjepsen:


kfranqueiro:
Do you think it might make sense to put this above width/height stuff since it can influence their meaning?

PR #1241, opened 2017-09-05
feat(fab): Implement enter/exit transitions

kfranqueiro:


packages/mdc-fab/README.md
```diff
@@ -141,7 +141,8 @@ in all browsers. See [this Stackoverflow post](http://stackoverflow.com/posts/35
 
 CSS Class | Description
 --- | ---
-`mdc-fab--mini` | Makes the fab smaller (40 x 40 pixels)
+`mdc-fab--exited` | When this class is added, the FAB will shrink out of view (exit);<br>when it is removed, the FAB will grow into view (enter)
```
lynnjepsen: mdc-fab-withdrawn??

drop the 'when this class is added'...I think thats implied with CSS classes

how about 'makes the FAB withdraw from view. When `mdc-fab--withdrawn` is removed, the FAB will return to view.

You don't have to mention grow/shrink, since I think thats an implementation detail

packages/mdc-fab/_mixins.scss
```diff
@@ -40,7 +40,10 @@
   width: 56px;
   height: 56px;
   padding: 0;
-  transition: box-shadow 280ms $mdc-animation-standard-curve-timing-function;
+  transition:
+    box-shadow 280ms $mdc-animation-standard-curve-timing-function,
```
lynnjepsen: should this be 270ms?

packages/mdc-fab/_mixins.scss
```diff
@@ -40,7 +40,10 @@
...
+    opacity 15ms linear 30ms,
+    mdc-animation-enter(transform, 270ms);
```
lynnjepsen: is 270ms here the 180ms + 90ms from below? If so, maybe reference some private variables directly

packages/mdc-fab/_mixins.scss
```diff
@@ -96,4 +99,19 @@
...
   width: 100%;
+  will-change: transform;
+  transition: mdc-animation-enter(transform, 180ms, 90ms);
+}
+
+@mixin mdc-fab--exited_ {
+  transform: scale(0);
+  transition:
+    opacity 15ms linear 150ms,
+    mdc-animation-exit-permanent(transform, 180ms);
```
lynnjepsen: is this the same 180ms from above? If so, maybe reference a private variable directly

packages/mdc-fab/README.md
kfranqueiro: I'll admit that in plain English, "withdrawn" sounds less weird than "exited", but I picked "exited" because "exit" is the specific word we use to refer to this motion, and it is a word that is meaningful to animators. WDYT?

packages/mdc-fab/_mixins.scss
kfranqueiro: This is existing code and matches `$mdc-elevation-transition-duration`, but I'm not sure why we aren't just using that variable directly, so I'll go fix that :)

packages/mdc-fab/_mixins.scss
kfranqueiro: The specs don't necessarily directly call this out (which is why this didn't occur to me), but the idea makes sense to me since the two motions do line up.

packages/mdc-fab/_mixins.scss
kfranqueiro: I'm not sure that these are meant to be strongly correlated (one is the duration of just the FAB's icon transitioning in, while the other is the duration of the entire FAB transitioning out). I'll ping one of our motion designers to find out. For now I'd rather err on the side of not factoring these together.

kfranqueiro: I got confirmation that these two should not be correlated to a common variable, so I'll leave my updated PR as-is, only correlating the enter durations.

packages/mdc-fab/README.md
lynnjepsen: talked offline, --exited makes sense

PR #1242, opened 2017-09-06
chore(animation): Removing mixins and CSS classes

lynnjepsen:
BREAKING CHANGE: Removes mdc-animation mixins and CSS classes, please reference mdc-animation Sass variables directly.


kfranqueiro:
I know we discussed removing the mixins, but are we sure about removing the CSS classes as well, especially WRT non-Sass users?

Either way this definitely needs to include BREAKING CHANGE when we commit this.

PR #1244, opened 2017-09-06
feat(button): create theme mixin for button

yeelan0319:
# Overview

## Private base mixins 
Those styles that irrelevant to theme, such as `padding` etc.. It will be used in `button.scss` as well as cases where user don't want to inherit our baseline but define their own.
- mdc-button-base_
- mdc-button--raised_
- mdc-button--compact_
- mdc-button--dense_

## Public theme mixins
- mdc-button-container-fill-color
- mdc-button-stroke-color
- mdc-button-ink-color
- mdc-button-ripple (current it also define `tap-highlight`. This will be addressed in separate PR in state work)
- mdc-button-stroke-width
- mdc-button-stroke-style
- mdc-button-corner-radius

## A few notes:
1. `@at-root #{selector-append("a", &)}` besides `&:enabled` is to support link button since technically speaking link doesn't have `disabled` or `enabled` state. (In FF/Safari, they won't apply `&:enabled` style to link button
2. As we using `enabled`, I also corrected `disabled` selector by removing `fieldset` parent selector. This will not affect what we have for fieldset because based on what mentioned in [HTML5 Spec](https://dev.w3.org/html5/spec-preview/the-fieldset-element.html#attr-fieldset-disabled)
> The disabled attribute, when specified, causes all the form control descendants of the fieldset element, excluding those that are descendants of the fieldset element's first legend element child, if any, to be disabled.
3. For the same reason, background for disabled text button and stroked button are also specified explicitly now.
4. Compared the generated `mdc.button.css`. Besides selector name and order changes, the `tap-highlight` color is now correct to use ripple color.

It might be extra, but I also deployed to our internal development server [mdc-web-dev] to help for a quick review.


packages/mdc-button/_mixins.scss
```diff
@@ -17,25 +17,120 @@
...
-    background-color: white;
+@mixin mdc-button-ripple($ripple-config) {
+  @at-root #{selector-append("a", &)}, &:enabled {
+    @include mdc-ripple-base;
+    @include mdc-ripple-bg(map-merge((pseudo: "::before"), $ripple-config));
+    @include mdc-ripple-fg(map-merge((pseudo: "::after"), $ripple-config));
 
     &:not(.mdc-ripple-upgraded) {
-      -webkit-tap-highlight-color: rgba(black, .18);
+      $tap-highlight-color: rgba(map-get($ripple-config, base-color), map-get($ripple-config, opacity));
```
kfranqueiro: Is it really safe to assume that both base-color and opacity are set? mdc-ripple/_mixins.scss has defaults for these but here we have nothing to fall back on.

packages/mdc-button/mdc-button.scss
```diff
@@ -23,102 +23,51 @@
...
-    @include mdc-ripple-base;
-    @include mdc-ripple-bg((pseudo: "::before", base-color: white, opacity: .16));
-    @include mdc-ripple-fg((pseudo: "::after", base-color: white, opacity: .16));
-    @include mdc-theme-prop(color, text-primary-on-dark);
-
-    &:not(.mdc-ripple-upgraded) {
-      -webkit-tap-highlight-color: rgba(white, .18);
-    }
+    @include mdc-button-ink-color(text-primary-on-dark);
+    @include mdc-button-ripple((base-color: white, opacity: .16));
```
kfranqueiro: Should we make variables for the various ripple configs we reuse in baseline styles?

packages/mdc-button/mdc-button.scss
```diff
@@ -195,9 +146,9 @@
 }
 
 .mdc-button--stroked {
-  fieldset:disabled &,
   &:disabled {
     border-color: rgba(black, .38);
+    background-color: transparent;
```
kfranqueiro: Is this not already covered by the style in .mdc-button above?

packages/mdc-button/_mixins.scss
yeelan0319: I think that is a totally valid concern. 
My suggestion is when we implement state, we move this css fall back as part of ripple. At the end of the day, this is css only fallback for ripple, which might not belong to this place in the first place? What do you think?

packages/mdc-button/mdc-button.scss
yeelan0319: Could we consider do that in a separate PR? 
We might consider this PR is to implementing the mixin. I appended this to the list of follow up PRs to make sure we come to this thing later.

kfranqueiro: Discussed with Yiran offline and it makes sense to look at this later, pending design discussions to nail down the exact values and which values are common.

kfranqueiro:
FWIW the switch from 2 selectors to `:not(:disabled)` alleviates the increase in CSS size I observed before (in addition to avoiding the issue Lynn pointed out with using these mixins within descendant selectors). When gzipped, it's actually even 2 bytes smaller than master. 🤔

yeelan0319:
I also compared the size of the newly generate css file.

To explain the actual size being larger but gzip smaller, I think there are two factors playing there:

(1) The new version has more selectors since we nested the `:not(:disabled)` selector within the mixin v.s. originally this two thing will mixed together. Here is an example of that:
```
/* Code generated by container fill mixin*/
.mdc-button--raised:not(:disabled),
.mdc-button--unelevated:not(:disabled) {
  background-color: black; }

/* Code generated by ink color mixin*/
.mdc-button--raised:not(:disabled),
.mdc-button--unelevated:not(:disabled) {
  /* @alternate */
  color: white;
  color: var(--mdc-theme-text-primary-on-dark, white); }
```

(2) On the opposite, get rid of `fieldset :disabled` half the selectors we needed for any of our disabled style and make the file shorter.

Based on gzip's description, they use **DEFLATE** algorithm, which
> ### Duplicate string elimination
> Within compressed blocks, if a duplicate series of bytes is spotted (a repeated string), then a back-reference is inserted, linking to the previous location of that identical string instead. An encoded match to an earlier string consists of an 8-bit length (3–258 bytes) and a 15-bit distance (1–32,768 bytes) to the beginning of the duplicate. Relative back-references can be made across any number of blocks, as long as the distance appears within the last 32 KB of uncompressed data decoded (termed the sliding window).
> 
> If the distance is less than the length, the duplicate overlaps itself, indicating repetition. For example, a run of 10 identical bytes can be encoded as one byte, followed by a duplicate of length 9 beginning 1 byte ago.

So duplicate selectors generated by nested mixin should be able to be optimized by their algorithm as some of them are pretty close to each other (I mean expect one really), so the decrease in that meant we benefit from removing the fieldset and haven't create too much of a concern by introducing nested mixins.

PR #1249, opened 2017-09-06
chore(fab): Remove the mdc-fab--plain modifier

### LINKED ISSUE #1143: Update FAB color to primary color
- [x] Create color mixins
- [ ] Delete --plain modifier
- [ ] Include "breaking change" message about removing modifiers, point to color mixin as deprecation replacement.

lynnjepsen:
Resolves #1143, Update demo to use the mdc-fab-accessible mixin, since this is the migration path away from mdc-fab--plain.

BREAKING CHANGE: Removes mdc-fab--plain, please update your code to use mdc-fab-accessible mixin instead.


demos/fab.html
```diff
@@ -93,52 +90,84 @@
...
+        <legend>FABs with Ripple</legend>
+        <div class="demo-fabs">
+          <figure>
+            <button class="mdc-fab material-icons" aria-label="Favorite">
+              <span class="mdc-fab__icon">
+                favorite_border
+              </span>
+            </button>
+            <figcaption>
+              <div>Themed FAM</div>
```
kfranqueiro: Typo: FAM -> FAB

(Insert "it's lit" meme here)

demos/fab.html
```diff
@@ -93,52 +90,84 @@
...
+            </figcaption>
+          </figure>
+          <figure>
+            <button class="mdc-fab mdc-fab--mini material-icons" aria-label="Favorite">
+              <span class="mdc-fab__icon">
+                favorite_border
+              </span>
+            </button>
+            <figcaption>
```
kfranqueiro: Was your thought that including these captions on the first set of buttons is sufficient? (Just noting that they're only present on the ripple set, not the CSS-only set.)

demos/demos.scss
```diff
@@ -23,6 +23,14 @@
 $mdc-theme-primary: #212121 !default; // Grey 900
 $mdc-theme-secondary: #64dd17 !default; // Light Green A700
 
+@import "../packages/mdc-theme/color_palette";
+@import "../packages/mdc-fab/mixins";
+
+
```
kfranqueiro: Not sure this double-empty-line was intentional?

demos/fab.html
lynnjepsen: Yes, I think it is sufficient. (And whenever I copy paste stuff across sections I run the risk of forgetting to update it in a later PR.)

kfranqueiro: Yup, can definitely agree with that logic and figured that might be part of the reason :)

joshiste:
this wasn't updated in the components docs

kfranqueiro:
Thanks @joshiste; the docs mirrored on material.io should be up to date now.

PR #1250, opened 2017-09-06
style: Specify where cursor properties are in the order

lynnjepsen:


packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -33,13 +33,13 @@
   transform: translate(-50%, 100%);
   transition: mdc-animation-exit-permanent(transform, .25s);
   background-color: $mdc-snackbar-background-color;
+  pointer-events: none;
```
kfranqueiro: This specific change seems unfortunate since it "splits up" the background-color style and the dark override for it, but I'm not sure it's avoidable...

PR #1252, opened 2017-09-07
docs(fab): Update README to match best practices

lynnjepsen:


lynnjepsen:
Please wait for #1241 before merging

packages/mdc-fab/README.md
```diff
@@ -39,68 +37,71 @@ npm install --save @material/fab
...
+### CSS Classes
 
-By default the FAB rests in the page where it sits in the content flow.
-Developers must position it as needed within their application's design.
+CSS Class | Description
+--- | ---
+`mdc-fab` | Mandatory, for the button element
+`mdc-fab__icon` | Mandatory, for the icon element
+`mdc-fab--mini` | Optional, modifies the FAB to a smaller size
+`mdc-fab--exited` | Optional, animates the FAB out of view.<br>When this class is removed, the FAB will return to view.
```
kfranqueiro: Thanks for already factoring this in here. :sunny:

packages/mdc-fab/README.md
```diff
@@ -39,68 +37,71 @@ npm install --save @material/fab
...
 
-### Mini
+### HTML Structure
 
 ```html
-<button class="mdc-fab mdc-fab--mini material-icons" aria-label="Favorite">
+<button class="mdc-fab material-icons" aria-label="Foo label">
   <span class="mdc-fab__icon">
-    favorite
+    foo_icon
```
kfranqueiro: Is there a reason not to use a valid material-icons value here? (We also still reference `favorite` in the absolute-positioning example.)

packages/mdc-fab/README.md
```diff
@@ -39,68 +37,71 @@ npm install --save @material/fab
...
   </span>
 </button>
 ```
 
-### Absolutely positioned
+### CSS Classes
```
kfranqueiro: Is it normal for this to be as far removed from the Sass sections of the readme as it is? (e.g. should we consider moving the Absolutely Positioned and Adding MDC Ripple sections further down below the Sass stuff?)

lynnjepsen: I think users will use HTML and CSS. Then they will probably also want to position their FAB. They will probably also want to add a Ripple, though maybe not if JS is too heavy. The least likely thing they want to do is customize the FAB's color (least likely after the positioning and ripple). So I've sorted the features to the top which I think most users will want.

PR #1256, opened 2017-09-07
feat(button): Add mdc-button-filled-accessible mixin

yeelan0319:


packages/mdc-button/_mixins.scss
```diff
@@ -113,6 +113,22 @@
   line-height: $mdc-dense-button-height;
 }
 
+@mixin mdc-button-filled-accessible($container-color) {
```
lynnjepsen: $container-fill-color

lynnjepsen: I made the mistake in the FAB PR to not add demos at the same time I added the public mixin. Can you add a demo of using mdc-button-filled-accessible to the demo page in this PR?

yeelan0319: Good point. Let me confirm with @kfranqueiro if this will conflict with his work for separating `demos.scss`.

yeelan0319:
Good point about demo the usage. Let me confirm with @kfranqueiro tomorrow morning if this will conflict with his work for separating `demos.scss`.

kfranqueiro:
Don't worry about my demo splitting stuff. Presumably you'll be adding to demos.scss in this PR, and I'll bring any of those additions into button.scss in my branch when I rebase (since I'm not at the point of sending a PR yet).

PR #1262, opened 2017-09-07
fix(slider): Add two test cases to cover give default value to step for discrete slider

yeelan0319:
Following up for PR #1245 to capture issue mentioned in #1173


test/unit/mdc-slider/foundation.test.js
```diff
@@ -117,6 +117,21 @@ test('#init checks if slider is discrete and if display track markers', () => {
   raf.restore();
 });
 
+test('#init sets step to one if slider is discrete but with step equals to zero', () => {
```
kfranqueiro: nit: shorten "but with step equals to zero" -> "but step is zero"

yeelan0319:
Thanks for fixing my English 🥇 
Ready for a second pass.

PR #1265, opened 2017-09-08
style: Order our mixins above declarations

lynnjepsen:


.stylelintrc.yaml
```diff
@@ -18,6 +18,16 @@ rules:
   # Double-quotes are our convention throughout our codebase within (S)CSS. They also reflect how
   # attribute strings are normally quoted within the DOM.
   string-quotes: double
+  order/order:
+    -
```
kfranqueiro: Sorry if I'm being ignorant of some feature of yaml, but what happened with these dashes on their own lines here?

packages/mdc-ripple/_mixins.scss
```diff
@@ -107,7 +109,9 @@
 
       content: "";
     }
-  } @else {
+  }
+
+  @else {
```
kfranqueiro: Is this really our standard practice for `@else` blocks (on their own line with an empty line in between), and was this performed by our fix task when updating for this PR or is it unrelated?

Tangentially, it appears we have 3 different styles of `@else` - cuddled, on its own line with no empty line in between, and this. Personally I find the empty line a bit confusing since the `@else` is conceptually connected to what's above it and empty lines usually serve the purpose of disconnecting from what's above.

amsheehan: I personally don't love this style. Is there a style guide you can point to that recommends it?

lynnjepsen: Agreed, this is a weird format change. I think it snuck it here with an auto-fix.

I think I need to make another PR that lints our @if and @else (a la https://github.com/kristerkari/stylelint-scss/blob/master/docs/examples/if-else.md). Then we can cuddle our @elses across the code base

.stylelintrc.yaml
lynnjepsen: the dashes create lists, and non dash is like a key in a map. Its basically JSON, but weird (I think).

And I need this format in order for the stylelint-order plugin to parse it

packages/mdc-toolbar/_mixins.scss
```diff
@@ -25,17 +25,19 @@
...
 
   border: none;
   background-color: inherit;
   color: inherit;
   text-decoration: none;
+
+  // stylelint-disable at-rule-empty-line-before, block-closing-brace-newline-after
+  @if $padding-type == menu {
+    padding: $mdc-toobar-element-vertical-padding;
+  }
```
kfranqueiro: It looks like you covered the existing ones, but should we un-space this new if/else too?

PR #1267, opened 2017-09-08
feat(tabs): tab bar scroller foundation - publicize scrollToTabAtIndex method

danielraggs:
This pull request removes the underscore from the method `scrollToTabAtIndex_()` on the `MDCTabBarScroller` foundation so that the framework using it can call this method and scroll to a specific tab if needed. Additional documentation has also been added to the README for MDC tabs.

**Example use case:**
A tab bar with tabs that can become active by clicking a different button elsewhere on the page. We need to communicate to the scroller to bring an active tab into view if it is hidden.


packages/mdc-tabs/tab-bar-scroller/foundation.js
```diff
@@ -144,16 +144,16 @@ export default class MDCTabBarScrollerFoundation extends MDCFoundation {
 
     for (let i = 0; i < this.adapter_.getNumberOfTabs(); i++) {
       const tabOffsetLeftAndWidth =
-        this.adapter_.getComputedLeftForTabAtIndex(i) + this.adapter_.getComputedWidthForTabAtIndex(i);
+          this.adapter_.getComputedLeftForTabAtIndex(i) + this.adapter_.getComputedWidthForTabAtIndex(i);
```
kfranqueiro: I'm not sure why there are indentation changes in this commit, but they seem unrelated to what was actually changed. Can you revert them?

packages/mdc-tabs/README.md
```diff
@@ -637,6 +637,10 @@ Scrolls the tab bar such that the rightmost tab traverses the scroll frame and b
 
 > **NOTE:** Due to a quirk in event behavior, we allow the rightmost tab to be partially occluded even when tabbed to because clicking on such an element would shift the frame on the `focus` event. This would result in a scenario where the ripple persists and the intended tab would not be selected due to the tab bar shifting before the `mouseup` or `click` events get dispatched.
 
+#### MDCTabBarScrollerFoundation.scrollToTabAtIndex(index) => index
```
kfranqueiro: This method doesn't return anything, right? So, `=> void`?

danielraggs:
@kfranqueiro Thanks for the review, I have made your suggested changes. Not sure how those indentations got in there.

packages/mdc-tabs/README.md
```diff
@@ -637,6 +637,10 @@ Scrolls the tab bar such that the rightmost tab traverses the scroll frame and b
 
 > **NOTE:** Due to a quirk in event behavior, we allow the rightmost tab to be partially occluded even when tabbed to because clicking on such an element would shift the frame on the `focus` event. This would result in a scenario where the ripple persists and the intended tab would not be selected due to the tab bar shifting before the `mouseup` or `click` events get dispatched.
 
+#### MDCTabBarScrollerFoundation.scrollToTabAtIndex(index) => void
```
kfranqueiro: Sorry for not noticing this at the same time as my other comment here, but we need to specify the type of the parameter here: `(index: number)` (I guess that might've gotten mixed up into the return value earlier).

danielraggs: Thanks, good catch. I have updated this as well.

kfranqueiro:
Thanks! 👍

PR #1268, opened 2017-09-08
docs(toolbar): Fix off-by-one error in toolbar demo RTL buttons

kfranqueiro:
Fixes a bug introduced in #1068, where a new example was added but the index passed to toggleRTL wasn't properly updated in the added example or the ones appearing after it.

This fixes the issue by removing the need to pass the correct index altogether, so this problem should not happen again.

### DIFF
--- a/demos/toolbar/index.html
+++ b/demos/toolbar/index.html
@@ -118,65 +118,61 @@
 
       <section class="examples">
         <div class="example">
-          <h2>Normal Toolbar <button type="button" onclick="toggleRTL(0)">Toggle RTL</button></h2>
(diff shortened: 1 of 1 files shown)


kfranqueiro:
I had been wondering which way to do that and thought I saw other fix(demos) but maybe that was only when things touch multiple demos.

I was thinking it made more sense to list demo fixes separately since it doesn't indicate a bug in the component, only in the component's demo page. I guess we can discuss that later though. I'm sure I've already made this mistake in at least one commit message though.

kfranqueiro:
Thanks for the comment Yiran, and sorry I misread it initially on Friday. I updated the commit message on the branch and the title in this PR.

PR #1269, opened 2017-09-08
docs(button): Make call in stylesheet button.html to create primary and secondary example

yeelan0319:
I renamed `accent` to correctly using `secondary` in the new class selector.
In the next PR, I will remove the stale modifier.

I split `stroked` and `text` button in order to show how to completely customize the style of each button type, if that makes sense.

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -141,9 +141,6 @@
             <button class="mdc-button mdc-button--dense">
               Dense
             </button>
-            <button class="mdc-button primary-text-button">
-              Primary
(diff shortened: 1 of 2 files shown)


kfranqueiro:
My comment in chat was about possibly combining some of the selectors, but Yiran had a good point in intentionally organized everything separately to provide standalone SCSS for each example, which is clearer than saving bytes in a way that would force users to hunt into selectors used for multiple examples.  Otherwise this LGTM.

Issue #991, opened 2017-07-20
Parameterize buttons


yeelan0319:
# Customizable attributes
|Mixins|
| --- |
|$mdc-button-ink-color|
|$mdc-button-ripple|
|$mdc-button-container-fill|
|$mdc-button-stroke-color|
|$mdc-button-corner-radius|
|$mdc-button-stroke-width|

See prototype [here](https://codepen.io/yeelan0319/pen/jwQaxw)

# Break up with the following PRs
- [x] Create theme mixin for button #1244 
- [x] Move disabled style into private base mixin since it is not customizable #1255 
- [x] Add `accessibility` mixin #1256 
- [x] Reorder the mixins #1271 
- [x] Make call in stylesheet `button.html` to create primary and secondary example #1269 
- [x] Remove the primary and accent modifier and update documents #1270 
- [x] Make reused styles into variables #1280 
- [x] Make button support icon #1281
- [x] House cleaning for `--primary` and `--secondary` #1312 
- [x] Retract `button-style` for the moment #1305 
- [x] Make corner radius back to `2px` #1307 
- [x] Add stroke-width demo #1307 
- [x] Add icon usage in README #1313 
- [x] Add a line of experimental in demo page #1307

PR #1270, opened 2017-09-08
refactor(button): Remove primary and accent modifier

yeelan0319:
BREAKING CHANGE: Remove support of `mdc-button--primary` and `mdc-button--accent` modifier classes. For custom and theme button implementation, use button mixins instead. See `demos.scss` for details.


yeelan0319:
## Related to #991 

I updated the doc to meet the best practice as in #1252

packages/mdc-button/README.md
```diff
@@ -41,47 +41,30 @@ npm install --save @material/button
...
+### CSS Classes
 
-```html
-<button class="mdc-button mdc-button--stroked">
-  Stroked button
-</button>
-```
+| Class                 | Description                                             |
+| --------------------- | ------------------------------------------------------- |
+| `mdc-button`          | A text button that is flush with the surface.           |
```
kfranqueiro: Should this be documented as Mandatory and the others documented as Optional?

yeelan0319: That's a great question!
IMHO, the html structure have already demonstrated this point, so I might not feel confused. But it should be useful to someone new to this library, I will add it.

kfranqueiro: The reason I mentioned both "Mandatory" and "Optional" is it would match the practice e.g. in the FAB docs: https://github.com/material-components/material-components-web/tree/master/packages/mdc-fab#css-classes

PR #1274, opened 2017-09-11
chore: Use stylelint --fix instead of stylefmt

lynnjepsen:
stylefmt was formatting our code in weird ways, and stylelint --fix is more in sync with our `npm run lint`


kfranqueiro:
If you run `npm i` does it update `package-lock.json` to remove dependencies? If so, that should be included in this PR.

Also, does running `stylelint --fix` produce no changes within our codebase as-is right now?

lynnjepsen:
> If you run npm i does it update package-lock.json to remove dependencies? If so, that should be included in this PR.

Fixed!

> Also, does running stylelint --fix produce no changes within our codebase as-is right now?

Yes...I got all the changes done in the previous PRs.

PR #1276, opened 2017-09-11
chore: Update npm-run-all to the latest version

lynnjepsen:


package.json
```diff
@@ -71,7 +71,7 @@
     "mkdirp": "^0.5.1",
     "mocha": "^3.2.0",
     "node-sass": "^4.0.0",
-    "npm-run-all": "^3.1.2",
+    "npm-run-all": "^4.1.0",
```
yeelan0319: Nit: Might consider bumping it to `^4.1.1` since the only changes introduced by this [patch](https://github.com/mysticatea/npm-run-all/releases/tag/v4.1.1) version is to fix an bug?

> fbd4024 fixed a regression bug of --print-name.

PR #1280, opened 2017-09-12
refactor(button): Make ripple configure using variables

yeelan0319:
## Related to #991 

This PR is made to make customize button with mixins easier:
1. Expose `filled` and `unfilled` button ripple opacity value, so they could use those value if they only wish to adapt ripple color to custom ink color 
2. Allow theme-variable to be passed as `base-color` in `mdc-button-ripple`, so that `mdc-button-filled-accessible` will ensure ink color is always aligned with ripple color to prevent potential error.

*Note*: It worths considering move `tap-color` as well as `base-color` logic into ripple, but I suggest we do that separately when we refactor ripple. I added a todo here to indicate this is something that we should revisit.

### DIFF
--- a/demos/button.scss
+++ b/demos/button.scss
@@ -17,17 +17,18 @@
 @import "./common";
 @import "../packages/mdc-button/mdc-button";
 @import "../packages/mdc-button/mixins";
+@import "../packages/mdc-button/variables";
 @import "../packages/mdc-checkbox/mdc-checkbox";
 @import "../packages/mdc-form-field/mdc-form-field";
 @import "../packages/mdc-typography/mdc-typography";
 
 .mdc-button.secondary-text-button {
   @include mdc-button-ink-color($mdc-theme-secondary);
-  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: .16));
+  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: $mdc-unfilled-button-ripple-opacity));
 
   @include mdc-theme-dark(".mdc-button") {
     @include mdc-button-ink-color($mdc-theme-secondary);
-    @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: .16));
+    @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: $mdc-unfilled-button-ripple-opacity));
   }
 }
 
@@ -42,12 +43,12 @@
 .mdc-button.secondary-stroked-button {
   @include mdc-button-ink-color($mdc-theme-secondary);
   @include mdc-button-stroke-color($mdc-theme-secondary);
-  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: .16));
+  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: $mdc-unfilled-button-ripple-opacity));
 
   @include mdc-theme-dark(".mdc-button") {
     @include mdc-button-ink-color($mdc-theme-secondary);
     @include mdc-button-stroke-color($mdc-theme-secondary);
-    @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: .16));
+    @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: $mdc-unfilled-button-ripple-opacity));
   }
 }
 
--- a/demos/button.scss
+++ b/demos/button.scss
@@ -24,11 +24,12 @@
 
 .mdc-button.secondary-text-button {
   @include mdc-button-ink-color($mdc-theme-secondary);
-  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: $mdc-unfilled-button-ripple-opacity));
+  @include mdc-button-ripple((theme-style: secondary, opacity: $mdc-unfilled-button-ripple-opacity));
 
   @include mdc-theme-dark(".mdc-button") {
+    // TODO(yiranm): Revisit ripple initialization after ripple refactoring
     @include mdc-button-ink-color($mdc-theme-secondary);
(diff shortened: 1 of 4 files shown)


packages/mdc-button/_mixins.scss
```diff
@@ -55,6 +55,14 @@
 
 @mixin mdc-button-ripple($ripple-config) {
   &:not(:disabled) {
+    // TODO(yiranm): Move base-color and tap-highlight-color into ripple
```
kfranqueiro: Currently this code is only needed because we're always relying on `base-color` in the ripple config map (we're using it for `$tap-highlight-color` below).

However, Yiran and I did a brief dive to investigate why we're even setting `$tap-highlight-color` specifically here, when ripple ordinarily sets it to fully transparent. `-webkit-tap-highlight-color` only affects behavior of links on mobile, so this results in `<a class="mdc-button">` elements having a more stark effect upon being tapped on mobile browsers. However, this also makes them behave inconsistently with all other MDC Buttons, whereas removing the `-webkit-tap-highlight-color` here would make them completely consistent. We already have focused and active states for CSS-only buttons, so the buttons should still appear responsive to user interaction.

IMO the ideal scenario here is we could just switch from using `base-color` to `theme-style` above (which already expects a property from the theme map), but we're thinking of checking with someone more experienced in a11y to be sure that removing `-webkit-tap-highlight-color` here is okay first.

packages/mdc-button/_variables.scss
```diff
@@ -16,3 +16,6 @@
 
 $mdc-button-height: 36px;
 $mdc-dense-button-height: 32px;
+
+$mdc-unfilled-button-ripple-opacity: .16;
```
lynnjepsen: You should reference these in our demo.scss as well.

packages/mdc-button/_mixins.scss
```diff
@@ -26,10 +26,10 @@
 
   @if ($light-or-dark == "light") {
     @include mdc-button-ink-color(text-primary-on-dark);
-    @include mdc-button-ripple((base-color: white, opacity: .32));
+    @include mdc-button-ripple((base-color: text-primary-on-dark, opacity: $mdc-filled-button-ripple-opacity));
```
lynnjepsen: Can you do

@include mdc-button-ripple((base-color: map-get($mdc-theme-property-values, $text-primary-on-dark), opacity: $mdc-filled-button-ripple-opacity))

And then remove the @if statement from mdc-button-ripple

yeelan0319: We could do this. Though the argument I have here is since `theme` already agnostically taking either hex value or theme string, it makes it feel inconsistent from the standpoint of API.

On the other end, the other argument we had in the morning is it might worth we address it later in the ripple work and instead of `base-color` use `theme-color` temporarily since it is an existing API.

packages/mdc-button/_variables.scss
yeelan0319: Good point!

packages/mdc-button/_mixins.scss
lynnjepsen: I think using`theme-style` in the existing ripple config is better.

mdc-button-ripple should just take whatever "ripple config" we would use in mdc-ripple-* mixins

kfranqueiro: The only thing preventing us from doing that right now is we're relying directly on the presence of `base-color` in this mixin, and `base-color` and `theme-style` are generally mutually exclusive, but yes, we'd like to ideally stop needing to do that. See my [other comment](https://github.com/material-components/material-components-web/pull/1280#discussion_r138396949)

yeelan0319:
Designer said "links-masquerading-as-buttons need to be differentiated visually. As long as the link has sufficient contrast and behaves like a button with touch feedback, a11y users will be fine. "

So I am going to move forward with our original plan.
1. Remove `tap-highlight-color` from our mixin (and style no highlight in base mixin)
2. Change `base-color` into `theme-color`
3. Remove the `@if` statement

Sounds good?

demos/button.scss
```diff
@@ -17,17 +17,18 @@
...
 @import "../packages/mdc-button/mixins";
+@import "../packages/mdc-button/variables";
 @import "../packages/mdc-checkbox/mdc-checkbox";
 @import "../packages/mdc-form-field/mdc-form-field";
 @import "../packages/mdc-typography/mdc-typography";
 
 .mdc-button.secondary-text-button {
   @include mdc-button-ink-color($mdc-theme-secondary);
-  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: .16));
+  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: $mdc-unfilled-button-ripple-opacity));
```
kfranqueiro: Come to think of it, is it possible to use `theme-style` instead of `base-color` for all of these?  This would be `theme-style: secondary` rather than `base-color: $mdc-theme-secondary`, right?

yeelan0319: Yeah. I kept this one because I thought it might be a good idea to demonstrate both method (with a theme value or with a HEX value), but I don't hold a very strong opinion to that if you insist?

How about if I add a comment here say:
> Besides we would also like to demonstrate the you could use HEX value or variables represent a HEX value to initialize a ripple

How do you feel this way?

kfranqueiro: Since we're using actual theme slots here I'd sooner show us doing it the right way, but if you want we could add a comment above like...

```
// These examples use theme-style to map to a specific theme color.
// You can also specify any hex color via base-color instead of theme-style.
// See MDC Ripple documentation for more information on the ripple config map.
```

yeelan0319: Addressed this offline. The plan is to make ripple support HEX and theme value agnostically as theme do right now. I will left a todo there to make sure we revisit this parts once we start working on ripple.

PR #1281, opened 2017-09-12
feat(button): Support icon in button

yeelan0319:
## Related to #991 

[Codepen Demo](https://codepen.io/yeelan0319/pen/pWzxYZ)

Dev server: 1281-dot-mdc-web-dev.xxxx.com

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -20,10 +20,10 @@
     <title>Button - Material Components Catalog</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="icon" type="image/png" href="/images/logo_components_color_2x_web_48dp.png">
-    <script src="/assets/button.css.js"></script>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto+Mono">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <script src="/assets/button.css.js"></script>
     <style>
       .demo-wrapper {
         padding-bottom: 100px;
--- a/demos/button.html
+++ b/demos/button.html
@@ -147,6 +147,10 @@
             <a href="javascript:void(0)" class="mdc-button">
               Link
             </a>
+            <button class="mdc-button">
+              <i class="material-icons mdc-button__icon">favorite</i>
+              Icon
+            </button>
           </div>
         </fieldset>
 
@@ -168,6 +172,10 @@
             <a href="javascript:void(0)" class="mdc-button mdc-button--raised">
               Link
             </a>
+            <button class="mdc-button mdc-button--raised">
+              <i class="material-icons mdc-button__icon">favorite</i>
+              Icon
+            </button>
           </div>
         </fieldset>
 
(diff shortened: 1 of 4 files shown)


demos/button.html
```diff
@@ -20,10 +20,10 @@
     <title>Button - Material Compoonents Catalog</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="icon" type="image/png" href="/images/logo_components_color_2x_web_48dp.png">
-    <script src="/assets/demo-styles.css.js"></script>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto+Mono">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <script src="/assets/demo-styles.css.js"></script>
```
kfranqueiro: I'm not sure why this line was moved in this PR, but it will no longer exist the next time you merge with master anyway. The styles have been split out, so now the ones for this demo are in `button.scss`.

demos/button.html
```diff
@@ -147,6 +147,10 @@ <h1 class="mdc-typography--display2">Ripple Enabled</h1>
             <a href="javascript:void(0)" class="mdc-button">
               Link
             </a>
+            <button class="mdc-button">
```
kfranqueiro: We wanted to keep link last since it's the only one that doesn't get disabled with the checkbox at the top of the demo page, right?

packages/mdc-button/_mixins.scss
```diff
@@ -130,6 +130,15 @@
   }
 }
 
+@mixin mdc-button__icon_ {
+  display: inline-block;
+  width: 16px;
+  height: 16px;
+  margin-right: 8px;
+  font-size: 16px;
+  vertical-align: middle;
```
kfranqueiro: These icons don't actually look like they're vertically centered; their tops align, but they look like they're extending below the baseline, which is going to look awkward at least with all-caps text like our buttons typically have. I'm not sure what the design goal is here though - technically right now I think our button text vertical placement is centered disregarding anything hanging below the baseline.

Also, I'm wondering how this will interact with dense button, or is it not intended to ever be combined with that?

demos/button.html
yeelan0319: I will rebase this PR on #1285 .

To address your question though, this line introduced a funny bug. We defined `mdc-button__icon` in `button.scss`, which is sourced as part of `material-component-web.css` in `demo.scss`. However, we added "Material+Icons" stylesheet below that to introduce `material-icon`. The latter introduced a `font-size: 24px` that override what we provided in `demo.scss`(`button.scss`) and that's the reason why I moved it to the bottom because we definitely don't want a general library to override our own styles.

It shouldn't be a concern any longer after I rebased it.

demos/button.html
yeelan0319: I don't hold a strong opinion here, but this is a reasonable ask since that looks cleaner. Will move it.

packages/mdc-button/_mixins.scss
yeelan0319: Let me see if I could address this. The problem here is the button text is 14px (and 13px for dense button). However, the targeted icon side is 16px. I was not sure about the intention and may check with design tomorrow in the sign off.

yeelan0319:
Thanks for checking it! I will wait a bit for CI to finish...

PR #1292, opened 2017-09-12
fix(snackbar): Drop mdc-button from snackbar's dependency

yeelan0319:
BREAKING CHANGE: Removed the dependency of mdc-button from DOM structure of snackbar.

Dev server: 1292-dot-mdc-web-dev.xxxx.com

## Background
The rationale behind dropping button from snackbar is `action-button` is not technically a `MDC button` from the design perspective.
- It doesn’t have ripple
- It doesn’t have state
- It doesn’t have height and width restriction
- It doesn’t need to have icon etc..(after #1281 merged)

So to prevent future regression, we remove the dependency of button from snackbar by directly styling `action button`. The changes increase CSS file by 10 lines, but the actual size remained same 10KB.

### DIFF
--- a/packages/mdc-snackbar/mdc-snackbar.scss
+++ b/packages/mdc-snackbar/mdc-snackbar.scss
@@ -129,19 +129,25 @@
   /* stylelint-enable plugin/selector-bem-pattern */
 
   &__action-button {
+    @include mdc-typography(button);
     @include mdc-theme-prop(color, secondary);
 
     @include mdc-theme-dark(".mdc-snackbar") {
       @include mdc-theme-prop(color, primary);
     }
 
-    @include mdc-rtl-reflexive-box(margin, right, -16px, ".mdc-snackbar");
-
-    min-width: auto;
-    height: inherit;
     transition: mdc-animation-exit-permanent(opacity, .3s);
+    border: none;
+    outline: none;
+    background-color: transparent;
     opacity: 0;
+    user-select: none;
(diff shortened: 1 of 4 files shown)


packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -129,20 +129,26 @@
...
 
   &__action-button {
+    @include mdc-typography(button);
     @include mdc-theme-prop(color, secondary);
 
     @include mdc-theme-dark(".mdc-snackbar") {
       @include mdc-theme-prop(color, primary);
     }
 
-    @include mdc-rtl-reflexive-box(margin, right, -16px, ".mdc-snackbar");
```
kfranqueiro: Why was this line removed? This effectively shifts the action button to the left 16px.

yeelan0319: This line used to be the mechanism to reverse `padding` introduce by button, since now there is no longer `padding: 0 16px`, we also could safely remove the `margin-direction: -16px` here. 

Could you take a look at the deployed dev server and see if this style looks correct to you?

amsheehan:
screen shot 2017-09-14 at 11 08 56 am
screen shot 2017-09-14 at 11 08 59 am

yeelan0319:
Good catch, seems like Chrome add some padding by default to button. My overlook indeed! Thanks @amsheehan

PR #1294, opened 2017-09-13
fix(dialog): Fix $compound param in mdc-theme-dark

### LINKED ISSUE #1031: `mdc-button--theme-dark` with `:disabled`, doesn't properly set `background-color` or `color`
Check out the latest code and run `npm run dev`, navigate to http://localhost:8080/button.html (The latest version of the demo page is not available on https://material-components-web.appspot.com/button.html yet)

The `mdc-theme--dark` section shows a disabled button with grey text

ikgvjcpzimp

However, the `mdc-button--theme-dark` section shows a disabled button with black text.

b8fyn9xemo3

The second section, `mdc-button--theme-dark`, has the bug. The disabled button should have grey text.

The fix is to correctly use the `@include mdc-theme-dark(".mdc-button")` within the `:disabled` context of mdc-button.scss. It needs to be updated to

```
@include mdc-theme-dark(".mdc-button", true /* compound */)
```

[`mdc-theme-dark`](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#mdc-theme-darkroot-selector-compound) mixin requires `$compound` to be set to true if the 

> current selector is a compound selector with the base selector

...and `:disabled` is a compound selector with `.mdc-button`.

### LINKED ISSUE #1032: `mdc-dialog--theme-dark` with `mdc-dialog____body`, doesn't properly set `color`
Check out the latest code and run `npm run dev`, navigate to http://localhost:8080/dialog.html (The latest version of the demo page is not available on https://material-components-web.appspot.com/dialog.html yet)

The `mdc-theme--dark` section shows grey text in the dialog

yyyejwvngs7

However, the `mdc-dialog--theme-dark` section black text.

r6kaddbqgpc

The second section, `mdc-dialog--theme-dark`, has the bug. The dialog should have grey text.

The fix is to correct the `@include mdc-theme-dark(".mdc-dialog", true)` within the `.mdc-dialog__body` context of mdc-dialog.scss. It needs to be updated to

```
@include mdc-theme-dark(".mdc-dialog")
```

[`mdc-theme-dark`](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#mdc-theme-darkroot-selector-compound) mixin requires `$compound` to be set to true if the 

> current selector is a compound selector with the base selector

...and `.mdc-dialog__body` is NOT a compound selector with `.mdc-dialog`.

teddyward:
resolves #1031
resolves #1032

$compound param was set to true in dialog, when it should not have been. Conversely, it was not set in mdc-button, when it should have been. Fixing the first caused the second to show up as broken in the dialog demo, so I fixed both in one commit instead.

### DIFF
--- a/packages/mdc-button/mdc-button.scss
+++ b/packages/mdc-button/mdc-button.scss
@@ -29,7 +29,7 @@
   @include mdc-button-ink-color(text-primary-on-light);
   @include mdc-button-ripple((base-color: black, opacity: .16));
 
-  @include mdc-theme-dark(".mdc-button") {
+  @include mdc-theme-dark(".mdc-button", true /* compound */) {
     @include mdc-button-ink-color(text-primary-on-dark);
     @include mdc-button-ripple((base-color: white, opacity: .16));
   }
@@ -40,7 +40,7 @@
   @include mdc-button--filled_;
   @include mdc-button-filled-accessible(black);
 
-  @include mdc-theme-dark(".mdc-button") {
+  @include mdc-theme-dark(".mdc-button", true /* compound */) {
     @include mdc-button-filled-accessible(white);
   }
 }
@@ -55,7 +55,7 @@
   @include mdc-button-stroke-style(solid);
   @include mdc-button-stroke-color(text-primary-on-light);
 
-  @include mdc-theme-dark(".mdc-button") {
+  @include mdc-theme-dark(".mdc-button", true /* compound */) {
     @include mdc-button-stroke-color(text-primary-on-dark);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/mdc-button/mdc-button.scss
```diff
@@ -29,7 +29,7 @@
   @include mdc-button-ink-color(text-primary-on-light);
   @include mdc-button-ripple((base-color: black, opacity: .16));
 
-  @include mdc-theme-dark(".mdc-button") {
+  @include mdc-theme-dark(".mdc-button", true /* compound */) {
```
lynnjepsen: I dont think this one is compound....this will end up creating a selection like `.mdc-button--theme-dark.mdc-button`. I think in this case you just need to write `@include mdc-theme-dark` with no arguments.

PR #1301, opened 2017-09-14
docs(toolbar): Fix Toolbar with Menu demo's icon padding

yeelan0319:
Fix a small regression introduced by #1068 

Before
screen shot 2017-09-14 at 14 48 11

After
screen shot 2017-09-14 at 14 48 49

### DIFF
--- a/demos/toolbar/menu-toolbar.html
+++ b/demos/toolbar/menu-toolbar.html
@@ -52,17 +52,17 @@
         <section class="mdc-toolbar__section mdc-toolbar__section--align-end" role="toolbar">
           <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Download" alt="Download">file_download</a>
           <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Print this page" alt="Print this page">print</a>
-          <div class="mdc-menu-anchor">
           <a href="#" class="material-icons mdc-toolbar__icon toggle" aria-label="Bookmark this page" alt="Bookmark this page">more_vert</a>
-          <div class="mdc-simple-menu" tabindex="-1" id="demo-menu">
-            <ul class="mdc-simple-menu__items mdc-list" role="menu" aria-hidden="true" style="transform: scale(1, 1);">
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.069s;">Back</li>
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.124s;">Forward</li>
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.179s;">Reload</li>
-              <li class="mdc-list-divider" role="separator"></li>
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.236s;">Save As...</li>
-            </ul>
-          </div>
+          <div class="mdc-menu-anchor">
+            <div class="mdc-simple-menu" tabindex="-1" id="demo-menu">
+              <ul class="mdc-simple-menu__items mdc-list" role="menu" aria-hidden="true" style="transform: scale(1, 1);">
+                <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.069s;">Back</li>
+                <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.124s;">Forward</li>
+                <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.179s;">Reload</li>
+                <li class="mdc-list-divider" role="separator"></li>
+                <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.236s;">Save As...</li>
+              </ul>
+            </div>
           </div>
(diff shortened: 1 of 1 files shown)


demos/toolbar/menu-toolbar.html
```diff
@@ -52,17 +52,17 @@
...
-          <div class="mdc-simple-menu" tabindex="-1" id="demo-menu">
-            <ul class="mdc-simple-menu__items mdc-list" role="menu" aria-hidden="true" style="transform: scale(1, 1);">
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.069s;">Back</li>
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.124s;">Forward</li>
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.179s;">Reload</li>
-              <li class="mdc-list-divider" role="separator"></li>
-              <li class="mdc-list-item" role="menuitem" tabindex="0" style="transition-delay: 0.236s;">Save As...</li>
-            </ul>
-          </div>
+          <div class="mdc-menu-anchor">
```
kfranqueiro: I'm not sure this fix is correct... the purpose of `mdc-menu-anchor` is to exist inside the `mdc-simple-menu` root node to be detected by it to determine how to position the menu, right?

demos/toolbar/menu-toolbar.html
```diff
@@ -52,17 +52,17 @@
         <section class="mdc-toolbar__section mdc-toolbar__section--align-end" role="toolbar">
           <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Download" alt="Download">file_download</a>
           <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Print this page" alt="Print this page">print</a>
-          <div class="mdc-menu-anchor">
           <a href="#" class="material-icons mdc-toolbar__icon toggle" aria-label="Bookmark this page" alt="Bookmark this page">more_vert</a>
```
kfranqueiro: Incidentally, there's some copypasted label/alt attribute values from that original commit here too.

demos/toolbar/menu-toolbar.html
yeelan0319: I figured out this might be a valid way from [simple-menu's README](https://github.com/material-components/material-components-web/tree/master/packages/mdc-menu#automatic-positioning)

```
<div class="toolbar mdc-menu-anchor">
  <div class="mdc-simple-menu">
  ...
  </div>
</div>
```

Also from the running demo, I feel this looks right compare to our catalog page?
Now on catalog
screen shot 2017-09-14 at 17 30 03

Verse
This change
screen shot 2017-09-14 at 17 29 28

screen shot 2017-09-14 at 17 29 37

demos/toolbar/menu-toolbar.html
yeelan0319: Sorry, I am not sure if I followed this comment...

demos/toolbar/menu-toolbar.html
kfranqueiro: Looking at my comment earlier I seem to have misworded it with regard to the parent/child relationship, sorry.

I was thinking the code as it was structured was intentionally putting the overflow icon within the anchor element (there's another example in the menu readme that shows a button as well as the menu within the anchor). If it doesn't make a difference either way then I suppose either way is fine here. I'll have to play around with menu sometime to see what the difference is between the two approaches.

demos/toolbar/menu-toolbar.html
kfranqueiro: the label and alt on the overflow button say "Bookmark this page" when they should probably say "More".

demos/toolbar/menu-toolbar.html
yeelan0319: TL;DR this is what caused the regression.

Let me try explain the rationale behind the fix.
We expect the last `toolbar__icon` is have a larger padding than the others (16px than 8px). 

The way we implemented that is through `:last-of-type`. However, based on CSS3 spec, the pseudo selector `:last-of-type` is:
> The :last-of-type selector allows you to target the last occurence of an element within its container.
And in this case, the last element we want to apply 16px padding is nested within `simple-menu`'s anchor, so the style applied to both the actual last element as well as the secondary last element.

I think about if there is actual ways to select only the real last element, but it seems pretty hard if the assumption is to allow any level of nesting. `:last-child` also not helping here since technically speaking the dotdotdot is not even the last child in this anchor container. 

I agree this is a example shown in `simple-menu` but I feel it might not applicable due to how toolbar is structured. Any suggestions?

demos/toolbar/menu-toolbar.html
yeelan0319: Nice catch. It seems one of earlier commit updated the icon but not aria tag.

I am thinking changing this back to `bookmark` instead of `more_vert` because `more_vert` seems a pattern that user expect menu will pop out?

kfranqueiro:
Yiran and I discussed this more yesterday, and after trying a few other things, we decided to go with Yiran's original approach to address this specific issue in the demo. Unfortunately there are some limits/quirks to what `:last-of-type` can do.

Issue #1140, opened 2017-08-17
Add 20 components to theme demo page


lynnjepsen:
- [x] button
- [x] checkbox
- [x] drawer
- [x] FAB
- [x] icon toggle
- [x] linear progress
- [x] radio
- [x] ripple
- [x] select
- [x] slider
- [x] switch
- [x] tabs
- [x] text field
- [x] toolbar

PR #1312, opened 2017-09-15
docs(button): Remove --primary and --accent modifier from demos and docs

### LINKED ISSUE #1287: Remove/fix remaining instances of mdc-button--primary and mdc-button--secondary
We recently removed these classes since we now expose Sass mixins that can be used for the same purpose. However, there are still a number of references to these classes throughout demos and docs.

There's also at least one visible related regression, in snackbar.html with the action button text:

snackbar html

Some other demos (e.g. tabs.html) use `mdc-button--primary` but it's probably okay for us to drop the class in cases like that.

**Edit:**

Yiran pointed out yesterday that the snackbar regression specifically is actually due partly to slightly earlier work, which caused snackbar's own action button selector to no longer override styles in mdc-button. So that's a bit of a separate issue which I just happened to find at the same time.

So there's now two tasks here:

* [x] Resolve snackbar action button styling
* [ ] Remove or update other general uses of `mdc-button--primary` and `mdc-button--secondary` in demo pages

yeelan0319:
Closes #1287

## Related to #991 
Note: I intentionally left those in `demos/theme/index.html` because #1140


docs/theming.md
```diff
@@ -110,16 +110,15 @@ Here's the markup:
...
         </section>
       </div>
     </div>
   </body>
 </html>
 ```
 
-You'll see that we have a number of pretty empty looking cards, with black text on a white background. The only hint of
-color comes from the buttons, which we've made use the primary color by adding the `mdc-button--primary` class.
+You'll see that we have a number of pretty empty looking cards, with black text on a white background. For buttons, they adopt the baseline color by default.
```
lynnjepsen: Maybe we should change this to use the MDC Fab, which defaults to primary color?

packages/mdc-theme/README.md
```diff
@@ -111,7 +111,7 @@ CSS Custom property | Description
 
 ### CSS Classes
```
lynnjepsen: I would just replace this entire CSS Classes section with a section about our Sass color mixins

packages/mdc-theme/README.md
```diff
@@ -111,7 +111,7 @@ CSS Custom property | Description
...
 
-Some components can change their appearance when a theme-based modifier CSS class is applied. For example, `mdc-button--primary` will make the MDC Button the primary color. For more documentation on these modifier classes, consult the documentation for each component.
+Some components can change their appearance when a theme-based modifier CSS class is applied. For example, `mdc-icon-toggle--primary` will make the MDC Icon Toggle the primary color. For more documentation on these modifier classes, consult the documentation for each component.
```
kfranqueiro: This edit works for now, but I suspect we'll have to revise this more later.

docs/theming.md
yeelan0319: Sounds good, that seems better meet the purpose here actually.

packages/mdc-theme/README.md
yeelan0319: I am a bit worried doing so since we have only parameterized 2 elements: `button` and `fab`. The majority of the library is still not using the sass mixin way for customize theming. Could we consider mention this possibility and redirect our users to the documentation for each component instead?

Here is what I am thinking
> ### CSS Classes
>
>MDC-Web provides SASS mixins to some components to help users do customization, such as `@mixin mdc-button-filled-accessible($container-fill-color)`. For more documentation on these modifier classes, consult the documentation for each component.
>
>If you want to modify an element, which is not a Material Design component, you can apply the following modifier CSS classes.
> ....

How do you feel about this approach?

kfranqueiro: (Continuing this review for Lynn who's feeling under the weather. Hope you feel better soon, Lynn!)

I think this sounds reasonable for the time being. I agree RE replacing the current first paragraph since we'll be tending away from offering `--primary`/etc. classes, and I also agree it's a bit early to completely remove/retool this section since we've only tackled two components thus far.

Suggestion on wording for the new paragraph: "MDC-Web provides SASS mixins, such as `mdc-button-filled-accessible`, to facilitate customization of some components. For more information, consult the documentation for each component."

docs/theming.md
```diff
@@ -69,6 +72,12 @@ Here's the markup:
       .element-card > .mdc-card__media {
         height: 9em;
       }
+      #demo-absolute-fab {
```
kfranqueiro: Nit: Add an empty line before this selector

docs/theming.md
```diff
@@ -382,7 +402,7 @@ In addition, we also define custom properties for known dark and light backgroun
 
 
 Ideally, we should set all of the text colors on primary, since we never know which one an MDC-Web component might use.
-Since we're just using buttons, though, let's keep it simple for now:
+So let's keep it simple for now:
```
kfranqueiro: Suggestion: "Since our cards only contain text and no components, let's keep it simple for now:"

docs/theming.md
```diff
@@ -419,10 +439,10 @@ backgrounds), but the default is always light.
 > Note: When using a dark theme, you probably want to choose a dark color as the background for your page, and adjust
 the MDC-Web `background` color to match.
 
-In order to apply a dark theme to a single element, you can use its `--dark` class. For example, for a button:
+In order to apply a dark theme to a single element, you can use its `--theme-dark` class. For example, for a button:
 
 ```html
-<button class="mdc-button mdc-button--raised mdc-button--dark">
+<button class="mdc-button mdc-button--raised mdc-button--theme-dark">
```
kfranqueiro: Nice catch!

packages/mdc-theme/README.md
yeelan0319: SG! Thanks Ken for helping out the review

docs/theming.md
yeelan0319: DONE

docs/theming.md
yeelan0319: Thanks for catching it!

yeelan0319:
Thanks @kfranqueiro again and @lynnjepsen feel better.

Issue #983, opened 2017-07-20
Decide the future of our framework examples


lynnjepsen:
Our [framework examples](https://github.com/material-components/material-components-web/tree/master/framework-examples) currently serve as example code for how to wrap MDC Web in various frameworks

- Angular 2
- Aurelia
- Preact
- React
- Vue

This has some benefits
- We keep Real Code to test our Foundation/Adapter architecture against
- Lets us document some examples of how Foundation/Adapter works with frameworks

But it also has one key problem: We often have to triage issues in the community that are specific to one of these frameworks. We do not have "experts" for each of these frameworks who can help resolve these issues, so we often have unsatisfying replies to these issues.

@amsheehan suggested that we start a list on our README with links to _other_ repositories which wrap MDC Web in a framework. For example. @prateekbh could submit a PR to the MDC Web README, which adds a link to his [preact material components](https://github.com/prateekbh/preact-material-components). Then MDC Web could direct all preact specific issues to that repository.

And, of course, @prateekbh can still send MDC Web PRs whenever our Foundation/Adapter code needs tweaking to better support preact.

I really like this suggestion, but wanted to open it up as a discussion, in case there was other feedback.

prateekbh:
Sounds good! 
Just one thing: maybe before we add the links to README, we can look into the project with the most minimum requirement that the package does not break the _al-la-catré_ delivery model. I have seen projects which just bundle everything in one js file and hurt more than gains.
Thoughts?

trimox:
Great idea. I'd suggest grouping by type of MCW implementation method (e.g. using Foundation API or wrappers of the vanilla components).

Glad to see the discussion and the opportunity to demonstrate MCW using Angular..
Sign me up, 
[Angular MDC](https://github.com/trimox/angular-mdc-web)

amsheehan:
@trimox There's an example in our demos that will prevent the FOUC when toggling the start aligned snackbar 😄 

Angular MDC looks great!

Garbee:
I would urge when doing a "list of implementations" to have some quality control mechanisms in place. Something to the effect of:

* Project submissions must have existed for longer than 6 weeks.
* Should show continued maintenance over time
* Provide usage documentation to users

[...]

PR #1314, opened 2017-09-15
docs: Update docs around framework-examples directory

lynnjepsen:
Related to #983

I've started updating our docs with requirements for each library which wants to be listed on our main README. Follow up PRs will finish removing all of the framework-examples directory


docs/integrating-into-frameworks.md
```diff
@@ -14,11 +14,10 @@ frameworks.
 
 ## Examples
 
-Our [framework-examples](../framework-examples) directory contains sample applications built in various frameworks,
-such as [react](../framework-examples/react) and [angular2](../framework-examples/angular2) that show how to wrap our
-components. If there's an example framework you'd like to see, or if you're a framework author who
-wants to provide an example, please consider letting us know and we'll work with you to have it
-contributed!
+We maintain a list of component libraries, which wrap MDC Web for another framework, on our main [README](../README.md). Each library must:
```
amsheehan: "for other frameworks"

Too me that makes it clear that we're not looking for one particular framework that may or may not be developed by Google.

docs/integrating-into-frameworks.md
```diff
@@ -14,11 +14,10 @@ frameworks.
...
+- Serve components in an a-la-catré delivery model
```
yeelan0319: Love the French correctness 🔮

kfranqueiro: typos: à-la-carte

docs/integrating-into-frameworks.md
```diff
@@ -14,11 +14,10 @@ frameworks.
...
+- Have existed for longer than 6 weeks and show continued maintenance over time
+- Provide usage documentation to users per component
```
kfranqueiro: Consider dropping "to users"? I find it a little easier to read without

docs/integrating-into-frameworks.md
```diff
@@ -14,11 +14,10 @@ frameworks.
 
 ## Examples
```
yeelan0319: Question: do we want to maintain the list ourselves, or we are open to pull request since you listed the criteria here?

docs/integrating-into-frameworks.md
yeelan0319: 😿

lynnjepsen:
> Do we want to mention that we need component parity with MDC Web or is that too restrictive?

I don't think we could actually maintain this standard. We need some buffer time for us to add a new component...and then dependent libraries will follow with wrapped components. So yes, too restrictive.

> Question: do we want to maintain the list ourselves, or we are open to pull request since you listed the criteria here?

Open to pull requests! Just have to meet the criteria listed here.

docs/integrating-into-frameworks.md
kfranqueiro: To Yiran's point, would it make sense to re-word or augment "More coming soon!" to make it clear we're open to pull requests if someone thinks they have a good example repository?  "More coming soon!" might make it sound like we're planning to make them ourselves.

PR #1315, opened 2017-09-15
chore(demos): Fix CSS specificity of Material Icons in theme demo

acdvorak:
Previously, the Material Icons CSS had a higher specificity because it was imported after the theme page CSS. As a result, icons were sometimes displayed incorrectly.


acdvorak:
Tests pass locally, and this PR doesn't touch any JS; something must be funky with Travis.

PR #1316, opened 2017-09-15
chore(demos): Fix button colors on new theme demo page

acdvorak:
The `mdc-button--primary` and `mdc-button--secondary` CSS classes no longer exist. Use Sass mixins instead.


demos/theme/index.scss
```diff
@@ -101,6 +101,36 @@ $demo-section-margin: 48px;
 // Button demo
 //
 
+.mdc-button.secondary-text-button {
+  @include mdc-button-ink-color($mdc-theme-secondary);
+  @include mdc-button-ripple((base-color: $mdc-theme-secondary, opacity: .16));
+
+  @include mdc-theme-dark(".mdc-button") {
```
kfranqueiro: Are we actually showing dark-themed buttons on this page right now?

acdvorak: Not yet, but I will add a toggle for dark mode after I finish adding the remaining components to the page.

I just copied this Sass verbatim from `demos/button.scss`, which already contained the `mdc-theme-dark` bit; eventually I'd like to de-duplicate demo styles, but for now it seems simpler to just copy/paste it.

PR #1320, opened 2017-09-16
chore(demos): Make all drawer menu icons clickable

acdvorak:


demos/theme/index.html
```diff
@@ -295,7 +295,7 @@ <h2 class="mdc-typography--headline demo-component-section__heading">
           <a href="#drawer" class="demo-component-section__permalink" title="Permalink to the theme demo for the drawer component">#</a>
         </h2>
 
-        <p class="mdc-typography--body1">Click the menu icon (<i class="material-icons demo-icon--inline">menu</i>) in the toolbar at the top of the page to open the drawer.</p>
+        <p class="mdc-typography--body1">Tap the menu icon (<i class="material-icons demo-drawer-toggle">menu</i>) in the toolbar at the top of the page to open the drawer.</p>
```
kfranqueiro: I guess there's no longer a need to say "in the toolbar at the top of the page"?

acdvorak: I considered removing that bit of text, but I think it's still useful to direct users to the canonical location of the menu icon, while also giving them an inline copy so that A) they know what to look for, B) it's clear that you can open/close the drawer programmatically, and C) they don't have to move the mouse as far if they're as lazy as I am 😛

PR #1327, opened 2017-09-18
fix(toolbar): Fix toolbar padding on desktop and mobile

yeelan0319:
According to guideline the padding seems not right:
https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-keylines-spacing

After the changes, it became:
Desktop:
screen shot 2017-09-18 at 18 10 37

Mobile:
screen shot 2017-09-18 at 18 19 29


packages/mdc-toolbar/_variables.scss
```diff
@@ -17,8 +17,12 @@
 $mdc-toolbar-row-height: 64px;
 $mdc-toolbar-mobile-row-height: 56px;
 $mdc-toolbar-mobile-landscape-row-height: 48px;
-$mdc-toolbar-padding: 20px 28px;
-$mdc-toobar-element-vertical-padding: 16px;
+
+$mdc-toolbar-element-vertical-padding: 16px;
+$mdc-toolbar-element-horizontal-padding-desktop: 24px;
+$mdc-toolbar-element-horizontal-padding-mobile: 16px;
+$mdc-toolbar-title-margin-to-icon-menu-desktop: 8px;
```
kfranqueiro: Nit: would these names read better as `title-margin-to-menu-icon` (rather than `icon-menu`)?

kfranqueiro:
Also, this should be fix(toolbar), not fix(button), I think, when we merge.

packages/mdc-toolbar/_variables.scss
yeelan0319: It seems this won't introduce confusion with our modifier class name.

yeelan0319:
Thanks for catching the commit message error!

PR #1333, opened 2017-09-19
feat(auto-init): return initialized components

### LINKED ISSUE #1329: (Feature Request) Make autoInit return the initialized components
## Feature Request

Return the components initialized by a call to `autoInit`

**Motivation**

Currently, when integrating mdc with a framework or with custom elements is necessary to individually create the components, store its references and destroy when appropriate. With `autoInit` is possible to accomplish the first step with one call (create the components) but will not provide the means to destroy the component instances

**Use case**

By `autoInit` returning the initialized components we could write components like the below which will initialize (and) finalize all mdc components with auto-init attribute

```

class MDAutoComponent extends HTMLElement {
  connectedCallback() {
    this.mdcComponents = mdc.autoInit(this)
  }
  disconnectedCallback() {
    this.mdcComponents.forEach((comp) => comp.destroy())
  }
}
```

blikblum:
Make `autoInit` return initialized components.

Fixes #1329


blikblum:
I signed it!

Garbee:
What exactly is the reason for wanting this change? Prevent memory leaks? Keep track of changes somehow?

As far as I can tell, even in the issue example, there is no compelling use-case for this to be needed. When the nodes are removed, as long as they have no references left in the JS the related things will be cleaned up.

We fell for a little trap of storing references internally in MDL's auto init system and it was only problematic in the end. I'd rather MCW avoids allowing for that pitfall without tangible gain on having it included.

blikblum:
> What exactly is the reason for wanting this change? Prevent memory leaks? Keep track of changes somehow?

Keep a reference of the initialized MD components in an **wrapper component** so we can destroy them as advised in [Framework Integration guide](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md). It even cites the custom element example:

> Note that this approach will also work for custom elements. Use connectedCallback for initialization and disconnectedCallback for destruction.

As a concrete example see this [pen](https://codepen.io/blikblum/pen/JrKYWO). It wraps the [Getting Started](https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md) demo in a custom element. The MD components are initialized with `autoInit` (using the element as root). Per the integration guide the MD components should be destroyed when the **component wrapper** is destroyed but currently is not possible because there's not a way to get their references. 

The alternative would be initialize and store a reference of each MD component to be destroyed in `disconnectedCallback`. This is counter productive and error prone. Say that i want to add or remove a new text field or button in the form: i would have to update the template, initialize the component, store a reference and use it to call destroy. 

> As far as I can tell, even in the issue example, there is no compelling use-case for this to be needed. When the nodes are removed, as long as they have no references left in the JS the related things will be cleaned up.

So is not necessary to call `MDComponent.destroy` when element is removed? In the [pen](https://codepen.io/blikblum/pen/JrKYWO) the form (custom element) can be removed and recreated. It will not cause memory leaks?

blikblum:
I confirm

amsheehan:
@Garbee Can you update the thread on what the pitfalls in MDL were specifically?

blikblum:
> While this directly isn't doing that, the change does open the gateway for that very easy mistake to make to be done by external parties.

It's user responsibility to handle such situations. This is the case for JavaScript in general, not MDC specific. Even without this feature is currently possible to have leaks with the MDC API as in the example below:

```
import {MDCTextField} from '@material/textfield';

window._mdcComps = [];

const el = document.querySelector('.mdc-text-field')

const textField = new MDCTextField(el);

window._mdcComps.push(textField)

//later
el.parentNode.removeElement(el);
```
> Destructing should be optional anyways

This contradicts the [documentation](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md):

"When the wrapper component is destroyed (e.g. it is unbound and detached from the DOM), call mdcComponent.destroy() to clean up the MDC-Web component."

> Also, if you need to destroy a component, you should be capable of calling the destruction on the node itself without needing to keep the references from auto initialization.

I can't see in a clean way to do that with autoInit

It would need to call `querySelectorAll('[data-mdc-auto-init]')` in the root node and walk through the nodes checking for each component name:
```
function cleanMDCAutoInit(el) { 
  const nodes = el.querySelectorAll('[data-mdc-auto-init]');
  for (let i = 0, node; (node = nodes[i]); i++) {
     if (node.MDCTextField) {
       node.MDCTextField.destroy()
      } else if (node.MDCRipple) {
      node.MDCRipple.destroy()
     } // etc
  }
}
```

kfranqueiro:
Sorry this PR lapsed for so long. The changeset looks reasonable to me and I agree with the thoughts expressed by @amsheehan and @blikblum. The CLA was signed and it looks like this can still be merged without conflict, so I'm going to go ahead with it.

PR #1342, opened 2017-09-21
feat(tabs): tab bar foundation - publicize layoutIndicator method

danielraggs:
This pull request removes the underscore from the method `layoutIndicator_()` on the `MDCTabBar` foundation so that the framework using it can call this method and force an update to the indicator's styles. Additional documentation has also been added to the README for MDC tabs.

**Example use case:**
The user visits a route containing a tab bar of tabs `[ A ][ B ][ C ]`, in which tab B is active. The state of the app changes (without the route changing), causing the tab bar to change to `[ D ][ E ][ F ]`. Simultaneously, tab E becomes active. However, since E has the same index as B, the `switchToTabAtIndex()` method thinks the tabs never changed, so the indicator is not updated. We need to communicate to the tab bar to update the indicator's width and position based on the newly active tab.

### DIFF
--- a/packages/mdc-tabs/README.md
+++ b/packages/mdc-tabs/README.md
@@ -557,6 +557,10 @@ classes, an adapter object must be provided.
 
 Sets layout for the Tab Bar component.
 
+#### MDCTabBarFoundation.layoutIndicator() => void
+
+Sets layout and styles for the Tab Bar indicator based on active tab width and position.
+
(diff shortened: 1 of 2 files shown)


danielraggs:
So the functionality I need can actually be achieved with the method `.layout()` on the foundation, so I no longer have a need for `.layoutIndicator_()` to be publicized. 👍

PR #1343, opened 2017-09-21
chore(icon-toggle): Make MDCIconToggleAdapter not a default export

lynnjepsen:
mdc-icon-toggle/adapter.js exports two things: MDCIconToggleAdapter and IconToggleEvent


lynnjepsen:
I actually did this change in order to better support [goog.module format](https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide). In order to simplify the transpiling from our current format to a goog.module format, I need to co-locate all the export declarations in one line in our source code. Either like `export default Foo;` or `export {Foo, Bar};`. This is transpiled to either `exports = Foo;` or `exports = {Foo, Bar}` respectively.

I can't mix and match default and non-default exports...which is what mdc-icon-toggle/adapter.js currently is doing.

PR #1347, opened 2017-09-22
chore: Transpile our JS to use goog.module format

### LINKED ISSUE #1337: Update closure tests to rewrite JS with goog.module
We need to support the goog.module format

lynnjepsen:
First we will have to rewrite some of the source code to be more transpile-friendly
- Move all ‘export’ expressions to the end of the file. Otherwise replacing that expression with a new expression drops the leading closure comment.
- Instead of `import {foo} from './util'`, use `import * as util from './util'`. This is only necessary if there is only ONE object within the curly brackets. Unfortunately the AST parser does see the difference between `import {foo}` and `import foo`. However, `import {foo, bar}` is okay.

Both of these source code changes are nitpicky, but it makes the transpiling script easier to maintain.

The actual transpiling scripting does the following
- Added goog.module to the top
- ExportNamedDeclaration, aka export {foo, bar}, becomes exports = {foo, bar}
- ExportDefaultDeclaration, aka export default Foo, becomes exports = Foo
- ImportDeclaration
  - import Foo from ‘./foo’, becomes const Foo = goog.require(‘mdc.foo’)
  - import {foo, bar} from ‘./util’, becomes const {foo, bar} = goog.require(‘mdc.util)
  - import Foo from ‘@material/foo’, becomes const Foo = goog.require(‘mdc.foo’)

Resolves #1337

### DIFF
--- a/scripts/rewrite-decl-statements-for-closure-test.js
+++ b/scripts/rewrite-decl-statements-for-closure-test.js
@@ -95,14 +95,47 @@ function transform(srcFile, rootDir) {
   });
 
   traverse(ast, {
-    'ImportDeclaration|ExportNamedDeclaration'({node}) {
-      if (node.source) {
-        rewriteDeclarationSource(node, srcFile, rootDir);
-      }
+    ExportNamedDeclaration(path) {
+      const properties = [];
+      path.node.specifiers.forEach((specifier) => {
+        properties.push(t.objectProperty(specifier.exported, specifier.exported, false, true, []));
+      });
+      const right = t.objectExpression(properties);
+      const expression = t.assignmentExpression('=', t.identifier('exports'), right);
+      path.replaceWith(t.expressionStatement(expression));
     },
   });
 
-  const {code: outputCode} = recast.print(ast, {
+  traverse(ast, {
+    ExportDefaultDeclaration(path) {
+      const expression = t.assignmentExpression('=', t.identifier('exports'), path.node.declaration);
+      path.replaceWith(t.expressionStatement(expression));
+    },
+  });
+
+  traverse(ast, {
+    ImportDeclaration(path) {
+      const callee = t.memberExpression(t.identifier('goog'), t.identifier('require'), false);
+      const packageStr = rewriteDeclarationSource(path.node, srcFile, rootDir);
+      const callExpression = t.callExpression(callee, [t.stringLiteral(packageStr)]);
+
+      let variableDeclaratorId;
+      if (path.node.specifiers.length > 1) {
+        const properties = [];
+        path.node.specifiers.forEach((specifier) => {
+          properties.push(t.objectProperty(specifier.local, specifier.local, false, true, []));
+        });
+        variableDeclaratorId = t.objectPattern(properties);
+      } else {
+        variableDeclaratorId = path.node.specifiers[0].local;
+      }
+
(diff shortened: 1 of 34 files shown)


scripts/rewrite-decl-statements-for-closure-test.js
```diff
@@ -95,14 +95,47 @@ function transform(srcFile, rootDir) {
   });
 
   traverse(ast, {
-    'ImportDeclaration|ExportNamedDeclaration'({node}) {
-      if (node.source) {
-        rewriteDeclarationSource(node, srcFile, rootDir);
+    ExportNamedDeclaration(path) {
```
kfranqueiro: Given that we're adding a bunch of stuff here, should we add to the docs above in this file?

Also, should we add notes about the new requirements to our docs about writing closure-friendly code?

scripts/rewrite-decl-statements-for-closure-test.js
```diff
@@ -95,14 +95,47 @@ function transform(srcFile, rootDir) {
...
+  });
+
+  traverse(ast, {
+    ImportDeclaration(path) {
+      const callee = t.memberExpression(t.identifier('goog'), t.identifier('require'), false);
+      const packageStr = rewriteDeclarationSource(path.node, srcFile, rootDir);
+      const callExpression = t.callExpression(callee, [t.stringLiteral(packageStr)]);
+
+      let variableDeclaratorId;
+      if (path.node.specifiers.length > 1) {
```
kfranqueiro: I figured out a possible solution to the problem of requiring globbing in place of referencing a single export. You're right, it seems like the AST doesn't reflect the `{` and `}` in the specifier ever, but it does reflect the source locations...

```js
        if (path.node.specifiers.length > 1 ||
          (src.substr(path.node.specifiers[0].start - 1, 1) === '{' && src.substr(path.node.specifiers[0].end, 1) === '}')) {
```

scripts/rewrite-decl-statements-for-closure-test.js
lynnjepsen: Added some docs

scripts/rewrite-decl-statements-for-closure-test.js
lynnjepsen: Yay! Works!

scripts/rewrite-decl-statements-for-closure-test.js
kfranqueiro: I see the docs added in markdown, but should there be anything in a block comment here / above summarizing what we're doing?

lynnjepsen: woops...forgot to push that change

scripts/rewrite-decl-statements-for-closure-test.js
```diff
@@ -15,17 +15,24 @@
  */
 
 /**
- * @fileoverview Rewrites import statements such that:
+ * @fileoverview Rewrites JS to match a goog.module format. That means
+ *  * Add goog.module to the top of the source code
+ *  * Rewrite export {foo, bar} to exports = {foo, bar}
+ *  * Rewrite export default Foo, to exports = Foo
+ *  * Rewrite import Foo from ‘./foo’ to const Foo = goog.require(‘mdc.foo’)
```
kfranqueiro: Please kill smart quotes here and on next line (preferably with 🔥 )

Issue #1345, opened 2017-09-22
Button no longer inherits color from theme


gpoitch:
According to the readme:
> By default an MDC Button will inherit its color from the theme

```css
:root {
  --mdc-theme-primary: #3fa9f5;
}
```

Since 0.21.0, this is no longer happening and the buttons are all black.

Lc0rE:
Exactly! I came here to open an issue for this and I can confirm this behaviour.

Inspecting the CSS code, it seems that `mdc-button--primary`, `mdc-button--secondary`, `mdc-button--accent` are completely missing.
I don't know if I'm doing something wrong but I just approached MDC for the first time following the official Getting Started guide and I encounter this strange behaviour.

Has anyone some information about?

hebbet:
> Inspecting the CSS code, it seems that mdc-button--primary, mdc-button--secondary, mdc-button--accent are completely missing.

they aren't missing, they were removed, see [Changelog](https://github.com/material-components/material-components-web/blob/master/CHANGELOG.md#0210-2017-09-18):

> button: Remove support of mdc-button--primary and mdc-button--accent modifier classes. For custom and theme button implementation, use button mixins instead. See demos.scss for details.

Lc0rE:
Hey, @hebbet thanks for the useful comment!

As I said, I came here just after having read the documentation which is (sadly) not mentioning this.
As a general rule, I think that it should be a good approach to update docs when something showed up in the very first tutorial is removed or modified...

I also found terribly annoying that the example linked in the theming section **docs/theming/** which is pointing to the plunker https://plnkr.co/edit/qxt7qpPsXgkt9UbMnE36?p=preview, contains the old (I guess old since I just approached the repository) `mdl-` classes.

Anyway, we don't have now the possibility to use CSS Custom Properties? Just SASS?

Garbee:
The library is still alpha level. You shouldn't be using it unless you're prepared for changes and removals from release to release. The change log is the point of reference for these as well. If the docs were updated then there would be quite a large list for each component which is distracting and off putting.

There is another issue discussing why it was removed and possible remedies. On mobile so having a hard time finding it right now. Continue discussion in that issue please to keep things centralized.

Lc0rE:
Sure @Garbee, not my intention to criticize. 
I'm aware of the early stage of the project and I opened a quick pull-request to avoid confusion around the theming of buttons and save times to newcomers.
Next, I'll update the plunker and the theming guide as well.

[...]

PR #1350, opened 2017-09-25
docs: Update getting-started themin section

Lc0rE:
With reference to the issue #1345, is no longer possible to use CSS Variables to style buttons

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -234,11 +234,14 @@ following to the `<head>` tag of `index.html`:
 </style>
 ```
 
-If you're using any supported browser, besides IE 11, you'll see that the button background as well as the focused underline and label on text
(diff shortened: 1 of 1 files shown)


Lc0rE:
I signed it!

Garbee:
I don't think this kind of addition should be included here. The changelog describes removed things. While the project is still in pre-RC form developers should look there for breakages in their existing builds. This is a slippery precedent for people wanting any future removal/modifications to be clearly called out in the doc pages. Which, while in pre-release form can become quite hefty.

The changelog is what exists for this type of content currently, keep things focused there for the best quality. When the time comes to things being more stable, then we should consider this type of system.

amsheehan:
We want to fix #1345 rather than document the issue here. Closing the pull request.

PR #1351, opened 2017-09-25
docs: Recommend yarn as the package manager

appsforartists:
Yarn has many advantages over NPM.  It's fast (using a shared local cache), but more importantly, it's deterministic.  With Yarn, you'll never check out a project and find that `yarn install` no longer works because a transitive dependency had a breaking change.  Unfortunately, it does happen with `npm install`.

I haven't made any changes to your lerna config to introduce yarn there; that can be a separate change.  However, `yarn` is a best practice, and the docs should reflect that.


Garbee:
> yarn is a best practice

This is actually just highly encouraged but not really "best practice". While yarn has its advantages it also has its dis-advantages. Also, yarn will sometimes break things in an update silently and leave you scratching at why unless you figure out how to rollback yarn to get things working again.

NPM is just fine for docs. Since it is what is provided with node and has no extra setup involved. Yarn should remain an opt-in choice of developers that wish to use it.

kfranqueiro:
Agreed with @Garbee; npm is available OOTB and requires no extra installation (and also features package locks in v5), and MDC Web should remain unopinionated in this regard.

Issue #1102, opened 2017-08-09
mdc-button in Edge browser is colorless


Loggan08:
Renders properly in Chrome and FF
![untitled](https://user-images.githubusercontent.com/7104301/29122868-0ea0ea8c-7d14-11e7-8a81-22181c55ded7.png)

touficbatache:
Hey, thanks for filing an issue. But please consider using the provided template to submit bugs / issues as it makes it more clear what the issue is, which browser is it affecting, which OS you're using, etc... so it can be tested and fixed as soon as possible.

acdvorak:
Definitely looks like a bug. I was able to repro it on our [button demo page](https://material-components-web.appspot.com/button.html) in Edge 15.

It's not just dark mode, either: all `--raised` buttons seem to lack any background color at all in Edge.

As @touficbatache suggested, in the future, please try to use the provided issue template and attach a CodePen (or in this case, a link to the public demo page). It will help us get to the bottom of the issue faster.

Thanks!

RobJacobs:
@acdvorak I think this has to do with when the global css variables are set/imported.  On the demo page, I used the dev tools to look at the 'Buttons - RAISED WITH PRIMARY' element - the computed tab revealed: 

![image](https://user-images.githubusercontent.com/4559806/29274769-f00c0f2c-80d6-11e7-8f11-76551d4b9df0.png)

For the background-color property, notice the .mdc-button--primary.mdc-button--raised: var(---mdc-theme-primary, #212121) from the assests/demo-styles.css.js is being overruled by the .mdc-button: transparent rule from the assets/material-components-web.css.js.

Odly enough, simply unchecking the top level background-color checkbox will cause the rule to evaluate correctly.  Rechecking will then apply the rules:

![image](https://user-images.githubusercontent.com/4559806/29275196-4bb16a9c-80d8-11e7-8dc1-99366deb1464.png)

Same applies to the button color rule.

yeelan0319:
@acdvorak @Loggan08 ,

This seems to be a Edge 15 only bug because its wrong implementation of pseudo element style. I will try to lock down this further.

ctafield:
FWIW, you can easily repro these with the demo page... https://material-components-web.appspot.com/button.html - especially with the raised / accent styles.

RobJacobs:
@yeelan0319 @acdvorak I was hoping PR #1113 would fix the problem, but no luck.

kfranqueiro:
[...]

PR #1356, opened 2017-09-26
fix(button): Default to primary color

### LINKED ISSUE #1345: Button no longer inherits color from theme

kfranqueiro:
Fixes #1345.

I should point out a couple of caveats with this approach, which differs from the prior approach of providing --primary and --accent (now AKA secondary) as opt-in classes...

* By setting a hard default to primary, light-vs-dark considerations are dropped, and in general, contrast considerations will be left to the user when customizing theme if customizing via CSS variables.
  * I'm a little concerned especially about flat buttons using primary for text/ink color, in cases where the primary color is relatively light.
* Current versions of Edge with (buggy) CSS variable support will have problems displaying this. I'm pondering looking into #1102 to see if this can be addressed.

### DIFF
--- a/demos/common.scss
+++ b/demos/common.scss
@@ -24,6 +24,9 @@
 $mdc-theme-primary: $material-color-grey-900 !default;
 $mdc-theme-secondary: $material-color-light-green-a700 !default;
 
+// Import button mixins to apply overrides for dark theme
+// TODO(kfranqueiro): Pending further design discussion around how to manage dark theme
+@import "../packages/mdc-button/mixins";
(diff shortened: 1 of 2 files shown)


kfranqueiro:
Added another commit to address contrast issues with dark theme (perfect example of one of the caveats above)... I'll definitely want to talk over dark theme with designers some more as to the direction we should take that.

image

Issue #664, opened 2017-05-17
Edge: Text Field issue


djensen47:
When using/testing mdc-web on the demo site, there seems to be an issue with the text field.

### What MDC-Web Version are you using?

0.11.1 via this page:
https://material-components-web.appspot.com/

### What browser(s) is this bug affecting?

Microsoft Edge 41.16193.1001.0

### What OS are you using?

Windows 10 Build 16193

### What are the steps to reproduce the bug?

Open Edge to https://material-components-web.appspot.com/textfield.html and click inside of a text field and start typing.

### What is the expected behavior?

It behaves normally.

### What is the actual behavior?
The background turns black.

![text-field-edge](https://cloud.githubusercontent.com/assets/527865/26179863/d1f588b2-3b1a-11e7-817b-85486b634fce.gif)

kbrownhbs:
I have been seeing the same issue in Edge as well.  I'm using version  40.15063.0.0

KayvandenHeuvel:
We are using MDC version 0.16 and we still have this problem. Checkboxes & radio buttons got the same problem.

I think it has something to do with the ::after elements that are set.

Derekjohnson277:
Any workarounds or solutions?

dpraul:
Running MDC v0.17.0 on Edge 40.15063.0.0. Can reproduce this problem on text fields, checkboxes, radio buttons, and CSS-only tabs.

jordyboytjuh:
Same issues over here. MDC components in Edge are very buggy. Hope some bugfixes will resolve this.

Kerrick:
We're also seeing the same issue [here](https://secondstreet.github.io/ember-material-components-web/), for EdgeHTML 15.15.063 / Edge 40.15063.0.0, including radio buttons and checkboxes.

PR #1364, opened 2017-09-28
Fix Edge CSS variable glitches in specific cases

### LINKED ISSUE #1102: mdc-button in Edge browser is colorless

### LINKED ISSUE #1212: Checkbox filled with black in Edge browser
0.19.0

Edge  40.15063.0.0

### What OS are you using?
Windows 10

Check this demo in Edge:
https://material-components-web.appspot.com/checkbox.html

The checkbox should be outlined.

The checkbox is filled with black, especially notable in the non dark theme examples.

kfranqueiro:
This PR has multiple commits across packages, which I intend to merge *without* squashing for the sake of changelog and versioning.

This PR adds support to `mdc-theme-prop` for a new optional parameter which will emit `@supports` clauses around its rules to exclude Edge.  The rest of the Sass updates in this PR use that parameter in specific places that singularly cause display glitches in Edge in those components. I mention this in the READMEs so that hopefully users aren't surprised if they change CSS variables and it takes effect everywhere except a few places.

Fixes #1102
Fixes #1212

FYI, #664 will be addressed separately by #998, after which its problematic selector will no longer exist.

### DIFF
--- a/packages/mdc-button/README.md
+++ b/packages/mdc-button/README.md
@@ -136,3 +136,10 @@ Mixin | Description
 `mdc-button-ripple` | Sets the ripple to the given [ripple configuration](https://github.com/material-components/material-components-web/blob/master/packages/mdc-ripple/README.md)
 `mdc-button-corner-radius` | Sets the corner radius to the given number (defaults to 2px)
 `mdc-button-stroke-width` | Sets the stroke width to the given number (defaults to 2px)
+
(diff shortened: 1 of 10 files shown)


kfranqueiro:
Merged these commits manually as mentioned above. (In retrospect, maybe multiple PRs would actually be easier...)

PR #1365, opened 2017-09-29
feat(checkbox): Add color theme mixins and update default color to secondary

### LINKED ISSUE #1146: Update checkbox color to secondary color
- [ ] Create color mixins
- [ ] Call color mixin from stylesheet, with secondary color
- [ ] Update README to document color mixins
- [x] Delete any --primary --accent modifiers

acdvorak:
- Removed `[aria-disabled="true"]` selector, as it doesn't appear to be necessary
- Fixed preexisting state transition color animation bugs

Resolves #1146


packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,356 @@
...
+
+//
+// Public
+//
+
+// TODO(acdvorak): Figure out why a pre-checked checkbox janks when `indeterminate` is toggled for the first time.
+@mixin mdc-checkbox-color($color) {
+  $ripple-config: (base-color: $color);
+
+  // TODO(acdvorak): Refactor ripple mixin to do this @if check automatically, or use mdc-theme-prop()
```
kfranqueiro: Yeah, I want to do something about ripple's base-color/theme-style split...

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,334 +14,137 @@
...
-    will-change: background-color, border-color;
-
-    @include mdc-theme-dark(".mdc-checkbox") {
-      border-color: $mdc-checkbox-border-color-dark;
-    }
-
-    // The frame's ::before element is used as a focus indicator for the checkbox
-    &::before {
-      @include mdc-checkbox-cover-element;
-      @include mdc-theme-prop(background, primary);
```
kfranqueiro: FYI, #1364 is adding $edgeOptOut: true here, so you will want to do likewise wherever the equivalent is (not sure if that will require doing it in all invocations of a mixin now, which will generate more CSS...)

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,333 +14,135 @@
...
-      transition:
-        mdc-checkbox-transition-exit(opacity, 0ms, $mdc-checkbox-transition-duration),
-        mdc-checkbox-transition-exit(transform, 0ms, $mdc-checkbox-transition-duration);
-      opacity: 0;
-
-      &__path {
-        stroke-dashoffset: 0;
-      }
+    @include mdc-checkbox-background-unfilled-disabled {
+      @include mdc-checkbox-stroke-color($mdc-checkbox-disabled-color-dark);
```
lynnjepsen: just reference `border-colo`r directly

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,333 +14,135 @@
...
     }
 
-    .mdc-checkbox__mixedmark {
-      transform: scaleX(1) rotate(0deg);
-      opacity: 1;
+    @include mdc-checkbox-background-filled-disabled {
+      @include mdc-checkbox-container-color($mdc-checkbox-disabled-color-dark);
```
lynnjepsen: just reference `background-color` directly

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,333 +14,135 @@
...
-      // stylelint-enable selector-max-compound-selectors, selector-max-specificity
-    }
+  @include mdc-checkbox-background-filled-enabled {
+    @include mdc-checkbox-stroke-color(secondary);
+    @include mdc-checkbox-container-color(secondary);
   }
 }
 
+// Needed to disable hover effects on CSS-only (non-JS) checkboxes
 .mdc-checkbox--disabled {
```
lynnjepsen: how is `mdc-checkbox--disabled` different from `mdc-checkbox-background-filled-disabled` and `mdc-checkbox-background-unfilled-disabled`?

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,333 +14,135 @@
...
 
-  &__mixedmark {
-    width: 100%;
-    height: floor($mdc-checkbox-mark-stroke-size);
-    transform: scaleX(0) rotate(0deg);
-    transition: mdc-checkbox-transition-exit(opacity), mdc-checkbox-transition-exit(transform);
-    background-color: $mdc-checkbox-mark-color;
-    opacity: 0;
+  @include mdc-checkbox-background-unfilled-disabled {
+    @include mdc-checkbox-stroke-color($mdc-checkbox-disabled-color);
```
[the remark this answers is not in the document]
lynnjepsen: I should add: `mdc-checkbox-filled-stroke-color` is meant for customizing the stroke color of a filled *enabled* checkbox, so using that mixin to set styles for a disabled checkbox doesn't make sense.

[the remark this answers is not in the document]
kfranqueiro: I was about to say something like this :)

The nice thing about the button mixins is the user doesn't have to think about "enabled" at all, we apply it for them.

Also, we might be able to just use `:enabled` instead of `:not(:disabled)` here - the reason for `:not(:disabled)` on button was to also support its use for `a` elements across browsers.

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,400 @@
...
+// NOTE: Child selector MUST begin with `.mdc-checkbox__background`
+@mixin mdc-checkbox-background-filled-disabled {
+  .mdc-checkbox__native-control:disabled:checked ~,
+  .mdc-checkbox__native-control:disabled:indeterminate ~ {
+    @content;
+  }
+}
+
+@mixin mdc-checkbox-stroke-color($color) {
+  .mdc-checkbox__background {
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,400 @@
...
+    @include mdc-theme-prop(border-color, $color);
+  }
+}
+
+@mixin mdc-checkbox-container-color($color) {
+  .mdc-checkbox__background {
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,400 @@
...
+    @include mdc-theme-prop(background-color, $color);
+  }
+}
+
+@mixin mdc-checkbox-focus-color($color) {
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,400 @@
...
+    // for line animation to work properly.
+    @include mdc-theme-prop(stroke, $color, $important: true);
+  }
+
+  .mdc-checkbox__mixedmark {
+    @include mdc-theme-prop(background-color, $color);
+  }
+}
+
+@mixin mdc-checkbox-ripple-color($color, $opacity: .14) {
```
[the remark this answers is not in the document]
acdvorak: That would make the API inconsistent with the other color mixins (which only take a `$color` param).

If all the mixins take a single `$color` param (and possibly an optional second param), it makes it easy for users to try all of our mixins without needing to read the documentation.

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,400 @@
...
+  @include mdc-ripple-bg(map-merge((pseudo: "::before"), $ripple-config));
+  @include mdc-ripple-fg(map-merge((pseudo: "::after"), $ripple-config));
+
+  &::before,
+  &::after {
+    @include mdc-checkbox-ripple-size_;
+  }
+}
+
+@mixin mdc-checkbox-state-transition-color($color) {
```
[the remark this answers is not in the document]
acdvorak: Done.

For consistency, I still allow callers to pass a single argument for both colors, because I doubt most people will want separate container and stroke colors for a marked checkbox (though I need to confirm that with design). You can disable that behavior by explicitly passing `false` for `$stroke-color`.

demos/checkbox.scss
```diff
@@ -17,4 +17,19 @@
 @import "./common";
 @import "../packages/mdc-checkbox/mdc-checkbox";
 @import "../packages/mdc-form-field/mdc-form-field";
+// TODO(acdvorak): Rename file to `_color-palette`
+@import "../packages/mdc-theme/color_palette";
 @import "../packages/mdc-typography/mdc-typography";
+
+$demo-custom-color: $material-color-red-500;
+
+.demo-checkbox--custom {
```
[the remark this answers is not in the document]
acdvorak: Why?

packages/mdc-checkbox/mdc-checkbox.scss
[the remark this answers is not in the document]
acdvorak: Done. Too bad there's not an `:unchecked` or `:determinate` selector.

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,333 +14,135 @@
...
 
-.mdc-checkbox__native-control:focus {
-  ~ .mdc-checkbox__background::before {
-    transform: scale(2.75, 2.75);
-    transition:
-      mdc-checkbox-transition-enter(opacity, 0ms, 80ms),
-      mdc-checkbox-transition-enter(transform, 0ms, 80ms);
-    opacity: .26;
+  @include mdc-checkbox-background-filled-disabled {
+    @include mdc-checkbox-stroke-color(transparent);
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -14,333 +14,135 @@
...
+    @include mdc-checkbox-container-color($mdc-checkbox-disabled-color);
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/mdc-checkbox.scss
acdvorak: Done

packages/mdc-checkbox/mdc-checkbox.scss
acdvorak: Done

packages/mdc-checkbox/mdc-checkbox.scss
acdvorak: `mdc-checkbox--disabled` is a modifier class for CSS-only buttons that prevents them from displaying a ripple or pointer cursor on hover.

`mdc-checkbox-unfilled-background-selector-disabled_` and friends are convenience mixins that wrap a block of Sass in the big ugly CSS selectors needed to target the 4 different state combinations. This will also hopefully help us keep the relatively complex selectors in sync by putting them all in one place.

packages/mdc-checkbox/_variables.scss
```diff
@@ -29,5 +29,5 @@ $mdc-checkbox-transition-duration: 90ms;
 $mdc-checkbox-item-spacing: 4px;
 
 /* Manual calculation done on SVG */
```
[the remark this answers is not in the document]
acdvorak: Done

demos/checkbox.scss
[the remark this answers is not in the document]
lynnjepsen: In our demos the demo CSS comes after the MDC Web CSS....but on some clients order guarantee is difficult to implement. We want to document how to handle that case, which is to have a more specific CSS selector (two classes).

demos/checkbox.scss
```diff
@@ -17,4 +17,16 @@
...
+  @include mdc-checkbox-marked-stroke-color($demo-custom-color);
+  @include mdc-checkbox-marked-container-fill-color($demo-custom-color);
```
lynnjepsen: You dont have any "checkbox-unmarked" mixin calls here. We should demonstrate that customizable feature on our demo pages as well.

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,435 @@
...
+    // for line animation to work properly.
+    @include mdc-theme-prop(stroke, $color, $important: true);
+  }
+
+  .mdc-checkbox__mixedmark {
+    @include mdc-theme-prop(background-color, $color);
+  }
+}
+
+@mixin mdc-checkbox-ripple-color($color, $opacity: $mdc-checkbox-ripple-opacity) {
```
lynnjepsen: mdc-checkbox-ripple(ripple-config)

packages/mdc-checkbox/_mixins.scss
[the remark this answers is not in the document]
lynnjepsen: With button we just take in ripple-config. Ripple config is an existing concept in MDD Web, with its own complications. Our work on states should help improve the state of ripple-config, but lets not tackle that same problem in this PR.

Also when we tried to use a $color param directly with the `mdc-button-ripple` mixin we hit a lot of weird issues (around theme-ing versus non-theme colors) and I don't want to run into those *again*.

Also also, it should be just `mdc-checkbox-ripple`

packages/mdc-checkbox/_mixins.scss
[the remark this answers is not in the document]
lynnjepsen: I think I would make both container-color and stroke-color required...its a simpler API that way. Otherwise `mdc-checkbox-check-animation` might be a little too magical.

packages/mdc-checkbox/_mixins.scss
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/_mixins.scss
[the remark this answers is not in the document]
acdvorak: Done

demos/checkbox.scss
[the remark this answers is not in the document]
acdvorak: Done

demos/checkbox.scss
acdvorak: Done

packages/mdc-checkbox/_mixins.scss
acdvorak: Done

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,414 @@
...
+
+@mixin mdc-checkbox-ripple($ripple-config) {
+  $ripple-config: map-merge((opacity: $mdc-checkbox-ripple-opacity), $ripple-config);
+
+  @include mdc-ripple-bg(map-merge((pseudo: "::before"), $ripple-config));
+  @include mdc-ripple-fg(map-merge((pseudo: "::after"), $ripple-config));
+
+  &::before,
+  &::after {
+    // This needs to be included immediately after mdc-ripple-bg and mdc-ripple-fg
```
kfranqueiro: Is it not feasible to accomplish this by overriding `$radius` in the ripple config instead of overriding these styles? (So far we never customize radius internally but I'd be interested if we have our own use cases for it.)

packages/mdc-checkbox/_mixins.scss
```diff
@@ -0,0 +1,414 @@
...
+    // to override their default dimensions.
+    @include mdc-checkbox-ripple-size_;
+  }
+}
+
+@mixin mdc-checkbox-mark-animation(
```
kfranqueiro: This was probably discussed and I missed it, but are we concerned at all that you need to invoke multiple mixins with the same parameters?

Also, I'm confused by unmarked-container-color, since there's no individual mixin for that and I thought the idea was it's transparent?

demos/checkbox.scss
```diff
@@ -17,4 +17,24 @@
 @import "./common";
 @import "../packages/mdc-checkbox/mdc-checkbox";
 @import "../packages/mdc-form-field/mdc-form-field";
+@import "../packages/mdc-theme/color_palette";
 @import "../packages/mdc-typography/mdc-typography";
+
+$demo-custom-color: $material-color-red-500;
```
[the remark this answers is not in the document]
acdvorak: Good call. It might not be a problem right now, but I plan to eventually reuse component demo styles in multiple places (the component demo page and the theme demo page), at which point it will become a problem. Fixed!

packages/mdc-checkbox/README.md
```diff
@@ -111,7 +111,7 @@ easily position checkboxes and their labels.
 <label for="basic-disabled-checkbox" id="basic-disabled-checkbox-label">This is my disabled checkbox</label>
 ```
 
-Note that `mdc-checkbox--disabled` is necessary on the root element to prevent hover states from activating.
+Note that `mdc-checkbox--disabled` is necessary on the root element of CSS-only checkboxes to prevent hover states from activating. Checkboxes that use the JavaScript component do not need this class.
```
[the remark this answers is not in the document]
acdvorak: Good idea. Done.

packages/mdc-checkbox/README.md
```diff
@@ -255,9 +255,25 @@ not return an object.
...
 
+### Sass Mixins
+
+The following mixins apply only to _enabled_ checkboxes. It is not currently possible to customize the color of a _disabled_ checkbox.
+
+Mixin | Description
+--- | ---
+`mdc-checkbox-unmarked-stroke-color($color)` | Sets the stroke color of an _unmarked_ checkbox
+`mdc-checkbox-marked-stroke-color($color)` | Sets the stroke color of a _marked_ checkbox
+`mdc-checkbox-marked-container-fill-color($color)` | Sets the container a fill color of _marked_ checkbox
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-checkbox/README.md
```diff
@@ -255,9 +255,25 @@ not return an object.
...
+`mdc-checkbox-ink-color($color)` | Sets the ink color of the checked and indeterminate icons
```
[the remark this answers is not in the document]
acdvorak: Great question. I'll ask design on Tuesday 10/10, but I'm guessing it should be fully customizable.

The dark theme section of the updated demo page in this PR shows a good use case for custom ink color. Personally, I think a dark checkmark icon is easier to see than a light icon when it's on top of a dark background. Ditto for light icons on light backgrounds.

packages/mdc-checkbox/_mixins.scss
acdvorak: Good catch! I just blindly assumed that `$radius` referred to `border-radius`, but clearly I was wrong. Updated!

packages/mdc-checkbox/_mixins.scss
acdvorak: I'm definitely concerned about it. Animating from one color to another requires specifying both the start color and end color, so I'm not sure how we could work around this. Suggestions welcome!

packages/mdc-checkbox/README.md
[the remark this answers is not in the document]
kfranqueiro: That's pretty interesting...and for that matter, also makes me wonder if we're meeting contrast requirements for the white-on-green icons to begin with. Not sure offhand what the requirements are when icons are involved. Definitely worth picking designers' brains.

packages/mdc-checkbox/_mixins.scss
acdvorak: To reduce confusion, I removed the non-animation stroke/container mixins. There's now just one mixin that does all 4: `mdc-checkbox-stroke-and-container-fill-color()`. It also optionally generates the `@keyframes` needed by the JavaScript component.

demos/checkbox.scss
```diff
@@ -17,4 +17,39 @@
...
 @import "../packages/mdc-typography/mdc-typography";
+
+.mdc-checkbox.demo-checkbox--custom-stroke-and-fill {
+  $color: $material-color-red-500;
+
+  @include mdc-checkbox-focus-indicator-color($color);
+  @include mdc-checkbox-ripple((base-color: $color, opacity: $mdc-checkbox-ripple-opacity));
+  @include mdc-checkbox-stroke-and-container-fill-color(
+    $unmarked-stroke-color: $color,
+    $unmarked-container-fill-color: rgba($color, .25),
```
lynnjepsen: Should this .25 be a public variable in mdc-checkbox?

demos/checkbox.scss
```diff
@@ -17,4 +17,39 @@
...
+    $marked-stroke-color: $color,
+    $marked-container-fill-color: $color);
+
+  @include mdc-theme-dark(".mdc-checkbox", $compound: true) {
+    @include mdc-checkbox-ink-color(black);
+    @include mdc-checkbox-stroke-and-container-fill-color(
+      $unmarked-stroke-color: $color,
+      $unmarked-container-fill-color: rgba($color, .35),
```
lynnjepsen: How about .35?

packages/mdc-checkbox/README.md
```diff
@@ -255,8 +255,28 @@ not return an object.
...
-aware.
+MDC Checkboxes use the theme's secondary color by default for "marked" states (i.e., checked or indeterminate), and are completely dark theme aware.
+
+### Sass Mixins
+
+The following mixins apply only to _enabled_ checkboxes. It is not currently possible to customize the color of a _disabled_ checkbox.
+
+Mixin | Description
+--- | ---
+`mdc-checkbox-stroke-and-container-fill-color($unmarked-stroke-color, $unmarked-container-fill-color, $marked-stroke-color, $marked-container-fill-color, $animate)` | Generates CSS classes to set and animate the stroke color and/or container fill color of a checkbox
```
lynnjepsen: I hate using "and" in API names...

How about `mdc-checkbox-marked-indicator`?

packages/mdc-checkbox/README.md
```diff
@@ -255,8 +255,28 @@ not return an object.
...
+
+[ripple-readme]: https://github.com/material-components/material-components-web/blob/master/packages/mdc-ripple/README.md
+
+#### `mdc-checkbox-stroke-and-container-fill-color($unmarked-stroke-color, $unmarked-container-fill-color, $marked-stroke-color, $marked-container-fill-color, $animate)`
+
+Generates CSS classes to set the stroke color and/or container fill color of a checkbox in its marked and unmarked states.
+
+All parameters are optional, and if left unspecified will use their default values.
+
+If you plan to use CSS-only checkboxes, set `$animate` to `false` to prevent the mixin from generating `@keyframes` and CSS classes used by the JavaScript component.
```
lynnjepsen: I think it should be $keyframeAnimate...not just $animate. Doesn't the CSS-only checkbox still do some animation? Just not with keyframes?

packages/mdc-checkbox/README.md
kfranqueiro: But it takes both marked and unmarked styles? Coming up with a good descriptive name for this is hard...

packages/mdc-checkbox/README.md
kfranqueiro: Getting super-pedantic here, but would that be transition rather than animation in that case? In which case animate might be enough distinction, but I could go either way.

demos/checkbox.scss
acdvorak: IMHO probably not - this is just a demo page, and we don't have any official material guidelines about what opacity to use for fill color of an unmarked checkbox.

demos/checkbox.scss
acdvorak: Ditto

packages/mdc-checkbox/README.md
acdvorak: Agreed, I'm not wild about the name either...

The mixin covers both marked and unmarked states, so putting `marked` in the name would probably be misleading. And I'm not sure I would call this an `indicator`, exactly.

Since _stroke_ and _fill_ are both properties of the container, what about:
`mdc-checkbox-container-colors`?

Making the mixin name plural will hopefully make it obvious that it takes multiple color arguments, and help differentiate it from the other theme mixins that only take one color argument.

packages/mdc-checkbox/README.md
acdvorak: I originally named it `$generate-keyframes`, so I changed it back to that. Does that work?

You're correct - the CSS-only checkbox uses `transition` to tween from marked to unmarked states instead of `animation`. As you suggested, it's probably more accurate to name the parameter something with "keyframes", because that's what it really controls.

packages/mdc-checkbox/README.md
lynnjepsen: mdc-checkbox-container-colors sounds right to me.

packages/mdc-checkbox/README.md
lynnjepsen: generate-keyframes does sound better to me.

demos/checkbox.scss
```diff
@@ -15,6 +15,36 @@
...
+      $unmarked-fill-color: rgba($color, .35),
+      $marked-fill-color: $color);
+  }
+}
+
+.mdc-checkbox.demo-checkbox--custom-stroke-and-fill {
+  $color: $material-color-blue-500;
+
+  @include mdc-checkbox-container-colors(
+    $unmarked-stroke-color: $color,
```
kfranqueiro: I would suggest making this color different to better illustrate how this mixin's parameters work now. I was seriously confused for several minutes thinking we were applying the same stroke color to both states, but IIUC that's not what we're doing; we're making the marked stroke color match the marked fill color, which makes sense.

packages/mdc-checkbox/README.md
```diff
@@ -255,8 +255,28 @@ not return an object.
...
+`mdc-checkbox-container-colors($unmarked-stroke-color, $unmarked-fill-color, $marked-fill-color, $generate-keyframes)` | Generates CSS classes to set and animate the stroke color and/or container fill color of a checkbox
+`mdc-checkbox-ink-color($color)` | Sets the ink color of the checked and indeterminate icons
+`mdc-checkbox-focus-indicator-color($color)` | Sets the color of the focus indicator
+`mdc-checkbox-ripple($ripple-config)` | Sets the ripple to the given [ripple configuration][ripple-readme]
+
+[ripple-readme]: https://github.com/material-components/material-components-web/blob/master/packages/mdc-ripple/README.md
+
+#### `mdc-checkbox-container-colors($unmarked-stroke-color, $unmarked-fill-color, $marked-fill-color, $generate-keyframes)`
+
+Generates CSS classes to set the stroke color and/or container fill color of a checkbox in its marked and unmarked states. In the unmarked state, stroke and fill color may be customized independently; in the marked state, the stroke disappears, and only the fill may be customized.
```
kfranqueiro: This sounds a little misleading... the stroke doesn't technically disappear, but it intentionally matches the marked fill color, they can't be customized separately.

demos/checkbox.scss
acdvorak: Done

packages/mdc-checkbox/README.md
acdvorak: Done

PR #1366, opened 2017-09-29
chore: Update Sauce Labs username/key

kfranqueiro:


kfranqueiro:
This PR is no longer necessary, as Alex has finally gotten to the bottom of Travis' shenanigans and the existing user still works.

PR #1369, opened 2017-09-29
style(theme): Disable linter for Egde cases

lynnjepsen:


kfranqueiro:
Also PS Egde -> Edge in commit message

PR #1370, opened 2017-09-29
chore: Preserve comments when replacing imports for goog.module format

### LINKED ISSUE #1368: Retain license comment when transpiling to goog.module format
The license comment is currently being dropped in the transpile process.

lynnjepsen:
Also adding @license to the comment blocks, which is required by our internal infrastructure.

Resolves #1368

(Includes a commit from #1369, just to get the linter to pass)

### DIFF
--- a/scripts/rewrite-decl-statements-for-closure-test.js
+++ b/scripts/rewrite-decl-statements-for-closure-test.js
@@ -138,7 +138,13 @@ function transform(srcFile, rootDir) {
       }
 
       const variableDeclarator = t.variableDeclarator(variableDeclaratorId, callExpression);
-      path.replaceWith(t.variableDeclaration('const', [variableDeclarator]));
+      const variableDeclaration = t.variableDeclaration('const', [variableDeclarator]);
+      if (path.node.comments && path.node.comments.length > 0) {
+        const commentValue = path.node.comments[0].value;
+        variableDeclaration.comments = variableDeclaration.comments || [];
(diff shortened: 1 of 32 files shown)


scripts/rewrite-decl-statements-for-closure-test.js
```diff
@@ -138,7 +138,13 @@ function transform(srcFile, rootDir) {
       }
 
       const variableDeclarator = t.variableDeclarator(variableDeclaratorId, callExpression);
-      path.replaceWith(t.variableDeclaration('const', [variableDeclarator]));
+      const variableDeclaration = t.variableDeclaration('const', [variableDeclarator]);
```
kfranqueiro: Question, since I don't necessarily fully understand all the stuff we do in here:

Is this assuming that we will always have at least one import in a file, and relying on the first import to find the license comment? If so, will we encounter problems for files with no imports, e.g. */constants.js?

lynnjepsen: If there are no imports to replace, then the entire 'ImportDeclaration' callback doesn't trigger...which means no path/node is replaces with a 'goog.require' version.

The replacing of a node was what caused the comment information to drop, because I didn't copy the comments onto the new node. So actually constants.js was fine, and had the license information in the transpiled version. But any file with import statements was losing the license information.

kfranqueiro: Oh, I see. Thanks for explaining. Maybe we should add a comment to this code e.g. "Preserve comments when replacing imports"?

PR #1372, opened 2017-10-01
Update README.md

jamesmfriedman:
Added my React Wrapper project for MDC.


jamesmfriedman:
I signed it!

jamesmfriedman:
@amsheehan sorry I never saw your comment!

1) Yeah, I just saw that the other day, I'm currently fixing with the latest release
2) What do you think about just inheriting the version directly from you guys? So I'd publish your semver and then just a simple increment. i.e. 0.22.0-1, 0.22.0-2, 0.22.0-rc1

I think it would be too confusing to try to indicate major, minor, patch for both libs.

jamesmfriedman:
@amsheehan just upgraded the lib to 0.22.0. There are a lot of hidden undocumented changes, but I pulled in the majority of the ones I could find and made some other improvements.

I also corrected the doc errors and added the version to the readme.

I really need to design and build out a nice version of the docs, just haven't had the time to invest in that part, I've been busy building things with the lib :).

Let me know if anyone is interested in contributing. Thanks!

jamesmfriedman:
Hi @amsheehan any word on this? Excited to get some more feedback on it before I move past the beta designation.

lynnjepsen:
Hey @jamesmfriedman, sorry for the delay.

One of our requirements for adding a library to our main README is "Serve components in an à-la-carte delivery model"

https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md#examples

https://github.com/jamesmfriedman/rmwc seems to serve all components in one giant bundle. Is there a way for you to split up the delivery model?

jamesmfriedman:
Sure. Technically you could pull anything directly out of source but that’s not entirely obvious. The actual UMD lib was only 15kb gzipped so I hadn’t thought too much about it. 

It will take a bit of refactoring for the build, but I should be able to do it. I was actually looking at material-UI the other day, their build looks like one I could mimic. I’ll look into tackling that shortly :)

lynnjepsen:
Hey @jamesmfriedman, any progress on refactoring the build of React Material Web Components

jamesmfriedman:
Hi @lynnjepsen, I got it started and got sidetracked by the day job. Gonna try to wrap it up in the next couple of days!

jamesmfriedman:
Welp, that was a hell of a refactor! I broke apart the dependencies as follows:
- Most practical way was to just move the component directories out of src into the project root and rename them to their official component name
- I'm re-exporting them from the main index file so people can still import easily from the project root if they don't care about tree shaking
- I'm directly importing from the individual packages now instead of destructuring from material-components-web, this make tree shaking work properly
-  The current 'al-a-carte' files are ES6. I'm a little conflicted on distributing es6 code vs the babelfied version, and I'm at a bit of a loss without jumping through hoops to do so, but...
   - The current Preact library you have listed distributes as ES6
   - MDC does as well
   - this isnt a huge deal, but a minor annoyance of requiring babel to process this stuff on a production build
   - not really a problem for development anymore since Chrome is ES6 proficient

```javascript
// should import only the button code
import { Button } from 'rmwc/Button'

// not 100% how tree shaking works in this situation, but I believe
// this would pull in all of rmwc. Either way, it works.
import { Button } from 'rmwc'
```

lynnjepsen:
> I'm a little conflicted on distributing es6 code vs the babelfied version, and I'm at a bit of a loss without jumping through hoops to do so, but...

Like you said, MDC Web is in the same boat, so we'll both have to figure this out some day. But for today...This looks great!

jamesmfriedman:
Awesome, thanks @lynnjepsen!

Issue #992, opened 2017-07-21
Toolbar Menu Icon wrong class structure


Garbee:
### What MDC-Web Version are you using?

Current master build

### What browser(s) is this bug affecting?

N/A to browsers. This is a Developer API problem.

### What is the expected behavior?

Menu icon class should *modify* the `mdc-toolbar__icon` class. 

### What is the actual behavior?

Currently it is a standalone class on its own that inherits the primary class and adds the extra bits for the menu.

### Any other information you believe would be useful?

In BEM syntax `--` is a *modifier* of the root class. So it only appends new rules to tweak the look. This keeps the output more minimal since you aren't repeating rules for selectors everywhere. 

### Recommended solution

`mdc-toolbar__icon--menu` can either be made a true modifier and therefore become required with `mdc-toolbar__icon` or it can be made its own element, say `mdc-toolbar__menu-icon` for example. However, we should really assess whether or not this is needed at all. Why does this extra modifier exist for this one type of trigger?

yeelan0319:
@Garbee I definitely agree with what you said and this is definitely a bug not following BEM.

The reason we have `mdc-toolbar__icon` and `mdc-toolbar__menu-icon` separated is because the style behavior is different for those two, so instead of having one modify the other by overriding style rules, we decided two go with have them include the shared style via mixin.

To correct this, we would like to accept a PR for this bug. The concrete steps we should take:
- [ ] Change `mdc-toolbar__icon--menu` to `mdc-toolbar__menu-icon`
- [ ] Update the README.md and demo to reflect the changes

Garbee:
Yup, those changes LGTM to address the issue.

PR #1373, opened 2017-10-02
 fix(toolbar): update menu-icon className (#992)

bhongy:
Fixes [Toolbar Menu Icon wrong class structure #992](https://github.com/material-components/material-components-web/issues/992).


demos/toolbar/toolbar.html
```diff
@@ -0,0 +1,86 @@
+<!DOCTYPE html>
```
lynnjepsen: Why do you need this new file? Can you attach a screenshot of the demo pages before and after? I'm having trouble understanding why this is necessary.

bhongy: You are correct. I committed it by mistake (used it to test out the different mocks). Sorry for the confusion! Will squash.

PR #1376, opened 2017-10-02
refactor(infrastructure): Cut recompile time in half (12s to 5s)

acdvorak:
Don't build dist files in demo mode (they're not used).


webpack.config.js
```diff
@@ -170,58 +119,113 @@ module.exports = [{
...
+      toolbar: [path.resolve('./packages/mdc-toolbar/index.js')],
+    },
+    output: {
+      path: OUT_PATH,
+      publicPath: PUBLIC_PATH,
+      filename: 'mdc.[name].' + (IS_PROD ? 'min.' : '') + 'js',
+      libraryTarget: 'umd',
+      library: ['mdc', '[name]'],
+    },
+    // See https://github.com/webpack/webpack-dev-server/issues/882
```
kfranqueiro: Is the `devServer` block here and for `css` below even necessary anymore if we're not going to run these targets when we run the dev server?

kfranqueiro: (Additional thought, if we remove `devServer` on these, maybe move the comment that's above one of them to the first `devServer` config that's still relevant.)

acdvorak: Good call! Turns out you only need one `devServer` block, and it _must_ be in the first entry of the webpack config. It is completely ignored in later entries. Removed redundant blocks.

kfranqueiro: Seriously? Wouldn't have guessed that one... unintentionally dodged a bullet.

PR #1386, opened 2017-10-04
feat(textfield): Annotate textfield for Closure Compiler.

### LINKED ISSUE #345: Annotate mdc-textfield for closure

bwobrien:
Annotates MDC Textfield for the Closure Compiler.

Resolves #345

### DIFF
--- a/packages/mdc-textfield/adapter.js
+++ b/packages/mdc-textfield/adapter.js
@@ -74,7 +74,7 @@ class MDCTextfieldAdapter {
 
   /**
    * Returns true if classname exists for a given target element.
-   * @param {HTMLElement} target
+   * @param {?HTMLElement} target
    * @param {string} className
    * @return {boolean}
    */
@@ -83,14 +83,14 @@ class MDCTextfieldAdapter {
   /**
    * Registers an event handler on the root element for a given event.
    * @param {string} type
-   * @param {function(!Event)} handler
+   * @param {function(!Event): undefined} handler
    */
   registerTextFieldInteractionHandler(type, handler) {}
 
   /**
    * Deregisters an event handler on the root element for a given event.
    * @param {string} type
-   * @param {function(!Event)} handler
+   * @param {function(!Event): undefined} handler
    */
   deregisterTextFieldInteractionHandler(type, handler) {}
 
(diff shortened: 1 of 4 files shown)


packages/mdc-textfield/adapter.js
```diff
@@ -0,0 +1,194 @@
...
+
+  /**
+   * Returns an object representing the native text input element, with a
+   * similar API shape. The object returned should include the value, disabled
+   * and badInput properties, as well as the checkValidity() function. We never
+   * alter the value within our code, however we do update the disabled
+   * property, so if you choose to duck-type the return value for this method
+   * in your implementation it's important to keep this in mind. Also note that
+   * this method can return null, which the foundation will handle gracefully.
+   * @return {HTMLInputElement|NativeInputType}
```
acdvorak: Add [explicit nullable operator(s)](https://google.github.io/styleguide/jsguide.html#jsdoc-nullability):

`@return {?HTMLInputElement|?NativeInputType}`
or
`@return {!HTMLInputElement|!NativeInputType|null}`
or
`@return {HTMLInputElement|NativeInputType|null}`

(I'm not sure which variation is more readable, correct, and Closure-friendly...)

packages/mdc-textfield/foundation.js
```diff
@@ -233,17 +316,28 @@ export default class MDCTextfieldFoundation extends MDCFoundation {
...
 
+  /**
+   * @return {!HTMLInputElement|!NativeInputType} The native text input from the
+   * host environment, or a dummy if none exists.
+   * @private
+   */
   getNativeInput_() {
-    return this.adapter_.getNativeInput() || {
+    return this.adapter_.getNativeInput() ||
+    /** @type {NativeInputType} */ ({
```
acdvorak: Add [explicit nullability operator](https://google.github.io/styleguide/jsguide.html#jsdoc-nullability):

`@type {!NativeInputType}`

packages/mdc-textfield/index.js
```diff
@@ -103,6 +156,15 @@ export class MDCTextfield extends MDCComponent {
     };
   }
 
+  /**
+   * @return {!{
```
acdvorak: Should we create `@record` definitions for these collections of methods?

kfranqueiro: These subsets of the adapter API are kind of arbitrary (I assume we only split it like this as an attempt at readability); the only thing that really matters is the collective adapter API. Kind of unfortunate that we have to annotate these at all...

packages/mdc-textfield/adapter.js
bwobrien: Done. I think `null` is generally discouraged when `?` would suffice, so I went with `{?A|?B}`.

packages/mdc-textfield/foundation.js
bwobrien: Done, found some more and fixed throughout.

PR #1390, opened 2017-10-05
fix(drawer): Fixed drawer-toolbar-spacer align-items

stephenway:
`flex-center` isn't a valid align-items value, so I changed it to `center`.


stephenway:
@acdvorak @kfranqueiro  No problem, thanks for accepting it!

PR #1394, opened 2017-10-06
feat(textfield): Add textfield to the Closure whitelist.

bwobrien:
I missed this change with my previous PR #1386 for annotating Textfield for Closure.


bwobrien:
Fixed the compiler errors. PTAL.

PR #1395, opened 2017-10-06
docs(base): Update Closure Compiler annotation doc

bwobrien:
Updates the Closure Compiler annotation doc:

1. Applies Closure style to the examples.
2. Updates the examples with the doc's own recommendations (e.g. export statements go at the end), particularly for the provided skeleton.

### DIFF
--- a/docs/closure-compiler.md
+++ b/docs/closure-compiler.md
@@ -140,19 +140,21 @@ export {getFoo, getBar};
 
 ```js
 /** @enum {string} */
-export const cssClasses = {
+const cssClasses = {
   // ...
 };
 
 /** @enum {string} */
-export const strings = {
+const strings = {
   // ...
 };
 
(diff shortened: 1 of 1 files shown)


docs/closure-compiler.md
```diff
@@ -379,14 +396,14 @@ export default class MDCExampleAdapter {
   /**
    * Registers an event listener `handler` for event type `type` on the root element.
    * @param {string} type
-   * @param {!Function} handler
+   * @param {function(!Event): undefined} handler
```
kfranqueiro: Nit/question: Is there a preference RE `undefined` vs. `void` in closure type annotations?

bwobrien: I don't know for sure, but I assume undefined is preferred. I have never used void.

PR #1397, opened 2017-10-07
Add CDN usage

LukasDrgon:
I added [jsDelivr CDN links](https://www.jsdelivr.com/package/npm/material-components-web) to your readme to make it easier to use. jsDelivr is the fastest opensource CDN available and built specifically for production usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,17 @@ MDC-Web strives to seamlessly incorporate into a wider range of usage contexts, 
 
 Install the library
 
+Trough npm:
+
 ```
 npm install --save material-components-web
+```
+
+Or include it via CDN:
+
+```html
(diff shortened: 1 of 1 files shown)


Garbee:
I'd urge against including a CDN reference like this for now. Once the packages are closer to release candidate stages then CDNs can be assessed and decided on. For now, things are still in quite a bit of flux and we don't have data comparing CDNs all around. Even if this one is the "fastest" (which there is no data to back up this claim provided) there are still other things to think about like time to deploy new versions.

While supporting a CDN is certainly a good way to get new people onboarded more easily. It is not something that I feel needs to be addressed at this time. It should be handled with more care and data to support the choice to support rather than being the first one or two PRs submitted to support one.

lynnjepsen:
I agree with @Garbee

nebrelbug:
Are there any updates on having a CDN?

Issue #1258, opened 2017-09-07
Possible to see snackbars in hidden state on any platform with rubber-band scrolling at edges


kfranqueiro:
### What MDC-Web Version are you using?

0.20.0

### What browser(s) is this bug affecting?

Chrome or Safari on OS X

### What are the steps to reproduce the bug?

1. Go to http://material-components-web.appspot.com/snackbar.html
2. Scroll to the bottom and continue scrolling to cause the browser to rubber-band past the end of the screen

### What is the expected behavior?

It shouldn't be possible to see the hidden snackbars.

### What is the actual behavior?

The boxes from the "hidden" snackbars are visible.

![image](https://user-images.githubusercontent.com/398379/30181469-53af419e-93e1-11e7-8dc6-20de7553ed98.png)

### Any other information you believe would be useful?

Presumably we're leaving these elements in a convenient resting position to transition in. Possible solutions:

* Switch to using animations and set up keyframes so that the resting place is further offscreen but it immediately moves to the current resting place at the beginning of the animation
* Since this is a JS component, programmatically update the transition style (or apply another class) before becoming active / after the transition ends after becoming inactive

pgbross:
In #928 which is essentially seems to occur for the same reason as this one, one suggestion is that the visibility of the snack bar at rest could be set to "none". I implemented this in a customised adapter for a Vue snack bar component and it works for me and eliminates the artefacts.

I do not know if that is a good solution, but would involve only a minor change to implement it in the foundation if that idea was reviewed and passed accessibility and other requirements.

deap82:
FYI I've added this in my own scss file to solve the issue for now:


```
.mdc-snackbar {
	transition: mdc-animation-exit-permanent(transform, .25s), mdc-animation-exit-permanent(opacity, 0ms, .25s);
	opacity: 0;

	&.mdc-snackbar--active {
		transition: mdc-animation-enter(transform, .25s);
		opacity: 1;
	}
}
```

PR #1403, opened 2017-10-09
fix(snackbar): ensure snackbar is hidden when off-screen (#1258) [FOR DISCUSSION]

bhongy:
Address: #1258. This is a simple fix (see long-term suggestion under "notes") and ensure enter/exit transitions still work.

**Concerns:**
- couple style behavior to accessibility behavior. It relies on `aria-hidden` attribute. Changing accessibility behavior (unlikely) would cause style behavior to break unexpectedly.
- does not work without Javascript. Although, the widget will still work (degrade to this issue), and I am not sure how a snackbar widget would be used in JS-disabled environments.

**Alternatives, variations:**
- [deap82's suggestion](https://github.com/material-components/material-components-web/issues/1258#issuecomment-334368890) also works well. It does require ensuring the delay duration and transition duration matches (which could be achieved by using variables). Also, I believe you can still "tab to" the button (the subtree is not visibility-hidden or displayed none) —can be addressed with `tabindex="-1"` but would be more complex to use.
- I looked into using transition + delay for visibility (since it's transition-able). With the snippet below. But the enter transition did not work (just appears without transition for transform—due to `visibility: hidden` when the transition kicks off) (exit transition works fine). Same issue can be observed in [the demo on CSS-Tricks](https://css-tricks.com/snippets/css/toggle-visibility-when-hiding-elements/).

```scss
.mdc-snackbar {
  transform: translate(-50%, 100%);
  visibility: hidden;
  transition:
    visibility 0 linear .25s,
    mdc-animation-exit-permanent(transform, .25s);
  
  &--active {
    transform: translate(0);
    visibility: visible;
    transition:
      visibility 0 linear 0,
      mdc-animation-enter(transform, .25s);
  }
}
```

**Note:** knowing two states for animation generally is not enough (show, hide). We generally want  to do some work (imperatively) right before/after animation and sometimes we want to know the direction of the animation (action) ([action driven animation vs state driven animation](http://tobiasahlin.com/blog/meaningful-motion-w-action-driven-animation/)).

An idea for a longer-term fix might be creating a Javascript module that handles animation of an item entering/exiting the DOM. For mounting (showing), we inject the widget to the DOM (to a target container) then trigger enter transition. For unmounting (hiding), we trigger exit transition, wait for `transitionend` event then remove the widget from the DOM.

### DIFF
--- a/demos/snackbar.html
+++ b/demos/snackbar.html
@@ -73,8 +73,7 @@
       <section class="hero">
         <div class="mdc-snackbar mdc-snackbar--active"
              aria-live="assertive"
-             aria-atomic="true"
-             aria-hidden="true">
+             aria-atomic="true">
           <div class="mdc-snackbar__text">Message sent</div>
           <div class="mdc-snackbar__action-wrapper">
             <button type="button" class="mdc-snackbar__action-button">Undo</button>
--- a/packages/mdc-snackbar/mdc-snackbar.scss
+++ b/packages/mdc-snackbar/mdc-snackbar.scss
@@ -36,6 +36,12 @@
   pointer-events: none;
   will-change: transform;
 
+  // to match both <div aria-hidden="true"> and <div aria-hidden>
+  // if client manually add the attribute to the markup
+  &[aria-hidden] {
+    visibility: hidden;
+  }
+
   @include mdc-theme-dark(".mdc-snackbar") {
     background-color: $mdc-snackbar-background-color-on-dark;
   }


demos/snackbar.html
```diff
@@ -73,8 +73,7 @@
       <section class="hero">
         <div class="mdc-snackbar mdc-snackbar--active"
              aria-live="assertive"
-             aria-atomic="true"
-             aria-hidden="true">
```
bhongy: **Concern:** client knows deeply about the style implementation (expect aria-hidden to not exists so `visibility: hidden` does not apply).

Garbee:
> does not work without Javascript. Although, the widget will still work (degrade to this issue), and I am not sure how a snackbar widget would be used in JS-disabled environments.

Snackbars are a dynamic-oriented component. They aren't intended to work without JS. If you're working in that kind of an environment, your server needs to inject a notification into a good part of the page that will get the users attention on an action being completed.

PR #1405, opened 2017-10-09
Override default browser background-color <select> & <option> for 'CSS only' dark theme

trollderius:
Override default browser background-color ```<select>``` & ```<option>``` for 'CSS only' dark theme:

```
.mdc-select {
  [...]
  background-color: transparent; 
  [...]

  option.mdc-list-item{
    @include mdc-theme-dark {
      background-color: #424242;
    }
  }
}
```
  
(And add .mdc-list-item in <option>'s in demo)

### DIFF
--- a/demos/select.html
+++ b/demos/select.html
@@ -143,13 +143,13 @@
       <section class="example">
         <h2 class="mdc-typography--title">CSS Only</h2>
         <select class="mdc-select">
-          <option value="" selected>Pick a food group</option>
-          <option value="grains">Bread, Cereal, Rice, and Pasta</option>
-          <option value="vegetables" disabled>Vegetables</option>
-          <option value="fruit">Fruit</option>
-          <option value="dairy">Milk, Yogurt, and Cheese</option>
-          <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
-          <option value="fats">Fats, Oils, and Sweets</option>
+          <option class="mdc-list-item" value="" selected>Pick a food group</option>
+          <option class="mdc-list-item" value="grains">Bread, Cereal, Rice, and Pasta</option>
+          <option class="mdc-list-item" value="vegetables" disabled>Vegetables</option>
+          <option class="mdc-list-item" value="fruit">Fruit</option>
+          <option class="mdc-list-item" value="dairy">Milk, Yogurt, and Cheese</option>
+          <option class="mdc-list-item" value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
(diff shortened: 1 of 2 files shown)


packages/mdc-select/mdc-select.scss
```diff
@@ -95,6 +96,12 @@
     white-space: nowrap;
     overflow: hidden;
   }
+
+  option.mdc-list-item {
```
Garbee: Why have the `option` selector on this at all?

trollderius: Because it is not necessary to override the option background in a 'js select' like ```li.mdc-list-item```. His parent ```.mdc-theme--dark .mdc-simple-menu``` already has this background.

lynnjepsen:
@trollderius could you maybe provide a code pen or screenshot of exactly what it is you're trying to change?

demos/select.html
```diff
@@ -143,13 +143,13 @@ <h2 class="mdc-typography--title">Fully-Featured Component</h2>
...
         <h2 class="mdc-typography--title">CSS Only</h2>
         <select class="mdc-select">
-          <option value="" selected>Pick a food group</option>
-          <option value="grains">Bread, Cereal, Rice, and Pasta</option>
-          <option value="vegetables" disabled>Vegetables</option>
-          <option value="fruit">Fruit</option>
-          <option value="dairy">Milk, Yogurt, and Cheese</option>
-          <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
-          <option value="fats">Fats, Oils, and Sweets</option>
+          <option class="mdc-list-item" value="" selected>Pick a food group</option>
```
amsheehan: I don't think this will actually do anything

packages/mdc-select/mdc-select.scss
amsheehan: Lets take a step back... Is this even doing anything?

Regardless we want to avoid tag selectors at all cost.

amsheehan:
I see now that you are targeting the CSS-only multi-select. If that's the case, then instead of selecting on `option`, please select on `.mdc-list-item`s that are descendants of `.mdc-multi-select`. E.g.:

```
.mdc-multi-select {
  <...>

  .mdc-list-item {
    @include mdc-theme--dark(".mdc-select",  true) {
        @include mdc-theme-prop(background-color, text-secondary-on-dark);
    }

    <...>
  }
}
```

Further, we already have the dark theme mixin set up for multi-select list items [here](https://github.com/material-components/material-components-web/blob/5dabcdf72606e182d44bdaddd386e47ae2da0d77/packages/mdc-select/mdc-select.scss#L336)

lynnjepsen:
We've removed mdc-theme--dark, so this PR is out of date.

PR #1406, opened 2017-10-10
refactor(ripple): Combine base-color and theme-style into `color`

acdvorak:
- Similar to the `mdc-theme-prop` mixin, this change makes it simpler for callers to use a single property for all theme/color values
- Update CSS 4 color function name from `color` to `color-mod` per [spec](https://drafts.csswg.org/css-color/#modifying-colors)
- Backward compatible with existing usages


acdvorak:
Closing this because @kfranqueiro will be doing the same thing in a separate PR.

PR #1408, opened 2017-10-10
fix(demos): Make unsupported ripple message visible and improve detection

acdvorak:
Message was previously:
- Hidden behind the toolbar
- Not displayed in Edge, even though Edge doesn't support ripple

Before:
image

After:
image


acdvorak:
Good call. Updated the list of browsers in the message. PTAL!

PR #1418, opened 2017-10-12
Ripple/events2

aprigogin:
Added optional event parameters to activate/deactivate methods.
Changed non-static getter to be a private method to keep jscompiler happy.


aprigogin:
I missed some changes that are needed while wrangling git. Made them locally, but so far was not able to merge them into this PR.

packages/mdc-ripple/foundation.js
```diff
@@ -266,6 +267,9 @@ class MDCRippleFoundation extends MDCFoundation {
     });
   }
 
+  /**
```
kfranqueiro: I would suggest:

```js
activate(event = null) {
  this.activate_(event);
}
```

And the same for deactivate.

I'd prefer avoiding single-letter variable names because they tend to be harder to quickly understand as well as search.

packages/mdc-ripple/foundation.js
```diff
@@ -188,6 +177,18 @@ class MDCRippleFoundation extends MDCFoundation {
...
 
   /**
+   * We compute this property so that we are not querying information about the client
+   * until the point in time where the foundation requests it. This prevents scenarios where
+   * client-side feature-detection may happen too early, such as when components are rendered on the server
+   * and then initialized at mount time on the client.
+   * @return {boolean}
+   * @private
+   */
+  isSupported_() {
```
kfranqueiro: I thought we were going to examine making this a static getter? But given that it's internal I can go either way on it.

aprigogin: I thought the intent is to have implementer be able to override this via adapter. As such it cannot be static method - needs instance. If the intent is to have this always call util.browserSupportsCssVars, we can do it from here. Let me know what you think.

kfranqueiro: Oh because the adapter is connected to the instance... yup, that would be why I thought I remembered there being a reason why we couldn't do that.

PR #1423, opened 2017-10-13
 refactor(textfield): De-dupe two style rules.

bwobrien:
De-dupe two style rules in mdc-textfield. Remove one that was never applied, and mark the other as @alternate.


packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -266,7 +267,6 @@
   }
 
   &.mdc-textfield--disabled {
-    @include mdc-theme-prop(background-color, $mdc-textfield-light-background);
```
kfranqueiro: I'm actually wondering whether this style (which is fully-opaque white) or the hard-coded one below is intended to be used for disabled textfield boxes... cc @amsheehan

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -53,6 +53,7 @@
     border: none;
     border-bottom: 1px solid $mdc-textfield-underline-on-light-idle;
     background: none;
+    /* @alternate */
```
kfranqueiro: Would it be better to remove `font-size` from the `@each` a few lines up instead, then we can probably just remove `font-size: inherit` too? (Assuming I'm not being an idiot, and this doesn't cause a display regression)

packages/mdc-textfield/mdc-textfield.scss
bwobrien: Confirmed that black 2% opacity is correct. Made it a variable for easier changes in the future.

packages/mdc-textfield/mdc-textfield.scss
bwobrien: Looked into it a bit, and I'm going to leave as-is for now. The `font-size` in the `@each` above is the correct one, but it is always overriden by this definition here. So the correct fix is to remove this `font-size`, but that might cause regressions and this is just a refactor.

bwobrien: Closing this out: I removed the first `font-size` rule since it never applies currently, with a TODO to re-add it later. Thus I can remove the `@alternate` entirely.

PR #1438, opened 2017-10-17
fix(drawer): Fix typo on `align-items` value

acdvorak:
There is no `flex-center` value; only `flex-start`, `flex-end`, and `center`.

See the [MDN docs on `align-items`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items).


acdvorak:
Duplicate of #1390

PR #1446, opened 2017-10-18
feat(textfield): Add mixin allowing customization of border radii

### LINKED ISSUE #1422: Allow developers to customize the corner radius of text field

amsheehan:
resolves #1422


packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -163,10 +163,13 @@
 }
 
 .mdc-textfield--box {
+  @include mdc-text-field-corner-radius-all(0);
```
kfranqueiro: Is it really necessary to have 2 separate mixins for this? It is possible to specify the 4 corners separately with `border-radius` as well, which would also avoid needing to define 3 styles instead of one here.

packages/mdc-textfield/README.md
```diff
@@ -279,6 +279,19 @@ text, validation, and dense UI.
...
 ```
 
+### Sass Mixins
+
+To customize a Text Field's border radii, you can use the following mixins.
+
+#### `mdc-text-field-corner-radius($position, $radius)`
+
+This mixin is provided for customizing the border radius for a corner of a Text Field at the given position,
+e.g.: `@include mdc-text-field-corner-radius(top-left, 10px)` will apply `border-top-left-radius: 10px` to that Text Field.
```
lynnjepsen: don't document what the mixin will output. These are implementation details.

packages/mdc-textfield/mdc-textfield.scss
lynnjepsen: +1, I think we should just remove mdc-text-field-corner-radius-all....its not necessary for the feature request, and it doesn't provide much utility.

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -176,12 +179,13 @@
 
   &::before,
   &::after {
+    @include mdc-text-field-corner-radius-all(50%);
```
lynnjepsen: don't use the mixin for ::before and ::after elements. Then `mdc-text-field-corner-radius` mixin only applies to the `mdc-textfield` element...not it's before and after pseudo elements.

just use border-radius: 50%

packages/mdc-textfield/README.md
```diff
@@ -279,6 +279,19 @@ text, validation, and dense UI.
 </div>
 ```
 
+### Sass Mixins
+
+To customize a Text Field's border radii, you can use the following mixins.
+
+#### `mdc-text-field-corner-radius($position, $radius)`
```
lynnjepsen: Does this mixin only actually work (Aka show a rounded corner) if you're using `mdc-textfield--box` or `mdc-textfield--textarea`?

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -510,18 +516,18 @@
...
   border: 1px solid $mdc-textarea-border-on-light;
-  border-radius: 4px;
 
   @include mdc-theme-dark(".mdc-textfield") {
     border-color: $mdc-textarea-border-on-dark;
   }
 
   // stylelint-disable plugin/selector-bem-pattern
   .mdc-textfield__input {
+    @include mdc-text-field-corner-radius-all(2px);
```
lynnjepsen: I THIIIINK you need to update your `mdc-text-field-corner-radius` mixin to be like this

```
@mixin mdc-text-field-corner-radius($position, $radius) {
  .mdc-textfield--box, .mdc-textfield--textarea {
    border-#{$position}-radius: $radius;
  }
  .mdc-textfield--textarea .mdc-textfield__input {
    border-#{$position}-radius: $radius - 2;
  }
}
```

Or something like that. The mixin should definitely handle keeping `.mdc-textfield--textarea` and `.mdc-textfield__input` in sync.

packages/mdc-textfield/mdc-textfield.scss
```diff
@@ -682,7 +688,7 @@
 
   &::before,
   &::after {
-    border-radius: 0;
+    @include mdc-text-field-corner-radius-all(0);
```
lynnjepsen: Again, just use border-radius: 0;

packages/mdc-textfield/mdc-textfield.scss
kfranqueiro: Actually I was arguing the opposite - the `all` mixin (maybe without the `-all` in the name) is the one thing we need, because I think it should be feasible to send it anything that will work with `border-radius`, whether that be 1 value or 4.

packages/mdc-textfield/mdc-textfield.scss
kfranqueiro: Protip: these styles will no longer exist once I update things to use new ripple mixins.

But yeah, I don't think this should be needed at all, because it should already be clipped by the parent element?

packages/mdc-textfield/mdc-textfield.scss
kfranqueiro: Hm. Then my other idea (of only needing one mixin that does what `-all` does) might fly out the window.

packages/mdc-textfield/README.md
amsheehan: Sure does

packages/mdc-textfield/README.md
amsheehan: I was using this as shorthand to demonstrate that the positional argument should be one of the four possible positions in css, e.g.: top-left, top-right, bottom-left, bottom-right

Open to changing this though, as it seems we will take the approach of making the mixin more component specific, than a general purpose border radius mixin

amsheehan:
Alright. Let me change a bunch of stuff, one sec...

amsheehan:
Alright, so now we have two mixins, one for filled text field and one for textarea (I'll add an outlined text field mixin to the experimental branch)

One thing that deviates from other components mixins is that the default value for the border radius here is a sass variable marked `!default`. The `mdc-textfield.scss` file passes this when it uses either mixin. So what happens is, if the user wants to style textarea's and text fields differently with respect to border radius, they can use the custom class/mixin variant combo. If they want all text fields (either filled or textarea) to have the _same_ border radius, they can just override the variable.
 
Thoughts?

packages/mdc-textfield/_mixins.scss
```diff
@@ -14,6 +14,31 @@
...
 
+@mixin mdc-text-field-box-corner-radius($radius) {
+  border-radius: $radius $radius 0 0;
+}
+
+@mixin mdc-text-field-textarea-corner-radius($radius) {
+  border-radius: $radius;
+
+  .mdc-textfield__label {
+    border-radius: $radius;
```
kfranqueiro: Might want to flatten the bottom of this to prevent parts of text from "poking" out with larger border radius values

packages/mdc-textfield/README.md
```diff
@@ -279,6 +279,20 @@ text, validation, and dense UI.
...
 
+### Sass Mixins
+
+To customize a Filled Text Field or a Text Field `textarea`'s border radius, you can use the following mixins.
+Alternatively, if you would like to change the border radius for every instance of MDC Text Field uniformly, regardless
+of the variant, you can override $mdc-text-field-border-radius in Sass.
+
+#### `mdc-text-field-box-corner-radius($radius)`
+
+This mixin is provided for customizing the border radius for Text Field Box.
```
kfranqueiro: Shorten? "Customizes the border radius for a Text Field Box."

packages/mdc-textfield/README.md
```diff
@@ -279,6 +279,20 @@ text, validation, and dense UI.
...
+
+#### `mdc-text-field-textarea-corner-radius($radius)`
+
+This mixin is provided for customizing the border radius for a Text Field `textarea`.
```
kfranqueiro: Shorten? "Customizes the border radius for a Text Field `textarea`."

packages/mdc-textfield/_mixins.scss
```diff
@@ -14,6 +14,31 @@
...
+  }
+
+  .mdc-textfield__input {
+    // The input element is required to have 2 pixels shaved off
+    // of the radius of its parent. This prevents an ugly space of
+    // background between the two borders in each corner when the
+    // textarea is focused. This also means we need to guard against
+    // invalid output with @if/@else directives.
+    @if ($radius >= 2) {
```
kfranqueiro: This could be condensed: `border-radius: max($radius - 2, 0);`

packages/mdc-textfield/_variables.scss
```diff
@@ -49,3 +49,5 @@ $mdc-textarea-light-background: rgba(white, 1);
 $mdc-textarea-dark-background: rgba(48, 48, 48, 1);
 $mdc-textarea-disabled-background-on-light: rgba(249, 249, 249, 1);
 $mdc-textarea-disabled-background-on-dark: rgba(47, 47, 47, 1);
+
+$mdc-text-field-border-radius: 4px !default;
```
kfranqueiro: My only comment RE this is we might want to go back to other things we've mixin-ified and evaluate whether we want to do this for any parameters in those components.

packages/mdc-textfield/README.md
amsheehan: Eh

packages/mdc-textfield/README.md
amsheehan: eh

amsheehan:
D'oh! Removed.

Issue #1104, opened 2017-08-09
Consider making a common package for browser utility functions


kfranqueiro:
I noticed we currently have 3 packages (drawer, ripple, toolbar) that define `util.applyPassive` for detecting passive event support and making use of it, and at least 2 packages (dialog and ripple) that define a function to detect support for custom CSS properties.

This leads to redundant implementation and test code, and in the latter case also has led to the two implementations skewing apart from each other.

We should consider putting common utilities like this in a centralized reusable package.  We should also probably check whether there is a reason (other than an oversight) why `supportsCssCustomProperties` in dialog's util hasn't kept up with the edge cases added to ripple's util.

Spetnik:
My two cents: the lack of such a package threw me for a loop when replacing Array.prototype.find here: https://github.com/material-components/material-components-web/pull/1461/commits/5aa2b6b7d9e1dcc5383ae5bd90d4861d9b9dae31#diff-ab2630d7dd0cee3811f920a34f5efc3b

I felt like I should create a simple polyfill function (e.g. https://github.com/Spetnik/ie-Array.find/blob/master/README.md) but I really couldn't find a place where it made sense to add it. I would assume that this proposed package would be the perfect place for such a function.

acdvorak:
We don't use polyfills in MDC-Web because we don't want to conflict with other libs or create weird hard-to-trace bugs, but I would love to create a centralized set of ponyfills (similar to polyfills, but without monkey-patching global prototypes) and other common utilities like `resize` event throttling.

abhiomkar:
We now have `applyPassive` in @material/dom/events

Closing this.

PR #1448, opened 2017-10-18
feat(shape): Adding MDCShape, an experimental subsystem

lynnjepsen:
Also updated our webpack config and dependency test to allow demo-only JS.


demos/shape/heart/foundation.js
```diff
@@ -0,0 +1,121 @@
...
+      + ' ' + (-1 * realWidth) + ',' + (-1 * this.generateBaseCurveHeight2_(height, padding))
+      + ' ' + (-1 * realWidth)
+      + ',' + (-1 * (this.generateBaseHeight_(height, padding) - flatHeight));
+  }
+  generateRightBase_(width, height, padding) {
+    const flatWidth = this.generateBaseFlatWidth_(width, padding);
+    const flatHeight = this.generateBaseFlatHeight_(height, padding);
+    const curveHeight = this.generateBaseHeight_(height, padding) - flatHeight;
+    const realWidth = this.generateBaseWidth_(width, padding) - flatWidth;
+    return 'c 0,' + (curveHeight - this.generateBaseCurveHeight2_(height, padding))
```
Garbee: This looks like a good place to use template strings in JS to make this return statement a bit simpler.

demos/shape/heart/foundation.js
```diff
@@ -0,0 +1,121 @@
...
+      + (-1 * (realWidth - this.generateBaseCurveWidth_(width, padding)))
+      + ',' + (curveHeight - this.generateBaseCurveHeight_(height, padding))
+      + (-1 * realWidth) + ',' + curveHeight
+      +'l ' + (-1 * flatWidth) + ',' + flatHeight;
```
Garbee: nit: space after first `+` operator.

kfranqueiro: Another nit... do we not prefer trailing operators rather than leading? (Maybe moot if we reorganize this and use a template string.)

packages/mdc-shape/_variables.scss
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+$mdc-shape-shadow-padding: 56px;
+$mdc-shape-shadow-negative-padding: -1 * $mdc-shape-shadow-padding;
```
kfranqueiro: Is there a motive behind making this a variable rather than simply doing -1 * mdc-shape-shadow-padding in the 2 cases where this is needed? The presence of a separate variable makes me think there'd be more to it than simply negating the other one.

packages/mdc-shape/foundation.js
```diff
@@ -0,0 +1,419 @@
...
+  get finalUmbraShadowBlur_() {
+    return UMBRA_BLUR[this.elevation_];
+  }
+
+  get finalUmbraShadowSpread_() {
+    return UMBRA_SPREAD[this.elevation_];
+  }
+}
+
+const UMBRA_OFFSET_Y = [
```
kfranqueiro: Why are these constants here rather than in a constants file? Also, are these more directly related to elevation rather than shapes? I suspect these correspond to styles in mdc-elevation/_variables.scss and it would be *really* easy for these to get out of sync with those, being in a completely separate package.

Is there any other stuff here that should be considered for mdc-elevation instead? Should shapes be dependent on elevation?

packages/mdc-shape/component.js
```diff
@@ -0,0 +1,57 @@
...
+      getCanvasHeight: () => this.canvas_.offsetHeight,
+      getDevicePixelRatio: () => window.devicePixelRatio,
+      create2dRenderingContext: () => this.canvas_.getContext('2d'),
+      createPath2D: (path) => new Path2D(path),
+      setClipPathData: (value) => this.path_.setAttribute('d', value)};
+  }
+
+  initialSyncWithDOM() {
+    this.foundation_.setBackground(getBackgroundFromDOM(this.root_), false);
+    this.foundation_.setElevation(getElevationFromDOM(this.root_), false);
```
kfranqueiro: AFAICT, `getElevationFromDOM` assumes that CSS custom properties work, which in turn makes me suspect we need to be feature-detecting this the same way that we're doing in ripple (and IIRC a couple of other places, that probably need updating)...

Maybe it's time for #1104?

What's our plan for browsers that don't support custom CSS properties?

OTOH, is there a reason that setting elevation needs to be done this way rather than actually reusing mdc-elevation classes (esp. if there's any desire to support no-JS shapes), or just using a data attribute if this is solely for the purpose of initialization via JS?

packages/mdc-shape/foundation.js
```diff
@@ -0,0 +1,419 @@
...
+
+  setBackground(background) {
+    this.background_ = background;
+  }
+
+  setElevation(value, animate) {
+    this.elevating_ = false;
+    this.elevation_ = value;
+
+    if (!animate) {
```
kfranqueiro: Suggestion: flip the if/else so that the `if` performs a positive comparison?

scripts/check-pkg-for-release.js
```diff
@@ -44,13 +44,16 @@ const MASTER_PKG = require(path.join(process.env.PWD, MASTER_PKG_PATH));
...
+const NOT_PUBLISHABLE = ['shape'];
 
 main();
 
 function main() {
   checkPublicConfigForNewComponent();
   if (pkg.name !== MASTER_PKG.name) {
     checkNameIsPresentInAllowedScope();
+  }
+  if (NOT_PUBLISHABLE.indexOf(pkg.name) !== -1) {
```
kfranqueiro: Uh... is this inverted? IIUC this *will only* execute the statements below if the package is *not* publishable.

demos/shape/shape.html
```diff
@@ -0,0 +1,104 @@
+<!DOCTYPE html>
```
kfranqueiro: Nit: Rename this to shape/index.html?

demos/shape/shape.html
```diff
@@ -0,0 +1,104 @@
...
+          let elevatableHeart;
+          for (var i=0; i<shapes.length; i++) {
+            const demoHeart = DemoHeart.attachTo(shapes[i]);
+            if (shapes[i].id === "elevatable-shape") {
+              elevatableHeart = demoHeart;
+            }
+          }
+          const elevationInput = document.querySelector("#elevation-input");
+          elevationInput.addEventListener("change", () => {
+            elevatableHeart.elevation = elevationInput.value;
```
kfranqueiro: I'm noticing that there is sometimes visible jank when changing this elevation and I'm not sure whether it's a flicker of the clip-path resetting or what.

demos/shape/shape.scss
```diff
@@ -0,0 +1,22 @@
...
+@import "../../packages/mdc-shape/mdc-shape";
+
+fieldset {
+  margin: 24px;
+  margin-top: 0;
+  margin-bottom: 16px;
+}
+.mdc-shape {
+  --mdc-shape-elevation: 4;
+  --mdc-shape-background: #FF00FF;
```
kfranqueiro: Nit (here and anywhere else applicable): lowercase hex values

packages/mdc-shape/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+TODO
+
+## Usage
+
+### DOM Structure
+
+```html
+<div class="mdc-shape">
+  <canvas class="mdc-shape__canvas"></canvas>
```
kfranqueiro: Why is this a canvas and does it always have to be? Is shape expected to eventually work on other components?

packages/mdc-shape/mdc-shape.scss
```diff
@@ -0,0 +1,39 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "./variables";
+
+$mdc-shape-shadow-double-padding: 2 * $mdc-shape-shadow-padding;
```
kfranqueiro: I'm confused why we're doing this here yet we have mdc-shape-shadow-negative-padding in variables.scss. For that matter, maybe it'd be easier to understand if we simply set all of top/left/bottom/right to the same thing, instead of requiring this math?

I'm also not sure I understand what the padding on each side of the canvas is useful for?

packages/mdc-shape/util.js
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!mdcShapeElevation) {
+    mdcShapeElevation = el.style.getPropertyValue('--mdc-shape-elevation');
+  }
+
+  if (mdcShapeElevation) {
+    elevation = parseInt(mdcShapeElevation);
+  }
+
+  return elevation;
```
kfranqueiro: This is going to silently return 0 in the case where elevation can't be computed by these means. Is that what we want?

packages/mdc-shape/README.md
```diff
@@ -0,0 +1,100 @@
...
+mdcShape.background = '#FOO';
+```
+
+To modify the elevation of a shape, call the elevation setter
+
+```
+mdcShape.elevation = 4;
+```
+
+The second way is to use custom CSS properties.
```
kfranqueiro: Should we really expose 2 ways to do the same thing, especially if the 2nd way won't work in all browsers? (I have more thoughts on the CSS custom properties approach in another comment)

demos/shape/heart/foundation.js
amsheehan: My Outlined Text Field prototype uses leading operators, and the king of SVG on the web Mike Bostock always uses leading operators. I am open to more clearly defining our convention here, and am not married to leading operators.

kfranqueiro: I guess it comes down to whether we want to adhere to [Google's JS style guide](https://google.github.io/styleguide/jsguide.html#formatting-where-to-break) for this, which promotes breaking after the operator, not before.

Garbee: I think the closer we are to the Google style guide the better. Since that is the companies encouraged policy on coding style.

However, I still think shifting these to template literals would net a simpler codebase to maintain long-term.

packages/mdc-shape/foundation.js
```diff
@@ -0,0 +1,419 @@
...
+      this.ambientShadowOffsetY_ = this.finalAmbientShadowOffsetY_;
+      this.penumbraShadowBlur_ = this.finalPenumbraShadowBlur_;
+      this.penumbraShadowOffsetY_ = this.finalPenumbraShadowOffsetY_;
+      this.umbraShadowBlur_ = this.finalUmbraShadowBlur_;
+      this.umbraShadowOffsetY_ = this.finalUmbraShadowOffsetY_;
+      this.umbraShadowSpread_ = this.finalUmbraShadowSpread_;
+      finished = true;
+    }
+
+    this.redraw();
```
amsheehan: Seems like stuff is happening in `init()` that doesn't need to happen every time `redraw()` is called. Doesn't seem destructive yet, just seems unnecessary.

packages/mdc-shape/foundation.js
```diff
@@ -0,0 +1,419 @@
...
+
+    if (!finished && this.elevating_) {
+      this.animationFrameId_ = requestAnimationFrame(() => this.elevate_());
```
amsheehan: Is this going to be hard to test?

packages/mdc-shape/foundation.js
```diff
@@ -0,0 +1,419 @@
...
+    } else {
+      this.elevating_ = false;
+    }
+  }
+
+  clear_() {
+    this.ctx_.clearRect(0, 0, this.adapter_.getCanvasWidth(), this.adapter_.getCanvasHeight());
```
amsheehan: I think we should rethink `redraw()` :)

packages/mdc-shape/foundation.js
amsheehan: Plus One

lynnjepsen:
Resolves #1535

demos/shape/shape.html
lynnjepsen: There was a bug with parsing the elevation value into an int, this was causing the penumbra shadow to draw in the wrong place. Fixed the bug and now I see now jank.

FYI, the clip-path doesn't change when the elevation changes .... but the entire canvas redraws. Its not GPU accelerated, but thats kinda the deal with shapes.

packages/mdc-shape/README.md
lynnjepsen: It does always have to be a canvas.

Initially we had an implementation that clipped divs with clip-path, then applied a drop-shadow filter. This has a couple performance problems
* Calculating drop-shadow requires iterating over every pixel within that div. So the larger the div the slower the performance
* The drop-shadow is repainted (and therefore recalculated) whenever the content underneath the clipped div changed. (Technically, whenever any content in the same GPU tile changes.)  This means if you were scrolling content under the clipped div, the scroll could become janky.

Also for working with other components...yes some day. Since this is still experimental we've only built some offhand prototypes (like for button). In the future most components will probably have some "shape" version, in addition to the default version, which has a modified HTML structure to include canvas. Obviously, these will be *very* heavy weight components and should be used with discretion.

packages/mdc-shape/README.md
lynnjepsen: Fair enough, I removed the custom CSS properties approach.

packages/mdc-shape/_variables.scss
lynnjepsen: uh nope, this was silly. I removed it.

packages/mdc-shape/component.js
lynnjepsen: Theres no support for no-JS shapes...Shape JS has to draw the shape on a canvas, and it has to draw all the shadows on the canvas. The existing Sass mixins for elevation won't work with shapes because the shadows won't properly follow the irregular edges of a shape.

packages/mdc-shape/foundation.js
lynnjepsen: Moved it to mdc-elevation!

packages/mdc-shape/mdc-shape.scss
lynnjepsen: I think I was struggling with how to use a Sass variable within a calc function....fixed it now

The padding is to make sure the canvas is large enough to cover the highest elevation (24) with the largest blur (56px). This way there will always be enough room on the canvas to draw the shadow, regardless of what elevation a shape is at.

scripts/check-pkg-for-release.js
lynnjepsen: oops, fixed this up

packages/mdc-shape/README.md
```diff
@@ -0,0 +1,100 @@
...
+  <svg class="mdc-shape__svg">
+    <clipPath id="FOO_ID">
+      <path class="mdc-shape__path"/>
+    </clipPath>
+  </svg>
+</div>
+```
+
+We recommend you put any content for the shape inside the "clipped" element *after* the `mdc-shape__canvas` element. The "clipped" element must contain the style `clip-path:url(#FOO_ID)`, where `FOO_ID` corresponds to the value you assign to the `clipPath` element's `id` attribute. Using the same ID between the "clipped" element and the `clipPath` element effectively clips any content to the shape. It is important `mdc-shape__canvas` is before the "clipped" element, otherwise the clipping effect will clip the shadows drawn by `MDCShapeFoundation`.
+
```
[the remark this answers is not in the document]
lynnjepsen: any suggestions? Not sure which direction you want to take this... "cropped" or "limited to this area covered by svg"?

packages/mdc-shape/README.md
```diff
@@ -0,0 +1,100 @@
...
+| Method Signature | Description |
+| --- | --- |
+| `setCanvasWidth(value: number) => void` | Sets the width of the canvas element. |
+| `setCanvasHeight(value: number) => void` | Sets the height of the canvas element. |
+| `getCanvasWidth() => number` | Returns the width of the canvas element. |
+| `getCanvasHeight() => number` | Returns the height of the canvas element. |
+| `getDevicePixelRatio() => number` | Returns the device pixel ratio. |
+| `create2dRenderingContext() => {shadowColor: string, shadowBlur: number, shadowOffsetY: number, fillStyle: string, scale: (number, number), clearRect: (number, number, number, number), fill: (Path2D)}` | Returns an object which has the shape of a CanvasRenderingContext2d instance. An easy way to achieve this is simply `this.root_.querySelector(mdc.shape.MDCShapeFoundation.SHAPE_SELECTOR).getContext('2d');`. |
+
```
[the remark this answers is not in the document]
lynnjepsen: I moved it down to another section, hopefully that reads better

packages/mdc-shape/foundation.js
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  static get defaultAdapter() {
+    return {
+      setCanvasWidth: ( /* value: number */ ) => {},
+      setCanvasHeight: ( /* value: number */ ) => {},
+      getCanvasWidth: () => /* number */ 0,
+      getCanvasHeight: () => /* number */ 0,
+      getDevicePixelRatio: () => /* number */ 1,
+      create2dRenderingContext: () =>
```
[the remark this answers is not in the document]
lynnjepsen: closure is not complaining yet, but it will once we add it to the whitelist, and then it will need all these methods. So I think I'll leave them for that day.

packages/mdc-shape/foundation.js
lynnjepsen: I moved some of the canvas stuff to a separate method. This way we don't re-create2dRenderingContext on every call to redraw.

The sync with canvas is important though, because the canvas could have changed size.

packages/mdc-shape/foundation.js
lynnjepsen: I wrote a test!

test/unit/mdc-shape/foundation.test.js
```diff
@@ -0,0 +1,234 @@
...
+test('default adapter returns a complete adapter implementation', () => {
+  verifyDefaultAdapter(MDCShapeFoundation, [
+    'setCanvasWidth', 'setCanvasHeight', 'getCanvasWidth', 'getCanvasHeight', 'getDevicePixelRatio',
+    'create2dRenderingContext', 'createPath2D', 'setClipPathData',
+  ]);
+
+  const renderingContext = MDCShapeFoundation.defaultAdapter.create2dRenderingContext();
+  const renderingContextMethods =
+    Object.keys(renderingContext).filter((k) => typeof renderingContext[k] === 'function');
+  assert.deepEqual(renderingContextMethods.slice().sort(), ['scale', 'clearRect', 'fill'].slice().sort());
```
kfranqueiro: Is either `slice` call here required?

test/unit/mdc-shape/foundation.test.js
```diff
@@ -0,0 +1,234 @@
...
+    shadowOffsetYs.push(mockContext.shadowOffsetY);
+  };
+
+  foundation.init();
+  foundation.setElevation(2, true);
+  mockRaf.flush();
+
+  assert.equal(fillCount, 4);
+  assert.equal(shadowBlurs[0], 1);
+  assert.equal(shadowBlurs[1], 0.4);
```
kfranqueiro: Should some of these values (either here or below after flushing) be looked up from the constants in mdc-elevation?

test/unit/mdc-shape/foundation.test.js
```diff
@@ -0,0 +1,234 @@
...
+  assert.equal(shadowBlurs[2], 0.2);
+  assert.equal(shadowOffsetYs[0], 0.2);
+  assert.equal(shadowOffsetYs[1], 0.4);
+  assert.equal(shadowOffsetYs[2], 0.6000000000000001);
+  td.verify(mockAdapter.createPath2D('m 56.2 56.2 h 100 v 200'));
+
+  // requires 9 frames to finish animation
+  for (let i=0; i<9; i++) {
```
kfranqueiro: I'm a bit confused why this is necessary... isn't our mock raf implementation supposed to flush all pending animation frames at once?

PR #1449, opened 2017-10-18
feat(elevation): Update mixin to accept theme color param

### LINKED ISSUE #1534: Update elevation/shadow Sass mixins to customize color
We're experimenting with letting users customize the color of their shadows. This means the mdc-elevation Sass mixins need to support color as an input argument

acdvorak:
Resolves #1534

Screenshot of custom elevation color


packages/mdc-elevation/_mixins.scss
```diff
@@ -14,25 +14,43 @@
...
+
+  @if type-of($color) != color {
+    @error "$color must be a valid color, but '#{$color}' is of type #{type-of($color)}";
+  }
+
+  @if alpha($color) < 1 {
+    @error "$color must have an alpha of 1, but '#{$color}' has an alpha of #{alpha($color)}";
+  }
+
+  $umbra-color: rgba($color, .2);
```
kfranqueiro: Should these alpha values also be listed as variables (constants)?

acdvorak: Done

lynnjepsen:
Resolves #1534

acdvorak:
Got design approval

demos/elevation.scss
```diff
@@ -18,3 +18,7 @@
 
 @import "../packages/mdc-elevation/mdc-elevation";
 @import "../packages/mdc-typography/mdc-typography";
+
+.demo-elevation--custom-color {
+  @include mdc-elevation(16, $color: secondary-dark, $opacity-boost: .1);
```
kfranqueiro: Nit: why did you include the variable names here, when you're specifying all of them, in order, anyway?

acdvorak: I could probably remove the `$color` name since it's pretty obvious what `secondary-dark` is. But for `$opacity-boost`, I think it's more readable to have the param name; without it, `.1` is just a magical number that means who-knows-what, and you have to go read the documentation or source code to figure out what it does.

acdvorak: Removed `$color`!

PR #1450, opened 2017-10-18
chore(theme): Add mdc-theme-prop-value function

acdvorak:
This will unfortunately make search/replace of the `mdc-theme-prop` mixin more difficult, because the mixin name is now a prefix of the function name, but `mdc-theme-prop-value` feels like the most natural name for this function.


packages/mdc-theme/_variables.scss
```diff
@@ -140,3 +140,22 @@ $mdc-theme-property-values: (
...
+//
+// This is mainly useful in situations where `mdc-theme-prop` cannot be used directly (e.g., `box-shadow`).
+//
+// Examples:
+//
+// 1. mdc-theme-prop-value(primary) => "#3f51b5"
+// 2. mdc-theme-prop-value(blue) => "blue"
+//
+// NOTE: This function must be defined in _variables.scss instead of _functions.scss to avoid circular imports.
+@function mdc-theme-prop-value($property) {
```
kfranqueiro: Should we make this mirror the functionality of mdc-theme-prop where it checks if it's a color and treats that as a passthrough, otherwise it checks the map and will actively cause an error if the key doesn't exist? I'm a bit concerned this could lead to silent failures as written.

Also, I guess there's no reasonable way for us to factor CSS custom properties into this function; maybe we should also note that as a difference between this and the mixin?

acdvorak: Done. Good call!

PR #1452, opened 2017-10-19
feat(ripple): Add new simpler/smaller mixins and remove unused CSS vars

### LINKED ISSUE #1433: Add new ripple mixins (and docs) to mdc-ripple, without removing old ones
- [ ] mdc-ripple-surface
- [ ] mdc-ripple-color($color, $opacity)
- [ ] mdc-ripple-radius($radius)

kfranqueiro:
Currently the existing mixins are left intact; I'll remove them after updating all other packages to use the new ones.

Resolves #1433


packages/mdc-ripple/mdc-ripple.scss
```diff
@@ -17,13 +17,14 @@
...
+@import "./variables";
 
 // postcss-bem-linter: define ripple-surface
 
 .mdc-ripple-surface {
-  @include mdc-ripple-base;
-  @include mdc-ripple-bg((pseudo: "::before"));
-  @include mdc-ripple-fg((pseudo: "::after"));
+  @include mdc-ripple-surface;
+  @include mdc-ripple-color;
```
lynnjepsen: So you have to include mdc-ripple-color and mdc-ripple-radius even if your color is default and your radius is default? (Actually, is there a default radius?)

packages/mdc-ripple/_mixins.scss
```diff
@@ -241,6 +243,142 @@
...
+    // Assume $color is a theme property name
+    $theme-value: map-get($mdc-theme-property-values, $color);
+    $css-var: var(--mdc-theme-#{$color}, $theme-value);
+
+    background-color: rgba($theme-value, $opacity);
+
+    // See: https://drafts.csswg.org/css-color/#modifying-colors
+    // While this is currently unsupported as of now, it will begin to work by default as browsers
+    // begin to implement the CSS 4 color spec.
+    @supports (background-color: color(green a(10%))) {
```
lynnjepsen: mmmm...this might be outside the scope of this PR...but could we just remove the @supports? It's odd we use this technology before it is supported across all our supported browsers

packages/mdc-ripple/_mixins.scss
```diff
@@ -241,6 +243,142 @@
...
+    opacity: 0;
+  }
+
+  // Note: when :active is applied, :focus is already applied, which will effectively double the effect.
+  &:not(.mdc-ripple-upgraded) {
+    &:hover::before,
+    &:focus::before,
+    &:active::after {
+      transition-duration: 85ms;
+      opacity: .6;
```
lynnjepsen: should this reference the $opacity value? or no?

packages/mdc-ripple/_mixins.scss
```diff
@@ -241,6 +243,142 @@
...
+    }
+  }
+
+  &.mdc-ripple-upgraded--background-focused::before {
```
lynnjepsen: why included this code in the `mdc-ripple-color` mixin? It kinda seems like its a hardcoded aspect of MDC Ripple Surface. Could you move it to `mdc-ripple-surface`?

Same question for the stuff below.

packages/mdc-ripple/_mixins.scss
```diff
@@ -241,6 +243,142 @@
...
+    top: var(--mdc-ripple-top, calc(50% - #{$radius / 2}));
+    left: var(--mdc-ripple-left, calc(50% - #{$radius / 2}));
+    width: var(--mdc-ripple-fg-size, $radius);
+    height: var(--mdc-ripple-fg-size, $radius);
+    transform: scale(var(--mdc-ripple-fg-scale, 0));
+  }
+
+  // Foreground ripple styles
+  &.mdc-ripple-upgraded::after {
+    top: 0;
```
lynnjepsen: if top and left is hardcoded to 0, should this be moved to the `mdc-ripple-surface` mixin?

packages/mdc-ripple/_mixins.scss
```diff
@@ -241,6 +243,142 @@
...
+  }
+
+  &.mdc-ripple-upgraded--foreground-deactivation::after {
+    animation: 83ms mdc-ripple-fg-opacity-out;
+    // Retain transform from mdc-ripple-fg-radius-in activation
+    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
+  }
+}
+
+@mixin mdc-ripple-radius($radius: 100%) {
```
lynnjepsen: Gah I really like this mixin. It makes it so much more obvious all the element we need to keep in sync when we change the ripple radius.

packages/mdc-ripple/_mixins.scss
kfranqueiro: I definitely had thoughts of eliminating this, but I also thought it might make sense for me to evaluate this later with work on states, because I'm wondering if there's even a convincing reason for us to be using RGBA colors rather than relying more directly on the opacity property, which would also make this completely moot.

packages/mdc-ripple/_mixins.scss
kfranqueiro: I planned to look at this later with work on states, since I'll need to validate all of the opacity values then. It's not yet clear to me why this value is currently what it is for the CSS-only ripple.

packages/mdc-ripple/_mixins.scss
kfranqueiro: This was already part of the existing -bg and -fg mixins, and after the hitches that Yiran ran into needing to redundantly @include mdc-ripple-base due to opacity styles there, I wanted to avoid any more issues like that.

It is possible you're right, but I think some of this deserves a later sweep after we're sure that states behavior works as we want it.

packages/mdc-ripple/_mixins.scss
kfranqueiro: Yeah... it also made it clear to me that bounded/unbounded-specific styles aren't that huge, which is why I left that alone for now, but we can think about splitting that later.

packages/mdc-ripple/_mixins.scss
kfranqueiro: That might work, but it made more sense to me to put it here, juxtaposed against the unbounded styles below. If we were to split out bounded vs. unbounded styles, I think I'd want to keep these together.

There are definitely some things I'd like to ultimately see if I can move either into the surface mixin, or out of mixins entirely and just style once for all upgraded ripples.

packages/mdc-ripple/mdc-ripple.scss
kfranqueiro: Yeah, just like you had to always include -fg and -bg before.

And yes, there absolutely is a default radius (100%) - in fact, we _never_ customized this in our code prior to my realizing that checkbox and radio should have been doing so (they were manually overriding the exact styles involved rather than changing $radius in the ripple config).

PR #1453, opened 2017-10-19
fix(snackbar): Incorrect width using border-box

### LINKED ISSUE #1440: Snackbar incorrect rendering with border-box

0.23
(as on https://material-components-web.appspot.com, but I'm not sure since 0.23 was just released)

Chrome 62

`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.62 Safari/537.36`

### What OS are you using?

- Windows 10 RS3  10.0.16299.19
- Android 7.1.2

1. Open https://material-components-web.appspot.com/snackbar.html
2. In the inspector, switch the device size emulator to a phone, e.g. iPhone 6. Or simply hook up an Android phone in the remote inspector.
3. In the inspector, add the following custom css:
```css
* {
    box-sizing: border-box;
}
```
3. Click "SHOW" to show a snackbar.

A snackbar shows up, filling the whole width of the screen.
(I have changed to the dark theme to make it more clear)

material-components-web appspot com_snackbar html iphone 6 3

A snackbar shows up, leaving its right side a margin.

material-components-web appspot com_snackbar html iphone 6 2

It seems the Snackbar component ( as well as some other components) are designed with `box-sizing: content-box;` in mind. 
However, in the case when `box-sizing: border-box;` is set globally ( which happens very often ), the box-sizing of the Snackbar gets overwritten easily.
Adding this css seems to address this problem, but can it be done more elegantly?
```css
.mdc-snackbar {
  box-sizing: content-box;
}
```

RobertSmits:
Resolves #1440


kfranqueiro:
I think this also actually makes the snackbar adhere more accurately to the spec RE min/max width as well. Thanks!

PR #1457, opened 2017-10-20
fix(textfield): unwanted background color

Shyam-Chen:
Before:

After:


kfranqueiro:
The label's background color style is necessary in order to ensure that text doesn't clobber each other after you enter a value in the field.

Here's what it looks like now:

image

Here's what happens without that style:

image

So right now you should actually change both of those background colors together. Eventually we'll refactor textfield's styles into Sass mixins that will make this easier.

PR #1465, opened 2017-10-23
docs(dialog): Update README wording RE scrollable dialogs

### LINKED ISSUE #1442: Dialog body style guide (scrollable)
What is the reason for having a scrollable and a not scrollable Dialog body? And then the Dialog body height is set to a very small 195px on scrollable body and has overflow: scroll so the scrollbar is always visible, even if the content does not use all the space.
Would it not make more sense to make it dynamic? e.g. Dialog body is filling the screen until a certain % and after that, the body will scroll. 
Also liked that the dialog of MDL got to fullscreen on mobile (just don't forget to add a close / cancel button :-) )

kfranqueiro:
Resolves #1442 by clarifying docs.


bwobrien:
LGTM

kfranqueiro:
Pinging @googlebot to attempt to get CLA check to go through...

PR #1479, opened 2017-10-24
feat(theme): rename all color_palette instances

bonniezhou:
Rename to `color-palette` for consistency with all other SCSS files


bonniezhou:
BREAKING CHANGE: `_color_palette.scss` has been renamed to `_color-palette.scss` in mdc-theme

bonniezhou:
Added @google.com email (with CLA) for commits

PR #1481, opened 2017-10-25
feat(theme): Allow overriding of text themes

gline9:
When theme colors are very close to the border of light vs dark theme, the chosen text color can be different that what you want. Currently there is no way to change which text theme was chosen for a particular color. This can be accomplished by adding the !default tag to the chosen theme and allowing the variable to be overridden similarly to how the theme colors can be.


gline9:
I signed it!

PR #1485, opened 2017-10-25
feat(text-field): rename textfield to text-field

bonniezhou:
Fixes https://github.com/material-components/material-components-web/issues/1445


demos/text-field.html
```diff
@@ -136,28 +136,28 @@
...
         </div>
-        <p class="mdc-textfield-helptext mdc-textfield-helptext--persistent mdc-textfield-helptext--validation-msg"
+        <p class="mdc-text-field-helptext mdc-text-field-helptext--persistent mdc-text-field-helptext--validation-msg"
            id="pw-validation-msg">
           Must be at least 8 characters long
         </p>
       </section>
       <section class="example">
-        <h2>Textfield box</h2>
+        <h2>TextField box</h2>
```
kfranqueiro: Should we be referring to this as "Text field" in docs, to match the [guidelines](https://material.io/guidelines/components/text-fields.html)?

packages/mdc-textfield/adapter.js
```diff
@@ -28,16 +28,16 @@
 let NativeInputType;
 
 /**
- * Adapter for MDC Textfield.
+ * Adapter for MDC TextField.
```
kfranqueiro: If other adapters are any indication, we should use "Text Field" here and L34 (like "Form Field" and "Simple Menu" in their respective adapters).

packages/mdc-textfield/foundation.js
```diff
@@ -127,7 +127,7 @@ class MDCTextfieldFoundation extends MDCFoundation {
   }
 
   /**
-   * Handles all user interactions with the Textfield.
+   * Handles all user interactions with the TextField.
```
kfranqueiro: "Text Field" instead of "TextField" in method comments in this file too.

packages/mdc-textfield/index.js
```diff
@@ -85,39 +85,39 @@ class MDCTextfield extends MDCComponent {
   }
 
   /**
-   * Initiliazes the Textfield's internal state based on the environment's
+   * Initiliazes the TextField's internal state based on the environment's
```
kfranqueiro: "Text Field" in these comments as well.

While we're here, there's a typo (Initializes)

packages/mdc-textfield/package.json
```diff
@@ -6,7 +6,7 @@
   "keywords": [
     "material components",
     "material design",
-    "textfield"
+    "text field"
```
kfranqueiro: Nice catch 👍

acdvorak: We probably want both "textfield" and "text field"

demos/text-field.html
```diff
@@ -136,28 +136,28 @@
...
         </div>
-        <p class="mdc-textfield-helptext mdc-textfield-helptext--persistent mdc-textfield-helptext--validation-msg"
+        <p class="mdc-text-field-helptext mdc-text-field-helptext--persistent mdc-text-field-helptext--validation-msg"
            id="pw-validation-msg">
           Must be at least 8 characters long
         </p>
       </section>
       <section class="example">
-        <h2>Textfield box</h2>
+        <h2>Text field box</h2>
```
kfranqueiro: Nit: Capital F in Field, since all the other headings on this page do that

packages/mdc-textfield/constants.js
```diff
@@ -1,5 +1,4 @@
 /**
- * @license
```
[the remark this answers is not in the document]
bonniezhou: Oops, bad merge?

j-o-d-o:
I guess this change is for consistency (e.g. form-field). But shouldn't the folder then also be renamed to mdc-text-field?

bonniezhou:
The package name can't be changed because if it does, then anyone who lists it as a dependency won't receive updates for this package any more.

j-o-d-o:
You could mark it as deprecated:

`>> npm deprecate @material/textfield@"< latest-version" "This package was renamed to @material/text-field"`

PR #1487, opened 2017-10-26
fix(button): Button Size Is Incorrect

### LINKED ISSUE #1483: Dialog Button Size Incorrect
**What MDC-Web Version are you using?**
Online dialog example: https://material-components-web.appspot.com/dialog.html 

**What browser(s) is this bug affecting?**
All

**What are the steps to reproduce the bug?**
1. Review link: https://material-components-web.appspot.com/dialog.html 
2. Inspect dialog buttons
3. .mdc-button is using min-width: 88px;

**What is the expected behavior?**
Dialog buttons have a min-width: 64px; See Material Design documentation for "Button width and padding" https://material.io/guidelines/components/dialogs.html#dialogs-specs 

screen shot 2017-10-25 at 8 19 49 am

touficbatache:
Fixes and closes #1483

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -15,7 +15,7 @@ feature requests.
 If you'd like to work on a component, please ensure that you *submit an Engineering Outline before
 submitting a pull request*. You can read more about this in our [contributing docs](https://github.com/material-components/material-components-web/blob/master/CONTRIBUTING.md#building-components).
 
-MDC-Web is still under active development. You can see our current progress on [master](https://github.com/material-components/material-components-web/tree/master) as well as an overview of the direction we're headed in our [developer guide](https://github.com/material-components/material-components-web/blob/master/docs/DEVELOPER.md).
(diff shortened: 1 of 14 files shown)


touficbatache:
Yes, I saw that error on the guidelines page but the diagram is for the flat button. Should I change the `min-width` of all buttons in general or only the flat button's one?

touficbatache:
Changed the title as the fix is for the button in general and not only for the dialog one.

MeyerUX:
@kfranqueiro what's the process for the design team updating documentation? Do they do it on the fly, or during a release on: https://material.io/guidelines/material-design/whats-new.html ?

touficbatache:
Well, anyway they _should_ do it on the fly...

touficbatache:
Well, anyway they _should_ do it on the fly...

MeyerUX:
@touficbatache agreed, but also making sure we have a good way to track documentation updates. Which right now it looks like the "what's new" is only tracking major releases. This button change would be a big nice-to-know for the guideline side.

PR #1508, opened 2017-10-30
chore: Update eslint to 4.10 and fix lint error

kfranqueiro:
h/t @touficbatache for flagging this in #1487

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3454,20 +3454,20 @@
       }
     },
     "eslint": {
-      "version": "4.3.0",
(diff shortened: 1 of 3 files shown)


touficbatache:
Hahaha, that's why it was complaining! I was searching for code error but couldn't find any! 😂

kfranqueiro:
I'm pretty sure the lint error was caused by fixes in eslint 4.10, which was only released 3 days ago.

(This is why package-lock is a thing, but Travis is running with npm 4 at the moment, which doesn't know that package-lock exists...)

touficbatache:
Oh okay, thanks for the explanation.

PR #1511, opened 2017-10-31
Update Readme.md

Chinoms:
I think it should be "Material Components for the web (MDC-Web) helps" and not ". . . help". This is because Material Components for the web (MDC-Web) is the name of a collection and therefore, is a collective singular noun even though it appears to be plural.  As an example, you wouldn't say "The Chinese Association of Monks help people in their community." You'd rather say "The Chinese Association of Monks helps people in their community."

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Material Components for the web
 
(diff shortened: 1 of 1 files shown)


Chinoms:
Hello.
I've signed it.

PR #1514, opened 2017-10-31
docs: Fix grammar for mdc-toolbar modifiers

nickmccurdy:
There are some minor conjugation and consistency issues in the mdc-toolbar modifier docs.


nickmccurdy:
It seems like TODO comments in other parts of the docs (which I did not change) are failing the build, can we ignore that or fix it separately?

kfranqueiro:
It looks like what failed was a test timeout, so don't worry about that. The TODO comment warnings alone won't cause CI to fail.

PR #1522, opened 2017-11-01
fix(checkbox): JavaScript console warning in IE

### LINKED ISSUE #1504: MDCCheckbox, javascript console warning in IE

0.23.0

Internet Explorer 11, Edge

### What OS are you using?

Windows 10

Open the page in IE: https://material-components-web.appspot.com/checkbox.html

There will be multiple errors there that read:
```
HTML1500: Tag cannot be self-closing. Use an explicit closing tag.
```

The errors are caused by this element:
```
<path class="mdc-checkbox__checkmark__path"
      fill="none"
      stroke="white"
```

touficbatache:
Fixes and closes #1504

### DIFF
--- a/demos/checkbox.html
+++ b/demos/checkbox.html
@@ -77,7 +77,7 @@
               <path class="mdc-checkbox__checkmark__path"
                     fill="none"
                     stroke="white"
-                    d="M1.73,12.91 8.1,19.28 22.79,4.59"/>
+                    d="M1.73,12.91 8.1,19.28 22.79,4.59"></path>
             </svg>
             <div class="mdc-checkbox__mixedmark"></div>
           </div>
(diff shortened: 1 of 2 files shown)


lynnjepsen:
Hey @touficbatache , I had to revert this after discussing with my team more. Totally my fault, I gave poor guidance on the bug. The self-closing <path> element is valid HTML, and the <path></path> is actually NOT valid (despite IE complaining).

Theres some more info here: https://stackoverflow.com/questions/24299969/closing-svg-tags-explicit-or-self-closing/31120364#31120364

Could you still help us out though? We need to add a caveat to the README for MDC Checkbox. It should explain to our users that <path></path> is necessary on IE to avoid the console warnings.

touficbatache:
@lynnjepsen Yes sure!

PR #1532, opened 2017-11-02
chore: Remove framework-examples/react now that we have RMWC as a recommended library

lynnjepsen:


mateuszdraco:
I might be good to update https://material.io/components/web/docs/framework-integration/ as well. Currently the docs point to missing github page and the only reference to e.g. rmwc is in the commits.

PR #1533, opened 2017-11-02
fix(checkbox): JavaScript console warning in IE

### LINKED ISSUE #1504: MDCCheckbox, javascript console warning in IE

touficbatache:
Fixes and closes #1504. @lynnjepsen I made the changes you asked for in #1522 :smiley:


packages/mdc-checkbox/README.md
```diff
@@ -89,6 +89,31 @@ easily position checkboxes and their labels.
 </div>
 ```
 
+Please note that if you are using IE, you have to close the `<path>` tag to avoid console warnings.
+
+```html
```
kfranqueiro: Nit: it's kind of confusing to read this at a glance, seeing two very similar large code blocks together with one very important sentence sandwiched in the middle.

I'd suggest to just make a note instead after the existing code block and leave out the 2nd code block:

```
> **Note**: If you are using IE, you need to include a closing `</path>` tag if you wish to avoid console warnings.
```

cc @lynnjepsen WDYT?

lynnjepsen: SGTM

PR #1538, opened 2017-11-03
fix(button): Stroked buttons should change the padding of the button

lynnjepsen:


packages/mdc-button/_mixins.scss
```diff
@@ -64,12 +64,20 @@
   // Note: line height is adjusted for stroke button because borders are not
   // considered as space available to text on the Web
   line-height: $mdc-button-height - $stroke-width * 2;
+  $padding-value: max($mdc-button-horizontal-padding - $stroke-width, 0);
```
kfranqueiro: Might want a comment above this too, e.g. "Adjust padding to maintain consistent width with non-stroked buttons"

:+1: to centralizing this though; this actually makes the demo page line up nicer, in addition to allowing for stroke width changes across states for a single button without affecting its width.

packages/mdc-button/_variables.scss
```diff
@@ -15,7 +15,9 @@
 //
 
 $mdc-button-height: 36px;
+$mdc-button-horizontal-padding: 16px;
 $mdc-dense-button-height: 32px;
+$mdc-compact-button-horizontal-padding: 16px;
```
kfranqueiro: This was `8px`, I'm assuming the change was accidental :)

lynnjepsen: Ooops, fixed

PR #1543, opened 2017-11-03
feat(textfield): Convert some foundation methods from private to public

bwobrien:
This is to allow integration into other web frameworks.

### DIFF
--- a/packages/mdc-textfield/README.md
+++ b/packages/mdc-textfield/README.md
@@ -447,7 +447,7 @@ Deactivates the focus state of the Text Field. Normally called in response to th
 Animates the bottom line. The animation expands outward from the user's click or touch. Expects an
 event with clientX/clientY properties.
 
-##### MDCTextfieldFoundation.handleBottomLineAnimationEnd
+##### MDCTextfieldFoundation.handleBottomLineAnimationEnd(evt: !Event)
 
 Handles the end of the bottom line animation, performing actions that must wait for animations to
 finish. Expects a transition-end event.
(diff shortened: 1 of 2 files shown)


packages/mdc-textfield/foundation.js
```diff
@@ -202,9 +198,8 @@ class MDCTextfieldFoundation extends MDCFoundation {
    * Fires when animation transition ends, performing actions that must wait
    * for animations to finish.
    * @param {!Event} evt
-   * @private
    */
-  transitionEnd_(evt) {
+  transitionEnd(evt) {
```
kfranqueiro: I'm kind of surprised that we'd be publicizing this method in particular... would there be a better name for this that would give better context to when it'd make sense to directly invoke it?

bwobrien: Gave this and setBottomLineTransformOrigin more semantic names.

packages/mdc-textfield/README.md
```diff
@@ -426,6 +426,32 @@ Returns a boolean specifying whether or not the input is disabled.
 
 Updates the input's disabled state.
 
+##### MDCTextfieldFoundation.setValid(isValid: boolean)
+
+Sets the validity state of the Text Field. Triggers custom validity checking.
+
+##### MDCTextfieldFoundation.handleTextFieldInteraction(evt: !Event)
```
kfranqueiro: Nit: FWIW, I'm pretty sure we follow typescript conventions when documenting APIs in READMEs, so the `!`s here might not make sense...

bwobrien: Done.

PR #1544, opened 2017-11-04
feat(slider): Add color theme mixins; default to secondary; remove `--off`

### LINKED ISSUE #1151: Add Sass color mixin to slider
- [x] Create color mixins
- [x] Delete any —primary --accent modifiers

acdvorak:
BREAKING CHANGE: The `mdc-slider--off` modifier class has been removed as it is being removed from the spec.

Resolves #1151


demos/slider.scss
```diff
@@ -17,3 +17,14 @@
 @import "./common";
 @import "../packages/mdc-slider/mdc-slider";
 @import "../packages/mdc-typography/mdc-typography";
+@import "../packages/mdc-theme/color-palette";
+
+.demo-slider--custom {
+  $main-color: $material-color-red-700;
+  $tick-color: $material-color-red-300;
```
kfranqueiro: Is this supposed to be used?

packages/mdc-theme/_functions.scss
```diff
@@ -40,21 +40,35 @@ $_mdc-theme-tonal-offset: 7%;
...
+// Determine whether to use dark or light text on top of given color to meet accessibility standards for contrast.
+// Returns "dark" if the given color is light and "light" if the given color is dark.
+@function mdc-theme-contrast-tone($color) {
+  @return if(mdc-theme-tone($color) == "dark", "light", "dark");
+}
+
+// DEPRECATED. Use mdc-theme-contrast-tone instead.
+@function mdc-theme-light-or-dark($color) {
+  // stylelint-disable indentation
+  @warn "The 'mdc-theme-light-or-dark' mixin is DEPRECATED and will be REMOVED in a future version. " +
```
kfranqueiro: I assume we should merge #1546 first, assuming that the boatload of warnings this produces are resolved by the replacements performed in that PR...

demos/slider.scss
```diff
@@ -17,3 +17,14 @@
...
+
+  @include mdc-slider-rail-color($main-color);
```
kfranqueiro: Given that it seems to be ridiculously common to set all of these to the same color, should we have a mixin that sets all of them? Maybe takes a color + an ink color for the popup indicator?

demos/slider.scss
acdvorak: Good catch! Removed. I originally had a separate mixin for tick mark color, but then I removed it because I wasn't sure if it was supposed to be customizable. I figured we can always add that mixin back if necessary.

packages/mdc-theme/_functions.scss
acdvorak: Done!

demos/slider.scss
acdvorak: Done!

I decided to keep it simple and pick the ink color automatically, just like `mdc-slider-value-pin-fill-color-accessible`.

The new combined mixin is `mdc-slider-color-accessible`.

acdvorak:
FYI I updated the colors of the custom slider demo. PTAL!

packages/mdc-slider/mdc-slider.scss
```diff
@@ -17,29 +17,33 @@
...
-$mdc-slider-dark-theme-grey: #5c5c5c;
-
 // postcss-bem-linter: define slider
 
-@keyframes mdc-slider-emphasize {
-  0% {
-    animation-timing-function: ease-out;
-  }
+.mdc-slider {
+  @include mdc-slider-rail-color(secondary);
```
kfranqueiro: Can this technically use the new higher-level `accessible` mixin? (Since IIUC `text-primary-on-secondary` should be the same value that the mixin computes when passed `secondary`?)

packages/mdc-slider/_mixins.scss
```diff
@@ -0,0 +1,105 @@
...
+  @include mdc-slider-thumb-color($color);
+  @include mdc-slider-focus-halo-color($color);
+  @include mdc-slider-value-pin-fill-color-accessible($color);
+}
+
+//
+// Private
+//
+
+@mixin mdc-slider-rail-color-impl_($color, $important: false) {
```
kfranqueiro: Nit (from discussion in chat): Omit `-impl` in these private mixins (just make them the public mixin name plus underscore suffix)

packages/mdc-slider/_mixins.scss
```diff
@@ -0,0 +1,105 @@
...
+@mixin mdc-slider-thumb-color-impl_($color, $important: false) {
+  .mdc-slider__thumb {
+    @include mdc-theme-prop(fill, $color, $important: $important);
+    @include mdc-theme-prop(stroke, $color, $important: $important);
+  }
+}
+
+@mixin mdc-slider-thumb-stroke-cutout_($color) {
+  // We use !important here to override all styles without having to use excessive
+  // :not() selectors or intricate compound selectors.
```
lynnjepsen: What is "excessive :not() selectors"? How many :not selectors would you need to add?

acdvorak:
> [Question](#pullrequestreview-74778782): Do you think we should remove either the Dark Theme or Custom BG Color checkbox from the demo page since both pretty much do the same thing now?

Dark Theme and Custom BG are actually different enough that I think it makes sense to keep both. The Dark Theme checkbox shows that we provide a default value for `--mdc-slider-bg-color-behind-component` in dark mode, whereas the Custom BG checkbox shows how to override that default value.

packages/mdc-slider/mdc-slider.scss
acdvorak: We certainly could, but IIRC @lynnjepsen wanted our baseline component styles to explicitly call every individual theme mixin so that users know what all the options are. I'll verify this with her offline.

packages/mdc-slider/_mixins.scss
acdvorak: Done!

packages/mdc-slider/_mixins.scss
acdvorak: I actually just copied that comment from the original Sass code. At the time, I hadn't removed the `--off` modifier yet, so on the surface it seemed plausible that the comment was still accurate.

However, I just tried using `:not()` selectors, and it was actually super easy, so I updated the PR to do that instead of using `!important` at all.

Good catch!

packages/mdc-slider/mdc-slider.scss
acdvorak: I was wrong 😄   Lynn prefers your approach of using the high-level `accessible` mixin. Done!

acdvorak:
Added mixins to allow the user to customize rail color/opacity and tick mark color separately from highlight color. This was in response to feedback from design that those elements should be customizable.

PTAL!

PR #1545, opened 2017-11-04
fix(slider): Don't hide focus ring on discrete sliders

### LINKED ISSUE #1427: discrete mdc-slider doesn't have a visual focus indication

0.22.2

Tested on chrome

### What OS are you using?

Doesn't matter

1. Browse to https://material-components-web.appspot.com/slider.html
2. Use the tab key to change the focus to the different inputs on the page

All sliders should have a visual indication when they receive focus. (See the guidelines https://material.io/guidelines/components/sliders.html ).

The discrete sliders do not have a visual indication when they receive focus. This doesn't conform with the guidelines on https://material.io/guidelines/components/sliders.html

This is caused by the styling overrides for .mdc-slider--discrete.mdc-slider--focus .mdc-slider__focus-ring in the .scss files. Please note that for a discrete slider with markers, the marker should be shown on focus (according to the guidelines linked above), for that there doesn't seem to be any styling at all.

acdvorak:
The old behavior is not in the M1 or M2 specs. Even if it was, it's terrible for accessibility and is inconsistent with the continuous slider behavior.

Resolves #1427

### DIFF
--- a/packages/mdc-slider/mdc-slider.scss
+++ b/packages/mdc-slider/mdc-slider.scss
@@ -324,11 +324,6 @@ $mdc-slider-dark-theme-grey: #5c5c5c;
     .mdc-slider__thumb {
       animation: none;
     }
-
-    .mdc-slider__focus-ring {
-      transform: none;
-      opacity: 0;
-    }
   }
 
   &.mdc-slider--display-markers {


kfranqueiro:
Seems like this should resolve #1427?

acdvorak:
Great catch on #1427! And good question about showing/hiding the popup on focus/blur - I added that to my list of questions for the next design office hours. Thanks Ken!

Issue #1521, opened 2017-11-01
framework integration criteria (vuejs)


stasson:
Hello MDC Team

I have been working on this vuejs integration project: [vue-mdc-adapter](https://stasson.github.io/vue-mdc-adapter) for the last 6 month. I must say I have been very impressed by the work you guys do. I think my project has reached a point where it may be worth sharing. 

vue-mdc-adapter is using the foundation aproach. Goal is to use Vue's reactivity system wherever It make sense, and keep the markup/apis as simple and as close as possible to MDC. For now I have been mostly focused on figuring out the right vue patterns and get a first working implementation of each component. I guess next comes api refining and a big validation phase. if the project takes off, I'll have to take care of the perf as well.  

There are a couple of integration tricks and issues I found while integrating. I guess I'll gather my notes and share it with you. Are github issues the preferred way to send you some feedback ?

Also today I'm wondering If I should do a PR and add my library to your list of integrations: I definitely need some feedback on the api's  and it may help me find people who'd contribute... I went through the [criteria](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md) and I'm not sure about the following statement: 
> - Serve components in an à-la-carte delivery model  

As of today I have only one package. people can cherry pick es6 components using webpack/rollup, but I didn't create one npm package per component. Do I need to do so to meet the criteria ? 

thxs for your feedback.

lynnjepsen:
Go ahead and start a PR, and you'll get some Vue experts from our community to comment on your PR. We can work through the "a la carte" part with the Vue experts as well.

anlexN:
i don’t want to use any framework integration or wrapper!
[How to use material-components-web?](https://forum.vuejs.org/t/how-to-use-material-components-web/78469)

PR #1552, opened 2017-11-06
docs: Update README adding vue-mdc-adapter to frameworks integrations

stasson:


stasson:
Following up with #1521 (I have added per component distribution in latest version)

stasson:
I signed it!

stasson:
Hi MDC Team, 

Is there anything else I need to do or change before review ?

Thxs

lynnjepsen:
Hey @stasson , Thanks for the PR!

Lets take a look at [criteria](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md):

> Have existed for longer than 6 weeks and show continued maintenance over time

Looks like Material Components for Vue has existed for more than 6 weeks. Do you plan on continuing maintenance?

> Provide usage documentation per component

Looks like you have documentation per component. Great!

> Serve components in an à-la-carte delivery model

Honestly, I have no idea what the Vue system is for downloading and install libraries. Have you thought about how other Vue users will download your library? What do they do if they only want one component?

I was hoping some members of our community might jump on this PR if they had opinions on Vue, but we don't seem to have any comments yet. So in the meantime, could you compare your library to [Material Components Vue](https://github.com/matsp/material-components-vue) for me? @matsp submitted PR #1575, so I'm curious if there are any big differences between your two approaches.

stasson:
Hi @lynnjepsen 

> Do you plan on continuing maintenance?  

Yes.
 
> Serve components in an à-la-carte delivery model

- Vue components do work exactly like react components:  you import/require your components and then can mount them in the dom or agregeate them in other components:

```javascript
import {VueMDCButton} from 'vue-mdc-adapter/components/button'
// then do something with the component
```

- Vue also has the notion of [plugin](https://vuejs.org/v2/guide/plugins.html) which allows to do about the same as the `mdc.autoInit()`

```javascript
import Vue from 'vue'
import VueMDCButtonPlugin from 'vue-mdc-adapter/components/button'

Vue.use(VueMDCButtonPlugin)

// mdc-button is now registered I can use it in the markup 
```
Also when building umd or commonjs vue component library distribution for the CDN, Authors usually add the code that will automatically register the plugin if a Vue context is detected. so  you can also use the components without bundlers:

```html
 <script src="https://unpkg.com/vue/dist/vue.js"></script>
 <script src="https://unpkg.com/vue-mdc-adapter/dist/vue-mdc-button.js"></script>
```

Bottom line:  you can have a look at my [distribution](https://unpkg.com/vue-mdc-adapter/dist/). You will see the main distribution `vue-mdc-adapter.js` and the _a la carte_ distribution for each component `vue_mdc_button.js`, ...  There is a 1-to-1 match between MDC components and my adapters. 

> could you compare your library to Material Components Vue for me?

Here is the Vue/MDC integration landscape as I understand it today:

- [vue-mdc](https://github.com/posva/vue-mdc) from @posva. This is the legacy project inherited from `vue-mdl` which was the most used project for `material-light`.  Unfortunately the project is not really active yet. But I believe this could be one of the big player in the future if the project resumes dev.

- [robertkern/vue-material](https://github.com/robertkern/vue-material) from @robertkern Probably one of the best integration of MDC I've seen,  I have learned a lot about Vue reading his code. not sure where he stands regarding the criteria and the feature set, but probably worth to follow.


Then you have  @matsp 's and my project. 

-  [Material Components Vue](https://github.com/matsp/material-components-vue) 
- [vue-mdc-adapter](https://github.com/stasson/vue-mdc-adapter)

We both have about the same feature set and same goals (giving the users a very clean and simple documented api, follow strictly MDC's integration guideline, and allow customization.) 

The main difference is that @matsp 's project follows the Simple Approach (Wrapping MDC-Web vanilla components) while I'm using the Advanced Approach (Using foundations and adapters.) 

@matsp 's project will probably be more stable because he directly relies on Vanilla components. 
My project will probably leverage the Vue reactive system better and thus be better on the rendering perf because I'm using adapters. Although for now I may be a little bit behind regarding stability and have to work hard to keep up with MDC breaking changes... 

As of today I believe @matsp's and my projects both fit the criteria  and have the right feature set (#number of components implemented.)

matsp:
@stasson: Thanks for positive feedback and friendly coexistence :) - sadly this is becoming increasingly rare between developers. I think I have learned some thinks from your approach as well ;-)

posva:
> This is the legacy project inherited from vue-mdl which was the most used project for material-light. Unfortunately the project is not really active yet. But I believe this could be one of the big player in the future if the project resumes dev.

Thanks for pinging me!
To be clear, vue-mdc is a complete rewrite. It shares only the repo  with the old vue-mdl, there's no legacy there 🙂 
The project is right now non-active, indeed, I have limited time due to personal matters and I'm focusing more on Vue core and Vue + Firestore integrations. I plan to continue the development and more specifically making it easy for people to contribute when something is missing.

Looking forward to merging all the projects together and working together in the future, vue-mdc repo gets some traffic but it's not complete right now 🙁 , so that traffic and visibility is wasted

README.md
```diff
@@ -15,6 +15,7 @@ Material Components for the web is the successor to [Material Design Lite](https
   - [Preact Material Components](https://github.com/prateekbh/preact-material-components)
   - [RMWC: React Material Web Components](https://github.com/jamesmfriedman/rmwc)
   - [Blox Material](https://blox.src.zone/material): Angular Integration Library.
+  - [Vue MDC Adapter](https://github.com/stasson/vue-mdc-adapter): Vue Integration Library.
```
lynnjepsen: Can you add a note about how this uses the Foundation/Adapter approach. This will help clarify the main differences between this Vue library, and matsp's version

lynnjepsen:
Thanks for the reply. I think we should call out a little more about the foundation/adapter approach versus using vanilla components, so I requested a note to the README.

Otherwise this LGTM. 

One more question though....@matsp pointed out that he needs to understand how theming and Sass works. How do you plan on supporting theming, and custom colors, within Vue MDC Adapter. Does Vue and Sass play nicely together?

MishUshakov:
Hello, @posva @stasson @matsp nice to see you guys here in one place. Also hi to @lynnjepsen, thanks for making the library better. I use both material and Vue. My opinion on that topic is that there are many material design components libraries for vue out there, but most of them are not exactly like a material, so they just act like they are and many of them doesn't follow the guidelines or have too many unnecessary and heavy things. Im my personal vue Mr @posva  made a really great effort to Material Design Lite implementation for vue, he also started working on the vue-mdc project, but it looks like it is not being actively developed. I took some time to search npm and GitHub to find material components implementation for Vue. I noticed Mr @stasson made a great job on doing that, thank you very much. But the thing I also noticed, that his amazing effort was unaccessible and hard to find, so I created an issue on his repo and asked him to pr you guys. Mr @matsp has also created great lightweight library for MDC, I really think, that, if we take them two together and create a custom settings on @stasson's library, it will really achieve best results. Mr @matsp showed us css-only approach, which is great for PWA and AMP-friendly websites. If we could implement this approach and make this to a custom variable for vue-mdc-adapter it will be the greatest thing ever. For example Vue.use(VueMDC, {settings: {"cssonly": true}}) will disable javascript. Thank you

@lynnjepsen oh, looks like you are not familiar with Vue, yes Vue has built-in support for sass, less,scss, stylus and more if using Webpack of course.

README.md
stasson: @lynnjepsen 

I have amended the change with the note.

stasson: > regarding Theming: 

I guess the guys integrating for angular and react or vue are all asking themselves the same question: what's the best way to allow customization for one component. Should we just delegate to MDC-Web sass or css variables (easiest/cleanest way) or should we also provide a pattern which rely on component properties (kinda complex...) For now the reasonable answer probably is to just rely on MDC-Web sass.

This is a vast subject which may deserve it's whole discussion thread. 


>  Does Vue and Sass play nicely together?

Yes, Kinda. There is indeed a nice feature which comes with vue + webpack that allows you to embed css or sass code in the component itself, and even the ability to scope your styles to the component. This is nice to style or customize a component quickly, but won't work well with a full sass framework like MDC. 

So I decided to not use this specific Vue feature because
- I do not want to rely on webpack specifics (I'm using rollup for UMD distribution, same way most Vue libraries are built)
- It's easier to keep a separate single sass tree to enable theming while avoiding duplicates and rules ordering  issues...

>  supporting theming, and custom colors

I fully rely on MDC-Web sass theming. I have very few customization on top of MDC-Web sass. I only added couple of layouts to help with toolbar/drawer positioning. I follow MDC's color strategy to the letter. 

`dark-theme` is WIP and will be supported as a component property for the components who have a specific dark-them class. I also plan to provide a mdc-dark-theme component wrapper.

For now I do not use css variables, Although it seems really promising and i want to investigate further and see if we can leverage vue reactive system there...

PR #1567, opened 2017-11-08
chore(textfield): Stop listening to transition end of bottom line

lynnjepsen:
I don't actually notice any visual difference when I remove this code. I think there is already CSS which gracefully handles the change between `mdc-textfield__bottom-line` and `mdc-textfield__bottom-line--active`. 

I think we can just add `mdc-textfield__bottom-line--active` when `mdc-textfield__input` is focused, and remove `--active` when `mdc-textfield__input` is blurred. There's no need to listen to the transitionEnd event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,9 @@ Then simply include the correct files, write some HTML, and call `mdc.autoInit()
   </head>
   <body>
     <h2 class="mdc-typography--display2">Hello, Material Components!</h2>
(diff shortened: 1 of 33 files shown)


lynnjepsen:
Ah I see, I still need to remove the bottom-line active class when the text field is blurred, otherwise subsequent focus events won't show the grow animation. How about if we just remove the active class within `deactivateFocus` method

lynnjepsen:
Ah nope, not intended. Now I understand why this code is there. Thanks for the review.

PR #1568, opened 2017-11-08
fix(textfield): Should not be in both disabled and invalid state

lynnjepsen:
Disabled state means "user cannot interact with this text field". The invalid state means "user has to interact with this text field before we can proceed". It doesn't make sense to be in both states at once.


lynnjepsen:
Ya! Thats what I'm going for.

I think we should only show the invalid state to users _after_ they have interacted with it. The first time you load the page, even required text fields are still not "invalid". So if you take a disabled text field and undisable it, I think it should return to the same default state as when the page first loaded. This gives users another chance to focus on the text field, since it is now undisabled.

I made a diagram:

text field state diagram

Basically the only state a disabled text field can change to, is back to default.

kfranqueiro:
I was wondering if you might've intended this, and yeah I can see the point of doing it that way. Do you think we should check with design as to the intended behavior in this situation?

FWIW, HTML5 validation shows invalid state immediately upon re-enabling... but then it also shows invalid state immediately upon page load, not only after user input, which I agree is bad.

aaronhudon:
@lynnjepsen I'm assuming the changes you are discussing will not preclude the control to show an invalid state if the markup for the control is generated at runtime?  E.g. render an invalid textfield from a server response (e.g. ajax).

lynnjepsen:
Talked to our design team, and they agree this is the safest way to go for the default text field behavior.

@aaronhudon , you can still change a text field to the invalid state dynamically by calling `MDCTextField.valid = false`. See documentation here:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#mdctextfieldvalid

PR #1575, opened 2017-11-10
docs: adding material-components-vue to frameworks

matsp:


matsp:
I signed it!

MishUshakov:
https://github.com/stasson/vue-mdc-adapter is much better

matsp:
@MishUshakov Thanks for your feedback - but keep in mind that the two repositories are not comparable because they are implementing different [approaches](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md#approaches). I see pros and cons with both approaches.

lynnjepsen:
Hey @matsp , Thanks for the PR!

Lets take a look at [criteria](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md):

> Have existed for longer than 6 weeks and show continued maintenance over time

Looks like material-components-vue has existed for more than 6 weeks. Do you plan on continuing maintenance?

> Provide usage documentation per component

Looks like you have documentation per component. Great!

> Serve components in an à-la-carte delivery model

Honestly, I have no idea what the Vue system is for downloading and install libraries. Have you thought about how other Vue users will download your library? What do they do if they only want one component?

I was hoping some members of our community might jump on this PR if they had opinions on Vue, but we don't seem to have any comments yet. So in the meantime, could you compare your library to [Vue MDC Adapter](https://github.com/stasson/vue-mdc-adapter) for me? @stasson submitted PR #1552, so I'm curious if there are any big differences between your two approaches.

matsp:
Hi @lynnjepsen,

thanks for the reaction on this PR.

> Do you plan on continuing maintenance?

Yes. At work we are already using my components which gives me great feedback for bugs or usage improvements for the future.

> Have you thought about how other Vue users will download your library? What do they do if they only want one component?

Users actually have to do the following steps I described in the main README:

#### installation

```shell
npm install --save material-components-vue

yarn add material-components-vue
```

#### import all components
```javascript
import MaterialComponentsVue from 'material-components-vue'

Vue.use(MaterialComponentsVue)
```

#### import specific components
```javascript
import Button from 'material-components-vue/components/button'
import Card from 'material-components-vue/components/card'

Vue.use(Button)
Vue.use(Card)
```

At the moment I don't provide a component distribution á la carte because I have to think about theme customization for each component while maintaing a homogeneously UI. At the moment I don't know how to provide a custom theme for each component without modifying CSS variables where I get to the next problem: Browser compatibility. When I unterstand issue #1604 correct the are some plans in your team. So at the moment you just have to download one package. Afterwards you can import each component separately, which register their custom tags in the vue instance - I think @stasson has given a really good and complete description in his PR #1552.

I had set up a component distribution in the past but I recognized that this has to many limitations at the moment. The only advantage was that you can use the components without any build process like rollup, webpack and so on. In my opinion most users want to build a web app and using more than one component which they want to customize with some theming. I fully support this use case. Later I will implement á la carte distribution when it makes sense.

Again I have to reference #1552 because @stasson does are good comparison of our implementations. The featureset should be the same while implementing the mdc-web components with different approaches. I don't think that there are really performance differences between both approaches. I'm only creating my components for the web to provide a solid and simple solution for building PWAs in the future.

README.md
```diff
@@ -15,6 +15,7 @@ Material Components for the web is the successor to [Material Design Lite](https
   - [Preact Material Components](https://github.com/prateekbh/preact-material-components)
   - [RMWC: React Material Web Components](https://github.com/jamesmfriedman/rmwc)
   - [Blox Material](https://blox.src.zone/material): Angular Integration Library.
+  - [Material Components Vue](https://github.com/matsp/material-components-vue)
```
lynnjepsen: Can you add a note about how this uses the Vanilla components approach. This will help clarify the main differences between this Vue library, and stasson's version

lynnjepsen:
Thanks for the reply. I think we should call out a little more about using vanilla components, so I requested a note to the README.

Otherwise this LGTM.

> When I unterstand issue #1604 correct the are some plans in your team. So at the moment you just have to download one package

So we've begun to realize/suspect that plain CSS is not feature-rich enough to support material design. Sass as a css-preprocessor lets us do a lot more with CSS, which helps us build maintainable components.

I know very little about Vue, but I believe having SOME Sass build chain is going to be critical to the success of most Material libraries. Do you have any ideas on how a Vue developer could use Sass within their Vue project?

README.md
lynnjepsen: Can you add a note about how this uses the Vanilla components approach. This will help clarify the main differences between this Vue library, and stasson's version

matsp:
@lynnjepsen I think I described my concerns too complex ;-) Using SASS with Vue is not a problem at all but since SASS need to be compiled in some kind of build chain you can't use it directly in the browser. Because of this it's not possible to ship ready to use components **without** compiling them before usage. I just wanted to explain why I don't distribute my components separately. When I always need a build chain for e.g. changing the primary color than I have somethink like webpack and a sass compiler I can download a bigger package and just use the components I need.

PR #1577, opened 2017-11-11
feat(icon-toggle): add adapter creator

hvolschenk:
Add static createAdapter method for easier extension


hvolschenk:
I signed it!

PR #1589, opened 2017-11-14
feat(infrastructure): Add build command for static demo assets

acdvorak:
New command: `npm run build:demo`.

Outputs static `.css`, `.js`, and `.html` demo files to the `build/` directory. These files can be uploaded to a static file hosting service like [Firebase Hosting](https://firebase.google.com/docs/hosting/) to create a publicly-visible demo.

Note that this command emits `.css` files, _not_ `.css.js` files. As such, it also rewrites the HTML to include `.css` instead of `.css.js`.


webpack.config.js
```diff
@@ -92,12 +98,58 @@ const createCssLoaderConfig = () =>
...
+  return new PostCompilePlugin(() => {
+    if (!BUILD_STATIC_DEMO_ASSETS || !fs.existsSync(OUT_DIR_ABS)) {
+      return;
+    }
+
+    const demosDirAbs = path.resolve('./demos');
+    const tmpDirAbs = fs.mkdtempSync(path.join(os.tmpdir(), 'mdc-web-demo-output-'));
+
+    const copyOptions = {
+      // eslint-disable-next-line no-unused-vars
```
kfranqueiro: Why does this need to be here and why does `dest` need to be declared below?

webpack.config.js
```diff
@@ -92,12 +98,58 @@ const createCssLoaderConfig = () =>
...
+    };
+
+    fsx.copySync(demosDirAbs, tmpDirAbs, copyOptions);
+    fsx.copySync(OUT_DIR_ABS, path.join(tmpDirAbs, DEMO_ASSET_DIR_REL), copyOptions);
+
+    if (!WRAP_CSS_IN_JS) {
+      glob.sync(path.join(tmpDirAbs, '**/*.html'))
+        .forEach((absPath) => {
+          const oldHtml = fs.readFileSync(absPath, {encoding: 'utf8'});
+          const newHtml = oldHtml.replace(/<script src="([^"]+\.css\.js[^"]*)"><\/script>/ig, (match, oldPath) => {
```
kfranqueiro: Can't this whole `replace` call be reduced to this: 

```js
const newHtml = oldHtml.replace(
  /<script src="([^"]+\.css)\.js"><\/script>/ig,
  '<link rel="stylesheet" href="$1">');
```

* If you don't capture the `\.js` then you don't have to strip it manually
* I'm not sure why you need the `[^"]*` at the end at all; all instances of `.css.js` I see in our HTML files are immediately followed by `"`
* `$1` works within the replacement string to refer to the capture group

webpack.config.js
```diff
@@ -92,12 +98,58 @@ const createCssLoaderConfig = () =>
...
+}
+
+const createStaticBuildPlugin = () => {
+  return new PostCompilePlugin(() => {
+    if (!BUILD_STATIC_DEMO_ASSETS || !fs.existsSync(OUT_DIR_ABS)) {
+      return;
+    }
+
+    const demosDirAbs = path.resolve('./demos');
+    const tmpDirAbs = fs.mkdtempSync(path.join(os.tmpdir(), 'mdc-web-demo-output-'));
```
kfranqueiro: Is there a reason this ends with a dash?

webpack.config.js
```diff
@@ -92,12 +98,58 @@ const createCssLoaderConfig = () =>
...
+          const oldHtml = fs.readFileSync(absPath, {encoding: 'utf8'});
+          const newHtml = oldHtml.replace(/<script src="([^"]+\.css\.js[^"]*)"><\/script>/ig, (match, oldPath) => {
+            const newPath = oldPath.replace('.css.js', '.css');
+            return `<link rel="stylesheet" href="${newPath}">`;
+          });
+          fs.writeFileSync(absPath, newHtml, {encoding: 'utf8'});
+        });
+    }
+
+    fsx.removeSync(OUT_DIR_ABS);
```
kfranqueiro: Might want to comment here or at the top as to the relationship of some of the directories to make it clear why we're deleting this folder then recreating it.

webpack.config.js
```diff
@@ -16,23 +16,29 @@
 
 'use strict';
 
+const fs = require('fs');
```
kfranqueiro: According to fs-extra's documentation, it copies all of fs's methods to itself, so you probably don't need to require this?

webpack.config.js
acdvorak: Yes - [`mkdtempSync`](https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback) appends 6 random characters to the end of the `prefix` string passed in.

webpack.config.js
acdvorak: Removed

webpack.config.js
acdvorak: _**Wow.**_ Derp. Updated. Thanks!

(I originally added the `[^"]*` at the end because I was planning to eventually replace Webpack's stupid CSS hot reloading with a query string param along the lines of
`?timestamp=${Date.now()}`. But now that I think about it, the search/replace doesn't need to worry about the query string, because it would be added dynamically by JS, not in the static HTML...)

webpack.config.js
acdvorak: Done

webpack.config.js
acdvorak: Done

Issue #1619, opened 2017-11-20
Drawer Component detaching event handlers for embedded links


jonhue:
### What MDC-Web Version are you using?

0.24.0

### What browser(s) is this bug affecting?

Chrome (Version 62.0.3202.94 (Official Build) (64-bit))

### What OS are you using?

Windows 10

### What are the steps to reproduce the bug?

-> https://codepen.io/anon/pen/wPyxNG

### What is the expected behavior?

Opening a link which is embedded in the drawer 'remotely' should work as well as it does if it lives outside of it.

### What is the actual behavior?

Links inside the drawer with `data-remote="true"` don't work as expected, whereas a regular link outside of the component works well.

amsheehan:
Though not the focus of this issue, there is another [Github issue](https://github.com/material-components/material-components-web/issues/579) that mentions a library called [Turbolinks](https://github.com/turbolinks/turbolinks) (I'm not a RoR expert, forgive me) that has the same issue.

@jamesmfriedman appears to have a solution. Lets continue the conversation on this issue

jamesmfriedman:
@amsheehan Yeah, racked my brain against this one for a while. Monkey patch to the rescue.

https://github.com/jamesmfriedman/rmwc/blob/master/src/Base/withMDCDrawer.js#L36

kfranqueiro:
This should be resolved now that the `stopPropagation` has been refactored out via 3e173e1ddc3bff5c7c9c2d794eeb36a9354ec98c (based on #1138).

jonhue:
@kfranqueiro Great news!

Issue #579, opened 2017-05-03
Menus in drawer do not close when clicking on the drawer


neclepsio:
### What MDC-Web Version are you using?
0.10

### What browser(s) is this bug affecting?
I use Chrome, but it maybe affects others.

### What OS are you using?
Windows 7.

### What are the steps to reproduce the bug?
Have a menu inside a temporary drawer. 

### What is the expected behavior?
When the menu is open and click outside the menu but in the drawer, the menu should close.

### What is the actual behavior?
The menu stays open.

### Any other information you believe would be useful?
I think the problem may be that both drawer and menu use document click listener to close.

mhuebert:
Probably caused by issue #225 - TemporaryDrawer is calling stopPropagation on all events that propagate upward from child elements.

amsheehan:
This is the fourth issue we've seen regarding `stopPropagation` so this is clearly an issue. We will investigate ways to improve this behavior and update the community with our findings here. Thanks!

whodidthis:
There's also something going on with drawer that makes [turbolinks](https://github.com/turbolinks/turbolinks) not work. 😱 

If you use new MDCPersistentDrawer or MDCTemporarydrawer on your drawer none of the links work via turbolinks anymore.

ju1ius:
Hi,
Using `stopPropagation` also breaks event delegation on drawer child elements.
AFAICT, in this component `stopPropagation` is only used to prevent the drawer from closing when clicking inside the `.mdc-temporary-drawer__drawer` element.
One way to avoid this could be to use a real element for the drawer's backdrop instead of a pseudo-element.

```html
<aside class="mdc-temporary-drawer mdc-theme--dark">
  <div class="mdc-temporary-drawer__backdrop"></div>
  <nav class="mdc-temporary-drawer__drawer">
    ...
  </nav>
</aside>
```

jamesmfriedman:
I found a solution to the issue but I wanted to get feedback before I submit a PR. Basically, React (and other libs) use event delegation and the drawer click handler on this line https://github.com/material-components/material-components-web/blob/65a5936a3e4ec82b09fd620d2053f761536916d1/packages/mdc-drawer/slidable/foundation.js#L57 prevents events from ever reaching the root node. StopProp can be evil.

[...]

PR #1138, opened 2017-08-17
fix(drawer): Change how click events are handled

j-o-d-o:
Removing stopPropagation() for the drawer and adding conditional closing (similar to the Dialog).

Fixing issues #1004 and #644. Was not sure what to do with drawerClickHandler_. I left it in there with an empty function.
I tried to do it similar to the fix for the the dialog click events as the issues are similar.

PR #1621, opened 2017-11-21
Fix for #1619 & #579 & #225

jonhue:
@amsheehan 

Building on @jamesmfriedman 's proposal, I found that this solution is working for me.

Fixing #1619, #579

### DIFF
--- a/packages/mdc-drawer/slidable/foundation.js
+++ b/packages/mdc-drawer/slidable/foundation.js
@@ -55,9 +55,9 @@ export class MDCSlidableDrawerFoundation extends MDCFoundation {
     this.inert_ = false;
 
     this.drawerClickHandler_ = (evt) => {
-      let path = evt.composedPath ? evt.composedPath() : evt.deepPath || evt.path;
-      let el = path[1]
-      let drawerWasClicked = el.tagName == 'NAV' && el.classList.contains('mdc-temporary-drawer__drawer');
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
@kfranqueiro it seems that this PR was based off some work I did in my React lib. In my code, I do a check in the event path to see whether or not I should close it.

@jonhue, take a look.

https://github.com/jamesmfriedman/rmwc/blob/master/src/Base/withMDCDrawer.js#L36

jonhue:
@kfranqueiro 

It should work as intended now.

jonhue:
This should fix #225 as well.

jonhue:
@kfranqueiro 

I'd love it, if this could be merged!

jonhue:
@acdvorak @lynnjepsen 

Why doesn't this get merged?

kfranqueiro:
We've merged 3e173e1ddc3bff5c7c9c2d794eeb36a9354ec98c (based on #1138) which resolves this by completely removing `stopPropagation` from the drawer code.

jonhue:
@kfranqueiro When can I expect the changes to be released?

kfranqueiro:
Our next release should be next Monday (we skipped one release cycle due to the holidays).

jamesmfriedman:
Awesome, I'm looking forward to this one! It also seems like the temporary drawer is suffering from some animation issues on the newest iOS safari, but it could very likely be related to the hacky workaround I had to implement to fix this in RMWC.

PR #1624, opened 2017-11-21
feat(ripple): Add new states mixins

kfranqueiro:
This adds mixins to control the styling of hover, focus, and press states' overlays, both overall and individually. These mixins will ultimately replace `mdc-ripple-color`.

Other packages will be updated to use these mixins in future PRs.


packages/mdc-ripple/_mixins.scss
```diff
@@ -141,21 +199,57 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+  // cases where this mixin is used to override another inherited use of itself,
+  // without needing to re-include mdc-ripple-surface.
+  &::before,
+  &::after {
+    @include mdc-ripple-color_($color, $opacity);
+
+    opacity: 0;
+  }
+
+  // Note: when :active is applied, :focus is already applied, which will effectively double the effect.
```
acdvorak: I noticed that some combined states (e.g., active + focus, active + hover) are a bit darker now. I'm assuming that's intentional?

kfranqueiro: That shouldn't be the case using the existing mixins? I just did a test here and pressing a button on both the latest release and this branch result in the same color.

Using the new mixins, yes, some opacities are going to change, and that's intended.

packages/mdc-ripple/_keyframes.scss
```diff
@@ -41,14 +41,14 @@
     }
 
     to {
-      opacity: 1;
+      opacity: var(--mdc-ripple-fg-opacity);
```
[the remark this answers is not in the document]
kfranqueiro: I had thought about this and figured anything I put in here will be haphazard anyway, but I've gone ahead and put the dark ink pressed opacity as a default.

packages/mdc-ripple/README.md
```diff
@@ -61,15 +61,38 @@ General notes:
...
 --- | ---
 `mdc-ripple-surface` | Adds base styles for a ripple surface
-`mdc-ripple-color($color, $opacity)` | Adds styles for the color and opacity of the ripple effect
 `mdc-ripple-radius($radius)` | Adds styles for the radius of the ripple effect,<br>for both bounded and unbounded ripples
 
+##### Basic States Mixin
+
+Mixin | Description
+--- | ---
+`mdc-states-color($color, $include-focus-within)` | Adds state and ripple styles for the indicated color, deciding opacities based on whether the passed color is light or dark. Optionally includes `::focus-within` selector for focused state of CSS-only components.
```
lynnjepsen: We shouldn't document the internals of `$include-focus-within` as adding `::focus-within` selector. Thats an internal implementation detail. Could you turn this argument into '$css-only-component'?

packages/mdc-ripple/README.md
```diff
@@ -61,15 +61,38 @@ General notes:
...
+
+##### Advanced States Mixins
+
+Mixin | Description
+--- | ---
+`mdc-states-base-color($color)` | Sets up base state styles using the provided color
```
lynnjepsen: I dont think I understand the difference between mdc-states-color and mdc-states-base-color.

packages/mdc-ripple/_mixins.scss
```diff
@@ -53,64 +55,120 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+  // Background wash styles, for both CSS-only and upgraded stateful surfaces
+  &:hover::before {
+    opacity: $opacity;
   }
+}
+
+@mixin mdc-states-focus-opacity($opacity, $include-focus-within: false) {
+  // Focus overrides hover by reusing the ::before pseudo-element.
+  // :focus-within generally works on non-MS browsers and matches when a *child* of the element has focus.
+  // It is useful for cases where a component has a focusable element within the root node, e.g. text field,
```
lynnjepsen: maybe the argument should be `$focusable-element-within-root-element`?

packages/mdc-ripple/_mixins.scss
```diff
@@ -53,64 +55,120 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+    $include-focus-within,
+    "&:not(.mdc-ripple-upgraded):focus::before, &:not(.mdc-ripple-upgraded):focus-within::before",
+    "&:not(.mdc-ripple-upgraded):focus::before"
+  );
 
+  #{$cssOnlyFocusSelector},
   &.mdc-ripple-upgraded--background-focused::before {
-    opacity: .99999;
+    // Note that this duration is only effective on focus, not blur
+    transition-duration: 75ms;
```
lynnjepsen: why do you specify the transition-duration within the mdc-states-focus-opacity mixin? Shouldn't this been in a a more "base" mixin?

packages/mdc-ripple/_mixins.scss
```diff
@@ -53,64 +55,120 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
   }
+}
 
-  &.mdc-ripple-upgraded--foreground-deactivation::after {
-    animation: 83ms mdc-ripple-fg-opacity-out;
-    // Retain transform from mdc-ripple-fg-radius-in activation
-    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
-  }
+// Simple mixin which automatically selects opacity values based on whether the ink color is light or dark.
+@mixin mdc-states-color($color: black, $include-focus-within: false) {
```
lynnjepsen: hmm okay I think I see the difference between base-color and color now....

What if we renamed mdc-states-base-color to just mdc-states-color. And then renamed mdc-states-color to mdc-general-states. I talked to Michelle and she can add the term "general states" to the article to cover hover, focused, and pressed

packages/mdc-ripple/README.md
kfranqueiro: This isn't a blanket thing for all CSS-only components, and these mixins are only invoked once to cover both CSS-only and JS components. The distinction is that `::focus-within` is only needed for CSS-only components because we still handle focus/blur in JS when it is used in order to support all browsers (`::focus-within` isn't supported in IE/Edge).

Your comment later about `$focusable-element-within-root-element` could apply equally here; this is simply forwarding to that parameter when invoking the focus mixin. Maybe `$has-nested-focusable-element`? Coming up with a short name for this is hard...

packages/mdc-ripple/_mixins.scss
kfranqueiro: I did it here because all of the focus styles are here (including the logic for whether to include `:focus-within`, which we'd have to also parameterize elsewhere otherwise), and this needs to be specifically within the selector for focus styles.

packages/mdc-ripple/_mixins.scss
kfranqueiro: Before we decide on that, I should probably point out that I was planning to also include optional flags in this mixin for optionally emitting styles for activated and selected states. Would that term still apply then?

packages/mdc-ripple/README.md
lynnjepsen: $has-nested-focusable-element is better. Its okay it is long.

packages/mdc-ripple/_mixins.scss
lynnjepsen: Andy did some clever things with complicated selectors in MDC Checkbox. Like `mdc-checkbox-marked-background-selector-enabled_` Could we follow that pattern?

packages/mdc-ripple/_mixins.scss
lynnjepsen: We talked offline some. I thing mdc-state-hover-opacity is clear...and mdc-states for the "simple" mixins is also clear.

packages/mdc-ripple/_mixins.scss
kfranqueiro: The selectors Andy managed with that were completely static, though, whereas the selectors used here are dynamic based on the `$has-nested-focusable-element` param. If we were to move these styles somewhere else e.g. `mdc-ripple-surface`, we'd have to add this parameter to that as well, and moreover we'd have to repeat these selectors, which is more verbose than the one extra style rule here. (This transition duration is intended to only apply to focus state; there is already a default transition duration for washes in general in `mdc-ripple-surface`.)

packages/mdc-ripple/mdc-ripple.scss
```diff
@@ -35,11 +35,11 @@
   }
 
   &--primary {
-    @include mdc-states-color(primary);
+    @include mdc-states(primary, true);
```
acdvorak: Can we use a named param for `true`? E.g.:

```scss
@include mdc-states(primary, $has-nested-focusable-element: true);
```

At some point I'd like to go through all of our Sass mixins and use named params for all non-obvious boolean values.

packages/mdc-ripple/mdc-ripple.scss
```diff
@@ -35,11 +35,11 @@
   }
 
   &--primary {
-    @include mdc-ripple-color(primary, $mdc-ripple-pressed-dark-ink-opacity);
+    @include mdc-states(primary, true);
   }
 
   &--accent {
-    @include mdc-ripple-color(secondary, $mdc-ripple-pressed-dark-ink-opacity);
+    @include mdc-states(secondary);
```
acdvorak: Why does `--primary` set `$has-nested-focusable-element: true` but `--accent` does not?

kfranqueiro: Because I'm dumb and forgot to remove something I put in simply to verify that the mixin worked as intended (since the only component that will actually use that parameter is text field). :D Fixed.

packages/mdc-ripple/mdc-ripple.scss
kfranqueiro: Yeah, this was actually just for testing and I forgot to remove it... and I sort of intentionally broke exactly this rule to try to remind myself to remove it... but instead you reminded me!

packages/mdc-ripple/_mixins.scss
```diff
@@ -141,21 +199,57 @@ $mdc-ripple-common-styles-emitted_: false !default;
   }
 
   // Foreground ripple styles
+
```
acdvorak: Is this blank line intentional?

packages/mdc-ripple/_mixins.scss
```diff
@@ -141,21 +199,57 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+    &:hover::before,
+    &:focus::before,
+    &:active::after {
+      transition-duration: 85ms;
+      opacity: .6;
+    }
+  }
+
+  &.mdc-ripple-upgraded--background-focused::before {
+    opacity: .99999;
```
acdvorak: Should we add a comment explaining why we're using `.99999` instead of `1`? Or do we just assume that most front end devs will know it's related to GPU acceleration something something?

packages/mdc-ripple/_mixins.scss
kfranqueiro: It matches the one after the "Background ripple styles" comment.

packages/mdc-ripple/_mixins.scss
kfranqueiro: This code is identical to what existed previously. This is in the legacy mixin that will be going away in a couple of weeks hopefully :)

PR #1631, opened 2017-11-23
chore(text-field): Split out input into subelement

### LINKED ISSUE #1595: Split out input from text field parent
Part of our effort to split text field into sub directories, to reduce merge conflicts and generally improve the readability of text field code.

bonniezhou:
Refactoring input logic into a subdirectory for readability.
Resolves #1595.


packages/mdc-textfield/adapter.js
```diff
@@ -196,6 +175,12 @@ class MDCTextFieldAdapter {
    * @return {?MDCTextFieldHelperTextFoundation}
    */
   getHelperTextFoundation() {}
+
+  /**
+   * Returns the foundation for the input element.
+   * @return {?MDCTextFieldInputFoundation}
```
lynnjepsen: I think the input foundation is actually required (unlike helper text, which was optional). So this should be `@return {MDCTextFieldInputFoundation}` (no ?)

packages/mdc-textfield/foundation.js
```diff
@@ -91,16 +90,17 @@ class MDCTextFieldFoundation extends MDCFoundation {
   init() {
     this.adapter_.addClass(MDCTextFieldFoundation.cssClasses.UPGRADED);
     // Ensure label does not collide with any pre-filled value.
-    if (this.getNativeInput_().value) {
+    const input = this.adapter_.getInputFoundation();
+    if (input && input.getValue()) {
```
lynnjepsen: input is required. So you should only have to check input.getValue

packages/mdc-textfield/foundation.js
```diff
@@ -128,11 +128,13 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
    */
   handleTextFieldInteraction(evt) {
-    if (this.adapter_.getNativeInput().disabled) {
+    if (this.isDisabled()) {
       return;
     }
-
-    this.receivedUserInput_ = true;
+    const input = this.adapter_.getInputFoundation();
+    if (input) {
```
lynnjepsen: input is required, so you don't need this if statement

packages/mdc-textfield/foundation.js
```diff
@@ -244,23 +235,27 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
+    return input ? input.isBadInput() : false;
   }
 
   /**
    * @return {boolean} True if the Text Field is disabled.
    */
   isDisabled() {
-    return this.getNativeInput_().disabled;
+    const input = this.adapter_.getInputFoundation();
+    return input ? input.isDisabled() : false;
```
lynnjepsen: input is required, so you don't have to check if it exists

packages/mdc-textfield/foundation.js
```diff
@@ -244,23 +235,27 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
   }
 
   /**
    * @param {boolean} disabled Sets the text-field disabled or enabled.
    */
   setDisabled(disabled) {
     const {DISABLED, INVALID} = MDCTextFieldFoundation.cssClasses;
-    this.getNativeInput_().disabled = disabled;
+    const input = this.adapter_.getInputFoundation();
+    if (input) {
```
lynnjepsen: input is required, dont need the if statement

packages/mdc-textfield/index.js
```diff
@@ -61,11 +63,17 @@ class MDCTextField extends MDCComponent {
...
+   * creates a new MDCTextFieldInput.
    */
   initialize(
     rippleFactory = (el) => new MDCRipple(el),
-    bottomLineFactory = (el) => new MDCTextFieldBottomLine(el)) {
-    this.input_ = this.root_.querySelector(strings.INPUT_SELECTOR);
+    bottomLineFactory = (el) => new MDCTextFieldBottomLine(el),
+    inputFactory = (el) => new MDCTextFieldInput(el)) {
+    const inputElement = this.root_.querySelector(strings.INPUT_SELECTOR);
+    if (inputElement) {
```
lynnjepsen: I think we should assume the input element exists

packages/mdc-textfield/index.js
```diff
@@ -98,17 +106,12 @@ class MDCTextField extends MDCComponent {
     if (this.helperText_) {
       this.helperText_.destroy();
     }
+    if (this.input_) {
```
lynnjepsen: input is required, so you dont need the if statement

packages/mdc-textfield/index.js
```diff
@@ -161,6 +164,16 @@ class MDCTextField extends MDCComponent {
       registerTextFieldInteractionHandler: (evtType, handler) => this.root_.addEventListener(evtType, handler),
       deregisterTextFieldInteractionHandler: (evtType, handler) => this.root_.removeEventListener(evtType, handler),
       notifyIconAction: () => this.emit(MDCTextFieldFoundation.strings.ICON_EVENT, {}),
+      registerInputEventHandler: (evtType, handler) => {
+        if (this.input_) {
```
lynnjepsen: dont need if statement

packages/mdc-textfield/index.js
```diff
@@ -34,7 +36,7 @@ class MDCTextField extends MDCComponent {
    */
   constructor(...args) {
     super(...args);
-    /** @private {?Element} */
+    /** @private {?MDCTextFieldInput} */
```
lynnjepsen: you can remove the ? here too

packages/mdc-textfield/input/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Text Field Input
+
+The input element is where the user enters text.
+
+## Design & API Documentation
+
+<ul class="icon-list">
+  <li class="icon-list-item icon-list-item--spec">
+    <a href="https://material.io/guidelines/components/text-fields.html#text-fields-states">Material Design guidelines: Text Fields States</a>
```
lynnjepsen: I think this link

https://material.io/guidelines/components/text-fields.html#text-fields-field-types

might be a better place to link to

packages/mdc-textfield/adapter.js
kfranqueiro: Usually in cases like this, we should have `!` instead of `?`, right? I mean, it shouldn't be null either?

bonniezhou: Yeah, `{MDCTextFieldInputFoundation}` implies `{?MDCTextFieldInputFoundation}`, so I changed the `?` to `!`

packages/mdc-textfield/input/README.md
bonniezhou: Done

bonniezhou:
Made input a required element and removed unnecessary checks for its existence. Also updated tests with a mock input, and added one to assert an error if the `mdc-text-field__input` element isn't found.

packages/mdc-textfield/index.js
acdvorak: I think we want to replace `?` with `!` since input is required

acdvorak:
LGTM!

**Caveat:** I have limited experience with our JS (I've mainly been in Sass land), so Lynn's review is much more trustworthy 😛

packages/mdc-textfield/input/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+True if the input fails validity checks.
+
+##### MDCTextFieldInputFoundation.getValue() => string
+
+Returns the value in the input.
+
+##### MDCTextFieldInputFoundation.checkValidity() => boolean
+
+Returns the result of the checkValidity() call in the native input.
```
lynnjepsen: add a comment about what this returns if the client is using customized validity checking

packages/mdc-textfield/foundation.js
```diff
@@ -128,11 +128,11 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
    */
   handleTextFieldInteraction(evt) {
-    if (this.adapter_.getNativeInput().disabled) {
+    if (this.isDisabled()) {
       return;
     }
-
-    this.receivedUserInput_ = true;
+    const input = this.adapter_.getInputFoundation();
+    input.setReceivedUserInput(true);
```
lynnjepsen: Can this setReceivedUserInput be handled internally by the MDCTextFieldInputFoundation?

bonniezhou: Changed the input method to handleTextFieldInteraction()

packages/mdc-textfield/input/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+True if the input fails validity checks.
+
+##### MDCTextFieldInputFoundation.getValue() => string
+
+Returns the value in the input.
+
+##### MDCTextFieldInputFoundation.checkValidity() => boolean
+
+Returns the result of the checkValidity() call in the native input. This can be the result of customized validity checking.
```
kfranqueiro: This isn't true, is it? The `checkValidity` foundation method simply results in calling `checkValidity` directly on the native input element. (Don't we already have an existing issue related to there currently being no way to retrieve custom validity set by `setValid`?)

packages/mdc-textfield/input/foundation.js
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  notifyPressedAction_() {
+    this.adapter_.notifyPressedAction();
+  }
+
+  /**
+   * @return {boolean} True if the input fails validity checks.
+   */
+  isBadInput() {
```
kfranqueiro: I realize this code existed prior to this split, but...I'm confused why this exists. What is this doing distinct from checkValidity? cc @amsheehan

packages/mdc-textfield/input/foundation.js
```diff
@@ -0,0 +1,181 @@
...
+    this.receivedUserInput_ = true;
+  }
+
+  /**
+   * @return {!Element|!NativeInputType} The native text input from the
+   * host environment, or a dummy if none exists.
+   * @private
+   */
+  getNativeInput_() {
+    return this.adapter_.getNativeInput() ||
```
kfranqueiro: I realize this existed prior to the split, but...is there ever not an input element??

packages/mdc-textfield/input/foundation.js
```diff
@@ -0,0 +1,181 @@
...
+   */
+  isBadInput() {
+    const input = this.getNativeInput_();
+    return input.validity ? input.validity.badInput : input.badInput;
+  }
+
+  /**
+   * @return {string} Returns the value in the input.
+   */
+  getValue() {
```
kfranqueiro: These APIs seem awkward for a foundation when they're just passing through a straight-up DOM operation. Is this something that would no longer be needed after #1675?

(I realize some of these e.g. isDisabled also existed prior to this PR...)

packages/mdc-textfield/input/README.md
lynnjepsen: Ya I think you're right Ken. I guess we don't even reference custom validity in the input subelement foundation. We only reference custom validity in the parent MDCTextFieldFoundation. So we should drop this from the README.

Sorry for the confusion Bonnie, but revert this back to how it was.

packages/mdc-textfield/input/foundation.js
lynnjepsen: Lets not architecture in this PR. This PR is just refactoring. If it illuminates some problems with the old MDCTextFieldFoundation code...great! But file issues and we will resolve in another PR.

packages/mdc-textfield/input/foundation.js
lynnjepsen: The only other "non-awkward", option I can see is to have the foundation return `NativeInputType`. Like a `getNativeInputType()`. But I think I prefer we proxy the methods on `NativeInputType` as methods on the `MDCTextFieldInputFoundation` (aka .getValue and .isDisabled)

We would still need this after #1675. That would only change how we initiate these classes.

packages/mdc-textfield/input/foundation.js
lynnjepsen: Its annotated in the `MDCTextFieldInputAdapter` as `{?Element|?NativeInputType}`. So technically it is optional. But again, lets not architecture in this PR. This PR is just refactoring. If it illuminates some problems with the old MDCTextFieldFoundation code...great! But file issues and we will resolve in another PR.

PR #1636, opened 2017-11-23
demo(text-field): Move script into head

anton-kachurin:
This removes the console error message caused by scripts trying to use the global variable `mdc`  before it was created:
`text-field.html:184 Uncaught ReferenceError: mdc is not defined`

### DIFF
--- a/demos/text-field.html
+++ b/demos/text-field.html
@@ -58,6 +58,7 @@
       }
 
     </style>
+    <script src="/assets/material-components-web.js"></script>
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


lynnjepsen:
I filed an issue, #1808, because I was able to reproduce this quite quickly on our main demo site.

But it we should not merge this PR until we know what the root cause of the issue is. @anton-kachurin do you know why this is happening in the first place?

acdvorak:
Should be fixed by #1919 and #1920. Feel free to reopen if you still see the issue. Thanks! 😀

PR #1639, opened 2017-11-24
feat(snackbar): Emit show or hide event. fixes #1603

### LINKED ISSUE #1603: Notification when snackbar closes?
Is there any way to know when an mdc-snackbar has closed (because the timeout expired)? I am using this component in a React application and need to handle state updates when the snackbar is no longer visible but I cannot find a reliable way to do this other than a separate timer which seems a bit delicate. 

Ideally we would be able to listen for a special event triggered when the mdc-snackbar closes, in the same way as we do for the mdc-dialog comoponent, e.g.:

```js
this.MDCComponent = new MDCSnackbar(myDOMElement);
this.MDCComponent.listen("MDCSnackbar:close", myCloseHandler);
```

If there is a way to coerce the component into doing this with some adapter magic then that would be great but I could not see anything obvious in the docs.

rpatil26:
Add feature to emit `MDCSnackbar:show` when shown or `MDCSnackbar:hide` when
dismissed.


Garbee:
I think using `shown` and `hidden` for the names here would be prudent. These are in essence already completed actions that can't be cancelled or prevented. So using past tense to help solidify this within code seems best. This then allows the addition of pre-completed events in the future using active present tense verbs such as `showing` and `hiding` that would allow a developer to cancel the actions. Without needing to rename things later or have any confusion on the verbage.

rpatil26:
I agree with you @Garbee and wanted to do the same. Simple form of the verb sounds more like a request `to` component, whereas past tense sounds better for notification `from` component. 

I was surprised to see such event naming used by other components too. [Drawer](https://github.com/material-components/material-components-web/tree/master/packages/mdc-drawer) `open` and `close` for example.

I doubt we'll change convention just for this component. Let's see what reviewers think.

rpatil26:
I wrongly used forked master to send this. Closing this and sending another clean one.

PR #1640, opened 2017-11-25
fix(snackbar): Apply visibility hidden when at rest. fixes #1258

### LINKED ISSUE #1258: Possible to see snackbars in hidden state on any platform with rubber-band scrolling at edges

rpatil26:


rpatil26:
Send fresh one instead. https://github.com/material-components/material-components-web/pull/1752.

Issue #1609, opened 2017-11-17
MDC-Card border


MrChriZ:
Is there any reason why the mdc-card__actions does not have a border option like the MDL-Card did - 'mdl-card--border'?

Thanks,
Chris.

amsheehan:
Are you referring to the 1dp stroke above the action buttons [here](https://material.io/guidelines/components/cards.html#cards-actions)?

Nothing in particular from our side, might have been an oversight. Would you mind making a pull request with this change?

MrChriZ:
I am yes.  I've taken a look at the SCSS and wouldn't feel comfortable making that change at present - I'm kind of new on some of these front end side of things, and the MDC SCSS differs from the MDL for the card component that it doesn't look like it can be a copy/paste job. I don't have much time to fully get my head around it at present.  I may revisit when things are less busy if no one has got too it before me but that maybe a month or so away.

amsheehan:
No worries. When you would like to take a crack at it, I believe it would require three small changes:

1. In `packages/mdc-card/mdc-card.scss`, after the `&--vertical {}` block, add the following:
```css
&--with-border {
  border-top: 1px solid rgba(0, 0, 0, .1);
}
```

2. In `demos/card.html`, copy the second demo card (the first one with actions), but add the new class `mdc-card__actions--with-border` to the `<section>` element with class `mdc-card__actions`. It should now look like: 
`class="mdc-card__actions mdc-card__actions--with-border"`

3. In `packages/mdc-card/README.md` document the new modifier class and explain when to use it [here](https://github.com/material-components/material-components-web/tree/master/packages/mdc-card#actions)

I think that would be it. Let us know if you have any questions when you get around to it.

coreymgilmore:
The `mdl-card--border` class was also used on `mdl-card__supporting-text` elements, not just `mdl-card__actions`.  

Instead of adding the `&--with-border` class that would only affect the `mdc-card__actions`, would it not be best to just re-implement the `mdl-card--border` as `mdc-card--border` that can be added to either the supporting text OR the actions?

Suggested:
`
.mdc-card--border {
  border-top: 1px solid rgba(0, 0, 0, 0.1);
}
`

This could just be added to the mdc-card.scss at the end.

Garbee:
We'd need `--border` on each of those elements. Having the global modifier affect the lower children isn't proper BEM. Modifiers go onto the elements they affect directly, not children or parents of the elements in the block.

[...]

PR #1645, opened 2017-11-27
Implement borders between mdc-card content blocks.

coreymgilmore:
Per issue #1609 

MDL had support for adding a 1px divider line between supporting-text and action content blocks.  This adds this functionality to MDC.  Demo cards are updated to show new functionality and README describes usage to users.

The divider line on a .mdc-card__supporting-text element needs to add some padding to separate the content from the divider line.  The size of this padding is the same as the bottom padding on a .mdc-card__supporting-text element.

### DIFF
--- a/packages/mdc-card/README.md
+++ b/packages/mdc-card/README.md
@@ -159,7 +159,7 @@ Be sure to include the `mdc-card__action` class on every action for correct posi
 `mdc-button--compact` class on buttons for correct alignment.
 
 You can use the `mdc-card__actions--border` option to place a divider line to separate the `mdc-card__actions` from 
-the previous element (ex.: a `mdc-card__supporting-text`).  This should not be used with `mdc-card__actions--vertical`.
+the previous element (ex.: a `mdc-card__supporting-text`).
 
 #### Primary title / text
 
--- a/packages/mdc-card/README.md
+++ b/packages/mdc-card/README.md
@@ -158,6 +158,8 @@ You can use the `mdc-card__actions--vertical` option to lay actions out vertical
 Be sure to include the `mdc-card__action` class on every action for correct positioning. Also, be sure to include the
 `mdc-button--compact` class on buttons for correct alignment.
 
+You can use the `mdc-card__actions--border` option to place a divider line to separate the `mdc-card__actions` from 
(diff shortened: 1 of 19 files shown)


coreymgilmore:
I signed It!

coreymgilmore:
I signed it!

packages/mdc-card/README.md
```diff
@@ -158,6 +158,8 @@ You can use the `mdc-card__actions--vertical` option to lay actions out vertical
 Be sure to include the `mdc-card__action` class on every action for correct positioning. Also, be sure to include the
 `mdc-button--compact` class on buttons for correct alignment.
 
+You can use the `mdc-card__actions--border` option to place a divider line to separate the `mdc-card__actions` from 
+the previous element (ex.: a `mdc-card__supporting-text`).  This should not be used with `mdc-card__actions--vertical`.
```
Garbee: Instead of saying it shouldn't be used with vertical mode, why not use the `:not()` operator defining what it shouldn't be used with in order to outright prevent browsers from using the styling where it shouldn't be. i.e. `&--border:not(&--vertical)`

packages/mdc-card/mdc-card.scss
```diff
@@ -66,6 +66,12 @@
     &:last-child {
       padding-bottom: 24px;
     }
+
+    // Divider line similar to MDL
```
Garbee: We don't need comments like this. They aren't useful in any context.

packages/mdc-card/mdc-card.scss
```diff
@@ -104,6 +110,11 @@
         margin-bottom: 0;
       }
     }
+
+    // Divider line above the action elements similar to MDL
```
Garbee: See above comment on comments.

coreymgilmore:
@Garbee Your suggestions are 100% valid.  No idea why I didn't use the `:not()` operator....brain fart.

coreymgilmore:
Changes made to reflect new method of showing divider.  Instead of using classes on mdl-card--action element, this new method uses and `<hr>` element.  This method also allows us to implement the inset divider per the Material Design docs/guide.

PR #1659, opened 2017-11-30
chore(demos): Make simple-menu demo less ugly

acdvorak:
- Use mdc-typography to make text look nicer
- Fix invalid markup
- Reformat markup
- Add a space between radio buttons and their label text
- Use light colored text in dark mode

### DIFF
--- a/demos/simple-menu.html
+++ b/demos/simple-menu.html
@@ -14,7 +14,7 @@
   See the License for the specific language governing permissions and
   limitations under the License
 -->
-<html>
+<html class="mdc-typography">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


demos/simple-menu.html
```diff
@@ -60,14 +60,15 @@
 
       .mdc-theme--dark {
         background-color: #333;
+        color: #ddd;
```
patrickrodee: Legible text is nice feature 👍 

Updating L43 from `margin-top: 16px;` to `padding: 16px 0;` would remove the white gap between the hero and the demo when dark mode is activated. Makes it look just a tad cleaner!

acdvorak: Good call! Done.

PR #1660, opened 2017-11-30
chore(list): Rename CSS class to follow BEM

acdvorak:
BREAKING CHANGE: The `mdc-list-item__text__secondary` class was renamed to `mdc-list-item__secondary-text` to adhere to BEM conventions.

### DIFF
--- a/demos/index.html
+++ b/demos/index.html
@@ -42,63 +42,63 @@
             <span class="catalog-list-icon mdc-list-item__start-detail"><img class="catalog-component-icon" src="/images/ic_button_24px.svg" /></span>
             <a href="button.html" class="mdc-list-item__text">
               Button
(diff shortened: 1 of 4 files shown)


acdvorak:
According to the [BEM FAQ](http://getbem.com/faq/#what-would-be-a-class-name-for-an-element-inside-another-element-block__el1__el2-):

> block structure should be flattened; you do not need to reflect nested DOM structure of the block... Besides the fact that the classes look much nicer, it makes the elements be dependent on the block only. So, you can easily move them across the block when providing changes to the interface. The changes of the block DOM structure would not need corresponding changes to the CSS code.

To me, that quote sort of captures the essence of the BEM philosophy: decoupling your DOM structure from your CSS selectors (as much as reasonably possible) generally makes refactoring faster, safer, easier, and more predictable.

PR #1663, opened 2017-11-30
feat(list): Add color theme mixins & --selected/--activated modifiers

### LINKED ISSUE #1662: Add Sass color mixins to list
- [x] Create color mixins
- [x] Call color mixins from baseline stylesheet
- [x] Add `--selected` modifier for `mdc-list-item` (using primary color)

acdvorak:
Resolves #1662


demos/list.html
```diff
@@ -125,9 +39,9 @@
     <main>
       <div class="mdc-toolbar-fixed-adjust"></div>
       <section class="hero">
-        <ul class="mdc-list mdc-list--two-line mdc-list--avatar-list two-line-avatar-text-icon-demo">
+        <ul class="mdc-list mdc-list--two-line mdc-list--avatar-list demo-list--two-line-avatar-meta-icon demo-list">
```
kfranqueiro: Nit: demo-list before demo-list--modifierthing?

demos/list.html
```diff
@@ -486,10 +485,10 @@ <h3>Example - Avatar with Text and icon</h3>
                 </a>
               </li>
               <li class="mdc-list-item">
-                <img class="mdc-list-item__start-detail grey-bg"
+                <img class="mdc-list-item__start-detail"
                      src="/images/animal2.svg"
-                     width="56" height="56" alt="Sleuth"></span>
-                Sleuth
+                     width="56" height="56" alt="Sloth">
```
kfranqueiro: FYI I think this was also hit in another PR? Did that get merged already?

demos/list.html
```diff
@@ -613,23 +612,23 @@ <h3>Start Detail</h3>
           </section>
           <section>
             <h3>Start Detail (Dense)</h3>
-            <ul class="mdc-list mdc-list--two-line mdc-list--dense">
+            <ul class="mdc-list mdc-list--two-line mdc-list--dense demo-list  demo-list--icon-placeholders">
```
kfranqueiro: Double space -> single space

demos/list.html
```diff
@@ -639,23 +638,23 @@ <h3>Start Detail (Dense)</h3>
           </section>
           <section>
             <h3>Avatar List</h3>
-            <ul class="mdc-list mdc-list--two-line mdc-list--avatar-list">
+            <ul class="mdc-list mdc-list--two-line mdc-list--avatar-list demo-list  demo-list--icon-placeholders">
```
kfranqueiro: Double space -> single space

demos/list.scss
```diff
@@ -19,3 +19,107 @@
...
+
+  @include mdc-theme-dark {
+    @include mdc-list-item-icon-ink-color(#303030);
+    @include mdc-list-item-icon-fill-color(text-icon-on-dark);
+  }
+}
+
+.demo-list--meta-placeholders {
+  .mdc-list-item__end-detail {
+    background-color: currentColor;
```
kfranqueiro: Is this not expressible via a mixin now? (Relevant to another PR that I think you merged?)

demos/list.scss
```diff
@@ -19,3 +19,107 @@
...
+.demo-list--custom {
+  @include mdc-list-item-primary-text-ink-color($material-color-red-600);
+  @include mdc-list-item-secondary-text-ink-color($material-color-orange-500);
+  @include mdc-list-item-icon-fill-color($material-color-yellow-600);
+  @include mdc-list-item-icon-ink-color($material-color-yellow-50);
+  @include mdc-list-item-meta-ink-color($material-color-pink-300);
+}
+
+.demo-list--two-line-avatar-meta-icon .mdc-list-item__start-detail {
+  display: inline-flex;
```
kfranqueiro: Is there anything here that should be centralized in our CSS? (I'm eyeballing this off-the-cuff right now so I could be missing something.)

demos/list.scss
```diff
@@ -19,3 +19,107 @@
...
+}
+
+#demo-wrapper h3 {
+  margin-bottom: 0.8em;
+}
+
+.hero .mdc-list {
+  background-color: white;
+  min-width: 320px;
+  box-shadow:
```
kfranqueiro: Does this map to one of our elevation presets?

demos/list.html
acdvorak: Yes, that already got merged, but they missed the trailing `</span>`

demos/list.html
acdvorak: I have brought shame and dishonor to my family. Fixed 😛

demos/list.html
acdvorak: More shame and dishonor

demos/list.html
acdvorak: More shame and dishonor

demos/list.scss
acdvorak: I don't have a mixin for `background-color` on `__end-detail`, because I didn't see any mention of it in the spec.

And you're right that `mdc-theme-prop` now supports `currentColor` values; I just didn't see a need to use that mixin here since the value is hard-coded and not programmatically- or user-modifiable.

packages/mdc-list/README.md
```diff
@@ -384,3 +384,15 @@ And the basic CSS is relatively trivial
...
+
+### Sass Mixins
+
+Mixin | Description
+--- | ---
+`mdc-list-item-primary-text-ink-color($color)` | Sets the ink color of the primary text
+`mdc-list-item-secondary-text-ink-color($color)` | Sets the ink color of the secondary text
+`mdc-list-item-icon-fill-color($color)` | Sets the fill color of the icon (start detail)
+`mdc-list-item-icon-ink-color($color)` | Sets the ink color of the icon (start detail)
+`mdc-list-item-meta-ink-color($color)` | Sets the ink color of the metadata (end detail)
```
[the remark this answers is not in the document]
acdvorak: I generally want to migrate the names of things to match the spec, cuz it can get really confusing when the names don't match. I was planning to do the actual renaming part in a separate PR, but it felt like it would still make sense to have the documentation match the spec.

The spec calls the first/leading the "Supporting visuals tile", which is super ugly for several reasons, so I didn't use any variations of that phrase.

I figured that because the mixins are only used for icons (they are meaningless when an image avatar is used), having `icon` in the name made more sense than `start-detail` or `leading-tile` or something. But it does have the unfortunate side effect of suggesting that it will change the color of _all_ icons, including trailing icons, which is not actually the case.

So I'm open to suggestions for naming, but I want to try to follow the spec if possible. If the names used in the spec make that difficult or awkward, we can always talk to the designers/writers about updating the spec with better, shorter names for elements.

demos/list.scss
acdvorak: Yes! Great call!

I suppose it's possible that someone would want to put text in the leading tile (rather than an image/icon), and they might not want the text centered, but that seems like a much less likely scenario than images/icons.

demos/list.scss
acdvorak: Yes! `4`. Great catch!

packages/mdc-list/README.md
```diff
@@ -384,3 +384,15 @@ And the basic CSS is relatively trivial
...
+`mdc-list-divider-ink-color($color)` | Sets the ink color of the divider
```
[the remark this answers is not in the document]
acdvorak: Done. I went back and forth on this myself.

packages/mdc-list/_mixins.scss
```diff
@@ -0,0 +1,84 @@
...
+
+@import "@material/rtl/mixins";
+@import "@material/theme/mixins";
+@import "./variables";
+
+//
+// Public
+//
+
+// TODO(acdvorak): Talk to Design about renaming "primary" and "secondary" to something not related to color theming
```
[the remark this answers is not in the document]
acdvorak: Removed. I will worry about renaming it after this PR gets merged.

packages/mdc-list/_mixins.scss
```diff
@@ -0,0 +1,84 @@
...
+  .mdc-list-group__subheader {
+    @include mdc-theme-prop(color, $color);
+  }
+}
+
+//
+// Private
+//
+
+// Sets the width and height of a detail element to the specified dimension.
```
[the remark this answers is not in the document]
acdvorak: Done.

demos/list.html
acdvorak: BTW, I only noticed the invalid markup because IntelliJ is awesome and automatically highlights stuff like that

packages/mdc-list/README.md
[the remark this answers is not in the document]
acdvorak: Renamed "icon" to "graphic" after talking to our designer. The spec will be updated to call it a "supporting graphic" instead of "supporting visual".

packages/mdc-list/mdc-list.scss
```diff
@@ -75,6 +76,11 @@ $mdc-list-side-padding: 16px;
   justify-content: flex-start;
   height: 48px;
 
+  &--selected {
```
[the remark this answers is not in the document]
acdvorak: I believe the `mdc-list-item-graphic-fill-color` and `mdc-list-item-graphic-ink-color` mixins should cover that use case, unless I'm misunderstanding your question.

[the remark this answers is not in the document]
kfranqueiro: I meant a way to do it by emitting fewer styles rather than extra styles that are kind of redundant overrides. But if we think it's more common to color both then we should probably go ahead and do that.

PR #1664, opened 2017-11-30
feat(typography): Support custom properties in mdc-typography mixin

pndewit:
Apply all styles from `$mdc-typography-styles` instead of just a select set of properties.

This allows users to override `$mdc-typography-styles` and add custom properties which will then be used everywhere. For example, adding `color` or a different `font-family`.


packages/mdc-typography/_mixins.scss
```diff
@@ -31,12 +31,13 @@
...
-  font-weight: #{map-get($style-props, font-weight)};
-  letter-spacing: map-get($style-props, letter-spacing);
-  line-height: map-get($style-props, line-height);
-  text-decoration: map-get($style-props, text-decoration);
-  text-transform: map-get($style-props, text-transform);
+  @if not $adjust-margin {
+    $style-props: map_remove($style-props, margin);
+  }
+
+  @each $style, $value in $style-props {
```
acdvorak: `$style` is already being used by a param; maybe rename this to `$prop`?

packages/mdc-typography/_mixins.scss
```diff
@@ -31,12 +31,13 @@
...
   @include mdc-typography-base;
 
-  font-size: map-get($style-props, font-size);
-  font-weight: #{map-get($style-props, font-weight)};
-  letter-spacing: map-get($style-props, letter-spacing);
-  line-height: map-get($style-props, line-height);
-  text-decoration: map-get($style-props, text-decoration);
-  text-transform: map-get($style-props, text-transform);
+  @if not $adjust-margin {
+    $style-props: map_remove($style-props, margin);
```
acdvorak: Our convention is to use underscores for Sass function names: `map-remove`

packages/mdc-typography/_mixins.scss
```diff
@@ -22,7 +22,7 @@
   -webkit-font-smoothing: antialiased;
 }
 
-@mixin mdc-typography($style) {
+@mixin mdc-typography($style, $adjust-margin: false) {
```
acdvorak: Please update the signature of `mdc-typography` in README.md (under ["Sass Variables and Mixins"](https://github.com/pndewit/material-components-web/blob/86ac580c9eb8280c1622b5ea0efc0f933aaca3f4/packages/mdc-typography/README.md#sass-variables-and-mixins)) to include this new param

acdvorak:
Thanks for the PR!

I'm not sure if I would consider `mdc-typography-adjust-margin` deprecated, because it's fairly common to want to change only the margin of an element (especially a heading). Let me double check with the team, though.

pndewit:
Fair enough and thanks for the review :)

packages/mdc-typography/_mixins.scss
pndewit: done

pndewit:
@acdvorak is everything ok now?

pndewit:
@acdvorak Can this be merged now?

acdvorak:
Hey @pndewit, thanks for your patience, and sorry for the delay.

I talked it over with the team, and we would be happy to merge this PR with one small change: Remove the `$adjust-margin` param and always call `map-remove($style-props, margin)`.

Instead of creating a new param, you can just add the `mdc-typography--adjust-margin` class to any element that needs to set a margin, or call the `mdc-typography-adjust-margin` mixin.

I realize it might feel a little cumbersome to need to specify two separate CSS classes or call two separate mixins, but this separation allows us to A) use the `mdc-typography--*` classes on both UI elements (e.g., buttons) _and_ body text, and B) avoid "parameter creep" and mixin overlap.

pndewit:
I understand, updated readme and removed argument.

acdvorak:
LGTM! Thanks @pndewit!

Issue #1564, opened 2017-11-07
Update MDC Button to use mdc-states-color


lynnjepsen:
Use the mdc-states-color mixin created in #1556

MDC Button already uses mdc-ripple-color mixin:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-button/mdc-button.scss#L28

this should be replaced with the new mdc-states-color mixin, with any relevant opacity tweaks.

kfranqueiro:
Implemented in #1668

PR #1668, opened 2017-12-01
feat(button): Use new mdc-states mixin for button styles

kfranqueiro:
BREAKING CHANGE: The $mdc-*-button-ripple-opacity variables have been removed, as these values are now available via the state opacity maps in mdc-ripple.


kfranqueiro:
Fixes #1564

Issue #1562, opened 2017-11-07
Update MDC Fab to use mdc-states-color


lynnjepsen:
Use the mdc-states-color mixin created in #1556

MDC Fab already uses mdc-ripple-color mixin:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-fab/_mixins.scss#L30

this should be replaced with the new mdc-states-color mixin, with any relevant opacity tweaks.

kfranqueiro:
Implemented in #1669.

PR #1669, opened 2017-12-01
feat(fab): Use new mdc-states mixin for fab styles

kfranqueiro:


kfranqueiro:
Fixes #1562

packages/mdc-fab/mdc-fab.scss
```diff
@@ -22,7 +22,7 @@
   @include mdc-fab-base_;
   @include mdc-fab-container-color(secondary);
   @include mdc-fab-ink-color(text-primary-on-secondary);
-  @include mdc-ripple-color(text-primary-on-secondary, $mdc-ripple-pressed-dark-ink-opacity);
+  @include mdc-states(text-primary-on-secondary);
 
   &:not(.mdc-ripple-upgraded) {
     -webkit-tap-highlight-color: rgba(0, 0, 0, .18);
```
amsheehan: Is there are reason this remains a literal value and does not use the `mdc-theme-prop()` mixin?

amsheehan: Is there are reason this remains a literal value and does not use the mdc-theme-prop() mixin?

kfranqueiro: mdc-states already uses mdc-theme-prop-value internally so you can pass a theme property here. mdc-ripple-color used its own logic analogous to that.

amsheehan: Ah, I was referring to line 28

kfranqueiro: Oh. Actually I'm not sure why we never outright removed that style, since we removed it from button styles several releases ago, and it actually looks super-awkward for FAB (it creates a rectangular shade over the FAB when the FAB is round). So I've removed it.

PR #1673, opened 2017-12-01
feat(radio): Use new mdc-states mixin for radio styles

### LINKED ISSUE #1559: Update MDC Radio to use mdc-states-color
Use the mdc-states-color mixin created in #1556

MDC Radio already uses mdc-ripple-color mixin:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-radio/mdc-radio.scss#L36

this should be replaced with the new mdc-states-color mixin, with any relevant opacity tweaks.

kfranqueiro:
Fixes #1559.

### DIFF
--- a/packages/mdc-radio/mdc-radio.scss
+++ b/packages/mdc-radio/mdc-radio.scss
@@ -33,8 +33,8 @@
   @include mdc-radio-ink-color($mdc-radio-baseline-theme-color);
   @include mdc-radio-focus-indicator-color($mdc-radio-baseline-theme-color);
   @include mdc-ripple-surface;
-  @include mdc-states($mdc-radio-baseline-theme-color);
   @include mdc-ripple-radius($mdc-radio-ripple-radius);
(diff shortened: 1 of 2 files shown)


packages/mdc-radio/mdc-radio.scss
```diff
@@ -33,7 +33,7 @@
   @include mdc-radio-ink-color($mdc-radio-baseline-theme-color);
   @include mdc-radio-focus-indicator-color($mdc-radio-baseline-theme-color);
   @include mdc-ripple-surface;
-  @include mdc-ripple-color($mdc-radio-baseline-theme-color, $mdc-radio-ripple-opacity);
+  @include mdc-states($mdc-radio-baseline-theme-color);
```
amsheehan: nit: Put this after all of the `mdc-ripple-*` mixins unless there is some magical states reason not to.

kfranqueiro: Done.

Issue #1558, opened 2017-11-07
Update MDC Text field to use mdc-states-color


lynnjepsen:
Use the mdc-states-color mixin created in #1556

MDC Text field already uses mdc-ripple-color mixin:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/mdc-textfield.scss#L167

this should be replaced with the new mdc-states-color mixin, with any relevant opacity tweaks.

kfranqueiro:
Implemented in #1677

PR #1677, opened 2017-12-01
feat(textfield): Use mdc-states mixin and add support for focus shade

kfranqueiro:


packages/mdc-textfield/README.md
```diff
@@ -285,13 +285,13 @@ can be controlled by passing a ripple factory argument to the constructor.
 
 ```js
 const textFieldBoxEl = document.querySelector('.mdc-text-field--box');
-const textField = new MDCTextField(textFieldBoxEl, /* foundation */ undefined, (el) => {
-  // do something with el...
-  return new MDCRipple(el);
+const textField = new MDCTextField(textFieldBoxEl, /* foundation */ undefined, (el, foundation) => {
```
amsheehan: My only request is that we make it more clear that we mean the ripple foundation and not the component foundation. Otherwise LGTM

kfranqueiro: Done.

kfranqueiro:
Oops... forgot to mention, fixes #1558

PR #1684, opened 2017-12-04
chore(text-field): Pass subelement foundations through MDCTextField super constructor

### LINKED ISSUE #1675: Pass the subelement foundations through MDCTextField's super constructor
Right now we are referencing Subelement Foundations from MDCTextFieldAdapter. This is really just so that MDCTextFieldFoundation can reference the Subelement Foundations.

Instead of passing these foundation through the MDCTextFieldAdapter...we should pass them at initiation time, aka when we create a new MDCTextFieldFoundation, pass all Subelement Foundations as constructor arguments.

MDCTextFieldFoundation is created by MDCTextField, and MDCTextField extends MDCComponent. MDCComponent, the superclass, takes care of  creating the Foundation, so we need to pass Subelement Foundations through the super constructor within MDCTextField

bonniezhou:
Instead of getting Subelement Foundations through MDCTextFieldAdapter, pass them in directly as constructor arguments to MDCTextFieldFoundation through the MDCTextField super constructor.

Resolves #1675.

### DIFF
--- a/packages/mdc-textfield/README.md
+++ b/packages/mdc-textfield/README.md
@@ -285,13 +285,13 @@ can be controlled by passing a ripple factory argument to the constructor.
 
 ```js
 const textFieldBoxEl = document.querySelector('.mdc-text-field--box');
-const textField = new MDCTextField(textFieldBoxEl, /* foundation */ undefined, (el) => {
-  // do something with el...
-  return new MDCRipple(el);
+const textField = new MDCTextField(textFieldBoxEl, /* MDCTextFieldFoundation */ undefined, (el, foundation) => {
+  // Optionally do something with the element or the Ripple foundation...
+  return new MDCRipple(el, foundation);
 });
 ```
 
-By default the ripple factory simply calls `new MDCRipple(el)` and returns the result.
+By default the ripple factory simply calls `new MDCRipple(el, foundation)` and returns the result.
 
 #### MDCTextField API
 
--- a/packages/mdc-textfield/README.md
+++ b/packages/mdc-textfield/README.md
@@ -336,8 +336,8 @@ complicated.
 | registerBottomLineEventHandler(evtType: string, handler: EventListener) => void | Registers an event listener on the bottom line element for a given event |
 | deregisterBottomLineEventHandler(evtType: string, handler: EventListener) => void | Deregisters an event listener on the bottom line element for a given event |
 | getNativeInput() => {value: string, disabled: boolean, badInput: boolean, checkValidity: () => boolean}? | Returns an object representing the native text input element, with a similar API shape. The object returned should include the `value`, `disabled` and `badInput` properties, as well as the `checkValidity()` function. We _never_ alter the value within our code, however we _do_ update the disabled property, so if you choose to duck-type the return value for this method in your implementation it's important to keep this in mind. Also note that this method can return null, which the foundation will handle gracefully. |
-| getBottomLineFoundation() => MDCTextFieldBottomLineFoundation | Returns the instance of the bottom line element's foundation |
-| getHelperTextFoundation() => MDCTextFieldHelperTextFoundation | Returns the instance of the helper text element's foundation |
+
+MDC TextField has multiple optional sub-elements: bottom line and helper text. The foundations of these sub-elements must be passed in as constructor arguments for the MDC TextField foundation. Since the MDCTextField component takes care of creating its foundation, we need to pass sub-element foundations through the MDCTextField component. This is typically done in the component's implementation of `getDefaultFoundation()`.
 
 #### The full foundation API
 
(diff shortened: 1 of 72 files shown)


packages/mdc-textfield/foundation.js
```diff
@@ -15,9 +15,11 @@
  * limitations under the License.
  */
 
+/* eslint-disable no-unused-vars */
```
patrickrodee: To ensure checking isn't disabled for the entire file, this line could be removed and a line-specific lint disable comment could be added above `MDCTextFieldHelperTextFoundation`

packages/mdc-textfield/foundation.js
```diff
@@ -15,9 +15,11 @@
...
 import MDCFoundation from '@material/base/foundation';
 import {MDCTextFieldAdapter, NativeInputType} from './adapter';
 import MDCTextFieldBottomLineFoundation from './bottom-line/foundation';
+import MDCTextFieldHelperTextFoundation from './helper-text/foundation';
```
patrickrodee: See above. Adding this line above line 22 ensures that the `no-unused-vars` check is only disabled for this one instance and not for the entire file.
`// eslint-disable-next-line no-unused-vars`

packages/mdc-textfield/index.js
```diff
@@ -18,11 +18,12 @@
 import MDCComponent from '@material/base/component';
 import {MDCRipple} from '@material/ripple';
 
+/* eslint-disable no-unused-vars */
```
patrickrodee: Same as above. Could be removed to ensure unused var check is only removed for specific lines

packages/mdc-textfield/index.js
```diff
@@ -18,11 +18,12 @@
...
 import {cssClasses, strings} from './constants';
 import {MDCTextFieldAdapter} from './adapter';
 import MDCTextFieldFoundation from './foundation';
-import {MDCTextFieldBottomLine} from './bottom-line';
-import {MDCTextFieldHelperText} from './helper-text';
+import {MDCTextFieldBottomLine, MDCTextFieldBottomLineFoundation} from './bottom-line';
+import {MDCTextFieldHelperText, MDCTextFieldHelperTextFoundation} from './helper-text';
```
patrickrodee: Same as above. Adding `// eslint-disable-next-line no-unused-vars` ensures only specific lines are allowed to ignore rule

packages/mdc-textfield/README.md
```diff
@@ -336,8 +336,8 @@ complicated.
...
-| getBottomLineFoundation() => MDCTextFieldBottomLineFoundation | Returns the instance of the bottom line element's foundation |
-| getHelperTextFoundation() => MDCTextFieldHelperTextFoundation | Returns the instance of the helper text element's foundation |
+
+MDC TextField has multiple optional sub-elements: bottom line and helper text. The foundations of these sub-elements must be passed in as constructor arguments for the MDC TextField foundation. Since the MDCTextField component takes care of creating its foundation, we need to pass sub-element foundations through the MDCTextField component. This is typically done in the component's implementation of `getDefaultFoundation()`.
```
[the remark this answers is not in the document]
bonniezhou: I'm not sure if there is a formal standard, but I think when it's referenced as a concept then I use sentence case, and when it's referenced as code I use markdown. I changed some references above to be in markdown.

packages/mdc-textfield/foundation.js
bonniezhou: Done! Here and in index.js

bonniezhou:
Changed the foundation constructor arguments to be a single map of subfoundations upon @kfranqueiro's suggestion

packages/mdc-textfield/foundation.js
```diff
@@ -57,17 +59,22 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
       deregisterBottomLineEventHandler: () => {},
       getNativeInput: () => {},
-      getBottomLineFoundation: () => {},
-      getHelperTextFoundation: () => {},
     });
   }
 
   /**
    * @param {!MDCTextFieldAdapter=} adapter
+   * @param {!FoundationMapType=} foundationMap Map of all subcomponents to subfoundations.
```
kfranqueiro: Nit: "all" here might seem misleading since they're not always required?

Maybe "Map from subcomponents' names to their subfoundations"?

packages/mdc-textfield/adapter.js
```diff
@@ -31,6 +31,14 @@ import MDCTextFieldHelperTextFoundation from './helper-text/foundation';
  */
 let NativeInputType;
 
+/**
+ * @typedef {{
+ *   bottomLine: (!MDCTextFieldBottomLineFoundation | undefined),
```
acdvorak: Remove the space before and after the `|` (ditto for `helperText` below)

packages/mdc-textfield/foundation.js
```diff
@@ -57,17 +59,22 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
    */
-  constructor(adapter = /** @type {!MDCTextFieldAdapter} */ ({})) {
+  constructor(adapter = /** @type {!MDCTextFieldAdapter} */ ({}),
+    foundationMap = /** @type {!FoundationMapType} */ ({})) {
     super(Object.assign(MDCTextFieldFoundation.defaultAdapter, adapter));
 
+    /** @type {!MDCTextFieldBottomLineFoundation|undefined} */
```
acdvorak: FYI, this is the correct way to annotate an [optional property/param](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#optional-parameter-in-a-param-annotation). I talked to Brendan (our resident Closure Compiler expert) and crawled through the [JS Style Guide](https://google.github.io/styleguide/jsguide.html#jsdoc-nullability) and [Closure Compiler JSDoc wiki](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#optional-parameter-in-a-param-annotation), and they all seem to suggest that your style is the right one.

The rest of our code base uses a nullable type instead of `undefined`
(e.g., `{?MDCTextFieldBottomLineFoundation}`), but that's technically incorrect, and our internal Closure Compiler usually complains about it.

We should probably update the rest of our components to use your style of annotations at some point.

Issue #1561, opened 2017-11-07
Update MDC Icon Toggle to use mdc-states-color


lynnjepsen:
Use the mdc-states-color mixin created in #1556

MDC Icon Toggle already uses mdc-ripple-color mixin:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-icon-toggle/mdc-icon-toggle.scss#L41

this should be replaced with the new mdc-states-color mixin, with any relevant opacity tweaks.

kfranqueiro:
We should also migrate the demo styles to scss and use `mdc-states-color` there to apply the correct opacity styles for the varying-color demo cases.

kfranqueiro:
Implemented in #1685

PR #1685, opened 2017-12-04
feat(icon-toggle): Use new mdc-states mixin for icon-toggle styles

kfranqueiro:
Also simplifies styling of the demo page a bit, in addition to applying opacities based on the examples in the [guidelines](https://material.io/guidelines/components/buttons.html#buttons-toggle-buttons).

Note that this does nothing in terms of the `--primary` or `--secondary` modifiers since I expect that's part of a separate theming-related task.


kfranqueiro:
Fixes #1561

PR #1691, opened 2017-12-05
feat(menu): Add new anchor positioning functionality

aprigogin:


kfranqueiro:
I'm not sure how but it looks like this branch somehow diverged from master and is repeating commits/changes that are already on master? (For example some of my states commits, and the entirety of the mdc-select UX update...)

aprigogin:
@Ken All I was doing I did a rebase. I was also surprised to see all that...

demos/index.html
```diff
@@ -59,6 +59,13 @@
               <span class="mdc-list-item__secondary-text">Multi-selection controls</span>
             </a>
           </li>
+          <li class="mdc-list-item">
```
lynnjepsen: Do we really need a demo page for both context menus and simple menu? I think the updates you did to the simple menu demo page show the anchor positioning better than the context menu demo...so I kinda just want to drop the entire context menu demo

packages/mdc-menu/simple/adapter.js
```diff
@@ -163,8 +163,11 @@ class MDCSimpleMenuAdapter {
   * }} position */
   setPosition(position) {}
 
+  /** @param {string} height */
+  setMaxHeight(height) {}
```
lynnjepsen: remember....before you hit the "squash and merge button" to add a "breaking change" comment to the commit.

packages/mdc-menu/simple/foundation.js
```diff
@@ -149,6 +165,20 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
 
+  /**
+   * @param {Corner} corner Default anchor corner alignment of top-left menu corner.
+   */
+  setAnchorCorner(corner) {
+    this.anchorCorner_ = corner;
+  }
+
+  /**
+   * @param {AnchorMargin} margin 4-plet of margins from anchor.
```
lynnjepsen: hmmm. 4-plet is not really a word...though I see what you are trying to say here. Maybe just add an example of what an argument would look like?

packages/mdc-menu/simple/foundation.js
```diff
@@ -149,6 +165,20 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
     this.adapter_.deregisterBodyClickHandler(this.documentClickHandler_);
   }
 
+  /**
+   * @param {Corner} corner Default anchor corner alignment of top-left menu corner.
+   */
+  setAnchorCorner(corner) {
```
lynnjepsen: Document this in the README

packages/mdc-menu/simple/foundation.js
```diff
@@ -149,6 +165,20 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    this.anchorCorner_ = corner;
+  }
+
+  /**
+   * @param {AnchorMargin} margin 4-plet of margins from anchor.
+   */
+  setAnchorMargin(margin) {
```
lynnjepsen: Document this in the README

packages/mdc-menu/simple/foundation.js
```diff
@@ -411,37 +441,79 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
 
     const anchor = this.adapter_.getAnchorDimensions();
     const windowDimensions = this.adapter_.getWindowDimensions();
-
-    const topOverflow = anchor.top + this.dimensions_.height - windowDimensions.height;
-    const bottomOverflow = this.dimensions_.height - anchor.bottom;
+    const topShift = (this.anchorCorner_ & CornerBit.BOTTOM) ?
```
lynnjepsen: This entire `animationLoop_` method...can you break it up into smaller private methods? There is too much going on right now.

packages/mdc-menu/simple/index.js
```diff
@@ -61,6 +62,21 @@ class MDCSimpleMenu extends MDCComponent {
     this.foundation_.close();
   }
 
+  /**
+   * @param {Corner} corner Default anchor corner alignment of top-left
+   *     menu corner.
+   */
+  setAnchorCorner(corner) {
```
lynnjepsen: Document this in the README

packages/mdc-menu/simple/index.js
```diff
@@ -61,6 +62,21 @@ class MDCSimpleMenu extends MDCComponent {
...
+    this.foundation_.setAnchorCorner(corner);
+  }
+
+  /**
+   * @param {AnchorMargin} margin
+   */
+  setAnchorMargin(margin) {
```
lynnjepsen: Document this in the README

demos/index.html
aprigogin: ok

packages/mdc-menu/simple/adapter.js
aprigogin: So adding, a method is considered a breaking change?

packages/mdc-menu/simple/foundation.js
aprigogin: Done

packages/mdc-menu/simple/adapter.js
kfranqueiro: When it comes to adapter methods, any change is considered breaking, since it impacts authors of libraries which wrap our components (since they have to implement the adapter API themselves).

packages/mdc-menu/simple/foundation.js
kfranqueiro: If this comment has been addressed, I'm not sure the latest commits have been pushed to this remote branch?

packages/mdc-menu/simple/foundation.js
acdvorak: Add nullability qualifier to type (e.g., `@param {!AnchorMargin}`

packages/mdc-menu/simple/index.js
aprigogin: done

packages/mdc-menu/simple/index.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
aprigogin: Refactored... no animationLoop_ method. All is done in css. Auto-positioning logic has been broken up into smaller methods.

packages/mdc-menu/simple/foundation.js
aprigogin: tests have examples of how the argument looks like. I added flexibility to specify just the non-zero (default) values.

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
```diff
@@ -150,109 +168,27 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
   }
 
   /**
-   * Calculates transition delays for individual menu items, so that they fade in one at a time.
-   * @private
+   * @param {!Corner} corner Default anchor corner alignment of top-left menu corner.
```
kfranqueiro: This isn't always controlling the top-left menu corner, right? Doesn't it depend on anchor position within viewport?

packages/mdc-menu/simple/foundation.js
```diff
@@ -150,109 +168,27 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
-   * Removes transition delays from menu items.
-   * @private
+   * @param {!AnchorMargin} margin 4-plet of margins from anchor.
    */
-  removeTransitionDelays_() {
-    const numItems = this.adapter_.getNumberOfItems();
-    for (let i = 0; i < numItems; i++) {
-      this.adapter_.setTransitionDelayForItemAtIndex(i, null);
+  setAnchorMargin(margin) {
+    if (typeof margin.top == 'number') {
```
kfranqueiro: Nit: triple-equals for comparisons (not technically necessary here, but consistent with other MDC code)

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
 
-  /** @private */
-  autoPosition_() {
-    if (!this.adapter_.hasAnchor()) {
-      return;
-    }
+  /**
+   * @return {AutoLayoutMeasures} Measurements used to position menu popup.
+   */
+  getAutoLayoutMeasures_() {
```
kfranqueiro: Maybe rename "Measures" to "Measurements" throughout?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
     // Defaults: open from the top left.
-    let vertical = 'top';
-    let horizontal = 'left';
-
-    const anchor = this.adapter_.getAnchorDimensions();
-    const windowDimensions = this.adapter_.getWindowDimensions();
+    let corner = Corner.TOP_LEFT;
+
+    const {anchorRect, viewportMargin, anchorHeight, anchorWidth, menuHeight, menuWidth} = this.measures_;
+    const bottomAligned = this.anchorCorner_ & CornerBit.BOTTOM;
```
kfranqueiro: Nit: rename to `isBottomAligned` to make it more clear this is a boolean?

Ditto for `leftAligned` below

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    const availableTop = bottomAligned ? anchorRect.top + anchorHeight + this.anchorMargin_.bottom
```
kfranqueiro: I'm a bit confused why there is one reference to `anchorMargin_.bottom` for `availableTop` but never a reference to `anchorMargin_.top` for `availableBottom`. Is there a mistake here or am I just not understanding this fully?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+   * @return {AutoLayoutMeasures} Measurements used to position menu popup.
+   */
+  getAutoLayoutMeasures_() {
+    const anchorRect = this.adapter_.getAnchorDimensions();
+    const viewport = this.adapter_.getWindowDimensions();
+
+    return {
+      anchorRect: anchorRect,
+      viewport: viewport,
+      viewportMargin: {
```
kfranqueiro: Is this actually describing distance from anchor to edge of viewport? Maybe `viewportDistance` is a better term?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+      corner |= CornerBit.BOTTOM;
+    }
 
-    const topOverflow = anchor.top + this.dimensions_.height - windowDimensions.height;
-    const bottomOverflow = this.dimensions_.height - anchor.bottom;
-    const extendsBeyondTopBounds = topOverflow > 0;
+    const leftAligned = this.adapter_.isRtl() ? (this.anchorCorner_ & (CornerBit.FLIP_RTL | CornerBit.RIGHT)) :
+      !(this.anchorCorner_ & CornerBit.RIGHT);
+    const availableLeft = leftAligned ? viewportMargin.left + this.anchorMargin_.left :
+      viewportMargin.left + anchorWidth + this.anchorMargin_.right;
```
kfranqueiro: Similar question to above, why does this reference `anchorMargin_.right` rather than `left`?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    const availableRight = leftAligned ? anchorWidth - this.anchorMargin_.left + viewportMargin.right :
+      viewportMargin.right - this.anchorMargin_.right;
 
-    if (extendsBeyondTopBounds) {
-      if (bottomOverflow < topOverflow) {
-        vertical = 'bottom';
+    const leftOverflow = menuWidth - availableLeft;
+    const rightOverflow = menuWidth - availableRight;
+    if (this.adapter_.isRtl()) {
+      if (leftOverflow < 0 && this.anchorCorner_ & CornerBit.FLIP_RTL & ~CornerBit.RIGHT) {
```
kfranqueiro: There's no else to this inner `if`? Can this simply be flattened into the `if` above?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    const verticalAlignment = (corner & CornerBit.BOTTOM) ? 'bottom' : 'top';
+
+    if (this.anchorCorner_ & CornerBit.BOTTOM) {
+      if (verticalAlignment === 'top') {
+        maxHeight = viewportMargin.bottom - this.anchorMargin_.bottom;
+      } else {
+        maxHeight = viewportMargin.top + this.anchorMargin_.top;
+      }
+    }
+    // perhaps clip at viewport based maxHeight?
```
kfranqueiro: Is this supposed to be a TODO?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+  }
+
+  /**
+   * @param {Corner} corner Origin corner of the menu.
+   * @return {number} Vertical offset of menu origin corner from corresponding anchor corner.
+   * @private
+   */
+  getVerticalOriginOffset_(corner) {
+    const {anchorRect, viewport, viewportMargin, anchorHeight, menuHeight} = this.measures_;
+    const verticalAlignment = (corner & CornerBit.BOTTOM) ? 'bottom' : 'top';
```
kfranqueiro: I'm not sure there's really a point to this variable here, could just inline `corner & CornerBit.BOTTOM` below in the `if`?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
 
+  /**
+   * @param {Corner} corner Origin corner of the menu.
+   * @return {number} Maximum height of the menu, based on available space. 0 indicates should not be set.
+   * @private
+   */
+  getMenuMaxHeight_(corner) {
+    let maxHeight = 0;
+    const {viewportMargin} = this.measures_;
+    const verticalAlignment = (corner & CornerBit.BOTTOM) ? 'bottom' : 'top';
```
kfranqueiro: I'm not sure there's really a point to this variable here, could just inline `corner & CornerBit.BOTTOM` below in the `if`?

Also, I'm not sure I understand why we're only computing max height if bottom-aligned? Shouldn't this be equally relevant to top aligned? Or is that what the maybe-a-TODO below is about?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
     const position = {
-      [horizontal]: '0',
-      [vertical]: '0',
+      [horizontalAlignment]: horizontalOffset ? horizontalOffset + 'px' : '0',
+      [verticalAlignment]: verticalOffset ? verticalOffset + 'px' : '0',
     };
+    const {anchorWidth, menuHeight, menuWidth} = this.measures_;
+    // center align when anchor width is comparable or greater than menu, otherwise keep corner.
+    if (anchorWidth / menuWidth > numbers.ANCHOR_TO_MENU_WIDTH_RATIO) {
+      horizontalAlignment = 'center';
```
kfranqueiro: Reassigning this variable seems confusing given that it was already used for determining a `position` key above.

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    const horizontalOffset = this.getHorizontalOriginOffset_(corner);
+    const verticalOffset = this.getVerticalOriginOffset_(corner);
     const position = {
-      [horizontal]: '0',
-      [vertical]: '0',
+      [horizontalAlignment]: horizontalOffset ? horizontalOffset + 'px' : '0',
+      [verticalAlignment]: verticalOffset ? verticalOffset + 'px' : '0',
     };
+    const {anchorWidth, menuHeight, menuWidth} = this.measures_;
+    // center align when anchor width is comparable or greater than menu, otherwise keep corner.
```
kfranqueiro: Nit: capitalize Center

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    if (anchorWidth / menuWidth > numbers.ANCHOR_TO_MENU_WIDTH_RATIO) {
+      horizontalAlignment = 'center';
+    }
+    // adjust vertical origin when menu can and significantly overlaps anchor.
```
kfranqueiro: Explain why this isn't simply centering as opposed to horizontal above, and initial-cap first word of comment

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+      verticalAlignment = Math.round(originPercent * 100) / 100 + '%';
+    }
 
-    this.adapter_.setTransformOrigin(`${vertical} ${horizontal}`);
+    this.adapter_.setTransformOrigin(`${horizontalAlignment} ${verticalAlignment}`);
     this.adapter_.setPosition(position);
-  }
+    this.adapter_.setMaxHeight(maxMenuHeight ? maxMenuHeight + 'px' : '');
 
+    // clear measures after positioning is complete.
```
kfranqueiro: Nit: capital C in Clear

packages/mdc-menu/simple/foundation.js
```diff
@@ -480,11 +549,13 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
     }
 
     this.adapter_.deregisterBodyClickHandler(this.documentClickHandler_);
-    this.adapter_.addClass(MDCSimpleMenuFoundation.cssClasses.ANIMATING);
+    this.adapter_.addClass(MDCSimpleMenuFoundation.cssClasses.ANIMATING_CLOSE);
```
kfranqueiro: Nit: should this be `ANIMATING_CLOSED` (where "open" and "closed" are adjectives as opposed to "open" and "close" being verbs)?

packages/mdc-menu/simple/mdc-simple-menu.scss
```diff
@@ -53,50 +55,48 @@ $mdc-simple-menu-item-fade-duration: .3s;
...
     display: inline-block;
     transform: scale(1);
     opacity: 1;
+    > .mdc-simple-menu__items {
+      transform: scale(1);
+    }
   }
 
-  &--animating {
+  &--animating-close {
```
kfranqueiro: See earlier comment RE renaming this to `--animating-closed`

packages/mdc-menu/simple/mdc-simple-menu.scss
```diff
@@ -53,50 +55,48 @@ $mdc-simple-menu-item-fade-duration: .3s;
...
+    > .mdc-simple-menu__items {
+      transform: scale(1.25);
+    }
+  }
+
   &--open {
     display: inline-block;
     transform: scale(1);
     opacity: 1;
+    > .mdc-simple-menu__items {
```
kfranqueiro: Is this necessary, or would this be the default style anyway due to the style below under `&__items`?

demos/simple-menu.html
```diff
@@ -82,57 +97,121 @@
...
 
         <div class="demo-controls-container">
           <div class="demo-controls">
-            <p>Button position:</p>
-            <div>
-              <label><input type="radio" name="position" value="top left" checked> Top left</label>
+            <div class="left-column-controls">
+              Button Position:
+              <div>
+                <label><input type="radio" name="position" value="top left" checked> Top left</label>
```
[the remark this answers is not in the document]
aprigogin: done

packages/mdc-menu/README.md
```diff
@@ -108,8 +108,9 @@ or a wrapper element that contains the actual visible element to attach to:
 the menu is positioned and displayed correctly.
 
 The menu will check if its parent element has the `mdc-menu-anchor` class set, and if so, it will automatically position
-itself relative to this anchor element. It will open from the top left (top right in RTL) corner of the anchor by
-default, but will choose an appropriate different corner if close to the edge of the screen.
+itself relative to this anchor element. Unless anchor corner is specified, menu will open from the top left
+(top right in RTL) corner of the anchor by default, but will choose an appropriate different corner if close
```
lynnjepsen: Does it actually make sense to swap this in RTL? Sometimes RTL really only makes sense for text based content. Where a menu opens is not as "text-y". (But if theres guidance in our Material Design articles to do this RTL swap, then it makes sense)

packages/mdc-menu/index.js
```diff
@@ -16,5 +16,6 @@
  */
 
 import * as util from './util';
-import {MDCSimpleMenu, MDCSimpleMenuFoundation} from './simple';
-export {MDCSimpleMenu, MDCSimpleMenuFoundation, util};
+import * as constants from './simple/constants';
+import {MDCSimpleMenu, MDCSimpleMenuFoundation, CornerBit, Corner, AnchorMargin} from './simple';
+export {MDCSimpleMenu, MDCSimpleMenuFoundation, CornerBit, Corner, AnchorMargin, util, constants};
```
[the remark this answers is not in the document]
lynnjepsen: They should be hung off the foundations.

packages/mdc-menu/simple/adapter.js
```diff
@@ -163,8 +139,8 @@ class MDCSimpleMenuAdapter {
   * }} position */
   setPosition(position) {}
 
-  /** @return {number} */
-  getAccurateTime() {}
+  /** @param {string} height */
+  setMaxHeight(height) {}
```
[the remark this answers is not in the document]
lynnjepsen: it is REALLY odd that setMaxHeight doesnt take a number as input. Are you suppose to specify "100px"? Could we change it to just take a number?

packages/mdc-menu/simple/foundation.js
```diff
@@ -96,27 +116,23 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
     /** @private {boolean} */
     this.isOpen_ = false;
     /** @private {number} */
-    this.startScaleX_ = 0;
+    this.openAnimationEndTimerId_ = 0;
```
lynnjepsen: `openAnimationEndTimerId` is a number?

I think `openAnimationEndTimerId`, `closeAnimationEndTimerId`, *and* `selectedTriggerTimerId` all need better inline documentation for what they represent

packages/mdc-menu/simple/foundation.js
lynnjepsen: I think we generally need to be more clear that the corner clients set is the "ideal" corner. But there is internal logic which will change the "actual" corner depending on the view port.

packages/mdc-menu/simple/foundation.js
```diff
@@ -15,10 +15,34 @@
  * limitations under the License.
  */
 
+/**
```
lynnjepsen: Just in general...`MDCSimpleMenuFoundation` is now 570 lines of code. Which is quite a lot. Is there any way to break out some piece of this into a new file? Open to ideas here...

packages/mdc-menu/simple/adapter.js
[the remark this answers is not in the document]
kfranqueiro: I initially had that reaction too, but there also needs to be the possibility to remove it, which would require making the adapter method implementation "smarter" to interpret 0 specifically as `''`. I think it'll seem less awkward with a string parameter with `Style` appended to the method name.

packages/mdc-menu/simple/foundation.js
kfranqueiro: Might be feasible to genericize some of the broken-out auto-positioning functions to util functions?

demos/simple-menu.html
```diff
@@ -66,6 +62,25 @@
...
+      .left-column-controls {
+         display: inline-block;
+      }
+
+      .right-column-controls {
+        display: inline-block;
+        margin-left: 2em;
+      }
+
+      .longAnchorWidth,
```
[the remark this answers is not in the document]
aprigogin: done.

demos/simple-menu.html
```diff
@@ -182,6 +261,97 @@
...
+                break;
+              case "bottom_end":
+                menu.setAnchorCorner(mdc.menu.constants.Corner.BOTTOM_END);
+                break;
+            }
+          }
+        });
+      }
+
+      var inputs = document.querySelectorAll('input[type="text"]');
```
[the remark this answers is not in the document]
aprigogin: done

demos/simple-menu.html
```diff
@@ -182,6 +261,97 @@
...
+      for (var i = 0; i < inputs.length; i++) {
+        inputs[i].addEventListener('change', () => {
+          const topMarginInput = document.querySelector('input[name="topMargin"]');
```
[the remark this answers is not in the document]
aprigogin: Changed for the inputs, but left for radio buttons. Renamed names to be kebab-case.

demos/simple-menu.html
```diff
@@ -182,6 +261,97 @@
...
+        }
+      });
+
+      var radios = document.querySelectorAll('input[name="menulength"]');
+      for (var i = 0; i < radios.length; i++) {
+        radios[i].addEventListener('change', function(evt) {
+          if (evt.target.checked) {
+            switch (evt.target.value) {
+              case "small":
+                document.querySelector('.longMenuItems').style.display = 'none';
```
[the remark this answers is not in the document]
aprigogin: done

demos/simple-menu.html
```diff
@@ -182,6 +261,97 @@
...
+        });
+      }
+
+      var radios = document.querySelectorAll('input[name="anchorwidth"]');
+      for (var i = 0; i < radios.length; i++) {
+        radios[i].addEventListener('change', function(evt) {
+          if (evt.target.checked) {
+            switch (evt.target.value) {
+              case "tiny":
+                document.querySelector('.normalAnchorWidth').style.display = 'none';
```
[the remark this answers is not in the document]
aprigogin: done

packages/mdc-menu/README.md
```diff
@@ -311,6 +307,17 @@ Closes the menu.
 
 Returns whether or not the menu is open.
 
+#### MDCSimpleMenuFoundation.setAnchorCorner(Corner = Corner.TOP_START) => void
+
+Specifies the anchor corner to which top start (top left in ltr, top right in rtl) should align given there is enough
```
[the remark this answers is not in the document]
aprigogin: I updated comment and added more verbiage around flipping and non covering.

packages/mdc-menu/index.js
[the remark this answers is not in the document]
aprigogin: I will look at this when I kill /simple directory and move all of that directly under /menu

packages/mdc-menu/simple/adapter.js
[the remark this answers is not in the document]
aprigogin: added to the README. I followed setTransformOrigin and setPosition name pattern. Open to change them all, but they need to be consistent, I think.

packages/mdc-menu/simple/constants.js
```diff
@@ -40,17 +41,48 @@ const numbers = {
...
 };
 
-export {cssClasses, strings, numbers};
+/**
+ * Enum for bits in the {@see Corner) bitmap.
+ * @enum {number}
+ */
+const CornerBit = {
+  BOTTOM: 1,
+  CENTER: 2,
```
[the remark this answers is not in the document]
aprigogin: CENTER is not used, yet. I am thinking of using it for select. Also this enum is compatible with goog.positioning, so leaving it as is for now.

packages/mdc-menu/simple/foundation.js
aprigogin: it is whatever, setTimeout returns and what you can pass to clearTimeout. Yeah it is a number - see https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout.

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
aprigogin: neah, just needs removal :)

packages/mdc-menu/simple/foundation.js
aprigogin: This just makes origin coincide with anchor, so that scale animation looks like it originates from the anchor.

packages/mdc-menu/simple/foundation.js
aprigogin: I guess it can be confusing... do not want to intro yet another variable though.

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
aprigogin: ok changed to adj meaning, however, I was thinking those as being verbs in this case. Like in "animating whirl".

packages/mdc-menu/simple/mdc-simple-menu.scss
aprigogin: done

packages/mdc-menu/simple/mdc-simple-menu.scss
aprigogin: yes. This is the end state that gets applied after animating-open class is added.

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
aprigogin: keeping my name, unless you feel strongly about that.

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
aprigogin: I think you found a bug... :)

packages/mdc-menu/simple/foundation.js
aprigogin: Actually, here it is correct, because it appears as if menu is aligned with the right side, so right margin is taken into account.

packages/mdc-menu/simple/foundation.js
aprigogin: because of the else if below....

packages/mdc-menu/simple/foundation.js
aprigogin: I think it is a little bit more readable that way.

packages/mdc-menu/simple/foundation.js
aprigogin: I think that having a variable there makes it little bit more readable.
top aligned means you can overlap the anchor, in that case max-height of the menu is driven by css from vh - 32 style.

packages/mdc-menu/README.md
aprigogin: I think so, in RTL - EVERYTHING is mirrored. I do not think there is any guidance in spec about RTL, I will ask Jefferson.

packages/mdc-menu/simple/foundation.js
aprigogin: Separate PR, plz!

packages/mdc-menu/simple/foundation.js
kfranqueiro: `Margin` tends to denote a gap of space where things _aren't_ supposed to appear. The reason I suggested `Distance` instead is because when I was first reading the code that used this, I was thinking this referred to some magical value of pixels we wanted to keep the menu away from the edge of the viewport; it took a while to dawn on me that it actually is strictly the distance to the edge of the viewport from the anchor element.

packages/mdc-menu/simple/foundation.js
kfranqueiro: Actually, can't this be represented a bit more concisely like so, rather than having multiple variably-nested branches that result in the same action?

```js
const shouldUseRightCorner =
  (this.adapter_.isRtl() && leftOverflow < 0 && this.anchorCorner_ & CornerBit.FLIP_RTL & ~CornerBit.RIGHT) ||
  (!this.adapter_.isRtl() && rightOverflow > 0 && leftOverflow < rightOverflow);
if (shouldUseRightCorner) {
  corner |= CornerBit.RIGHT;
}
```

packages/mdc-menu/simple/foundation.js
kfranqueiro: If you want a separate variable, I'd go for something like `isBottomAligned` again instead, that way you can simply do `if (isBottomAligned)` rather than comparing to arbitrary strings. Same for the next instance below.

packages/mdc-menu/simple/foundation.js
kfranqueiro: Add a comment above the outer if below to indicate that the alternative case is handled by CSS (if I'm understanding correctly)

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,202 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
-      [vertical]: '0',
+      [horizontalAlignment]: horizontalOffset ? horizontalOffset + 'px' : '0',
+      [verticalAlignment]: verticalOffset ? verticalOffset + 'px' : '0',
     };
+    const {anchorWidth, menuHeight, menuWidth} = this.measures_;
+    // Center align when anchor width is comparable or greater than menu, otherwise keep corner.
+    if (anchorWidth / menuWidth > numbers.ANCHOR_TO_MENU_WIDTH_RATIO) {
+      horizontalAlignment = 'center';
+    }
+    // Adjust vertical origin when menu can and significantly overlaps anchor.
```
kfranqueiro: When I asked to explain what this meant, I meant elaborate in the comment so people have an easier time figuring out the "why" of this later :)

packages/mdc-menu/simple/foundation.js
kfranqueiro: Yeah, I'll create an issue, since I also realized that I think there are now some unused `util` functions anyway.

demos/simple-menu.html
```diff
@@ -66,6 +62,43 @@
...
+        display: none;
+      }
+
+      .demo-button--normal .demo-button__normal-text,
+      .demo-button--long .demo-button__normal-text,
+      .demo-button--long .demo-button__long-text {
+        display: inline;
+      }
+
+      .mdc-simple-menu__long-items,
```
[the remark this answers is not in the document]
aprigogin: makes sense. done.

demos/simple-menu.html
```diff
@@ -175,13 +278,109 @@
 
               var vertical = evt.target.value.split(' ')[0];
               var horizontal = evt.target.value.split(' ')[1];
-              anchor.style.setProperty(vertical, '0');
+              if (vertical == 'middle') {
```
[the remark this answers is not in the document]
aprigogin: done

demos/simple-menu.html
```diff
@@ -175,13 +278,109 @@
...
+                break;
+            }
+          }
+        });
+      }
+
+      const marginInputsEl = document.querySelector('#margin-inputs');
+      var inputs = marginInputsEl.querySelectorAll('input[type="text"]');
+      for (var i = 0; i < inputs.length; i++) {
+        inputs[i].addEventListener('change', () => {
```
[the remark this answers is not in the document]
aprigogin: done.

packages/mdc-menu/README.md
```diff
@@ -295,7 +292,7 @@ The adapter for simple menu must provide the following functions, with correct s
 | `isRtl() => boolean` | Returns boolean indicating whether the current environment is RTL. |
 | `setTransformOrigin(value: string) => void` | Sets the transform origin for the menu element. |
 | `setPosition(position: { top: string, right: string, bottom: string, left: string }) => void` | Sets the position of the menu element. |
-| `getAccurateTime() => number` | Returns a number representing the number of milliseconds (and fractional milliseconds, as the decimal part) since a given point in time, which should remain constant during the component's lifecycle. Ideally, this should be provided by `window.performance.now()`, which has enough precision for high frequency animation calculations. Using `Date.now()` or equivalent may result in some aliasing in animations. |
+| `setMaxHeight(value: string) => void` | Sets maximum height style for the menu element. |
```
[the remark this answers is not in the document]
aprigogin: done

packages/mdc-menu/README.md
```diff
@@ -311,6 +308,21 @@ Closes the menu.
 
 Returns whether or not the menu is open.
 
+#### MDCSimpleMenuFoundation.setAnchorCorner(Corner = Corner.TOP_START) => void
+
+Specifies the anchor corner to which top start of the menu (top left in ltr, top right in rtl) should align given there
+is enough space to show the full menu.
+When menu display is constrained by viewport edge, TOP can be flipped to BOTTOM. Similarly START can flip to END.
```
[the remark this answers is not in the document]
aprigogin: done

packages/mdc-menu/README.md
```diff
@@ -311,6 +308,21 @@ Closes the menu.
...
+Specifying BOTTOM_START, TOP_END, and BOTTOM_END positions, indicates that the anchor cannot be covered by the menu.
+In such cases maximum height of the menu is enforced.
```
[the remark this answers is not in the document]
aprigogin: done, and good catch.

packages/mdc-menu/README.md
```diff
@@ -311,6 +308,21 @@ Closes the menu.
...
+
+#### MDCSimpleMenuFoundation.setAnchorMargin({top: number, right: number, bottom: number, left: number}) => void
+
+Specifies pixel margins that the menu should be offset from all four sides of an anchor. When menu overlaps an
+anchor (default anchor corner) and menu is taller than would fit (not extending beyond viewport edges), margin is
+ignored and menu allowed to display fully overlapping the anchor. In such cases menu is not strictly aligned to an
```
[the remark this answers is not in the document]
aprigogin: If a menu is aligned TOP_START for instance, and is taller than would fit if you aligned top-left of menu with top-left of anchor, then menu is actually allowed to be taller and be either bottom (or top) aligned to viewport. This happens in cases where increasing menu's height just helps bring more of the menu items on screen.

demos/simple-menu.html
```diff
@@ -81,58 +109,126 @@
     </header>
 
     <main>
-      <div class="mdc-toolbar-fixed-adjust"></div>
-      <section class="hero">
```
kfranqueiro: Why was the hero section removed? If this was for button positioning, shouldn't it be possible to work within the `demo-content` element for that and leave the hero section intact?

packages/mdc-menu/simple/foundation.js
kfranqueiro: Are you planning to rename this to `viewportDistance` from the earlier discussion?

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,209 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+   * @private
+   */
+  getOriginCorner_() {
     // Defaults: open from the top left.
-    let vertical = 'top';
-    let horizontal = 'left';
+    let corner = Corner.TOP_LEFT;
+
+    const {anchorRect, viewportMargin, anchorHeight, anchorWidth, menuHeight, menuWidth} = this.measures_;
+    const isBottomAligned = this.anchorCorner_ & CornerBit.BOTTOM;
```
kfranqueiro: Nit: `Boolean(...);` on right side

demos/simple-menu.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+demo-menu__long-items<!DOCTYPE html>
```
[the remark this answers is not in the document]
aprigogin: fixed.

demos/simple-menu.html
aprigogin: brought it back. Still need to figure out how to size demo section correctly.

demos/simple-menu.html
```diff
@@ -163,25 +259,126 @@
...
+              } else {
+                anchor.style.setProperty(vertical, '0');
+              }
               anchor.style.setProperty(horizontal, '0');
             }
           }
         });
       }
 
+
```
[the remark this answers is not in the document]
aprigogin: done.

packages/mdc-menu/simple/foundation.js
aprigogin: done

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +335,209 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    const topOverflow = menuHeight - availableTop;
+    const bottomOverflow = menuHeight - availableBottom;
+    if (bottomOverflow > 0 && topOverflow < bottomOverflow) {
+      corner |= CornerBit.BOTTOM;
+    }
 
-    const anchor = this.adapter_.getAnchorDimensions();
-    const windowDimensions = this.adapter_.getWindowDimensions();
+    const isRtl = this.adapter_.isRtl();
+    const isFlipRtl = !!(this.anchorCorner_ & CornerBit.FLIP_RTL);
```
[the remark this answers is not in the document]
aprigogin: done

demos/simple-menu.html
aprigogin: added it back. Ignoring issues with height of the demo content portion.

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +340,209 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
-        vertical = 'bottom';
-      }
+  /**
+   * @param {Corner} corner Origin corner of the menu.
+   * @return {number} Horizontal offset of menu origin corner from corresponding anchor corner.
+   * @private
+   */
+  getHorizontalOriginOffset_(corner) {
+    const {anchorRect} = this.measures_;
+    const horizontalAlignment = (corner & CornerBit.RIGHT) ? 'right' : 'left';
```
kfranqueiro: Nit: Would an `isRightAligned` boolean be simpler here, like `isBottomAligned` in the next method below? (Maybe I only commented on that one earlier...)

Also, given that there's really just one assignment within each branch, the `*Offset` variables aren't really necessary.

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -20,23 +20,55 @@ import td from 'testdouble';
...
+ *   Approximate viewport corner where anchor is located.
+ * @param {boolean=} isRtl Indicates whether layout is RTL.
+ * @param {number=} menuHeight Optional height of the menu.
+ */
+function initAnchorLayout(mockAdapter, anchorDimensions, isRtl, menuHeight) {
+  td.when(mockAdapter.hasAnchor()).thenReturn(true);
+  td.when(mockAdapter.getWindowDimensions()).thenReturn({height: 1000, width: 1000});
+  td.when(mockAdapter.getAnchorDimensions()).thenReturn(anchorDimensions);
+
+  if (isRtl) {
```
kfranqueiro: This seems awkward since it's not going to explicitly stub whatsoever if you don't pass `isRtl`, resulting in it being `undefined` (and warranting the coersion here).

It'd make more sense to default `isRtl` to `false` in the function parameters, then simply do `td.when(mockAdapter.isRtl()).thenReturn(isRtl);` (since we always pass a boolean to it).

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -20,23 +20,55 @@ import td from 'testdouble';
...
+    td.when(mockAdapter.isRtl()).thenReturn(!!isRtl);
+  }
+
+  if (menuHeight) {
+    td.when(mockAdapter.getInnerDimensions()).thenReturn({height: menuHeight, width: 100});
```
kfranqueiro: Default `menuHeight` to `200` in the function's parameters, then you don't need this if/else at all, just this line from inside the `if` block.

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -62,216 +94,358 @@ test('exports numbers', () => {
...
   });
 
 testFoundation('#open focuses the menu at the end of the animation', ({foundation, mockAdapter, mockRaf}) => {
-  td.when(mockAdapter.getAccurateTime()).thenReturn(0);
-
   foundation.open();
   mockRaf.flush();
-  mockRaf.flush();
-
-  td.when(mockAdapter.getAccurateTime()).thenReturn(500);
```
kfranqueiro: Given that this test is only waiting for the very next frame, the test name seems inaccurate - are we actually focusing before the animation finishes? (That might be okay, but we should probably remove that part from the test name.)

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -62,216 +94,358 @@ test('exports numbers', () => {
...
-  mockRaf.flush();
   td.verify(mockAdapter.focusItemAtIndex(0), {times: 0});
 });
 
-testFoundation('#open anchors the menu on the top left in LTR, given enough room',
+/** Testing various layout cases for autopositoining */
```
kfranqueiro: Typo: autopositioning

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -62,216 +94,358 @@ test('exports numbers', () => {
...
+    foundation.open();
     mockRaf.flush();
-    td.verify(mockAdapter.setTransformOrigin('top right'));
-    td.verify(mockAdapter.setPosition({right: '0', top: '0'}));
+    td.verify(mockAdapter.setTransformOrigin('right bottom'));
+    td.verify(mockAdapter.setPosition({right: '40px', bottom: '0'}));
   });
 
-testFoundation('#open anchors the menu on the top right in RTL, given enough room',
+testFoundation('#open from small anchor in right bottom of viewport, BOTTOM_START) anchor corner, LTR',
```
kfranqueiro: Typo: remove `)` in test name

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -62,216 +94,358 @@ test('exports numbers', () => {
...
+    initAnchorLayout(mockAdapter, smallBottomLeft, true);
+    foundation.setAnchorCorner(Corner.TOP_END);
+    foundation.open();
     mockRaf.flush();
+    td.verify(mockAdapter.setTransformOrigin('left bottom'));
+    td.verify(mockAdapter.setPosition({left: '40px', bottom: '0'}));
+  });
 
-    td.when(mockAdapter.getAccurateTime()).thenReturn(500);
+testFoundation('#open from small anchor in left bottom of viewport, BOTTOM_START) anchor corner, RTL',
```
kfranqueiro: Typo: remove `)` in test name

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -20,23 +20,55 @@ import td from 'testdouble';
...
 }
 
+// Various anchor dimensions.
+const smallTopLeft = {height: 20, width: 40, top: 20, bottom: 40, left: 20, right: 60};
+const smallBottomLeft = {height: 20, width: 40, top: 920, bottom: 940, left: 20, right: 60};
+const smallBottomRight = {height: 20, width: 40, top: 920, bottom: 940, left: 920, right: 960};
+const smallCenter = {height: 20, width: 40, top: 490, bottom: 510, left: 480, right: 520};
+const smallAboveMiddleLeft = {height: 20, width: 40, top: 400, bottom: 420, left: 20, right: 60};
+const smallBelowMiddleLeft = {height: 20, width: 40, top: 600, bottom: 620, left: 20, right: 60};
+const wideCenter = {height: 20, width: 100, top: 490, bottom: 510, left: 450, right: 550};
```
kfranqueiro: IIUC these configure the anchor to be the same width as the menu. Do we need some tests for when the anchor is wider than the menu to test which side the menu adheres to in LTR vs. RTL for cases where it overlaps the anchor?

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -62,216 +94,358 @@ test('exports numbers', () => {
...
+    td.when(mockAdapter.hasAnchor()).thenReturn(true);
+    td.when(mockAdapter.isRtl()).thenReturn(true);
+    td.when(mockAdapter.getInnerDimensions()).thenReturn({height: 200, width: 100});
+    td.when(mockAdapter.getWindowDimensions()).thenReturn({height: 1000, width: 1000});
+    td.when(mockAdapter.getAnchorDimensions()).thenReturn({
+      height: 20, width: 40, top: 900, bottom: 920, left: 910, right: 950,
+    });
+    foundation.setAnchorCorner(Corner.BOTTOM_START);
+    foundation.setAnchorMargin({top: 5, left: 0, bottom: 10, right: 7});
+    foundation.setAnchorMargin({top: '5px', left: 'zz', bottom: '10dp', right: '7em'});
```
kfranqueiro: Why two `setAnchorMargin` calls?

And what's the motivation for testing invalid input for this specific API? (I doubt we really do that in general...)

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -287,37 +461,30 @@ testFoundation('#close does nothing if event target has aria-disabled set to tru
...
-  mockRaf.flush();
-  td.verify(mockAdapter.addClass('mdc-simple-menu--animating'));
+  td.verify(mockAdapter.addClass(cssClasses.ANIMATING_CLOSED));
 });
 
 testFoundation('#close removes the open class from the menu', ({foundation, mockAdapter, mockRaf}) => {
-  td.when(mockAdapter.hasClass('mdc-simple-menu--open')).thenReturn(true);
-  td.when(mockAdapter.hasClass('mdc-simple-menu--open-from-bottom-left')).thenReturn(true);
-
+  td.when(mockAdapter.hasClass(cssClasses.OPEN)).thenReturn(true);
```
kfranqueiro: Is this line even necessary, here and in the next test?

demos/simple-menu.html
```diff
@@ -28,33 +28,28 @@
...
       body,
       main {
-        display: flex;
+        position: relative;
         flex-direction: column;
-        height: 100%;
-      }
-
-      .demo-header-toolbar {
-        z-index: 10;
```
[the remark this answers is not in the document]
aprigogin: Well, true. However, popup menus should show over the title bar. I would argue that z-indexing of the bar itself is not right.

packages/mdc-menu/README.md
aprigogin: yeah that got fixed

demos/simple-menu.html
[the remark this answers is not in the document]
kfranqueiro: I guess you could argue for changing the z-index of the specific hero menu instance instead?  One way or another, this menu shouldn't overlap the toolbar; it breaks the perception of the toolbar being "over" the page.

For future thought: I'm not sure how typical it is for pages to scroll while menus are open, but it seems more common to have a menu within the body of a page rather than in its fixed toolbar and in that case you wouldn't want the menu to cover the toolbar, so ultimately menu-in-toolbar may warrant a special-case `z-index` of its own instead.

packages/mdc-menu/simple/foundation.js
```diff
@@ -150,109 +173,27 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
   /**
-   * Removes transition delays from menu items.
-   * @private
+   * @param {!AnchorMargin} margin 4-plet of margins from anchor.
    */
-  removeTransitionDelays_() {
-    const numItems = this.adapter_.getNumberOfItems();
-    for (let i = 0; i < numItems; i++) {
-      this.adapter_.setTransitionDelayForItemAtIndex(i, null);
+  setAnchorMargin(margin) {
```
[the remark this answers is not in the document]
aprigogin: I wanted to allow people to call setAnchorMargin({top: 5, bottom: -5}) without having to worry about right/left. This API kind of follows the css style specification in that it does not reset margin to 0 for sides you have not specified.

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +340,209 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+   */
+  getOriginCorner_() {
     // Defaults: open from the top left.
-    let vertical = 'top';
-    let horizontal = 'left';
+    let corner = Corner.TOP_LEFT;
+
+    const {anchorRect, viewportDistance, anchorHeight, anchorWidth, menuHeight, menuWidth} = this.measures_;
+    const isBottomAligned = Boolean(this.anchorCorner_ & CornerBit.BOTTOM);
+    const availableTop = isBottomAligned ? anchorRect.top + anchorHeight + this.anchorMargin_.bottom
```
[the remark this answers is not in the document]
aprigogin: I think those are always equal, but good catch for consistency.

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +340,209 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
+    }
 
-    const anchor = this.adapter_.getAnchorDimensions();
-    const windowDimensions = this.adapter_.getWindowDimensions();
+    const isRtl = this.adapter_.isRtl();
+    const isFlipRtl = Boolean(this.anchorCorner_ & CornerBit.FLIP_RTL);
+    const avoidHorizontalOverlap = Boolean(this.anchorCorner_ & CornerBit.RIGHT);
+    const isAlignedRight = (avoidHorizontalOverlap && !isRtl) ||
+      (!avoidHorizontalOverlap && isFlipRtl && isRtl);
+    const availableLeft = isAlignedRight ? viewportDistance.left + anchorWidth + this.anchorMargin_.right :
```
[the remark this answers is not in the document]
aprigogin: Suppose we are aligning to the TOP-LEFT. When menu should appear aligned to the right edge of the anchor, we need to add or subtract right margin. The difference between addition vs subtraction comes from asymmetry with respect to direction of an axis.

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: sure, makes sense. I guess I am not used to setting default values on params (yet)!

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: done

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: done.

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: done

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: done

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: code coverage is one of the motivations.

packages/mdc-menu/simple/foundation.js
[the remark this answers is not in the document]
kfranqueiro: I'd argue that in your example, people _do_ have to worry about whether right/left have been set previously. i.e. previous calls to this `set` API can have side effects on future calls, and if they want to be sure that _only_ the margins they pass are in effect, they have to pass all 4 sides and zero out the ones they don't want, which seems less convenient in the long run.

IMO it's easier to consume `set` APIs that function as resets, not potentially additive across subsequent calls.

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: done

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: strictly speaking it is not needed :).

test/unit/mdc-menu/simple.foundation.test.js
kfranqueiro: Oh, for the `typeof` operations? This may become unnecessary then, if we revise the API RE my other comment (i.e. the branch this hits will no longer exist, and those branches are really intended for nonexistent values rather than invalid ones, right?).

packages/mdc-menu/simple/foundation.js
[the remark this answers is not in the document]
aprigogin: I do not expect people to change margins more than once, frankly. It has to do with anchor geometry and should probably not change between invocations. So, being able to specify just the margins that one cares about is useful. One can always set all 4 to be sure. If you feel strongly about this, I can add assert for all 4 fields to be there on each set.
I do not think that not including a side should reset that side to 0 margin.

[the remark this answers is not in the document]
kfranqueiro: Discussed offline, but in summary: I definitely agree that this will be called only once in most cases, which makes me feel a little guilty about harping on it :) but resetting the unspecified states seems cleaner from a setter API perspective, and also can be accomplished with less code, which seems like a win-win to me.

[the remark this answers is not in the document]
aprigogin: Single liner breaks closure test, since margin is of type AnchorMargin and not Object.

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -1,14 +1,14 @@
 /**
  * Copyright 2016 Google Inc. All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the 'License');
```
kfranqueiro: I assume some kind of mass find/replace altered these comments in the copyright; we should revert these 2 edits

demos/simple-menu.html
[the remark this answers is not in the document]
aprigogin: Ok I put back z-index on the toolbar and added z-index for the popup menu. I still think it sets a bad example moving forward - "z-index war" is a loosing war.

demos/simple-menu.html
```diff
@@ -163,25 +272,124 @@
...
+          const margin = {top: parseInt(topMarginInput.value, 10),
+                          right: parseInt(rightMarginInput.value, 10),
+                          bottom: parseInt(bottomMarginInput.value, 10),
+                          left: parseInt(leftMarginInput.value, 10)};
+          menu.setAnchorMargin(margin);
+        });
+      }
+
+      const rtl = document.querySelector('input[name="is-rtl"]');
+      rtl.addEventListener('change', () => {
```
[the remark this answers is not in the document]
aprigogin: done.

test/unit/mdc-menu/simple.foundation.test.js
aprigogin: great eye!

packages/mdc-menu/simple/foundation.js
```diff
@@ -399,69 +332,208 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
...
-    }
+  /**
+   * @return {AutoLayoutMeasurements} Measurements used to position menu popup.
+   */
+  getAutoLayoutMeasurements_() {
+    const anchorRect = this.adapter_.getAnchorDimensions();
+    const viewport = this.adapter_.getWindowDimensions();
+
+    return {
+      anchorRect: anchorRect,
```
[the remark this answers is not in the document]
aprigogin: Good point, refactored affected methods.

demos/simple-menu.html
```diff
@@ -32,6 +32,14 @@
         flex-direction: column;
       }
 
+      .demo-header-toolbar {
+        z-index: 10;
+      }
+
+      #demo-menu {
```
kfranqueiro: I was thinking this menu should _not_ overlap the toolbar since the toolbar is supposed to be thought of as being above the page.

If you want this to overlap the toolbar, it'd be simpler to only override the z-index on the menu in the hero section, instead of overriding the z-index of both the toolbar and this menu.

Also, generally avoid ID selectors in demo CSS.

aprigogin: done

PR #1694, opened 2017-12-05
fix(linear-progress): changed default to 0

### LINKED ISSUE #1528: Initial value of linear progress component

> 0.24.0

> All

### What OS are you using?

> Windows 10

> https://codepen.io/corneschouten/pen/RjWOay
> The solution is already applied in the CSS area of the codepen.

> Start the linear progress with an initial value of 0%. 
> This could be easily achieved by adding the following css:
> `.mdc-linear-progress__primary-bar { transform: scaleX(0); }`

> The linear progress starts with an initial value of 100% .

patrickrodee:
Fixes bug where determinate linear-progress was defaulting to 100% when progress value was not set.

Fixes #1528


patrickrodee:
Since linear progress has no `getter` for `progress`, setting a default value of 0 for the `progress` value is unnecessary.

patrickrodee:
Re- adding tests for the default `transform: scaleX(0);`

Since the tests run without loading the styles, we'd have to duplicate the fix inside the foundation by manually setting `transform` to `scaleX(0)` to ensure it is accessible and passes. Seems like duplicate code just to ensure we have a test case.

Issue #1556, opened 2017-11-07
Add new mdc-states-color mixin


lynnjepsen:
Expose a mixin which, given an ink color, determines whether to apply state opacities for light UI (for dark ink) or colorized UI (for light ink).

This mixin will handle Enabled, Hovered, Focused, and Pressed

kfranqueiro:
This is committed in #1624. However, we'll need to clear out the old mixins once everything's converted, and I also need to add activated/selected support in a later iteration. Do we want to track that separately or keep this open?

lynnjepsen:
Filed #1713 for the activated/selected. And a chore in pivotal to delete the deprecated mixins

kfranqueiro:
OK, then I'll close this one since the remaining items are accounted for elsewhere.

PR #1696, opened 2017-12-05
feat(ripple): Add support for activated and selected states

### LINKED ISSUE #1713: Update states to handle activated/selected state
The states subsystem should handle activated and selected states, in addition to the existing hover, focus, and pressed states

kfranqueiro:
Related to #1556; resolves #1713.

This adds support for optionally specifying a color for activated and/or selected states in the simple `mdc-states` mixin. If the values are not present, the styles are not emitted for these modifiers.

In cases where you're using the advanced mixins, you would simply re-apply those mixins within e.g. a `.foo--activated` or `.foo--selected` selector.


lynnjepsen:
Resolves #1713

packages/mdc-ripple/_mixins.scss
```diff
@@ -169,6 +168,41 @@ $mdc-ripple-common-styles-emitted_: false !default;
   @include mdc-states-hover-opacity(map-get($opacity-map, "hover"));
   @include mdc-states-focus-opacity(map-get($opacity-map, "focus"), $has-nested-focusable-element);
   @include mdc-states-press-opacity(map-get($opacity-map, "press"));
+
+  @if $activated-color {
```
lynnjepsen: Can you create a `mdc-activated-state($color, $has-nested-focusable-element)` mixin which just handles activated state? It seems like its isolated logic

packages/mdc-ripple/_mixins.scss
```diff
@@ -169,6 +168,41 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+    &--activated {
```
lynnjepsen: does --activated work with all existing components? Or will this be a new modifier for some components?

packages/mdc-ripple/_mixins.scss
```diff
@@ -169,6 +168,41 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+      @include mdc-states-hover-opacity(map-get($opacity-map, "activated") + map-get($opacity-map, "hover"));
+      @include mdc-states-focus-opacity(
+        map-get($opacity-map, "activated") + map-get($opacity-map, "focus"),
+        $has-nested-focusable-element
+      );
+      @include mdc-states-press-opacity(map-get($opacity-map, "activated") + map-get($opacity-map, "press"));
+    }
+  }
+
+  @if $selected-color {
```
lynnjepsen: Can you create a mdc-selected-state($color, $has-nested-focusable-element) mixin which just handles selected state? It seems like its isolated logic

packages/mdc-ripple/_mixins.scss
```diff
@@ -169,6 +168,41 @@ $mdc-ripple-common-styles-emitted_: false !default;
...
+    &--selected {
```
lynnjepsen: does --selected work with all existing components? Or will this be a new modifier for some components?

kfranqueiro:
I've updated the branch to separate the basic mixin into 3 (one for normal interaction states, one for when a surface is activated, and one for when it is selected).

RE the questions around working with all existing components, no, these will be used in the future by components intended to support these states. Currently drawer and menu have some modifiers for selected state, but those need to be tweaked a bit because they're modifying the wrong block. Andy has also done some work in https://github.com/material-components/material-components-web/commit/6ea948f266b340f6733e3b005c0e25e93028955c to make way for the activated/selected modifiers already within MDC List, which will be the first thing to use these mixins.

PR #1697, opened 2017-12-06
chore(text-field): Split out icon into subelement

### LINKED ISSUE #1597: Split out icon from text field parent
Part of our effort to split text field into sub directories, to reduce merge conflicts and generally improve the readability of text field code.

bonniezhou:
Refactoring icon logic into a subdirectory for readability.
Resolves #1597.


packages/mdc-textfield/README.md
```diff
@@ -117,37 +117,6 @@ By default an input's validity is checked via `checkValidity()` on blur, and the
 accordingly. Set the MDCTextField.valid field to set the input's validity explicitly. MDC Text Field
 automatically appends an asterisk to the label text if the required attribute is set.
 
-### Leading and Trailing Icons
```
kfranqueiro: I don't think we should remove this from the top level readme... The top-level readme is all people should need to look at in order to use the component, and they're less likely to look at the subelements' readmes unless they're interested specifically in implementing their own adapters/components.

packages/mdc-textfield/foundation.js
```diff
@@ -141,13 +142,8 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
 
-    const {target, type} = evt;
-    const {TEXT_FIELD_ICON} = MDCTextFieldFoundation.cssClasses;
-    const targetIsIcon = this.adapter_.eventTargetHasClass(target, TEXT_FIELD_ICON);
-    const eventTriggersNotification = type === 'click' || evt.key === 'Enter' || evt.keyCode === 13;
-
-    if (targetIsIcon && eventTriggersNotification) {
-      this.adapter_.notifyIconAction();
+    if (this.icon_) {
+      this.icon_.handleTextFieldInteraction(evt);
```
kfranqueiro: This is still listening at the text field's root element which means the icon code still needs to filter it. Wasn't our thought the other day to actually listen to this event at the icon level? Then we might not even need the `eventTargetHasClass` icon adapter API...

packages/mdc-textfield/icon/README.md
```diff
@@ -0,0 +1,76 @@
...
+excerpt: "Icons describe the type of input a text field requires"
+iconId: text_field
+path: /catalog/input-controls/text-fields/icon/
+-->
+
+# Text Field Icon
+
+Icons describe the type of input a text field requires. They can also be interaction targets.
+
+## Design & API Documentation
```
kfranqueiro: See comment in top-level README; I think the templating aspects should be left there and just document the foundation/adapter here.

packages/mdc-textfield/icon/README.md
```diff
@@ -0,0 +1,76 @@
...
+and the css will ensure the cursor is set to default, and that actioning on an icon doesn't
+do anything unexpected.
+
+#### MDCTextFieldIcon API
+
+##### MDCTextFieldIcon.foundation
+
+MDCTextFieldIconFoundation. This allows the parent MDCTextField component to access the public methods on the MDCTextFieldIconFoundation class.
+
+### Using the foundation class
```
kfranqueiro: The table under this heading is the adapter API, not the foundation, right? We should either re-word this heading or add a sentence explaining that.

packages/mdc-textfield/icon/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Method Signature | Description
+--- | ---
+setAttr(attr: string, value: string) => void | Sets an attribute with a given value on the icon element
+eventTargetHasClass(target: EventTarget, className: string) => boolean | Returns true if classname exists for a given target element
+notifyIconAction() => void | Emits a custom event "MDCTextField:icon" denoting a user has clicked the icon
```
kfranqueiro: We might want to call out that this event should bubble, so that it can be listened for from the text field's root element.

packages/mdc-textfield/icon/foundation.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Handles a text field interaction event
+   * @param {!Event} evt
+   */
+  handleTextFieldInteraction(evt) {
+    const {target, type} = evt;
+    const {TEXT_FIELD_ICON} = MDCTextFieldIconFoundation.cssClasses;
+    const targetIsIcon = this.adapter_.eventTargetHasClass(target, TEXT_FIELD_ICON);
```
kfranqueiro: I think this check shouldn't be necessary if we actually handle the event on this node rather than passing one in from the root text field element. Moreover, then `cssClasses` shouldn't be necessary at all for this subcomponent either. (However, we'd need `[de]registerInteractionHandler` on the icon adapter instead.)

packages/mdc-textfield/icon/index.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * @return {!MDCTextFieldIconFoundation}
+   */
+  getDefaultFoundation() {
+    return new MDCTextFieldIconFoundation(/** @type {!MDCTextFieldIconAdapter} */ (Object.assign({
+      setAttr: (attr, value) => this.root_.setAttribute(attr, value),
+      eventTargetHasClass: (target, className) => target.classList.contains(className),
+      notifyIconAction: () => this.emit(MDCTextFieldIconFoundation.strings.ICON_EVENT, {}),
```
kfranqueiro: IMO this event should bubble, so that it's simple to listen for from the root text field element rather than needing to pry into this subelement. The text field root component/element should still be considered the public-facing API that users will directly consume.

test/unit/mdc-textfield/mdc-text-field-icon.test.js
```diff
@@ -0,0 +1,64 @@
...
+  root.classList.add('foo');
+  assert.equal(component.getDefaultFoundation().adapter_.eventTargetHasClass(root, 'foo'), true);
+});
+
+test('#adapter.eventTargetHasClass returns false if classname does not exist for a given target element', () => {
+  const {root, component} = setupTest();
+  assert.equal(component.getDefaultFoundation().adapter_.eventTargetHasClass(root, 'foo'), false);
+});
+
+test('#adapter.notifyIconAction emits ' + `${MDCTextFieldIconFoundation.strings.ICON_EVENT}`, () => {
```
kfranqueiro: Perhaps we should also verify that this event is configured to bubble in this test?

packages/mdc-textfield/README.md
bonniezhou: My understanding was that the top-level README has everything that's required to get a basic working text field, and all the "extras" fall into their subelement READMEs. 
I'm not too adamant on having it one way or the other, but if we go with your case then we should also move stuff in helper-text README back to the top-level one.

bonniezhou: @lynnjepsen suggested that we move as much as possible to the subelement READMEs but mention each subelement in the toplevel README with a "See here for more info" link.

kfranqueiro: Yeah, links would be a good idea at the very least. My main concern is that we're making docs on certain features harder to find otherwise.

packages/mdc-textfield/icon/README.md
bonniezhou: Will address this in a separate PR once all the subelements have been split out, and reorganize the entire README so it's more readable.

packages/mdc-textfield/icon/README.md
```diff
@@ -0,0 +1,77 @@
...
+##### MDCTextFieldIcon.foundation
+
+MDCTextFieldIconFoundation. This allows the parent MDCTextField component to access the public methods on the MDCTextFieldIconFoundation class.
+
+### Using the foundation class
+
+Method Signature | Description
+--- | ---
+setAttr(attr: string, value: string) => void | Sets an attribute with a given value on the icon element
+registerInteractionHandler(evtType: string, handler: EventListener) => void | Registers an event listener on the input element for a given event
```
kfranqueiro: Remove "on the input element" from this and the next line (this wording is old, and there's really only one element in this subcomponent so we probably don't need a distinction)

packages/mdc-textfield/README.md
```diff
@@ -117,36 +117,16 @@ By default an input's validity is checked via `checkValidity()` on blur, and the
...
-<div class="mdc-text-field mdc-text-field--box mdc-text-field--with-leading-icon">
-  <i class="material-icons mdc-text-field__icon" tabindex="0">event</i>
-  <input type="text" id="my-input" class="mdc-text-field__input">
-  <label for="my-input" class="mdc-text-field__label">Your Name</label>
-  <div class="mdc-text-field__bottom-line"></div>
-</div>
-```
+The helper text provides supplemental information and/or validation messages to users. It appears on input field focus
+and disappears on input field blur by default, or it can be persistent. 
+See <a href="/catalog/input-controls/text-fields/helper-text/">here</a> for more information on using helper text.
```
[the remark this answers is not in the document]
bonniezhou: Thanks!

packages/mdc-textfield/icon/README.md
bonniezhou: Nice catch, done.

kfranqueiro:
This PR shouldn't include any package-lock changes/additions, since it doesn't actually affect dependencies at all. You should be able to restore master's onto this branch by doing `git checkout master -- package-lock.json`. I think mdc-elevation's package-lock file in this PR was added and didn't already exist, so it can just be removed.

(I thought I'd commented on this before but I can't find that comment now...)

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -96,17 +133,50 @@ test('#destroy cleans up the ripple if present', () => {
...
+  td.verify(component.bottomLine_.destroy());
+});
+
+test('#destroy cleans up the helper text if present', () => {
+  const root = getFixture();
+  root.querySelector('.mdc-text-field__input').setAttribute('aria-controls', 'helper-text');
+  const helperText = getHelperTextElement();
+  document.body.appendChild(helperText);
+  const component =
+    new MDCTextField(root, undefined, undefined, undefined, (el) => new FakeHelperText(el));
```
kfranqueiro: Should we bake these other fake factories into `setupTest` so that we can then reference them directly from its return map rather than needing to dig into arguably internal instance properties (e.g. `helperText_` here, `icon_` and `label_` below)?

bonniezhou: Good call, done!

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -88,6 +113,36 @@ test('#constructor instantiates a helper text on the element with id specified i
...
+  assert.instanceOf(component.icon_, MDCTextFieldIcon);
+});
+
+test('#constructor instantiates a label on the `.mdc-text-field__label` element if present', () => {
+  const root = getFixture();
+  const component = new MDCTextField(root);
+  assert.instanceOf(component.label_, MDCTextFieldLabel);
+});
+
+function setupTest(fixture) {
```
kfranqueiro: I think you can do `fixture = getFixture()` for the parameter here instead of `fixture || getFixture()` for root below.

Issue #24, opened 2016-12-05
Investigate making the Tooltip Component


traviskaufman:
Spec reference: https://material.io/go/design-tooltips

> Copied from https://github.com/google/material-design-lite/issues/4495

MrUnknownGamer:
Any update on this? Wanting to use this for some of my drawer list items.

menelike:
Maybe I've missed something, but currently `:before` and `:after` classes are used for ripple effects on [buttons](https://github.com/material-components/material-components-web/blob/e1686e0e2cccfe42becc320b70af0d552e7212af/packages/mdc-button/mdc-button.scss#L26-L28). Adding tooltips could be a hard task as they will conflict with that.

Imho ripple should be an opt-in and not included per default on components.

btakita:
Looks like Tooltips are a ways down in the priority list.

https://www.pivotaltracker.com/story/show/128619029

raupes:
Please add the tooltip and the other missing controls that are already available at MDL

Garbee:
The flow isn't to match MDL as rapidly as possible. It's to build the foundation components first and build up from those to the extraneous. Tooltip's are something that need to get tested in various scenarios across multiple components to make sure they work in all the places developers will want them. Which is why it is one of the last planned items.

j-o-d-o:
Is anybody working on this?

rhroyston:
My though is that tooltips are exclusive to mouse cursors hovering. I can't imagine how a tooltip would work on a touch (mobile or tablet). Noone holds their finger a quarter away from the touchscreen do they? Yet, [Material Design specifies](https://material.io/guidelines/components/tooltips.html#) for those use cases (mobile/tablet)??? I think [Microsoft is of the same mind](https://msdn.microsoft.com/en-us/library/windows/desktop/bb760250(v=vs.85).aspx):

> Tooltips appear automatically, or pop up, when the user pauses the mouse pointer over a tool or some other UI element.

I think if MDC-Web tackled desktop (keyboard attached) devices only it really could be coded in under a day w/ CSS `position`ing. A default for top with modifier class for bottom, left, and right.

Anyways, just sayin...

rhroyston:
@Zodiase You are correct. Good point. However, the press-and-hold gesture feature is awful no matter the manufacturer/device. There is a reason that the mouse/mousepad exists and persists. press-and-hold is plain awful in my experience. In fact, press-and-hold has probably resulted in thousands of driving related deaths.

[...]

PR #1701, opened 2017-12-06
feat(tooltip): Implementation for Tooltips

j-o-d-o:
I found some time yesterday so I decided I set up a first draft for a Tooltip component for MDC. Even thou most of the MDC architecture is quite straightforward, I am quite new to it and I might have messed up some things here and there.

------

**So here is a little technical breakdown:**
The tooltip is represented by a `<span/>` and can either be placed inside the controlling element or outside with the 'for' attribute pointing to the ID of the controlling element. If it is placed inside, the tooltip element will be moved outside of the controlling element. That way the controlling element can have `overflow: hidden` and the tooltip can still show up properly. The position will be calculated with the width and height of the controlling element and the tooltip element as well as the offsetTop and offsetLeft from the controlling element. The tooltip itself has the `position: absolute`. 
The tooltip shows on hover, focus and touchstart. On these events, it will move from the center of the controlling element to the specified direction. When it hides it just fades out. After fading out it will be recentered to the controlling element. 
I tried to follow https://material.io/guidelines/components/tooltips.html as good as possible.
For more infos on how to use them look at the readme or the demo site. 

**ToDos/Unclear**
- should there be a delay after hover to show the tooltips?
- should the tooltips scale up down on show hide like in MDL?
- check if tooltip move out of window -> if yes -> chose another direction
- what z-index should the tooltip element have?
- add Closure compatibility 

#24


j-o-d-o:
Just to make it clear, it is ready for review ; )

lynnjepsen:
Hey @j-o-d-o , sorry this is a very large PR for use to review...so thanks for your continued patience. In the meantime could you fix up the merge conflicts? We will try and find someone to review this more thoroughly and work with you.

j-o-d-o:
No problem. I will try to keep the branch up to date with master.

packages/mdc-tooltip/README.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+## Usage
+
+```html
+<button class="mdc-fab material-icons" aria-label="Favorite">
+  <span class="mdc-fab__icon">
+    arrow_downward
+  </span>
+  <span class="mdc-tooltip">My Tooltip</span>
```
developerdizzle: Since tooltips aren't supposed to contain rich content/images, I wonder if it would make sense to use the `title` attribute instead?  I know the MDL library uses your same technique, though.

j-o-d-o: Do you mean something like this:
```html
<span class="mdc-fab__icon" data-mdc-tooltip="My Tooltip">
  arrow_downward
</span>
```
and then using CSS to create the tooltip? If so, there has been a short discussion about it here: https://github.com/material-components/material-components-web/issues/1741

If not, could you maybe show a short example of what you have in mind?

developerdizzle: Hey @j-o-d-o, I had something like this in mind:

moog16:
@j-o-d-o sorry that we dropped the ball on this one. Would you please update this PR with master? I'd like to work with you on getting this into master. Let me know if you're still interested. 

Just to let you know it could be a long process getting an entirely new component into master. I scroll through briefly. It looks like the Readme is not currently up to our standard: https://github.com/material-components/material-components-web/blob/master/docs/code/readme_standards.md

moog16:
@j-o-d-o just checked it out and fixed the conflicts. A few things:

1. In the screenshot the tooltip is positioned off screen
1. The animation in is sliding, but in the spec it is a fade in/out

screenshot from 2018-07-03 10-15-10

j-o-d-o:
Exchanged the slide animation with the fade in/out + scale in/out, did some cleanup as well as styled everything according to spec. Additionally I added an option for a delay not only to hide it as it states in the spec but also in the beginning (which is set to 0ms per default).
Next up:
1) Detect offscreen position and move tooltip to other side
2) If user interacts with the controlling element (e.g. by click), hide the tooltip

Things I don't understand in the spec:
It says 

> "Continuously display the tooltip as long as the user long-presses or hovers over the element."

and then

> "Display the tooltip for 1.5 seconds."

I decided to go with the second one and hide the tooltips after 1.5 seconds of hovering.
Also, what is the correct media query for "Mobile" when the tooltips are supposed to have a different size?

moog16:
NICE! I'll take a look in a bit. 

> "Continuously display the tooltip as long as the user long-presses or hovers over the element."

They mean if a user hovers over the anchoring element, it should continuously show the tooltip.

RE: 1.5 seconds. I think they are saying there should be 1.5s delay on the tooltip hiding after hover or long press. Although from the videos in the spec show it disappearing faster. It could be a sped up video, but I will double check for you.

packages/mdc-tooltip/README.md
```diff
@@ -1,4 +1,144 @@
...
+tooltip.show();
+```
+
+## Usage within Web Frameworks
+
+If you are using a JavaScript framework, such as React or Angular, you can create a Tooltip for your framework. Depending on your needs, you can use the _Simple Approach: Wrapping MDC Web Vanilla Components_, or the _Advanced Approach: Using Foundations and Adapters_. Please follow the instructions [here](../../docs/integrating-into-frameworks.md).
+
+### `MDCTolltipAdapter`
+
+The `root` element is the tooltip element e.g. `<span class='mdc-tooltip'>tooltip</span>`. The `controller` is the element the tooltip is for. Therefore it "controls" the tooltip when and where the tooltips should display and hide.
```
moog16:
```
The `controller` is the element the tooltip is for. Therefore it "controls" the tooltip when and where the tooltips should display and hide.
``` 

--> 

```
The `controller` is the element the tooltip anchors to. It "controls" the tooltips positioning and display.
```

packages/mdc-tooltip/_variables.scss
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "@material/theme/color-palette";
+
+$mdc-tooltip-background: $material-color-grey-700 !default;
```
moog16: remove this in favor of sass mixins

packages/mdc-tooltip/_variables.scss
```diff
@@ -0,0 +1,22 @@
...
+$mdc-tooltip-transition-length-show: 150ms !default;
+$mdc-tooltip-transition-length-hide: 75ms !default;
+$mdc-tooltip-mobil: 750px !default;
```
moog16: `$mdc-tooltip-mobile`

Where did you get 750px?

packages/mdc-tooltip/adapter.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
moog16: 2018

packages/mdc-tooltip/_variables.scss
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright 2017 Google Inc. All Rights Reserved.
```
moog16: 2018

packages/mdc-tooltip/adapter.js
```diff
@@ -0,0 +1,85 @@
...
+   * @param {string} propertyName
+   * @param {string} value
+   */
+  setStyle(propertyName, value) { }
+
+  /**
+   * @param {string} evtType
+   * @param {!Function} handler
+   */
+  registerListener(evtType, handler) { }
```
moog16: remove the `registerlistener` and `deregisterlistener` methods. We now attach/remove events in the component level since its easier for React and other 3rd party libraries to use.

See list for an example
https://github.com/material-components/material-components-web/blob/master/packages/mdc-list/index.js#L55

packages/mdc-tooltip/constants.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
moog16: 2018

packages/mdc-tooltip/constants.js
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @enum {string} */
+const cssClasses = {
+  ROOT: 'mdc-tooltip',
+  SHOW: 'mdc-tooltip--show',
+  HIDE: 'mdc-tooltip--hide',
```
moog16: instead of having a --show and a --hide class, can we default to hiding and remove the --hide class?

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
moog16: 2018

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+      removeClass: (/* className: string */) => {},
+      getClassList: () => /* [classNames: string] */ [],
+      getRootWidth: () => /* type: number */ 0,
+      getRootHeight: () => /* type: number */ 0,
+      getControllerWidth: () => /* type: number */ 0,
+      getControllerHeight: () => /* type: number */ 0,
+      getControllerOffsetTop: () => /* type: number */ 0,
+      getControllerOffsetLeft: () => /* type: number */ 0,
+      setStyle: (/* propertyName: string, value: string */) => {},
+      registerListener: (/* type: string, handler: EventListener */) => {},
```
moog16: remove this and put logic into component level.

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+      left = ctrlOffsetLeft - tooltipWidth - this.gap;
+    } else {
+      top = ctrlOffsetTop + ctrlHeight + this.gap;
+    }
+
+    this.adapter_.setStyle('top', top.toString() + 'px');
+    this.adapter_.setStyle('left', left.toString() + 'px');
+  }
+
+  addEventListeners_() {
```
moog16: move this all to component layer

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+    this.adapter_.registerListener('mouseenter', this.showDelayed_.bind(this));
+    this.adapter_.registerListener('focus', this.showDelayed_.bind(this));
+    this.adapter_.registerListener('touchstart', this.showDelayed_.bind(this));
+    this.adapter_.registerListener('mouseleave', this.hide_.bind(this));
+    this.adapter_.registerListener('blur', this.hide_.bind(this));
+    this.adapter_.registerListener('touchend', this.hide_.bind(this));
+  }
+
+  showDelayed_() {
+    this.showTimeout_ = setTimeout(() => {
```
moog16: clearTimeout in destroy method

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+      this.show_();
+    }, this.showDelay)
+  }
+
+  show_() {;
+    this.setDirection_();
+    this.displayed_ = true;
+    this.adapter_.removeClass(cssClasses.HIDE);
+    this.adapter_.addClass(cssClasses.SHOW);
+    this.hideTimeout_ = setTimeout(() => {
```
moog16: clearTimeout in destroy method

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+  showDelayed_() {
+    this.showTimeout_ = setTimeout(() => {
+      this.show_();
+    }, this.showDelay)
+  }
+
+  show_() {;
+    this.setDirection_();
+    this.displayed_ = true;
+    this.adapter_.removeClass(cssClasses.HIDE);
```
moog16: I think you can get rid of the HIDE class

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  hide_() {
+    if(this.displayed_) {
+      this.adapter_.removeClass(cssClasses.SHOW);
+      this.adapter_.addClass(cssClasses.HIDE);
+    }
+    this.displayed_ = false;
+
+    if(this.showTimeout_ != null){
```
moog16: !==

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+      clearTimeout(this.showTimeout_);
+      this.showTimeout_ = null;
+    }
+    if(this.hideTimeout_ != null){
```
moog16: !==

packages/mdc-tooltip/foundation.js
```diff
@@ -0,0 +1,173 @@
...
+    const tooltipWidth = this.adapter_.getRootWidth();
+    const ctrlOffsetTop = this.adapter_.getControllerOffsetTop();
+    const ctrlOffsetLeft = this.adapter_.getControllerOffsetLeft();
+    const ctrlHeight = this.adapter_.getControllerHeight();
+    const ctrlWidth = this.adapter_.getControllerWidth();
+
+    let top = ctrlOffsetTop + ctrlHeight / 2 - tooltipHeight / 2;
+    let left = ctrlOffsetLeft + ctrlWidth / 2 - tooltipWidth / 2;
+
+    if (this.direction_ === 'top') {
```
moog16: Still I'm seeing the tooltip being cutoff in smaller screen resolutions
screenshot from 2018-07-06 16-12-32

It looks like there is enough space to the right, but its still cut off. I'm not sure what the correct behavior in this case would be. Maybe the gap should decrease? I'll have to check with design.

packages/mdc-tooltip/index.js
```diff
@@ -0,0 +1,131 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
moog16: 2018

packages/mdc-tooltip/index.js
```diff
@@ -0,0 +1,131 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import MDCComponent from '@material/base/component';
+import MDCTooltipFoundation from './foundation';
+// import MDCRippleAdapter from './adapter';
```
moog16: delete

packages/mdc-tooltip/index.js
```diff
@@ -0,0 +1,131 @@
...
+import {getCorrectEventName} from '@material/animation/index';
```
moog16: delete

packages/mdc-tooltip/index.js
```diff
@@ -0,0 +1,131 @@
...
+   * @return {!MDCTooltipAdapter}
+   */
+  static createAdapter(instance) {
+    return {
+      addClass: (className) => instance.root_.classList.add(className),
+      removeClass: (className) => instance.root_.classList.remove(className),
+      getClassList: () => instance.root_.classList,
+      getRootWidth: () => instance.root_.offsetWidth,
+      getRootHeight: () => instance.root_.offsetHeight,
+      getControllerWidth: () => instance.controller_.offsetWidth,
```
moog16: i saw you removed getBoundingClientRect. Why is that?

packages/mdc-tooltip/mdc-tooltip.scss
```diff
@@ -0,0 +1,101 @@
+//
+// Copyright 2017 Google Inc. All Rights Reserved.
```
moog16: 2018

packages/mdc-tooltip/index.js
j-o-d-o: Because getClientBoundingRect() considers the transform also which is used to scale the tooltip up and down.

packages/mdc-tooltip/_variables.scss
j-o-d-o: Just random. Didn't know what the right condition for "Mobile" from the specs is. 
>Also, what is the correct media query for "Mobile" when the tooltips are supposed to have a different size?

packages/mdc-tooltip/foundation.js
j-o-d-o: I have not yet implemented the detection if a tooltip is moving outside of the viewport.

package.json
```diff
@@ -187,7 +187,8 @@
           "typography",
           "demos",
           "infrastructure",
-          "package"
+          "package",
+          "tooltip"
         ],
```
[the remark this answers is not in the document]
j-o-d-o: Yep, running node v8.11.1 and npm v5.6.0. I will update that.

packages/mdc-tooltip/constants.js
j-o-d-o: The issue I had with that, was that the tooltips where just blinking up briefly on page load. 
I did change it to only having a show class and as a workaround I added an "animation" class which adds the animation for the default mode (hidden mode) after the first show animation. 
If you have a more elegant solution to this, let me know.

moog16: I see - that sounds reasonable.

packages/mdc-tooltip/index.js
j-o-d-o: I actually had to add it again for the controlling element in order to probably calculate the position in case any of the parent elements is scrolling. I added this case to the example page to quickly test that usecase.
I now move all of the tooltips directly under body and make them fixed position in order to guarantee that there are no corner cases where the position might be off.

j-o-d-o:
New update which fixes corner cases in positioning as well as adding the feature to change direction in case it the tooltip would move out of the viewport.

Need to fix unit tests as well as some linting, will do that the following days.

moog16:
I'll rereview when you fix the tests and are ready. Please ping me in this PR once ready.

moog16:
@j-o-d-o talked to design about the 1.5 seconds. We believe that was a mistake on their end, which they are correcting in the spec. What they mean to say is to debounce the tooltip from appearing after 1.5s. This makes sense, because if you move your mouse around the page with multiple tooltips, you would not want all of them to appear/reappear.

Does that answer you question?


------------------------------------------------------------------------------------------------------------------------------
Other question you had regarding auto-positioning. They said we shouldn't really design for that. Tooltips should short and concise, and should never flow off screen. The end developer/designer should place the tooltip where it will always appear on the screen (not cropped off the page). Sorry for the confusion on my part.

j-o-d-o:
Okay, just to verify that I get it right. If a signal is debounced for 1.5 seconds, it will result in a delay of 1.5s from user e.g. hovering the controlling element to actually showing the tooltip.

moog16:
@j-o-d-o that's correct. The user will need to hover over the control element in order for the tooltip to appear

j-o-d-o:
Some updates regarding the recent discussions. Unittests, Code Coverage and linting should be fixed as well as the latest changes we decided on. 

I will be away for the next 3 weeks (without a laptop). So we will need to put this on hold for the time being or at least I will not be able to contribute in the meantime.

moog16:
@j-o-d-o Thanks for the update. If you're going to be out that long we might continue to push changes and try to merge it. If we do that I'll try to make sure that you get the credit :)

moog16:
@j-o-d-o can you add me to this repo so I can continue to push to this? Or better yet the MDC-Web team?

j-o-d-o:
@moog16 I am not sure how to add a whole team. But I added a view from the MDC team that where active recently as collaborators to the repo. Of course feel free to change anything as you see fit.

moog16:
Thanks @j-o-d-o...I think myself should be good enough.

j-o-d-o:
I am not really sure how to proceed with this. Let me know in case there are any updates.

j-o-d-o:
Cleaning up some of my github PRs and Issues. Closing this as it is already too old. Probably better to start with a new PR from scratch or taking only bits and pieces from this one.

PR #1703, opened 2017-12-06
fix(demos): add back button to header on drawer demos page

moog16:
Fixes https://www.pivotaltracker.com/story/show/148981415

Screenshot of fix below:

screen shot 2017-12-06 at 10 58 25


demos/drawer/persistent-drawer.html
```diff
@@ -95,6 +95,9 @@
       <header class="mdc-toolbar mdc-elevation--z4">
         <div class="mdc-toolbar__row">
           <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
+            <span class="catalog-back">
```
bwobrien: Nit: .catalog-back has a padding-right which I think is a bit too much here, given there's another toolbar button. Can you remove the padding here?

moog16: Yes, I agree. Will remove

Issue #1608, opened 2017-11-17
Textarea border is partially cut by the floating label


ricardobrandao:
### What MDC-Web Version are you using?

> 0.25.0 (except for base that is 0.24.0

### What browser(s) is this bug affecting?

> Chrome 62.0.3202.94 (64 bits)

### What OS are you using?

> MacOS 10.13.1

### What are the steps to reproduce the bug?

> Please write the steps which need to be taken in order to reproduce the bug. These steps should be
> as detailed as possible, e.g.
>
> 1. Go to [this url](https://codepen.io/ricardobrandao/pen/xPPymW).
> 2. Start typing on the textarea
> 3. Observe the component's behavior: the top border is partially overlapped by the label.

### What is the expected behavior?

> The top border should stay the same

### What is the actual behavior?

> The top border is partially cut near the label

### Any other information you believe would be useful?

<img width="291" alt="screen shot 2017-11-17 at 15 37 41" src="https://user-images.githubusercontent.com/2292745/32955384-b8e93692-cbad-11e7-9aeb-ae3220eab6cc.png">

amsheehan:
Looks like a bug. Putting in tracker.

mellogarrett:
Textarea for Safari (Mobile & Desktop) is even worse ☹️ 

<img width="746" alt="screen shot 2017-12-04 at 12 35 25 pm" src="https://user-images.githubusercontent.com/9908831/33566985-b56b68b0-d8ef-11e7-8a07-a0a08b7f8bf1.png">

amsheehan:
We have found that zoom levels and pixel densities of monitors have had a detrimental effect on these types of features. @mellogarrett Thank you for the screen shot, could you also provide the specs for the machine(s) you're using?

mellogarrett:
Seen on both 2017 MacBook Pro & iPhone 8+

moog16:
While looking at the dev tools, it seems like the <textarea /> and the `mdc-text-field` wrapper both have a border. The <textarea /> element does not have border-radius, and since the label is position: absolute, it overlaps the border. 

Proposed solution is to remove the textarea border. Another is to remove all the classes from the mdc-text-field, but it then not be consistent with the text field.

moog16:
@ricardobrandao In the MDC demo normalize.css is being used, but is turned off in your codepen. I noticed that margin is added by default on both Chrome and Safari. Safari (v11.0.1) however adds by default to the textfield.

```
margin-top: 0em;
[...]

PR #1715, opened 2017-12-07
fix(textfield): Fix textarea label from overlapping border.

moog16:
Removed textarea border to prevent double borders.
Added line-height to textarea label.

Fixes https://github.com/material-components/material-components-web/issues/1608
https://www.pivotaltracker.com/story/show/153034989


packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -445,7 +445,6 @@
   .mdc-text-field__input {
     padding: $padding-inset;
     padding-top: $padding-inset * 2;
-    border: 1px solid transparent;
```
havenchyk: I'm not 100% sure, but in active state textarea should have `2px` border and if you just remove this border, the whole border will become `1px`

screencast 2017-12-08 15-39-06

havenchyk:
@moog16 can transparent background solve the initial problem instead of line-height? https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/mdc-text-field.scss#L479

kfranqueiro:
I believe the reason for the non-transparent background is so that the label is still readable in cases where the text entered into the textarea overlaps it due to overflow/scrolling.

moog16:
@

So a few things are being addressed in this PR:

1. I updated the label background to be transparent as opposed to white. This is what was causing the overlap with the border as well as some text in situations where zoom was used.
1. There was another issue when moving the fullwidth textarea handle horizontally, and the outer element would not collapse/expand with the textarea element. I fixed this with the `resize; vertical;` property.
1. The 2nd issue would also come up when it was not fullwidth. The textarea wrapper element needed the css property `width: fit-content`, which allows for the outer div to take the width of the children.
1. I also added line-height to the label, which was also used in normalize.css. There is a conflict with Safari's default <label /> line-height that caused the overlap in the top border.

PR #1716, opened 2017-12-07
feat(list): Rename elements to match spec; don't set size of meta

### LINKED ISSUE #1644: Add support in List for text-based end details
As described here https://material.io/guidelines/components/lists.html#lists-content (check image on the right side)
There should be a possibility for adding not only icon, but also text-based end-detail for list items.

At this moment as the width of the `.mdc-list-item__end-detail` set to 24px it's not possible without overriding the property to add for example timestamp content here.

acdvorak:
BREAKING CHANGE: `__start-detail` has been renamed to `__graphic`, and `__end-detail` has been renamed to `__meta`. In addition, meta data tiles no longer have a default width/height (fixes #1644).

Also:
- Format mdc-list README
- Capitalize headings in mdc-list README and demo


packages/mdc-list/README.md
```diff
@@ -105,103 +108,103 @@ While in theory you can add any number of "lines" to a list item, you can use th
...
+### List "Tile" Elements
 
-As mentioned in the spec, lists can contain primary and secondary actions. It can also contain
-things such as avatars, icons, interactive controls, etc. We call these items *detail* items. Lists
-can contain 1 **start detail** item and/or 1 **end detail** item that are positioned at the start
-and end of the list, respectively. These items are correctly flipped in RTL contexts.
+As mentioned in the spec, lists can contain primary and secondary actions. They can also contain
+things such as avatars, icons, interactive controls, etc. We call all of these items *tiles*. Lists
+can contain 1 **supporting graphic** tile and/or 1 **meta data** tile that are positioned at the start
+and end of the list, respectively. These tiles are correctly flipped in RTL contexts.
```
kfranqueiro: "of the list" should be "of the list item", right?

acdvorak: Done

demos/list.html
```diff
@@ -116,47 +116,47 @@
         <section>
           <h2>Custom Colors</h2>
           <section>
-            <h3>Example - Two-line Lists, Avatars, end detail, inset dividers</h3>
+            <h3>Example - Two-Line Lists, Avatars, Meta Data, Inset Dividers</h3>
```
kfranqueiro: If we're making metadata one word now then we need to update these instances in the demo page.

acdvorak: Done

PR #1720, opened 2017-12-07
feat(toolbar): Add theme color mixins

### LINKED ISSUE #1154: Add Sass color mixin to toolbar
- [x] Create color mixins
- [x] Delete any —primary --accent modifiers

acdvorak:
This PR also makes the demo page a little prettier.

Fixes #1154


demos/common.scss
```diff
@@ -34,20 +35,19 @@ $mdc-theme-secondary: $material-color-light-green-a700 !default;
 @import "../packages/mdc-toolbar/mdc-toolbar";
 @import "../packages/mdc-typography/mdc-typography";
 
+$dark-button-color: $material-color-light-green-a200;
```
moog16: is this for dark-theme? Is this change needed?

moog16: Nevermind - I see you just moved it up

demos/toolbar/default-flexible-toolbar.html
```diff
@@ -25,36 +25,11 @@
...
-        color: red;
-        background-color: rgba(0, 0, 0, 0.8);
-      }
     </style>
   </head>
-  <body class="mdc-typography mdc-toolbar-demo">
-    <header class="mdc-toolbar mdc-toolbar--flexible mdc-toolbar--flexible-default-behavior">
-      <div class="mdc-toolbar__row">
+  <body class="mdc-typography">
+    <header class="mdc-toolbar mdc-toolbar--flexible mdc-toolbar--flexible-default-behavior demo-toolbar">
```
moog16: should this have the class demo-toolbar--flexible to be consistent with the custom toolbar?

moog16: on second thought...its just the demo so nevermind

demos/toolbar/waterfall-flexible-toolbar.html
```diff
@@ -80,7 +53,7 @@
         </p>
       </div>
     </main>
-    <footer>
+    <footer class="demo-toolbar-floating-footer">
```
moog16: I think it would be more BEM-ish to be `demo-toolbar__floating-footer` or `demo-toolbar__footer--floating` if we are trying to be BEM

demos/toolbar/waterfall-flexible-toolbar-custom-style.html
```diff
@@ -80,7 +60,7 @@
         </p>
       </div>
     </main>
-    <footer>
+    <footer class="demo-toolbar-floating-footer">
```
moog16: `demo-toolbar__floating-footer` or `demo-toolbar__footer--floating` comment from below

demos/toolbar/waterfall-toolbar-fix-last-row.html
```diff
@@ -88,7 +63,7 @@
         </p>
       </div>
     </main>
-    <footer>
+    <footer class="demo-toolbar-floating-footer">
```
moog16: change here too if we decide to go BEM

packages/mdc-toolbar/_mixins.scss
```diff
@@ -14,14 +14,38 @@
...
+
+@mixin mdc-toolbar-ink-color($color) {
+  @include mdc-theme-prop(color, $color);
+}
+
+@mixin mdc-toolbar-fill-color($color) {
+  @include mdc-theme-prop(background-color, $color);
+}
+
+@mixin mdc-toolbar-fill-color-accessible($color) {
```
moog16: Confused what it means to be accessible regarding color. Is this like high contrast mode? If so, that seems to me like another theme (like dark-theme).

demos/toolbar/waterfall-flexible-toolbar-custom-style.html
acdvorak: If it were a child of `demo-toolbar`, yes, absolutely! In this particular case, though, it's a sibling of `demo-toolbar`, so it gets its own separate block class.

demos/toolbar/waterfall-flexible-toolbar.html
acdvorak: Ditto above

demos/toolbar/waterfall-toolbar-fix-last-row.html
acdvorak: Ditto above

packages/mdc-toolbar/_mixins.scss
acdvorak: Kind of, but not quite. It's not a separate "mode", per se, but simply a convenience mixin for clients who want their button's fill (bg) color to be X, but don't care what the ink (text) color is as long as it's high enough contrast to pass accessibility requirements
([4.5:1 contrast ratio](https://webaim.org/resources/contrastchecker/) or higher).

Assuming that the client has accessible values for `primary` in `$mdc-theme-text-colors`, this mixin guarantees that the ink color it chooses will be accessible on the given fill color.

moog16: That makes sense...should I also be doing this for drawer?

acdvorak: Yes - generally speaking, all `-fill-color` mixins should also have a `-fill-color-accessible` version that sets the ink color automatically.

PR #1729, opened 2017-12-07
chore(text-field): Export subcomponents from root index.js

### LINKED ISSUE #1714: Bug: mdc-textfield's root `index.js` file does not export the foundations or default implementations of its sub-components
Currently (as of `v0.26.0`), `mdc-textfield`'s root `index.js` file does not export any of the following:

- `MDCTextFieldBottomLine`
- `MDCTextFieldBottomLineFoundation`
- `MDCTextFieldHelperText`
- `MDCTextFieldHelperTextFoundation`

Imports for these symbols were added to the root `index.js` as part of #1684, however they were not subsequently exported.

By not exporting these symbols from the root `index.js`, we provide no way for consumers to access them from `dist/mdc.textfield.js` bundle. Consumers are instead forced to access them via their ES6 modules, which can be problematic for some (see #929).

bonniezhou:
Fixes #1714.

### DIFF
--- a/packages/mdc-textfield/index.js
+++ b/packages/mdc-textfield/index.js
@@ -231,4 +231,6 @@ class MDCTextField extends MDCComponent {
   }
 }
 
-export {MDCTextField, MDCTextFieldFoundation};
+export {MDCTextField, MDCTextFieldFoundation,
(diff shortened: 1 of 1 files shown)


tyronedougherty:
Cheers for this! Any idea when this will make it into a release on the NPM repo? (current is `0.26.0`)

Issue #1145, opened 2017-08-17
Add Sass color mixin to drawer


lynnjepsen:
- [ ] Create color mixins
- [ ] Call color mixin from stylesheet, with primary color
- [ ] Update README to document color mixins

acdvorak:
Duplicate of #943

Issue #943, opened 2017-07-13
Developers should be able to override MDC Drawer with any background color


lynnjepsen:
Came up in #902 . The design guidance is that MDC Drawer can be any color. Right now our code has it set to the Theme "background" color. We need to make it possible for developers to override the MDC Drawer background color with any color.

acdvorak:
We need 3 mixins:

`mdc-drawer-fill-color($color)`
`mdc-drawer-ink-color($color)`
`mdc-drawer-scrim-color($color, $opacity)`

PR #1730, opened 2017-12-07
feat(drawer): custom sass mixins for color, background, scrim

moog16:
Fixes:
https://github.com/material-components/material-components-web/issues/1145
https://github.com/material-components/material-components-web/issues/943


packages/mdc-drawer/_mixins.scss
```diff
@@ -147,3 +147,15 @@
   @include mdc-drawer-header_;
   @include mdc-drawer-list_;
 }
+
+@mixin mdc-drawer-fill-color($color) {
+  background-color: $color;
```
acdvorak: We typically use the `mdc-theme-prop` mixin for these component theme mixins, because it supports our special variable names like `primary`, `secondary-light`, `text-hint-on-dark`, etc. in addition to color literals like `#aaccdd`.

packages/mdc-drawer/_mixins.scss
```diff
@@ -147,3 +147,15 @@
...
+}
+
+@mixin mdc-drawer-ink-color($color) {
+  color: $color;
+}
+
+@mixin mdc-drawer-scrim-color($color, $opacity) {
+  background-color: rgba($color, $opacity);
```
acdvorak: You'll need to use the `mdc-theme-prop-value` function here to resolve special aliases like `primary` to their literal color values.

Since the `rgba` Sass function returns a literal color (and not an alias), you _don't_ need to use the `mdc-theme-prop` mixin here - just the similarly-named function to get the value.

Does that make sense?

packages/mdc-drawer/_variables.scss
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "@material/theme/variables";
+
+$mdc-temporary-drawer-background: $mdc-theme-background !default;
```
acdvorak: Rename to `$mdc-temporary-drawer-fill-color` to match spec

packages/mdc-drawer/_variables.scss
```diff
@@ -0,0 +1,30 @@
...
+$mdc-temporary-drawer-ink-color: map-get($mdc-theme-property-values, text-primary-on-primary) !default;
```
acdvorak: Wouldn't this be `text-primary-on-background`?

Ditto below

packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
```diff
@@ -45,15 +46,16 @@
...
+    @include mdc-drawer-scrim-color($mdc-temporary-drawer-scrim-color, $mdc-temporary-drawer-scrim-opacity);
     opacity: 0;
     opacity: var(--mdc-temporary-drawer-opacity, 0);
     content: "";
     will-change: opacity;
   }
 
   &__drawer {
-    @include mdc-theme-prop(background, background);
+    @include mdc-drawer-ink-color($mdc-temporary-drawer-ink-color);
```
acdvorak: We've generally been `@include`-ing the theme mixins on the root element rather than child elements, and then putting the child element selectors in the mixins.

That approach allows clients to keep all (or most) of their theme customizations in one place.

packages/mdc-drawer/_mixins.scss
moog16: Think so...something like:

```
background-color: rgba(@include mdc-theme-prop-value($color), $opacity);
```

acdvorak: Really close:

```scss
background-color: rgba(mdc-theme-prop-value($color), $opacity);
```

In Sass, `@include` is used when you call a `@mixin` like `mdc-theme-prop`.

A `@function` like `mdc-theme-prop-value`, on the other hand, doesn't get `@include`.

packages/mdc-drawer/_mixins.scss
moog16: fixed

packages/mdc-drawer/_mixins.scss
moog16: fixed

packages/mdc-drawer/_variables.scss
moog16: fixed

packages/mdc-drawer/_variables.scss
moog16: fixed

packages/mdc-drawer/temporary/mdc-temporary-drawer.scss
moog16: fixed

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -51,14 +52,18 @@
...
-      .mdc-permanent-drawer {
+      .mdc-drawer--permanent {
         z-index: 2;
       }
 
       .demo-main {
         padding-left: 16px;
       }
 
+      .mdc-demo__control {
```
acdvorak: Let's rename this to `demo-form-field`.

`mdc-form-field` is a special RTL-aware class that wraps a checkbox, radio, or switch, plus its adjacent `<label>`.

And our new convention is to prefix CSS class names for demo pages with `demo-`.

_(I generally tend to just `s/^mdc-/demo-/` the class name I'm customizing for the demo 🙂 )_

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -122,6 +127,17 @@
       <main class="demo-main">
         <h1 class="mdc-typography--display1">Permanent Drawer</h1>
         <p class="mdc-typography--body1">It sits to the left of this content.</p>
+
+        <div class="mdc-demo__control">
```
acdvorak: Add `mdc-form-field` class, and make sure it's `@import`-ed by `demos/drawer.scss`

Ditto for other usages

demos/drawer/persistent-drawer.html
```diff
@@ -52,16 +53,21 @@
 
       .demo-main {
         padding-left: 16px;
+        padding-bottom: 16px;
+      }
+
+      .demo-main__control {
```
acdvorak: 1. Rename to `demo-switch`
2. Move this class to `demos/drawer/drawer.scss`
3. File a follow-up chore in Pivotal to consolidate and move existing common/shared styles out of the `<style>` blocks in each individual drawer demo page into `demos/drawer/drawer.scss`

Ditto for other files

demos/drawer/permanent-drawer-below-toolbar.html
```diff
@@ -135,6 +152,16 @@ <h1 class="mdc-typography--display1">Permanent Drawer</h1>
       document.querySelector('#toggle-tall').addEventListener('click', function() {
         extraTall.style.display = extraTall.style.display ? '' : 'none';
       });
+
+      var drawerEl = document.querySelector('.mdc-drawer');
```
acdvorak: For future-proofing, I would probably give both these elements an `id` and select on that.

E.g., `id="demo-drawer"` and `id="demo-theme-switch"`

Ditto for other files

packages/mdc-drawer/_mixins.scss
```diff
@@ -20,7 +20,7 @@
 @import "@material/typography/mixins";
 
 @mixin mdc-drawer-toolbar-spacer_() {
-  &__toolbar-spacer {
+  .mdc-drawer__toolbar-spacer {
```
acdvorak: 👍

packages/mdc-drawer/_mixins.scss
```diff
@@ -149,13 +149,22 @@
...
+  .mdc-drawer__drawer {
+    @include mdc-theme-prop(background-color, $color);
+  }
 }
 
 @mixin mdc-drawer-ink-color($color) {
-  color: $color;
+  @include mdc-theme-prop(color, $color);
+
+  .mdc-list-item {
```
acdvorak: Let's remove this rule block. There's already a separate set of `mdc-list-item-*` mixins, so our clients (and our demos) should just use those.

(We generally try to keep each component as isolated as possible to make them easier to maintain.)

packages/mdc-drawer/persistent/mdc-persistent-drawer.scss
```diff
@@ -28,14 +28,13 @@ $mdc-persistent-drawer-dark-theme-bg-color: #212121 !default;
...
   @include mdc-drawer-base_;
-
+  @include mdc-drawer-ink-color($mdc-persistent-drawer-ink-color);
+  @include mdc-drawer-fill-color($mdc-persistent-drawer-fill-color);
   width: 0;
 
-  &__drawer {
-    @include mdc-drawer-ink-color($mdc-persistent-drawer-ink-color);
-    @include mdc-drawer-fill-color($mdc-persistent-drawer-background);
+  .mdc-drawer__drawer {
```
acdvorak: 👍

acdvorak:
Nit: `feat` in the PR title should be lowercase 🙂

demos/drawer/permanent-drawer-above-toolbar.html
kfranqueiro: I think you mean `demo-form-field`... and maybe we should pick a class name that isn't conflatable with a component it has nothing to do with?

demos/drawer/drawer.scss
```diff
@@ -14,7 +14,15 @@
...
 //
 
+@import "@material/list/mixins";
 @import "../common";
 @import "../../packages/mdc-drawer/mdc-drawer";
+@import "../../packages/mdc-drawer/mixins";
 @import "../../packages/mdc-elevation/mdc-elevation";
 @import "../../packages/mdc-list/mdc-list";
+
+.demo-drawer.mdc-drawer {
```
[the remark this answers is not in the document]
moog16: Yup you're totally correct. For some reason I saw something different last night. Removing .mdc-foo

demos/drawer/permanent-drawer-above-toolbar.html
moog16: gotcha makes sense

demos/drawer/permanent-drawer-above-toolbar.html
moog16: Ok, so rule of thumb is if a component is followed by a <label /> it should be wrapped in mdc-form-field class?

moog16:
also adding mdc-drawer-fill-color-accessible($color)

demos/drawer/drawer.scss
```diff
@@ -14,7 +14,24 @@
...
 @import "../common";
 @import "../../packages/mdc-drawer/mdc-drawer";
+@import "../../packages/mdc-drawer/mixins";
 @import "../../packages/mdc-elevation/mdc-elevation";
 @import "../../packages/mdc-list/mdc-list";
+
+.demo-drawer {
+  @include mdc-drawer-fill-color(#3f51b5);
+  @include mdc-drawer-ink-color(#fff);
+  @include mdc-list-item-graphic-ink-color(#fff);
```
acdvorak: We should probably call the `primary-text-ink-color` mixin too
(it needs to be called on the `.mdc-list-item` element, btw;
otherwise it will overwrite the value set by `mdc-drawer-fill-color`.)

So we probably want something like:

```scss
.demo-drawer--custom {
  @include mdc-drawer-fill-color($material-color-teal-900);
  @include mdc-drawer-ink-color(#fff);

  .demo-drawer-list-item {
    @include mdc-list-item-graphic-ink-color(text-icon-on-dark);
    @include mdc-list-item-primary-text-ink-color(text-secondary-on-dark);
  }

  .demo-drawer-list-item.mdc-list-item--selected {
    @include mdc-list-item-graphic-ink-color(text-primary-on-dark);
    @include mdc-list-item-primary-text-ink-color(text-primary-on-dark);
  }
}
```

demos/drawer/drawer.scss
```diff
@@ -14,7 +14,24 @@
...
+}
+
+.demo-form-field {
```
acdvorak: Add this class name in the HTML

demos/drawer/drawer.scss
```diff
@@ -14,7 +14,24 @@
...
+  margin: 16px 0;
+}
+
+.demo-switch {
+  margin-top: 16px;
```
acdvorak: This looks weird - can we remove it?

demos/drawer/drawer.scss
```diff
@@ -14,7 +14,24 @@
 // limitations under the License.
 //
 
+@import "@material/list/mixins";
+@import "@material/form-field/mdc-form-field";
```
acdvorak: For demo pages, we use relative paths in our Sass `@import`s (demo pages don't need to be distributed via npm).

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -73,11 +74,11 @@
...
   </head>
   <body class="mdc-typography demo-body">
-    <nav class="mdc-permanent-drawer">
-      <div class="mdc-permanent-drawer__toolbar-spacer"></div>
+    <nav id="demo-theme-switch" class="mdc-drawer mdc-drawer--permanent">
+      <div class="mdc-drawer__toolbar-spacer"></div>
       <div class="mdc-list-group">
         <nav class="mdc-list">
-          <a class="mdc-list-item mdc-permanent-drawer--selected" href="#">
+          <a class="mdc-list-item mdc-drawer--selected" href="#">
```
acdvorak: `mdc-drawer--selected` should be `mdc-list-item--selected`

Ditto in other files

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -122,6 +123,17 @@
...
         <p class="mdc-typography--body1">It sits to the left of this content.</p>
+
+        <div class="mdc-form-field demo-form-field">
+          <div class="mdc-switch demo-switch">
+            <input type="checkbox" id="basic-switch" class="mdc-switch__native-control" />
+            <div class="mdc-switch__background">
+              <div class="mdc-switch__knob"></div>
+            </div>
+          </div>
+          <label for="basic-switch" class="mdc-switch-label">Theme off/on</label>
```
acdvorak: Reword label text to "Custom theme off/on"

Ditto in other files

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -144,6 +156,16 @@ <h1 class="mdc-typography--display1">Permanent Drawer</h1>
       document.querySelector('#toggle-tall').addEventListener('click', function() {
         extraTall.style.display = extraTall.style.display ? '' : 'none';
       });
+
+      var drawerEl = document.querySelector('#demo-theme-switch');
+      var switchEl = document.querySelector('.mdc-switch');
+      switchEl.addEventListener('change', function(e) {
+        if(e.target.checked) {
+          drawerEl.classList.add('demo-drawer');
```
acdvorak: `'demo-drawer--custom'`

Ditto in other files

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -144,6 +156,16 @@ <h1 class="mdc-typography--display1">Permanent Drawer</h1>
       document.querySelector('#toggle-tall').addEventListener('click', function() {
         extraTall.style.display = extraTall.style.display ? '' : 'none';
       });
+
+      var drawerEl = document.querySelector('#demo-theme-switch');
+      var switchEl = document.querySelector('.mdc-switch');
```
acdvorak: Let's use an ID instead of a class name

Ditto in other files

demos/drawer/permanent-drawer-below-toolbar.html
```diff
@@ -78,10 +79,10 @@
     </header>
 
     <div class="demo-content mdc-toolbar-fixed-adjust">
-      <nav class="mdc-permanent-drawer">
+      <nav id="demo-theme-switch" class="mdc-drawer mdc-drawer--permanent">
```
acdvorak: Rename `demo-theme-switch` to `demo-drawer`

Ditto in other files

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -144,6 +156,16 @@ <h1 class="mdc-typography--display1">Permanent Drawer</h1>
       document.querySelector('#toggle-tall').addEventListener('click', function() {
         extraTall.style.display = extraTall.style.display ? '' : 'none';
       });
+
+      var drawerEl = document.querySelector('#demo-theme-switch');
```
acdvorak: This should probably be `#demo-drawer`

Ditto in other files

packages/mdc-drawer/persistent/mdc-persistent-drawer.scss
```diff
@@ -27,13 +28,13 @@ $mdc-persistent-drawer-dark-theme-bg-color: #212121 !default;
   --mdc-persistent-drawer-dark-theme-bg-color: #{$mdc-persistent-drawer-dark-theme-bg-color};
 }
 
-.mdc-persistent-drawer {
+.mdc-drawer--persistent {
   @include mdc-drawer-base_;
-
+  @include mdc-drawer-ink-color($mdc-persistent-drawer-ink-color);
+  @include mdc-drawer-fill-color($mdc-persistent-drawer-fill-color);
   width: 0;
```
acdvorak: Nit: While we're here, add a newline before `width` to make the linter happy

demos/drawer/drawer.scss
```diff
@@ -14,7 +14,24 @@
...
 @import "../common";
 @import "../../packages/mdc-drawer/mdc-drawer";
+@import "../../packages/mdc-drawer/mixins";
 @import "../../packages/mdc-elevation/mdc-elevation";
 @import "../../packages/mdc-list/mdc-list";
+
+.demo-drawer {
```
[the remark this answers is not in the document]
moog16: Initially it seemed strange to have a class with no styles on it though, but yup that makes sense as it is more BEM like.

demos/drawer/drawer.scss
moog16: ah ok - ya initially I had this in the mixin, but then we took it out. After removing I should have just put it in the demo classes. Makes sense.

moog16: Should I also do the same changes on `.demo-drawer--accessible`? If so, should I also create mdc-list accessible mixins?

demos/drawer/drawer.scss
moog16: ah forgot to remove

demos/drawer/permanent-drawer-above-toolbar.html
moog16: don't know what I was thinking there...changing

demos/drawer/permanent-drawer-above-toolbar.html
acdvorak: @kfranqueiro Yes, I believe that's right.
@moog16 Yes, we should always use `mdc-form-field` when we need to add a label next to a radio, checkbox, or switch

demos/drawer/permanent-drawer-above-toolbar.html
```diff
@@ -122,6 +123,40 @@
       <main class="demo-main">
         <h1 class="mdc-typography--display1">Permanent Drawer</h1>
         <p class="mdc-typography--body1">It sits to the left of this content.</p>
+
+        <div id="demo-radio-buttons">
+          <div class="mdc-form-field">
+            <div class="mdc-radio">
+              <input class="mdc-radio__native-control" type="radio" id="default" name="theme" checked>
```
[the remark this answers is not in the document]
kfranqueiro: I think `theme-radio-default` would be good enough? This is a demo page so `demo-` in IDs seems redundant, and it's a demo for drawer so ditto for `drawer-`.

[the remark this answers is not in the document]
acdvorak: Fair enough - `theme-radio-default` SGTM 🙂  I've been using `demo-` prefixes for CSS class names to make it clear that they're not library classes, but we don't really use IDs in our lib, so I guess there's no need to worry about potential overlap.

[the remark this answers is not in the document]
moog16: Ok will update!

PR #1733, opened 2017-12-08
fix(checkbox): Respect BEM when outputting the base stylesheet

lynnjepsen:
Also update our demo pages from `.mdc-checkbox.demo-checkbox--custom-all` to .demo-checkbox--custom-all`, since we don't need to be overly specific with our CSS selectors


packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -23,18 +23,21 @@
...
   @include mdc-states($mdc-checkbox-baseline-theme-color);
   @include mdc-ripple-radius($mdc-checkbox-ripple-radius);
-  @include mdc-checkbox-container-colors();
 
   @include mdc-theme-dark {
     @include mdc-checkbox-container-colors($unmarked-stroke-color: $mdc-checkbox-border-color-dark);
   }
 }
 
+@at-root {
```
acdvorak: I'll be darned! I thought you could only `@include` mixins in Sass if you were already inside a CSS selector (even if the mixin itself has a CSS selector in it).

But I guess Sass treats `@at-root` like a selector, so ... awesome!

packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -23,18 +23,21 @@
...
+  @include mdc-checkbox-ink-color($mdc-checkbox-mark-color);
+  @include mdc-checkbox-focus-indicator-color($mdc-checkbox-baseline-theme-color);
+  @include mdc-checkbox-container-colors();
```
acdvorak: While we're here, I think we normally omit empty parens on mixins

packages/mdc-checkbox/mdc-checkbox.scss
lynnjepsen: WEeeee!

PR #1736, opened 2017-12-08
chore(ripple): move common ripple styles out of mixins and into mdc-ripple

lynnjepsen:
When we pull all the components into material-components-web.scss, we need to de-dupe the common ripple styles. We don't want material-components-web.css to have unnecessary duplicate styles.

Right now we do this by de-duping within ripple/_mixins.scss. But I would like to keep ripple/_mixins.scss purely for mixins. I would like to move the de-dupe logic from ripple/_mixins.scss to ripple/mdc-ripple.scss


lynnjepsen:
Yep, will add

> BREAKING CHANGE: Please update all components which use MDC Ripple to import the new @material/ripple/common file

packages/mdc-ripple/common.scss
```diff
@@ -0,0 +1,46 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "./keyframes";
+@import "./mixins";
```
kfranqueiro: I don't think `mixins` and `variables` are needed here?

Issue #1560, opened 2017-11-07
Update MDC List to use mdc-states-color


lynnjepsen:
Use the mdc-states-color mixin created in #1556

MDC List already uses mdc-ripple-color mixin:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-list/mdc-list.scss#L158

this should be replaced with the new mdc-states-color mixin, with any relevant opacity tweaks

kfranqueiro:
Note that list presently only applies ripple styles for upgraded ripples with JS component instances. As part of states work, we will want to apply states styles to _all_ list items by default, with an optional modifier class for opting out of this.

However, there is an issue with list / menu / drawer styles: currently, menu and drawer nuke the padding that `mdc-list` normally applies, which causes problems due to the styles applied for ripples to counteract that padding. (I'm currently under the impression that menu and drawer are applying these styles to properly support separators that extend to both edges of the list, but separators are technically part of mdc-list along with mdc-list-group, so we should probably seek to fix this in mdc-list itself instead.)

kfranqueiro:
Resolved by #1737 + #1738 + #1739.

Issue #1455, opened 2017-10-19
Ensure all components specify box-sizing to avoid bugs due to external styles


kfranqueiro:
We could probably manually test this by throwing in a `* { box-sizing: ... }` rule (set to content-box and border-box) in `demos/common.scss` to see if anything obviously breaks.

It looks like several components already account for box-sizing but I'll leave everything unchecked here until we actually test.

* [ ] mdc-button
* [ ] mdc-card
* [ ] mdc-checkbox
* [ ] mdc-dialog
* [ ] mdc-drawer
* [ ] mdc-fab
* [ ] mdc-form-field
* [ ] mdc-grid-list
* [ ] mdc-icon-toggle
* [ ] mdc-layout-grid
* [ ] mdc-linear-progress
* [ ] mdc-list
* [ ] mdc-menu
* [ ] mdc-radio
* [ ] mdc-ripple
* [ ] mdc-select
* [ ] mdc-slider
* [ ] mdc-switch
* [ ] mdc-tabs
* [ ] mdc-textfield
* [ ] mdc-toolbar

RobertSmits:
I have checked all demo pages for errors relating to box-sizing.
Now these are fixed but there are still two "errors".
- mdc-list: 
The demo list is a bit smaller due to css specific to the demo page.
- mdc-dialog:
I have not done any work on this but when setting ``* { box-sizing: border-box }`` the body of the scrollable dialog gets a little smaller. Reason for not jet fixing this is #1442 

What should I do with these two?

kfranqueiro:
Thanks for looking into this.

RE mdc-list, I wouldn't be particularly worried about the hero styles in the demo page, but switching to border-box does affect how the max-width on all lists is applied. However, I'd have to check whether the max-width we set is arbitrary, since I don't see mention of width in the guidelines.

I'm curious if there's a reason you chose border-box for mdc-list in your branch, whereas you chose content-box for the others? Wouldn't using content-box for mdc-list preserve its current behavior? (Using border-box necessitates the extra `calc` usage as well which could be a little harder to follow than the current alternative.)

RE dialog, I think applying border-box will end up affecting how the height of the scrolling container behaves, since it also has padding. Granted, that height is already arbitrary so this might not matter, but I'd have the same question as above RE choosing border-box vs. preserving the current content-box behavior?

kfranqueiro:
[...]

PR #1737, opened 2017-12-08
feat(list): Use states mixins; change padding behavior to support them

kfranqueiro:
Refs #1560. There will be separate PRs to update/clean up menu and drawer styles which depend on list.

Also refs #1455 since these updates resolve discrepancies when box-sizing is specified on all elements.

Also updates demo index page DOM and styles to accommodate new padding behavior and make entire rows clickable to match states' shade region.

BREAKING CHANGE: List padding is now per-item rather than across the entire list. Separators now span the entire list width by default, with the addition of a mdc-list-divider--padded modifier class to achieve the old default behavior.


demos/index.scss
```diff
@@ -17,16 +17,10 @@
 @import "./common";
 @import "../packages/mdc-list/mdc-list";
 
-.catalog-list {
-  padding-left: 28px;
-}
-.catalog-list-icon {
-  margin-right: 24px;
+.catalog-list a {
```
acdvorak: Can we possibly rename this to `.demo-catalog-list`?

I prefer that convention because it makes it obvious that the CSS class is _only_ used
by a demo page, and not for any other purpose (as opposed to, say, `.material-icons`).

demos/index.scss
```diff
@@ -17,16 +17,10 @@
...
+  color: #212121;
 }
 
-ul.catalog-list {
-  a {
-    text-decoration: none;
-    color: #212121;
-  }
+.catalog-list-icon {
```
acdvorak: Ditto

packages/mdc-list/README.md
```diff
@@ -256,20 +262,11 @@ To use within lists, simply add the `mdc-list-divider` class to a list item.
...
-<ul class="mdc-list">
-  <li class="mdc-list-item">Item 1 - Division 1</li>
-  <li class="mdc-list-item">Item 2 - Division 1</li>
-  <li class="mdc-list-item">Item 3 - Division 1</li>
-  <li role="separator" class="mdc-list-divider mdc-list-divider--inset"></li>
-  <li class="mdc-list-item">Item 1 - Division 2</li>
-  <li class="mdc-list-item">Item 1 - Division 2</li>
-</ul>
-```
+Dividers are designed to span the full width of the list by default (especially useful in the context of drawers and menus).
```
acdvorak: 👍

packages/mdc-list/mdc-list.scss
```diff
@@ -163,25 +164,27 @@ a.mdc-list-item {
...
-  padding: 0 $mdc-list-side-padding;
   overflow: hidden;
 
   &:focus {
     outline: none;
   }
+}
+
+// List items should support states by default, but it should be possible to opt out.
+:not(.mdc-list--non-interactive) > .mdc-list-item {
```
acdvorak: Is the direct child selector (`>`) necessary to prevent the styles from leaking
into nested child `.mdc-list-item` elements?

packages/mdc-list/mdc-list.scss
```diff
@@ -195,12 +198,19 @@ a.mdc-list-item {
   border-bottom-style: solid;
 }
 
+.mdc-list-divider--padded {
```
acdvorak: Nit: to me, the word `padded` in the class name implies `padding`, but the implementation actually uses `margin` to achieve the effect.

Do you think that difference might confuse or surprise people?

packages/mdc-list/README.md
```diff
@@ -80,6 +80,12 @@ attached to the root element, or the element has an ancestor with class `mdc-the
 </html>
 ```
 
+#### Non-interactive lists
+
+By default, list items receive styles for hover, focus, and press states (including the ripple effect if
```
acdvorak: 👍

packages/mdc-list/README.md
```diff
@@ -80,6 +80,12 @@ attached to the root element, or the element has an ancestor with class `mdc-the
...
+`MDCRipple` is instantiated). It is possible to opt out of these styles by adding the `mdc-list--non-interactive`
```
acdvorak: Should we mention that the focus styles only apply if the element is one that
naturally receives focus (`<a href="">`, `<button>`, etc.) or has `tabindex="0"`?
Or do we think most clients figure that out?

packages/mdc-list/mdc-list.scss
kfranqueiro: Without the direct child selector, this selector always matches, because while the direct parent might have the `non-interactive` modifier applied, there are plenty of ancestors that won't.

packages/mdc-list/mdc-list.scss
kfranqueiro: Can we think of a better word to use? Usually I'm against using terms that can seem contradictory in the source language, but I'm not sure what else works here. `inset` is already used for a different purpose.

demos/index.html
```diff
@@ -37,216 +37,216 @@
     </header>
     <main>
       <nav class="mdc-toolbar-fixed-adjust">
-        <ul class="catalog-list mdc-list mdc-list--two-line">
-          <li class="mdc-list-item">
+        <div class="catalog-list mdc-list mdc-list--two-line">
```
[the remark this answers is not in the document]
amsheehan: I also wonder if this will effect page rank at all. Is the crawler smart enough to parse all this without the help of a site map or something. Can someone more enlightened chime in?

packages/mdc-list/README.md
kfranqueiro: I'll add a comment.

demos/index.html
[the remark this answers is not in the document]
kfranqueiro: My main concern as to why I did this was to make the entire lines clickable because the hover state will make users think the whole line is clickable, when in fact right now it's not. If we're concerned about the markup, what I can do instead is revert this and simply set the parent to `non-interactive` so that this UX issue doesn't occur.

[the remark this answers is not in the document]
amsheehan: https://html.spec.whatwg.org/multipage/sections.html#the-nav-element

Seems like `<nav>` is enough for user agents, but no mention of anything else in the living standard.

packages/mdc-list/mdc-list.scss
kfranqueiro: It also occurs to me that although in this particular case `margin` is used, the purpose of the `--padded` modifier is to match the `padding` that is applied to individual list items. We need to use `margin` here since the intent here is to restrict the border.

kfranqueiro:
It has just come to my attention while testing that there may be some unintended interactions between list styles and select styles...

image

I'll try to figure out what's going on here. Ideally I should fix it before merging this.

demos/index.html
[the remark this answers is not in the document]
acdvorak: For now, maybe we should just add `role="list"` and `role="listitem"` and call it a day.

For the record, I totally understand why you made this change, and I _love_ the UX improvement it provides. I just want to make sure we don't accidentally break accessibility or SEO 🙂 

The change in DOM structure _might_ affect PageRank, but I'm not certain. I've generally seen `<nav>` elements contain a `<ul><li>`, which suggests it's a standard practice for some reason, but I don't know if it's for SEO or accessibility (or both).

From doing a few quick searches, I don't think our SEO can get any worse - Google Search doesn't appear to display any sitelinks for our demo site (likely because we have so many different navigation destinations), and the material.io catalog site seems to get ranked higher. So I'm not _too_ worried about SEO impact.

Once I get a Windows machine, I can definitively test the accessibility aspect in JAWS, but for now, I suspect adding the `role` attributes ought to suffice.

packages/mdc-list/mdc-list.scss
acdvorak: I see - let's keep `--padded` then.

acdvorak:
Ouch! That's a nasty bug. I was able to repro it in Chrome 62 on macOS Sierra (10.12.6), but not in Safari 11.0.1 or Firefox 57.0.1, so I suspect you're right about it being a Chrome issue.

It looks like this PR also affects the padding of CSS-only multiple-select:

pr 1737 select css-only indentation

kfranqueiro:
Yes, this PR affects the multiselect demo padding in Chrome only, because Chrome does some inexplicably hacky nonsense involving using `::before` pseudo-elements' `content` on `option` elements within `optgroup`s to add indentation, but our ripple/states styles set `content` to `''`.

No other browser does what Chrome does. Safari also indents `option`s inside `optgroups`, but by other means, and I have to wonder why Chrome didn't simply use padding or something. Some other browsers don't indent options in `optgroup`s at all to begin with. So I'm not sure it's something worth fixating on.

demos/index.html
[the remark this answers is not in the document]
kfranqueiro: When I saw this conversation unfolding I had reverted the new DOM structure, but I've put it back and added roles.

kfranqueiro:
For posterity: Andy also noticed yesterday that Firefox was getting a horizontal scrollbar in the Select demo page with this update. It turns out that Firefox seems to ignore `position: relative` on option elements, causing our ripple/shade pseudo-elements to span the entire body and then some:

image

Between this issue and others that have already existed (namely, Chrome completely ignoring our background styles for focused selected option in multiselect and using the OS selection color instead), I'm going to nuke the before/after content for multiselect option elements.

Issue #1600, opened 2017-11-15
Add outline subelement, and wire it into parent text field adapter


lynnjepsen:
Add outline subelement, and wire it into parent text field adapter

... also add a demo of the outlined text field.

bonniezhou:
Resolved by #1749

PR #1749, opened 2017-12-12
feat(text-field): Add outline subelement and demo for outlined text field

bonniezhou:
Add an outline subcomponent and wire it into the parent component for the outlined variation of MDC Text Field. It uses an SVG path when the text field is focused and a div border when it's idle.
The demo of outlined text field includes:
- disabled
- RTL
- usage with helper text

These will come in later PRs:
- ripple
- leading/trailing icons
- dense variant
- CSS-only variation

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -14,6 +14,10 @@
 // limitations under the License.
 //
 
+@mixin mdc-text-field-outlined-corner-radius($radius) {
+  border-radius: $radius $radius $radius $radius;
+}
+
 @mixin mdc-text-field-box-corner-radius($radius) {
   border-radius: $radius $radius 0 0;
 }
@@ -45,7 +49,7 @@
 
     33% {
       animation-timing-function: cubic-bezier(.5, 0, .701732, .495819);
-      transform: translateX(10px) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(5px) translateY(-#{$positionY}) scale(.75, .75);
     }
 
     66% {
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -15,7 +15,7 @@
 //
 
 @mixin mdc-text-field-outlined-corner-radius($radius) {
-  border-radius: $radius $radius $radius $radius;
+  border-radius: $radius;
 }
 
 @mixin mdc-text-field-box-corner-radius($radius) {
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -58,9 +58,6 @@ class MDCTextFieldFoundation extends MDCFoundation {
       registerBottomLineEventHandler: () => {},
       deregisterBottomLineEventHandler: () => {},
       getNativeInput: () => {},
-      getWidth: () => {},
-      getHeight: () => {},
-      getLabelWidth: () => {},
       getIdleOutlineStyleValue: () => {},
       isRtl: () => {},
     });
@@ -154,13 +151,13 @@ class MDCTextFieldFoundation extends MDCFoundation {
    * Updates the focus outline for outlined text fields.
    */
   updateOutline() {
-    if (!this.outline_) {
+    if (!this.outline_ || !this.label_) {
       return;
     }
-    // The label is scaled 75% when it floats above the text field.
-    const labelWidth = this.adapter_.getLabelWidth() * 0.75;
+    const labelWidth = this.label_.getFloatingWidth();
+    // Fall back to reading a specific corner's style because Firefox doesn't report the style on border-radius.
     const radiusStyleValue = this.adapter_.getIdleOutlineStyleValue('border-radius') ||
-      /* Firefox requires specifying the corner. */ this.adapter_.getIdleOutlineStyleValue('border-top-left-radius');
+      this.adapter_.getIdleOutlineStyleValue('border-top-left-radius');
     const radius = parseFloat(radiusStyleValue);
     const isRtl = this.adapter_.isRtl();
     this.outline_.updateSvgPath(labelWidth, radius, isRtl);
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -162,7 +162,7 @@ class MDCTextFieldFoundation extends MDCFoundation {
     // The label is scaled 75% when it floats above the text field.
     const labelWidth = this.adapter_.getLabelWidth() * 0.75;
     const radiusStyleValue = this.adapter_.getIdleOutlineStyleValue('border-radius') ||
-      this.adapter_.getIdleOutlineStyleValue('border-top-left-radius') /* Firefox requires specifying the corner. */;
+      /* Firefox requires specifying the corner. */ this.adapter_.getIdleOutlineStyleValue('border-top-left-radius');
     const radius = parseFloat(radiusStyleValue);
     const isRtl = this.adapter_.isRtl();
     this.outline_.updateSvgPath(width, height, labelWidth, radius, isRtl);
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -172,7 +172,9 @@ class MDCTextFieldFoundation extends MDCFoundation {
     const height = this.adapter_.getHeight();
     // The label is scaled 75% when it floats above the text field.
     const labelWidth = this.adapter_.getLabelWidth() * 0.75;
-    const radius = parseFloat(this.adapter_.getIdleOutlineStyleValue('border-radius'));
+    const radiusStyleValue = this.adapter_.getIdleOutlineStyleValue('border-radius') ||
+      this.adapter_.getIdleOutlineStyleValue('border-top-left-radius') /* Firefox requires specifying the corner. */;
+    const radius = parseFloat(radiusStyleValue);
     const isRtl = this.adapter_.isRtl();
     this.outline_.updateSvgPath(width, height, labelWidth, radius, isRtl);
   }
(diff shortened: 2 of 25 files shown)


bonniezhou:
Resolves #1600.

packages/mdc-textfield/index.js
```diff
@@ -188,6 +214,19 @@ class MDCTextField extends MDCComponent {
             this.bottomLine_.unlisten(evtType, handler);
           }
         },
+        getWidth: () => this.root_.offsetWidth,
+        getHeight: () => this.root_.offsetHeight,
+        getLabelWidth: () => {
+          const labelElement = this.root_.querySelector(strings.LABEL_SELECTOR);
```
kfranqueiro: Is this safe? We treat label as optional overall...but I suppose it's required for outlined text fields?

(I can't actually remember in what cases label isn't provided, off the top of my head... Full-width?)

I'm also wondering whether this belongs in the label subcomponent's adapter, though that might not be possible without making it more onerous to use from this module.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -115,6 +117,102 @@
...
+
+  &.mdc-text-field--focused .mdc-text-field__idle-outline {
+    opacity: 0;
+  }
+
+  &.mdc-text-field--focused .mdc-text-field__outline {
+    opacity: 1;
+  }
+
+  .mdc-text-field__label--float-above ~ .mdc-text-field__idle-outline {
```
kfranqueiro: These can be combined with the blocks above (i.e. group the 0's and the 1's)

packages/mdc-textfield/outline/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+##### MDCTextFieldOutline.foundation
+
+MDCTextFieldOutlineFoundation. This allows the parent MDCTextField component to access the public methods on the MDCTextFieldOutlineFoundation class.
+
+### Using the foundation class
+
+Method Signature | Description
+--- | ---
+setOutlinePathAttr(value: string) => void | Sets the SVG path of the outline element
```
kfranqueiro: Should this specifically say it sets the `d` attribute of the path within the outline SVG?

packages/mdc-textfield/foundation.js
```diff
@@ -142,6 +150,24 @@ class MDCTextFieldFoundation extends MDCFoundation {
     this.receivedUserInput_ = true;
   }
 
+  /**
+   * Updates the focus outline for outlined text fields.
+   */
+  updateOutline() {
```
kfranqueiro: It bugs me that all of this logic has to be exposed at the top level, though I realize that it's probably impossible to isolate everything reasonably within the outline subcomponent (e.g. getting label size).

One thing I'm curious about is whether it'd be sufficient to get the width/height of the svg element rather than the root element, and if at least that part of this could be isolated within outline so you have less to compute and pass from here? I'd also suspect that `getIdleOutlineStyleValue` could be moved within the subcomponent too.

It's quite possible I'm not thinking this all the way through and don't know what you already tried, so let me know if this doesn't seem at all feasible.

packages/mdc-textfield/foundation.js
```diff
@@ -142,6 +150,24 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
+    if (!this.outline_) {
+      return;
+    }
+    const width = this.adapter_.getWidth();
+    const height = this.adapter_.getHeight();
+    // The label is scaled 75% when it floats above the text field.
+    const labelWidth = this.adapter_.getLabelWidth() * 0.75;
+    const radiusStyleValue = this.adapter_.getIdleOutlineStyleValue('border-radius') ||
+      /* Firefox requires specifying the corner. */ this.adapter_.getIdleOutlineStyleValue('border-top-left-radius');
```
kfranqueiro: Maybe move this to a line comment above this statement to make the code flow better? e.g. // Fall back to reading a specific corner's style because Firefox doesn't report the style on border-radius"

packages/mdc-textfield/outline/foundation.js
```diff
@@ -0,0 +1,90 @@
...
+   * @param {number} labelWidth
+   * @param {number} radius
+   * @param {boolean=} isRtl
+   */
+  updateSvgPath(width, height, labelWidth, radius, isRtl = false) {
+    let path;
+    if (!isRtl) {
+      path = 'M' + (radius + 2.1 + Math.abs(10 - radius) + labelWidth + 8) + ',' + 1
+        + 'h' + (width - (2 * (radius + 2.1)) - labelWidth - 8.5 - Math.abs(10 - radius))
+        + 'a' + radius + ',' + radius + ' 0 0 1 ' + radius + ',' + radius
```
kfranqueiro: If I'm reading this right, the middle 7 lines here are identical in both if/else? Maybe store that part in a common variable to make it clearer what's actually different?

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -115,6 +117,102 @@
   // stylelint-enable plugin/selector-bem-pattern
 }
 
+.mdc-text-field--outlined {
+  height: 56px;
+  border: none;
+
+  .mdc-text-field__outline {
```
kfranqueiro: Isn't this redundant, and can't this be moved into the outline subcomponent's scss? i.e. you're not going to have this element within anything besides a `.mdc-text-field--outlined` so this doesn't need to be nested within that selector.

test/unit/mdc-textfield/mdc-text-field-outline.foundation.test.js
```diff
@@ -0,0 +1,77 @@
...
+
+test('defaultAdapter returns a complete adapter implementation', () => {
+  verifyDefaultAdapter(MDCTextFieldOutlineFoundation, [
+    'setOutlinePathAttr',
+  ]);
+});
+
+const setupTest = () => setupFoundationTest(MDCTextFieldOutlineFoundation);
+
+test('#updateSvgPath sets the path of the outline element when isRtl=false by default', () => {
```
kfranqueiro: I'm not sure how useful it is to verify the exact argument passed to setOutlinePathAttr, given that it basically requires carbon-copying the implementation and is thus a change-detector test.

packages/mdc-textfield/_mixins.scss
```diff
@@ -45,7 +49,7 @@
 
     33% {
       animation-timing-function: cubic-bezier(.5, 0, .701732, .495819);
-      transform: translateX(10px) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(5px) translateY(-#{$positionY}) scale(.75, .75);
```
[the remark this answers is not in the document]
bonniezhou: Outlined requires a smaller value because otherwise the label would overlap the outline, and I think it's better to keep the values the same if possible. I don't think this is redlined so... 5px for all text fields looks good to me. Thoughts?

packages/mdc-textfield/foundation.js
bonniezhou: Good idea to move getting width/height into the outline subcomponent, I'll update that.
However `getIdleOutlineStyleValue` can't be moved to the subcomponent because the `idle-outline` element is outside of the subcomponent's root `outline` element, same with `labelWidth` and `isRtl`.

packages/mdc-textfield/index.js
bonniezhou: Hmm yeah it makes more sense to get the label width from the label's adapter, even if it means it'll have to be passed to the label foundation, then the main foundation, then the outline foundation.

packages/mdc-textfield/mdc-text-field.scss
bonniezhou: Whoops, accidentally added ripple stuff here. Will remove and put into the next PR to keep this one clean, and yep I agree it should've been in the subcomponent's scss instead of here anyways.

test/unit/mdc-textfield/mdc-text-field-outline.foundation.test.js
bonniezhou: Yeah I agree...maybe just confirming that it sets the outline from the adapter is as much as we can test.

packages/mdc-textfield/_mixins.scss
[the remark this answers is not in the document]
amsheehan: That's probably fine, I would make a quick screen grab of it in Quicktime to send to Jonas for approval. He is the owner of Text Field animations.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -115,6 +117,97 @@
...
+
+      transform: scale(.75) translateY(-170%);
+
+      &.mdc-text-field__label--shake {
+        animation: invalid-shake-float-above-outlined 250ms 1;
+      }
+    }
+  }
+
+  &.mdc-text-field--focused {
```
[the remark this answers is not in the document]
bonniezhou: Ohh I see, this does look better. Done.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -116,6 +118,87 @@
...
+      transform: scale(.75) translateY(-170%);
+
+      &.mdc-text-field__label--shake {
+        animation: invalid-shake-float-above-outlined 250ms 1;
+      }
+    }
+  }
+
+  &.mdc-text-field--focused .mdc-text-field__idle-outline, 
+  .mdc-text-field__label--float-above ~ .mdc-text-field__idle-outline{
```
kfranqueiro: Missing space before `{`

bonniezhou:
I considered that too but decided that being able to access more logic from within the subcomponent isn't worth the complexity of adding yet another DOM element to group the outlines, especially if like you said we'd still have to go to the outer component for label.

PR #1752, opened 2017-12-13
fix(snackbar): Apply visibility hidden when at rest. fixes #1258

### LINKED ISSUE #1258: Possible to see snackbars in hidden state on any platform with rubber-band scrolling at edges

rpatil26:


rpatil26:
@amsheehan Can you please check this?

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -69,6 +69,10 @@
     }
   }
 
+  &[aria-hidden] {
```
amsheehan: Why not select on `.mdc-snackbar` and `.mdc-snackbar--active` for this?

rpatil26:
Thanks @amsheehan for the suggestion.

I had tried that already, but show or hide animation was getting messed up. To keep it sync, we need to apply CSS transition (on opacity may be because visibility animation is little tricky) and make sure the timing is in sync. Alternately we can also use JS and wait for transition end to alter the visibility. I feel it is increasing code size and introducing unnecessary complexity. Currently aria hidden is on transition end already and to me that looked the right place to hook it.

Trying to understand your rationale and thoughts, why the current one might not be a good idea?

rpatil26:
Thanks Alex. Yes, it's good to specifically target aria-hidden `true` to make it work reliably. Done.

rpatil26:
Hi @amsheehan, can you please merge it if it looks good now.

packages/mdc-snackbar/mdc-snackbar.scss
rpatil26: Done!

rpatil26: Changes done as discussed!

PR #1756, opened 2017-12-13
fix(menu): Close menu even when a list-item was clicked.

### LINKED ISSUE #1747: Select menu does not close when clicking

0.27.0

Crome 62.0.3202.94 and any browser

### What OS are you using?

Windows 10

1. Go to https://codepen.io/YuoMamoru/full/rpNxMB
2. Click a "mdc-select" in the page
3. Click the list under "Steps to Reproduce the Bug"

If you click the point on "li" tag with class="mdc-list-item", the menu should close.

If you click the point on "li" tag with class="mdc-list-item", the menu does not close.

I think that it is an issue not to check whether the element clicked is part of its own menu in handleDocumentClick_(evt) on mdc-menu/simple/foundation.js.

simono:
Previously clicks on a list-item in the page would leave the menu open. Now we check if a item in this menu was clicked and leave the menu open.

Fixes #1747 

@amsheehan You introduced this behavior in 02fe795139. Is my fix correct?

### DIFF
--- a/packages/mdc-menu/README.md
+++ b/packages/mdc-menu/README.md
@@ -172,7 +172,6 @@ Method Signature | Description
 `hasClass(className: string) => boolean` | Returns a boolean indicating whether the root element has a given class.
 `hasNecessaryDom() => boolean` | Returns boolean indicating whether the necessary DOM is present (namely, the `mdc-menu__items` container).
 `getAttributeForEventTarget(target: EventTarget, attributeName: string) => string` | Returns the value of a given attribute on an event target.
-`eventTargetHasClass: (target: EventTarget, className: string) => boolean` | Returns true if the event target has a given class.
 `getInnerDimensions() => {width: number, height: number}` | Returns an object with the items container width and height.
 `hasAnchor: () => boolean` | Returns whether the menu has an anchor for positioning.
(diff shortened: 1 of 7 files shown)


packages/mdc-menu/simple/foundation.js
```diff
@@ -282,7 +281,7 @@ class MDCSimpleMenuFoundation extends MDCFoundation {
     let el = evt.target;
 
     while (el && el !== document.documentElement) {
-      if (this.adapter_.eventTargetHasClass(el, cssClasses.LIST_ITEM)) {
+      if (this.adapter_.getIndexForEventTarget(el) >= 0) {
```
patrickrodee: Small nitpick: I'd suggest updating the JSDoc description to explain that it cancels if it's not a child item instead of just a list item.

patrickrodee: Also, perhaps just personal preference, but I prefer checks to be as explicit as possible so it might be a good idea to change the comparator from `>= 0` to `!== -1`

simono:
@patrickrodee Thanks for the Review. I updated the PR.

simono:
@patrickrodee I rebase my branch on master. Anything else I can do to get this merged? Thanks!

simono:
@williamernest I rebased my branch once again and it should apply cleanly now. Let me know if there's anything I can help out with. Thanks!

PR #1757, opened 2017-12-13
fix(select): Work around glitch with new list styles in Chrome

kfranqueiro:
This resolves the Chrome-specific select menu glitch I discovered while testing #1737 (see screenshot in https://github.com/material-components/material-components-web/pull/1737#issuecomment-351105800).

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -263,6 +263,7 @@ $mdc-select-menu-transition: transform 180ms $mdc-animation-standard-curve-timin
     }
 
     // Styles to override mdc-ripple-radius from mdc-list, which causes display glitches in Chrome
+    // (See https://github.com/material-components/material-components-web/pull/1737#issuecomment-351105800)
     @include mdc-ripple-radius(50%);
 
     &::before,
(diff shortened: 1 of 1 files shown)


packages/mdc-select/mdc-select.scss
```diff
@@ -261,6 +261,14 @@ $mdc-select-menu-transition: transform 180ms $mdc-animation-standard-curve-timin
         @include mdc-theme-prop(color, text-primary-on-dark);
       }
     }
+
+    // Styles to override mdc-ripple-radius from mdc-list, which causes display glitches in Chrome
+    @include mdc-ripple-radius(50%);
+
+    &::before,
+    &::after {
+      border-radius: 0;
```
acdvorak: Is this needed? When I comment it out, the display glitch still seems to be fixed.

packages/mdc-select/mdc-select.scss
```diff
@@ -261,6 +261,14 @@ $mdc-select-menu-transition: transform 180ms $mdc-animation-standard-curve-timin
         @include mdc-theme-prop(color, text-primary-on-dark);
       }
     }
+
+    // Styles to override mdc-ripple-radius from mdc-list, which causes display glitches in Chrome
```
acdvorak: Can we add a bit more context to this comment? E.g.:

```
// Override the mdc-ripple-radius set by mdc-list. This fixes display glitches in Chrome 62+
// (see PR #1757 for more info).
```

packages/mdc-select/mdc-select.scss
acdvorak: Never mind, I'm dumb. This is definitely needed to fix the ripple on press.

Issue #1687, opened 2017-12-04
Respect BEM when outputting the base stylesheet


amsheehan:
Previously, I believe the specificity was warranted because of some other demo pages outside of the MDC Web catalog page. Since this is no longer the case, I think we should change it back to the demo selector.

Instead of `.mdc-radio.demo-radio--custom {}`, use `.demo-radio--custom {}`

Also if we're trying to use BEM here, we are using it incorrectly. For there to be a modifier class, we need a block or element to modify.

For instance, in `demos/radio.html` we have elements with:

`class="mdc-radio demo-radio--custom"`

This should read: `class="mdc-radio demo-radio demo-radio--custom"` to adhere to BEM style.

bonniezhou:
To the first point, yes we should be using only the demo selector in CSS.
To the second point, we decided that for just the demo pages, better readability is more important than strict adherence to BEM style.

lynnjepsen:
See #1733 for an example on how to fix the Sass mixins and demo pages. This PR is for checkbox only

moog16:
I will also remove the <style> tags from all the demo files and place into the .scss files. I think @acdvorak mentioned wanting to do that in #1730.

PR #1760, opened 2017-12-14
Chore/remove mdc demo classes

moog16:
addresses: https://github.com/material-components/material-components-web/issues/1687, which was to remove the `.mdc-*` naming convention.

This is only a portion of the components, but I wanted to get the ones I have into a PR before moving forward with the remainder.

Still remaining:
1. drawer
1. images
1. theme
1. toolbar
1. ripple


moog16:
Closing for smaller PRs

demos/dialog.html
```diff
@@ -166,17 +132,17 @@ <h2 id="mdc-dialog-with-list-label" class="mdc-dialog__header__title">
...
             </footer>
           </div>
           <div class="mdc-dialog__backdrop"></div>
         </aside>
       </div>
       <section class="example">
-        <button id="default-dialog-activation" class="mdc-button mdc-button--raised">Show Dialog</button>
-        <button id="colored-footer-button-dialog-activation" class="mdc-button mdc-button--raised">Show Colored Footer Button Dialog</button>
-        <button id="dialog-with-list-activation" class="mdc-button mdc-button--raised">Show Scrolling Dialog</button>
+        <button id="default-dialog-activation" class="mdc-button demo-button mdc-button--raised demo-button--raised">Show Dialog</button>
```
moog16: not sure how you feel about these classes, but needed them because of some class selectors in common.scss

demos/ripple.scss
```diff
@@ -17,3 +17,89 @@
...
+  left: 0;
+  top: 0;
+  width: 100%;
+}
+
+button.demo-surface {
+  display: inline-block; /* Safari does not center button text otherwise :( */
+  background: none;
+  border: none;
+  -webkit-appearance: none;
```
moog16: I think autoprefixer takes care of this stuff for us, correct? Will remove

demos/dialog.html
amsheehan: Yeah, that's fine. I would just order them so `demo-` is next to `demo-`

It's only a matter of style, but I (and Ken) feel this is more readable.

moog16:
Closing this since there are so many merge conflicts. Also will not be focusing on the demo css selectors, but instead ensuring that we do not have overly specific BEM selectors. See https://github.com/material-components/material-components-web/issues/1687

PR #1782, opened 2017-12-14
chore(slider): annotate for closure compiler

patrickrodee:
Create adapter.js under MDC Slider as an interface. Annotate index.js, foundation.js, adapter.js, and constants.js under MDC Slider for closure compiler. Update package.json closure whitelist to test MDC Slider.


packages/mdc-slider/adapter.js
```diff
@@ -0,0 +1,192 @@
...
+   * Removes a class from the slider Element
+   * @param {string} className
+   */
+  removeClass(className) {}
+
+  /**
+   * Returns a string if attribute name exists on the slider Element,
+   * otherwise returns null
+   * @param {string} name
+   * @return {?string}
```
lynnjepsen: Have Andy do a pass on your nulls and undefineds...He's more of an expert than me...and apparently these is a difference

patrickrodee: @acdvorak plz help me, i am uncertain :fearful:

packages/mdc-slider/adapter.js
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Removes all track markers fromt he track mark container element
+   */
+  removeTrackMarkers() {}
+
+  /**
+   * Sets a style property of the last track marker to the passed value
+   * @param {string} propertyName
+   * @param {string|number} value
```
acdvorak: I'm not sure if our internal JS compiler will complain about `string|number` not matching the type of `setProperty`'s `value` param... The JS extern definition for `setProperty` says `value` can only be a `string`.

I guess we can try it with `string|number` for now, and if the compiler complains, we can convert `value` to `String(value)` before passing it to `setProperty`.

packages/mdc-slider/adapter.js
```diff
@@ -0,0 +1,192 @@
...
+   */
+  setLastTrackMarkersStyleProperty(propertyName, value) {}
+
+  /**
+   * Returns true if RTL is enabled, false if not
```
acdvorak: `Returns true if the root element is RTL, otherwise false`

packages/mdc-slider/foundation.js
```diff
@@ -75,11 +84,16 @@ export default class MDCSliderFoundation extends MDCFoundation {
...
       setLastTrackMarkersStyleProperty: (/* propertyName: string, value: string */) => {},
       isRTL: () => /* boolean */ false,
-    };
+    });
   }
 
-  constructor(adapter = {}) {
+  /**
+   * Creates a new instance of MDCSliderFoundation
+   * @param {MDCSliderAdapter} adapter
```
acdvorak: Add explicit nullability qualifier

packages/mdc-slider/foundation.js
```diff
@@ -244,10 +268,16 @@ export default class MDCSliderFoundation extends MDCFoundation {
     }
   }
 
+  /**
+   * Creates a handler for mouse/touch/pointer down events
+   * @param {string} moveEvt
+   * @param {(function(!Event): !number)=} getPageX
```
acdvorak: Primitive values like `number` are non-null by default, so you can remove the `!` qualifier (ditto below)

packages/mdc-slider/foundation.js
```diff
@@ -273,18 +303,28 @@ export default class MDCSliderFoundation extends MDCFoundation {
...
+      this.setValueFromEvt_(evt, /** @type {function(!Event): !number} */ (getPageX));
     };
 
     return downHandler;
   }
 
+  /**
+   * Sets the slider value from an event
+   * @param {!Event} evt
+   * @param {!function(!Event): !number} getPageX
```
acdvorak: `function` types are non-null by default, so you can remove the `!` qualifier before `function`

packages/mdc-slider/foundation.js
```diff
@@ -297,6 +337,10 @@ export default class MDCSliderFoundation extends MDCFoundation {
     return min + pctComplete * (max - min);
   }
 
+  /**
+   * Handles keydown events
+   * @param {Event} evt
```
acdvorak: Add explicit nullability qualifier

packages/mdc-slider/foundation.js
```diff
@@ -311,6 +355,11 @@ export default class MDCSliderFoundation extends MDCFoundation {
     this.adapter_.notifyChange();
   }
 
+  /**
+   * Returns the computed name of the event
+   * @param {Event} kbdEvt
```
acdvorak: Add explicit nullability qualifier

packages/mdc-slider/index.js
```diff
@@ -17,51 +17,75 @@
...
+ * @extends MDCComponent<!MDCSliderFoundation>
+ */
+class MDCSlider extends MDCComponent {
   static attachTo(root) {
     return new MDCSlider(root);
   }
 
+  constructor(...args) {
+    super(...args);
+    /** @type {Element} */
```
acdvorak: Add explicit nullability qualifier (ditto below)

packages/mdc-slider/adapter.js
kfranqueiro: I would advocate tolerating only `string` (in fact that's how it's documented inline in the foundation). It looks like we only use this API once to set `flex` so we're currently getting away with implicitly coercing number to string, but I'd be +1 on wrapping the value we're passing it with `String(...)` rather than loosening the type here.

patrickrodee:
Updated from code review

packages/mdc-slider/foundation.js
```diff
@@ -177,7 +177,7 @@ class MDCSliderFoundation extends MDCFoundation {
       if (indivisible) {
         const lastStepRatio = (max - numMarkers * step) / step + 1;
         const flex = getCorrectPropertyName(window, 'flex');
-        this.adapter_.setLastTrackMarkersStyleProperty(flex, lastStepRatio);
+        this.adapter_.setLastTrackMarkersStyleProperty(flex, String(lastStepRatio));
```
acdvorak: It might make more sense to do the `String()` conversion _inside of_ `setLastTrackMarkersStyleProperty()` (right before it gets passed to `setProperty()`). That way, callers can pass a `string` _or_ a `number`, and `setLastTrackMarkersStyleProperty()` will deal with the conversion to make Closure Compiler happy.

packages/mdc-slider/foundation.js
```diff
@@ -339,7 +339,7 @@ class MDCSliderFoundation extends MDCFoundation {
 
   /**
    * Handles keydown events
-   * @param {Event} evt
+   * @param {?Event} evt
```
acdvorak: This param should probably be non-null, because we call `evt.preventDefault()` below without checking if `evt` is non-null.

packages/mdc-slider/foundation.js
```diff
@@ -357,7 +357,7 @@ class MDCSliderFoundation extends MDCFoundation {
 
   /**
    * Returns the computed name of the event
-   * @param {Event} kbdEvt
+   * @param {?Event} kbdEvt
```
acdvorak: The function body appears to assume this param is non-null, so we should mark it as such in the JSDoc.

packages/mdc-slider/foundation.js
kfranqueiro: That would require making an adapter API smarter, and we want our adapter APIs to be as simple and easy to implement as possible.

patrickrodee: In either case, Closure Compiler was happy. Even when run with the internal blaze compiler, it was fine with both the coerced string and the `string|number`.

packages/mdc-slider/foundation.js
patrickrodee: So make `kbdEvt` non-nullable, is that correct?

packages/mdc-slider/foundation.js
patrickrodee: Done

patrickrodee:
PR updated with review comments

lynnjepsen:
Resolves #1731

packages/mdc-slider/foundation.js
acdvorak: `@param {!Event} kbdEvt`

patrickrodee:
Updated from review

packages/mdc-slider/foundation.js
patrickrodee: Done

PR #1787, opened 2017-12-14
chore(demos): Use demo-* class names for custom styles instead of mdc-* in theme demo

moog16:
partial fix of https://github.com/material-components/material-components-web/issues/1687

### DIFF
--- a/demos/theme/index.html
+++ b/demos/theme/index.html
@@ -88,32 +88,32 @@
 
           <div>
             <h3 class="mdc-typography--body1">Default (primary theme color)</h3>
-            <button class="mdc-button">
+            <button class="mdc-button demo-button">
               Text
             </button>
-            <button class="mdc-button mdc-button--raised">
+            <button class="mdc-button demo-button mdc-button--raised">
               Raised
             </button>
-            <button class="mdc-button mdc-button--unelevated">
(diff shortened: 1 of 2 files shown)


demos/theme/index.scss
```diff
@@ -411,15 +387,15 @@ figure {
   @include mdc-states($demo-custom-color);
 }
 
-.mdc-ripple-surface[data-mdc-ripple-is-unbounded] {
+.de3mo-ripple-surface[data-mdc-ripple-is-unbounded] {
```
amsheehan: de3mo ~> demo

demos/theme/index.html
```diff
@@ -88,32 +88,32 @@ <h2 class="mdc-typography--headline demo-component-section__heading">
 
           <div>
             <h3 class="mdc-typography--body1">Default (primary theme color)</h3>
-            <button class="mdc-button">
+            <button class="mdc-button demo-button">
```
amsheehan: It is unnecessary to add `demo-*` as a namespace to any of these elements. We never require them for selector specificity in the `css`.

To clarify, just because we have `mdc-*`, doesn't mean we _always_ need `demo-*`. There are of course cases where we need a class specifically for the demo to select on, (`secondary-text-button` for instance), but we do not _also_ need `demo-button` on it as well.

PR #1788, opened 2017-12-14
chore(textfield): Update README to follow best practices

lynnjepsen:
I've cleaned up each README of each subelement follow the structure of
* Intro (simple one line from spec)
* HTML Structure
* Usage within `mdc-text-field`
* CSS Classes
* `MDCTextFieldFoo` API documentation
* `MDCTextFieldFooAdapter` API documentation
* `MDCTextFieldFooFoundation` API documentation

I've also cleaned up the main text field README to follow a very similar structure. There are a lot of 
> _NOTE_

in packages/textfield/README.md...which is concerning...but it represents how complicated the HTML structure is.

The idea, going forward, is that any breaking change to `MDCTextField` or its sub elements will require a change to these READMEs.

CHANGELOG only allows for a very short description of a breaking change. The idea is to have READMEs be the more detailed documentation for what APIs changed.

### DIFF
--- a/packages/mdc-textfield/README.md
+++ b/packages/mdc-textfield/README.md
@@ -125,8 +125,6 @@ included as part of the DOM structure of a full width text field.
 > _NOTE_: Only use `mdc-text-field__label` within `mdc-text-field--textarea` _if you plan on using
 > Javascript_. Otherwise, use the `placeholder` attribute, as shown below.
 
-##### CSS Only
-
 ```html
 <div class="mdc-text-field mdc-text-field--textarea">
   <textarea id="textarea-css-only"
--- a/packages/mdc-textfield/README.md
+++ b/packages/mdc-textfield/README.md
@@ -137,6 +137,8 @@ included as part of the DOM structure of a full width text field.
 
 #### Disabled
 
+Add the `disabled` attribute to `<input>` if the `mdc-text-field` is disabled. You also need to add `mdc-text-field--disabled` to the `mdc-text-field`.
+
 ```html
 <div class="mdc-text-field mdc-text-field--disabled">
   <input type="text" id="disabled-text-field" class="mdc-text-field__input" disabled>
@@ -161,7 +163,6 @@ See [here](icon/) for more information on using icons.
 CSS Class | Description
 --- | ---
 `mdc-text-field` | Mandatory
-`mdc-text-field__input` | Mandatory
 `mdc-text-field--upgraded` | Indicates the text field is upgraded, normally by JavaScript
 `mdc-text-field--box` | Styles the text field as a box text field
 `mdc-text-field--fullwidth` | Styles the text field as a full width text field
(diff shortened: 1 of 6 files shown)


packages/mdc-textfield/README.md
```diff
@@ -149,210 +135,100 @@ See [here](icon/) for more information on using icons.
...
-
-Sets the content of the helper text, if it exists.
-
-### Theming
-
-MDC Text Field components use the configured theme's primary color for its underline and label text
-when the input is focused.
-
-MDC Text Field components support dark themes.
+### `MDCTextFieldAdapter`
```
bonniezhou: Not sure if we want to add it here or in the future, but it'd be nice to separate what's useful for someone who wants to use MDC (i.e. all the stuff above this), or someone who wants to build a wrapper around MDC (i.e foundation/adapter implementation).

packages/mdc-textfield/README.md
```diff
@@ -149,210 +135,100 @@ See [here](icon/) for more information on using icons.
...
 
-#### MDCTextField API
+Mixin | Description
+--- | ---
+`mdc-text-field-box-corner-radius($radius)` | Customizes the border radius for a box text field
+`mdc-text-field-textarea-corner-radius($radius)` | Customizes the border radius for a `<textarea>` text field
 
-Similar to regular DOM elements, the `MDCTextField` functionality is exposed through accessor
-methods.
+### `MDCTextField`
```
bonniezhou: I think we should at least link to docs that explain importing/instantiation if we don't provide it under every component.

packages/mdc-textfield/README.md
```diff
@@ -149,210 +135,100 @@ See [here](icon/) for more information on using icons.
...
 
-##### MDCTextField.disabled
+#### `MDCTextField.disable`
```
bonniezhou: Open question: would it be better to put these methods into a table like above?

packages/mdc-textfield/README.md
```diff
@@ -149,210 +135,100 @@ See [here](icon/) for more information on using icons.
 </div>
 ```
 
-### Full-width
+#### Disabled
```
bonniezhou: I think we can remove this, it seems that a line under "CSS Classes" will suffice for any variations that don't require further explanation.

packages/mdc-textfield/README.md
```diff
@@ -149,210 +135,100 @@ See [here](icon/) for more information on using icons.
...
-  // Optionally do something with the element or the Ripple foundation...
-  return new MDCRipple(el, foundation);
-});
-```
+### CSS Classes
+
+CSS Class | Description
+--- | ---
+`mdc-text-field` | Mandatory
+`mdc-text-field__input` | Mandatory
```
bonniezhou: This can be removed. Seems like this table should just be for root class modifiers.

packages/mdc-textfield/README.md
lynnjepsen: I added some more notes about why this is in the "HTML Structure" section. Basically you need a 'disabled' attribute.

packages/mdc-textfield/README.md
lynnjepsen: I think we should think about this for the future. Right now we only have one README per component, so we put this all into one README.

packages/mdc-textfield/README.md
```diff
@@ -186,199 +164,97 @@ behave normally.
...
+### `MDCTextFieldFoundation`
 
-MDC Text Field components use the configured theme's primary color for its underline and label text
-when the input is focused.
+Method Signature | Description
+--- | ---
+`isDisabled() => boolean` | Returns whether or not the input is disabled
+`setDisabled(disabled: boolean) => void` | Updates the input's disabled state
+`setValid(isValid: boolean) => void` | Sets the validity state of the Text Field. Triggers custom validity checking
+`setHelperTextContent(content: string) => void` |
```
bonniezhou: Forgot to remove this?

packages/mdc-textfield/README.md
```diff
@@ -186,199 +164,97 @@ behave normally.
...
+See [here](icon/) for more information on using icons.
 
-Updates the input's disabled state.
+### CSS Classes
+
+CSS Class | Description
+--- | ---
+`mdc-text-field` | Mandatory
+`mdc-text-field--upgraded` | Indicates the text field is upgraded, normally by JavaScript
+`mdc-text-field--box` | Styles the text field as a box text field
```
bonniezhou: Add `mdc-text-field--outlined` class + description

PR #1799, opened 2017-12-18
fix(list): Respect BEM when outputting the base stylesheet.

### LINKED ISSUE #1748: List class selector specificity increase
Since 0.27.0, the specificity of some list related class selector has increased with unnecessary nesting, breaking low-specificity style overrides.

Previously: `.mdc-list-group__subheader`, `.mdc-list-item__start-detail`

Now: `.mdc-list-group .mdc-list-group__subheader`, `.mdc-list .mdc-list-item__start-detail`

There appears to be no reason to warrant this specificity increase. Could this be reverted?

simono:
Fixes #1748.

@lynnjepsen I replicated what you did in #1733.
The dark-theme selectors are still nested: e.g. `.mdc-theme--dark .mdc-list-group .mdc-list-group__subheader`. I didn't come up with an elegant solution :-/


lynnjepsen:
And we'll circle back to how to elegantly handle dark theme soon ;)

PR #1800, opened 2017-12-18
fix(ripple): Listen for up events at document level

### LINKED ISSUE #1124: Button with ripples stays focused after mouse is moved away
Clicking and holding a button, then releasing or clicking away from it, keeps the button in focused state.

An example on <https://material-components-web.appspot.com/button.html>.
mdc-button

The behavior is different with "CSS Only" buttons. You have to click outside of button's bounds to remove the "background tint", which also doesn't seem correct. With Material Design Lite it works as expected.

kfranqueiro:
This fixes #1124 by registering touch/mouse up event handlers at the documentElement level at the time that a down event is triggered.

It also adds an instance property to track the previous event for a bit of time to avoid triggering activation animation twice for the same event due to browsers [firing mouse events after touch events](https://patrickhlauke.github.io/touch/tests/results/).

This PR was inspired partly by discussion on #1461 which previously attempted to resolve this issue.


Spetnik:
Functionality looks great to me. The only issue I notice is on an <a> element, if I move the mouse before the mouse-up, it stays highlighted. But links in general have a very different behavior altogether from other elements, so I don't know how big of a deal that is.

packages/mdc-ripple/foundation.js
```diff
@@ -365,48 +426,38 @@ class MDCRippleFoundation extends MDCFoundation {
     this.adapter_.computeBoundingRect();
   }
 
+  resetActivationState_() {
+    // Take note of previous activation event type to ignore follow-on event for the same interaction on touch devices
+    this.previousActivationEvent_ = this.activationState_.activationEvent;
+    this.activationState_ = this.defaultActivationState_();
+    setTimeout(() => this.previousActivationEvent_ = null, 100);
```
patrickrodee: Is the `setTimeout` to account for delays between when same-interaction "follow-on" events are triggered?

kfranqueiro: Yup. Do you think my comment above these lines needs improvement?

packages/mdc-ripple/foundation.js
```diff
@@ -219,16 +212,73 @@ class MDCRippleFoundation extends MDCFoundation {
...
+    ACTIVATION_EVENT_TYPES.forEach((type) => {
+      this.adapter_.registerInteractionHandler(type, this.activateHandler_);
     });
+    this.adapter_.registerInteractionHandler('focus', this.focusHandler_);
+    this.adapter_.registerInteractionHandler('blur', this.blurHandler_);
     this.adapter_.registerResizeHandler(this.resizeHandler_);
   }
 
+  /**
+   * @param {Event} e
```
patrickrodee: Should be a non-nullable annotation for Event

packages/mdc-ripple/foundation.js
```diff
@@ -127,26 +123,21 @@ class MDCRippleFoundation extends MDCFoundation {
...
-        () => this.adapter_.removeClass(MDCRippleFoundation.cssClasses.BG_FOCUSED)
-      ),
-    };
+    /** @private {function(!Event)} */
+    this.activateHandler_ = (e) => this.activate_(e);
+
+    /** @private {function(!Event)} */
+    this.deactivateHandler_ = (e) => this.deactivate_(e);
+
+    /** @private {function(!Event)} */
```
patrickrodee: Event annotation should be nullable since it's not required inside the function

packages/mdc-ripple/foundation.js
```diff
@@ -127,26 +123,21 @@ class MDCRippleFoundation extends MDCFoundation {
...
+    this.focusHandler_ = () => requestAnimationFrame(
+      () => this.adapter_.addClass(MDCRippleFoundation.cssClasses.BG_FOCUSED)
+    );
+
+    /** @private {function(!Event)} */
```
patrickrodee: Same as above

test/unit/mdc-ripple/foundation-activation.test.js
```diff
@@ -234,6 +234,26 @@ testFoundation('adds activation classes on programmatic activation', ({foundatio
   td.verify(adapter.addClass(cssClasses.FG_ACTIVATION));
 });
 
+testFoundation('programmatic activation immediately after interaction', ({foundation, adapter, mockRaf}) => {
```
patrickrodee: A more descriptive name would be helpful (something like "programmatic activation that occurs immediately after interaction adds the FG activation class")

kfranqueiro: If I make the description any longer it will overflow our maximum line length, and that's usually where I draw the line...

patrickrodee: Hahaha fair enough :) Approved!

Issue #1152, opened 2017-08-17
Add Sass color mixin to tabs


lynnjepsen:
- [ ] Create color mixins
- [ ] Call color mixin from stylesheet, with primary color
- [ ] Delete any —primary --accent modifiers
- [ ] Update README to document color mixins
- [ ] Include "breaking change" message about removing modifiers, point to color mixin as deprecation replacement.

moog16:
Spoke to designers and ran by @kfranqueiro  and @acdvorak on the mixins I'm creating:

```
$mdc-tab-label-ink-color
$mdc-tab-icon-ink-color
$mdc-tab-underline-ink-color

$mdc-tab-label-activated-ink-color
$mdc-tab-underline-activated-ink-color
$mdc-tab-icon-activated-ink-color

$mdc-tab-underline-height
$mdc-tab-width```

PR #1801, opened 2017-12-19
feat/tabs/sass color mixins

moog16:
fixes: https://github.com/material-components/material-components-web/issues/1152

### DIFF
--- a/packages/mdc-tabs/_mixins.scss
+++ b/packages/mdc-tabs/_mixins.scss
@@ -1,0 +1,39 @@
+@import "@material/theme/mixins";
+
+@mixin mdc-tab-label-ink-color($color) {
+  .mdc-tab {
+    @include mdc-theme-prop(color, $color);
+  }
+}
+
+@mixin mdc-tab-icon-ink-color($color) {
+  
+}
+
+@mixin mdc-tab-underline-ink-color($color) {
+
+}
+
+@mixin mdc-tab-label-activated-ink-color($color) {
+  .mdc-tab--active {
+    @include mdc-theme-prop(color, $color);
+  }
+}
+
+@mixin mdc-tab-underline-activated-ink-color($color) {
+  .mdc-tab-bar__indicator {
+    @include mdc-theme-prop(background-color, $color);
+  }
+}
+
+@mixin mdc-tab-icon-activated-ink-color($color) {
+
+}
+
+@mixin mdc-tab-underline-height($color) {
+
+}
+
+@mixin mdc-tab-width($color) {
+
+}
--- a/packages/mdc-tabs/_mixins.scss
+++ b/packages/mdc-tabs/_mixins.scss
@@ -69,20 +69,3 @@
     @include mdc-theme-prop(color, $color);
   }
 }
-
-@mixin mdc-tab-indicator-height($height) {
-  .mdc-tab-bar__indicator {
-    height: $height;
-  }
-
-  &:not(.mdc-tab-bar-upgraded) .mdc-tab::after {
-    bottom: 1px;
-    height: $height;
-  }
-}
-
-@mixin mdc-tab-width($width) {
-  .mdc-tab {
-    min-width: $width;
-  }
-}
--- a/packages/mdc-tabs/_mixins.scss
+++ b/packages/mdc-tabs/_mixins.scss
@@ -15,57 +15,51 @@
 @import "@material/theme/mixins";
 @import "../variables";
 
+@mixin mdc-tab-ink-color($color) {
+  @include mdc-tab-label-ink-color($color);
+  @include mdc-tab-icon-ink-color($color);
+}
+
+@mixin mdc-tab-activated-ink-color($color) {
+  @include mdc-tab-label-activated-ink-color($color);
+  @include mdc-tab-icon-activated-ink-color($color);
+  @include mdc-tab-indicator-ink-color($color);
+}
+
 @mixin mdc-tab-label-ink-color($color) {
   .mdc-tab:not(.mdc-tab--active) {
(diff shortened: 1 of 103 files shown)


packages/mdc-tabs/_mixins.scss
```diff
@@ -0,0 +1,88 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@import "@material/theme/mixins";
+@import "../variables";
+
+@mixin mdc-tab-label-ink-color($color) {
+  .mdc-tab:not(.mdc-tab--active) {
```
lynnjepsen: I go back and forth on the best way to support states within an element. In this case its how to support customization of both the inactive and active state of a tab.

In mdc-button, we did *not* want to allow customization of the disabled button, so we added a `:not(:disabled)` to our public mixins.

However, in this mdc-tab case, I think we want to support customization of active and inactive tabs separately.

I propose this:
* one public sass mixins `mdc-tab-label-ink-color`
* drop `mdc-tab-label-activated-ink-color`
* change mdc-tab-bar.scss to have:

```
.mdc-tab {
  @include mdc-tab-label-ink-color(text-secondary-on-light);
}
.mdc-tab--active {
  @include mdc-tab-label-ink-color(text-primary-on-light);
}
```

Then clients should override the the active inactive state like this
```
.foo-tab:not(.mdc-tab--active) {
  @include mdc-tab-label-ink-color(pink);
}
.foo-tab.mdc-tab--active {
  @include mdc-tab-label-ink-color(purple);
}
```

I think it is more important that clients keep their active and inactive states in sync, then it is for clients to keep their subcomponents in sync. For example this makes a lot of sense:
```
.foo-tab:not(.mdc-tab--active) {
  @include mdc-tab-label-ink-color(pink);
  @include mdc-tab-icon-ink-color(pink);
}
.foo-tab.mdc-tab--active {
  @include mdc-tab-label-ink-color(purple);
  @include mdc-tab-icon-ink-color(purple);
}
```

But this is harder to read
```
.foo-tab {
  @include mdc-tab-label-ink-color(pink /*inactive color*/, purple /*active color*/);
  @include mdc-tab-icon-ink-color(pink /*inactive color*/, purple /*active color*/);
}
```

packages/mdc-tabs/_mixins.scss
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+@mixin mdc-tab-icon-activated-ink-color($color) {
+  .mdc-tab.mdc-tab--active .mdc-tab__icon {
+    @include mdc-theme-prop(color, $color);
+  }
+}
+
+@mixin mdc-tab-indicator-height($height) {
```
lynnjepsen: actually, don't bother with indicator height and width right now. just focus on the color mixins for now

moog16: Ok will remove from PR

packages/mdc-tabs/_mixins.scss
```diff
@@ -0,0 +1,88 @@
...
+
+@import "@material/theme/mixins";
+@import "../variables";
+
+@mixin mdc-tab-label-ink-color($color) {
+  .mdc-tab:not(.mdc-tab--active) {
+    @include mdc-theme-prop(color, $color);
+  }
+
+  @include mdc-tab-icon-ink-color($color);
```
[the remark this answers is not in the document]
moog16: This way actually works where you can customize the ink and the icon color with these two mixins. But I agree it isn't clear. I'll go with the three mixins.

packages/mdc-tabs/_mixins.scss
moog16: If we go that route won't have mixins on the parent class, which I thought was what we wanted. Also the label-ink-color mixin would be a proxy to mdc-theme-prop:

```
@mixin mdc-tab-label-ink-color($color) {
}
```

The icon would also just be:
```
@mixin mdc-tab-icon-ink-color($color) {
}
```

What is the main purpose for these scss mixins? Is it to allow clients to customize the components quicker? To make our styles DRY'er? In either case if our scss mixins for other components are designed to live on the parent class, we should try to be consistent here as well. Clients and external developers can easily see the mixins we've created and know where to put them for quick configuration. 

I'll probably pose the question on theming since you are on vacayy ;)

packages/mdc-tabs/_mixins.scss
```diff
@@ -0,0 +1,88 @@
...
+    @include mdc-theme-prop(color, $color);
+  }
+}
+
+@mixin mdc-tab-indicator-ink-color($color, $opacity: .87) {
+  .mdc-tab-bar__indicator {
+    @include mdc-theme-prop(background-color, $color);
+  }
+
+  &:not(.mdc-tab-bar-upgraded) .mdc-tab {
```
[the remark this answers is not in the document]
moog16: you're right confusing. I think it is better to go with something like:
```
@mixin mdc-tab-indicator-ink-color($color, $opacity);
@mixin mdc-tab-label-ink-color($color, $opacity);
@mixin mdc-tab-icon-ink-color($color, $opacity);
```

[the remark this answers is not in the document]
moog16: Nevermind...we don't need $opacity. Developers and clients can just go like:

```
@include mdc-tab-indicator-ink-color(rgba(0,0,0, 0.9))
```

moog16:
Closing as I am addressing changes in https://github.com/material-components/material-components-web/pull/1851

PR #1807, opened 2017-12-20
feat(text-field): Add ripple to outlined text field

bonniezhou:
Add ripple to Outlined Text Field.
The ripple is instantiated on the outline element instead of the root element to allow the label to be seen beyond the outline while still clipping the ripple at the outline.

### DIFF
--- a/packages/mdc-textfield/index.js
+++ b/packages/mdc-textfield/index.js
@@ -109,6 +109,13 @@ class MDCTextField extends MDCComponent {
     const outlineElement = this.root_.querySelector(strings.OUTLINE_SELECTOR);
     if (outlineElement) {
       this.outline_ = outlineFactory(outlineElement);
+      const adapter = Object.assign(MDCRipple.createAdapter(this.outline_), {
+        isSurfaceActive: () => this.input_[MATCHES](':active'),
+        registerInteractionHandler: (type, handler) => this.input_.addEventListener(type, handler),
+        deregisterInteractionHandler: (type, handler) => this.input_.removeEventListener(type, handler),
+      });
+      const foundation = new MDCRippleFoundation(adapter);
+      this.ripple = this.outline_.createRipple(rippleFactory, foundation);
     }
     if (this.input_.hasAttribute(strings.ARIA_CONTROLS)) {
       const helperTextElement = document.getElementById(this.input_.getAttribute(strings.ARIA_CONTROLS));
--- a/packages/mdc-textfield/index.js
+++ b/packages/mdc-textfield/index.js
@@ -16,7 +16,9 @@
  */
 
 import MDCComponent from '@material/base/component';
-import {MDCRipple, MDCRippleFoundation} from '@material/ripple';
+/* eslint-disable no-unused-vars */
+import {MDCRipple, MDCRippleFoundation, RippleCapableSurface} from '@material/ripple';
+/* eslint-enable no-unused-vars */
 import {getMatchesProperty} from '@material/ripple/util';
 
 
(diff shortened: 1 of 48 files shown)


packages/mdc-textfield/outline/README.md
```diff
@@ -21,22 +21,64 @@ The outline is a border around all sides of the text field. This is used for the
...
+  </div>
+  <div class="mdc-text-field__idle-outline"></div>
+</div>
+```
+
+### CSS Classes
+
+CSS Class | Description
+--- | ---
+`mdc-text-field__outline` | Mandatory
```
kfranqueiro: No actual descriptions?

packages/mdc-textfield/outline/README.md
```diff
@@ -21,22 +21,64 @@ The outline is a border around all sides of the text field. This is used for the
...
-getWidth() => number | Returns the width of the outline element
-getHeight() => number | Returns the height of the outline element
-setOutlinePathAttr(value: string) => void | Sets the "d" attribute of the outline element's SVG path
+`MDCTextFieldOutline.createRipple(rippleFactory, rippleInputMethods) => MDCRipple` | Returns an `MDCRipple` instance set on the `mdc-text-field__outline` element
+
+##### `MDCTextFieldOutline.foundation`
+
+This allows the parent `MDCTextField` component to access the public methods on the `MDCTextFieldOutlineFoundation` class.
+
+##### `MDCTextFieldOutline.createRipple(rippleFactory: (function(!Element, !MDCRippleFoundation): !MDCRipple), foundation: !MDCRippleFoundation)`
```
kfranqueiro: I think we tend to not include the `!` in our notation in the READMEs which follows TypeScript syntax.

packages/mdc-textfield/outline/mdc-text-field-outline.scss
```diff
@@ -33,6 +33,10 @@
 }
 
 .mdc-text-field__outline {
+  @include mdc-ripple-surface;
+  @include mdc-states-base-color(black);
```
kfranqueiro: Remove this line; the next line will already result in including `mdc-states-base-color(text-primary-on-light)`.

packages/mdc-textfield/outline/mdc-text-field-outline.scss
```diff
@@ -33,6 +33,10 @@
...
+  @include mdc-states(text-primary-on-light);
+  @include mdc-ripple-radius;
```
kfranqueiro: Nit: Move this above `mdc-states`, below `mdc-ripple-surface`

packages/mdc-textfield/index.js
```diff
@@ -120,6 +109,23 @@ class MDCTextField extends MDCComponent {
...
+        isSurfaceActive: () => this.input_[MATCHES](':active'),
+        registerInteractionHandler: (type, handler) => this.input_.addEventListener(type, handler),
+        deregisterInteractionHandler: (type, handler) => this.input_.removeEventListener(type, handler),
+      };
+      if (this.root_.classList.contains(cssClasses.BOX)) {
+        const adapter = Object.assign(MDCRipple.createAdapter(this), rippleInputMethods);
+        const foundation = new MDCRippleFoundation(adapter);
+        this.ripple = rippleFactory(this.root_, foundation);
+      } else if (this.root_.classList.contains(cssClasses.OUTLINED)) {
+        this.ripple = this.outline_.createRipple(rippleFactory, rippleInputMethods);
```
kfranqueiro: It seems a bit overboard to me to have a separate method on the outline subcomponent that does exactly the same thing as above, and requiring us to pass everything through to it, when the only difference is which element the ripple is instantiated on.

I also find it a bit strange since the ripple is still destroyed from the top level component, and there's nothing really "owning" it from the outline subcomponent.

Given that the branching logic is here anyway, should we simply conditionalize which element the ripple is created on here and otherwise have the same code for both in one place?

Another thought: should the decision of which element hosts the ripple be determined within an API for the top-level foundation, so that component implementors have less logic to worry about?

kfranqueiro:
Question about intended behavior for outlined text fields: is this really supposed to have hover and focus shade, or is it just supposed to have the ripple on press? It looks particularly odd seeing a shade when the label floats and the outline has the gap at the top.

If this should actually only have pressed state, you should be able to get away with including only `mdc-states-press-opacity` specifically instead of `mdc-states`, but you'll need to manually specify the correct opacity from the maps in ripple's variables.

packages/mdc-textfield/index.js
bonniezhou: The reason for the separate method on the outline subcomponent is because we need access to the outline's root element to instantiate the ripple on. The other option was to have a public `getRoot()` method in the outline, which seems to break best practices....although that might seem less weird than having an entire `createRipple()` method on a subcomponent that doesn't even own the ripple. Thoughts?

Moving the ripple host decision to the foundation seems even more complicated since the main foundation would have to somehow access the outline subcomponent (don't know if this is possible, maybe through the outline subadapter?...)

bonniezhou:
You're right, it should only have the ripple on press. Thanks for catching!

packages/mdc-textfield/index.js
kfranqueiro: I would argue that we already have access to the outline subcomponent's root element: `outlineElement`. We just passed it _to_ the subcomponent constructor about 30 lines above in this same method.

And...yeah my idea about making the foundation decide ends up being pretty awkward, so nevermind about that one.

packages/mdc-textfield/outline/mdc-text-field-outline.scss
kfranqueiro: True (I hadn't tested the behavior prior to writing that comment), but should it be using `text-primary-on-light` like you were doing with `mdc-states`, or `black`?

packages/mdc-textfield/index.js
bonniezhou: D'oh, huge brain fart on my part. Updated it to share the same code except for conditionalizing on which element the ripple is created on.

packages/mdc-textfield/outline/mdc-text-field-outline.scss
```diff
@@ -33,6 +33,10 @@
 }
 
 .mdc-text-field__outline {
+  @include mdc-ripple-surface;
+  @include mdc-ripple-radius;
+  @include mdc-states-base-color(text-primary-on-light);
+  @include mdc-states-press-opacity(0.12);
```
kfranqueiro: We should reference `map-get($mdc-ripple-dark-ink-opacities, "press")` (which is .16, not .12).

Also, we should probably include styles for dark mode passing `text-primary-on-dark` and `map-get($mdc-ripple-light-ink-opacities, "press")`.

(There's no dark mode toggle under the outlined text field example at the moment either; should we add one?)

bonniezhou: Good call.

I plan on adding dark mode along with dense mode to the PR for adding leading/trailing icons.

Issue #1505, opened 2017-10-30
Textfield fullwidth placeholder text disappears on focus instead of on keydown


djensen47:
According to the Material design spec, "Placeholder text can be displayed in an empty text field until input is entered." Also, "As the user types into the text field, input text replaces any placeholder text."

When there is a fullwidth textfield with a placeholder, it currently clears the placeholder on focus. The expected result is that the placeholder stays until the use starts typing input.

dessant:
I can confirm this behavior. It seems to be a regression, the placeholder stays visible on focus with textfield 0.4.1.

lynnjepsen:
Looks like a bug! Thanks for filing.

For posterity, here is a codepen reproducing the issue:
https://codepen.io/anon/pen/KydoyJ

djensen47:
I found the issue. The wrong color is being assigned to :focus::placeholder

here:
https://github.com/material-components/material-components-web/commit/45c6cf68aad9a0985b1ca0b4ad81ed466298809f#diff-0d82c2a11524792f13f48f99ef46fd4cL69

and here
https://github.com/material-components/material-components-web/commit/45c6cf68aad9a0985b1ca0b4ad81ed466298809f#diff-0d82c2a11524792f13f48f99ef46fd4cL86

Then take a look at https://github.com/material-components/material-components-web/blob/251338e5812461df6fa717c40b68b5d21e291806/packages/mdc-textfield/_variables.scss#L35

Those values are light text on a light background and dark text on a dark background.

---
The previous values are still around. For instance, the **light** theme used to use ` rgba(black, .54)`:
https://github.com/material-components/material-components-web/blob/57581edc4f643e39c0bcfffa95bab4ffde68b865/packages/mdc-theme/_variables.scss#L132

And the **dark** theme used to use `rgba(white, .7)`
https://github.com/material-components/material-components-web/blob/57581edc4f643e39c0bcfffa95bab4ffde68b865/packages/mdc-theme/_variables.scss#L138

djensen47:
@lynnjepsen I can submit a PR.

I'm not sure what @amsheehan was intending to do but my _guess_ is that those colors should be swapped.

 ```
$mdc-textfield-light-placeholder: rgba(black, .38);
$mdc-textfield-dark-placeholder: rgba(white, .3);
```

Or the values used in the :focus::placeholder should be swapped.

Thoughts?

djensen47:
I generally like to follow the guideline of asking before making a PR.  😄 

[...]

PR #1813, opened 2017-12-20
fix(textfield): placeholder colors

moog16:
Fixes: https://github.com/material-components/material-components-web/issues/1505. Also added a dark theme and disabled checkbox on the CSS Only textfield to showcase fix.

### DIFF
--- a/demos/text-field.html
+++ b/demos/text-field.html
@@ -319,8 +319,8 @@
 
       <section class="example">
         <h2>CSS Only Text Field</h2>
-        <div class="mdc-form-field mdc-form-field--align-end">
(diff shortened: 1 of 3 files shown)


patrickrodee:
Aight, I know this is out of scope for the PR but since you're already modifying the demo file, could you rename the input's `id` and the corresponding `for` to a non-conflicting `id`? Chrome's getting unhappy in the console.

https://github.com/material-components/material-components-web/blob/5f00f0e4c0b462cdc8449b556e35641af7642403/demos/text-field.html#L93-L94

moog16:
I'm glad you said that, I was about to delete them in another PR. Updating...

Issue #1529, opened 2017-11-02
Textfield doesn't respect autofocus


acdvorak:
### What MDC-Web Version are you using?

> 0.24.0

### What browser(s) is this bug affecting?

> All (tested in Chrome, Safari, and Firefox)

### What OS are you using?

> All (tested in macOS and Linux)

### What are the steps to reproduce the bug?

> 1. Add `autofocus` attribute to a textfield
> 2. Run the demo server
> 3. Go to localhost:8080/textfield.html
> 4. Observe the component's behavior

[CodePen](https://codepen.io/acdvorak/full/zPvQam) _**(TIMING DEPENDENT - may be hard to repro on a slow website with lots of external network fetches like CodePen. Repros reliably on my local demo server.)**_

### What is the expected behavior?

> Label should animate upward like it normally does when focused

### What is the actual behavior?

> The label stays at its initial position, on top of the caret and on top of any text the user enters. If you <kbd>tab</kbd> and then <kbd>shift</kbd>+<kbd>tab</kbd>, the label position adjusts correctly.

![mdc-web-bug---textfield-autofocus](https://user-images.githubusercontent.com/409245/32320529-43b5ddc8-bf7b-11e7-8206-f2eb83e661b6.gif)

### Any other information you believe would be useful?

I'm guessing that `focus` occurs _before_ the JavaScript event handler gets bound to the input.

lynnjepsen:
looks like a bug, thanks for filing

anton-kachurin:
In this particular case, moving `<script src="/assets/material-components-web.js"></script>` into `<head>` helps to solve the issue.

Nevertheless, we still need some solution for cases when `focus` happens before textfield's initialization, such as when scripts are loaded with `async` attribute or dynamically. Also, most browsers do focus elements loaded via ajax if they have `autofocus`.

`document.activeElement` may be of use:
https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement

Possibly we could extend the adapter with the method `getActiveElement` or `isInputFocused` and set the "focused" state on the Text Field on initialization?

Garbee:
In initialization just check if the input is focused already. If so, apply focus triggers immediately at the end of initialisation.

PR #1815, opened 2017-12-21
fix(textfield): added isFocused to adapter incase autofocus attr is present

moog16:
fixes: https://github.com/material-components/material-components-web/issues/1529


packages/mdc-ripple/foundation.js
```diff
@@ -463,6 +463,7 @@ class MDCRippleFoundation extends MDCFoundation {
     }
     this.layoutFrame_ = requestAnimationFrame(() => {
       this.layoutInternal_();
+
```
amsheehan: ?

moog16: removing...

Issue #1435, opened 2017-10-17
Missing Textfield label primary color application on focus


marioavs:
### What MDC-Web Version are you using?

> 0.23.0

### What are the steps to reproduce the bug?

> 1. Run the demo server (it is also present in the demo site https://material-components-web.appspot.com/textfield.html)
> 2. Go to localhost:8080/textfield.html
> 3. Check the Dark Theme option for the "Full Functionality JS Component" (second Textfield component). This step is only needed to make more evident the label coloring.
> 4. Tab-focus on the "Full Functionality JS Component" (second Textfield component)
> 5. Observe the label color

### What is the expected behavior?

> The label should be displayed with primary color for focused textfield components

### What is the actual behavior?

> The label is displayed with the normal / idle color as if it would be a non focused textfield component

### Any other information you believe would be useful?

> It looks like the following is needed and has been removed from packages/mdc-textfield/mdc-textfield.scss, for the `mdc-textfield--focused` class:
`@include mdc-theme-prop(color, primary);`

reddysridhar53:
Can i take it up?

dscalzi:
Still an issue on 0.25.0.

moog16:
Will fix as a result of https://github.com/material-components/material-components-web/issues/1526

PR #1820, opened 2017-12-21
fix(textfield): add primary color to textfield label on focus

williamernest:
fixes: #1435


williamernest:
@amsheehan 
Changed it to apply the label color from the input focus psuedo class.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -237,7 +241,6 @@
     position: absolute;
     bottom: 20px;
     width: calc(100% - #{$mdc-text-field-icon-padding});
-    color: $mdc-text-field-box-secondary-text;
```
acdvorak: Why does this line need to be removed? It doesn't appear to be related to the `:focused` state.

Also, it looks like `$mdc-text-field-box-secondary-text` isn't used anywhere else, so if we really do need to remove this line, we should also delete the variable from `_variables.scss`.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -246,10 +249,6 @@
     // when the ripple is activated behind it.
     will-change: transform;
 
-    @include mdc-theme-dark(".mdc-text-field") {
```
acdvorak: Why do these lines need to be removed? They don't appear to be related to the `:focused` state.

packages/mdc-textfield/mdc-text-field.scss
williamernest: The default `color` property is set in the mdc-text-field-label.scss file. My idea was to let the label scss file determine the default values, and then override them when we need to change them based on interaction with the other elements in the mdc-text-field.scss file.

PR #1824, opened 2017-12-21
feat(text-field): Add CSS-only version of outlined text field

### LINKED ISSUE #1601: Add CSS for css only version of outlined text field
Add CSS for css only version of outlined text field

... also add a demo

bonniezhou:
Fixes #1601.

### DIFF
--- a/packages/mdc-textfield/mdc-text-field.scss
+++ b/packages/mdc-textfield/mdc-text-field.scss
@@ -661,8 +661,8 @@
 
     height: 100%;
     padding: 0 0 0 12px;
-    border: 1px solid $mdc-text-field-outlined-idle-border;
     transition: mdc-text-field-transition(border);
+    border: 1px solid $mdc-text-field-outlined-idle-border;
 
     &:hover {
       border-color: $mdc-text-field-outlined-hover-border;
--- a/packages/mdc-textfield/mdc-text-field.scss
+++ b/packages/mdc-textfield/mdc-text-field.scss
@@ -653,16 +653,35 @@
   }
 }
 
-.mdc-text-field--box:not(.mdc-text-field--upgraded) {
-  height: 56px;
+.mdc-text-field:not(.mdc-text-field--upgraded) {
+  &.mdc-text-field--outlined {
+    height: 56px;
 
-  &::before,
-  &::after {
-    border-radius: 0;
+    .mdc-text-field__input{
+      @include mdc-text-field-outlined-corner-radius($mdc-text-field-border-radius);
+
+      height: 100%;
+      padding: 0 0 0 12px;
+      border: 1px solid $mdc-text-field-outlined-idle-border;
+      transition: mdc-text-field-transition(border);
+
+      &:hover, &:focus {
+        border-color: $mdc-text-field-outlined-hover-border;
+      }
+    }
   }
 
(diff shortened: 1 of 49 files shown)


bonniezhou:
Confirmed with designers that the outline should be darkened on hover, and changed to the primary color on focus.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -655,16 +655,39 @@
...
-.mdc-text-field--box:not(.mdc-text-field--upgraded) {
-  height: 56px;
+.mdc-text-field:not(.mdc-text-field--upgraded) {
+  &.mdc-text-field--outlined {
+    height: 56px;
 
-  &::before,
-  &::after {
-    border-radius: 0;
+    .mdc-text-field__input{
```
kfranqueiro: Missing space before `{`

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -655,16 +655,39 @@
   }
 }
 
-.mdc-text-field--box:not(.mdc-text-field--upgraded) {
-  height: 56px;
+.mdc-text-field:not(.mdc-text-field--upgraded) {
+  &.mdc-text-field--outlined {
```
kfranqueiro: Is there a reason you nested these selectors and increased their specificity?

Would this work just as well as `.mdc-text-field--box:not(...)` and `.mdc-text-field--outlined:not(...)`?

bonniezhou: `.mdc-text-field--outlined:not(...)` wasn't strong enough to override the `.mdc-text-field:not(.mdc-text-field--upgraded):not(.mdc-text-field--textarea) .mdc-text-field__input` selector from above, so I removed `:not(.mdc-text-field--textarea)` from that and put in its own rule.

So now we have 
- `.mdc-text-field--outlined:not(...)`
- `.mdc-text-field--box:not(...)`
- `.mdc-text-field--textarea:not(...)`

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -655,6 +653,27 @@
...
 
+.mdc-text-field--outlined:not(.mdc-text-field--upgraded) {
+  height: 56px;
+
+  .mdc-text-field__input {
+    @include mdc-text-field-outlined-corner-radius($mdc-text-field-border-radius);
+
+    height: 100%;
+    padding: 0 0 0 12px;
+    transition: mdc-text-field-transition(border);
```
kfranqueiro: See comment below about transitioning on border / border-color, so I'm wondering if we should forego transition on this. Depending on what your primary color is, transitioning between hover and focus can be rather subtle anyway.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -668,6 +687,12 @@
   }
 }
 
+.mdc-text-field--textarea:not(.mdc-text-field--upgraded) {
+  .mdc-text-field__input {
+    transition: mdc-text-field-transition(border-color);
```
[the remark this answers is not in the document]
bonniezhou: Oh I think you're right, removed all transitions for border.

PR #1827, opened 2017-12-22
fix(textfield): Use correct colors for invalid textfield with dark theme.

### LINKED ISSUE #1822: Different warning color values between stroke and helper text for outline text field

Latest Master

Chrome, all

### What OS are you using?

macOS Sierra

1. Fire up dev server
1. Turn on Outline Text Field's dark mode
1. F stuff up in the input

Red in the outline should match red in the helper text

They are different reds.

screen shot 2017-12-21 at 2 57 54 pm

simono:
Fixes #1822

### DIFF
--- a/packages/mdc-textfield/mdc-text-field.scss
+++ b/packages/mdc-textfield/mdc-text-field.scss
@@ -404,12 +404,42 @@
   .mdc-text-field__outline .mdc-text-field__outline-path {
     stroke: $mdc-text-field-error-on-light;
   }
+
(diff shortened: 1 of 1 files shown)


acdvorak:
Thanks for the PR!

It looks like this is only an issue in dark mode, which we're going to be removing very soon (you can just use our theme mixins or customize the value of $mdc-theme-primary and $mdc-theme-background instead).

Closing this PR for now; if it's still a problem in light mode, we can reopen it.

PR #1828, opened 2017-12-22
fix(textfield): Never show a border on the input of an outlined textfield.

### LINKED ISSUE #1817: Bottom line during hover - Outlined Text Field

> latest

> Chrome, probably all

### What OS are you using?

> macOS Sierra

1. Run the demo server
1. Add `mdc-theme--dark` as a class to the outline text field wrapper element. This will allow you to see the issue more clearly
1. Hover the cursor over the text field

The cursor should change, but nothing else

A bottom line gets added

dark-theme-textfield

simono:
Fixes #1817

### DIFF
--- a/packages/mdc-textfield/mdc-text-field.scss
+++ b/packages/mdc-textfield/mdc-text-field.scss
@@ -128,7 +128,8 @@
     display: flex;
     height: 30px;
     padding: 12px;
-    border: none;
+    // Use !important here to override all other border treatments
+    border: none !important;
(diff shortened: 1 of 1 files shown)


acdvorak:
It looks like this is only an issue in dark mode, which we're going to be removing very soon (you can just use our theme mixins or customize the value of $mdc-theme-primary and $mdc-theme-background instead).

Closing this issue for now; if it's still a problem in light mode, we can reopen it.

Issue #1598, opened 2017-11-15
Split up the "states" CSS classes in text field into private mixins


lynnjepsen:
Part of our effort to split text field to reduce merge conflicts and generally improve the readability of text field code.

lynnjepsen:
Should include:
* mdc-text-field--focused
* mdc-text-field--disabled
* mdc-text-field--invalid

PR #1831, opened 2017-12-22
Chore/textfield/split state classes

moog16:
fixes: https://github.com/material-components/material-components-web/issues/1598

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -41,24 +41,24 @@
   }
 }
 
-@mixin mdc-text-field-invalid-label-shake-keyframes_($modifier, $positionY) {
+@mixin mdc-text-field-invalid-label-shake-keyframes_($modifier, $positionY, $scale: .75) {
   @keyframes invalid-shake-float-above-#{$modifier} {
     0% {
-      transform: translateX(0) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(0) translateY(-#{$positionY}) scale(#{$scale});
     }
 
     33% {
       animation-timing-function: cubic-bezier(.5, 0, .701732, .495819);
-      transform: translateX(5px) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(5px) translateY(-#{$positionY}) scale(#{$scale});
     }
 
     66% {
       animation-timing-function: cubic-bezier(.302435, .381352, .55, .956352);
-      transform: translateX(-5px) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(-5px) translateY(-#{$positionY}) scale(#{$scale});
     }
 
     100% {
-      transform: translateX(0) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(0) translateY(-#{$positionY}) scale(#{$scale});
     }
   }
 }
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -63,11 +63,9 @@
   }
 }
 
-/*
-  * Private Mixins
-*/
+// Private Mixins
 
-// default
+// Common
 
 @mixin mdc-text-field-invalid_ {
   + .mdc-text-field-helper-text--validation-msg {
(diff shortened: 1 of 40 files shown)


packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,263 @@
     }
   }
 }
+
```
moog16: All of this code was moved from `/mdc-textfield/mdc-text-field.scss`. I made sure not to alter actual CSS. Pretty much a copy/paste, and extracted the state selectors (`--focused`, `--disabled`, etc)

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -670,39 +484,8 @@
 
 // postcss-bem-linter: end
 
-// postcss-bem-linter: define text-field-helper-text
-
-.mdc-text-field-helper-text {
```
moog16: moved this to `helper-text/mdc-text-field-helper-text.scss`

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,263 @@
...
+    @include mdc-theme-dark(".mdc-text-field", true) {
+      color: $mdc-text-field-error-on-dark;
+    }
+  }
+}
+
+// outline
+
+@mixin mdc-text-field-focused-outline_() {
```
bonniezhou: Parentheses can be omitted if there are no parameters, here and below

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -157,48 +155,15 @@
...
-    .mdc-text-field__icon {
-      @include mdc-theme-prop(color, $mdc-text-field-disabled-icon-on-light);
-    }
-  }
+.mdc-text-field--outlined.mdc-text-field--focused {
+  @include mdc-text-field-focused-outline_();
+}
 
-  // stylelint-enable plugin/selector-bem-pattern
+.mdc-text-field--outlined.mdc-text-field--focused {
```
bonniezhou: I think you meant for this block to be `--disabled` and the mixin to be `mdc-text-field-outlined-disabled_` (see comment on mixin above)

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -608,14 +425,13 @@
 }
 
 // Graceful degredation for css-only inputs
-
-.mdc-text-field {
-  &:not(.mdc-text-field--upgraded):not(.mdc-text-field--textarea) .mdc-text-field__input {
+.mdc-text-field:not(.mdc-text-field--upgraded) {
```
kfranqueiro: Changing this selector from how it was affects the usage of the `mdc-theme-dark` mixin inside of this block.

Previously, it selected on `.mdc-text-field--theme-dark...` but now it selects on `... .mdc-input--theme-dark`.

You might need to revert this selector change, but maybe passing a root selector to the `mdc-theme-dark` could work?

(I found this regression by diffing the output CSS on master running `npm build` against the output CSS from this branch.)

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,263 @@
...
+    background-color: $mdc-textarea-disabled-background-on-light;
+
+    @include mdc-theme-dark(".mdc-text-field") {
+      background-color: $mdc-textarea-disabled-background-on-dark;
+    }
+  }
+}
+
+// box
+@mixin mdc-text-field-box-disabled_ {
```
bonniezhou: Fix consistency: mixins above are named in order of state-variant, but this one is named variant-state.
(I actually prefer the variant-state naming more, but up to you as long as it's consistent.)

packages/mdc-textfield/mdc-text-field.scss
bonniezhou: I think we agreed that any CSS which requires a parent CSS class should not be moved into a subelement directory, since it's more important to keep all subelements in sync within any given parent state than it is to keep a subelement in sync with itself within all parent states.

packages/mdc-textfield/mdc-text-field.scss
moog16: Yes, correct! That was confusing naming on my part. Thanks for catching!

packages/mdc-textfield/_mixins.scss
moog16: I also like variant-state...changing

packages/mdc-textfield/label/mdc-text-field-label.scss
```diff
@@ -57,3 +57,8 @@
   }
 }
 // stylelint-enable plugin/selector-bem-pattern
+
+.mdc-text-field__input:required + .mdc-text-field__label::after {
```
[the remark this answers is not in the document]
moog16: Given @bonniezhou's comment found below, you are correct. If there is a parent selector needed for a subelement, it should be found with the parent selectors. 

What I was trying to do was make the mdc-text-field.scss file smaller, but I will revert as it is not the convention.

packages/mdc-textfield/mdc-text-field.scss
moog16: Thanks for finding! I thought I had covered the case when I manually tested, but I will revert.

packages/mdc-textfield/_mixins.scss
moog16: removed all instances

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,263 @@
...
+      color: $mdc-text-field-error-on-dark;
+    }
+  }
+}
+
+// outline
+
+@mixin mdc-text-field-focused-outline_() {
+  .mdc-text-field__idle-outline,
+  .mdc-text-field__label--float-above ~ .mdc-text-field__idle-outline {
```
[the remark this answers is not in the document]
moog16: I see what you're saying. I fixed up the others that were like this by adding in the state selector. What I ended up with for outline is the following:

```
.mdc-text-field--outline {
  ...
  @include mdc-text-field-outline-focused_;
  @include mdc-text-field-outline-disabled_;
  @include mdc-text-field-outline-float-above_;
}
```

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,263 @@
...
+  }
+
+  .mdc-text-field__icon {
+    @include mdc-rtl-reflexive-position(right, $mdc-text-field-dense-icon-position);
+  }
+}
+
+// shared
+
+@mixin mdc-text-field-icons-dense_ {
```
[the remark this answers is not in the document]
moog16: gotcha - I will combine them. My thinking was that it was cleaner to share the styles this way, but we don't want extra CSS.

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,295 @@
...
+
+    .mdc-text-field__label {
+      bottom: 8px;
+      cursor: default;
+    }
+  }
+}
+
+@mixin mdc-text-field-focused_ {
+  .mdc-text-field--focused {
```
lynnjepsen: You should not include the `mdc-text-field--focused` inside the mixin. Instead change mdc-text-field.scss to 

```
.mdc-text-field--focused {
  @include mdc-text-field-focused_;
}
```

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,295 @@
...
+    .mdc-text-field__input:required + .mdc-text-field__label::after {
+      color: $mdc-text-field-error-on-light;
+
+      @include mdc-theme-dark(".mdc-text-field", true) {
+        color: $mdc-text-field-error-on-dark;
+      }
+    }
+  }
+}
```
lynnjepsen: As part of `mdc-text-field-focused` you should add this logic

```
.mdc-text-field--focused + .mdc-text-field-helper-text:not(.mdc-text-field-helper-text--validation-msg) {
  opacity: 1;
}
```

See original logic here:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/mdc-text-field.scss#L727

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,295 @@
...
+    .mdc-text-field__label {
+      bottom: 8px;
+      cursor: default;
+    }
+  }
+}
+
+@mixin mdc-text-field-focused_ {
+  .mdc-text-field--focused {
+    .mdc-text-field__input:required + .mdc-text-field__label::after {
```
lynnjepsen: this CSS selector is complicated, and used in two places:
- here, within the `mdc-text-field-focused` mixin
- and in mdc-text-field.scss on line 291

Since we don't want these two places getting out of sync, you should create another mixin. I suggest this:

```
@mixin mdc-required-text-field-label-asterix_() {
  .mdc-text-field__input:required + .mdc-text-field__label::after {
    @content;
  }
}
```

You can use the mixin in the two places mentioned above.

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,295 @@
...
+    }
+
+    .mdc-text-field__label {
+      bottom: 8px;
+      cursor: default;
+    }
+  }
+}
+
+@mixin mdc-text-field-focused_ {
```
lynnjepsen: Can you sort these mixins in this order?

- `mdc-text-field-focused`
- `mdc-text-field-disabled`
- `mdc-text-field-invalid`

Lets come back to dense, textarea, and box in a later PR. I want to get these three figured out first.

packages/mdc-textfield/_mixins.scss
```diff
@@ -62,3 +62,295 @@
...
+
+    .mdc-text-field__label--float-above {
+      // NOTE: This is an eyeball'd approximation of what's in the mocks.
+      transform: translateY(calc(-100% - 2px)) scale(.923, .923);
+    }
+  }
+}
+
+@mixin mdc-text-field-disabled_ {
+  .mdc-text-field--disabled {
```
lynnjepsen: You should not include the `.mdc-text-field--disabled` inside the mixin. Instead change mdc-text-field.scss to 

```
.mdc-text-field--disabled {
  @include mdc-text-field-disabled_;
}
```

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -143,10 +140,13 @@
 
     position: absolute;
     bottom: 20px;
+    left: 16px;
```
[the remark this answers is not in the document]
kfranqueiro: This (and the next one) look like incorrect merge conflict resolutions after merged PRs by Bonnie, FYI.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -33,7 +33,6 @@
 @include mdc-text-field-invalid-label-shake-keyframes_(standard, 100%);
 @include mdc-text-field-invalid-label-shake-keyframes_(box, 50%);
 @include mdc-text-field-invalid-label-shake_keyframes_(outlined, 130%);
-@include mdc-text-field-invalid-label-shake_keyframes_(outlined-dense, 138%, .923);
```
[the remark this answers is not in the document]
moog16: Merge conflict. Was in my branch, but didn't have time to push. Starting fresh so it'll be there

packages/mdc-textfield/mdc-text-field.scss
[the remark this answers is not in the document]
moog16: Yes merge conflict....Starting fresh so we shouldn't see this issues

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -143,10 +140,13 @@
 
     position: absolute;
     bottom: 20px;
+    left: 16px;
     transition: transform 260ms ease;
     z-index: 2;
 
     &--float-above {
+      @include mdc-theme-prop(color, primary);
```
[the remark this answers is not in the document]
moog16: MR conflict as well

PR #1846, opened 2018-01-02
feat(text-field): Add dense mode to outlined text field

bonniezhou:
- Check to see if text field has `--dense` class and calculate label width accordingly
- Add `$scale` argument to `mdc-text-field-invalid-label-shake-keyframes_` mixin
- Add dense option to demo

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -41,24 +41,24 @@
   }
 }
 
-@mixin mdc-text-field-invalid-label-shake-keyframes_($modifier, $positionY) {
+@mixin mdc-text-field-invalid-label-shake-keyframes_($modifier, $positionY, $scale:.75) {
   @keyframes invalid-shake-float-above-#{$modifier} {
     0% {
-      transform: translateX(0) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(0) translateY(-#{$positionY}) scale(#{$scale}, #{$scale});
     }
 
     33% {
       animation-timing-function: cubic-bezier(.5, 0, .701732, .495819);
-      transform: translateX(5px) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(5px) translateY(-#{$positionY}) scale(#{$scale}, #{$scale});
     }
 
     66% {
       animation-timing-function: cubic-bezier(.302435, .381352, .55, .956352);
-      transform: translateX(-5px) translateY(-#{$positionY}) scale(.75, .75);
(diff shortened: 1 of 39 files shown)


packages/mdc-textfield/foundation.js
```diff
@@ -159,7 +160,10 @@ class MDCTextFieldFoundation extends MDCFoundation {
     if (!this.outline_ || !this.label_) {
       return;
     }
-    const labelWidth = this.label_.getFloatingWidth();
+
+    const isDense = this.adapter_.hasClass(cssClasses.DENSE);
+    const labelScale = isDense ? 0.923 : 0.75;
```
kfranqueiro: Should we create a `numbers` object in `constants.js` for these?

packages/mdc-textfield/_mixins.scss
```diff
@@ -41,24 +41,24 @@
   }
 }
 
-@mixin mdc-text-field-invalid-label-shake-keyframes_($modifier, $positionY) {
+@mixin mdc-text-field-invalid-label-shake-keyframes_($modifier, $positionY, $scale:.75) {
   @keyframes invalid-shake-float-above-#{$modifier} {
     0% {
-      transform: translateX(0) translateY(-#{$positionY}) scale(.75, .75);
+      transform: translateX(0) translateY(-#{$positionY}) scale(#{$scale}, #{$scale});
```
kfranqueiro: It should be fine to only specify one scale parameter, since `scale` defaults to using the same number for both dimensions if only one is specified. (See also the instances of this in mdc-text-field.scss)

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -173,6 +171,17 @@
     stroke-width: 2px;
   }
 
+  &:not(.mdc-text-field--invalid) {
```
kfranqueiro: Is this fixing a different issue?

bonniezhou: Yeah, I moved all the spec changes over to #1855

packages/mdc-textfield/foundation.js
bonniezhou: Good idea. Done!

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -198,6 +198,29 @@
...
+
+    .mdc-text-field__input {
+      padding: 12px 12px 7px;
+    }
+
+    .mdc-text-field__label {
+      bottom: 18px;
+
+      &--float-above {
+        transform: translateY(calc(-122% - 2px)) scale(.923, .923);
```
kfranqueiro: Remove second `.923`

Issue #1526, opened 2017-11-01
Add Sass color mixin to text field


lynnjepsen:
- [ ] Create color mixins
- [ ] Call color mixin from stylesheet, with primary color
- [ ] Delete any —primary --accent modifiers
- [ ] Update README to document color mixins
- [ ] Include "breaking change" message about removing modifiers, point to color mixin as deprecation replacement.

dscalzi:
Any progress?

PR #1847, opened 2018-01-02
BREAKING CHANGE: Move text-field colors to mixins

williamernest:
Breaking Change: Move text-field colors from text-field.scss to _mixins.scss. 
refs: #1526


lynnjepsen:
I'm starting to review this with Will offline. We're going to look into ways of splitting this up more. And we will probably wait for #1831 to go in.

PR #1851, opened 2018-01-03
feat/tab/sass color mixins

moog16:
fixes: #1152
related: https://github.com/material-components/material-components-web/pull/1801

### DIFF
--- a/demos/tabs.scss
+++ b/demos/tabs.scss
@@ -20,20 +20,31 @@
 @import "../packages/mdc-tabs/mdc-tabs";
 
 #basic-tab-bar-custom-label-color .mdc-tab--active,
-#css-tab-bar-custom-label-color .mdc-tab--active,
-#icon-tab-bar-custom-icon-color .mdc-tab--active,
-#css-icon-tab-bar-custom-icon-color .mdc-tab--active,
-#icon-tab-bar-custom-icon-color .mdc-tab--active,
-#icon-text-tab-bar-custom-colors .mdc-tab--active,
-#css-icon-text-tab-bar-custom-colors .mdc-tab--active {
+#css-tab-bar-custom-label-color .mdc-tab--active {
   @include mdc-tab-label-ink-color($material-color-indigo-500);
 }
 
-#css-tab-bar-custom-label-color .mdc-tab:not(.mdc-tab--active),
 #basic-tab-bar-custom-label-color .mdc-tab:not(.mdc-tab--active),
+#css-tab-bar-custom-label-color .mdc-tab:not(.mdc-tab--active) {
+  @include mdc-tab-label-ink-color($material-color-deep-purple-200);
+}
+
+#icon-tab-bar-custom-icon-color .mdc-tab--active,
+#css-icon-tab-bar-custom-icon-color .mdc-tab--active {
+  @include mdc-tab-icon-ink-color($material-color-indigo-500);
+}
+
 #icon-tab-bar-custom-icon-color .mdc-tab:not(.mdc-tab--active),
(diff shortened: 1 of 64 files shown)


packages/mdc-tabs/README.md
```diff
@@ -211,6 +211,19 @@ dots.addEventListener('click', function (evt) {
 })
 ```
 
+### Sass Mixins
+
+To customize the label ink color or the icon ink color you can use the following mixins. You would use these mixins on the `.mdc-tab:not(.mdc-tab--active)` selector or on the `.mdc-tab--active` selector to alter label or icon colors in the activated state.
+
+#### `mdc-tab-ink-color`
+Use this mixin to set the color of icon and label to have the same color. Do not use this if you are using the `mdc-tab-label-ink-color` or `mdc-tab-icon-ink-color` mixins.
```
lynnjepsen: Change this to:

Use this mixin to set the color of all ink on the tab.

...This give us flexibility to add new "ink" components in the future.

packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -46,7 +48,6 @@ $mdc-tab-with-icon-and-text-height: 72px;
   overflow: hidden;
   vertical-align: middle;
 
-  @include mdc-theme-prop(color, text-secondary-on-light);
 
   &:hover {
     @include mdc-theme-prop(color, text-primary-on-light);
```
lynnjepsen: this should use `@include mdc-tab-ink-color(text-primary-on-light)` as well

packages/mdc-tabs/README.md
```diff
@@ -211,6 +211,19 @@ dots.addEventListener('click', function (evt) {
 })
 ```
 
+### Sass Mixins
+
+To customize the label ink color or the icon ink color you can use the following mixins. You would use these mixins on the `.mdc-tab:not(.mdc-tab--active)` selector or on the `.mdc-tab--active` selector to alter label or icon colors in the activated state.
+
```
lynnjepsen: To customize the ink color of any part of the tab, use the following mixins. We recommend you apply these mixins within CSS selectors like `.foo-tab:not(.mdc-tab--active)` to select your inactive tabs, and `.foo-tab.mdc-tab--active` to select your active tabs.

moog16: I definitely have to get better at Readmes

Thanks!

demos/tabs.scss
```diff
@@ -24,27 +24,42 @@
   @include mdc-tab-label-ink-color($material-color-indigo-500);
 }
 
-#basic-tab-bar-custom-label-color .mdc-tab:not(.mdc-tab--active),
-#css-tab-bar-custom-label-color .mdc-tab:not(.mdc-tab--active) {
-  @include mdc-tab-label-ink-color($material-color-deep-purple-200);
+#basic-tab-bar-custom-label-color,
```
lynnjepsen: hmm, i didn't catch this in my first review, but I don't think we should be using ID CSS selectors in our demo pages. Its not what we would recommend our users to use. We would recommend they use CSS classes. So we should do the same.

so like:

````
.custom-label-color-tab {
  &:not(.mdc-tab--active) {
    @include mdc-tab-label-ink-color(purple);
  }
  &:hover {
    @include mdc-tab-label-ink-color(indigo);
  }
  &.mdc-tab--active {
    @include mdc-tab-label-ink-color(indigo);
  }
}

...and the same for all the CSS below

packages/mdc-tabs/README.md
lynnjepsen: Maybe also mention the `:hover` case as well.

PR #1858, opened 2018-01-03
feat(text-field): Handle leading/trailing icons in outlined text field

### LINKED ISSUE #1602: Update outlined text field to handle leading and trailing icon
Update outlined text field to handle leading and trailing icon. This is a CSS update. (I don't think there are any JS involved here....)

bonniezhou:
Fixes #1602.
- adjust label x-position
- add outlined text field with icons to demo
- remove unnecessary z-indices
- remove note about setting input width when using icons (the width of the parent text field can be set as usual, there's no special case for incorporating with icons)

### DIFF
--- a/demos/text-field.html
+++ b/demos/text-field.html
@@ -48,11 +48,7 @@
         margin: 20px 0;
         background-color: #f2f2f2;
       }
-
-      .custom-text-field-input {
-        width: 300px;
-      }
-
+      
       .full-width-textarea-example {
         margin-top: 16px;
       }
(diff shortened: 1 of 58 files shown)


demos/text-field.html
```diff
@@ -275,60 +271,71 @@ <h2>Text Field box</h2>
...
             <i class="material-icons mdc-text-field__icon" tabindex="0">delete</i>
             <div class="mdc-text-field__bottom-line"></div>
           </div>
         </div>
+        <div id="demo-tf-outlined-leading-wrapper">
+          <div id="tf-outlined-leading-example"
+            class="mdc-text-field mdc-text-field--outlined mdc-text-field--with-leading-icon" data-demo-no-auto-js>
+            <input type="text" id="tf-outlined-leading" class="mdc-text-field__input">
+            <label for="tf-outlined-leading" class="mdc-text-field__label">Your other name</label>
+            <i class="material-icons mdc-text-field__icon" tabindex="0">event</i>
```
kfranqueiro: This element should be before the input and label in order for tab order to make sense (focus should move to icon and then text field for leading icon).

demos/text-field.html
```diff
@@ -48,11 +48,7 @@
         margin: 20px 0;
         background-color: #f2f2f2;
       }
-
-      .custom-text-field-input {
-        width: 300px;
-      }
-
+
```
kfranqueiro: Remove trailing spaces on this line

demos/text-field.html
```diff
@@ -275,60 +271,71 @@ <h2>Text Field box</h2>
 
       <section class="example">
         <h2>Text Field - Leading/Trailing icons</h2>
-        <div class="demo-note">
```
kfranqueiro: Are we sure about removing this advice? I think the intent was to tell people how to make text fields consistent size both with and without icons in them, though that's not really visible from this demo example alone since we only use the class on the icon examples. It might make more sense to move this to a note within the icon readme?

bonniezhou: Yep, Alex confirmed it's fine since the input elements have `width:100%`.
Making text fields consistent size both with and without icons is done the same way (by setting the width of the top-level `mdc-text-field` element).

bonniezhou:
Moved the label shake changes into another branch (PR will be sent out after this one merges).
I kept the z-index changes since the icon's z-index has to be set higher than the input's, and the input's z-index has to be higher than outline elements.

bonniezhou:
Re: the icon README, will send a follow-up PR to add examples with outlined text field

Issue #1795, opened 2017-12-16
Add auto-scrolling to the mdc-drag-drop module


lynnjepsen:
When you start dragging an object and move the cursor to the bottom of the screen, it should scroll the page

lynnjepsen:
Re-opening this issue, since the code is not committed to master

lynnjepsen:
Closing for #1915

PR #1859, opened 2018-01-03
Experimental/dnd scroll

moog16:
fixes: https://github.com/material-components/material-components-web/issues/1795

### DIFF
--- a/packages/mdc-drag-drop/drag-collection/index.js
+++ b/packages/mdc-drag-drop/drag-collection/index.js
@@ -538,13 +538,12 @@ export class MDCDragCollection extends MDCComponent {
     // (e.originalEvent || e.detail.originalEvent).preventDefault();
     const scrollZone = MDCDragCollection.getScrollZone_(e, this.scrollZones_);
 
-    if(!scrollZone) {
-      this.stopScrolling();
-    } else {
+    if(scrollZone) {
       this.handleScrollZone_(scrollZone);
       return;
     }
 
+    this.stopScrolling();
     this.handleDragZone_(e);
   }
 
@@ -608,12 +607,11 @@ export class MDCDragCollection extends MDCComponent {
   handleDragStop_(e) {
     this.resetDropShifts_(this.draggableItemList_);
     this.removeDraggingStateElements_();
+    this.stopScrolling();
 
     if (this.activeDropZone_) {
       this.dropItLikeItsHot_(e);
     }
-
-    this.stopScrolling();
 
     if (this.sourceItemEl_ && this.sourceItemEl_.ripple) {
       // TODO(acdvorak): Submit PR to "draggable" repo to pass through originalEvent to drag:stop
@@ -825,13 +823,9 @@ export class MDCDragCollection extends MDCComponent {
 
   static getScrollZone_(e, scrollZones) {
     const pointerPositionInViewport = util.getPointerPositionInViewport(e);
-    for (let i = 0; i < scrollZones.length; i++) {
-      const curScrollZone = scrollZones[i];
-      if (curScrollZone.pointInsideOrExtendsRect(pointerPositionInViewport)) {
-        return curScrollZone;
-      }
-    }
-    return null;
+    return scrollZones.find(scrollZone => {
+       return scrollZone.pointInsideOrExtendsRect(pointerPositionInViewport);
+    });
   }
 
   /**
--- a/packages/mdc-drag-drop/drag-collection/index.js
+++ b/packages/mdc-drag-drop/drag-collection/index.js
@@ -538,12 +538,15 @@ export class MDCDragCollection extends MDCComponent {
     // (e.originalEvent || e.detail.originalEvent).preventDefault();
     const scrollZone = MDCDragCollection.getScrollZone_(e, this.scrollZones_);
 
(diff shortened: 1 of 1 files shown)


packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -584,6 +634,13 @@ export class MDCDragCollection extends MDCComponent {
     this.resetState_();
   }
 
+  stopScrolling() {
+    this.activeScrollZone_ = null;
+    if (this.activeScrollTimer_) {
```
acdvorak: I think we can skip this check - `cancelAnimationFrame` will silently returns if you pass it an invalid value.

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -756,6 +813,21 @@ export class MDCDragCollection extends MDCComponent {
...
+    return Object.keys(PageSides)
+      .map(k => {
+        const side = PageSides[k];
+        return new ScrollZone({type: side});
+      });
+  }
+
+  static getScrollZone_(e, scrollZones) {
+    const pointerPositionInViewport = util.getPointerPositionInViewport(e);
+    return scrollZones.find(scrollZone => {
```
acdvorak: IE 11 doesn't support `Array#find` 🙁

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -525,7 +536,18 @@ export class MDCDragCollection extends MDCComponent {
 
   handleDragMove_(e) {
     // (e.originalEvent || e.detail.originalEvent).preventDefault();
+    const scrollZone = MDCDragCollection.getScrollZone_(e, this.scrollZones_);
 
+    if(scrollZone) {
+      this.handleScrollZone_(scrollZone);
+      return;
```
acdvorak: I can't decide if we should handle both scrolling _and_ drag-drop at the same time... Currently, when the user scrolls down the page while dragging, they have to stop scrolling and then move their finger a bit to trigger the drop zone to activate. On the other hand, if we removed this `return`, the user might drop a card somewhere accidentally... Maybe add a `TODO(moog16)` comment here with a note explaining that we should revisit whether to remove the `return` statement.

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -525,7 +536,18 @@ export class MDCDragCollection extends MDCComponent {
 
   handleDragMove_(e) {
     // (e.originalEvent || e.detail.originalEvent).preventDefault();
+    const scrollZone = MDCDragCollection.getScrollZone_(e, this.scrollZones_);
 
+    if(scrollZone) {
```
acdvorak: Add a space between `if (`

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -525,7 +536,18 @@ export class MDCDragCollection extends MDCComponent {
...
+      this.handleScrollZone_(scrollZone);
+      return;
+    }
+
+    this.stopScrolling();
```
acdvorak: Move this above the `if (scrollZone)` check

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -555,9 +577,37 @@ export class MDCDragCollection extends MDCComponent {
     }
   }
 
+  handleScrollZone_(scrollZone) {
+    if(!this.scrollZones_) {
```
acdvorak: Add a space between `if (`

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -555,9 +577,37 @@ export class MDCDragCollection extends MDCComponent {
...
+      return;
+    }
+
+    const isActiveScrollZone = this.activeScrollZone_ && this.activeScrollZone_.type === scrollZone.type;
+
+    if (isActiveScrollZone) {
+      return;
+    }
+
+    this.stopScrolling();
```
acdvorak: Remove this line, since we should always be calling it in `handleDragMove_`

packages/mdc-drag-drop/drag-collection/index.js
```diff
@@ -555,9 +577,37 @@ export class MDCDragCollection extends MDCComponent {
...
+    this.activeScrollZone_ = scrollZone;
+
+    const scrollLoop = () => {
+      const { type } = scrollZone;
```
acdvorak: `const {type}` (remove spaces around `type`)

packages/mdc-drag-drop/drag-collection/index.js
moog16: Will do...yes I also wrestled with this. I switched the logic twice. The behavior I saw made me think that scrolling should overrule the dropzone because if the user activated a dropzone but is at very top/bottom page, the page wouldn't scroll. I did not try attempting both, since I thought it would be confusing to the user. And as you said would unknowingly drop a card.

packages/mdc-drag-drop/drag-collection/index.js
moog16: :( Ok I'll revert to a for loop

PR #1860, opened 2018-01-03
fix(text-field): updated dependency check test and added special case…

moog16:
fixes dependency pkg check test forn text-field package.json update

### DIFF
--- a/scripts/check-pkg-for-release.js
+++ b/scripts/check-pkg-for-release.js
@@ -130,10 +130,7 @@ function checkPkgDependencyAddedInMDCPackage() {
 
 function checkCSSDependencyAddedInMDCPackage() {
   const name = getPkgName();
-  let nameMDC = pkg.name.replace('@material/', 'mdc-');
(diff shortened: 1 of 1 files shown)


scripts/check-pkg-for-release.js
```diff
@@ -129,9 +129,9 @@ function checkPkgDependencyAddedInMDCPackage() {
 }
 
 function checkCSSDependencyAddedInMDCPackage() {
-  const name = pkg.name.split('/')[1];
+  const name = getPkgName();
   let nameMDC = pkg.name.replace('@material/', 'mdc-');
-  if (name === 'textfield') {
+  if (name === 'text-field') {
```
lynnjepsen: I don't think we need this if statement at all anymore. 'mdc-' plus 'text-field' equals 'mdc-text-field'

moog16: I removed the if condition and updated it to use `name` variable

PR #1740, opened 2017-12-08
feat(textfield): Expand the Foundation and Component public API

bwobrien:
Also adds private helper methods for applying style based on the component state.

PR #1873, opened 2018-01-04
feat(text-field): Add properties for value, disable, value, and required

lynnjepsen:
This is a copy from the original #1740...but fixed up after some merge-hell with mdc-textfield directory


packages/mdc-textfield/adapter.js
```diff
@@ -28,7 +28,10 @@ import MDCTextFieldLabelFoundation from './label/foundation';
  *   value: string,
  *   disabled: boolean,
  *   badInput: boolean,
- *   checkValidity: (function(): boolean)
+ *   validity: {
```
lynnjepsen: This will need a BREAKING CHANGE comment when you go to merge, so don't forget!

packages/mdc-textfield/foundation.js
```diff
@@ -181,21 +183,20 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
+    this.styleFocused_(this.isFocused_);
     if (this.bottomLine_) {
       this.bottomLine_.activate();
     }
     if (this.outline_) {
       this.updateOutline();
     }
     if (this.label_) {
-      this.label_.floatAbove();
+      this.label_.style(undefined, undefined, undefined, this.isFocused_);
```
lynnjepsen: I find this part so confusing....3 undefined arguments in a row :(

Is there a way to break up the `label.style` method so it doesn't need all 4 arguments?

packages/mdc-textfield/foundation.js
```diff
@@ -233,54 +234,63 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
     const shouldRemoveLabelFloat = !input.value && !this.isBadInput_();
-
-    this.isFocused_ = false;
-    this.adapter_.removeClass(FOCUSED);
+    const isValid = this.isValid();
+    this.styleValidity_(isValid);
+    this.styleFocused_(this.isFocused_);
     if (this.label_) {
-      this.label_.deactivateFocus(shouldRemoveLabelFloat);
+      this.label_.style(this.getValue(), isValid, this.isBadInput_(), this.isFocused_);
```
lynnjepsen: Here as well, I find it confusing

packages/mdc-textfield/label/foundation.js
```diff
@@ -58,32 +58,32 @@ class MDCTextFieldLabelFoundation extends MDCFoundation {
...
+   * Styles the label to match the supplied optional params. If a
+   * param is not supplied, the method makes no changes related to that param.
+   * @param {string=} optValue
+   * @param {boolean=} optIsValid
+   * @param {boolean=} optIsBadInput
+   * @param {boolean=} optIsFocused
    */
-  deactivateFocus(shouldRemoveLabelFloat) {
-    this.adapter_.removeClass(cssClasses.LABEL_SHAKE);
+  style(optValue, optIsValid, optIsBadInput, optIsFocused) {
```
lynnjepsen: Document this in the mdc-textfield/label/README.md

...actually see my above comment first, about maybe making this more clear...but then we should document all new public methods in mdc-textfield/label/README.md

packages/mdc-textfield/adapter.js
bwobrien: Ack!

packages/mdc-textfield/foundation.js
bwobrien: Broke up `style` into `styleFloat` and `styleShake`. Neither method allows `undefined` arguments anymore. I agree I think this is clearer now.

packages/mdc-textfield/foundation.js
bwobrien: Done.

packages/mdc-textfield/label/foundation.js
bwobrien: Done.

PR #1882, opened 2018-01-05
fix(text-field): Fix label shake animation

### LINKED ISSUE #1818: Incorrect translateY value for label shake animation

Latest

Chrome, all

### What OS are you using?

macOS Sierra

1. Fire up demo server
1. Type one letter into outline text field demo
1. Blur outline text field demo
1. Observe animation

Label will shake back and forth in the x direction as detailed in spec for text field box.

Label does shake in the x direction, but also moves in the y direction.

invalid-animation

### LINKED ISSUE #1867: Text field label shrinks in shake animation when on dense mode
Latest
Chrome, all
### What OS are you using?
macOS
1. Go to http://material-components-web.appspot.com/text-field.html
2. Select the 'Dense' checkbox under Text Field Box
2. Type a letter into Text Field Box
3. Blur Text Field Box
4. Observe label animation
The label will not shrink.
The label shrinks as part of the shake animation.

### LINKED ISSUE #1923: Textarea label shake position is incorrect for invalid fields

> 0.28.0

> All (Tested on Chome, Firefox, Mac Safari)

> 1. Create a mdc textarea as below.
```
<div class="mdc-text-field mdc-text-field--textarea">
  <textarea id="textarea" class="mdc-text-field__input" rows="8" cols="40"></textarea>
  <label for="textarea" class="mdc-text-field__label">Textarea Label</label>
</div>
```

> 2. Add html5 validation (**minlength="8"**).
```
<div class="mdc-text-field mdc-text-field--textarea">
  <textarea id="textarea" class="mdc-text-field__input" rows="8" cols="40" minlength="8"></textarea>
  <label for="textarea" class="mdc-text-field__label">Textarea Label</label>
</div>
```
> 3. Type fewer than 8 characters in textrea and tap outside. 
> 4. Invalid label will shake, but its position is incorrect.

> Textarea label should be visible when invalid.

> When textarea gets invalid, label position will not behave as expected.

Video: https://drive.google.com/open?id=1C_9bdYC9rnCMtWdcxAuouWiKv1MQu3ld

bonniezhou:
Fixes #1818
Fixes #1867
Fixes #1923

Added a "required" checkbox option to demo label shake animation with leading/trailing icons.
Cleaned up the CSS to be more consistent/readable.


demos/text-field.html
```diff
@@ -533,6 +537,18 @@ <h2>Full-Width Text Field and Textarea</h2>
...
         });
 
+        document.getElementById('required-leading-trailing').addEventListener('change', function(evt) {
+          var target = evt.target;
+          tfBoxLeadingEl.querySelector('.mdc-text-field__input').required = target.checked;
+          tfBoxTrailingEl.querySelector('.mdc-text-field__input').required = target.checked;
+          tfOutlinedLeadingEl.querySelector('.mdc-text-field__input').required = target.checked;
+          tfOutlinedTrailingEl.querySelector('.mdc-text-field__input').required = target.checked;
+
+          tfBoxLeadingEl.querySelector('.mdc-text-field__input').pattern = target.checked ? ".{8,}" : "*";
```
acdvorak: `*` is not a valid regex; we should probably just set it to an empty string instead (ditto below).

Also, I think we normally use single quotes instead of double quotes (ditto below).

acdvorak:
Other than the `pattern` comment this LGTM, but I wasn't able to repro either of the bugs that this is fixing, so perhaps someone else should take a look at this PR.

demos/text-field.html
bonniezhou: That would mean the input is valid iff it matches the empty string, so I changed it to `.*` instead.

acdvorak:
I'm able to repro now, but this PR doesn't seem to fix the issue for me 🙁 

Make sure to test this with textarea too (#1923)

bonniezhou:
Thanks! Re-ordered the transforms to be consistent (translate first, then scale). Also added the keyframes for textarea label shake to fix #1923.

packages/mdc-textfield/label/mdc-text-field-label.scss
```diff
@@ -20,7 +20,6 @@
 @import "@material/theme/mixins";
 
 @include mdc-text-field-invalid-label-shake_keyframes_(standard, 100%);
```
moog16: Should this one also be in `packages/mdc-textfield/_mixins.scss`?

PR #1886, opened 2018-01-05
feat(theme): Switch to new theme demo page

acdvorak:
The `MDC_GENERATE_DEMO_THEMES` env var is now opt-out (i.e., it's `true` by default)


packages/mdc-drawer/README.md
```diff
@@ -403,7 +403,7 @@ Temporary drawers can easily be initialized using their default constructors as
 ```javascript
 import {MDCTemporaryDrawer} from '@material/drawer';
 
-const drawer = new MDCTemporaryDrawer(document.querySelector('.mdc-temporary-drawer'));
+const drawer = new MDCTemporaryDrawer(document.querySelector('.mdc-drawer--temporary'));
```
lynnjepsen: doesnt look like this should be in this PR

acdvorak: This should have been updated in #1730 but we missed it :-\

PR #1890, opened 2018-01-07
feat(toolbar): Apply icon color to SVG icons' fill

### LINKED ISSUE #1821: [Toolbar] SVG Icons have wrong color
When I use SVG icons in a MDC toolbar (instead of the material-icon font based icons, which the MDC Toolbar demos use), the icons are always black, but should be white (or rather: `--mdc-theme-text-primary-on-primary`).

This [CodePen](https://codepen.io/MichaelTamm/pen/OzRXwZ) demonstrates the issue.

The reason is, that in the file `mdc-toolbar.scss` you only style the `color` property, but for SVG icons the `fill` property must be styled.

If you agree, that adding styling for the `fill` property (in order to support SVG icons) would be a worthy addition to the project, I would like to provide a pull request.

@material/toolbar 0.27.0

all

MichaelTamm:
Fixes #1821

### DIFF
--- a/packages/mdc-toolbar/_mixins.scss
+++ b/packages/mdc-toolbar/_mixins.scss
@@ -35,7 +35,8 @@
 }
 
 @mixin mdc-toolbar-icon-ink-color($color) {
-  @include mdc-theme-prop(color, $color);
+  @include mdc-theme-prop(color, $color); // for font icons
+  @include mdc-theme-prop(fill, $color); // for SVG icons
 }
 
 //


flundermicha:
Ping

MichaelTamm:
Ping

packages/mdc-toolbar/_mixins.scss
```diff
@@ -35,7 +35,8 @@
 }
 
 @mixin mdc-toolbar-icon-ink-color($color) {
-  @include mdc-theme-prop(color, $color);
+  @include mdc-theme-prop(color, $color); // for font icons
+  @include mdc-theme-prop(fill, $color); // for SVG icons
```
kfranqueiro: I think it should be possible to do this centrally rather than upon each application of this mixin, by setting `fill: currentColor` in the `mdc-toolbar-icon_` mixin instead?

Also CC @williamernest in case he knows of any reason setting `fill: currentColor` might be problematic in any future planned toolbar instances

williamernest:
Thanks for submitting this PR. We're fixing this using `fill: currentColor` in #2331.

PR #1898, opened 2018-01-08
feat(icon-toggle): Added support for SVG icons

MichaelTamm:


flundermicha:
Ping

MichaelTamm:
Ping

kfranqueiro:
Sorry for letting this PR lapse for so long. This has now become obsoleted by the mdc-icon-button package (see #2748), which now has basic SVG support.

However, I think this also attempted to implement the ability to support different SVGs for on/off states, which Icon Button currently doesn't. This has been identified in #3107 with a proposed alternative approach to the component's DOM structure.

PR #1911, opened 2018-01-09
chore(text-field): Move idle outline style method

patrickrodee:
Remove `getIdleOutlineStyleValue` method from textfield adapter. Add
`getIdleOutlineStyleValue` method to textfield outline adapter. Update
tests to check new functionality.

BREAKING CHANGE: Text field outline adapter now must implement the `getIdleOutlineStyleValue` method previously implemented in the text field adapter. The functionality is exactly the same and requires only small changes to accessing the outline node.

### DIFF
--- a/packages/mdc-textfield/outline/index.js
+++ b/packages/mdc-textfield/outline/index.js
@@ -52,6 +52,12 @@ class MDCTextFieldOutline extends MDCComponent {
         const path = this.root_.querySelector(strings.PATH_SELECTOR);
         path.setAttribute('d', value);
       },
+      getIdleOutlineStyleValue: (propertyName) => {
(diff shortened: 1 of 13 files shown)


packages/mdc-textfield/outline/index.js
```diff
@@ -52,6 +52,12 @@ class MDCTextFieldOutline extends MDCComponent {
         const path = this.root_.querySelector(strings.PATH_SELECTOR);
         path.setAttribute('d', value);
       },
+      getIdleOutlineStyleValue: (propertyName) => {
+        const idleOutlineElement = this.root_.parentNode.querySelector(strings.IDLE_OUTLINE_SELECTOR);
```
patrickrodee: This is the only real funkiness. Since the idle outline is a sibling of the outline root, we need to access the parent node to be able to query it.

PR #1919, opened 2018-01-10
feat(demos): Add global `demoReady()` function

acdvorak:
For this PR, I only updated the button demo page to use the new `demoReady()` function. #1920 updates the remaining demo pages.

Also note that our packaged JS is now being loaded with `async`, which speeds up initial rendering time over slow connections. Using "Mid-tier mobile" network emulation in Chrome devtools, the [TTFMP](https://developers.google.com/web/tools/lighthouse/audits/first-meaningful-paint) of the button demo page was reduced from ~10.5s to ~9.5s, and the time to render the Material icon font  dropped from ~12s to ~10.5s.

### DIFF
--- a/demos/ready.js
+++ b/demos/ready.js
@@ -87,7 +87,7 @@ window.demoReady = (function() {
   }
 
   function invokeHandlers() {
-    handlers.forEach((handler) => {
+    handlers.forEach(function(handler) {
       handler();
     });
   }
--- a/demos/ready.js
+++ b/demos/ready.js
@@ -61,10 +61,9 @@ window.demoReady = (function() {
   }
 
   function startTimer() {
-    if (isReady() || timer) {
+    if (timer) {
       return;
     }
-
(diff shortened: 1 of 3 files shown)


demos/ready.js
```diff
@@ -0,0 +1,104 @@
...
+  function removeDetectionDom() {
+    if (!testDom) {
+      return;
+    }
+    document.body.removeChild(testDom);
+    testDom = null;
+  }
+
+  function startTimer() {
+    if (isReady() || timer) {
```
bonniezhou: If I'm following the logic correctly...I think this conditional can be removed since
- it seems like `timer` will always be null here
- in the case where `isReady()` is true but `timer` is null, the handler will never get called

acdvorak: Good catch! Removed the `isReady()` check.

`timer` will only be `null` the first time `addHandler()` is called, so I think we need to keep that check.

PR #1928, opened 2018-01-10
chore(linear-progress): Create variable for bar height

acdvorak:


acdvorak:
I can remove the `!default` to keep them read-only, but I think it's useful to be able to query a component's dimensions - e.g., to animate it in and out of view (which I need to do on the new theme demo page).

acdvorak:
After discussing offline, I can avoid needing to read this value from the theme demo page. Closing PR.

PR #1963, opened 2018-01-16
feat(text-field): Move text-field outline colors to mixins

williamernest:
Breaking Change: Moves color customization of the outline text-field to SASS mixins. 

refs: #1526

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -15,7 +15,9 @@
 //
 
 @import "bottom-line/mixins";
+@import "outline/mixins";
 @import "label/mixins";
+@import "helper-text/mixins";
 
 @mixin mdc-text-field-outlined-corner-radius($radius) {
   border-radius: $radius;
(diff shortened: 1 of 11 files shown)


packages/mdc-textfield/_mixins.scss
```diff
@@ -72,6 +74,12 @@
   }
 }
 
+@mixin mdc-text-field-input-hover-outline-stroke-color($color) {
```
lynnjepsen: Ah this took me a while to figure out....I think you should add a comment about why this is not the same as `mdc-text-field-hover-outline-stroke-color`. Where does this get used?

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -151,7 +157,7 @@
     z-index: 2;
 
     &:hover ~ .mdc-text-field__idle-outline {
-      border: 1px solid $mdc-text-field-outlined-hover-border;
+      border: 1px solid;
```
lynnjepsen: Is this line necessary? Seems the same as line 136 above

williamernest: Line 136 is for the input:hover and this one is for icon:hover. 

They were both set to the same color, but I didn't want to assume that would always be the case.

PR #1965, opened 2018-01-16
chore(tabs): move indicator sass into custom mixins

moog16:
Partial fix of: #1152 
TODO: Update README.md to reflect changes


demos/tabs.scss
```diff
@@ -57,3 +58,11 @@
     @include mdc-tab-ink-color($material-color-indigo-500);
   }
 }
+
+.custom-indicator-tab-bar {
+  @include mdc-tab-bar-indicator-ink-color(primary);
+}
+
+.custom-toolbar-indicator .custom-indicator-toolbar-tab-bar {
+  @include mdc-tab-bar-indicator-ink-color(secondary);
```
lynnjepsen: this example seems to do the same thing as above...do we really need it?

packages/mdc-tabs/tab-bar/_mixins.scss
```diff
@@ -0,0 +1,8 @@
+@import "@material/theme/mixins";
+
+@mixin mdc-tab-bar-indicator-ink-color($color) {
+  .mdc-tab-bar__indicator,
+  &:not(.mdc-tab-bar-upgraded) .mdc-tab::after {
```
lynnjepsen: I worry about & used within a comma separated selector.

Can you double check this `&:not(.mdc-tab-bar-upgraded) .mdc-tab::after` works with the css only version of tabs on our demo page?

packages/mdc-tabs/tab-bar/mdc-tab-bar.scss
```diff
@@ -47,6 +48,7 @@
 
   // postcss-bem-linter: ignore
   .mdc-toolbar & {
```
lynnjepsen: What does all this code do...the stuff under `.mdc-toolbar &`....which I assume translates to `.mdc-toolbar .mdc-tab-bar`

Why do we need so much code to get a tab bar to work within a toolbar? Maybe we don't need this now that we're customizing color through mixins?

packages/mdc-tabs/tab-bar/mdc-tab-bar.scss
```diff
@@ -78,34 +78,3 @@
 .mdc-tab-bar--icons-with-text {
   height: 72px;
 }
-
-.mdc-tab-bar--indicator-primary,
```
lynnjepsen: removing these --primary and --accent is a breaking change, so remember to add that to the commit

demos/tabs.scss
moog16: Its similar, but a different color (primary vs secondary). Also we need it to target the toolbar specifier. If we don't have that we cannot override the color. If we remove the toolbar scss, we can probably get around this.

packages/mdc-tabs/tab-bar/mdc-tab-bar.scss
moog16: We could remove this and then have mixins to customize toolbar specific tab-bars. I like that idea. This will be a breaking change though, so I'll check the rest of the repo.

demos/tabs.scss
moog16: I'll update this after I remove toolbar specific sass

moog16: Addressed in https://github.com/material-components/material-components-web/pull/1979/

packages/mdc-tabs/tab-bar/mdc-tab-bar.scss
moog16: Addressed in https://github.com/material-components/material-components-web/pull/1979/

packages/mdc-tabs/tab-bar/_mixins.scss
```diff
@@ -2,7 +2,7 @@
 
 @mixin mdc-tab-bar-indicator-ink-color($color) {
   .mdc-tab-bar__indicator,
-  &:not(.mdc-tab-bar-upgraded) .mdc-tab::after {
+  &.mdc-tab-bar:not(.mdc-tab-bar-upgraded) .mdc-tab::after {
```
lynnjepsen: HMmmmmmmmmm...this is tricky

Remind me, why do we use `.mdc-tab::after` and not `.mdc-tab-bar__indicator` for the CSS only version of tabs?

moog16: Going to look into removing the .mdc-tab::after element in favor of .mdc-tab-bar__indicator. Will open pre-pr if possible

PR #1974, opened 2018-01-17
chore(demos): RTL-aware margins on theme page

acdvorak:
To test, add `dir="rtl"` to the `<html>` element in the Chrome dev tools.


acdvorak:
Fair point that RTL isn't theme-related 🙂   I can remove the RTL toggle button from #1975.

That said, we might as well merge this PR - it showcases best practices and doesn't increase complexity.

_FYI: I'm actually planning to add an RTL toggle to every demo page (except maybe theme and elevation). That should make it easier for us to find RTL issues, and will improve the consistency of our demo pages. Only a few pages have RTL toggles right now, and they're all in different places._

PR #1983, opened 2018-01-17
refactor(tabs): removed ::after for css-only indicators

moog16:
BREAKING CHANGE: all css-only mdc-tabs must have a .mdc-tab-bar__indicator
child element


packages/mdc-tabs/tab-bar/mdc-tab-bar.scss
```diff
@@ -18,6 +18,16 @@
 @import "@material/theme/mixins";
 @import "@material/rtl/mixins";
 
+.mdc-tab-bar__indicator,
+.mdc-tab__indicator {
```
moog16: not sure if this should be in mdc-tab.scss, or if its ok to be in here.

lynnjepsen: It should be in mdc-tab.scss.

Since you want to share CSS properties across the two classes, you should create a private mixin and re-use it across the two files.

packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -109,56 +109,33 @@ $mdc-tab-with-icon-and-text-height: 72px;
...
-  &:not(.mdc-tab--active):active::after {
+  &:not(.mdc-tab--active):active .mdc-tab__indicator {
     opacity: .87;
   }
 }
 // stylelint-enable plugin/selector-bem-pattern
 
-// stylelint-disable plugin/selector-bem-pattern
-.mdc-tab-bar--icons-with-text:not(.mdc-tab-bar-upgraded) .mdc-tab::after {
-  top: $mdc-tab-with-icon-and-text-height - 2px;
```
lynnjepsen: Did you keep this logic somewhere?

moog16: This is no longer needed since its addressed in the .mdc-tab__indicator { bottom: 2px; } rule

packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -110,28 +111,36 @@ $mdc-tab-with-icon-and-text-height: 72px;
...
-    opacity: .38;
-  }
+    &:not(.mdc-tab--active):hover .mdc-tab__indicator {
+      opacity: .38;
+    }
 
-  &--active,
-  &:not(.mdc-tab--active):active .mdc-tab__indicator {
-    opacity: .87;
+    &--active,
```
lynnjepsen: remove this `&` syntax, it is confusing

PR #1987, opened 2018-01-18
chore(checkbox): Remove !important modifiers

patrickrodee:
GSS chokes on !important and it's not good practice in general. Let's remove them.

### DIFF
--- a/packages/mdc-checkbox/_mixins.scss
+++ b/packages/mdc-checkbox/_mixins.scss
@@ -69,9 +69,7 @@
 
 @mixin mdc-checkbox-ink-color($color) {
   .mdc-checkbox__checkmark__path {
-    // !important is needed here because a stroke must be set as an attribute on the SVG in order
-    // for line animation to work properly.
-    @include mdc-theme-prop(stroke, $color, $important: true);
+    @include mdc-theme-prop(stroke, $color);
   }
 
   .mdc-checkbox__mixedmark {
@@ -137,11 +135,7 @@
 }
 
 @mixin mdc-checkbox__child--upgraded_ {
-  // Due to the myriad of selector combos used to properly style a CSS-only checkbox, all of
-  // which have varying selector precedence and make use of transitions, it is cleaner and more
-  // efficient here to simply use !important, since the mdc-checkbox--anim-* classes will take
-  // over from here.
-  transition: none !important;
+  transition: none;
 }
 
 // Animation


packages/mdc-checkbox/_mixins.scss
```diff
@@ -69,9 +69,7 @@
 
 @mixin mdc-checkbox-ink-color($color) {
   .mdc-checkbox__checkmark__path {
-    // !important is needed here because a stroke must be set as an attribute on the SVG in order
-    // for line animation to work properly.
-    @include mdc-theme-prop(stroke, $color, $important: true);
+    @include mdc-theme-prop(stroke, $color);
```
lynnjepsen: does line animation still work?

packages/mdc-checkbox/_mixins.scss
```diff
@@ -137,11 +135,7 @@
 }
 
 @mixin mdc-checkbox__child--upgraded_ {
-  // Due to the myriad of selector combos used to properly style a CSS-only checkbox, all of
-  // which have varying selector precedence and make use of transitions, it is cleaner and more
-  // efficient here to simply use !important, since the mdc-checkbox--anim-* classes will take
-  // over from here.
-  transition: none !important;
+  transition: none;
```
lynnjepsen: hmmm... this note above is not super helpful. Do the animations still work?

packages/mdc-checkbox/_mixins.scss
acdvorak: Nope - removing this `!important` breaks the line animation when toggling from `indeterminate` to `determinate` on an unchecked checkbox ☹️ 

If you slow animations down to 10% in the Chrome Dev Tools, and click "Toggle indeterminate" near an unchecked checkbox, you can see it:

pr 1987 indeterminate bug

It's supposed to look like this:

pr 1987 indeterminate prod

packages/mdc-checkbox/_mixins.scss
acdvorak: It looks like the animations still work.

AFAICT this `!important` was only needed for `--theme-dark`, which we're getting rid of anyway.

CSS-only checkboxes use the `transition` property to change states, whereas JS-enabled checkboxes use the `animation` property. This line simply turns off the CSS-only transition style for JS-enabled checkboxes.

packages/mdc-checkbox/_mixins.scss
patrickrodee: Good catch. I'll find a different way to accomplish this.

patrickrodee: FYI I think your comments apply to the opposite lines. Just for future reference. Gonna link them.

packages/mdc-checkbox/_mixins.scss
patrickrodee: https://github.com/material-components/material-components-web/pull/1987#discussion_r163142119

packages/mdc-checkbox/_mixins.scss
patrickrodee: https://github.com/material-components/material-components-web/pull/1987#discussion_r163142706

williamernest:
Bump! Is this something we still want, or something we should close?

patrickrodee:
Still want! Can’t use it internally without them removed. It’d be best to
not have to manually modify this every time we import it.

williamernest:
Sounds good. Can we get it in soon? It's been open for over 2 months.

patrickrodee:
Sure, it’s gonna require a considerable rework of checkbox animation
though. This can be labeled tech debt.

PR #2002, opened 2018-01-19
feat(text-field): Add more demos

williamernest:
Add more text-field demos for css only elements.

### DIFF
--- a/demos/text-field.html
+++ b/demos/text-field.html
@@ -112,11 +112,11 @@
           <label for="dense">Dense</label>
         </div>
         <div>
-          <input id="required" type="checkbox">
+          <input id="required" class="required" type="checkbox">
           <label for="required">Required</label>
         </div>
         <div>
-          <input id="alternate-colors" type="checkbox">
+          <input id="alternate-colors" class="alternate-colors" type="checkbox">
           <label for="alternate-colors">Alternate Colors</label>
         </div>
         <div>
@@ -168,8 +168,7 @@
             </div>
             <div class="mdc-text-field__idle-outline"></div>
           </div>
-          <p class="mdc-text-field-helper-text mdc-text-field-helper-text--validation-msg"
-            id="name-validation-msg">
+          <p class="mdc-text-field-helper-text mdc-text-field-helper-text--validation-msg">
             Must be at least 8 characters
           </p>
         </div>
@@ -205,10 +204,6 @@
               }
               tf.layout();
             });
-            document.getElementById('outlined-custom-colors').addEventListener('change', function(evt) {
-              tfEl.classList[evt.target.checked ? 'add' : 'remove']('demo-text-field-custom-colors');
-              tfEl.classList[evt.target.checked ? 'add' : 'remove']('demo-text-field-custom-error-colors');
-            });
             document.getElementById('outlined-dense').addEventListener('change', function(evt) {
               tfEl.classList[evt.target.checked ? 'add' : 'remove']('mdc-text-field--dense');
               tf.layout();
@@ -221,7 +216,7 @@
         <h2>Text Field box</h2>
         <div id="demo-tf-box-wrapper">
           <div id="tf-box-example" class="mdc-text-field mdc-text-field--box" data-demo-no-auto-js>
-            <input required pattern=".{8,}" type="text" id="tf-box" class="mdc-text-field__input"
+            <input type="text" id="tf-box" class="mdc-text-field__input"
                    aria-controls="name-validation-message">
             <label for="tf-box" class="mdc-text-field__label">Your Name</label>
             <div class="mdc-text-field__bottom-line"></div>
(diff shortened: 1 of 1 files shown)


demos/text-field.html
```diff
@@ -358,19 +361,26 @@ <h2>CSS Only Text Field</h2>
...
             <input type="text" class="mdc-text-field__input" id="css-only-text-field"
                    placeholder="Name">
-            <div class="mdc-text-field__bottom-line"></div>
           </div>
           <label for="css-only-text-field">Your name:</label>
         </div>
 
         <div>
-          <input id="css-only-tf-disabled" type="checkbox">
+          <input id="css-only-tf-disabled" class="demo-disabled" type="checkbox">
```
kfranqueiro: Curious why `demo-` prefix is used on the classes for disabled and dense checkboxes when it's not used for the others?

demos/text-field.html
```diff
@@ -382,6 +392,22 @@ <h2>CSS Only Text Field box</h2>
                    placeholder="Name">
           </div>
         </div>
+        <div>
+          <input type="checkbox" class="demo-disabled" id="css-only-box-disable">
```
kfranqueiro: Nit: This page currently has a mix of "disable" and "disabled" for checkbox IDs... can we make them consistent?

demos/text-field.html
```diff
@@ -679,6 +731,56 @@ <h2>Full-Width Text Field and Textarea</h2>
...
+                while (!box.classList.contains('example')) {
+                  if(box.parentElement === document.body) {
+                    return null;
+                  }
+                  box = box.parentElement;
+                }
+                return box;
+              })(checkbox);
+
+              parentSection.querySelectorAll(elementToExecuteCallbackOn)
```
kfranqueiro: Does this even need to use qSA or is it only ever being used to operate on a single element?

If you do need qSA, wrap in `[].slice.call` to work in all browsers, e.g. `[].slice.call(parentSection.querySelectorAll(...)).forEach(...)`

demos/text-field.html
```diff
@@ -679,6 +731,56 @@ <h2>Full-Width Text Field and Textarea</h2>
           section.classList[target.checked ? 'add' : 'remove']('mdc-theme--dark');
         });
       });
+
+      demoReady(function() {
+
+        // Initializes all checkboxes to find closest "example" class (section), finds the elementToExecuteCallbackOn
+        // and calls a function with the event and each element found.
+        var findElementAndAddClickHandlerToCheckBox = function(checkboxClassName, elementToExecuteCallbackOn, handlerCallback) {
```
kfranqueiro: Shorten some names here? `addCheckboxHandlers` for function name (we're not actually listening for click as the name currently suggests anyway)? `targetSelector` for second argument?

demos/text-field.html
```diff
@@ -679,6 +731,56 @@ <h2>Full-Width Text Field and Textarea</h2>
...
+        });
+
+        findElementAndAddClickHandlerToCheckBox('alternate-colors', '.mdc-text-field', function(evt, wrapper) {
+          var target = evt.target;
+          wrapper.classList[target.checked ? 'add' : 'remove']('demo-text-field-custom-colors');
+          wrapper.classList[target.checked ? 'add' : 'remove']('demo-text-field-custom-error-colors');
+        });
+
+        findElementAndAddClickHandlerToCheckBox('demo-disabled', '.mdc-text-field__input', function(evt, input) {
+          evt.target.checked ? input.setAttribute("disabled", "disabled") : input.removeAttribute("disabled");
```
kfranqueiro: Double quotes -> single quotes

demos/text-field.html
```diff
@@ -205,10 +204,6 @@ <h2>Outlined Text Field</h2>
               }
               tf.layout();
             });
-            document.getElementById('outlined-custom-colors').addEventListener('change', function(evt) {
```
kfranqueiro: You removed this code but it looks like you didn't add the equivalent class to use your new functions, so the "alternate colors" checkbox under outlined text field currently doesn't seem to do anything.

Moreover, should it also be possible to remove the outlined-dense and outlined-disable handlers here and use your centralized code for that, and ditto for the disabled, dense, and alternate colors checkboxes for the first example on the page?

demos/text-field.html
```diff
@@ -679,6 +731,56 @@ <h2>Full-Width Text Field and Textarea</h2>
           section.classList[target.checked ? 'add' : 'remove']('mdc-theme--dark');
         });
       });
+
+      demoReady(function() {
+
+        // Initializes all checkboxes to find closest "example" class (section), finds the elementToExecuteCallbackOn
+        // and calls a function with the event and each element found.
+        var findElementAndAddClickHandlerToCheckBox = function(checkboxClassName, elementToExecuteCallbackOn, handlerCallback) {
+          var checkboxes = Array.from(document.getElementsByClassName(checkboxClassName));
```
kfranqueiro: I'm pretty sure `Array.from` won't work in IE11; see my other comment about wrapping qSA in `[].slice.call`

PR #2007, opened 2018-01-19
fix(text-field): removed --float-above from --shake selectors

moog16:


packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -186,11 +186,10 @@
   }
 
   .mdc-text-field__label--float-above {
-    animation: invalid-shake-float-above-box 250ms 1;
```
moog16: This bug I introduced last PR, which needs to be removed

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -187,7 +187,7 @@
     transform: translateY(-50%) scale(.75);
```
bonniezhou: This seems to be a duplicate of the selector just above it.

moog16:
yup totally right...idk how i missed those all. Must have been a merge conflict. :|

packages/mdc-textfield/mdc-text-field.scss
moog16: :(

PR #2011, opened 2018-01-19
chore(demos): Combine JS module bundles on theme page

acdvorak:
The advantage of having a single bundle (compiled JS binary) is that we don't need to store global mutable state on `window` or `document`; JS modules can simply use closure-protected variables like normal.

The only downside is that `demo.common.util` and friends are no longer available for debugging in the browser dev tools.


acdvorak:
You're right, `HotSwapper` is basically a singleton. The problem is that the module gets compiled into two separate output JS files, so there's no way to share variables between those files without attaching something to `window` or `document`.

Is there a better way that I'm missing?

kfranqueiro:
Discussed offline, I think the current tale-of-2-hotswappers problem stems from theme/index.html loading common.js and theme/index.js individually; if the latter imports the former internally instead and theme/index.html only loads theme/index.js, I think there shouldn't be a problem.

acdvorak:
Done - PTAL! 🙂

PR #2014, opened 2018-01-20
feat(select): Move colors to mixins

williamernest:
Move select colors to mixins. Update demo.

### DIFF
--- a/demos/select.scss
+++ b/demos/select.scss
@@ -29,5 +29,11 @@
   // Focused colors
   @include mdc-select-focused-bottom-line-color(green);
   @include mdc-select-focused-label-color(green);
+
+  // Menu colors
+  @include mdc-select-menu-item-color(blue);
+  @include mdc-select-menu-selected-item-color(green);
+  @include mdc-select-menu-group-color(yellow);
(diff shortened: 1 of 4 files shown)


demos/select.scss
```diff
@@ -29,5 +29,11 @@
   // Focused colors
   @include mdc-select-focused-bottom-line-color(green);
   @include mdc-select-focused-label-color(green);
+
+  // Menu colors
+  @include mdc-select-menu-item-color(blue);
```
lynnjepsen: Instead of calling new mixins, and creating new mixins....just call the existing mdc-list mixins with the `mdc-select__menu` selector

```
.mdc-select__menu .mdc-list-item {
  @include mdc-list-item-primary-text-ink-color(blue);
}
```

...if a mixin in mdc-list does not exist, then we should wait for more customization in mdc-list. So don't add new mixins to mdc-select

demos/select.scss
```diff
@@ -29,5 +29,11 @@
...
+  @include mdc-select-menu-selected-item-color(green);
```
lynnjepsen: For selected menu items...lets wait for a resolution on #2031.

PR #2016, opened 2018-01-20
style(demos): Minor code cleanup

acdvorak:
- Reorder `ready.js` in markup
- Inline some variables that are only used once
- Simplify `[].slice.call(...).forEach(...)` to `[].forEach.call(...)`
- Add a few blank lines


demos/theme/index.html
```diff
@@ -979,31 +979,31 @@ <h3 class="mdc-typography--headline demo-component-section__heading">
          * Slider
```
bonniezhou: I think you missed Select?

acdvorak: Good catch!

PR #2018, opened 2018-01-20
fix(demos): Sanitize slider input values

acdvorak:
The demo page now checks that the user's input values are finite numbers before assigning them to any of the component's setter properties. This prevents an error from being thrown when the user deletes all the text in one of the inputs (which is a fairly common thing to do before entering a new number).

In addition, to make it clearer that the slider supports negative numbers for `min` and `max`, the demo page now allows those input values to go down to `-100` (previously they were bounded at `0`).


acdvorak:
Removed string parsing from the foundation.

The reason I added it is that many standard DOM APIs automatically coerce string values to numbers (e.g., `scrollTop = '50'`), which is useful in UIs where component parameters often depend on user input. But since there's no particular need for it, I'll remove it and just keep the demo changes.

demos/slider.html
```diff
@@ -309,25 +313,37 @@ <h2>Custom Colored Discrete Slider with Tick Marks</h2>
...
+          customDiscreteWMarkerSlider.min = minValue;
         });
 
-        max.addEventListener('input', function() {
-          continuousSlider.max = parseFloat(max.value);
-          discreteSlider.max = parseFloat(max.value);
-          discreteWMarkerSlider.max = parseFloat(max.value);
-          customDiscreteWMarkerSlider.max = parseFloat(max.value);
+        maxInput.addEventListener('input', function() {
+          var minValue = parseFloat(maxInput.value);
```
kfranqueiro: `maxValue`

acdvorak: Done

PR #2019, opened 2018-01-20
feat(demos): Add consistent RTL toggle to most demo pages

acdvorak:
Every component that supports RTL now has an RTL toggle button in the top-right corner of their demo page toolbar. Any existing RTL toggle buttons/checkboxes that were scattered around the page have been removed, as they are no longer needed.

On all pages except the drawer demos, the top toolbar is the only element on the page that is _not_ affected by toggling RTL. This ensures that you can flip back and forth quickly without having to move the mouse.

The specific demo pages now have an RTL toggle:

- Card
- Checkbox
- Dialog
- Drawer
- Grid List
- Layout Grid
- Linear Progress
- List
- Select
- Simple Menu
- Slider
- Snackbar
- Tabs
- Text Field
- Toolbar

Pages not specifically listed above do not require an RTL toggle because there's nothing to demonstrate.

### DIFF
--- a/demos/common.scss
+++ b/demos/common.scss
@@ -26,7 +26,7 @@ $mdc-theme-secondary: #018786 !default; // baseline teal, 600 tone
 
 // Import button and ripple mixins to apply overrides for dark theme
 // TODO(kfranqueiro): Pending further design discussion around how to manage dark theme
-@import "../packages/mdc-button/mixins";
+@import "../packages/mdc-button/mdc-button";
 @import "../packages/mdc-ripple/mixins";
 // All demo pages have a top toolbar, and most of them use theme variables.
 // Import these *after* setting theme colors to override defaults in mdc-theme.
--- a/packages/mdc-dialog/README.md
+++ b/packages/mdc-dialog/README.md
@@ -204,7 +204,7 @@ dialog.listen('MDCDialog:cancel', function() {
   console.log('canceled');
 })
 
-document.querySelector('#default-dialog-activation').addEventListener('click', function (evt) {
+document.querySelector('#show-dialog-button').addEventListener('click', function (evt) {
   dialog.lastFocusedTarget = evt.target;
   dialog.show();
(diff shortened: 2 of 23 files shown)


demos/common.scss
```diff
@@ -26,7 +26,7 @@ $mdc-theme-secondary: #018786 !default; // baseline teal, 600 tone
 
 // Import button and ripple mixins to apply overrides for dark theme
 // TODO(kfranqueiro): Pending further design discussion around how to manage dark theme
-@import "../packages/mdc-button/mixins";
+@import "../packages/mdc-button/mdc-button";
```
lynnjepsen: why did you make this change?

packages/mdc-dialog/README.md
```diff
@@ -204,7 +204,7 @@ dialog.listen('MDCDialog:cancel', function() {
   console.log('canceled');
 })
 
-document.querySelector('#default-dialog-activation').addEventListener('click', function (evt) {
+document.querySelector('#show-dialog-button').addEventListener('click', function (evt) {
```
lynnjepsen: why did you make this change?

demos/common.scss
kfranqueiro: Funny, I had wondered this same thing (and PS this will be removed when we finish removing dark mode anyway)... but I also talked to Andy about this yesterday, mainly about the fact that this PR seems to have a lot of unrelated changes in it, and I think the consensus was he would split this up?

acdvorak: A lot of our demo pages have buttons on them. But yes, I will split up this PR

packages/mdc-dialog/README.md
acdvorak: I just did a search-and-replace in the whole project and didn't catch this

acdvorak:
It's probably not worth the time and energy to split up this PR. Closing for now.

PR #2024, opened 2018-01-20
fix(base): `shouldBubble` param now defaults to `true` in `emit()`

acdvorak:
BREAKING CHANGE: The `shouldBubble` param of `MDCComponent#emit()` is now opt-out rather than opt-in. This matches the behavior of most standard Web events. Without this change, there's no way to listen to delegated events.

### DIFF
--- a/packages/mdc-base/README.md
+++ b/packages/mdc-base/README.md
@@ -172,7 +172,7 @@ export class MyComponent extends MDCComponent {
 | `destroy()` | Subclasses may override this method if they wish to perform any additional cleanup work when a component is destroyed. For example, a component may want to deregister a window resize listener. |
 | `listen(type: string, handler: EventListener)` | Adds an event listener to the component's root node for the given `type`. Note that this is simply a proxy to `this.root_.addEventListener`. |
 | `unlisten(type: string, handler: EventListener)` | Removes an event listener from the component's root node. Note that this is simply a proxy to `this.root_.removeEventListener`. |
-| `emit(type: string, data: Object, shouldBubble: boolean = false)` | Dispatches a custom event of type `type` with detail `data` from the component's root node. It also takes an optional shouldBubble argument to specify if the event should bubble. This is the preferred way of dispatching events within our vanilla components. |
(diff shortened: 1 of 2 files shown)


acdvorak:
My goal is to make all (or most) of our custom events behave like native events, which typically bubble.

Without bubbling, you need to manually bind an event listener to every individual component on the page, which is silly. Delegated events solve that problem, but not if we prevent bubbling by default.

I'd be curious to know if there's a reason why we weren't already doing this; perhaps it goofs up some wrapper framework?

kfranqueiro:
"Behave like native events" could mean bubble or don't bubble, depending on what sample set you pick. Native `input`, `change`, `focus`, and `blur` events don't bubble, and many of our custom events are analogous to `input` and `change`.

I checked with Travis who pointed out that custom events created via DOM APIs also don't bubble by default, which was one of the guiding factors for the current decision for our own API. Moreover, it probably makes sense to isolate events at the component level since they serve specific purposes anyway. This was another reason behind the current behavior.

I surveyed occurrences of `this.emit` and they generally seem correct as-is, not bubbling:

* dialog accept and cancel events - unlikely to merit delegation across dialogs
* drawer open and close events - unlikely to merit delegation (you'd probably only have one drawer)
* icon toggle change event - not bubbling is consistent with native change events
* menu selected/cancel events - selected is already effectively delegated across items in the menu; unlikely to want to delegate the same handler across menus?
* select change event - not bubbling is consistent with native change events
* slider input and change events - not bubbling is consistent with native input and change events
* snackbar show/hide events - unlikely to merit delegation since you can potentially reuse the same snackbar
* tab selected events - explicitly bubble
* tab bar change event - not bubbling is consistent with native change events
* text field bottom line animation end event - IIRC this is mainly for use by the top-level text field component so it doesn't make sense to expose it to the outside world via bubbling
* text field icon action event - explicitly bubbles, so that it _is_ exposed to the outside world (otherwise it's buried in a subcomponent); bubbling seems consistent with e.g. click events anyway
* flexible toolbar change event - unlikely to merit delegation (you'd probably only have one toolbar)

acdvorak:
Good points Ken 😄

PR #2025, opened 2018-01-20
feat(card): Add theme mixins; remove content layouts

### LINKED ISSUE #1126: mdc-card should have background color

0.17.0

All browsers

### What OS are you using?

Mac OS Sierra

You can view this behavior on the demo page. https://material-components-web.appspot.com/card.html
Just open the chrome inspector and change the background color
image

The material spec shows all cards have a background by default.

https://material.io/guidelines/components/cards.html#

transparent cards

acdvorak:
BREAKING CHANGE: All CSS classes for content layouts have been removed. Clients should decide what kind of layout is best for their specific use case. Dark theme CSS classes have been removed; use the Sass mixin or custom CSS instead.

Fixes #1126

Updated README to match best practices.

### Removed classes (12)

- `mdc-card--theme-dark`
- `mdc-card__primary`
- `mdc-card__title`
- `mdc-card__title--large`
- `mdc-card__subtitle`
- `mdc-card__supporting-text`
- `mdc-card__media-item`
- `mdc-card__media-item--1dot5x`
- `mdc-card__media-item--2x`
- `mdc-card__media-item--3x`
- `mdc-card__actions--vertical`
- `mdc-card__horizontal-block`

### New classes (8)

- `mdc-card__media--square`
- `mdc-card__media--16-9`
- `mdc-card__media-content`
- `mdc-card__actions--full-bleed`
- `mdc-card__action-buttons`
- `mdc-card__action-icons`
- `mdc-card__action--button`
- `mdc-card__action--icon`

### New mixins (3)

- `mdc-card-fill-color($color)`
- `mdc-card-corner-radius($radius)`
- `mdc-card-media-aspect-ratio($x, $y)`

### Screenshot (edited)

Annotated screenshot of new Card demo page


packages/mdc-card/README.md
```diff
@@ -39,200 +39,146 @@ npm install --save @material/card
...
+`mdc-card__action-buttons` | A group of action buttons, displayed on the left side of the card (in LTR), adjacent to `mdc-card__action-icons`
+`mdc-card__action-icons` | A group of supplemental action icons, displayed on the right side of the card (in LTR), adjacent to `__action-buttons`
+`mdc-card__action` | An individual action button or icon
+`mdc-card__action--icon` | An icon action
+
+### Sass Mixins
+
+Mixin | Description
+--- | ---
+`mdc-card-fill-color($color)` | Sets the fill color of a `mdc-card`
```
kfranqueiro: Just say card instead of `mdc-card` here, like you do on the next line?

packages/mdc-card/README.md
```diff
@@ -39,200 +39,146 @@ npm install --save @material/card
...
+`mdc-card-corner-radius($radius)` | Sets the corner radius of a card
+`mdc-card-media-aspect-ratio($width-unitless, $height-unitless)` | Automatically sets the height of the `mdc-card__media` subelement to maintain a given aspect ratio based on its width
```
kfranqueiro: `-unitless` sounds a bit awkward and I'm not sure it's needed... would just `$x` and `$y` seem any better?

Maybe it'd be better to re-word the description to something like "Maintains the given aspect ratio by setting the height of the `mdc-card__media` subelement based on its width"?

packages/mdc-card/mdc-card.scss
```diff
@@ -15,174 +15,151 @@
...
+.mdc-card__media:last-child {
+  border-bottom-left-radius: inherit;
+  border-bottom-right-radius: inherit;
+}
 
-    box-sizing: border-box;
-    padding: 8px 16px;
+.mdc-card__media--square {
+  @include mdc-card-media-aspect-ratio(1, 1);
+  @include mdc-card-media-aspect-ratio-base_;
```
kfranqueiro: Shouldn't `...-aspect-ratio-base_` actually be included under just `.mdc-card__media` instead of in each of these modifiers? Otherwise how would users properly customize this without also applying one of these two modifiers first in order to get the base styles (since the base mixin is private)?

There is also no example of this in the demo page to prove that the exposed mixin is sufficient.

On the other hand, if we really only want people to use these two ratios, then we shouldn't expose the aspect-ratio mixin at all.

packages/mdc-card/mdc-card.scss
```diff
@@ -15,174 +15,151 @@
...
+}
 
-    // postcss-bem-linter: ignore
-    .mdc-card__media-item--3x {
-      margin-bottom: 16px;
-    }
-  }
+// Icon toggles are taller than buttons, so we need to adjust their margins to prevent the action row from expanding.
+// stylelint-disable-next-line plugin/selector-bem-pattern
+.mdc-card__action--icon.mdc-icon-toggle {
```
kfranqueiro: Would it be preferable to use 2 classes both internal to this package if we need to override specificity, rather than needing to reference a class in another package? (I realize the alternative also breaks BEM though...) cc @lynnjepsen

demos/card.html
```diff
@@ -127,210 +38,226 @@
...
+                 tabindex="0"
+                 role="button"
+                 aria-pressed="false"
+                 aria-label="Add to favorites"
+                 title="Add to favorites"
+                 data-toggle-on='{"content": "favorite", "label": "Remove from favorites"}'
+                 data-toggle-off='{"content": "favorite_border", "label": "Add to favorites"}'>
+                favorite_border
+              </i>
+              <button class="mdc-button mdc-card__action mdc-card__action--icon" title="Share">
```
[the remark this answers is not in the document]
acdvorak: Removed the `mdc-button` wrappers. Now they're just plain icons with unbounded ripples.

I was originally going to add `overflow: hidden` to `mdc-card`, but that would break overflow menus, selects, and probably lots more stuff in the future (e.g., tooltips).

Also, it looks weird when the icon toggle ripple gets truncated by the card.

demos/card.html
```diff
@@ -127,210 +38,226 @@
...
-            <section class="mdc-card__actions">
-              <button class="mdc-button mdc-button--compact mdc-card__action">Action 1</button>
-            </section>
+      <section class="demo-card-collection">
+        <div class="mdc-card demo-card">
+          <div class="demo-card-article-group-heading mdc-typography--subheading2">Headlines</div>
+
+          <hr class="mdc-list-divider">
+
+          <a class="demo-card-article mdc-ripple-surface" href="javascript:">
```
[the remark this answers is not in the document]
acdvorak: Done

demos/card.html
```diff
@@ -127,210 +38,226 @@
...
-            <div class="mdc-card__horizontal-block">
-              <section class="mdc-card__primary">
-                <h1 class="mdc-card__title mdc-card__title--large">Title here</h1>
-                <h2 class="mdc-card__subtitle">Subtitle here</h2>
-              </section>
-              <img class="mdc-card__media-item mdc-card__media-item--1dot5x" src="images/1-1.jpg">
+          </section>
+          <section class="mdc-card__actions">
+            <div class="mdc-card__action-icons demo-card__action-icons--centered">
+              <i class="mdc-icon-toggle material-icons mdc-card__action mdc-card__action--icon"
```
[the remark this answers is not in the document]
acdvorak: Fixed; see [my comment](https://github.com/material-components/material-components-web/pull/2025#discussion_r163692909) above

demos/card.html
```diff
@@ -127,210 +38,226 @@
...
-      // formField.input = cb;
-      var demoWrapper = document.getElementById('demo-wrapper');
-      document.getElementById('toggle-rtl').addEventListener('change', function () {
-        if (this.checked) {
-          demoWrapper.setAttribute('dir', 'rtl');
-        } else {
-          demoWrapper.removeAttribute('dir');
-        }
+      demoReady(function() {
+        document.addEventListener('change', function(evt) {
```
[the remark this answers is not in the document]
acdvorak: Done

demos/card.html
```diff
@@ -127,210 +38,226 @@
...
+        [].forEach.call(document.querySelectorAll('.mdc-button'), function(surface) {
+          mdc.ripple.MDCRipple.attachTo(surface);
+        });
+
+        [].forEach.call(document.querySelectorAll('.mdc-icon-toggle'), function(iconToggle) {
+          mdc.iconToggle.MDCIconToggle.attachTo(iconToggle);
+        });
+
+        document.addEventListener('MDCIconToggle:change', function(evt) {
+          // TODO(acdvorak): Do this in icon-toggle
```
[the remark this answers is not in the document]
acdvorak: I'm not sure it's a bug so much as a missing feature. We don't use the `title` attribute anywhere in our JS.

Personally, I think `title` attributes are **essential** on icons; if you don't already know what a hamburger icon means, you have no idea what's going to happen when you click on it.

Removed the TODO because we should fix this sort of thing during an accessibility audit.

demos/card.scss
```diff
@@ -19,3 +19,149 @@
...
+}
+
+.demo-card-toggles {
+  margin: 24px;
+  padding: 24px;
+  text-align: center;
+}
+
+//
+// Small
```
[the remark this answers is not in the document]
acdvorak: Done

demos/card.scss
```diff
@@ -19,3 +19,149 @@
...
+.demo-card__action-icons--centered {
+  justify-content: center;
+}
+
+.demo-card__action-buttons--text {
+  padding: 8px;
+}
+
+//
+// Music
```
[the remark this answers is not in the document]
acdvorak: Done

demos/theme/index.html
```diff
@@ -359,46 +359,44 @@ <h3 class="mdc-typography--headline demo-component-section__heading">
...
+              </div>
             </section>
             <section class="mdc-card__actions">
-              <button class="mdc-button mdc-button--compact mdc-card__action">Action 1</button>
+              <div class="mdc-card__action-buttons">
+                <button class="mdc-button mdc-card__action">Read</button>
+                <button class="mdc-button mdc-card__action">Bookmark</button>
+              </div>
+              <div class="mdc-card__action-icons">
+                <i class="mdc-icon-toggle material-icons mdc-card__action mdc-card__action--icon"
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-card/README.md
```diff
@@ -39,200 +39,146 @@ npm install --save @material/card
...
-  <section class="mdc-card__actions">
-    <button class="mdc-button mdc-button--compact mdc-card__action">Action</button>
-  </section>
+<div class="mdc-card__actions">
+  <div class="mdc-card__action-buttons">
+    <button class="mdc-button mdc-card__action">Read</button>
+    <button class="mdc-button mdc-card__action">Bookmark</button>
+  </div>
+  <div class="mdc-card__action-icons">
+    <i class="mdc-icon-toggle material-icons mdc-card__action mdc-card__action--icon"
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-card/README.md
```diff
@@ -39,200 +39,146 @@ npm install --save @material/card
...
 
-#### Media items
+Be sure to include the `mdc-card__action` class on every action for correct positioning. In addition, icon actions
+should use the `mdc-card__action--icon` class.
 
-Media items are designed to be used in horizontal blocks, taking up a fixed height, rather than stretching to the width
-of the card.
-
-> Note: Unlike media blocks, media items are not designed to be used as optional containers.
+To have a single action button take up the entire width of the action row, use the `--full-bleed` modifier:
```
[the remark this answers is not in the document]
acdvorak: "Full bleed" is a standard term in printing and graphic design. I don't think the card spec mentions it by name, tho; it only has a mock.

packages/mdc-card/README.md
```diff
@@ -39,200 +39,146 @@ npm install --save @material/card
...
 
 ```html
-<div class="mdc-card__horizontal-block">
-  <section class="mdc-card__primary">
-    <h1 class="mdc-card__title mdc-card__title--large">Title here</h1>
-    <h2 class="mdc-card__subtitle">Subtitle here</h2>
-  </section>
-  <img class="mdc-card__media-item mdc-card__media-item--1x" src="image.jpg">
+<div class="mdc-card__actions mdc-card__actions--full-bleed">
+  <a class="mdc-button mdc-card__action" href="javascript:">
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-card/README.md
```diff
@@ -39,200 +39,146 @@ npm install --save @material/card
...
-| ------------------------------- | ---------------------------------------------------------------- |
-| `mdc-card__media-item`          | Defines a media item and sets it to the default height of 80px.  |
-| `mdc-card__media-item--1dot5x`  | Sets the media item height to 120px.                             |
-| `mdc-card__media-item--2x`      | Sets the media item height to 160px.                             |
-| `mdc-card__media-item--3x`      | Sets the media item height to 240px.                             |
+### CSS Classes
+
+CSS Class | Description
+--- | ---
+`mdc-card` | A card
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-card/README.md
acdvorak: Done

packages/mdc-card/README.md
acdvorak: Done

packages/mdc-card/mdc-card.scss
acdvorak: Done

packages/mdc-card/mdc-card.scss
acdvorak: In this particular case, I think that might be a tad overkill.

Clients already need to add `mdc-card__action` and `mdc-card__action--icon`,
so a third class (which is only necessary if the icon is _also_ an `mdc-icon-toggle`)
would get kinda verbose:

```html
<i class="material-icons mdc-card__action mdc-card__action--icon mdc-card__action--icon-toggle" tabindex="0" role="button" title="Share">share</i>
```

That said, if you or Lynn feel strongly that we should have another modifier class instead, I can certainly add one.

demos/card.html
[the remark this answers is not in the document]
acdvorak: Created #2085 to do this globally across all demo pages that import `demos/common.js`

demos/card.scss
```diff
@@ -19,3 +19,167 @@
...
+//
+// Focusable child elements
+//
+
+.demo-card__ripple-surface {
+  @include mdc-ripple-surface;
+  @include mdc-ripple-radius;
+  @include mdc-states;
+
+  &:not([data-mdc-ripple-is-unbounded]) {
```
kfranqueiro: Would it make more sense to just apply these styles to `.demo-card-article` instead?

demos/theme/index.html
```diff
@@ -360,46 +360,35 @@ <h3 class="mdc-typography--headline demo-component-section__heading">
...
+                   tabindex="0"
+                   role="button"
+                   aria-pressed="false"
+                   aria-label="Add to favorites"
+                   title="Add to favorites"
+                   data-toggle-on='{"content": "favorite", "label": "Remove from favorites"}'
+                   data-toggle-off='{"content": "favorite_border", "label": "Add to favorites"}'>
+                  favorite_border
+                </i>
+                <i class="material-icons mdc-card__action mdc-card__action--icon mdc-ripple-surface"
```
kfranqueiro: Can we do something to make the padding of these consistent with the icon-toggle? They currently mismatch (and I'm fixing the unbounded ripple size in #2092 which will make the icon-toggle ripple the correct size).

demos/card.scss
acdvorak: Potentially, but if we ever decide to add another card example that uses
`demo-card__ripple-surface` on a different element, it won't get the benefit of the
IE/Edge workaround (`position: relative`).

demos/theme/index.html
acdvorak: Done

acdvorak:
Good catch on the RTL corner radius. Fixed it in the demo.

packages/mdc-card/_mixins.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+//
+// Private
+//
+
+@mixin mdc-card-box-sizing_ {
+  &,
+  &::before,
```
kfranqueiro: Do we really need the `::before` and `::after` selectors on every single thing this mixin is applied to? I would only expect it to be needed on before/after for `__action` and possibly `card` itself assuming we have ripples on the whole card at some point.

packages/mdc-card/_mixins.scss
```diff
@@ -0,0 +1,66 @@
...
+  &::after {
+    box-sizing: border-box;
+  }
+}
+
+@mixin mdc-card-column-layout_($display: flex) {
```
kfranqueiro: What is the point of the `display` parameter existing if we never set it and we pretty much assume it's flex given the presence of flex-direction, anyway?

Actually, do we even need this mixin? AFAICT these styles don't affect anything on `__media` and `__media-content` so they're only actually needed once, on `mdc-card`...

packages/mdc-card/_mixins.scss
acdvorak: There's no direct need for it ATM, but I wanna try to start consistently applying `border-box` to all of our elements (and pseudo elements) so that there are no surprises in the future (for us _or_ our clients).

I can't tell you how many times I've gone to add `padding` to something in MDC Web and had it end up bigger than I expected because of stupid `content-box`. I want to get to a place where we can assume that all of our elements are `border-box` by default, unless we specifically override it for some reason.

WDYT?

kfranqueiro: I guess, but most of our elements don't have `::before` and `::after` unless they have ripples, which is why I thought that in particular seemed excessive. Keep in mind how much CSS this is adding because we're including this mixin multiple times.

Maybe add a parameter to this mixin to conditionalize the inclusion of before/after in the selector? (There's an example of how to conditionalize the selector in the focus state styles in mdc-ripple/_mixins)

packages/mdc-card/_mixins.scss
acdvorak: Removed! Good catch!

packages/mdc-card/_mixins.scss
acdvorak: As discussed offline: Libraries/frameworks generally should not mess with `::before` and `::after` unless absolutely necessary (e.g., mdc-ripple). This is to prevent conflicts with other libraries and client code which might _also_ set styles on the pseudos.

Removed the `mdc-card-box-sizing_` mixin and directly applied `box-sizing: border-box` to each element individually.

PR #2033, opened 2018-01-22
Docs/list/new readme

moog16:
Rewrote MDC List Readme.md to follow Readme Best Practices https://github.com/material-components/material-components-web/blob/master/docs/code/readme_standards.md


packages/mdc-list/README.md
```diff
@@ -1,49 +1,23 @@
-<!--docs:
```
bonniezhou: Was this removed by accident? I thought we need it as front matter for publishing this on material.io?

packages/mdc-list/README.md
```diff
@@ -1,49 +1,23 @@
...
-layout: detail
-section: components
-excerpt: "Lists present multiple line items vertically as a single continuous element."
-iconId: list
-path: /catalog/lists/
--->
-
 # Lists
 
-<!--<div class="article__asset">
```
bonniezhou: Same with these comments (see the [template](https://github.com/material-components/material-components-web/blob/master/docs/code/readme_template.md))

packages/mdc-list/README.md
```diff
@@ -1,49 +1,23 @@
...
-     href="https://material-components-web.appspot.com/list.html">
-    <img src="{{ site.rootpath }}/images/mdc_web_screenshots/lists.png" width="365" alt="Lists screenshot">
-  </a>
-</div>-->
-
-MDC List provides styles which implement [Material Design Lists](https://material.io/guidelines/components/lists.html) -
-"A single continuous column of tessellated subdivisions of equal width." Both single-line and two-line lists are
-supported (with three-line lists [planned](https://github.com/material-components/material-components-web/issues/31)).
-MDC Lists are designed to be accessible and RTL aware.
+MDC List provides styles which implement Material Design Lists - "A single continuous column of tessellated subdivisions of equal width." Both single-line and two-line lists are supported (with three-line lists planned). MDC Lists are designed to be accessible and RTL aware.
```
bonniezhou: Maybe for the first line:

MDC List provides styles which implement a single continuous column of tessellated subdivisions of equal width.

packages/mdc-list/README.md
```diff
@@ -1,49 +1,23 @@
...
 
-<ul class="icon-list">
-  <li class="icon-list-item icon-list-item--spec">
-    <a href="https://material.io/guidelines/components/lists.html">Material Design guidelines: Lists</a>
-  </li>
-  <li class="icon-list-item icon-list-item--link">
-    <a href="https://material-components-web.appspot.com/list.html">Demo</a>
-  </li>
-</ul>
+* [Material Design guidelines: List](https://material.io/guidelines/components/lists.html)
```
bonniezhou: This should also follow the [template](https://github.com/material-components/material-components-web/blob/master/docs/code/readme_template.md)

packages/mdc-list/README.md
```diff
@@ -1,49 +1,23 @@
...
 ```
 npm install --save @material/list
 ```
 
 ## Usage
 
-### Basic Lists
 
-A basic list consists simply of the list itself, and list items taking up one line.
+## Single-Line List
```
bonniezhou: I think for the Usage section, we should separate the sections first by lanaguage (HTML, then CSS/Sass) and then by type of list (single-line, two-line, etc). Then we can consolidate the CSS/Sass into a single reference table, instead having a CSS table for every type of list. E.g.
```
### HTML Structure
#### Single-line list
...
#### Two-line list
...
#### List groups
..
#### List dividers
...
### CSS Classes
...
### Sass Mixins
...
```
See the [Text Field README](https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/README.md) for an example of this.

The reason for this is because the first thing developers want to know is what types of lists we provide and the HTML structure that they can copy/paste to use it, which is why we should put all of that at the beginning. Then, if they're interested in understanding the structure or CSS behind it, they can look at the CSS/Sass reference tables. There's also a lot of overlap in the CSS tables (`mdc-list-item` appears in multiple tables, for example) so we can consolidate all the tables into one, to be as concise as possible.

But that's my thinking, what are your thoughts on this?

bonniezhou: Also maybe we should add a one-line description for all the different types of lists (single-line, two-line, etc).

packages/mdc-list/README.md
moog16: Ah yes, you're right. Thanks for catching.

packages/mdc-list/README.md
moog16: Yas, definitely. That makes sense to me. As someone new to MDC I would want all the classes in one place. I think the dev will be smart enough to figure out where the class should go.

And yes I'll add a one-liner for each type of list

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
-</ul>
-```
-
-### Two-Line List Items
-
-While in theory you can add any number of "lines" to a list item, you can use the `mdc-list--two-line` combined with
-some extra markup around the text to style a list in the two-line list style as defined by
-[the spec](https://material.io/guidelines/components/lists.html#lists-specs) (see "Two-line lists").
+#### Two-Line List
+While in theory you can add any number of "lines" to a list item, you can use the mdc-list--two-line combined with some extra markup around the text to style a list in the two-line list style as defined by [the spec](https://material.io/guidelines/components/lists.html#lists-specs) (see "Two-line lists").
```
bonniezhou: Style as inline code, i.e. `mdc-list--two-line`

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
-```
-
-### List Dividers
-
-MDC List contains an `mdc-list-divider` class which can be used as full-width or inset
-subdivisions either within lists themselves, or standalone between related groups of content.
-
-To use within lists, simply add the `mdc-list-divider` class to a list item.
+#### List Items
+List-items (rows) can contain primary and secondary actions. Lists-items can contain 1 supporting graphic tile and/or 1 metadata tile that are positioned at the start and end of the list item, respectively.
```
bonniezhou: nit: List items don't need to be hyphenated

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
-Inset dividers are useful when working with lists which have graphics.
-To add an inset divider, also add the `mdc-list-divider--inset` modifier class to the divider element.
-
-### List Groups
-
-Multiple related lists can be grouped together using the `mdc-list-group` class on a containing
-element. `mdc-list-divider` elements can be used in these groups _between_ lists to help
-differentiate them.
+#### List Groups
+Multiple related lists can be grouped together using the mdc-list-group class on a containing element.
```
bonniezhou: Style as inline code, i.e. `mdc-list-group`

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
-### Tips/Tricks
-
-#### Bordered Lists
-
-While hinted at within the spec, **bordered lists** - where each list item has a border around
-it - are not officially part of the list component spec. However, they seem to be used
-often in web applications, especially those suited more for desktop. The following example shows how
-to add borders to lists.
+#### List Dividers
+MDC List contains an mdc-list-divider class which can be used as full-width or inset subdivisions either within lists themselves, or standalone between related groups of content.
```
bonniezhou: Style as inline code, `mdc-list-divider`

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
+  <li class="mdc-list-item">Item 1 - Division 1</li>
+  <li class="mdc-list-item">Item 2 - Division 1</li>
+  <li role="separator" class="mdc-list-divider"></li>
+  <li class="mdc-list-item">Item 1 - Division 2</li>
+  <li class="mdc-list-item">Item 2 - Division 2</li>
 </ul>
 ```
 
-#### Control Tile Positions
+> Note the role="separator" attribute on the list divider. It is important to include this so that assistive technology can be made aware that this is a presentational element and is not meant to be included as an item in a list. Note that separator is indeed a valid role for li elements.
```
bonniezhou: nit: Style the note here similar to notes in the [Text Field readme](https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield) e.g.
> _NOTE_: add note here

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
+`mdc-list-item--activated` | Optional, styles the row in an activated* state
+`mdc-list-item__graphic` | Optional, the first tile in the row (in LTR languages, the first column of the list-item). Typically an icon or image.
+`mdc-list-item__meta`	| Optional, the last tile in the row (in LTR languages, the last column of the list-item). Typically small text, icon. or image.
+`mdc-list-group` | Optional, wrapper around two or more mdc-list elements to be grouped together
+`mdc-list-group__subheader` |	Optional, heading text displayed above each list in a group
+`mdc-list-divider` | Optional, for list divider element
+`mdc-list-divider--padded` | Optional, leaves gaps on each side of divider to match padding of list-item__meta
+`mdc-list-divider--inset` | Optional, increases the leading margin of the divider so that it does not intersect the avatar column
+
+\* **Note**: `mdc-list-divider` class can be used between list-items (example 1) *OR* between two lists (example 2)
```
bonniezhou: See comment above about styling notes

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
+
+\* **Note**: the difference between selected and activated states:
+
+* *Selected* state should be implemented on the list-item when it is likely to change soon. Eg., selecting one or more photos to share in Google Photos.
+* Multiple items can be selected at the same time when using the *selected* state
+* *Activated* state is similar to selected state, however should only be implemented once within a specific list.
+* *Activated* state is more permanent than selected state, and will **NOT** change soon relative to the lifetime of the page.
+
+### Sass Variables and Mixins
```
bonniezhou: Remove "Variables" since there aren't any here

packages/mdc-list/README.md
```diff
@@ -52,390 +52,132 @@ A basic list consists simply of the list itself, and list items taking up one li
...
-</script>
-```
-
-### List Dividers
-
-MDC List contains an `mdc-list-divider` class which can be used as full-width or inset
-subdivisions either within lists themselves, or standalone between related groups of content.
-
-To use within lists, simply add the `mdc-list-divider` class to a list item.
+#### List Items
```
bonniezhou: I'm inclined to remove this section and just put the paragraph describing list items under Single-line lists...but it's up to you

moog16: No I think you're right. the item is part of the list, and this section is quite bare

moog16: less is more

PR #2039, opened 2018-01-23
feat(card): Add primary action element with hover, focused, and pressed states

### LINKED ISSUE #1709: Add Hover, Focus, and Pressed states to MDC Card
MDC Cards should be clickable, and should use our states subsystem to show hover, focus, and pressed states

acdvorak:
Fixes #1709


packages/mdc-card/mdc-card.scss
```diff
@@ -81,6 +83,32 @@
...
+// Primary action
+//
+
+.mdc-card__primary-action {
+  @include mdc-ripple-surface;
+  @include mdc-ripple-radius-bounded;
+  @include mdc-states;
+
+  box-sizing: border-box;
+  outline: 0 none;
```
kfranqueiro: Isn't the `0` here redundant?

packages/mdc-card/README.md
```diff
@@ -172,6 +172,7 @@ CSS Class | Description
 --- | ---
 `mdc-card` | Mandatory, for the card element
 `mdc-card--stroked` | Removes the shadow and displays a hairline stroke instead
+`mdc-card__primary-action` | The main tappable area of the card. Typically contains most (or all) card content _except_ `mdc-card__actions`.
```
kfranqueiro: Might want to further qualify that this is only applicable to cards that have a primary action that the main surface should trigger?

acdvorak: Done

packages/mdc-card/mdc-card.scss
acdvorak: Done

packages/mdc-card/mdc-card.scss
```diff
@@ -48,7 +48,9 @@
 .mdc-card__media {
   @include mdc-card-media-aspect-ratio-base_;
 
+  display: flex;
```
kfranqueiro: I'm not sure this makes much sense here, but also it seems like it would make more sense (and also work) if we added display / flex-direction to `.mdc-card__primary-action` instead.

In fact the 3 styles (not counting mixins) currently applied to `mdc-card` should maybe just be put in a selector that applies to both `mdc-card` and `mdc-card__primary-action`?

PR #2051, opened 2018-01-23
style(text-field): Move -with-leading-icon to private mixins

lynnjepsen:


packages/mdc-textfield/_mixins.scss
```diff
@@ -406,6 +406,50 @@
...
+  .mdc-text-field__icon {
+    @include mdc-rtl-reflexive-position(left, $mdc-text-field-icon-position);
+  }
+
+  .mdc-text-field__label {
+    @include mdc-rtl-reflexive-position(left, $mdc-text-field-icon-padding);
+  }
+}
+
+@mixin mdc-text-field-outlined-with-leading-icon_ {
```
moog16: Do you want to group these with the outlined mixins?

packages/mdc-textfield/_mixins.scss
```diff
@@ -406,6 +406,50 @@
...
+      transform: translateX(27%) translateY(-130%) scale(.75);
+    }
+  }
+
+  .mdc-text-field__label--shake {
+    animation: invalid-shake-float-above-outlined-leading-icon 250ms 1;
+  }
+}
+
+@mixin mdc-text-field-outlined-dense-with-leading-icon_ {
```
moog16: same with this one

packages/mdc-textfield/_mixins.scss
lynnjepsen: I dont think so. I think I want to cover all cases of icons (leading or trailing) in the same code grouping. Lots of the styling seems to revolve more styling the icon.

PR #2058, opened 2018-01-23
fix(text-field): Remove unnecessary styling on label in disabled state

lynnjepsen:
The label already sets `bottom: 8px` in [mdc-text-field-label.scss](https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/label/mdc-text-field-label.scss#L26). So it seems unnecessary to override that in `mdc-text-field--disabled`


kfranqueiro:
Hm. Yeah I'm not sure why that style was there. Also I think removing this line might also allow you to remove lines for disabled box/outlined which are simply there to override this back to the value they would otherwise have inherited?

lynnjepsen:
Sure, removed extra styling from the disabled outlined and box...and I removed some extra styling from the dense box

PR #2060, opened 2018-01-23
fix(text-field): Remove dead code now that there is no CSS only version

lynnjepsen:


packages/mdc-textfield/label/_mixins.scss
```diff
@@ -25,8 +25,7 @@
 // Private
 
 @mixin mdc-text-field-label-color_($color) {
-  .mdc-text-field__label,
-  .mdc-text-field__input::placeholder { // placeholder used in place of label on css only version
```
williamernest: We use this for the fullwidth text-field.

PR #2061, opened 2018-01-24
chore(menu): Rename SimpleMenu to Menu

williamernest:
BREAKING CHANGE: Renames MDCSimpleMenu to MDCMenu. Renames all `mdc-simple-menu` classes to `mdc-menu`.


Kerrick:
This broke the util export from #824

williamernest:
@Kerrick Thanks for catching this. I'll be adding this back in the new `mdc-menu-surface` package coming in the next major release.

PR #2062, opened 2018-01-24
refactor(text-field): Move transform logic into mdc-text-field-label-float_

lynnjepsen:
I've moved positionX, positionY, and scale variables for the label into mdc-text-field/label/_variables.scss. And I created a new `mdc-text-field-label-float_` mixin in mdc-text-field/label/_mixins.scss. Then I call `mdc-text-field-label-float_` with the variables from mdc-text-field.scss.

This drops 30 lines of code from mdc-text-field/_mixins.scss


packages/mdc-textfield/label/_mixins.scss
```diff
@@ -46,14 +48,30 @@
...
+@include mdc-text-field-invalid-label-shake-keyframes_(box, $mdc-text-field-box-label-position-y);
+@include mdc-text-field-invalid-label-shake-keyframes_(box-dense, $mdc-text-field-box-dense-label-position-y, 0%, $mdc-text-field-dense-label-scale);
+@include mdc-text-field-invalid-label-shake_keyframes_(outlined, $mdc-text-field-outlined-label-position-y);
+@include mdc-text-field-invalid-label-shake_keyframes_(outlined-dense, $mdc-text-field-outlined-dense-label-position-y, 0%, $mdc-text-field-dense-label-scale);
+@include mdc-text-field-invalid-label-shake_keyframes_(outlined-leading-icon, $mdc-text-field-outlined-label-position-y, $mdc-text-field-outlined-with-leading-icon-label-position-x);
+@include mdc-text-field-invalid-label-shake_keyframes_(outlined-leading-icon-dense, $mdc-text-field-outlined-dense-label-position-y, $mdc-text-field-outlined-dense-with-leading-icon-label-position-x, $mdc-text-field-dense-label-scale);
+@include mdc-text-field-invalid-label-shake-keyframes_(textarea, $mdc-text-field-textarea-label-position-y, 0%, $mdc-text-field-textarea-label-scale);
+
+@mixin mdc-text-field-label-float_($positionY, $positionX: 0%, $scale: .75) {
```
moog16: Should the default of $positionX also be 0, since we are checking 0? I ran npm run build and tested it, and It looks like it works as intended. Really just a style decision.

packages/mdc-textfield/_mixins.scss
```diff
@@ -157,14 +157,12 @@
 }
 
 @mixin mdc-text-field-dense_ {
+  // NOTE: This is an eyeball'd approximation of what's in the mocks.
+  @include mdc-text-field-label-float_(110%, 0%, $mdc-text-field-dense-label-scale);
```
moog16: Did you not give 110% a var name since its only used here?

packages/mdc-textfield/label/_mixins.scss
```diff
@@ -46,14 +48,30 @@
...
+  @if $positionX == 0 {
+    .mdc-text-field__label--float-above {
+      transform: translateY(-1 * $positionY) scale($scale);
+    }
+  } @else {
+    .mdc-text-field__label--float-above {
+      transform: translateY(-1 * $positionY) translateX(-1 * $positionX) scale($scale);
+
+      @include mdc-rtl {
```
moog16: This probably fixed a bunch of bugs from the ones missing it. NICE 💯

packages/mdc-textfield/_mixins.scss
lynnjepsen: Ya...I think there's probably some bugs around the floating label with a dense text field...but I decided to keep this PR as a pure refactor

packages/mdc-textfield/label/_mixins.scss
lynnjepsen: I copied the arguments from `mdc-text-field-invalid-label-shake-keyframes_`...no idea if its the "right" decision, but at least we're consistent

PR #2066, opened 2018-01-24
chore(demos): Update list demo to show checkbox examples

moog16:
added to show examples for bugs we have recently seen.

screen shot 2018-01-23 at 15 00 18

screen shot 2018-01-23 at 15 00 05


demos/list.html
```diff
@@ -420,6 +482,68 @@ <h3>Metadata (Dense)</h3>
               </li>
             </ul>
           </section>
+          <section>
+            <h3>Metadata - Checkbox Example</h3>
```
acdvorak: Let's rename this heading to:

`Trailing Checkbox`

demos/list.html
```diff
@@ -307,13 +307,14 @@ <h3>Graphic Example - Icon with Text (Dark)</h3>
...
-            <ul class="mdc-list demo-list">
+            <h3>Leading Checkbox</h3>
+            <ul class="mdc-list demo-list" id="leading-checkbox-list">
               <li class="mdc-list-item">
                 <span class="mdc-list-item__graphic">
                   <div class="mdc-checkbox">
                     <input type="checkbox"
-                           class="mdc-checkbox__native-control"/>
+                           class="mdc-checkbox__native-control"
+                           label="leading-checkbox-blueberries"/>
```
acdvorak: `s/label/id/`

demos/list.html
```diff
@@ -483,14 +486,15 @@ <h3>Metadata (Dense)</h3>
             </ul>
           </section>
           <section>
-            <h3>Metadata - Checkbox Example</h3>
-            <ul class="mdc-list demo-list">
+            <h3>Trailing Checkbox</h3>
+            <ul class="mdc-list demo-list" id="trailing-checkbox-list">
               <li class="mdc-list-item">
-                Blueberries
+                <label for="leading-checkbox-blueberries">Blueberries</label>
```
acdvorak: `s/leading/trailing/`

demos/list.html
```diff
@@ -483,14 +486,15 @@ <h3>Metadata (Dense)</h3>
...
                 <span class="mdc-list-item__meta">
                   <div class="mdc-checkbox">
                     <input type="checkbox"
-                           class="mdc-checkbox__native-control"/>
+                           class="mdc-checkbox__native-control"
+                           label="leading-checkbox-blueberries"/>
```
acdvorak: `s/label/id/`
`s/leading/trailing/`

demos/list.html
```diff
@@ -505,11 +509,12 @@ <h3>Metadata - Checkbox Example</h3>
                 </span>
               </li>
               <li class="mdc-list-item">
-                Boysenberries
+                <label for="leading-checkbox-boysenberries">Boysenberries</label>
```
acdvorak: `s/label/id/`
`s/leading/trailing/`

demos/list.html
```diff
@@ -505,11 +509,12 @@ <h3>Metadata - Checkbox Example</h3>
...
                 <span class="mdc-list-item__meta">
                   <div class="mdc-checkbox">
                     <input type="checkbox"
-                           class="mdc-checkbox__native-control"/>
+                           class="mdc-checkbox__native-control"
+                           label="leading-checkbox-boysenberries"/>
```
acdvorak: `s/label/id/`
`s/leading/trailing/`

demos/list.html
```diff
@@ -524,11 +529,12 @@ <h3>Metadata - Checkbox Example</h3>
                 </span>
               </li>
               <li class="mdc-list-item">
-                Strawberries
+                <label for="leading-checkbox-strawberries">Strawberries</label>
```
acdvorak: `s/leading/trailing/`

demos/list.html
```diff
@@ -524,11 +529,12 @@ <h3>Metadata - Checkbox Example</h3>
...
                 <span class="mdc-list-item__meta">
                   <div class="mdc-checkbox">
                     <input type="checkbox"
-                           class="mdc-checkbox__native-control"/>
+                           class="mdc-checkbox__native-control"
+                           label="leading-checkbox-strawberries"/>
```
acdvorak: `s/label/id/`
`s/leading/trailing/`

demos/list.html
```diff
@@ -1324,6 +1330,24 @@ <h3>Example - Interactive List (Dark Theme)</h3>
             evt.preventDefault();
           });
         }
+
+        var clickHandler = function(e) {
+          var checkbox = this.querySelector('input[type="checkbox"]');
+          if(e.target !== checkbox) {
+            checkbox.checked = !checkbox.checked;
+          }
+        }
```
acdvorak: Add a `;` or use function declaration syntax instead
(e.g., `function clickHandler(e) { ... }`)

demos/list.html
```diff
@@ -1324,6 +1330,24 @@ <h3>Example - Interactive List (Dark Theme)</h3>
             evt.preventDefault();
           });
         }
+
+        var clickHandler = function(e) {
+          var checkbox = this.querySelector('input[type="checkbox"]');
```
acdvorak: `var checkbox = this.querySelector('.mdc-checkbox__native-control');`

demos/list.html
```diff
@@ -1324,6 +1330,24 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+          if(e.target !== checkbox) {
+            checkbox.checked = !checkbox.checked;
+          }
+        }
+
+        var integrateCheckboxListItems = function(list) {
+          list.querySelectorAll('.mdc-list-item').forEach(function(listItem) {
+            listItem.addEventListener('click', clickHandler.bind(listItem));
+          });
+        }
```
acdvorak: Add a `;` or use function declaration syntax instead
(e.g., `function integrateCheckboxListItems(e) { ... }`)

moog16:
Spoke with Michelle about toggling the checkbox when the list item is clicked: she said that this should not happen.

Spoke with jnaimark: He confirmed that the checkbox ripple should only be triggered on the checkbox click, and the list-item ripple should only be triggered on list-item click. Although this is not happening, I spoke with @kfranqueiro who said type of list should be non-interactive.

moog16:
@acdvorak spoke to michelle and convinced her otherwise! So if list-item is clicked, this should also toggle checkbox.

demos/list.html
```diff
@@ -1200,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+        var deactivateRippleHandler = null;
+        var checkboxListRippleSurfaceClass = 'checkbox-list-ripple-surface';
+
+        function clickHandler(evt, rippleFoundation, li) {
+          var checkbox = li.querySelector('.mdc-checkbox__native-control');
+          var isLabel = Array.prototype.some.call(checkbox.labels, function(label) { return label === evt.target; });
+          if(evt.target !== checkbox && !isLabel) {
+            checkbox.checked = !checkbox.checked;
+            li.classList.add(checkboxListRippleSurfaceClass);
+            rippleFoundation.activate();
```
moog16: If I pass the event here, I don't need the deactivateRippleHandler logic.

kfranqueiro: That might be a better idea cross-device, since that will allow the ripple foundation to register all types of pointer events for deactivation rather than just mouseup like you're doing here.

Alternatively, what form-field does is simply call deactivate on the next animation frame.

demos/list.html
```diff
@@ -1329,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+            });
+
+            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(rippleAdapter);
+            rippleFoundation.layout();
+
+            li.addEventListener('mousedown', e => clickHandler(e, rippleFoundation, li));
+          });
+        };
+
+        var leadingCheckboxList = document.getElementById('leading-checkbox-list');
```
acdvorak: You can probably inline these two variables

demos/list.html
```diff
@@ -1329,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+            var rippleAdapter = mdc.ripple.MDCRipple.createAdapter({
+              unbounded: false,
+              disabled: false,
+              root_: li
+            });
+
+            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(rippleAdapter);
+            rippleFoundation.layout();
+
+            li.addEventListener('mousedown', e => clickHandler(e, rippleFoundation, li));
```
acdvorak: IE 11 💩  doesn't support arrow functions ☹️   (It also lacks proper `const` or `let` support).

So our uncompiled demo JS needs to use ES5 syntax, sadly.

demos/list.html
```diff
@@ -1329,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+        };
+
+        function deactivateRipple(evt, rippleFoundation, li) {
+          li.classList.remove(checkboxListRippleSurfaceClass);
+          rippleFoundation.deactivate();
+          li.removeEventListener('mouseup', deactivateRippleHandler);
+        };
+
+        function addRippleToListItems(list) {
+          list.querySelectorAll('.mdc-list-item').forEach(function(li) {
```
acdvorak: Not all browsers support `NodeList.forEach()` (I'm looking at you, IE 11 💩 ).

So you have to do `[].forEach.call(function(li) { ... })` instead.

demos/list.html
```diff
@@ -1329,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+
+        function clickHandler(evt, rippleFoundation, li) {
+          var checkbox = li.querySelector('.mdc-checkbox__native-control');
+          var isLabel = Array.prototype.some.call(checkbox.labels, function(label) { return label === evt.target; });
+          if(evt.target !== checkbox && !isLabel) {
+            checkbox.checked = !checkbox.checked;
+            li.classList.add(checkboxListRippleSurfaceClass);
+            rippleFoundation.activate();
+
+            deactivateRippleHandler = e => deactivateRipple(e, rippleFoundation, li);
```
acdvorak: IE 11 💩  doesn't support arrow functions ☹️   (It also lacks proper `const` or `let` support).

So our uncompiled demo JS needs to use ES5 syntax, sadly.

demos/list.html
```diff
@@ -1329,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
             evt.preventDefault();
           });
         }
+
+        var deactivateRippleHandler = null;
+        var checkboxListRippleSurfaceClass = 'checkbox-list-ripple-surface';
+
+        function clickHandler(evt, rippleFoundation, li) {
+          var checkbox = li.querySelector('.mdc-checkbox__native-control');
+          var isLabel = Array.prototype.some.call(checkbox.labels, function(label) { return label === evt.target; });
```
acdvorak: Nit: Although `Array.prototype.some.call(...)` is technically The Right Way to do it, I tend to prefer:

`[].some.call(...)`

It's terser and (IMO) easier to read.

And unless you're writing performance-critical code, the cost of allocating a new object is negligible.

acdvorak: `checkbox.labels` is not supported in IE 11 💩  or Edge 🙁

demos/list.html
acdvorak: Yeah, let's pass the event to `activate()` so we don't need the `mouseup` handler.

demos/list.html
```diff
@@ -1329,6 +1329,48 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+
+        function deactivateRipple(evt, rippleFoundation, li) {
+          li.classList.remove(checkboxListRippleSurfaceClass);
+          rippleFoundation.deactivate();
+          li.removeEventListener('mouseup', deactivateRippleHandler);
+        };
+
+        function addRippleToListItems(list) {
+          list.querySelectorAll('.mdc-list-item').forEach(function(li) {
+            var rippleAdapter = mdc.ripple.MDCRipple.createAdapter({
```
acdvorak: We could probably inline this, but it's up to you 🙂

demos/list.html
moog16: definitely easier to read...changing

moog16: :( I'll use querySelector instead

demos/list.html
moog16: ah...right. I'm just so used to it I didn't even realize.

demos/list.html
moog16: reference for myself: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels

demos/list.html
moog16: You mean like

```
var leadingCheckboxList = document.getElementById('leading-checkbox-list'),
        trailingCheckboxList = document.getElementById('trailing-checkbox-list');
```

kfranqueiro: I assume he meant don't use variables, since they're used exactly once each. We never do multiple declarations in one var statement.

moog16: gotcha

demos/list.html
```diff
@@ -1330,47 +1330,34 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
-            });
-
-            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(rippleAdapter);
-            rippleFoundation.layout();
+          var listItems = list.querySelectorAll('.mdc-list-item');
+          [].forEach.call(listItems, function(li) {
+            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(
+              mdc.ripple.MDCRipple.createAdapter({
+                unbounded: false, disabled: false, root_: li
+              })
```
acdvorak: Add a `;`

demos/list.html
```diff
@@ -1330,47 +1330,34 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+            );
 
-            li.addEventListener('mousedown', e => clickHandler(e, rippleFoundation, li));
+            li.addEventListener('mousedown', function(e) {
+              clickHandler(e, rippleFoundation, li)
```
acdvorak: Add a `;`

demos/list.scss
```diff
@@ -75,6 +76,22 @@ a.material-icons {
   margin-left: 72px;
 }
 
+.checkbox-list-ripple-surface {
+  @include mdc-ripple-surface;
+  @include mdc-ripple-radius;
+  @include mdc-states;
+  @include mdc-states-base-color(black);
+
+  position: relative;
```
acdvorak: I think this property and `overflow: hidden` are already being set by `mdc-list-item`.

Is there a specific reason you needed to add them?

demos/list.scss
```diff
@@ -75,6 +76,22 @@ a.material-icons {
...
+  outline: none;
```
acdvorak: Is `outline: none` still necessary? The list item is no longer focusable.

demos/list.html
```diff
@@ -1200,6 +1329,35 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+          var listItems = list.querySelectorAll('.mdc-list-item');
+          [].forEach.call(listItems, function(li) {
+            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(
+              mdc.ripple.MDCRipple.createAdapter({
+                unbounded: false, disabled: false, root_: li
+              })
+            );
+
+            li.addEventListener('mousedown', function(e) {
+              clickHandler(e, rippleFoundation, li)
```
acdvorak: Add a `;`

demos/list.scss
```diff
@@ -75,6 +76,22 @@ a.material-icons {
...
+  overflow: hidden;
+}
+
+.checkbox-list-ripple-surface:not(.mdc-ripple-upgraded):active::after {
+  // disable checkbox click activating ripple effect
```
acdvorak: Nit: Reword to:

```
Prevent checkbox clicks from activating a second ripple on the list item.
```

demos/list.html
```diff
@@ -306,6 +306,71 @@ <h3>Graphic Example - Icon with Text (Dark)</h3>
...
+              <li class="mdc-list-item checkbox-list-ripple-surface">
+                <span class="mdc-list-item__graphic">
+                  <div class="mdc-checkbox">
+                    <input type="checkbox"
+                           class="mdc-checkbox__native-control"
+                           id="leading-checkbox-blueberries"/>
+                    <div class="mdc-checkbox__background">
+                      <svg class="mdc-checkbox__checkmark"
+                           viewBox="0 0 24 24">
+                        <path class="mdc-checkbox__checkmark__path"
```
acdvorak: Replace `__path` with `-path` in this entire file (it was renamed in #2096).

Otherwise, the checkboxes look like this:

pr 2066 capture 2

demos/list.html
moog16: This is inside the new mdc.ripple.MDCRippleFoundation function.

demos/list.scss
moog16: I didn't try taking them off. I copied from mdc-ripple-surface. Let me try removing

demos/list.scss
moog16: nope redundant.

demos/list.html
acdvorak: Oh woops! You're absolutely right 😀

demos/list.scss
moog16: I realize I dont even need these mixins either...removing

demos/list.html
```diff
@@ -1200,6 +1329,35 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+          }
+          if(evt.target !== checkbox) {
+            rippleFoundation.activate(evt);
+          }
+        };
+
+        function addRippleToListItems(list) {
+          var listItems = list.querySelectorAll('.mdc-list-item');
+          [].forEach.call(listItems, function(li) {
+            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(
```
kfranqueiro: Why do you need all of this code?  Wouldn't the following be enough?

```js
          [].forEach.call(listItems, function(li) {
            mdc.ripple.MDCRipple.attachTo(li);
          });
```

demos/list.html
```diff
@@ -1200,6 +1329,35 @@ <h3>Example - Interactive List (Dark Theme)</h3>
             evt.preventDefault();
           });
         }
+
+        function clickHandler(evt, rippleFoundation, li) {
```
kfranqueiro: Why do you need all of this code? Since you're instantiating a ripple on each li anyway, you shouldn't need to activate it manually.

Shouldn't the following be enough to work?

```js
          var checkbox = li.querySelector('.mdc-checkbox__native-control');
          if(evt.target === li) {
            checkbox.checked = !checkbox.checked;
          }
```

And then you don't even need `rippleFoundation`.

demos/list.html
```diff
@@ -1200,6 +1329,35 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+        function addRippleToListItems(list) {
+          var listItems = list.querySelectorAll('.mdc-list-item');
+          [].forEach.call(listItems, function(li) {
+            var rippleFoundation = new mdc.ripple.MDCRippleFoundation(
+              mdc.ripple.MDCRipple.createAdapter({
+                unbounded: false, disabled: false, root_: li
+              })
+            );
+
+            li.addEventListener('mousedown', function(e) {
```
kfranqueiro: I might be inclined to opt for simply listening on `click` here.

demos/list.html
```diff
@@ -1200,6 +1329,26 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
         }
+
+        function clickHandler(evt, li) {
+          var checkbox = li.querySelector('.mdc-checkbox__native-control');
+          if(evt.target === li) {
+            checkbox.checked = !checkbox.checked;
+          }
+        };
+
+        function addRippleToListItems(list) {
```
kfranqueiro: You should also forEach over `list.querySelectorAll('.mdc-checkbox')` to instantiate checkboxes (which will instantiate ripples on them) to benefit from #2123.

demos/list.html
```diff
@@ -1330,30 +1330,27 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
-              clickHandler(e, rippleFoundation, li);
+          [].forEach.call(listItems, function(li) {
+            mdc.ripple.MDCRipple.attachTo(li);
+            li.addEventListener('click', function(e) {
+              clickHandler(e, li);
             });
           });
+
+          [].forEach.call(checkboxes, function(checkbox) {
+            mdc.ripple.MDCRipple.attachTo(checkbox);
```
acdvorak: `mdc.checkbox.MDCCheckbox.attachTo`?

acdvorak: Also, this loop ends up creating a second ripple on `#toggle-rtl-checkbox` (the first is on [lines 1307-1309](https://github.com/material-components/material-components-web/blob/cf2b5a537510f362c9e9eb023e1bd404850a21a6/demos/list.html#L1307-L1309)).

I would remove lines 1307-1309, and replace this loop with:

```js
[].forEach.call(checkboxes, function(checkbox) {
  var cbInstance = mdc.checkbox.MDCCheckbox.attachTo(checkbox);
  if (checkbox.parentElement.classList.contains('mdc-form-field')) {
    var ffInstance = mdc.formField.MDCFormField.attachTo(checkbox.parentElement);
    ffInstance.input = cbInstance;
  }
});
```

demos/list.html
```diff
@@ -1200,6 +1329,32 @@ <h3>Example - Interactive List (Dark Theme)</h3>
...
+
+          [].forEach.call(listItems, function(li) {
+            mdc.ripple.MDCRipple.attachTo(li);
+            li.addEventListener('click', function(e) {
+              clickHandler(e, li);
+            });
+          });
+
+          [].forEach.call(checkboxes, function(checkbox) {
+            mdc.ripple.MDCRipple.attachTo(checkbox);
```
acdvorak: `mdc.checkbox.MDCCheckbox.attachTo`?

Also, this loop ends up creating a second ripple on `#toggle-rtl-checkbox` (the first is on [lines 1307-1309](https://github.com/material-components/material-components-web/blob/cf2b5a537510f362c9e9eb023e1bd404850a21a6/demos/list.html#L1307-L1309)).

I would remove lines 1307-1309, and replace this loop with:

```js
[].forEach.call(checkboxes, function(checkbox) {
  var cbInstance = mdc.checkbox.MDCCheckbox.attachTo(checkbox);
  if (checkbox.parentElement.classList.contains('mdc-form-field')) {
    var ffInstance = mdc.formField.MDCFormField.attachTo(checkbox.parentElement);
    ffInstance.input = cbInstance;
  }
});
```

moog16: Will also remove the id `toggle-rtl-checkbox` from the checkbox

demos/list.html
moog16: This is happening on line #1309. I wanted to group all the checkbox logic together.

demos/list.html
```diff
@@ -1200,6 +1335,27 @@ <h3>Example - Interactive List (Dark Theme)</h3>
             evt.preventDefault();
           });
         }
+
+        function clickHandler(evt, li) {
+          var checkbox = li.querySelector('.mdc-checkbox__native-control');
+          if(evt.target === li) {
```
kfranqueiro: Missing space between `if` and `(`

Also technically you only need to query for the checkbox inside the `if`

PR #2073, opened 2018-01-24
fix(text-field): Make outline visibility directly linked to floating labels

lynnjepsen:


packages/mdc-textfield/outline/mdc-text-field-outline.scss
```diff
@@ -63,3 +63,15 @@
   mdc-text-field-transition(opacity);
   fill: transparent;
 }
+
+// Never show the idle outline when the label is floating, otherwise the label
```
kfranqueiro: Is it correct to be putting this within the outline subcomponent CSS when it references classes outside that component? Otherwise this idea LGTM.

lynnjepsen: eh youre probably right. I moved it to mdc-text-field.scss

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -255,3 +255,15 @@
 .mdc-text-field--textarea.mdc-text-field--disabled {
   @include mdc-text-field-textarea-disabled_;
 }
+
+// Never show the idle outline when the label is floating, otherwise the label
```
kfranqueiro: One more thought, though I'm being annoying now because this contradicts my earlier one because I didn't think of this...

Do you think it would make more sense to put these styles within an `@at-root` within the `mdc-text-field-outlined_` mixin instead? It's kind of off in no-man's land here which makes it harder to spot when working in the context of outlined styles, otherwise.

lynnjepsen: These styles are no longer specific to `mdc-text-field--outlined`. They are general styles for any text field that uses both a floating label and an outline.

Right now the only text field that uses a floating label and an outline happens to be `mdc-text-field--outlined`. But thats an internal detail.

kfranqueiro: I had to re-read this and it took me a minute to realize what you were getting at here but I think I see what you're getting at :P

PR #2076, opened 2018-01-24
chore(checkbox): remove dark theme

moog16:


demos/snackbar.html
```diff
@@ -33,6 +33,14 @@
         background-color: #333;
       }
 
+      .mdc-theme--dark .mdc-checkbox .mdc-checkbox__native-control:enabled:not(:checked):not(:indeterminate) ~ .mdc-checkbox__background {
```
moog16: Should I leave as is? Or use sass mixins in the demo.scss file?

lynnjepsen: Remove this. There should be no reference to `mdc-theme--dark` and `mdc-checkbox`

lynnjepsen: In demos/snackbar.scss could you add a call to the checkbox mixins?

kfranqueiro: Given that this is in the snackbar demo html and not the checkbox styles themselves, would it make more sense to leave this for when we remove dark mode from snackbar instead? No sense creating extra work that we're going to remove...

moog16: Yup, spoke to lynn but didn't comment here. I am going to merge in snackbar first, then remove these before merging.

acdvorak: PR #2100 removes dark theme from snackbar, so does it really matter if we break this for like 1 day?

I vote we revert these two rules and leave it broken.

kfranqueiro: I think that's what Matt is saying he wants to do, though he was thinking we could just merge the snackbar PR first and not break anything.

acdvorak: Gotcha. SGTM!

acdvorak: Now that #2100 is merged, we can remove these two `mdc-theme--dark` CSS rules! 😀

PR #2081, opened 2018-01-25
chore(icon-toggle): remove dark theme

moog16:


acdvorak:
# Add BREAKING CHANGE to commit message before merging

demos/icon-toggle.html
```diff
@@ -166,20 +137,6 @@ <h2>Additional Color Combinations</h2>
               </div>
               <div>Dark icon on background</div>
             </div>
-            <div id="custom-on-dark" class="demo-color-combo mdc-theme--dark">
```
acdvorak: I think we actually want to keep this example, because it shows how to use custom colors. Just remove the `mdc-theme--dark` class.

demos/icon-toggle.html
```diff
@@ -166,20 +137,6 @@ <h2>Additional Color Combinations</h2>
...
-                   role="button"
-                   aria-label="Add to favorites"
-                   aria-pressed="false"
-                   tabindex="0"
-                   data-toggle-on='{"content": "favorite", "label": "Remove From Favorites"}'
-                   data-toggle-off='{"content": "favorite_border", "label": "Add to Favorites"}'>
-                  favorite_border
-                </i>
-              </div>
-              <div class="mdc-theme--text-primary-on-dark">Custom color on dark background</div>
```
acdvorak:
```html
<div>Custom color on background</div>
```

packages/mdc-icon-toggle/mdc-icon-toggle.scss
```diff
@@ -19,7 +19,6 @@
 @import "@material/animation/functions";
 @import "@material/ripple/common";
 @import "@material/ripple/mixins";
-@import "@material/theme/mixins";
```
kfranqueiro: Don't remove this. `mdc-theme-prop` is still used in this file (test with `npm run build` for now)

demos/icon-toggle.html
moog16: roger!

demos/icon-toggle.scss
```diff
@@ -75,6 +75,10 @@
   @include mdc-states(black);
 }
 
+#custom-on-dark {
+  background-color: #6200ee;
```
acdvorak: Can we just remove this rule altogether?

The purpose of this example is to show how to customize the color of the icon; a background color isn't really necessary for that.

FYI, a red icon against a dark purple background has a [contrast ratio of `1.8`](http://leaverou.github.io/contrast-ratio/#%23de442c-on-%236200ee), which is really hard to see.
[The spec](https://material.io/guidelines/usability/accessibility.html#accessibility-color-contrast) says small text and icons should have a contrast ratio of at least `4.5:1` against their background.

Also, in Sass files, we should use the official Material color palette variables instead of hard-coding the RGB. E.g.:

```sass
#custom-on-dark {
  background-color: $material-color-grey-100;
}
```

demos/icon-toggle.html
```diff
@@ -178,7 +149,7 @@ <h2>Additional Color Combinations</h2>
                   favorite_border
                 </i>
               </div>
-              <div class="mdc-theme--text-primary-on-dark">Custom color on dark background</div>
+              <div class="mdc-theme--text-primary-on-primary">Custom color on background</div>
```
acdvorak: I would suggest removing the background color (see my next comment), which means we can remove all CSS classes from this element. E.g.:

```html
<div>Custom color</div>
```

PR #2082, opened 2018-01-25
chore(list): remove dark theme

moog16:


packages/mdc-list/_variables.scss
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 $mdc-list-divider-color-light: rgba(0, 0, 0, .12) !default;
-$mdc-list-divider-color-dark: rgba(255, 255, 255, .2) !default;
```
kfranqueiro: I'm wondering one of two things here...

* Should we keep this variable and use `mdc-theme-contrast-tone` in our styles to determine whether to use light or dark based on the value of the background color?
* Should we remove this and also remove `-light` from the other variable?

cc @acdvorak

packages/mdc-list/mdc-list.scss
```diff
@@ -21,26 +21,12 @@
 @import "./mixins";
 @import "./variables";
 
-.mdc-list,
```
kfranqueiro: I'm pretty sure you can remove the import for `@material/theme/mixins` after removing the dark styles in this file? (We also missed this for mdc-button.scss I think, so might want to send a tiny separate PR for that, and check on the other open PRs)

moog16: Will do...good call.

packages/mdc-list/_variables.scss
moog16: So we would just keep this style in here that other people can reference if they create their own "dark-theme"?

acdvorak: I vote for option 1: keep this variable and use `mdc-theme-contrast-tone`.

E.g.:

```sass
@at-root {
  $divider-color: if(
    mdc-theme-contrast-tone($mdc-theme-background) == "dark",
    $mdc-list-divider-color-dark,
    $mdc-list-divider-color-light
  );

  @include mdc-list-divider-color($divider-color);
}
```

That way, if the client overrides `$mdc-theme-background`, we will automatically set the correct divider color for them. If they don't like the color we choose, they can always override it with the mixin.

moog16: Ok I see what you guys are saying now. I will update as stated here. Thanks!

packages/mdc-list/_variables.scss
```diff
@@ -14,6 +14,5 @@
 
 $mdc-list-divider-color-light: rgba(0, 0, 0, .12) !default;
```
acdvorak: The names of these variables are confusing - can we rename them while we're here? E.g.:

```sass
$mdc-list-divider-color-on-light-bg: rgba(0, 0, 0, .12) !default;
$mdc-list-divider-color-on-dark-bg: rgba(255, 255, 255, .2) !default;
```

We'll need to add `BREAKING CHANGE` to the PR description (and the commit message when we squash and merge).

packages/mdc-list/mdc-list.scss
```diff
@@ -191,7 +170,13 @@ a.mdc-list-item {
 }
 
 @at-root {
-  @include mdc-list-divider-color($mdc-list-divider-color-light);
+  $divider-color: if(
+    mdc-theme-contrast-tone($mdc-theme-background) == "dark",
```
acdvorak: My mistake - this should actually be `mdc-theme-tone($mdc-theme-background) == "dark"`

The `$mdc-list-divider-color-[light|dark]` variables are named confusingly, so I got them mixed up.

If you look at the "Full-Width Dividers" and "Inset Dividers" examples, the dividers should be visible; currently they're not (because the colors were flipped).

moog16: yup just noticed...looks better

PR #2083, opened 2018-01-25
feat(chips): Baseline chip and chip set

### LINKED ISSUE #2003: Baseline chip and chip set
- [ ] `mdc-chip`
- [ ] use `mdc-states` within `mdc-chip`
- [ ] `mdc-chip__text`
- [ ] `mdc-chip-set`
- [ ] create foundation/adapter for basic chip functionality

bonniezhou:
MDC Chips is a new package in MDC Web for the [chips](https://material.io/guidelines/components/chips.html) component. It is comprised of two parts: the chip itself and the set that the chip belongs to. 

This PR includes:
* basic styles (including ripple) for `mdc-chip`, `mdc-chip-set`, and `mdc-chip__text` classes
* adapter, foundation, and vanilla component for `MDCChip` and `MDCChipSet`
* tests for `MDCChip` and `MDCChipSet`
* documentation and a new catalog page for Chips

This PR doesn't include:
* chip set behaviour
* handling icons
* Sass customization

Note: MDC Chips has a similar model to MDC Tabs, which is comprised of multiple mandatory components, as opposed to MDC Text Field, which is a single main component with multiple optional components. Thus there is a single README for MDC Chips.

Resolves #2003


packages/mdc-chips/README.md
```diff
@@ -0,0 +1,98 @@
...
+`chips` | Array<`MDCChip`> | An array of the `MDCChip` objects that represent chips in the set
+
+### Adapters: `MDCChipAdapter` and `MDCChipSetAdapter`
+
+#### `MDCChipAdapter`
+Method Signature | Description
+--- | ---
+`registerInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event listener on the root element
+`deregisterInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener on the root element
+`notifyInteraction() => void` | Emits a custom event "MDCChip:interaction" denoting the chip has been interacted with
```
kfranqueiro: Mention that this event bubbles, like we did with [text field icon](https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield/icon)

packages/mdc-chips/chip/adapter.js
```diff
@@ -0,0 +1,58 @@
...
+   * Emits a custom "MDCChip:interaction" event denoting the chip has been
+   * interacted with (typically on click or keydown).
+   */
+  notifyInteraction() {}
+
+  /**
+   * Returns the text content in the chip.
+   * @return {string}
+   */
+  getText() {}
```
kfranqueiro: If this is intended to eventually support e.g. selection APIs, I'd be a bit concerned of the consequences of i18n (i.e. the returned text would vary by language)

packages/mdc-chips/chip/index.js
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * @param {!Element} root
+   * @return {!MDCChip}
+   */
+  static attachTo(root) {
+    return new MDCChip(root);
+  }
+
+  initialize() {
+    this.ripple_ = new MDCRipple(this.root_);
```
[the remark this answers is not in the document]
bonniezhou: We might... I can't think of any use cases right now but we can see if it's necessary as we build out more features on chips.

packages/mdc-chips/chip/adapter.js
bonniezhou: Yeah this is a good point. I'll remove it for now and wait until we have the actual selection API built determine the best way for querying chips.

packages/mdc-chips/README.md
```diff
@@ -0,0 +1,94 @@
...
+`chips` | Array<`MDCChip`> | An array of the `MDCChip` objects that represent chips in the set
+
+### Adapters: `MDCChipAdapter` and `MDCChipSetAdapter`
+
+#### `MDCChipAdapter`
+Method Signature | Description
+--- | ---
+`registerInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event listener on the root element
+`deregisterInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener on the root element
+`notifyInteraction() => void` | Emits a custom event "MDCChip:interaction" denoting the chip has been interacted with, which bubbles to the top-level text field element
```
kfranqueiro: "text field element"? 🤔 replace with chip set?

packages/mdc-chips/chip-set/index.js
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * @return {!MDCChipSetFoundation}
+   */
+  getDefaultFoundation() {
+    return new MDCChipSetFoundation(/** @type {!MDCChipSetAdapter} */ (Object.assign({
+      hasClass: (className) => this.root_.classList.contains(className),
+    })));
+  }
+
+  instantiateChips_(chipFactory) {
```
kfranqueiro: Does this need type information to compile with closure?

Also, this PR doesn't seem to be adding `mdc-chips` to `closureWhitelist` in package.json... should we be doing that from the outset with this package?

packages/mdc-chips/chip/constants.js
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @enum {string} */
+const strings = {
+  INTERACTION_EVENT: 'MDCChip:interaction',
+  TEXT_SELECTOR: '.mdc-chip__text',
```
kfranqueiro: This is unused now; should it be removed here?

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -0,0 +1,38 @@
...
+  color: rgba(0, 0, 0, 0.87);
+  padding: 7px 12px;
+  margin: 5px;
+  cursor: pointer;
+  overflow: hidden;
+}
+
+.mdc-chip__text {
+  display: inline-block;
+  vertical-align: middle;
```
kfranqueiro: FYI I noticed something funny with this style... with it, chips are 33px tall, but without it, they're 32px tall. (Only checked in Chrome so far.)

Not sure if we're aiming for a specific height for chips, or if this is moot pending the addition of icons anyway.

test/unit/mdc-chips/mdc-chip.foundation.test.js
```diff
@@ -0,0 +1,69 @@
...
+  td.verify(mockAdapter.registerInteractionHandler('click', td.matchers.isA(Function)));
+  td.verify(mockAdapter.registerInteractionHandler('keydown', td.matchers.isA(Function)));
+});
+
+test('#destroy removes event listeners', () => {
+  const {foundation, mockAdapter} = setupTest();
+  foundation.destroy();
+
+  td.verify(mockAdapter.deregisterInteractionHandler('click', td.matchers.isA(Function)));
+  td.verify(mockAdapter.deregisterInteractionHandler('click', td.matchers.isA(Function)));
```
kfranqueiro: Is this supposed to be `keydown`?

packages/mdc-chips/chip/mdc-chip.scss
bonniezhou: Interesting... the `mdc-chip__text` div is 18px but for some reason the `mdc-chip` content becomes 19px. 

From the `experimental/chip` branch it seems that both `mdc-chip` content and `mdc-chip__icon` height are 24px once icons are added (which is the expected result).

Not sure what's causing this, but a fix for now would be to make the `mdc-chip__text` content 17px since that won't affect anything that comes afterwards. Any thoughts?

kfranqueiro: My first question would be... are chips w/o icons expected to be less tall than chips w/ icons? This would be weird if there's any situation in which you can have a mix of some with icons and some without within the same set, but OTOH if they're all expected to eventually match the with-icons size then this point may become moot.

bonniezhou: Chips w/o icons are expected to be the same height as chips with icons.
But chips with icons are also 32px, which means that we need to fix this for chips w/o icons anyways.

bonniezhou:
Added a few extra chip set tests to check that the child chips are being properly instantiated/destroyed.

I'm getting weird errors when trying to add the module to the Closure whitelist, so I'm still working on that.

bonniezhou:
Added `mdc-chips` to Closure whitelist now!

packages/mdc-chips/chip-set/adapter.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+class MDCChipSetAdapter {
+  /**
+   * Returns true if the root element contains the given class name.
+   * @param {string} className
+   */
+  hasClass(className) {}
+}
+
+export {MDCChipSetAdapter};
```
kfranqueiro: I noticed you had to update your imports due to this exporting an object containing the adapter rather than just exporting the adapter directly. I'm pretty sure we generally don't export an object unless we need to expose other things in addition (e.g. with text field), so I think you can remove the curly braces in the chip and chip-set adapters as well as their respective imports.

bonniezhou: Done!

Issue #2031, opened 2018-01-22
MDCMenuFoundation add --selected to list items, so that states code is applied to menu


lynnjepsen:
Currently our mdc-menu does not show the selected state of a list-item.

This may need to be implemented in a variant of mdc-menu. Some menus, once an item is selected, the menu is never expected to be re-opened. But for mdc-select, the menu we attach to the select component can be re-opened. Right now we use `[aria-selected="true"]` instead of `mdc-list-item--selected` in our CSS selectors within mdc-select.

Issue #536, opened 2017-04-19
mdc-menu closes immediately after Enter is pressed to open it


traviskaufman:
### What MDC-Web Version are you using?

Latest (0.9.1)

### What browser(s) is this bug affecting?

Chrome Version 57.0.2987.133 (64-bit)

```
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36"
```

### What OS are you using?

macOS Sierra Version 10.12.3 (16D32)

### What are the steps to reproduce the bug?

1. Navigate to https://codepen.io/traviskaufman/pen/MmKOae
1. Click the "view" link
1. Verify that an `mdc-menu` opens
1. Click away from the menu to dismiss it, which returns focus to the "view" link
1. Press the Enter key

### What is the expected behavior?

Since pressing enter on a link triggers its "click" event, the menu should open and remain open

### What is the actual behavior?

The menu is dismissed immediately

### Any other information you believe would be useful?

This probably has to do with the following chain of events:

1. Pressing enter opens and automatically focuses the menu
1. A `keyup` event listener is registered on the menu for enter
1. The enter key is released, causing the `keyup` event to fire on the now-focused menu

The solve for this could be to listen for _both_ a keydown _and_ a keyup such that the menu only handles keyboard events when a full press and release have been initiated on the component itself.

PR #2084, opened 2018-01-25
feat(menu): Add --selected class to menu items

williamernest:
fixes: #2031 
related: #536
Adds `mdc-list-item--selected` class to menu items when `rememberSelection` is set to `true`. 
Auto-focuses on the last selected item when opened when `rememberSelection` is set to `true`.


packages/mdc-menu/adapter.js
```diff
@@ -141,6 +141,34 @@ class MDCMenuAdapter {
 
   /** @param {string} height */
   setMaxHeight(height) {}
+
+  /** @param {number} index */
+  getOptionAtIndex(index) {}
```
lynnjepsen: does this return something? It is a getter after all...

packages/mdc-menu/constants.js
```diff
@@ -22,6 +22,7 @@ const cssClasses = {
   ANIMATING_OPEN: 'mdc-menu--animating-open',
   ANIMATING_CLOSED: 'mdc-menu--animating-closed',
   LIST_ITEM: 'mdc-list-item',
+  SELECTED_LIST_ITEM: 'mdc-list-item--selected',
```
lynnjepsen: you should be able to reference the constants on MDCListFoundation.strings

packages/mdc-menu/foundation.js
```diff
@@ -137,6 +142,12 @@ class MDCMenuFoundation extends MDCFoundation {
     this.anchorMargin_ = {top: 0, right: 0, bottom: 0, left: 0};
     /** @private {?AutoLayoutMeasurements} */
     this.measures_ = null;
+    /** @private {number} */
+    this.selectedIndex_ = -1;
+    /** @private {boolean} */
+    this.rememberSelection_ = false;
+    /** @private {boolean} */
+    this.keyDownWithinMenu_ = false;
```
lynnjepsen: this keeps track of whether the user *currently* has a keydown interaction happening on the `mdc-menu` component?

These sorts of fields can be vague...so add a comment about exactly what it is this field represents

packages/mdc-menu/foundation.js
```diff
@@ -188,13 +199,24 @@ class MDCMenuFoundation extends MDCFoundation {
...
+  }
+
   /**
    * @param {?number} focusIndex
    * @private
    */
   focusOnOpen_(focusIndex) {
     if (focusIndex === null) {
-      // First, try focusing the menu.
+      // First focus the last selected item if we're remembering selections and it has been set
```
lynnjepsen: flip this
"If this instance of MDCMenu remembers selections, and the user has made a selection, then focus the last selected item"

packages/mdc-menu/foundation.js
```diff
@@ -241,6 +263,9 @@ class MDCMenuFoundation extends MDCFoundation {
     const isArrowUp = key === 'ArrowUp' || keyCode === 38;
     const isArrowDown = key === 'ArrowDown' || keyCode === 40;
     const isSpace = key === 'Space' || keyCode === 32;
+    const isEnter = key === 'Enter' || keyCode === 13;
+
+    this.keyDownWithinMenu_ = isEnter || isSpace;
```
lynnjepsen: explain whats going on here

packages/mdc-menu/foundation.js
```diff
@@ -188,13 +199,24 @@ class MDCMenuFoundation extends MDCFoundation {
...
    */
   focusOnOpen_(focusIndex) {
     if (focusIndex === null) {
-      // First, try focusing the menu.
+      // First focus the last selected item if we're remembering selections and it has been set
+      if (this.rememberSelection_ && this.selectedIndex_ >= 0) {
+        this.adapter_.focusItemAtIndex(this.selectedIndex_);
+        return;
+      }
+      // Then, try focusing the menu.
```
lynnjepsen: You probably dont need this comment.

packages/mdc-menu/foundation.js
```diff
@@ -297,7 +322,12 @@ class MDCMenuFoundation extends MDCFoundation {
     const isEscape = key === 'Escape' || keyCode === 27;
 
     if (isEnter || isSpace) {
-      this.handlePossibleSelected_(evt);
+      // If our menu didn't catch the keydown, then the user likely hit the
```
lynnjepsen: There is no "we" or "our" in comments. The program is the thing running. Try and be specific about which part of the program is doing what action.

For example,

"If MDCMenu does not see a keydown event, then it assumes the user triggered an enter or space event"

Also, how is enter and space NOT a keydown event??

packages/mdc-menu/foundation.js
```diff
@@ -564,6 +597,38 @@ class MDCMenuFoundation extends MDCFoundation {
   isOpen() {
     return this.isOpen_;
   }
+
+  /** @return {?Element} */
```
lynnjepsen: I think we try NOT to reference HTML Element in our Foundation classes. Maybe this public method is not necessary...

packages/mdc-menu/constants.js
williamernest: MDCListFoundation doesn't exist

PR #2089, opened 2018-01-25
chore(demos): Fixed styles being loaded twice

pndewit:
Toolbar and typography styles are already included in common.scss.


pndewit:
No problem at all, glad I can help. Thank you for the awesome work you and your team are doing! 👍

PR #2092, opened 2018-01-25
fix(ripple): Fix unbounded ripple sizes

kfranqueiro:
Adds styles to correct base CSS-only unbounded ripple styles, branches JS logic to behave appropriately for unbounded ripples separately from bounded ones, and fixes demo page styles (h/t @yeelan0319 who pointed out these spurious styles to me a while ago).

The simplest way to test this is to look at ripple.html and test both the JS and CSS-only unbounded ripples; both should be the same size now, and the ripples' maximum size shouldn't exceed the bounding box (including padding) of the surface. Meanwhile, bounded ripples should continue to behave as they did before.

The selector for the added non-upgraded unbounded ripple styles is pretty ugly in this PR, but I have a follow-up to split bounded and unbounded radius styles into separate mixins, which will obviate the need for this ugly selector.

Refs #1067; the original issue in toolbar is improved but not 100% resolved; need to look into toolbar icon paddings.


acdvorak:
The unbounded JS ripple appears to be slightly truncated on the right and bottom sides in Chrome (but not other browsers):

Screenshot

Increasing the browser's zoom level to 125% seems to get rid of the clipping.

packages/mdc-ripple/foundation.js
```diff
@@ -488,17 +485,20 @@ class MDCRippleFoundation extends MDCFoundation {
...
     const maxDim = Math.max(this.frame_.height, this.frame_.width);
-    const surfaceDiameter = Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2));
 
-    // 60% of the largest dimension of the surface
-    this.initialSize_ = maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE;
+    // Surface diameter is treated differently for unbounded vs. bounded ripples.
+    // Unbounded ripple diameter is calculated smaller since the surface is expected to already be padded appropriately
+    // to extend the pointer target, and the ripple is expected to meet the edges of the padded hitbox
+    // (which is typically square). Bounded ripples, on the other hand, are fully expected to expand beyond the
+    // surface's longest diameter (calculated based on the diagonal plus a constant padding).
```
acdvorak: Can we add a clause to this sentence that explains why bounded ripples are expected to expand beyond the surface's longest diameter?

E.g., something along the lines of:

```
// surface's longest diameter (calculated based on the diagonal plus a constant padding),
// and are clipped at the surface's border via `overflow: hidden`.
```

packages/mdc-ripple/foundation.js
```diff
@@ -488,17 +485,20 @@ class MDCRippleFoundation extends MDCFoundation {
...
+    this.maxRadius_ = this.adapter_.isUnbounded() ? maxDim :
+      Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2)) + MDCRippleFoundation.numbers.PADDING;
```
acdvorak: It might be more readable to refactor the "bounded" expression a bit:

```js
const getDiagonal = () => {
  const hypotenuse = Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2));
  return hypotenuse + MDCRippleFoundation.numbers.PADDING;
};

this.maxRadius_ = this.adapter_.isUnbounded() ? maxDim : getDiagonal();
```

Or `getDiagonal` could be a separate private method if you prefer.

packages/mdc-ripple/foundation.js
```diff
@@ -488,17 +485,20 @@ class MDCRippleFoundation extends MDCFoundation {
...
     this.frame_ = this.adapter_.computeBoundingRect();
-
     const maxDim = Math.max(this.frame_.height, this.frame_.width);
-    const surfaceDiameter = Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2));
 
-    // 60% of the largest dimension of the surface
-    this.initialSize_ = maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE;
+    // Surface diameter is treated differently for unbounded vs. bounded ripples.
+    // Unbounded ripple diameter is calculated smaller since the surface is expected to already be padded appropriately
+    // to extend the pointer target, and the ripple is expected to meet the edges of the padded hitbox
```
acdvorak: Maybe rephrase this to:

```
// to create a larger tap target, and ...
```

packages/mdc-ripple/foundation.js
```diff
@@ -488,17 +485,20 @@ class MDCRippleFoundation extends MDCFoundation {
...
+    // (which is typically square). Bounded ripples, on the other hand, are fully expected to expand beyond the
+    // surface's longest diameter (calculated based on the diagonal plus a constant padding).
+    this.maxRadius_ = this.adapter_.isUnbounded() ? maxDim :
+      Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2)) + MDCRippleFoundation.numbers.PADDING;
 
-    // Diameter of the surface + 10px
-    this.maxRadius_ = surfaceDiameter + MDCRippleFoundation.numbers.PADDING;
+    // Ripple starts as a fraction of the largest dimension of the surface and scales from 1x to a value based on radius
```
acdvorak: Just to clarify, does "1x" refer to the initial starting size (i.e., a fraction of the largest dimension of the surface)?

If so, can we reword this to something like:

```
// Ripple starts as a fraction of the largest dimension of the surface
// and scales up to the calculated maximum radius.
```

packages/mdc-ripple/foundation.js
```diff
@@ -488,17 +485,20 @@ class MDCRippleFoundation extends MDCFoundation {
   /** @private */
   layoutInternal_() {
     this.frame_ = this.adapter_.computeBoundingRect();
-
     const maxDim = Math.max(this.frame_.height, this.frame_.width);
-    const surfaceDiameter = Math.sqrt(Math.pow(this.frame_.width, 2) + Math.pow(this.frame_.height, 2));
 
-    // 60% of the largest dimension of the surface
-    this.initialSize_ = maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE;
+    // Surface diameter is treated differently for unbounded vs. bounded ripples.
```
acdvorak: This entire comment block is super helpful. Without it, I would have no idea why the code is doing what it's doing.

You, sir, are the king of clear, concise comments🤴

packages/mdc-ripple/foundation.js
kfranqueiro: I felt like saying "tap" was device-ist :P but I am considering just using the word hitbox instead... but that might be my inner gamer talking.

packages/mdc-ripple/foundation.js
kfranqueiro: I'm normally against this sort of thing for a single use, but...maths, so yeah.

packages/mdc-ripple/foundation.js
kfranqueiro: I was referring to how the CSS initially has scale set to 1 and then scales it up from there to `fgScale_`, which is computed based on the initial size and the max radius. Your rewording doesn't explain why fgScale_ exists, IMO, which is why I added the comment. I'll try re-wording it though...

(TBH I'd love to experiment with actually incorporating .6 and 1 into the CSS scale transform to see if we can simplify the JS even more, but I probably shouldn't invest time on that right now.)

kfranqueiro:
RE the clipping, I'm inclined to leave it for the moment rather than go on another wild goose chase, given the other issues we've seen Chrome working through regarding issues with subpixels and GPU-composited elements. Looking at the latest release, I suspect this was happening even before these changes, though it's harder to see there.

PR #2093, opened 2018-01-25
chore(button): removed unused imports from css

moog16:


packages/mdc-button/_mixins.scss
```diff
@@ -15,9 +15,7 @@
 //
 
 @import "@material/elevation/mixins";
-@import "@material/ripple/mixins";
```
kfranqueiro: This file is definitely using mdc-ripple and mdc-states mixins, so this should stay

packages/mdc-button/_mixins.scss
```diff
@@ -15,9 +15,7 @@
...
 @import "@material/theme/functions";
-@import "@material/theme/mixins";
```
kfranqueiro: This file is still using `mdc-theme-prop` so we should leave this

packages/mdc-button/_mixins.scss
moog16: Ya it is also in mdc-button.scss. I haven't seen any one pattern to determine if all should be in the mdc-foo/mdc-foo.scss file or the mdc-foo/_mixins.scss file. I will add these back

acdvorak:
# Add BREAKING CHANGE to commit message before merging

PR #2095, opened 2018-01-25
chore(select): Update demos

williamernest:
Add more demos for alternate colors and disabled state for css-only select.

### DIFF
--- a/demos/select.html
+++ b/demos/select.html
@@ -117,6 +117,9 @@
                 <li class="mdc-list-item" role="option" id="vegetables" aria-disabled="true" tabindex="0">
                   Vegetables
                 </li>
+                <li class="mdc-list-item" role="option" id="noodles" tabindex="0">
+                  Noodles
+                </li>
               </ul>
             </div>
           </div>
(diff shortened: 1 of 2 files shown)


demos/select.html
```diff
@@ -142,17 +145,31 @@ <h2 class="mdc-typography--title">Fully-Featured Component</h2>
...
-            <option value="dairy">Milk, Yogurt, and Cheese</option>
-            <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
-            <option value="fats">Fats, Oils, and Sweets</option>
-          </select>
-          <div class="mdc-select__bottom-line"></div>
+        <section class="box-demo-wrapper">
+          <div id="demo-css-only-select" class="mdc-select">
+            <select class="mdc-select__surface">
+              <option value="" selected>Pick a food group</option>
+              <optgroup label="First Group">
```
acdvorak: I'm not sure we want to use `<optgroup>` in our demos; the spec doesn't really have anything like it AFAIK.

demos/select.html
```diff
@@ -195,6 +213,13 @@ <h2 class="mdc-typography--title">CSS Only</h2>
         disabledCb.addEventListener('change', function() {
           select.disabled = disabledCb.checked;
         });
+
+        cssAlternateColors.addEventListener('change', function() {
+          selectCss.classList[cssAlternateColors.checked ? 'add' : 'remove']('demo-select-custom-colors');
+        });
+        cssDisabled.addEventListener('change', function() {
+          cssDisabled.checked ? selectCss.setAttribute('disabled', 'disabled') : selectCss.removeAttribute('disabled');
```
kfranqueiro: Can't this be simplified to `selectCss.disabled = cssDisabled.checked`?

williamernest: No, that doesn't work. The `selectCss` variable is the `div` container that contains the select and the bottom line. A div doesn't have a disabled value, so setting it requires that we add the attribute to the div tag itself. 

Our css styles apply to a disabled selector (`.mdc-select[disabled]`) to "Imitate native disabled functionality".

kfranqueiro: Oh, oops, I'm dumb and was assuming it was the select element itself, and forgot the context.

kfranqueiro: Just to track the conversation here, I discussed this more offline with Will and realized that we've actually got some awkwardness with disabled CSS-only selects, in that you actually need to apply the disabled attribute to both the top-level element (to make it look disabled) _and_ the select (to make it actually behave disabled). Needing to apply the `disabled` attribute to an element for which it has no native functional meaning is really weird, and likely a remnant of select's structure prior to #1097.

PR #2097, opened 2018-01-25
chore(menu): removed dark theme

moog16:


packages/mdc-menu/mdc-menu.scss
```diff
@@ -16,7 +16,6 @@
 
 @import "@material/animation/variables";
 @import "@material/elevation/mixins";
-@import "@material/theme/mixins";
```
kfranqueiro: Don't remove this; `mdc-theme-prop` is still used in this file.

packages/mdc-menu/mdc-menu.scss
```diff
@@ -118,38 +114,20 @@ $mdc-menu-scale-duration: .12s;
...
   .mdc-list-item__graphic {
     color: rgba(black, .54);
-
-    @include mdc-theme-dark(".mdc-menu", true) {
-      color: rgba(white, .54);
-    }
   }
 
   .mdc-list-item[aria-disabled="true"] {
     @include mdc-theme-prop(color, text-disabled-on-light);
```
acdvorak: `text-disabled-on-background`

packages/mdc-menu/mdc-menu.scss
```diff
@@ -118,38 +114,20 @@ $mdc-menu-scale-duration: .12s;
...
   }
 
-  .mdc-list-divider {
-    @include mdc-theme-dark(".mdc-menu", true) {
-      border-color: rgba(white, .12);
-    }
-  }
-
   .mdc-list-item__graphic {
     color: rgba(black, .54);
```
acdvorak: Replace this line with:

`@include mdc-theme-prop(color, text-secondary-on-background);`

packages/mdc-menu/mdc-menu.scss
```diff
@@ -27,10 +27,6 @@ $mdc-menu-scale-duration: .12s;
 .mdc-menu {
```
acdvorak: Can we change `background-color: white` to `background-color: $mdc-theme-background`?

(We'll need to add `@import "@material/theme/variables"` to the top of this file)

packages/mdc-menu/mdc-menu.scss
moog16: I traced it back through mdc-theme, so it should be the same, but we don't have a demo to showcase this. Should we add a menu item with a graphic?

PR #2098, opened 2018-01-25
chore(select): removed dark theme

moog16:


packages/mdc-select/mdc-select.scss
```diff
@@ -18,7 +18,6 @@
 @import "@material/animation/functions";
 @import "@material/animation/variables";
 @import "@material/typography/mixins";
-@import "@material/theme/mixins";
```
moog16: add back...use of mdc-theme-prop

PR #2100, opened 2018-01-25
chore(snackbar): remove dark theme

moog16:


packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -16,7 +16,6 @@
 
 @import "@material/animation/functions";
 @import "@material/rtl/mixins";
-@import "@material/theme/mixins";
```
kfranqueiro: Don't remove this; `mdc-theme-prop` is still used.

PR #2101, opened 2018-01-25
chore(switch): remove dark theme

moog16:


packages/mdc-switch/mdc-switch.scss
```diff
@@ -15,7 +15,6 @@
 //
 
 @import "@material/elevation/mixins";
-@import "@material/theme/mixins";
```
moog16: This one is ok...i ran npm run build

demos/switch.html
```diff
@@ -119,47 +107,6 @@ <h2>Switch on Light Theme - Disabled</h2>
           <label for="basic-switch--disabled" class="mdc-switch-label">off/on</label>
         </div>
       </section>
-
-      <section class="example mdc-theme--dark">
-        <h2>Switch on Dark Theme</h2>
```
acdvorak: Can we rename the two remaining headings?

* "Switch on Light Theme" -> "Enabled"
* "Switch on Light Theme - Disabled" -> "Disabled"

PR #2102, opened 2018-01-25
chore: Add `npm run build` to test command to fail on broken dist build

kfranqueiro:
This will help us catch errors like what recently happened on master with 2 recent merges.

(This is going to fail Travis CI until those recent issues are fixed.)


kfranqueiro:
This...didn't end up failing even when I expected it to. 🤔 going to close, amend branch, and open new PR.

Issue #1332, opened 2017-09-19
Please keep SASS optional


jamesmfriedman:
## Feature Requests
I keep up to date with changes to this lib, and have been loving using it. I was looking at the the most recent changelog and found that mdc-button--primary and accent were removed in 3e3c869ff46e914b78b02b7be4a18658c74c9940 and now the suggestion for the breaking change is to use SASS mixins to replicate the behavior.

I'd would personally really love for SASS to remain something that is optional for this library. Looking at the code that was removed for the buttons, it's not something that could be easily replicated without SASS, but primary and accent buttons are used frequently in any design I've ever worked on.

In the end, this will allow people with all different types of css processors (or no processors) to get the most out of the lib.

Cheers!
James

LeviHassel:
I second this. I'd love if there was a clean way to change the global primary and secondary colors without SAAS.

jamesmfriedman:
@LeviHassel there are using css vars. Just redeclare them on the :root element. Before this change, the buttons obeyed those vars which was excellent.

LeviHassel:
Oh thanks! I'm new to CSS, so I wasn't aware of the :root element and have just been declaring the colors for everything on the html element. I still agree that the buttons should obey those vars though.

lynnjepsen:
What if we added support for some custom CSS properties:

- `--mdc-button-ink-color`
- `--mdc-button-container-fill-color`
- etc.

Then you could configure a text button to use you're non-Sass theme like this:
```
--mdc-button-ink-color: var(--mdc-theme-secondary);
```

And you're raised button like this:
```
--mdc-button-container-fill-color: var(--mdc-theme-secondary);
--mdc-button-ink-color: var(--mdc-theme-text-primary-on-secondary);
```

In fact you don't even have to use `var(--mdc-theme-secondary)`...you could use `var(--tertiary-color-because-i-am-special)`. Material design felt that "primary" and "secondary" color theme slots might be too limiting. We wanted to open up customization of a button to work with a bigger theme palette.

Of course, IE/Edge don't really support custom CSS properties...so if you don't have Sass AND you want to support IE/Edge, we don't have an answer.

And, there are some Sass mixins which we cannot support with custom CSS properties. For example the `mdc-button-filled-accessible` requires if/else logic only possible in Sass, not runtime CSS. If any mixin's implementation changes such that we CANT support it with a custom CSS property...we would have to remove support of that custom CSS property.

Thoughts?

mrijken:
[...]

PR #2106, opened 2018-01-25
chore(text-field): remove dark theme

moog16:


aaronhudon:
Does this imply dark theme support is being pulled from controls, or is this just for the demo pages showing both versions at the same time...?

moog16:
@aaronhudon we are pulling dark theme styles out from scss. Its become too much to manage. We want the end developer to be using our sass mixins to style the components.

aaronhudon:
@moog16 so this means the `.mdc-theme--dark` modifier is also being removed?

acdvorak:
Thanks for asking, @aaronhudon - we always appreciate feedback from valued members of the community!

I know it's disappointing 🙁   But sadly it's necessary to keep our code healthy and maintainable, as @moog16 alluded to.

For a more detailed explanation of our motivation and reasoning, see [my response to #1332](https://github.com/material-components/material-components-web/issues/1332#issuecomment-335724659).

The TL;DR is that instead of using `--theme-dark` classes, you can customize the theme variables (`$mdc-theme-background`) and use our component theme mixins to achieve the same result.

Our goal is to make that relatively easy, so please let us know if it's not. (Keep in mind that we're still actively working on theming and customization, so it's very much a work in progress.)

aaronhudon:
@acdvorak thanks for the followup, I trust a guide on this new method of implementing a dark theme will be posted in the docs?

acdvorak:
@aaronhudon There's an example on our [theme demo page](https://material-components-web.appspot.com/theme/index.html?theme=dark) (the source is in
[`demos/theme/theme-dark.scss`](https://github.com/material-components/material-components-web/blob/f399384be2c607c5618fe181f8b5b049f46e072a/demos/theme/theme-dark.scss)).

In short, you need to A) override `$mdc-theme-background` to a dark color (`#333` or whatever), and B) call a few component theme mixins. We're working on removing the need to call theme mixins, but for now that's the easiest way to do it.

packages/mdc-textfield/_variables.scss
```diff
@@ -15,41 +15,27 @@
 //
 
 $mdc-text-field-error-on-light: #d50000;
```
acdvorak: We can probably remove the `-on-error` suffix from all of the remaining color variables

moog16: I removed all the -on-light as well as -light-* variable names.

packages/mdc-textfield/_variables.scss
```diff
@@ -14,15 +14,14 @@
...
-$mdc-text-field-underline-on-light-idle: rgba(black, .5);
-$mdc-text-field-light-background: rgba(white, 1);
-$mdc-text-field-light-label: rgba(black, .6);
-$mdc-text-field-light-placeholder: rgba(black, .38);
+$mdc-text-field-error: #d50000;
+$mdc-text-field-divider: rgba(black, .12);
+$mdc-text-field-disabled-border: rgba(#231f20, .26);
+$mdc-text-field-disabled-icon: rgba(black, .3);
+$mdc-text-field-underline: rgba(black, 1);
+$mdc-text-field-underline: rgba(black, .5);
```
acdvorak: Add `-idle` back

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -33,14 +33,14 @@
 // postcss-bem-linter: define text-field
 
 .mdc-text-field {
-  @include mdc-text-field-bottom-line-color($mdc-text-field-underline-on-light-idle);
-  @include mdc-text-field-hover-bottom-line-color($mdc-text-field-underline-on-light);
+  @include mdc-text-field-bottom-line-color($mdc-text-field-underline);
```
acdvorak: Add `-idle` back

packages/mdc-textfield/_variables.scss
```diff
@@ -14,15 +14,14 @@
...
-$mdc-text-field-underline-on-light: rgba(black, 1);
-$mdc-text-field-underline-on-light-idle: rgba(black, .5);
-$mdc-text-field-light-background: rgba(white, 1);
-$mdc-text-field-light-label: rgba(black, .6);
-$mdc-text-field-light-placeholder: rgba(black, .38);
+$mdc-text-field-error: #d50000;
+$mdc-text-field-divider: rgba(black, .12);
+$mdc-text-field-disabled-border: rgba(#231f20, .26);
+$mdc-text-field-disabled-icon: rgba(black, .3);
+$mdc-text-field-underline: rgba(black, 1);
```
acdvorak: Can we rename this to `$mdc-text-field-underline-hover`, since that's what it's used for?

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -33,14 +33,14 @@
...
+  @include mdc-text-field-hover-bottom-line-color($mdc-text-field-underline);
```
acdvorak: `$mdc-text-field-underline-hover`

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -33,14 +33,14 @@
...
   @include mdc-text-field-focused-bottom-line-color(primary);
   @include mdc-text-field-ink-color(text-primary-on-light);
-  @include mdc-text-field-label-color($mdc-text-field-light-label);
+  @include mdc-text-field-label-color($mdc-text-field-label);
   @include mdc-text-field-helper-text-color(text-hint-on-light);
-  @include mdc-text-field-fullwidth-bottom-line-color($mdc-text-field-divider-on-light);
-  @include mdc-text-field-icon-color($mdc-text-field-underline-on-light);
+  @include mdc-text-field-fullwidth-bottom-line-color($mdc-text-field-divider);
+  @include mdc-text-field-icon-color($mdc-text-field-underline);
```
acdvorak: `$mdc-text-field-underline-hover`

demos/text-field.html
```diff
@@ -231,11 +230,21 @@ <h2>Text Field box</h2>
...
+          <input id="box-persistent-helper-text" type="checkbox" disabled>
+          <label for="box-persistent-helper-text">Make helper text persistent</label>
+        </div>
         <script>
           demoReady(function() {
+            var section = document.getElementById('text-field-box-example');
             var tfEl = document.getElementById('tf-box-example');
             var tf = new mdc.textField.MDCTextField(tfEl);
             var wrapper = document.getElementById('demo-tf-box-wrapper');
+            var helperText = section.querySelector('.mdc-text-field-helper-text');
```
acdvorak: Let's give this element an `id` attribute and use `getElementById` instead

demos/text-field.html
```diff
@@ -231,11 +230,21 @@ <h2>Text Field box</h2>
           <input id="box-required" type="checkbox">
           <label for="box-required">Required</label>
```
acdvorak: While we're here, can we rename this label to "Required and must be at least 8 characters"?

demos/text-field.html
moog16: Also will remove line 243...no longer using section

theenoahmason:
@moog The dark theme is being removed from the entire package? as in even a global `.mdc-theme--dark` will not work?

If yeah, that's kind of a bummer! Will any of the CSS vars be removed like `...on-dark`, `...on-light`?

Will classes be removed like `...on-dark`, `...on-light`?

moog16:
@theenoahmason yes `.mdc-theme--dark` and `@mixin mdc-theme-dark` will be removed from the package. Having all the dark theme support was complicating our CSS, and as a team we decided to remove it. We have been moving towards sass mixins to achieve the same affect. 

We will not be removing CSS vars as you mentioned (`*-on-dark`, `*-on-light`, etc). These will remain as they do not require extra css rules to support them.

PR #2108, opened 2018-01-25
fix(checkbox): Delay checked-unchecked fade

patrickrodee:
Delay the fade of the checkbox background by 90ms during the transition from checked to unchecked. This makes the checked to unchecked animation a better inversion of the unchecked to checked animation.


patrickrodee:
## Before
checkbox-fade--before

## After
checkbox-fade--after

PR #2109, opened 2018-01-26
chore: Add `npm run build` to npm test and travis

kfranqueiro:
This will allow us to catch regressions that break the build, even if they don't break the demo server (which is possible e.g. with Sass due to the demo styles always including certain files which may mask missing imports elsewhere).

~~This branch is purposely behind master as of the fix to the build issues, in order to test that travis actually fails with this update.~~ Now that I've successfully confirmed that Travis CI fails when expected, I've reset the branch to only the configuration changes.


kfranqueiro:
Not sure why travis still isn't failing...so trying again with a change _on this branch_ that intentionally breaks the build instead.

kfranqueiro:
Confirmed that Travis CI fails now, by intentionally breaking the build in this PR by removing an import.

image

Now I'll reset this branch to just the changes to update `npm test` and Travis CI so someone can review it.

package.json
```diff
@@ -19,7 +19,7 @@
     "lint": "npm-run-all --parallel lint:*",
     "postinstall": "lerna bootstrap",
     "pretest": "npm run lint && node scripts/check-imports.js",
-    "test": "npm run test:unit && npm run test:closure && npm run test:dependency",
+    "test": "npm run test:unit && npm run test:closure && npm run test:dependency && npm run build",
```
acdvorak: Should we append ` && npm run clean` to delete the generated `build/` directory?

package.json
```diff
@@ -19,7 +19,7 @@
     "lint": "npm-run-all --parallel lint:*",
     "postinstall": "lerna bootstrap",
     "pretest": "npm run lint && node scripts/check-imports.js",
-    "test": "npm run test:unit && npm run test:closure && npm run test:dependency",
+    "test": "npm run test:unit && npm run test:closure && npm run test:dependency && npm run build && npm run clean",
```
acdvorak: What about `&& npm run build; npm run clean` so that it always runs `clean`, even if `build` fails?

package.json
```diff
@@ -3,13 +3,13 @@
   "description": "Material Components Web",
   "license": "Apache-2.0",
   "scripts": {
-    "build": "mkdirp build && webpack --progress --colors",
+    "build": "npm run clean && mkdirp build && webpack --progress --colors",
     "build:demo": "cross-env MDC_ENV=development MDC_BUILD_STATIC_DEMO_ASSETS=true MDC_WRAP_CSS_IN_JS=false npm run build",
     "build:min": "mkdirp build && cross-env MDC_ENV=production webpack -p --progress --colors",
     "changelog": "standard-changelog -i CHANGELOG.md -k packages/material-components-web/package.json -w",
-    "clean": "del-cli build/** build",
+    "clean": "del-cli build/** build .closure-tmp",
```
acdvorak: You _might_ need to do:

`"clean": "del-cli build/** build .closure-tmp/** .closure-tmp"`

To force the directory to be fully deleted on macOS.

Issue #1991, opened 2018-01-18
Changing all directory import statements to explicitly import index.js


jpray:
Hi, I love this project!  Would you all consider accepting a PR that changed all directory import statements from `import './foo'` to `import './foo/index.js'`?  I know node and webpack support the current style but other module loaders like systemjs don't.  Thanks for considering this.

bonniezhou:
MDC Web only intends to support the formats listed in [importing-js.md](https://github.com/material-components/material-components-web/blob/master/docs/importing-js.md).
However we encourage you to ask SystemJS to add support for the `import './foo'` syntax!

jpray:
Bummer, I mean, the syntax is still ES2015.  The issue is just that having /index.js be implicit isn't part of any standard loader spec that I know of.  So a module loader that sticks strictly to a standard like whatwg loader can't support this... and this makes since because it would be bad for a browser to make multiple calls to the server seeing if foo.js or foo/index.js existed.  

It wouldn't be a breaking change.  I'd include tests in the PR. :)

bonniezhou:
Ok...you've convinced us. :) However we still think that the `import {...} from 'path/to/index'` (without the `.js` extension) makes most sense, as per http://2ality.com/2014/09/es6-modules-final.html. Adding @kfranqueiro and @acdvorak to this thread in case they want to speak more on it.

Feel free to submit a pull request that:
- [ ] changes all directory imports from `import './foo'` to `import './foo/index'`

jpray:
Sweet!  I'm fine with omitting the '.js' part.  I've submitted https://github.com/material-components/material-components-web/pull/2111 for this issue.

PR #2111, opened 2018-01-26
refactor: Import index files explicitly to maximize compatibility with module loaders

jpray:
This PR addresses https://github.com/material-components/material-components-web/issues/1991

### DIFF
--- a/scripts/check-imports.js
+++ b/scripts/check-imports.js
@@ -41,7 +41,7 @@ let exitCode = 0;
 main();
 
 function main() {
-  globSync('packages/mdc-*/**/*.js')
+  globSync('packages/**/*.js')
     .filter((p) => !(/\/node_modules\/|\/dist\//).test(p))
     .forEach(check);
   process.exit(exitCode);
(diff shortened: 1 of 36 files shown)


jpray:
I signed it!

jpray:
Thanks for catching that.  I've made the update.

scripts/check-imports.js
```diff
@@ -41,7 +41,7 @@ let exitCode = 0;
 main();
 
 function main() {
-  globSync('packages/mdc-*/**/*.js')
+  globSync('packages/**/*.js')
```
bonniezhou: Was there a reason that this was changed?

jpray: Yes. That glob wasn't including packages/material-components-web/index.js, and all of these checks for imports should apply there as well.

scripts/check-imports.js
```diff
@@ -87,6 +88,30 @@ function assertImportResolvesWithinPkgRoot(source, pkgRoot, srcFile, node) {
   }
 }
 
+function assertNoLocalDirectoryImports(source, pkgRoot, srcFile, node) {
```
kfranqueiro: Thanks for adding this, btw :star:

jpray: My pleasure.  Thanks for blessing it and merging it : )

PR #2112, opened 2018-01-26
feat(ripple): Split radius mixin into bounded/unbounded versions

### LINKED ISSUE #2074: Split `mdc-ripple-radius` into a bounded and unbounded version
This is a follow up to the fix to #1067

kfranqueiro:
This leads to less code in each component that uses ripple, since generally you're only interested in a component's surfaces being either bounded or unbounded, not both.

Saves roughly 6.7K / 0.5K gzipped in material-components-web.css.

Fixes #2074

BREAKING CHANGE: mdc-ripple-radius is replaced by mdc-ripple-radius-bounded and mdc-ripple-radius-unbounded; use one or the other as appropriate for the surface


kfranqueiro:
This is semi-related to #1067; I came up with this in the process of investigating the root cause of oversized unbounded ripples.

kfranqueiro:
I noticed for some reason CSS-only checkbox/radio lost their hover styles somehow, so I need to investigate that.

PR #2114, opened 2018-01-26
refactor(text-field): Remove default empty adapter parameter from constructor

bonniezhou:


packages/mdc-textfield/bottom-line/foundation.js
```diff
@@ -54,7 +54,7 @@ class MDCTextFieldBottomLineFoundation extends MDCFoundation {
   /**
    * @param {!MDCTextFieldBottomLineAdapter=} adapter
```
acdvorak: We should probably remove the `=` since this param is no longer optional (ditto below)

PR #2116, opened 2018-01-26
fix(select): Remove css version

williamernest:
Remove css only verion of select.


packages/mdc-select/README.md
```diff
@@ -200,37 +200,11 @@ To disable a list item, set `aria-disabled` to `"true"`, and set `tabindex` to `
 </div>
 ```
 
-### Using the Pure CSS Select
```
kfranqueiro: We should presumably also remove the tips/tricks section in this readme since it no longer applies (there is no longer a css-only MDC Select)

packages/mdc-select/README.md
```diff
@@ -200,37 +200,11 @@ To disable a list item, set `aria-disabled` to `"true"`, and set `tabindex` to `
...
-  <div class="mdc-select__bottom-line"></div>
-</div>
-```
-
 #### CSS Classes
 
 | Class                    | Description                                     |
 | ------------------------ | ----------------------------------------------- |
-| `mdc-select`             | A pure css `select` element                     |
+| `mdc-select`             | Mandatory                                       |
```
kfranqueiro: Include trailing period here for consistency with other entries

packages/mdc-select/mdc-select.scss
```diff
@@ -188,8 +188,7 @@ $mdc-select-menu-transition: transform 180ms $mdc-animation-standard-curve-timin
   }
 }
 
-.mdc-select--disabled,
```
kfranqueiro: Just leaving a note here that the `@each` after this will be removed in #2098 anyway, so it probably doesn't merit touching in this PR.

kfranqueiro:
PS: This probably shouldn't just be a chore, since it directly affects users.

williamernest:
Should it be a feature or fix?

kfranqueiro:
As weird as it is calling it a "feature", it is something that may affect users of the component so I'd call it that, and it probably warrants a breaking change note.

(What'd we do when we removed CSS-only text field?)

demos/select.html
```diff
@@ -134,32 +134,6 @@ <h2 class="mdc-typography--title">Fully-Featured Component</h2>
...
-              <option value="fruit">Fruit</option>
-              <option value="dairy">Milk, Yogurt, and Cheese</option>
-              <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
-              <option value="fats">Fats, Oils, and Sweets</option>
-            </select>
-            <div class="mdc-select__bottom-line"></div>
-          </div>
-        </section>
-        <div>
-          <input type="checkbox" id="css-alternate-colors">
```
kfranqueiro: You should also remove the JS that references these checkboxes that are no longer present.

kfranqueiro:
I definitely share your concern, but at the same time, text field and select already had distinct DOM structures for CSS-only vs. JS components, which IMO already posed a problem for progressive enhancement.

We recently noticed some problems pertaining to certain states (e.g. disabled state) of CSS-only text fields and selects. Moreover, we're going to be doing more style work and refactoring in this area, and feel that it makes sense to reduce styles and scope until the heavy lifting is done, then re-evaluate what the best approach looks like to accomplish this.

kfranqueiro:
I found some more things you can remove:

This should be specific to actual `select` elements in IE/Edge:
https://github.com/material-components/material-components-web/blob/959398da21ba28d1f597e17f9196cbd87e9c32e5/packages/mdc-select/mdc-select.scss#L76-L78

And for `select` elements in general (keep the overflow hidden though):
https://github.com/material-components/material-components-web/blob/959398da21ba28d1f597e17f9196cbd87e9c32e5/packages/mdc-select/mdc-select.scss#L88-L89

PR #2120, opened 2018-01-29
Dialog scroll-lock bugfix

tompng:
When calling dialog.destroy() soon after dialog.close(), the whole window gets scroll-locked.
(when destroy() is called before the close transition ends)

`this.enableScroll_()` cannnot placed inside `if(this.isOpen_){}`
because `this.close()` (it turns `this.isOpen_ = false;`) does not `enableScroll_()`

```javascript
destroy() {
  if (this.isOpen_) {
    ...
    // this.enableScroll_(); // can't be here. scroll-lock state is not syncrhonized with the flag `isOpen_`.
  }
  this.enableScroll_();
}
```

other possibilities
```javascript
destroy() {
  if (this.isOpen_) {
    ...
    this.enableScroll_(); // can be here. scroll-lock state is syncrhonized with the flag `isOpen_`.
  }
}
close() {
  this.isOpen_ = false;
  ...
  this.enableScroll_(); // add this
}
```


tompng:
I signed it!

williamernest:
@tompng Thanks for submitting this PR. Is there a reason that `this.enableScroll_()` isn't just moved from the `handleTransitionEnd` function to line 97, just after 'this.isOpen_ = false;'. It seems like a better place to enable scrolling behavior than to always call it in the `destroy` function.

tompng:
@williamernest
> to line 97, just after 'this.isOpen_ = false;'.

I agree, fixed.

Little changes is that, scrolling is enabled before closing animation starts(originally, it was enabled after animation ends), but I think it's not a problem.

Issue #1702, opened 2017-12-06
Allow clients to specify that a menu should open *without* animation


lynnjepsen:
This should include options for mobile and desktop. Sometimes the desktop menus are quite big, so the animation should be less heavy

PR #2127, opened 2018-01-30
feat(menu): Add quickOpen option. 

williamernest:
Fixes: #1702 
Disables open and closing animations for the mdc-menu element.


packages/mdc-menu/foundation.js
```diff
@@ -208,6 +210,11 @@ class MDCMenuFoundation extends MDCFoundation {
     this.setSelectedIndex(-1);
   }
 
+  /** @param {boolean} quickOpen */
+  quickOpen(quickOpen) {
```
kfranqueiro: Should this be called `setQuickOpen`? The current name in the context of a method makes it sound like it's actually going to open the menu.

packages/mdc-menu/foundation.js
```diff
@@ -555,7 +562,10 @@ class MDCMenuFoundation extends MDCFoundation {
    */
   open({focusIndex = null} = {}) {
     this.adapter_.saveFocus();
-    this.adapter_.addClass(MDCMenuFoundation.cssClasses.ANIMATING_OPEN);
+
+    if (!this.quickOpen_) {
+      this.adapter_.addClass(MDCMenuFoundation.cssClasses.ANIMATING_OPEN);
```
kfranqueiro: We should probably also conditionalize the `setTimeout` code below (and same for within `close`), although it's effectively a no-op in this case right now.

test/unit/mdc-menu/simple.foundation.test.js
```diff
@@ -116,7 +116,14 @@ test('#init throws error when the necessary DOM is not present', () => {
 testFoundation('#open adds the animation class to start an animation',
   ({foundation, mockAdapter}) => {
     foundation.open();
-    td.verify(mockAdapter.addClass(cssClasses.ANIMATING_OPEN));
+    td.verify(mockAdapter.addClass(cssClasses.ANIMATING_OPEN), {times: 1});
+  });
+
+testFoundation('#open does not add the animation class to start an animation when quickOpen is false',
```
kfranqueiro: Missing a complementary test for `#close`? (Also test both `addClass` and `removeClass` after addressing my comment above.)

packages/mdc-menu/README.md
```diff
@@ -150,6 +153,7 @@ Property | Value Type | Description
 `open` | Boolean | Proxies to the foundation's `isOpen`/(`open`, `close`) methods. 
 `items` | Array<Element> | Proxies to the foundation's container to query for all `.mdc-list-item[role]` elements.
 `itemsContainer` | Element | Queries the foundation's root element for the `mdc-menu__items` container element.
+`quickOpen` | Boolean | Proxies to the foundation's `quickOpen()` method.
```
kfranqueiro: Description needs updating to `setQuickOpen()` now

kfranqueiro: It also occurs to me that `setQuickOpen` needs to be documented under the Foundation API... and make sure to mention that it affects both open _and_ close animations. (Which makes me think maybe this isn't the best name, but not sure what is... maybe flip it to `enableAnimations`, but maybe that's too vague?)

packages/mdc-menu/README.md
```diff
@@ -201,6 +205,7 @@ Method Signature | Description
 `open({focusIndex: ?number}) => void` | Opens the menu. Optionally accepts an object with a `focusIndex` parameter to indicate which list item should receive focus when the menu is opened. 
 `close(evt: ?Event)` | Closes the menu. Optionally accepts the event to check if the target is disabled before closing the menu.  
 `isOpen() => boolean` | Returns a boolean indicating whether the menu is open. 
+`setQuickOpen(quickOpen: boolean) => void` | Sets the menu to open and close without animation when the `open/close` methods are called.
```
kfranqueiro: Nit #1: "Whether the menu should open and close ..."
Nit #2: `open`/`close` (separate backticks around each word, leave the slash outside them)

williamernest: I changed it to "Sets whether the menu should open and close..." based on every other "set" method description in this readme starting with "Sets".

PR #2133, opened 2018-01-30
fix(select): updates styles to match text-field

pndewit:
If I'm correct Material Design doesn't really have a select component, but do have a text-field with dropdown capabilities: https://material.io/guidelines/components/text-fields.html#text-fields-layout

 So I updated the select component's layout to match (or at least be close to) the text-field component layout.

EDIT: it's also missing a margin that is present in the text-field, but I wasn't sure about adding it.

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -44,17 +44,17 @@ $mdc-select-menu-transition: transform 180ms $mdc-animation-standard-curve-timin
   align-items: center;
   justify-content: flex-start;
   box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


kmmbvnr:
Yep,  I have the same fix in my code, to make text field and select looks same when placed on one line.

It would be nice to see that in the core package.

williamernest:
Thanks for submitting this. We are currently working on implementing a non-box variant for the mdc-select to match the text-field styling and moving the box styles into a `mdc-select--box` class.
See #2149

williamernest:
Addressed with PR #2149

PR #2134, opened 2018-01-30
fix(text-field): bottom line height should be 1px

pndewit:
According to the Material Design guidelines, the bottom line should only be 1px high.


pndewit:
Whoops, only noticed after creating this PR that user interactions triggers bottom-line to be 2px high..

PR #2138, opened 2018-01-30
fix(toolbar): Fix icon padding for ripples, and vertical alignment in FF/IE/Edge

### LINKED ISSUE #1067: Correct unbounded ripple diameter

### LINKED ISSUE #1296: Vertical alignment issues in nav bar icons
Basic code: https://gist.github.com/Extarys/cba6618f4c76de8733d3ab222de51ec0
Result:
alignmentmaterial

I also notice the same issue in the demo website where the right github icon isnt aligned properly and the drawer icon on the left have the same issue.

They appear a little bit too high, not centered.

EDIT: I copied the wrong issue, damn.

kfranqueiro:
* Change `align-items` to `center` for toolbar section element to fix vertical alignment in IE/Edge and also mitigate the adjustments to vertical padding in this PR
* Adjust vertical padding on toolbar items so that ripples applied to icons don't extend further than they should
* Add margin instead of increasing padding on leading and last toolbar icons, again to avoid excessive ripple size

This should not cause a visible difference in icon padding/arrangement (other than vertically aligning it properly in IE/Edge).

Fixes #1067 (finally)
Also fixes #1296 related to vertical alignment (I re-checked and indeed it's not just IE/Edge, it affects FF too, huh.)


demos/toolbar/index.html
```diff
@@ -94,9 +95,9 @@
                 <span class="mdc-toolbar__title">Toolbar</span>
               </section>
               <section class="mdc-toolbar__section mdc-toolbar__section--align-end" role="toolbar">
-                <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Download" alt="Download">file_download</a>
-                <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Print this page" alt="Print this page">print</a>
-                <a href="#" class="material-icons mdc-toolbar__icon" aria-label="Bookmark this page" alt="Bookmark this page">bookmark</a>
+                <a href="#" class="material-icons mdc-toolbar__icon mdc-ripple-surface demo-toolbar-surface" data-mdc-ripple-is-unbounded aria-label="Download" alt="Download">file_download</a>
```
lynnjepsen: Should we call `@include mdc-states(text-primary-on-primary)` from within the CSS for `mdc-toolbar__icon`? And then remove this `mdc-ripple-surface` from the demo?

kfranqueiro: I can probably add ripple styles to `mdc-toolbar__icon` if it's safe to assume all toolbar icons are intended to be interactable (which I think is the case judging by the spec).

If I do that, I should probably also add logic to toolbar's JS to instantiate ripples for its icons.

lynnjepsen: I think we should do that.

kfranqueiro:
I just realized the fix this branch includes for #1296 actually causes problems for flexible toolbar, so I might need to find a different way to fix the vertical alignment...

kfranqueiro:
@williamernest was cool enough to find a resolution to the snag I ran into with flexible toolbar, and I've committed the rest of my stuff on top of that. Will's now hammering out one remaining issue but then I think this is ready for review again.

PR #2141, opened 2018-01-30
fix(tabs): centered and adjusted vertical placement of css-only tab indicator

### LINKED ISSUE #2117: CSS only tab indicator has a positioning bug, and is the wrong color when using .mdc-theme--dark

0.28

Should be all, this is a CSS issue, checked in chrome, edge and firefox.

### What OS are you using?

Windows 10

Visit the tab demos. 
**Positioning:** bug found in all CSS Only tabs. 
**Color:** bug found in all CSS Only tabs when using .mdc-theme--dark.

**Positioning:** CSS Only tab indicator should reach the bottom of the tab, and should be centered.
**Color:** CSS Only tab indicator should invert it's color when using `.mdc-theme--dark`.

**Positioning:** CSS Only tab indicator is 2px away from the bottom of the tab, and is aligned left.
**Color:** CSS Only tab indicator stays in it's light context when using `.mdc-theme--dark`

## Positioning

The CSS only tabs have the indicator **inside** the tab (`.mdc-tab__indicator`), whereas the JS tabs have their indicator **outside** the tab, in the tab bar itself (`.mdc-tab-bar__indicator`). 

Both, however, have a rule of `bottom: 2px`. This means that the tab indicator shows up a few pixels higher than the tab bar indicator, and is a few px too close to the tab text. The indicator is also 4px short, presumably so that the hover state and the active state don't touch as stated by `width: calc(100% - 4px);`. This means the indicator isn't centered.

poscapture 10

### Solution

Only the JS tab bar indicator needs the 2px bottom nudge, the CSS only tab indicator should sit at `bottom: 0;`. 

Also, Nudging the indicator 2px to the left splits the 4px "shortening" of the tab bar in half, and centers the indicator; while still leaving space between it, and the next one.

```
.mdc-tab__indicator {
    bottom: 0;
    left: 2px;
}
```

poscapturefix 23

## Color

Like the icon in [#2110](https://github.com/material-components/material-components-web/issues/2110), The color of the CSS Only indicator is wrong when using `.mdc-theme--dark`.

Again, probably an oversight, but you are currently using the following for the light context, which handles both the tab indicator, and the tab bar indicator:
```
.mdc-tab-bar__indicator, 
.mdc-tab__indicator {
    background-color: rgba(0, 0, 0, 0.87);
    background-color: var(--mdc-theme-text-primary-on-light, rgba(0, 0, 0, 0.87));
}
```

However, for the dark context, you are only inverting the JS tab bar indicator:
```
.mdc-tab-bar--theme-dark .mdc-tab-bar__indicator, 
.mdc-theme--dark .mdc-tab-bar__indicator {
    background-color: white;
    background-color: var(--mdc-theme-text-primary-on-dark, white);
}
```

### Solution

Just add `.mdc-tab__indicator` appropriately to the SASS that compiles the above block. It is also worth noting here that the hover state of `.mdc-tab__indicator` is handled by a plain `opacity` rule, so this fix will also apply for the hover state.

Employing both fixes:

tabindcolorfix

moog16:
fixes #2117


packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -95,10 +95,11 @@ $mdc-tab-with-icon-and-text-height: 72px;
 }
 
 .mdc-tab__indicator {
+  $mdc-tab-indicator-horizontal-padding: 2px;
```
moog16: Not sure if this is confusing. I could just inline these values.

williamernest: Do we really need this as a variable? If so, I think it should go next to the others on line 26.

moog16: Na we don't...I thought it might be clearer, but its not complicated math. I'll remove it

PR #2144, opened 2018-01-30
chore(infrastructure): Add stylelint config for screenshot tests

acdvorak:
This PR:
- Adds `test/**/*.scss` to `npm run fix:css` and `npm run lint:css`
- Requires all Sass symbols (variables/functions/mixins) to begin with `test-` or `custom-` rather than `mdc-`
- Moves `ignoreFiles` from `.stylelintrc.yaml` to `.stylelintignore`

    According to the [stylelint configuration docs](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#ignorefiles):
    > _Note that `ignoreFiles` is not an efficient method for ignoring lots of files._ If you want to ignore a lot of files efficiently, use `.stylelintignore` or adjust your files globs.
- Ignores all directories named `node_modules` or `out`, regardless of where they are
- Extends the root `.stylelintrc.yaml` file


test/screenshot/.stylelintrc.yaml
```diff
@@ -0,0 +1,17 @@
...
+  custom-property-pattern: ^(?:test|custom)-.+
+  selector-class-pattern:
+    - ^(?:test|custom)-.+
+    - resolveNestedSelectors: true
+  selector-id-pattern: ^(?:test|custom)-.+
+
+  # Sass
+  scss/dollar-variable-pattern:
+    - ^(?:test|custom)-.+
+    -
```
patrickrodee: Is this an extra line?

.stylelintignore
```diff
@@ -0,0 +1,3 @@
+**/node_modules
```
patrickrodee: Do we need another .styelintignore inside the `test/screenshot` directory or does this one handle all?

patrickrodee: Polite ping on this @advorak

test/screenshot/.stylelintrc.yaml
acdvorak: Removed!

.stylelintignore
acdvorak: This root file handles all of them. IIRC `stylelint` doesn't support merging multiple `.stylelintignore` files.

PR #2148, opened 2018-01-31
chore(checkbox): reduced specificity of scss

moog16:
related: #1687

example: 

WAS: 
`.mdc-checkbox .mdc-checkbox__native-control:disabled:not(:checked):not(:indeterminate) ~ .mdc-checkbox__background`

NOW:
`.mdc-checkbox__native-control:disabled:not(:checked):not(:indeterminate) ~ .mdc-checkbox__background`


packages/mdc-checkbox/_mixins.scss
```diff
@@ -102,7 +102,9 @@
   white-space: nowrap;
   cursor: pointer;
   vertical-align: bottom;
+}
 
+@mixin mdc-checkbox-background-selector-disabled_ {
```
kfranqueiro: This doesn't seem like the right name for this mixin. Maybe dropping `selector-` is enough, but cc @advorak for his opinion.

Mixins with "selector" in the name are general-purpose mixins used to embed `@content` within a particular selector, which is not what this does.

moog16: I see...I was just copying the names of the two mixins included in this one. How about `mdc-checkbox-disabled-background-base-fill-color`?

acdvorak: `mdc-checkbox-disabled-color_($color)` seems a little closer to what we want.

Then, in `mdc-checkbox.scss`, you can pass `$mdc-checkbox-disabled-color` as the parameter value.

acdvorak: After talking to @kfranqueiro, you don't need a `$color` param - you can leave it as-is and just rename the mixin to `mdc-checkbox-disabled-container-color_`.

Issue #1843, opened 2018-01-01
Style standard Select to match standard Text Field (move variant into --box)


djensen47:
Text field boxes are optional for text fields. According to the spec, the mdc-select component is a form of text field:  "Other components such as dropdown menus and pickers can be nested in text fields to make input easier."  https://material.io/guidelines/components/text-fields.html#text-fields-input-types

**Observed behavior:**
The current documented behavior for mdc-select is that it is always inside of a text field box.

**Expected behavior:**
A text field box should be optional for mdc-select

---
Please forgive me if this issue has already been raised. I searched through the issue but didn't find anything relevant but there still might be.  Also, I didn't read through the code to see if there is an undocumented feature.

MeyerUX:
Agree with @djensen47. Not having the box as optional is a killer. Expected outcome according to the Material guidelines is to have the box option to match regular text field boxes, and a non-box option to match regular text fields.

PR #2149, opened 2018-01-31
feat(select): Add non box version 

williamernest:
Moves the default implementation to the `mdc-select--box` class, and styles the default `mdc-select` class to match the default text-field style. 
fixes: #1843


demos/select.html
```diff
@@ -93,56 +93,98 @@
...
 
-      <section class="example">
-        <h2 class="mdc-typography--title">Fully-Featured Component</h2>
-        <section id="box-demo-wrapper">
-          <div id="js-select-box" class="mdc-select" role="listbox">
-            <div class="mdc-select__surface" tabindex="0">
-              <div class="mdc-select__label">Food Group</div>
-              <div class="mdc-select__selected-text"></div>
-              <div class="mdc-select__bottom-line"></div>
+        <section class="example">
```
bonniezhou: Every block here has an extra indent (I think you need to set your tabs to be 2 spaces instead of 4.)

robcaldecott:
Very excited about this. Will it make it into 0.30.0?

williamernest:
@robcaldecott It should make it in 0.30.0 when we release next week.

PR #2150, opened 2018-01-31
fix(list): Add missing import

patrickrodee:


patrickrodee:
Fixes missing import for this call

https://github.com/material-components/material-components-web/blob/a2c1bd01cc99e6f1755368af94dd7ae226f0aa65/packages/mdc-list/mdc-list.scss#L172-L177

PR #2152, opened 2018-01-31
chore(grid-list): lessen specificity of scss

moog16:
related: #1687


kfranqueiro:
I think we want to investigate further into what caused this regression, to figure out for sure the best way to structure the affected styles.

moog16:
I found that the margin in the `.mdc-grid-list__tiles` selector was being overwritten by a margin: 0. This was never applied before my change, so I just removed it.

packages/mdc-grid-list/mdc-grid-list.scss
```diff
@@ -48,15 +48,14 @@ $mdc-grid-list-tile-secondary-icon-size: 24px;
 }
 
 // postcss-bem-linter: define grid-list
-.mdc-grid-list {
-  @include mdc-grid-list-tile-aspect(1);
+@at-root {
   @include mdc-grid-list-tile-gutter(4px);
+  @include mdc-grid-list-tile-aspect(1);
```
kfranqueiro: Nit: I don't think this line needs to be reordered? So the diff in this section only has to include changing `.mdc-grid-list` to `@at-root`

moog16: Yup I will reorder

PR #2155, opened 2018-02-01
chore(switch): reduce sass specificity

moog16:
related: #1687


packages/mdc-switch/mdc-switch.scss
```diff
@@ -45,79 +38,86 @@
...
+    transition:
+      mdc-switch-transition(transform),
+      mdc-switch-transition(background-color);
+    border-radius: $mdc-switch-focus-ring-diameter / 2;
+    opacity: .2;
+    content: "";
   }
 }
 
+@at-root {
```
acdvorak: Let's move this block immediately after the `.mdc-switch` rule above.

(We usually try to keep all the theme-related stuff as close to the top of the file as possible, to make it easier for readers to figure out how theming works.)

packages/mdc-switch/mdc-switch.scss
```diff
@@ -45,79 +38,86 @@
...
-      bottom: 0;
-      left: 0;
-      transition:
-        mdc-switch-transition(opacity),
-        mdc-switch-transition(background-color);
-      border-radius: 7px;
-      opacity: .38;
-      content: "";
-    }
+// postcss-bem-linter: ignore
```
acdvorak: Is this still needed?

packages/mdc-switch/mdc-switch.scss
```diff
@@ -45,79 +38,86 @@
...
+  height: $mdc-switch-knob-diameter;
+  transform: translateX(0);
+  transition:
+    mdc-switch-transition(transform),
+    mdc-switch-transition(background-color);
+  border-radius: $mdc-switch-knob-diameter / 2;
+  z-index: 1;
+
+  // Focus indicator
+  // postcss-bem-linter: ignore
```
acdvorak: Is this still needed?

packages/mdc-switch/mdc-switch.scss
```diff
@@ -45,79 +38,86 @@
...
-      width: $mdc-switch-focus-ring-diameter;
-      height: $mdc-switch-focus-ring-diameter;
-      transform: scale(1);
-      transition:
-        mdc-switch-transition(transform),
-        mdc-switch-transition(background-color);
-      border-radius: $mdc-switch-focus-ring-diameter / 2;
-    }
+  // Focus indicator
+  .mdc-switch__knob::before {
```
acdvorak: AFAICT, most of the properties below are redundant. They're already being set in
`.mdc-switch__knob::before` (or `mdc-switch-tap-target_`) above.

Keep `transform` and remove the rest (but make sure it doesn't break anything 😄 )

packages/mdc-switch/mdc-switch.scss
```diff
@@ -45,79 +38,86 @@
...
     }
   }
+}
+
+.mdc-switch__background {
+  display: block;
+  position: relative;
+  width: $mdc-switch-track-width;
+  height: $mdc-switch-track-height;
+  border-radius: 7px;
```
acdvorak: Once this is merged, can you create a follow up PR that changes _**all**_ the `border-radius: Npx` properties in this file to `border-radius: 50%`? AFAIK there's no need for specific `px` values.

packages/mdc-switch/mdc-switch.scss
moog16: Good catch...I didn't pay attention to the body of this rule. I think all I did was change the indentation.

I also noticed a few more duplications which I'm removing.

packages/mdc-switch/mdc-switch.scss
moog16: nope - removed!

PR #2158, opened 2018-02-01
chore(slider): reduce scss specificity

moog16:
related: #1687


kfranqueiro:
Confirmed, this breaks disabled styles.

What it should look like:

image

What it looks like in this PR:

image

Closing this, since the only change this makes causes a regression.

PR #2160, opened 2018-02-01
fix(ripple): Only deduplicate ripple on parents whose children activated

kfranqueiro:
This amends #2123 to check for parent/child relationship before suppressing multiple ripples on the same frame, which should allow for rare cases where it makes sense for ripples to happen on multiple unrelated surfaces at the same time (e.g. multitouch, and also helps for screenshot testing).

Checked timing on all browsers this JS loads for (non-IE/Edge) and it takes less than 1ms (generally _much_ less in Chrome/Safari).

BREAKING CHANGE: Adds `containsEventTarget(target)` API to the ripple adapter.


kfranqueiro:
Discussed with Patrick offline; the reason this tracks across instances is in order to avoid 2 ripples happening from the same single user interaction due to both existing in the same DOM hierarchy, so we need to track this across instances. This amends the original fix which simply ignored subsequent calls on the same frame, which was too restrictive.

PR #2162, opened 2018-02-01
chore(tabs): fix css output and reduce scss specificity

### LINKED ISSUE #1866: Tabs: Fix usage of ripple/states mixins
I noticed yesterday that mdc-tabs is using the ripple/states mixins in an odd way: https://github.com/material-components/material-components-web/blob/fc57923923b72d6d48a0faa37691a62adfbf9147/packages/mdc-tabs/tab/mdc-tab.scss#L161

This is weird, since the ripple/states mixins themselves are responsible for selecting on `.mdc-ripple-upgraded`. This sort of has the side effect of having no effect on non-upgraded ripples, but it also results in spurious selectors in some places.

We used to do something like this in mdc-list as well, which was [resolved](https://github.com/material-components/material-components-web/commit/c8772ea5f32ef971f4ae049379dc114fa2f0ec84#diff-d5738923465a810c2132e773f88f3d23) in #1737.

Moreover, I'm unsure why we're treating tabs differently between JS and CSS-only. Right now we're actually re-coloring the browser's focus outline for CSS-only tabs rather than using states, which seems like an odd thing to do (and this is the only component we style outline color in).

image

I'm planning to show this current behavior to our states designer today to confirm whether we should be using states even for the case of CSS-only tabs. If that's indeed the right thing to do, hopefully it means we can simplify tabs' styles and remove some.

moog16:
related: #1687
fixes #1866

Removed unused scss imports and condensed .mdc-tab rules together in scss. By removing `.mdc-ripple-upgraded` from `.mdc-tab.mdc-ripple-upgraded` selector, we fix the css output from creating selectors like `.mdc-tab.mdc-ripple-upgraded.mdc-ripple-upgraded` and `.mdc-tab.mdc-ripple-upgraded:not('.mdc-ripple-upgraded')`.

Also as #1866 references, we should be using mdc-ripple's hover state in favor of having an outline.


packages/mdc-tabs/tab/mdc-tab.scss
```diff
@@ -101,11 +101,7 @@ $mdc-tab-with-icon-and-text-height: 72px;
   width: calc(100% - 4px);
 }
 
-.mdc-tab-bar:not(.mdc-tab-bar-upgraded) {
-  .mdc-tab {
-    position: relative;
-  }
-
+:not(.mdc-tab-bar-upgraded) {
```
moog16: Looking at this..might need to revert this change.

moog16: adding back .mdc-tab-bar

PR #2167, opened 2018-02-02
docs(toolbar): Avoid demo page jumping when clicking toolbar icons

kfranqueiro:
This adds ready.js and common.js to the toolbar demo pages so that they get # link canceling behavior, and fixes an issue with ready.js so it also works for demos that only need styles and no JS.

### DIFF
--- a/demos/ready.js
+++ b/demos/ready.js
@@ -37,7 +37,8 @@ window.demoReady = (function(root) {
       return true;
     }
     ensureDetectionDom();
-    isReadyCached = Boolean(window.mdc) && getComputedStyle(testDom).position === 'relative';
+    isReadyCached = getComputedStyle(testDom).position === 'relative' &&
+      (Boolean(window.mdc) || !root.querySelector('script[src$="material-components-web.js"]'));
     return isReadyCached;
   }
 
--- a/demos/ready.js
+++ b/demos/ready.js
@@ -38,7 +38,7 @@ window.demoReady = (function(root) {
     }
     ensureDetectionDom();
     isReadyCached = getComputedStyle(testDom).position === 'relative' &&
-      (Boolean(window.mdc) || !root.querySelector('script[src$="material-components-web.js"]'));
+      (Boolean(window.mdc) || !root.querySelector('script[src*="material-components-web.js"]'));
     return isReadyCached;
   }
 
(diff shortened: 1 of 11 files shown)


demos/ready.js
```diff
@@ -37,7 +37,8 @@ window.demoReady = (function(root) {
       return true;
     }
     ensureDetectionDom();
-    isReadyCached = Boolean(window.mdc) && getComputedStyle(testDom).position === 'relative';
+    isReadyCached = getComputedStyle(testDom).position === 'relative' &&
+      (Boolean(window.mdc) || !root.querySelector('script[src$="material-components-web.js"]'));
```
acdvorak: It might be more resilient if we use a `[src*="value"]` selector instead of `[src$="value"]`.

That way, if we ever need to add a query string to the end of the URL for some reason (e.g., cache busting for statically-hosted PR screenshot tests), this won't suddenly break (which might be hard to debug, especially for anyone who joins the team after this code was added).

demos/toolbar/default-flexible-toolbar.html
```diff
@@ -55,6 +55,7 @@
       <span>Flexible Expansion Ratio: <span id="ratio">0</span></span>
     </footer>
 
+    <script src="/assets/common.js" async></script>
```
acdvorak: Nit: Can we move this line below `material-components-web.js`? Ditto for other files.

Both for consistency with other demo pages, and because it feels more readable (IMO) to keep the order of imports as:

1. MDC Web library file(s)
2. Demo-specific files

demos/toolbar/default-toolbar.html
```diff
@@ -50,5 +51,6 @@
         Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam egestas semper. Aenean ultricies mi vitae est. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Vestibulum tortor quam, feugiat vitae, ultricies eget, tempor sit amet, ante. Donec eu libero sit amet quam egestas semper. Aenean ultricies mi vitae est.
       </p>
     </main>
+    <script src="/assets/common.js" async></script>
```
acdvorak: Nit: Can we add a blank line before `<script>`? Ditto for other files.

I find it more readable to have the scripts separated from the visible elements with a little whitespace.

demos/toolbar/default-flexible-toolbar.html
kfranqueiro: Oh, thanks for pointing that out. Somehow didn't realize that was the order we were doing it elsewhere.

PR #2168, opened 2018-02-02
fix(menu): Rename test files

williamernest:


kfranqueiro:
This is presumably conflicting now due to merging #1756. Not sure if it's simpler to just re-do the rename from master now.

PR #2174, opened 2018-02-02
fix(demos): Too much space between toolbar title and left icon.

touficbatache:
There's too much space between the toolbar title and left icon. This PR fixes it.

demo


touficbatache:
I didn't remove it @williamernest take a look at the code.

demos/common.scss
```diff
@@ -79,7 +79,6 @@ fieldset {
 
 .catalog-logo,
 .catalog-back {
-  padding-right: 24px;
```
williamernest: We should keep the `padding-right` property for the catalog-logo class.

demos/button.html
```diff
@@ -74,9 +74,7 @@
     <header class="mdc-toolbar mdc-toolbar--fixed">
       <div class="mdc-toolbar__row">
         <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
-          <span class="catalog-back">
```
williamernest: We shouldn't be removing this span to change the `padding-right`.

touficbatache: Why not removing the `span`? I kept the class :smiley:

demos/common.scss
touficbatache: I don't think we should... There is too much space between the logo and the title...

touficbatache: Ah wait, there is already a `padding-right:24px;` applied to the remaining class

williamernest: Yup, not needed anymore.

touficbatache: That's why I removed it :smiley:

demos/button.html
williamernest: I guess it's not needed anymore :)

demos/common.scss
williamernest: Yup, no longer needed.

PR #2177, opened 2018-02-02
feat(chips): Add Sass mixins for customization

### LINKED ISSUE #2004: Sass mixins to customize mdc-chip
- [ ] `mdc-chip-border-radius`
- [ ] `mdc-chip-fill-color`
- [ ] `mdc-chip-ink-color`
- [ ] `mdc-chip-stroke-color`

bonniezhou:
Resolves #2004


packages/mdc-chips/_mixins.scss
```diff
@@ -0,0 +1,33 @@
...
+
+@mixin mdc-chip-fill-color($color: $mdc-chip-fill-color-default) {
+  background: $color;
+}
+
+@mixin mdc-chip-ink-color($color: $mdc-chip-ink-color-default) {
+  color: $color;
+}
+
+@mixin mdc-chip-stroke($width: 0, $style: solid, $color: transparent) {
```
kfranqueiro: Are chips really expected to have anything besides solid stroke?

In mdc-button we exposed separate mixins for stroke width and color only. (The button mixin for stroke-width also attempts to counteract the border width increasing the size of the overall component, though I'm not sure whether we want to do that for chips.)

packages/mdc-chips/_mixins.scss
```diff
@@ -0,0 +1,33 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "./variables";
+
+@mixin mdc-chip-border-radius($radius: $mdc-chip-border-radius-default) {
```
kfranqueiro: It seems odd to me for the parameters to these public mixins to have default values (it's more arguable for multi-parameter mixins, but see comment below RE stroke). If someone is calling one of these mixins, they clearly have a value in mind to pass to it.

I'm also wondering if we want to provide something similar to mdc-button-filled-accessible for chips? (At least, if chips are typically expected to have black or white ink in many cases.)

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -16,16 +16,17 @@
...
 @import "@material/ripple/mixins";
+@import "../mixins";
 
 .mdc-chip {
   @include mdc-ripple-surface;
   @include mdc-ripple-radius-bounded;
   @include mdc-states;
+  @include mdc-chip-border-radius;
+  @include mdc-chip-fill-color;
+  @include mdc-chip-ink-color;
```
kfranqueiro: We should update the invocation of `mdc-states` to pass the chip ink color to it for consistency.

bonniezhou: I changed `mdc-chip-ink-color` to automatically update the ripple color (the client can always override it).

packages/mdc-chips/_mixins.scss
bonniezhou: Makes sense, removed the default values and made them baseline arguments provided by `mdc-chip`.

The accessibility thing is a great idea! I added a `mdc-chip-accessible-color` mixin.

packages/mdc-chips/README.md
```diff
@@ -60,6 +60,21 @@ CSS Class | Description
...
 `mdc-chip-set` | Mandatory. Indicates the set that the chip belongs to.
 
+### Sass Mixins
+
+To customize the colors of any part of the chip, use the following mixins. 
+
+Mixin | Description
+--- | ---
+`mdc-chip-corner-radius($radius)` | Customizes the corner radius for a chip
+`mdc-chip-accessible-color($fill-color)` | Customizes the background fill color for a chip, and updates the chip's ink and ripple color to meet accessibility standards
```
kfranqueiro: Rename to `mdc-chip-fill-color-accessible` for consistency with other similar mixins in other packages (e.g. slider and toolbar). Then you can rename the parameter to just `$color` like the others here.

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -16,17 +16,19 @@
...
+@import "../mixins";
+@import "../variables";
 
 .mdc-chip {
   @include mdc-ripple-surface;
   @include mdc-ripple-radius-bounded;
   @include mdc-states;
+  @include mdc-chip-corner-radius($mdc-chip-border-radius-default);
+  @include mdc-chip-fill-color($mdc-chip-fill-color-default);
+  @include mdc-chip-ink-color($mdc-chip-ink-color-default);
```
kfranqueiro: Remove the `mdc-states` include 3 lines above this, since now it's handled within the `ink-color` mixin.

packages/mdc-chips/_mixins.scss
kfranqueiro: That seems like probably a good idea, especially if the default is no stroke whatsoever, but then should the default width be 1 rather than 0?

I would think we shouldn't include styles from one mixin in the other because that would lead to redundant styles being output (e.g. you'd end up with `border-style` both from the -style mixin and from the -width mixin).

bonniezhou:
Added custom-themed chips to the catalog page!

packages/mdc-chips/_mixins.scss
```diff
@@ -0,0 +1,70 @@
...
+@mixin mdc-chip-fill-color($color) {
+  @include mdc-theme-prop(background-color, $color);
+}
+
+@mixin mdc-chip-ink-color($color) {
+  @include mdc-states($color);
+  @include mdc-theme-prop(color, $color);
+}
+
+@mixin mdc-chip-stroke($width: 0, $style: solid, $color: black) {
```
kfranqueiro: Default `$width` to 1? Presumably if they're calling this, they want the stroke to be visible.

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -16,17 +16,18 @@
...
 
   display: inline-block;
-  border-radius: 16px;
-  background: rgba(0, 0, 0, 0.08);
-  color: rgba(0, 0, 0, 0.87);
-  padding: 7px 12px;
+  padding: $mdc-chip-vertical-padding $mdc-chip-horizontal-padding;
```
kfranqueiro: `padding` should come after `margin` according to our stylelint rules. (I'm not sure why `npm run lint` isn't catching this, but SublimeLinter did...)

PR #2179, opened 2018-02-03
fix(text-field): move script tags below mdc.js tag

moog16:
~This PR is to allow text-field demo page to render properly. This is only a temporary fix as the real fix is to update demoReady with a listener for DOMContentLoaded.~

#2180 Fixed the demo pages. Purpose of this is to just condense all the <script> together into one.


kfranqueiro:
#2180 fixes `demoReady` FYI

Issue #1994, opened 2018-01-18
Extract webpack.demo.js from webpack.config.js


lynnjepsen:
This is a prerequisite to our new screenshot testing directory

PR #2181, opened 2018-02-03
chore: Add tests for webpack.config.js module exports

acdvorak:
- Adds an `npm run test:buildconfig` command, and calls it in `npm test`
- Replaces `;` with `&&` in `package.json` (Windows `cmd.exe` does not support `;`)

To test this PR, simply change a single character in a `test/build/goldens/*.json` file.
Then run `npm run test:buildconfig` in the console.

This PR is a prerequisite for #1994.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,12 +19,13 @@
     "lint": "npm-run-all --parallel lint:*",
     "postinstall": "lerna bootstrap",
     "pretest": "npm run lint && node scripts/check-imports.js",
-    "test": "npm run test:unit && npm run test:closure && npm run test:dependency && npm run build; npm run clean",
+    "test": "npm run test:buildconfig && npm run test:unit && npm run test:closure && npm run test:dependency && npm run build && npm run clean",
     "posttest": "istanbul report --root coverage text-summary && istanbul check-coverage --lines 95 --statements 95 --branches 95 --functions 95",
     "test:watch": "karma start --auto-watch",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -19,12 +19,13 @@
     "lint": "npm-run-all --parallel lint:*",
     "postinstall": "lerna bootstrap",
     "pretest": "npm run lint && node scripts/check-imports.js",
-    "test": "npm run test:unit && npm run test:closure && npm run test:dependency && npm run build; npm run clean",
+    "test": "npm run test:buildconfig && npm run test:unit && npm run test:closure && npm run test:dependency && npm run build && npm run clean",
```
lynnjepsen: Move this to the end of the tests. Most people just run Unit tests and then force quit because they don't want to wait (at least thats what happens to me)

acdvorak: `test:buildconfig` takes less than 1 second to run.

I originally had it at the end, but I got tired of waiting for `test:unit` to finish :grinning:

kfranqueiro: People shouldn't be force-quitting after running unit tests though? At least, not if they want to make sure that everything actually works... (And if they really only want to run unit tests... `npm run test:unit`)

There might actually be solid argument for putting the fast-running things (e.g. `test:buildconfig` and `build`) first since if we manage to do something silly that fails fast, it'd be better for Travis CI to get it over with quick and move on to other jobs. We can do that separately from this PR though.

PR #2183, opened 2018-02-05
fix(snackbar): Doesn't close while other element is focused

### LINKED ISSUE #1918: Snackbar keyboard issue

0.28.0

Chrome/63.0.3239.132 (also others)

### What OS are you using?

Mac OS X 10_13_2

If I Operate the radio button with the keyboard, snackbar will now work properly.

1. Go to https://jsfiddle.net/yuheiy/d1e0j490/
2. Operate the radio button with the keyboard
3. Snackbar will not be hidden

piotr-cz:
Snackbar doesn't close when any element on page receives focus event while snackbar is opened and snackbar has empty action text.

Fixes #1918 

Steps to reproduce:

1) Go to https://material-components-web.appspot.com/snackbar.html
2) Delete Action Text value
3) Click on Show
4) Press Shift-Tab to focus some element on the page (like Action Text input element)

Snackbar should close, but it doesn't.


piotr-cz:
One test failed (*TEST_SUITE=unit*): but it looks like tests configuration problem: `The SUID sandbox helper binary was found, but is not configured correctly`

PR #2185, opened 2018-02-05
docs(line-ripple): Fix typos and add hyphen to improve readability

kfranqueiro:
Also fixes inconsistencies between things ending with a period or not in tables.

### DIFF
--- a/packages/mdc-line-ripple/README.md
+++ b/packages/mdc-line-ripple/README.md
@@ -2,13 +2,13 @@
 title: "Line Ripple"
 layout: detail
 section: components
-excerpt: "The line ripple is used to highight user specified text above it."
+excerpt: "The line ripple is used to highlight user-specified text above it."
 path: /catalog/input-controls/line-ripple/
 -->
 
 # Line Ripple
 
-The line ripple is used to highlight user specified text above it. When a the line ripple is active, the line’s color and thickness vary.
+The line ripple is used to highlight user-specified text above it. When a line ripple is active, the line’s color and thickness vary.
 
 ## Design & API Documentation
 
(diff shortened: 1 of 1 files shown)


packages/mdc-line-ripple/README.md
```diff
@@ -42,15 +42,15 @@ CSS Class | Description
...
+`mdc-line-ripple--deactivating` | Styles the line ripple as a deactivating line ripple
 
 ### Sass Mixins
 
 ### `MDCLineRipple`
 
-`activate() => void` | Proxies to the foundations `activate()` method.
-`deactivate() => void` |  Proxies to the foundations `deactivate()` method.
-`setRippleCenter(number: xCoordinate) => void` | Proxies to the foundations `setRippleCenter(number: xCoordinate)` method.
+`activate() => void` | Proxies to the foundation's `activate()` method
```
moog16: Is this supposed to be a table? I think you need the `--- | ---` above

kfranqueiro: You, sir, are very correct, and wowzers this sure looks very broken right now. Good catch.

packages/mdc-line-ripple/README.md
```diff
@@ -42,15 +42,15 @@ CSS Class | Description
 --- | ---
 `mdc-line-ripple` | Mandatory
 `mdc-line-ripple--active` | Styles the line ripple as an active line ripple
-`mdc-line-ripple--deactivating` | Styles the line ripple as a deactivating line ripple. 
+`mdc-line-ripple--deactivating` | Styles the line ripple as a deactivating line ripple
 
 ### Sass Mixins
```
moog16: I think this is unnecessary

kfranqueiro: Hm. Actually it looks like there _is_ a public mixin but it was never documented. Guess I'll add that in here.

PR #2191, opened 2018-02-05
feat(chips): Handle leading/trailing icon styles

### LINKED ISSUE #2005: Add mdc-chip__icon to mdc-chip
- [ ] Handle leading icon cases
- [ ] Plan for future case of a check icon for the filter chip set
- [ ] Handle trailing icon cases
- [ ] Plan for future case of a close icon for the entry chip set

bonniezhou:
Resolves #2005.

Adjust styles (mostly padding) to handle icons in chips. This PR does not attach JS functionality to any icons yet.


packages/mdc-chips/README.md
```diff
@@ -52,13 +52,39 @@ npm install --save @material/chips
 </div>
 ```
 
+#### Leading and Trailing Icons
+
+You can optionally add a leading icon (i.e. thumbnail) and/or a trailing icon to a chip. To add an icon, add the relevant class (either `mdc-chip--with-leading-icon` or `mdc-chip--with-trailing-icon`) to the chip element, add an `i` element with your preferred icon, and give it a class of `mdc-chip__icon`. If you're adding a trailing icon, also give the icon a class of `mdc-chip__icon--trailing`.
```
kfranqueiro: The CSS styles below seem to suggest that trailing icons can be individually hovered/focused. If so, we might want to consider using a `button` element (or barring that, an `a` element which we'd cancel click events on) for trailing icon rather than an `i` element, if clicking the trailing icon has a distinct action from clicking the rest of the chip. Otherwise you're never actually going to trigger the `:focus` style below without also going out of your way to specify `tabindex`.

On the other hand, if clicking the trailing icon does the same thing as clicking the rest of the chip, I'm not sure whether the icon itself should be individually focusable at all? We should determine which element[s] should be focusable and consider updating our DOM accordingly. Right now, nothing on the chips demo page is focusable.

packages/mdc-chips/README.md
```diff
@@ -52,13 +52,39 @@ npm install --save @material/chips
...
 CSS Class | Description
 --- | ---
+`mdc-chip-set` | Mandatory. Indicates the set that the chip belongs to
 `mdc-chip` | Mandatory.
-`mdc-chip__text` | Mandatory. Indicates the text content of the chip.
-`mdc-chip-set` | Mandatory. Indicates the set that the chip belongs to.
+`mdc-chip--with-leading-icon` | Optional. Indicates the chip contains a leading icon
+`mdc-chip--with-trailing-icon` | Optional. Indicates the chip contains a trailing icon
+`mdc-chip__text` | Mandatory. Indicates the text content of the chip
+`mdc-chip__icon` | Optional. Indicates a leading or trailing icon in the chip
```
acdvorak: Can we mention that "leading" is the default unless `--trailing` is used?

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -33,9 +33,41 @@
...
+
+.mdc-chip__icon {
+  color: $mdc-chip-icon-color;
+  vertical-align: middle;
+  margin: 0 4px;
+  border-radius: 16px;
+  outline: none;
+}
+
+.mdc-chip__icon.mdc-chip__icon--trailing {
```
kfranqueiro: I don't think most of these rules need the double class selector here?

We might want to use 2 classes for font-size specifically though, given the snag Brendan ran into when he can't guarantee ordering of material-icons and MDC stylesheets.

packages/mdc-chips/_variables.scss
```diff
@@ -19,3 +19,9 @@ $mdc-chip-fill-color-default: rgba(black, .08);
 $mdc-chip-ink-color-default: rgba(black, .87);
 $mdc-chip-horizontal-padding: 12px;
 $mdc-chip-vertical-padding: 7px;
+$mdc-chip-with-icon-padding: 4px;
+
+$mdc-chip-icon-color: rgba(0, 0, 0, .54);
+$mdc-chip-trailing-icon-size: 18px;
+$mdc-chip-trailing-icon-hover: rgba(0, 0, 0, .62);
+$mdc-chip-trailing-icon-focus:rgba(0, 0, 0, .87);
```
acdvorak: Nit: Add a space after `:`

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -33,9 +33,41 @@
   overflow: hidden;
 }
 
+.mdc-chip--with-leading-icon {
```
kfranqueiro: Is it not possible to accomplish this via padding or margin on the icons themselves, rather than needing to add one or two classes to every chip in addition to adding the icon?

packages/mdc-chips/_variables.scss
kfranqueiro: Missing space after `:` here. Also, add `-color` to the hover/focus variable names?

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -33,9 +33,41 @@
   overflow: hidden;
 }
```
acdvorak: I would suggest using `display: inline-flex` on `.mdc-chip` to collapse whitespace in the markup. This will give you more control over spacing.

Then, instead of `vertical-align: middle` on the child elements, you can just set `align-items: center` on `.mdc-chip`.

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -33,9 +33,41 @@
...
   vertical-align: middle;
   // Set line-height to be <18px to force the content height of mdc-chip to be 18px.
   line-height: 17px;
 }
+
+.mdc-chip__icon {
+  color: $mdc-chip-icon-color;
+  vertical-align: middle;
+  margin: 0 4px;
+  border-radius: 16px;
```
acdvorak: `border-radius: 50%` should do the trick, and it avoids hard-coding a specific pixel value 😀

packages/mdc-chips/chip/mdc-chip.scss
kfranqueiro: Based on a conversation I just had with Brendan, keep the width/height/font-size with the 2-class selector... the hover and focus are probably safe to just have `.mdc-chip__icon--trailing` though.

packages/mdc-chips/_variables.scss
```diff
@@ -19,3 +19,9 @@ $mdc-chip-fill-color-default: rgba(black, .08);
 $mdc-chip-ink-color-default: rgba(black, .87);
 $mdc-chip-horizontal-padding: 12px;
 $mdc-chip-vertical-padding: 7px;
+$mdc-chip-with-icon-padding: 4px;
+
+$mdc-chip-icon-color: rgba(0, 0, 0, .54);
```
acdvorak: I normally prefer to use the global theme variables when possible (e.g., `text-primary-on-background`), because it makes it much easier for clients to create a dark theme.

In this case, though, some of the other values in this file don't map directly to theme vars (e.g., `.08`, `.62`), so it might not be feasible.

Just something to keep in mind in the future 🙂

demos/chips.html
```diff
@@ -54,18 +54,82 @@
       </section>
 
       <section class="example">
+        <h2>Entry Chips</h2>
         <div class="mdc-chip-set">
-          <div class="mdc-chip">
-            <div class="mdc-chip__text">Add to Calendar</div>
+          <div class="demo-chip mdc-chip mdc-chip--with-leading-icon mdc-chip--with-trailing-icon">
+            <i class="material-icons mdc-chip__icon" tabindex="0">face</i>
```
acdvorak: If we're going to have `tabindex="0"` here, we should also add `:focus` and `:hover` styles for leading icons.

demos/chips.html
```diff
@@ -54,18 +54,82 @@
...
+            <div class="mdc-chip__text">Jane Smith</div>
+            <i class="material-icons mdc-chip__icon mdc-chip__icon--trailing" tabindex="0">more_vert</i>
```
acdvorak: Add ARIA and title attributes: `role="button" title="More options"`

Ditto for other elements that have a `tabindex="0"`

packages/mdc-chips/chip/mdc-chip.scss
acdvorak: Can we simplify this selector to just `.mdc-chip__icon--trailing`?

packages/mdc-chips/README.md
kfranqueiro: Okay, I just looked at this again after seeing Andy's comments, and realized things on the demo page _are_ focusable... but it's almost imperceptible that things are getting focused unless you know exactly where you should be looking. I think we may have a design problem...

packages/mdc-chips/chip/mdc-chip.scss
acdvorak: Why do we need both classes? All of the chips CSS rules are in a single file, so CSS load order shouldn't be a problem. What am I missing?

kfranqueiro: material-icons is not in the same file, or even the same repository. See #2242.

packages/mdc-chips/README.md
kfranqueiro: cc @acdvorak on this since he had some reservations about my `button` element idea in particular (mainly related to needing to override more styles, and maybe a concern about issues with padding in some browsers?)

bonniezhou: Good spotting, I'll ask designers about the imperceptible focus issue. 

I think it's fine to specify `tabindex` for icons, that's what we do for text field too. Using a `i` element makes the most sense semantically and specifying `tabindex` is a small price to pay.

We also need to specify `tabindex` for each chip, which leads to the question of whether we should use a `button` element for a chip. But from a design/semantics standpoint we probably want to separate the concept of a button from the concept of a chip as much as possible and not confuse clients...

packages/mdc-chips/chip/mdc-chip.scss
bonniezhou: Ohh so negative margins are a thing. Will do!

demos/chips.html
bonniezhou: Removed tabindex since interacting with the leading icon should be the same as interacting with the chip.

packages/mdc-chips/README.md
bonniezhou: Updated it so either `--leading` or `--trailing` is mandatory, which also removes the need for `mdc-chip--with-leading-icon` and `mdc-chip--with-trailing-icon`.

packages/mdc-chips/_variables.scss
bonniezhou: Replaced `mdc-chip-ink-color-default` with `text-primary-on-light`! I kept the icon colors though since the global variables are supposed to be for text.

packages/mdc-chips/chip/mdc-chip.scss
bonniezhou: Done, thanks for the tip!

demos/chips.html
bonniezhou: Should chips themselves also have `role="button"`? I guess this is related to [Ken's comment](https://github.com/material-components/material-components-web/pull/2191/files/9a3b9467d3ba4e4ec9e28713a177079566f3af16#r167957686) above.

packages/mdc-chips/README.md
acdvorak: I agree, chips are not buttons.

If anything, they're closer to checkboxes or radio buttons, but we can't really use those as the base element, so `<div tabindex="0">` or `<span tabindex="0">` seem reasonable to me.

Semantically, the "X" icon is a button, but it seems weird to use a `<button>` element for that, so I would also vote to stick with `<i tabindex="0" role="button">`.

kfranqueiro: It's true that we do have other components doing the same thing, and it's true that using the `i` element for icons is pretty pervasive nowadays, so yeah, I guess this discussion can be tabled.

The upside of a more semantically-correct element is not only getting tabindex for free but also potentially ARIA role (and technically we should probably have both?). However, it might also be the case that different roles make sense for different types of chips/icons, which would also discourage settling on a specific more semantic element. (I know Andy brought this up on the choice chips PR.)

FWIW the `i` element itself is arguably not semantic at all; it originally indicated italicized text back in the 90s, and was later discouraged in lieu of other tags like `em` that _do_ have semantic meaning.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i):

> The `<i>` represents a range of text that is set off from the normal text for some reason, for example, technical terms, foreign language phrases, or fictional character thoughts. It is typically displayed in italic type.

> Use this element only when there is not a more appropriate semantic element.

packages/mdc-chips/chip/mdc-chip.scss
acdvorak: Gotcha. Can we add a comment above each 2-class-selector explaining why we need the extra class?

demos/chips.html
kfranqueiro: I suspect that once all the variants of chips are in, we should document guidance on what roles make sense for what types of chips.

demos/chips.scss
```diff
@@ -17,6 +17,10 @@
 @import "./common";
 @import "../packages/mdc-chips/mdc-chips";
 
+.demo-chip {
```
kfranqueiro: I am wondering whether we should be building in at least the left/right margins if not also even top/bottom, since chips are intended to be in sets and I assume we have an opinion on how chips within a set are spaced apart? (Even if we make that style specific to chip-set...)

demos/chips.html
```diff
@@ -38,65 +38,129 @@
     <main class="mdc-toolbar-fixed-adjust">
       <section class="hero">
         <div class="mdc-chip-set">
-          <div class="mdc-chip">
+          <div class="demo-chip mdc-chip" tabindex="0">
```
kfranqueiro: We've given the trailing icons `role`, but what about chips themselves? (I forget if that's being handled in #2215)

packages/mdc-chips/README.md
```diff
@@ -40,25 +40,52 @@ npm install --save @material/chips
...
+  <div class="mdc-chip__text">Add to calendar</div>
+</div>
+```
+
+##### Trailing icon
+
+```html
+<div class="mdc-chip mdc-chip--with-trailing-icon">
+  <div class="mdc-chip__text">Jane Smith</div>
+  <i class="material-icons mdc-chip__icon mdc-chip__icon--trailing" tabindex="0">cancel</i>
```
kfranqueiro: Add `role="button"` to the example

packages/mdc-chips/README.md
```diff
@@ -40,25 +40,52 @@ npm install --save @material/chips
...
 ```
 
+#### Leading and Trailing Icons
+
+You can optionally add a leading icon (i.e. thumbnail) and/or a trailing icon to a chip. To add an icon, add an `i` element with your preferred icon, give it a class of `mdc-chip__icon`, and a class of either `mdc-chip__icon--leading` or `mdc-chip__icon--trailing`. If you're adding a trailing icon, also set `tabindex="0"` and `role="button"` to make it accessible by keyboard and screenreader.
+
+##### Leading icon
+
+```html
+<div class="mdc-chip mdc-chip--with-leading-icon">
```
kfranqueiro: Remove `--with-leading-icon` here and `--with-trailing-icon` below since they were re-engineered out of existence

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -24,18 +24,45 @@
...
+  outline: none;
+}
+
+.mdc-chip__icon--leading {
+  margin: -4px 4px -4px -4px;
+}
+
+.mdc-chip__icon--trailing {
+  margin: 0 -4px 0 4px;
+  &:hover {
```
kfranqueiro: Newline before this block? (Kinda surprised lint doesn't complain?)

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -24,18 +24,45 @@
...
+    color: $mdc-chip-trailing-icon-hover-color;
+  }
+
+  &:focus {
+    color: $mdc-chip-trailing-icon-focus-color;
+  }
+}
+
+.mdc-chip__icon.mdc-chip__icon--trailing {
```
kfranqueiro: Add a comment indicating these are more specific to ensure they override icon styles in other libraries e.g. material-icons (like in https://github.com/material-components/material-components-web/pull/2242/files)

demos/chips.html
acdvorak: I think @bonniezhou decided to do most of the a11y stuff (the tricky parts, anyway) in a separate follow up PR, cuz shoehorning it into this one was getting kinda hairy.

demos/chips.scss
acdvorak: +1, let's give each chip a baseline margin.

**However**, we should clarify with design whether the margins should also push the chips Npx away from the bounding box of their parent container (i.e., the chip set).

E.g.:

image

vs.:

image

(the grey borders are only to show the boundaries of the `.mdc-chip-set` element)

It makes a difference when there's a heading, or input, or some other element nearby that needs to be horizontally aligned.

I also strongly suggest that we change `.mdc-chip-set` to use `display: flex; flex-wrap: wrap;` to collapse the whitespace in the markup, because it will give us much better control over spacing (so we can match the spec).

bonniezhou: Design confirmed that the margins should also push the chips away from the chip-set border.
I've hard-coded a baseline margin on the chips for now and filed #2265 for a separate PR.

demos/chips.html
bonniezhou: Yep, it's tracked in #2259. Also, design confirmed that we should add a `role` but looking into options other than `button` (maybe `menuitem` or something similar).

bonniezhou:
@kfranqueiro Confirmed with design that clicking the trailing icon should not trigger a ripple (filed #2266).

bonniezhou:
@kfranqueiro @acdvorak Ready for another look! To summarize, nits are fixed and I filed #2259, #2265 and #2266 to be done in smaller PRs.

packages/mdc-chips/README.md
```diff
@@ -59,7 +59,7 @@ You can optionally add a leading icon (i.e. thumbnail) and/or a trailing icon to
 ##### Leading icon
 
 ```html
-<div class="mdc-chip mdc-chip--with-leading-icon">
+<div class="mdc-chip mdc-chip">
```
acdvorak: I really like chips too, but 2 is probably too much here :wink:

packages/mdc-chips/README.md
```diff
@@ -68,9 +68,9 @@ You can optionally add a leading icon (i.e. thumbnail) and/or a trailing icon to
 ##### Trailing icon
 
 ```html
-<div class="mdc-chip mdc-chip--with-trailing-icon">
+<div class="mdc-chip mdc-chip">
```
acdvorak: Ditto

packages/mdc-chips/README.md
```diff
@@ -40,25 +40,52 @@ npm install --save @material/chips
...
+  <div class="mdc-chip__text">Jane Smith</div>
+  <i class="material-icons mdc-chip__icon mdc-chip__icon--trailing" tabindex="0" role="button">cancel</i>
+</div>
+```
+
 ### CSS Classes
 
 CSS Class | Description
 --- | ---
+`mdc-chip-set` | Mandatory. Indicates the set that the chip belongs to
```
acdvorak: Can we change `.mdc-chip-set` to use `display: flex; flex-wrap: wrap` to collapse the whitespace in the markup?

It will give us much better control over spacing, so we can match the spec exactly.

bonniezhou: I put it in #2265 since it requires the creation of a new file (mdc-chip-set/mdc-chip-set.scss)

PR #2192, opened 2018-02-05
chore: Refactor environment JS into scripts/build/

### LINKED ISSUE #2171: Extract environment JS into scripts/build
- [ ] environment.js

acdvorak:
Fixes #2171

### DIFF
--- a/scripts/build/environment.js
+++ b/scripts/build/environment.js
@@ -1,0 +1,38 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+module.exports = class {
+  setBabelEnv() {
+    const event = this.getNpmLifecycleEvent_();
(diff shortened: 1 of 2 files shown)


scripts/build/environment.js
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+module.exports = class {
+  setBabelEnv() {
+    const event = this.getNpmLifecycleEvent_();
+    if (event === 'test' || event === 'test:watch') {
```
kfranqueiro: Not knowing much about this, is there a reason this checks for exactly 2 things? We have a lot of `test:...` targets...

acdvorak: Great question, Ken! I have no idea :stuck_out_tongue:

(I literally just copied this from the existing `webpack.config.js` file.)

I spent a few minutes Googling around for `BABEL_ENV test`, and it seems like it doesn't have any "special" built-in meaning to Babel/Webpack.

However, we provide our own custom config for `test` in `package.json`:

<pre>
  "babel": {
    "presets": [
      [
        "es2015",
        {
          <b>"modules": false</b>
        }
      ]
    ],
    "plugins": [
      "transform-object-assign"
    ],
    "env": {
      <b>"test"</b>: {
        "presets": [
          "es2015"
        ]
      }
    }
  }
</pre>

The more I read this config block, the more confused I get.

From reading the [`babel-preset-es2015` docs](https://www.npmjs.com/package/babel-preset-es2015#modules), it looks like our config _disables_
ES6->CommonJS module transformation for all `npm run` targets _except_
`npm run test` and `npm run test:watch`.
WAT?!

And then, in our `webpack.config.js`, we have `"libraryTarget": "umd"`.
_So which is it??_

What's more, `npm run test` is an alias that expands to
`npm run test:unit && npm run test:closure && ...`, so if you run
`npm run test:unit` by itself, presumably you _won't_ get `BABEL_ENV=test`...
WAT?!

I'm inclined to leave the `if(...)` check as-is in this PR, because I don't understand it and I don't want to break anything.

PR #2193, opened 2018-02-06
chore: Fix ordering of assert params

acdvorak:
Otherwise, the test output is backwards and confusing. E.g., if you add a line, it will show that you removed it, and vice versa.

To test, add or remove a line in a `test/build/goldens/*.golden.json` file and run:

```bash
npm run test:buildconfig
```


acdvorak:
Yeah, this PR is wrong. It seemed to fix the output in a different PR, but clearly it doesn't work here.

PR #2197, opened 2018-02-06
fix(text-field): Apply error color on bottom line of fullwidth field

### LINKED ISSUE #2165: Textfield - Invalid fullwidth textfields don't have red bottom border
Hi,

The bottom border of invalid fullwidth textfields remains grey instead of being red like standard textfields.

You can see it here : https://codepen.io/jfily/pen/GQoKaV

kfranqueiro:
Also fixes typos in a couple of private mixins (asterix -> asterisk, full-width -> fullwidth for consistency with other mixins incl. a public one)

Observable using the required checkbox under the fullwidth examples at the end of text-field.html.

Fixes #2165.

### DIFF
--- a/packages/mdc-textfield/mdc-text-field.scss
+++ b/packages/mdc-textfield/mdc-text-field.scss
@@ -173,7 +173,7 @@
   @include mdc-text-field-dense_;
 }
 
-@include mdc-required-text-field-label-asterix_ {
+@include mdc-required-text-field-label-asterisk_ {
   margin-left: 1px;
   content: "*";
 }
@@ -183,7 +183,11 @@
 }
 
 .mdc-text-field--fullwidth {
-  @include mdc-text-field-full-width_;
+  @include mdc-text-field-fullwidth_;
+}
+
+.mdc-text-field--fullwidth.mdc-text-field--invalid {
+  @include mdc-text-field-fullwidth-invalid_;
 }
 
 // postcss-bem-linter: define text-field-helper-text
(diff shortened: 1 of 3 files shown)


packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -183,7 +183,11 @@
 }
 
 .mdc-text-field--fullwidth {
-  @include mdc-text-field-full-width_;
+  @include mdc-text-field-fullwidth_;
+}
+
+.mdc-text-field--fullwidth.mdc-text-field--invalid {
```
williamernest: When the alternate colors class is added, the idle line is still red.

kfranqueiro: FWIW I would point out that the alternate color class was never doing anything to the idle bottom line color for the fullwidth text field, so this might actually be a separate issue, but I'm looking into it a bit.

kfranqueiro: So this seems like it was more of an oversight with the demo page than with these styles. I added styles to the demo page for the full-width custom color demo, and also fixed a style for the textfield custom color demo that seemed accidentally redundant (idle/focus used same opacity).

I suspect that those styles could be further cleaned up; I initially ended up confused as to where to put the styles for this specific part of the demo page, and realized that the custom color and custom error color styles probably don't actually need separate CSS classes since they're always applied together to the same elements...

Issue #2029, opened 2018-01-22
Text-field: Setting required with Component causes field to immediately be invalid. 


williamernest:
## Bugs

Follow the template below to ensure the quickest and most accurate response to your issue.

### What MDC Web Version are you using?

> 0.28.0

### What browser(s) is this bug affecting?

> Version 63.0.3239.132 (Official Build) (64-bit)

### What OS are you using?

> Debian 4.9.65-3+deb9u2 (2018-01-04) x86_64 GNU/Linux
> MacOS 10.13.2

### What are the steps to reproduce the bug?

1. Set one component to "required" using the JS component. 
2. Set another to required by adding "required" to the html element. 

### What is the expected behavior?

Setting a text-field as "required" should not immediately render the field as invalid until after the blur event. 

### What is the actual behavior?

The element set as a required element using the JS component has the --invalid class applied immediately. 

### Any other information you believe would be useful?
The trailing icon elements in the gif are using the Component to set the field as required. The leading icon elements are set to required using the html attribute 'required'. 
![final](https://user-images.githubusercontent.com/34519388/35235986-96f41472-ff5a-11e7-950d-dcae5b4cbe0c.gif)

theenoahmason:
Hey I was just looking at this as I remembered this had been fixed a couple versions ago... As far as I can tell this is invalid, the text field does not perform its validation untill after blur, as expected. 

Please correct me if I'm wrong but i think this can be closed.

theenoahmason:
**correction:** This problem is only happening on textareas now.

Also incase anyones wondering, currently the latest package as determined by unpkg for example, is 0.28, but the demo for the textfields is currently at 0.29.

williamernest:
Here's a code pen showing the issue for the text field. Setting required via the JS or the HTML5 attribute should behave the same way. 

https://codepen.io/williamernest/pen/ddPdxo

theenoahmason:
Oh I misunderstood, ignore my comments.

moog16:
2 cases that I discovered while playing around with my current fix:

1. mdc-text-field has the invalid class, but required attr is removed. In this instance it seems like the the invalid class on the text field should be removed.
1. mdc-text-field is dirty and has pattern=".{8,}", but not required. Then the required is toggled to true. In this instance I would expect the text field should get the invalid class.

Neither of these two examples are covered in the HTML 5 version. I will speak to designer to see what the correct behavior is.

moog16:
Just spoke with design:

[...]

PR #2201, opened 2018-02-06
fix(text-field): disable validation check in setRequired

moog16:
resolves https://github.com/material-components/material-components-web/issues/2029


demos/text-field.html
```diff
@@ -139,7 +139,7 @@ <h2>Password field with validation</h2>
         <h2>Outlined Text Field</h2>
         <div id="demo-tf-outlined-wrapper">
           <div id="tf-outlined-example" class="mdc-text-field mdc-text-field--outlined" data-demo-no-auto-js>
-            <input required pattern=".{8,}" type="text" id="tf-outlined-input" class="mdc-text-field__input"
+            <input pattern=".{8,}" type="text" id="tf-outlined-input" class="mdc-text-field__input"
```
kfranqueiro: This line is inconsistent with the JS code below (pattern would be `.{8,}` when required, but `.*` when not)

packages/mdc-textfield/foundation.js
```diff
@@ -301,9 +301,6 @@ class MDCTextFieldFoundation extends MDCFoundation {
    */
   setRequired(isRequired) {
     this.getNativeInput_().required = isRequired;
-    // Addition of the asterisk is automatic based on CSS, but validity checking
```
kfranqueiro: This solution is problematic, because it also leaves things as marked invalid if you change required from true to false... in which case they're no longer invalid if empty. At the very least I think we should still be calling styleValidity if isRequired is false.

Also, I think in the original issue you talked about distinguishing between whether the field was "dirty" or not. I'm not sure if we even have a check for that, though I think what you really mean is whether the input has ever received focus? This solution would just never update the state at all when turning required on.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -204,19 +204,15 @@ test('#setValid updates classes', () => {
 
 test('#setRequired updates CSS classes', () => {
   // Native validity checking does not apply in unittests, so manually mark as valid or invalid.
-  const {foundation, mockAdapter, nativeInput, helperText} =
+  const {foundation, mockAdapter, nativeInput} =
     setupValueTest('', /* isValid */ false);
 
   foundation.setRequired(true);
   assert.isOk(foundation.isRequired());
-  td.verify(mockAdapter.addClass(cssClasses.INVALID));
```
kfranqueiro: If we don't have a negative test for this (testing from required=true to required=false) we should probably add one (see my comment above).

demos/text-field.html
moog16: yes you're correct. I'll sift through the rest of the file and make sure that all works the same.

packages/mdc-textfield/foundation.js
moog16: I did notice your concern in the demo, and I did talk to Dave about that. In both cases the input is still marked as invalid (with or w/o `styleValidity` call). In the Text Field box example, this can be fixed by setting `input.pattern` before setting `input.required`. I thought having to do that was fragile. 

This then led me to question why do we choose to manage `required` in foundation when HTML5 already does this for us? I see `required` as a just another piece of validation, meaning why aren't we managing pattern, minlength, etc. Having to manage all of those properties isn't reasonable, so why do we only choose required?

kfranqueiro: Huh... you're right, even now on the demo page in 0.30.0 it doesn't clear the invalid style when setting required back to false, which seems like a problem... and that particular example doesn't even use `pattern` at all.

HTML5 marks a field as invalid immediately if it is required and blank, which is user-unfriendly and something we're specifically avoiding doing, right? Though I suppose the same is true with `pattern` if the text input is initially populated, but AFAICT pattern isn't applying `:invalid` if the field is blank...

https://codepen.io/kfranqueiro/pen/yvgjGb

moog16: Interesting...I didn't realize the invalid attribute is added. We aren't using :invalid to style invalid state, which is why we don't see this in the demos. In @williamernest's (codepen)[https://codepen.io/williamernest/pen/ddPdxo] we can see the current behavior and the HTML5 behavior, which is what this fix is trying to achieve. 

Do you have any thoughts as to just removing `setRequired` and `get required`?

kfranqueiro: In Will's demo, setting JS required from true to false properly clears the invalid state, reinforcing my original point that I think we should preserve the current logic specifically when setting required to false.

I made my codepen using pure HTML5 because I thought when you said "HTML5 already does this for us" you were referring to OOTB native behavior, not the behavior of our component WRT that attribute. I thought Will was originally only modifying the HTML5 attribute directly with our JS component as a workaround to avoid surfacing this bug in the demo anyway; I'm not sure it's something we should be actively comparing behavior to or seeking to match, since the JS component maintains state itself and setting the attribute directly skips that logic.

Ultimately I don't think we can remove the required setter/getter, given the problem regarding the input still being marked as invalid after it is no longer required. I also don't think we want to rely on the native HTML5 behavior directly rather than set our own class, since as I said above, generally users find that behavior undesirable because fields should not be highlighted as invalid before the user ever interacts with them.

moog16: I don't think Will's codepen is a common use case, but that doesn't mean we shouldn't design for it. Having the styleValidity call in `setRequired` is what is causing the original issue. Are you suggesting that developers use the HTML5 attr directly instead of our JS setter/getter if they see this issue?

kfranqueiro: I'm suggesting pretty much the opposite of that... We need to fix the programmatic API so that it works correctly in both directions (required -> not, not -> required) and developers should not ever have reason to programmatically toggle the HTML attribute directly on a JS text field. (It should work if the HTML attribute is initially set in markup though.)

Currently in our component on master, IIUC, required -> not works correctly programmatically, but not -> required doesn't. In the PR in its current state, the opposite is true. We want both to work.

If this needs further clarification, I will try to respond off-github.

moog16:
After much debate offline, the decision was made to remove set/get required from the component and remove isRequired and setRequired from foundation, and instead use a generic `getValidationAttribute`, `setValidationAttribute` and `removeValidationAttribute`. This will allow for other validation properties than just required. In our demo pages exists pattern and minlength which illustrate the benefit to this. This will also future proof the text-field if any new HTML validation attributes are added with support.

packages/mdc-textfield/index.js
```diff
@@ -203,17 +203,26 @@ class MDCTextField extends MDCComponent {
   }
 
   /**
-   * @return {boolean} True if the Text Field is required.
+   * @return {string} the attribute value from the input element
+   * @param {string} attrName is the name of the attribute on the input element
    */
-  get required() {
-    return this.foundation_.isRequired();
+  getValidationAttribute(attrName) {
```
kfranqueiro: I would still expect this to be exposed via `required` / `pattern` / etc. setters/getters on the component to make it match working with a native input element?

moog16: are you saying keep `get/set required`, but also add `get/set pattern` that proxy to `foundation_.getValidationAttribute` and `foundation_.setValidationAttribute`? Should we add others eg. minlength, maxlength?

kfranqueiro: Yeah, that's what I was thinking, and yes to min/maxlength too.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -666,12 +676,45 @@ test('touchstart on the input sets the line ripple origin', () => {
...
+  let attributeChange;
+  td.when(mockAdapter.registerValidationAttributeChangeHandler(td.matchers.isA(Function)))
+    .thenDo((handler) => attributeChange = handler);
+  foundation.init();
+
+  attributeChange([{attributeName: 'required'}]);
+  td.verify(mockAdapter.removeClass(cssClasses.INVALID));
+  td.verify(helperText.setValidity(true));
+
+  // None of these is affected by setValidationAttribute.
```
kfranqueiro: is -> are

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -666,12 +676,45 @@ test('touchstart on the input sets the line ripple origin', () => {
...
+  const {foundation, mockAdapter} = setupTest();
+  let attributeChange;
+  td.when(mockAdapter.registerValidationAttributeChangeHandler(td.matchers.isA(Function)))
+    .thenDo((handler) => attributeChange = handler);
+  foundation.init();
+
+  attributeChange([{attributeName: 'form'}]);
+  td.verify(mockAdapter.removeClass(cssClasses.INVALID), {times: 0});
+
+  // None of these is affected by setValidationAttribute.
```
kfranqueiro: is -> are

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -352,6 +352,47 @@ test('#adapter.deregisterTextFieldInteractionHandler removes an event handler fo
   td.verify(handler(td.matchers.anything()));
 });
 
+test('#adapter.registerValidationAttributeChangeHandler creates a working mutation observer', (done) => {
+  const {root, component} = setupTest();
+  const handler = td.func('ValidationAttributeChangeHandler');
+  td.when(handler(td.matchers.anything(), td.matchers.anything())).thenDo(() => {
+    assert.isNotOk(root.classList.contains('mdc-text-field--invalid'));
```
kfranqueiro: This is testing more logic than I think we want to be testing here; we just want to be testing that the attribute change called this function, right?

I think we might not actually need any assertions, just the `done`. (I'm not sure what kind of timeout is configured right now, if this were to fail...)

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -352,6 +352,47 @@ test('#adapter.deregisterTextFieldInteractionHandler removes an event handler fo
...
+    done();
+  });
+
+  component.foundation_.adapter_.registerValidationAttributeChangeHandler(handler);
+  root.classList.add('mdc-text-field--invalid');
+  root.querySelector('.mdc-text-field__input').required = true;
+});
+
+test('#adapter.registerValidationAttributeChangeHandler adds an observer to the component', () => {
```
kfranqueiro: This is testing implementation details; I would remove this test

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -352,6 +352,47 @@ test('#adapter.deregisterTextFieldInteractionHandler removes an event handler fo
...
+  const {component} = setupTest();
+  const handler = td.func('ValidationAttributeChangeHandler');
+
+  component.foundation_.adapter_.registerValidationAttributeChangeHandler(handler);
+  assert.equal(component.observers_.length, 2);
+});
+
+test('#adapter.deregisterValidationAttributeChangeHandler disconnects observers from component', () => {
```
kfranqueiro: I don't think we need both this and the next test? One of them should be enough.

test/unit/mdc-textfield/mdc-text-field.test.js
moog16: Timeout is 2000 ms, but the mutationObserver should never take the long. It will just get queued to the next event loop, but I see what you're saying. I could do a `verify(handler(....))`? It seems strange to not assert anything in a test.

moog16: As discussed, I will remove the assert since testing the removal of .mdc-text-field--invalid is not primary to the test. The fact that the done() is called will cause the test to pass. If it is not called, then the test will result in a failure.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -502,6 +481,36 @@ test('on focus makes helper text visible to the screen reader', () => {
...
 });
 
+test('on required validation attribute change should update invalid classes', () => {
+  const {foundation, mockAdapter} = setupTest();
+  let attributeMutate;
+  td.when(mockAdapter.registerValidationAttributeChangeHandler(td.matchers.isA(Function)))
+    .thenDo((handler) => attributeMutate = handler);
+  foundation.init();
+  attributeMutate([{attributeName: 'required'}]);
+  td.verify(mockAdapter.removeClass('mdc-text-field--invalid'));
```
kfranqueiro: Reference the constant in `cssClasses` instead of directly putting the class name inline here and in tests below

demos/text-field.html
```diff
@@ -167,19 +168,31 @@ <h2>Outlined Text Field</h2>
...
           <div id="tf-box-example" class="mdc-text-field mdc-text-field--box" data-demo-no-auto-js>
             <input type="text" id="tf-box" class="mdc-text-field__input"
-                   aria-controls="name-validation-message">
+                   aria-controls="box-name-validation-message">
             <label for="tf-box" class="mdc-text-field__label">Your Name</label>
             <div class="mdc-line-ripple"></div>
           </div>
           <p class="mdc-text-field-helper-text mdc-text-field-helper-text--validation-msg"
-            id="box-text-field-helper-text">
+            id="box-name-validation-message">
```
kfranqueiro: You changed this ID but didn't change references to it below in the JS, which caused the 'use helper text' checkbox under the Text Field Box heading to stop working.

demos/text-field.html
```diff
@@ -357,10 +378,24 @@ <h2>Full-Width Text Field and Textarea</h2>
...
+
       demoReady(function() {
         var tfEl = document.getElementById('tf-outlined-example');
+        var tfHelperText = document.getElementById('tf-outlined-validation-msg');
         var tf = new mdc.textField.MDCTextField(tfEl);
+        var outlinedInputEl = tfEl.querySelector('input');
         var wrapper = document.getElementById('demo-tf-outlined-wrapper');
+        var plainHelperText = 'Helper Text (possibly validation message)';
+        var minLengthError = 'Must be at least 8 characters';
+        var maxLengthError = 'Limit to 10 characters';
```
kfranqueiro: I'd suggest re-wording this message to 'Must not exceed 10 characters'

packages/mdc-textfield/adapter.js
```diff
@@ -106,6 +106,17 @@ class MDCTextFieldAdapter {
...
   deregisterInputInteractionHandler(evtType, handler) {}
 
+  /**
+   * Registers a validation attribute change listener on the input element.
+   * @param {function(!Array): undefined} handler
+   */
+  registerValidationAttributeChangeHandler(handler) {}
+
+  /**
+   * Disconnect all validation attrinute listeners on the input element.
```
kfranqueiro: Typo: attrinute -> attribute

packages/mdc-textfield/foundation.js
```diff
@@ -152,6 +158,23 @@ class MDCTextFieldFoundation extends MDCFoundation {
     this.receivedUserInput_ = true;
   }
 
+  /**
+   * Handles validation attribute changes
+   */
+  handleValidationAttributeMutation(mutationsList) {
+    // whitelist based off of https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Constraint_validation
+    // under section: `Validation-related attributes`
+    const validationAttrWhitelist = [
```
kfranqueiro: This can be hoisted to the top of the file above the class, rather than being defined every time we run this (and rename it `VALIDATION_ATTR_WHITELIST`)

packages/mdc-textfield/index.js
```diff
@@ -206,14 +209,103 @@ class MDCTextField extends MDCComponent {
...
   /**
    * @param {boolean} required Sets the Text Field to required.
    */
   set required(required) {
-    this.foundation_.setRequired(required);
+    this.input_.required = required;
+  }
+
+  /**
+   * @return {string} the input element's validation pattern.
```
kfranqueiro: Nit: Capitalize "The" on these getters

packages/mdc-textfield/index.js
```diff
@@ -206,14 +209,103 @@ class MDCTextField extends MDCComponent {
...
+   */
+  get maxLength() {
+    return this.input_.maxLength;
+  }
+
+  /**
+   * @param {number} maxLength Sets the input element's maxLength.
+   */
+  set maxLength(maxLength) {
+    if (maxLength < 0) {
```
kfranqueiro: I assume this normalization is here for IE and isn't present for minLength because it doesn't support that? I guess this is fair although it makes this smarter than simply being a proxy to the input. Add a comment explaining it though.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -666,12 +675,43 @@ test('touchstart on the input sets the line ripple origin', () => {
...
+test('should not call styleValidity_ on non-whitelisted attribute change', () => {
+  const {foundation, mockAdapter} = setupTest();
+  let attributeChange;
+  td.when(mockAdapter.registerValidationAttributeChangeHandler(td.matchers.isA(Function)))
+    .thenDo((handler) => attributeChange = handler);
+  foundation.init();
+
+  attributeChange([{attributeName: 'form'}]);
+  td.verify(mockAdapter.removeClass(cssClasses.INVALID), {times: 0});
+
```
kfranqueiro: Should this also be verifying that helpertext.setValidity was never called with anything, to match the previous test?

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -502,6 +481,36 @@ test('on focus makes helper text visible to the screen reader', () => {
   td.verify(helperText.showToScreenReader());
 });
 
+test('on required validation attribute change should update invalid classes', () => {
```
kfranqueiro: I'm confused why these 3 tests were added when the 2 below were also added. Aren't both sets testing the same thing? The ones below seem more appropriate.

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -417,9 +439,59 @@ test('get/set valid', () => {
 });
 
 test('get/set required', () => {
-  const {component, mockFoundation} = setupMockFoundationTest();
-  component.required;
-  td.verify(mockFoundation.isRequired());
+  const {component} = setupMockFoundationTest();
   component.required = true;
-  td.verify(mockFoundation.setRequired(true));
+  assert.isOk(component.required);
```
kfranqueiro: Change Ok/NotOk to True/False

packages/mdc-textfield/index.js
moog16: Well if I set maxLength to < 0, Chrome throws a warning. I will add a comment.

test/unit/mdc-textfield/foundation.test.js
moog16: True, these are basically duplicates while the 2 below handle the general case. I'll remove.

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -417,9 +439,59 @@ test('get/set valid', () => {
...
+  assert.equal(component.minLength, 0);
+});
+
+test('get/set maxLength', () => {
+  const {component} = setupMockFoundationTest();
+  component.maxLength = 10;
+  assert.equal(component.maxLength, 10);
+  component.maxLength = -1;
+  // IE11 has a different value for no maxlength property
+  assert.isFalse(component.maxLength === 10);
```
kfranqueiro: Nit: `assert.notEqual(component.maxLength, 10)`

kfranqueiro:
Also just noticed a bug: the helper text element below Text Field Box seems to initially be visible even though the checkbox is initially unchecked. This doesn't seem to be a problem with the first example on the page.

moog16:
Yes looks like it initially needs style='none'. I'll add it inline, since the checkbox toggles it on/off.

moog16:
It seems that Text Field Box is only used as a validation message, which is not given `opacity: 1` when focused on. I'll add the the `Use helper text as validation message` checkbox to make it consistent with the top text field.

packages/mdc-textfield/adapter.js
```diff
@@ -106,6 +106,17 @@ class MDCTextFieldAdapter {
...
+  /**
+   * Registers a validation attribute change listener on the input element.
+   * @param {function(!Array): undefined} handler
+   */
+  registerValidationAttributeChangeHandler(handler) {}
+
+  /**
+   * Disconnect all validation attribute listeners on the input element.
+   */
+  deregisterValidationAttributeChangeHandler() {}
```
kfranqueiro: This should probably just disconnect a specific observer that it is handed, and we should loop within the component's destroy method instead of building the iteration logic into the adapter.

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -320,10 +297,12 @@ test('#init adds event listeners', () => {
   td.verify(mockAdapter.registerInputInteractionHandler('touchstart', td.matchers.isA(Function)));
   td.verify(mockAdapter.registerTextFieldInteractionHandler('click', td.matchers.isA(Function)));
   td.verify(mockAdapter.registerTextFieldInteractionHandler('keydown', td.matchers.isA(Function)));
+  td.verify(mockAdapter.registerValidationAttributeChangeHandler(td.matchers.isA(Function)));
 });
 
 test('#destroy removes event listeners', () => {
   const {foundation, mockAdapter} = setupTest();
+  foundation.validationObserver_ = {test: 'test1'};
```
kfranqueiro: This can just be `{}`

PR #2208, opened 2018-02-07
fix(toolbar): Title position for desktops

pndewit:
When opening the following demo page (with a window which is at least 600px wide) and scrolling the toolbar until it is minimized (mdc-toolbar--flexible-space-minimized class is added), the title is aligned incorrectly with the menu icon. This PR fixes that issue.

https://material-components-web.appspot.com/toolbar/waterfall-toolbar-fix-last-row.html

### DIFF
--- a/packages/mdc-toolbar/mdc-toolbar.scss
+++ b/packages/mdc-toolbar/mdc-toolbar.scss
@@ -169,16 +169,9 @@
 
       align-self: flex-end;
       font-weight: #{map-get($mdc-typography-font-weight-values, normal)};
-      line-height: 1.5rem;
-    }
 
-    @media (min-width: 600px) {
-      &.mdc-toolbar--fixed-lastrow-only {
-        .mdc-toolbar__title,
-        .mdc-toolbar__menu-icon,
-        .mdc-toolbar__icon {
-          line-height: 2rem;
-        }
+      @media (max-width: $mdc-toolbar-mobile-breakpoint) { 
+        line-height: 1.5rem;
       }
     }
 
--- a/packages/mdc-toolbar/mdc-toolbar.scss
+++ b/packages/mdc-toolbar/mdc-toolbar.scss
@@ -172,6 +172,16 @@
       line-height: 1.5rem;
     }
 
+    @media (min-width: 600px) {
+      &.mdc-toolbar--fixed-lastrow-only {
+        .mdc-toolbar__title,
+        .mdc-toolbar__menu-icon,
+        .mdc-toolbar__icon {
+          line-height: 2rem;
+        }
+      }
+    }
+
     .mdc-toolbar__row:first-child::after {
       top: 0;
       left: 0;


packages/mdc-toolbar/mdc-toolbar.scss
```diff
@@ -172,6 +172,16 @@
       line-height: 1.5rem;
     }
 
+    @media (min-width: 600px) {
```
williamernest: The rest of this file uses `max-width: $mdc-toolbar-mobile-breakpoint`. It should stay consistent. 
I think it should be moved to the title class and replace line 172
```
@media (max-width: $mdc-toolbar-mobile-breakpoint) { 
  line-height: 1.5rem;
}
```

williamernest:
Thanks for submitting. We're addressing the icon padding issues with PR 2138. I'll take a look at this again once that gets resolved.

packages/mdc-toolbar/mdc-toolbar.scss
pndewit: True, I did consider using `$mdc-toolbar-mobile-breakpoint`, but that's not mobile first. So my question would then be, why is the entire component not developed mobile first?

Setting a line-height override using `max-width: $mdc-toolbar-mobile-breakpoint` would mean that it would use the `@include mdc-typography(display1);` settings for all screen sizes above `$mdc-toolbar-mobile-breakpoint`. This would mean it will take 2.5rem (https://github.com/material-components/material-components-web/blob/master/packages/mdc-typography/_variables.scss#L62) which is also not what we want it to be.
Besides, I think we only need the override specifically for the `.mdc-toolbar--fixed-lastrow-only` variant right?

williamernest: While I agree that this should be mobile first, this component isn't currently built that way, so I don't see any use in having one code block be different from the rest. 

This behavior also affects the `mdc-toolbar--flexible` variant. 

2.5rem gives the desired behavior in all screen sizes except for mobile, which is fixed by adjusting the line height to 1.5. The only issue remaining is the icon vertical alignment issue, which will be addressed by #2138

pndewit: @williamernest should be updated.

williamernest:
Thanks for submitting this PR. We've decided to move forward with deprecating the current toolbar implementation and recommend that developers switch to using the `top-app-bar` component, which is similar.

pndewit:
Alright, but if it is similar why even create a new component?
I noticed in the `toolbar` docs that only some of the functionality will be implemented in the `top-app-bar`. Am I correct in saying that the remaining functionality should be created in MDC specific implementations?

williamernest:
Yes, if you require functionality that hasn't been recreated in `top-app-bar` already, then you should consider implementing your own component based on the `toolbar` implementation. We have plans to implement a `flexible` variant, but we haven't gotten guidance from our design team yet as to what that should include and look like.

pndewit:
Thanks for the info. One more question if you don't mind, does that mean that the `top-app-bar` will eventually get the same functionalities as the current toolbar (and more)? If so, why was a new component created for this?
I am asking this because I am currently using the `toolbar` component for non-"top-app" header use cases as well which will make the component name kinda strange for those cases.

williamernest:
The `top-app-bar` will get some additional functionality that didn't exist in the previous `toolbar`, but it won't get all of the functionality, I believe. Some of the variants aren't written anywhere in spec, so it's difficult to determine what we want to keep about it. 

The new package was due to the `toolbar` implementation was actually just a `top-app-bar`. At some point we'd like to create additional functionality in the `toolbar` package that is more generic, and not for a specific style of toolbar.

PR #2215, opened 2018-02-08
feat(chips): Add `mdc-chip-set--choice` variant

### LINKED ISSUE #2008: mdc-chip-set--choice
- [ ] Add a new variant to `mdc-chip-set` called `mdc-chip-set--choice`
- [ ] Use mdc-states mixins  on `mdc-chip` for selected/activated state
- [ ] modify JS for single-select behavior

bonniezhou:
Choice chips are a variant of chips that allow single selection from a set of chip options.
The selected chip will have the `&--activated` modifier class attached to apply styles using the `mdc-states-activated` mixin.

Resolves #2008.


packages/mdc-chips/chip-set/adapter.js
```diff
@@ -31,8 +31,23 @@ class MDCChipSetAdapter {
...
+   * @param {string} evtType
+   * @param {function(!Event): undefined} handler
+   */
+  registerEventHandler(evtType, handler) {}
+
+  /**
+   * Deregisters an event handler on the root element for a given event.
+   * @param {string} evtType
+   * @param {function(!Event): undefined} handler
+   */
```
acdvorak: It looks like there are a few linter errors:

```
/packages/mdc-chips/chip-set/foundation.js
  57:8  error  Trailing spaces not allowed  no-trailing-spaces
/packages/mdc-chips/chip-set/index.js
  22:18  error  'MDCChipFoundation' is defined but never used. Allowed unused vars must match /^(?:(?:MDC(?:(?:[A-Z][a-z0-9]+)+)Adapter)|(?:(?:(?:[A-Z][a-z0-9]+)+)Type))$/  no-unused-vars
/packages/mdc-chips/chip/adapter.js
  49:1  error  Trailing spaces not allowed  no-trailing-spaces
/packages/mdc-chips/chip/foundation.js
  75:1  error  Trailing spaces not allowed  no-trailing-spaces
```

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -45,6 +53,42 @@ class MDCChipSetFoundation extends MDCFoundation {
    */
   constructor(adapter) {
     super(Object.assign(MDCChipSetFoundation.defaultAdapter, adapter));
+
+    /** 
+     * The active chips in the set. Only used for choice chip set or filter chip set.
+     * @private {!Array<!MDCChip>}
```
acdvorak: The Closure test is failing on this line:

```
.closure-tmp/packages/mdc-chips/chip-set/foundation.js:64: ERROR - Bad type annotation. Unknown type MDCChip
     * @private {!Array<!MDCChip>}
                         ^
```

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -45,6 +53,42 @@ class MDCChipSetFoundation extends MDCFoundation {
...
+  }
+
+  /**
+   * Handles a chip interaction event
+   * @param {!Object} evt
+   * @private
+   */
+  handleChipInteraction_(evt) {
+    const {chip} = evt.detail;
+    if (this.adapter_.hasClass(cssClasses.CHOICE)) {
```
acdvorak: Can we invert this `if` condition and bail out early if the root element does _not_ have the `--choice` class?

That'll reduce nesting and make the code a bit easier to read.

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -45,6 +53,42 @@ class MDCChipSetFoundation extends MDCFoundation {
...
+      if (this.activeChips_.length == 0) {
+        this.activeChips_[0] = chip;
+      } else if (this.activeChips_[0] != chip) {
```
acdvorak: Nit: Always use strict inequality (`!==`)

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -45,6 +53,42 @@ class MDCChipSetFoundation extends MDCFoundation {
...
+
+  /**
+   * Handles a chip interaction event
+   * @param {!Object} evt
+   * @private
+   */
+  handleChipInteraction_(evt) {
+    const {chip} = evt.detail;
+    if (this.adapter_.hasClass(cssClasses.CHOICE)) {
+      if (this.activeChips_.length == 0) {
```
acdvorak: Nit: Always use strict equality (`===`)

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -45,6 +53,42 @@ class MDCChipSetFoundation extends MDCFoundation {
    */
   constructor(adapter) {
     super(Object.assign(MDCChipSetFoundation.defaultAdapter, adapter));
+
+    /** 
+     * The active chips in the set. Only used for choice chip set or filter chip set.
+     * @private {!Array<!MDCChip>}
+     */
+    this.activeChips_ = [];
+    /** @private {function(!Event): undefined} */
```
acdvorak: Nit: Add a blank line before this one

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -22,17 +22,29 @@
...
+  @include mdc-states;
+  @include mdc-states-activated(black);
+  
   @include mdc-chip-corner-radius($mdc-chip-border-radius-default);
   @include mdc-chip-fill-color($mdc-chip-fill-color-default);
   @include mdc-chip-ink-color($mdc-chip-ink-color-default);
 
   display: inline-block;
+  border-radius: 16px;
+  background: rgba(0, 0, 0, 0.08);
```
acdvorak: Can you use the theme mixins for these properties?

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -22,17 +22,29 @@
...
+  padding: 7px 12px;
   margin: 5px;
   padding: $mdc-chip-vertical-padding $mdc-chip-horizontal-padding;
   cursor: pointer;
   overflow: hidden;
 }
 
+.mdc-chip:hover,
+.mdc-chip--activated{
+  color: rgba(0, 0, 0, 1);
```
acdvorak: Ditto about using theme mixin

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -45,6 +53,42 @@ class MDCChipSetFoundation extends MDCFoundation {
    */
   constructor(adapter) {
     super(Object.assign(MDCChipSetFoundation.defaultAdapter, adapter));
+
+    /** 
+     * The active chips in the set. Only used for choice chip set or filter chip set.
+     * @private {!Array<!MDCChip>}
+     */
+    this.activeChips_ = [];
```
acdvorak: I'm guessing this is an array because you'll need to support multiple selected chips in the future?

demos/chips.html
```diff
@@ -53,6 +53,27 @@
         </div>
       </section>
 
+      <section class="example">
+        <h2>Choice Chips</h2>
+        <div class="mdc-chip-set mdc-chip-set--choice">
+          <div class="demo-chip mdc-chip">
```
acdvorak: These chips should probably have `tabindex="0"` and one or more [ARIA attributes](https://www.w3.org/TR/wai-aria-1.1/#widget_roles).

I'm guessing you'll need something like this (but definitely double check first, cuz I'm not an ARIA expert):

```html
<div class="mdc-chip-set mdc-chip-set--choice" role="radiogroup">
  <div class="demo-chip mdc-chip" role="radio" aria-checked="false">
    <div class="mdc-chip__text">Extra Small</div>
  </div>
  ...
</div>
```

The JS will need to update the `aria-checked` attribute whenever a chip is selected/activated (I forget what the difference is between the two).

You'll also need to add styles for the `:focus` state.

kfranqueiro: FYI I left a similar comment related to what elements we should use and which elements should be focusable on #2191.

packages/mdc-chips/chip/mdc-chip.scss
bonniezhou: Whoops, bad merge. Done!

packages/mdc-chips/chip-set/foundation.js
bonniezhou: This is preemptively structured this way to account for the other chip set variants coming soon, but I'll invert it for now and change it in a later PR.

packages/mdc-chips/chip-set/foundation.js
bonniezhou: Yep, specifically in the `--filter` variant.

bonniezhou:
According to spec, baseline chips have a text color of 60% black on enabled, and 100% black on hover/focus/pressed/activated. The `mdc-chip-ink-color-activated` mixin was added to allow for customization for those states. It is also called inside the `mdc-chip-fill-color-accessible` mixin.

packages/mdc-chips/chip/mdc-chip.scss
bonniezhou: Done! Also see PR comment about activated text color.

demos/chips.html
bonniezhou: Ah yes, thanks for bring this up! Let's handle it in #2191.

packages/mdc-chips/_mixins.scss
```diff
@@ -40,9 +42,17 @@
 
 @mixin mdc-chip-ink-color($color) {
   @include mdc-states($color);
+  @include mdc-states-activated($color);
   @include mdc-theme-prop(color, $color);
 }
 
+@mixin mdc-chip-ink-color-activated($color) {
+  &:hover,
```
acdvorak: OOC, why does this set the ink color of both hover _and_ activated?

Actually, after reading your other PR comment, I'm guessing it's because the spec says that baseline chips have a text color of 60% black on enabled, and 100% black on hover/focus/pressed/activated.

However, clients are inevitably going to want to deviate from that guidance, guaranteed. So we need a way to do it for our own baseline styles, without being overly-restrictive for clients.

It seems like there are three options:

1. Create separate mixins for every state and/or combination of states.
    E.g.:
    ```
    .mdc-chip {
      @include mdc-chip-ink-color($color1);
      @include mdc-chip-ink-color-hover($color2);
      @include mdc-chip-ink-color-activated($color3);
      @include mdc-chip-ink-color-activated-hover($color4);
    }
    ```
2. Create one mixin that has an optional parameter controlling which state(s) it targets.
    E.g.:
    ```
    .mdc-chip {
      @include mdc-chip-ink-color($color1, $states: (idle));
      @include mdc-chip-ink-color($color2, $states: (hover, activated));
    }
    ```
3. Remove this `-activated` mixin and let clients call `mdc-chip-ink-color` from within their own CSS selectors.
    E.g.:
    ```
    .mdc-chip {
      @include mdc-chip-ink-color($color1);
    }
    .mdc-chip:hover {
      @include mdc-chip-ink-color($color2);
    }
    .mdc-chip--activated {
      @include mdc-chip-ink-color($color3);
    }
    .mdc-chip--activated:hover {
      @include mdc-chip-ink-color($color4);
    }
    ```

Personally, I'd go with option 3 because it's the simplest and most BEM-friendly - _especially_ if there are multiple (sometimes-)overlapping states like activated, focus, hover, and pressed.

Either way, we normally try to avoid having mixins do more than one thing unless it's absolutely necessary for technical reasons (e.g., checkbox needs to set `stroke` _and_ `background-color` for SVG).

Otherwise, the mixin names get confusing ("you mean `-activated()` sets the color of both activated _and_ hover states...?") or verbose (`-activated-and-hover`).

If we end up keeping the `-ink-color-activated()` mixin, we should probably also add a `-fill-color-activated()` mixin.

packages/mdc-chips/_mixins.scss
```diff
@@ -40,9 +42,17 @@
...
+  &--activated {
```
acdvorak: I think we normally try to avoid appending class names within mixins, because it doesn't always work intuitively for our clients (especially if they use non-BEM selectors like `#id`).

(Ripple seems to be the only exception to this rule, but I'm not sure why we needed to do it there.)

You could potentially do something like this, but it's kinda gross:

```scss
@mixin mdc-chip-ink-color-activated($color) {
  @if & {
    &:hover,
    &.mdc-chip--activated {
      @include mdc-chip-ink-color($color);
    }
  } @else {
    .mdc-chip:hover,
    .mdc-chip--activated {
      @include mdc-chip-ink-color($color);
    }
  }
}
```

This is one of the reasons I prefer option 3 above: It's a lot cleaner and less complicated, which makes our styles easier to maintain, support, and debug :slightly_smiling_face:

acdvorak:
It looks like you'll need to water your PR branch with master to fix the failing webpack test:

`npm run test:buildconfig`

packages/mdc-chips/_mixins.scss
kfranqueiro: Looking briefly at this I think I'd pick option 3 out of the 3 options. Definitely not 2.

For that matter... why is focus not documented? I think that's another design question.

FYI, the way the [mdc-ripple states mixins](https://github.com/material-components/material-components-web/tree/master/packages/mdc-ripple#advanced-states-mixins) handle this is by exposing specific mixins for hover/focus/press, then expecting you will use them yourselves within selected/activated because not all components have those states. This is a similar situation since not all chip types have activated/selected state, either.

Also, before I forget to ask this question again... Why are we using activated rather than selected here? Isn't choice a rather transient thing rather than something persistent/navigational?

bonniezhou: The reason we can't use option 3 is because `source-map` seems to get confused about the `--activated` selector.

Say I have
```
.mdc-chip--activated {
  @include mdc-chip-ink-color(black);
}
```
Then `mdc-chip-ink-color` calls `mdc-states-activated` which sets a style on the `&--activated` selector, which in the end resolves to a selector that is `mdc-chip--activated--activated`.

Here's the error, and correct me if I'm misunderstanding it (FYI line 147 in ripple/_mixins.scss is where the `&--activated` selector is styled in `mdc-states-activated`):
```
throw new Error('Invalid mapping: ' + JSON.stringify({
      ^

Error: Invalid mapping: {"generated":{"line":1269,"column":2},"source":"/Users/bonniez/material-components-web/packages/material-components-web/node_modules/@material/ripple/_mixins.scss","original":{"line":147,"column":-1},"name":null}
    at SourceMapGenerator_validateMapping [as _validateMapping] (/Users/bonniez/material-components-web/node_modules/source-map/lib/source-map-generator.js:277:13)
````

One alternative is to remove `mdc-states-activated` from `mdc-chip-ink-color` and just manually set the background color on `mdc-chip--activated`, in which case we can do option 3 that Andy proposed.  Thoughts?

Comment (bonniezhou) [reply packages/mdc-chips/_mixins.scss]: Ok since the activated state is kind of a special case (it's a class not a pseudo class unlike the other states) I think we should remove `mdc-states-activated` from `mdc-chip-ink-color` and bundle the ripple opacity change and the ink color change into one mixin called `mdc-chip-activated`:
```
@mixin mdc-chip-activated($activated-ink-color) {
  &.mdc-chip {
    // This changes the opacity of the ripple based on whether $activated-ink-color is light or dark.
    // It does not change the color of the ripple.
    @include mdc-states-activated($activated-ink-color);
  }

  &.mdc-chip--activated {
    @include mdc-chip-ink-color($activated-ink-color);
  }
}
```

Then everything else can be done as per option 3 above (except for activated/hover which , so:
```
.mdc-chip {
  @include mdc-chip-ink-color($color1);
  @include mdc-chip-activated($color3);
}
.mdc-chip:hover {
  @include mdc-chip-ink-color($color2);
  @include mdc-chip-activated($color4);
}
```

bonniezhou: Let me know if the solution in this [commit](https://github.com/material-components/material-components-web/pull/2215/commits/24fef1e23f25d7a761ca400159746aa2ce04c5c5) looks ok. Main changes:
- `mdc-chip-ink-color` calls `mdc-chip-activated-ink-color` and changes color on hover. This is because the baseline chip needs activated/hover ink color to be set to black, and I think it'd be really annoying for clients to call `mdc-chip-ink-color` and be surprised that the activated/hover styles didn't change along with it. All other states styles can be overridden manually.
- As mentioned above, `mdc-chip-activated-ink-color` bundles the ripple opacity change and the ink color change.

packages/mdc-chips/_mixins.scss
bonniezhou: I removed the `&--activated`, but I think pseudo-classes (`&:hover`) shouldn't be a problem, right?

packages/mdc-chips/_mixins.scss
bonniezhou: And to answer @kfranqueiro's question, we're using `activated` instead of `selected` only because the value of `activated` in the ripple opacity map matches the one in the guidelines (12% for dark, 24% for light).

kfranqueiro: Interesting. This sounds like something we should run by our states designer to be totally sure we're doing the right thing for chips.

packages/mdc-chips/_mixins.scss
acdvorak: Right, the pseudo `&:hover` should be fine!

packages/mdc-chips/_mixins.scss
acdvorak: Yeah, we wanna be consistent if possible.

Perhaps we should name the mixins and CSS classes `-selected`, and just call the `mdc-states-activated()` mixin internally, since it's really only an implementation detail.

kfranqueiro: That...doesn't sound consistent though, that sounds confusing. At that point we might as well just pass our own opacities independently instead of using the stock `activated` mixin.

Let's just make sure with design whether "activated" is the right concept to be using here.

bonniezhou: In design,
1. Choice chips are "activated" and filter chips are "selected" (yup, really).
2. Both activated choice chips and selected filter chips have a fill opacity of 12%.

In code,
1. The default ripple opacity in "activated" state is 12% and in "selected" state is 4%.

I agree that for consistency what we name the public mixins/CSS classes should be the same as the mixin what we call internally. I also think that we should have a single term to describe the interacted-with state of both choice and filter chips, since the opacity change is the same for both. Whether that term is "activated" or "selected" doesn't matter to me, but I used "activated" because its default ripple opacity was already 12%.

PR #2218, opened 2018-02-08
docs(menu): wrong boolean value

j-o-d-o:
I think it quickOpen should be set to true in order to turn off menu animation.


kfranqueiro:
Thanks for catching this!

PR #2229, opened 2018-02-11
Making slidable drawer open when swiping from the left side of the screen

jonhue:
_Expanding on #1976_

@acdvorak 

I finally had time to implement this feature.

It's still far from perfect though. Swiping from the left edge of the screen (or the right of used in RTL-mode) does not gradually open the drawer like in many Android apps. Instead it opens the drawer after a gesture ended.
It does use the same mechanism as the drawer closing gesture. It opens the drawer after a gesture completed that had a length of 50% of the drawers width or more. To check whether or not a gesture was made with the intention to open the drawer I settled on checking if it started <= than 35px away from the edge where the drawer is located.
Some mechanism that could be implemented as well (if you think that this would be beneficial), is to check whether a gesture which is currently considered to intend a drawer opening but also scrolled down ~200px or more should not be considered. This however could only be checked after a gesture completed.

I would love to hear what you think + some ideas how we could implement a gradual drawer opening as the gesture happens.

(Please excuse the changed package files. My briefly setup environment caused them to change.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,16 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "dev": true,
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
     "abbrev": {
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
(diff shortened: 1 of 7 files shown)


jonhue:
Oh, and you can test this out here: https://yaeme.com

touficbatache:
Take a look at http://materializecss.com It opens as you slide!!

touficbatache:
@jonhue This is a JS library that makes the drawer slidable, and it opens as you slide your finger: https://slideout.js.org

jonhue:
@touficbatache The performance seems to be pretty bad...

touficbatache:
@jonhue It works great on my Android (Android 8, HTC 10)

jonhue:
@touficbatache I have a fully speced XPS 9370 and was greeted by a 2s black screen when opening https://slideout.js.org.

touficbatache:
@jonhue Hmm ok, maybe you could change the code to make it faster? I'm just suggesting:

> I would love to hear what you think + some ideas how we could implement a gradual drawer opening as the gesture happens.

jonhue:
@touficbatache It would certainly get a little more complex, I'd have to make time for that.

jonhue:
Are there any plans on merging this?

moog16:
@jonhue thanks for your contribution, but we're actually in the middle of rebuilding drawer to match the current spec https://github.com/material-components/material-components-web/issues/2872. Design is also improve guidance for gestures, which may/may not include swiping open/closed the drawer.

PR #2235, opened 2018-02-12
docs(menu): Remove unsupported initially-open DOM example

### LINKED ISSUE #2217: Menu: initial open state -> wrong position of menu
0.30.0

Chrome 64.0.3282.140

### What OS are you using?
Win 10

Use the `mdc-menu--open` class on the menu.
https://codepen.io/j-o-d-o/pen/zppEdL

The menu is open as usual

The menu is shifted to the right of the button that is opening the menu.

I am not sure this is on purpose or if it is a bug?

kfranqueiro:
Setting a menu as initially open via the `--open` modifier class has probably not worked for a long time and is a questionable use case at best, but the docs still mentioned it.

(There was also apparently a generous helping of trailing whitespace in here, which this PR removes.)

Fixes #2217.

### DIFF
--- a/packages/mdc-menu/README.md
+++ b/packages/mdc-menu/README.md
@@ -40,9 +40,9 @@ npm install --save @material/menu
 
 ## Usage
 
-### HTML Structure 
+### HTML Structure
 
-A menu is usually closed, appearing when opened. It is appropriate for any display size.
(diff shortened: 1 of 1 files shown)


kfranqueiro:
Nice catch, I updated that line too.

PR #2237, opened 2018-02-12
chore(floating-label): separate label module from text-field

### LINKED ISSUE #2126: Split floating label into a new Node module
Split floating label into a new Node module

In the future, we will update MDCSelect to use floating label

moog16:
fixes #2126

BREAKING CHANGE: must use `.mdc-floating-label` selector instead of `.mdc-text-field__label`


packages/mdc-floating-label/_mixins.scss
```diff
@@ -48,22 +48,22 @@
...
+@include mdc-floating-label-invalid-shake-keyframes_(box-dense, $mdc-floating-label-box-dense-position-y, 0%, $mdc-floating-label-dense-scale);
+@include mdc-floating-label-invalid-shake-keyframes_(outlined, $mdc-floating-label-outlined-position-y);
+@include mdc-floating-label-invalid-shake-keyframes_(outlined-dense, $mdc-floating-label-outlined-dense-position-y, 0%, $mdc-floating-label-dense-scale);
+@include mdc-floating-label-invalid-shake-keyframes_(outlined-leading-icon, $mdc-floating-label-outlined-position-y, $mdc-floating-label-outlined-with-leading-icon-position-x);
+@include mdc-floating-label-invalid-shake-keyframes_(outlined-leading-icon-dense, $mdc-floating-label-outlined-dense-position-y, $mdc-floating-label-outlined-dense-with-leading-icon-position-x, $mdc-floating-label-dense-scale);
+@include mdc-floating-label-invalid-shake-keyframes_(textarea, $mdc-floating-label-textarea-position-y, 0%, $mdc-floating-label-textarea-scale);
 
-@mixin mdc-text-field-label-float_($positionY, $positionX: 0%, $scale: .75) {
+@mixin mdc-floating-label-position_($positionY, $positionX: 0%, $scale: .75) {
```
moog16: maybe this should be `mdc-floating-label-float-position`? Is that clearer?

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+
+@mixin mdc-floating-label-color($color) {
+  .mdc-floating-label {
+    @include mdc-theme-prop(color, $color);
+  }
+}
+
+// Private
+
+@mixin mdc-floating-label-invalid-shake-keyframes_($modifier, $positionY, $positionX: 0%, $scale: .75) {
```
moog16: I think it might make more sense to make this public, and move the @includes to the text-field scss file, along with the variables. Reason being select might not have the same variants (box, outlined, outlined-dense, etc). 

Want to see what reviewer thinks before doing this.

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+@mixin mdc-floating-label-shake_($modifier) {
+  .mdc-floating-label--shake {
+    animation: invalid-shake-float-above-#{$modifier} 250ms 1;
+  }
+}
+
+// Used for textarea in case of scrolling
+@mixin mdc-floating-label-background-color_($color) {
```
moog16: I think this needs to be public

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+      transform: translateY(-1 * $positionY) translateX(-1 * $positionX) scale($scale);
+
+      @include mdc-rtl {
+        transform: translateY(-1 * $positionY) translateX($positionX) scale($scale);
+      }
+    }
+  }
+}
+
+@mixin mdc-floating-label-shake_($modifier) {
```
moog16: This is used heavily in text-field, so I think this should actually be made public, along with reasons made in https://github.com/material-components/material-components-web/pull/2237/files#r167720790

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+</label>
+```
+
+> _NOTE_: Only place a `mdc-floating-label` inside of a text field _if you plan on using
+> Javascript_. Otherwise, the label must go outside of the text-field, as shown below.
+
+#### Single Line, CSS Only
+
+```html
+<label for="text-field-no-js">TextField with no JS: </label>
```
acdvorak: Nit: Remove trailing space just before `</label>`

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+--- | ---
+`mdc-floating-label` | Mandatory
+`mdc-floating-label--float-above` | Indicates the label is floating above the text field
+`mdc-floating-label--shake` | Shakes the label
+
+### SCSS Mixins
+
+Mixin | Description
+--- | ---
+`mdc-floating-label-color($color)` | Customizes the color of the label.
```
acdvorak:
```
`mdc-floating-label-ink-color($color)` | Customizes the ink color of the label.
```

* "Ink color" === CSS `color` property (usually)
* "Fill color" === CSS `background-color` property (usually)

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+`mdc-floating-label-fill-color($color)` | Customizes the fill color of the label.
+`mdc-floating-label-invalid-shake-keyframes($modifier, $positionY, $positionX, $scale)` | Creates a shake keyframe animation for invalid label shake. To be used with sass mixin `mdc-floating-label-shake($modifier)`.
```
acdvorak:
```
Generates a CSS `@keyframes` at-rule for an invalid label shake. Used in conjunction with the `mdc-floating-label-shake-animation` mixin.
```

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+`mdc-floating-label-shake($modifier)` | Applies shake keyframe animation to label.
```
acdvorak: Let's rename these two mixins for consistency:

```
mdc-floating-label-shake-keyframes($modifier, $positionY, $positionX, $scale)
mdc-floating-label-shake-animation($modifier)
```

Hopefully that makes their relationship a bit clearer, too.

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+`mdc-floating-label-float-position($positionY, $positionX, $scale)` | Sets position of label when floating.
+
+### `MDCFloatingLabel`
+
+Method Signature | Description
+--- | ---
+`shake(isValid: boolean, isFocused: boolean) => void` | Shakes label when invalid
```
acdvorak: If `isValid = false`, is this method a noop? Can we rephrase the description to clarify this?

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+`float(value: string, isFocused: boolean, isBadInput: boolean) => void` | Floats label above text field or select
+`getWidth() => number` | Returns the width of the label element
+
```
acdvorak: Remove this extra line

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+`addClass(className: string) => void` | Adds a class to the label element
+`removeClass(className: string) => void` | Removes a class from the label element
+`getWidth() => number` | Returns the width of the label element
+
+### `MDCFloatingLabelFoundation`
+
+Method Signature | Description
+--- | ---
+`getWidth() => number` | Returns the width of the label element
+`styleShake(isValid: boolean, isFocused: boolean)` | Styles the label to produce the shake effect when needed.
```
acdvorak: What does "when needed" mean?

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,107 @@
...
+`styleFloat(value: string, isFocused: boolean, isBadInput: boolean)` | Styles the label to float or defloat as necessary.
```
acdvorak: What does "as necessary" mean?

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+      transform: translateX(calc(-4% - #{$positionX})) translateY(-#{$positionY}) scale(#{$scale});
+    }
+
+    100% {
+      transform: translateX(calc(0 - #{$positionX})) translateY(-#{$positionY}) scale(#{$scale});
+    }
+  }
+}
+
+@include mdc-floating-label-invalid-shake-keyframes(standard, $mdc-floating-label-position-y);
```
acdvorak: We normally don't emit any CSS in a `_mixins` file. Move this line to `mdc-floating-label.scss`.

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,78 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "./variables";
+
+// Public
+
+@mixin mdc-floating-label-color($color) {
+  .mdc-floating-label {
```
acdvorak: This one is tricky...

We usually try to avoid putting _block_ CSS selectors in mixins because it violates BEM.

However, floating label is kind of a weird case, because although the class name looks like a block, it really acts more like a sub-element...

Let's chat about this with the rest of the team on Slack and see what they think.

Ditto for the mixins below.

packages/mdc-floating-label/index.js
```diff
@@ -0,0 +1,75 @@
...
+   * @param {boolean} isValid Whether the input's value is valid (passes all
+   *     validity checks).
+   * @param {boolean} isFocused Whether the input is focused.
+   */
+  shake(isValid, isFocused) {
+    this.foundation_.styleShake(isValid, isFocused);
+  }
+
+  /**
+   * Styles the label to float or defloat as necessary.
```
acdvorak: What does "as necessary" mean?

packages/mdc-floating-label/index.js
```diff
@@ -0,0 +1,75 @@
...
+   */
+  getWidth() {
+    return this.foundation_.getWidth();
+  }
+
+  /**
+   * @return {!MDCFloatingLabelFoundation}
+   */
+  getDefaultFoundation() {
+    return new MDCFloatingLabelFoundation(/** @type {!MDCFloatingLabelAdapter} */ (Object.assign({
```
acdvorak: Why is `Object.assign()` necessary?

packages/mdc-textfield/_mixins.scss
```diff
@@ -16,9 +16,9 @@
 
 @import "@material/line-ripple/mixins";
 @import "outline/mixins";
-@import "label/mixins";
-@import "label/variables";
+@import "@material/floating-label/mixins";
```
acdvorak: Keep the imports grouped and alphabetized (ditto line 21)

packages/mdc-textfield/adapter.js
```diff
@@ -146,6 +146,33 @@ class MDCTextFieldAdapter {
...
+
+  /**
+   * Shakes the floating label.
+   * @param {boolean} isValid
+   * @param {boolean} isFocused
+   */
+  shakeLabel(isValid, isFocused) {}
+
+  /**
+   * Floats the floating label element above the text-field
```
acdvorak: Nit: Add period

packages/mdc-textfield/adapter.js
```diff
@@ -146,6 +146,33 @@ class MDCTextFieldAdapter {
...
+  floatLabel(value, isFocused, isBadInput) {}
+
+  /**
+   * Sets the transform origin of the line ripple.
+   * @return {boolean}
+   */
+  hasLabel() {}
+
+  /**
+   * Get width of label in pixels
```
acdvorak: Nit: Add period

packages/mdc-textfield/adapter.js
```diff
@@ -146,6 +146,33 @@ class MDCTextFieldAdapter {
    * @param {number} normalizedX
    */
   setLineRippleTransformOrigin(normalizedX) {}
+
+  /**
+   * Shakes the floating label.
```
acdvorak: What happens if there's no label? Ditto for other methods below.

packages/mdc-textfield/adapter.js
```diff
@@ -146,6 +146,33 @@ class MDCTextFieldAdapter {
...
+
+  /**
+   * Sets the transform origin of the line ripple.
+   * @return {boolean}
+   */
+  hasLabel() {}
+
+  /**
+   * Get width of label in pixels
+   * @return {number}
```
acdvorak: What does this return if `hasLabel()` is `false`?

packages/mdc-textfield/foundation.js
```diff
@@ -109,8 +111,8 @@ class MDCTextFieldFoundation extends MDCFoundation {
   init() {
     this.adapter_.addClass(MDCTextFieldFoundation.cssClasses.UPGRADED);
     // Ensure label does not collide with any pre-filled value.
-    if (this.label_ && this.getValue()) {
-      this.label_.styleFloat(
+    if (this.getValue()) {
```
acdvorak: Should we keep the `this.hasLabel()` check?

Ditto below

test/unit/mdc-textfield/foundation.test.js
```diff
@@ -411,22 +395,22 @@ test('#updateOutline does nothing if no outline is present', () => {
...
     .thenDo((evtType, handler) => {
       input = handler;
     });
   foundation.init();
-  td.verify(label.styleFloat(), {times: 0});
+  td.verify(mockAdapter.floatLabel(), {times: 0});
   input();
-  td.verify(label.styleShake(/* isValid */ true, /* isFocused */ true));
-  td.verify(label.styleFloat(/* value */ '', /* isFocused */ true, /* isBadInput */ false));
+  td.verify(mockAdapter.shakeLabel(/* isValid */ true, /* isFocused */ true));
```
acdvorak: Let's check with the rest of the team on this, but I would much prefer to use a destructured object instead of individual boolean params.

E.g.:

```js
shakeLabel({isValid: true, isFocused: true})
```

packages/mdc-floating-label/README.md
moog16: Yes that sound much more concise

packages/mdc-floating-label/README.md
moog16: Should I also remove `invalid` from the keyframe animation name?

`invalid-shake-float-above-#{modifier}`?

moog16: renamed them to just `shake-float-above-#{modifier}`

packages/mdc-floating-label/README.md
moog16: How about:

"Label animation class is applied unless isValid or isFocused is true."

packages/mdc-floating-label/index.js
moog16: not necessary to my knowledge. Removing.

packages/mdc-textfield/adapter.js
moog16: Hmm good point...we have a guard in the component, but not in the foundation. I think what I'll do is move the guard into the foundation instead.

test/unit/mdc-textfield/foundation.test.js
moog16: Seems like the group has spoken and prefers inline comments, what say you?

packages/mdc-floating-label/_mixins.scss
moog16: looked at line ripple, which just removes the .mdc-line-ripple from its mixin. This should also work for floating-label

packages/mdc-floating-label/README.md
```diff
@@ -75,33 +75,32 @@ CSS Class | Description
...
 
 ### `MDCFloatingLabel`
 
 Method Signature | Description
 --- | ---
-`shake(isValid: boolean, isFocused: boolean) => void` | Shakes label when invalid
-`float(value: string, isFocused: boolean, isBadInput: boolean) => void` | Floats label above text field or select
-`getWidth() => number` | Returns the width of the label element
-
+`shake(isValid: boolean, isFocused: boolean) => void` | Label animation class is applied unless isValid or isFocused is true.
```
acdvorak: I think we should just have a single boolean param called `shouldShake`, and let the caller determine its value.

What you have now is a "leaky abstraction": text-field logic is leaking into a different component.

packages/mdc-floating-label/README.md
```diff
@@ -75,33 +75,32 @@ CSS Class | Description
...
+`float(value: string, isFocused: boolean, isBadInput: boolean) => void` | Floats label above text field or select.
```
acdvorak: Ditto - this method should take a single boolean param called `shouldFloat`.

packages/mdc-floating-label/README.md
```diff
@@ -75,33 +75,32 @@ CSS Class | Description
...
 
 ### `MDCFloatingLabelFoundation`
 
 Method Signature | Description
 --- | ---
-`getWidth() => number` | Returns the width of the label element
-`styleShake(isValid: boolean, isFocused: boolean)` | Styles the label to produce the shake effect when needed.
-`styleFloat(value: string, isFocused: boolean, isBadInput: boolean)` | Styles the label to float or defloat as necessary.
+`getWidth() => number` | Returns the width of the label element.
+`styleShake(isValid: boolean, isFocused: boolean)` | Will remove shake class if isValid _or_ isFocused is true, otherwise it will add shake class.
```
acdvorak: Ditto

packages/mdc-floating-label/README.md
```diff
@@ -75,33 +75,32 @@ CSS Class | Description
...
+`styleFloat(value: string, isFocused: boolean, isBadInput: boolean)` | Will add class to float label if isFocused, or if value is present. Otherwise it will remove classes unless isBadInput is true.
```
acdvorak: Ditto

packages/mdc-floating-label/_mixins.scss
```diff
@@ -19,13 +19,11 @@
...
-  .mdc-floating-label {
-    @include mdc-theme-prop(color, $color);
-  }
+  @include mdc-theme-prop(color, $color);
 }
 
-@mixin mdc-floating-label-invalid-shake-keyframes($modifier, $positionY, $positionX: 0%, $scale: .75) {
-  @keyframes invalid-shake-float-above-#{$modifier} {
+@mixin mdc-floating-label-shake-keyframes($modifier, $positionY, $positionX: 0%, $scale: .75) {
+  @keyframes shake-float-above-#{$modifier} {
```
acdvorak: CSS `@keyframes` are globally-scoped, so we need to prefix all of their names with `mdc-floating-label-` to avoid collisions.

packages/mdc-textfield/_mixins.scss
```diff
@@ -14,13 +14,13 @@
 // limitations under the License.
 //
 
-@import "@material/line-ripple/mixins";
-@import "outline/mixins";
-@import "@material/floating-label/mixins";
-@import "helper-text/mixins";
-@import "./variables";
 @import "icon/mixins";
```
acdvorak: 1. All subpaths of the current directory should start with `./`
2. Imports should be alphabetized
3. "External" imports (`@material/foo`) should go at the top

E.g.:

```scss
@import "@material/floating-label/mixins";
@import "@material/line-ripple/mixins";
@import "./helper-text/mixins";
@import "./icon/variables";
@import "./outline/mixins";
@import "./variables";
```

packages/mdc-floating-label/README.md
```diff
@@ -85,8 +85,8 @@ Mixin | Description
 
 Method Signature | Description
 --- | ---
-`shake(isValid: boolean, isFocused: boolean) => void` | Label animation class is applied unless isValid or isFocused is true.
-`float(value: string, isFocused: boolean, isBadInput: boolean) => void` | Floats label above text field or select.
+`shake(shouldShake: boolean) => void` | Shakes label if shouldShake is true.
```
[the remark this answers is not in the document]
moog16: I agree I think we should automatically remove the shake class. As we chatted about offline I will try using animationend event instead of a setTimeout

packages/mdc-floating-label/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@material/floating-label",
+  "description": "The Material Components for the web floating-label component",
+  "version": "0.30.0",
```
acdvorak: `0.31.0` now 😄

packages/material-components-web/package.json
```diff
@@ -25,6 +25,7 @@
     "@material/drawer": "^0.30.0",
     "@material/elevation": "^0.28.0",
     "@material/fab": "^0.31.0",
+    "@material/floating-label": "^0.30.0",
```
acdvorak: `0.31.0` now 😄

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,106 @@
...
+  <label class="mdc-floating-label" for="my-text-field-id">Hint text</label>
+  <div class="mdc-text-field__bottom-line"></div>
+</div>
+```
+
+<!-- TODO(mattgoo): add ### Usage within `mdc-select` once select uses mdc-floating-label -->
+
+#### Avoid Dynamic ID Generation
+
+It's also possible to wrap `mdc-text-field__input` within a `<label>` to avoid dynamic id generation:
```
acdvorak:
```markdown
If you're using the JavaScript-enabled version of floating label, you can avoid needing to assign
a unique `id` to each `<input>` by wrapping `mdc-text-field__input` within a `<label>`:
```

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+```html
+<label class="mdc-text-field">
+  <input type="text" class="mdc-text-field__input">
+  <span class="mdc-floating-label">Hint Text</span>
+  <div class="mdc-text-field__bottom-line"></div>
+</label>
+```
+
+> _NOTE_: Only place a `mdc-floating-label` inside of a text field _if you plan on using
```
acdvorak: Didn't we drop support for CSS-only text fields and selects?

If so, should we delete this paragraph and the "CSS Only" section below it?

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,74 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "./variables";
+
+// Public
```
acdvorak: Probably unnecessary if there aren't any private mixins

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -27,7 +27,7 @@
 @import "@material/line-ripple/mdc-line-ripple";
 @import "./helper-text/mdc-text-field-helper-text";
 @import "./icon/mdc-text-field-icon";
-@import "./label/mdc-text-field-label";
+@import "@material/floating-label/mdc-floating-label";
```
acdvorak: Group and sort these imports (`@material` first, then package-local imports)

packages/mdc-textfield/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@material/animation": "^0.25.0",
     "@material/base": "^0.29.0",
+    "@material/floating-label": "^0.30.0",
```
acdvorak: `0.31.0` now 😄

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,106 @@
...
+`mdc-floating-label-fill-color($color)` | Customizes the fill color of the label.
+`mdc-floating-label-shake-keyframes($modifier, $positionY, $positionX, $scale)` | Generates a CSS `@keyframes` at-rule for an invalid label shake. Used in conjunction with the `mdc-floating-label-shake-animation` mixin.
+`mdc-floating-label-shake-animation($modifier)` | Applies shake keyframe animation to label.
+`mdc-floating-label-float-position($positionY, $positionX, $scale)` | Sets position of label when floating.
+
+### `MDCFloatingLabel`
+
+Method Signature | Description
+--- | ---
+`shake(shouldShake: boolean) => void` | Shakes label if shouldShake is true.
```
acdvorak: We might want to clarify a few points in the descriptions for `shake()` and `float()`:

1. They also _stop_ shaking/floating if the argument is `false`
2. They proxy straight to the foundation

We might want to rename `styleShake()` to `shake()` and `styleFloat()` to `float()`, and use the same descriptions for both.

E.g.:

```markdown
### `MDCFloatingLabel`

--- | ---
`shake(shouldShake: boolean) => void` | Shakes or stops shaking the label, depending on the value of `shouldShake`. Proxies to the foundation method of the same name.
`float(shouldFloat: boolean) => void` | Floats or docks the label, depending on the value of `shouldFloat`. Proxies to the foundation method of the same name.
`getWidth() => number` | Returns the width of the label element.

...

### `MDCFloatingLabelFoundation`

--- | ---
`shake(shouldShake: boolean) => void` | Shakes or stops shaking the label, depending on the value of `shouldShake`.
`float(shouldFloat: boolean) => void` | Floats or docks the label, depending on the value of `shouldFloat`.
`getWidth() => number` | Returns the width of the label element.
```

packages/mdc-floating-label/package.json
kfranqueiro: New packages and dependencies on them should be initialized as 0.0.0 and lerna will rev them during the next release.

See e.g. https://github.com/material-components/material-components-web/commit/1dc0e857b7f117968c99cd46f41f4ae436557fed#diff-2aab7ec5d1d3bc64e9cd3cd98216fa15R4 from when line-ripple was added.

moog16: Thanks! I'll update

packages/mdc-floating-label/README.md
moog16: @acdvorak is renaming these from styleShake to shake and styleFloat to float a breaking change?

kfranqueiro: Isn't this entire package new to this PR? You can't break something that wasn't committed/released yet.

moog16: True. I did remove the label "package" from textfield. Is that a breaking change?

kfranqueiro: It would be from the perspective of anyone wrapping our components, so it's probably worth mentioning it moved to its own separate package.

packages/mdc-floating-label/mdc-floating-label.scss
```diff
@@ -26,7 +27,7 @@
   bottom: 8px;
   left: 0;
   transform-origin: left top;
-  transition: mdc-text-field-transition(transform), mdc-text-field-transition(color);
+  transition: mdc-floating-label-transition(transform), mdc-floating-label-transition(color);
   cursor: text;
 
   // stylelint-disable plugin/selector-bem-pattern
```
acdvorak: Should the block below this comment (`@include mdc-rtl(".mdc-text-field")`) get moved to text-field?

test/unit/mdc-floating-label/mdc-floating-label-foundation.test.js
```diff
@@ -23,10 +23,12 @@ import MDCFloatingLabelFoundation from '../../../packages/mdc-floating-label/fou
 
 const {cssClasses} = MDCFloatingLabelFoundation;
 
+const setupTest = () => setupFoundationTest(MDCFloatingLabelFoundation);
+
 suite('MDCFloatingLabelFoundation');
 
 test('exports cssClasses', () => {
-  assert.isOk('cssClasses' in MDCFloatingLabelFoundation);
+  assert.deepEqual(MDCFloatingLabelFoundation.cssClasses, cssClasses);
```
acdvorak: Why does `isOk()` need to be changed to `deepEqual()`?

moog16: Ken said isOk just checks for truthy-ness, while isTrue is more strict and checks for being _true_ (== vs ===). I changed to deepEqual since this covers 2 things:

1. Existence, which the isOk statement is checking
2. and equality

packages/mdc-floating-label/mdc-floating-label.scss
moog16: Keeping it, but removing mdc-text-field. Thanks for catching

test/unit/mdc-floating-label/mdc-floating-label-foundation.test.js
acdvorak: Interesting. It looks like our other packages use a mixture of `isOk`, `isTrue`, and `deepEqual`, which seems inconsistent and a bit odd. But OK :slightly_smiling_face: 

IMO, a deep equality check is unnecessary in this case, because it's literally comparing a value... to itself:

```js
const {cssClasses} = MDCFloatingLabelFoundation;
...
assert.deepEqual(MDCFloatingLabelFoundation.cssClasses, cssClasses);
```

The object destructuring syntax above hides the fact that it's actually doing this under the hood:

```js
assert.deepEqual(
  MDCFloatingLabelFoundation.cssClasses,
  MDCFloatingLabelFoundation.cssClasses);
```

I think the `isOk` check is probably sufficient for what the test is (presumably) trying to accomplish: making sure that we remember to export `cssClasses` from the foundation.

packages/mdc-floating-label/mdc-floating-label.scss
```diff
@@ -30,13 +30,11 @@
   transition: mdc-floating-label-transition(transform), mdc-floating-label-transition(color);
   cursor: text;
 
-  // stylelint-disable plugin/selector-bem-pattern
-  @include mdc-rtl(".mdc-text-field") {
+  @include mdc-rtl() {
```
acdvorak: Nit: We normally omit empty parens when defining and including Sass mixins.

Our code base is somewhat inconsistent about that, of course, since we don't have a stylelint rule for it, but just FYI :slightly_smiling_face:

moog16: ah gotcha - thanks :)

test/unit/mdc-floating-label/mdc-floating-label-foundation.test.js
moog16: @kfranqueiro was saying he wanted to get rid of all isOk's in the test. However I do realize that last line you wrote is what it boils down to.

kfranqueiro: I suppose equal vs. deepEqual is a wash in this case, but either one protects us from accidentally e.g. copypasting cssClasses to strings and forgetting to actually change what it exports.

And yeah, I was saying to Matt yesterday that at some point when someone has too much time on their hands I'd like to abolish isOk/isNotOk entirely.

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,72 @@
...
+
+@mixin mdc-floating-label-shake-animation($modifier) {
+  .mdc-floating-label--shake {
+    animation: mdc-floating-label-shake-float-above-#{$modifier} 250ms 1;
+  }
+}
+
+// Used for textarea in case of scrolling
+@mixin mdc-floating-label-fill-color($color) {
+  .mdc-floating-label {
```
acdvorak: Remove this selector here (we'll need to add it back to wherever the mixin is used)

packages/mdc-textfield/_variables.scss
```diff
@@ -38,3 +38,14 @@ $mdc-textarea-background: rgba(white, 1);
...
+
+$mdc-text-field-box-label-position-y: 50%;
+$mdc-text-field-box-dense-label-position-y: 90%;
+$mdc-text-field-dense-label-scale: .923;
+$mdc-text-field-outlined-label-position-y: 130%;
+$mdc-text-field-outlined-dense-label-position-y: 145%;
+$mdc-text-field-outlined-with-leading-icon-label-position-x: 32px;
+$mdc-text-field-outlined-dense-with-leading-icon-label-position-x: 21px;
+$mdc-text-field-textarea-label-position-y: 50%;
+// Note that the scale factor is an eyeball'd approximation of what's shown in the mocks.
```
acdvorak: Nit: "eyeballed"

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+### `MDCFloatingLabelFoundation`
+
+Method Signature | Description
+--- | ---
+`shake(shouldShake: boolean)` | Shakes or stops shaking the label, depending on the value of `shouldShake`.*
+`float(shouldFloat: boolean)` | Floats or docks the label, depending on the value of `shouldFloat`.
+`getWidth() => number` | Returns the width of the label element.
+
+\* **N.B.:** Multiple consecutive calls to `shake(true)` will only shake the label once.
```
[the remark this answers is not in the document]
moog16: It wasn't there before and it seems like people figured out how it worked. I'll remove.

packages/mdc-textfield/_mixins.scss
```diff
@@ -101,6 +101,12 @@
   }
 }
 
+@mixin mdc-text-field-label-color($color) {
```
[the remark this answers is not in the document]
moog16: Looks like a lot of other mixins were not in the text-field readme. Adding them.

demos/text-field.scss
```diff
@@ -28,7 +28,7 @@
   @include mdc-text-field-hover-bottom-line-color($hover-border);
   @include mdc-text-field-line-ripple-color($focused-border);
   @include mdc-text-field-ink-color(black);
-  @include mdc-text-field-label-color(rgba(blue, .5));
```
acdvorak: My bad - I didn't realize this mixin already existed. Let's undo the rename since it would be a breaking change unrelated to this PR.

packages/mdc-floating-label/README.md
```diff
@@ -0,0 +1,94 @@
...
+--- | ---
+`addClass(className: string) => void` | Adds a class to the label element.
+`removeClass(className: string) => void` | Removes a class from the label element.
+`getWidth() => number` | Returns the width of the label element.
+
+### `MDCFloatingLabelFoundation`
+
+Method Signature | Description
+--- | ---
+`shake(shouldShake: boolean)` | Shakes or stops shaking the label, depending on the value of `shouldShake`.*
```
acdvorak: Nit: Remove trailing `*`

packages/mdc-floating-label/_mixins.scss
```diff
@@ -0,0 +1,68 @@
...
+}
+
+@mixin mdc-floating-label-shake-animation($modifier) {
+  .mdc-floating-label--shake {
+    animation: mdc-floating-label-shake-float-above-#{$modifier} 250ms 1;
+  }
+}
+
+// Used for textarea in case of scrolling
+@mixin mdc-floating-label-fill-color($color) {
```
acdvorak: Nit: Move this mixin just after `mdc-floating-label-ink-color`

packages/mdc-floating-label/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@material/floating-label",
+  "description": "The Material Components for the web floating-label component",
+  "version": "0.0.0",
+  "license": "Apache-2.0",
+  "keywords": [
+    "material components",
+    "material design",
+    "floatinglabel",
+    "floating-label"
```
acdvorak: Nit: Replace `-` with a space

packages/mdc-textfield/_mixins.scss
```diff
@@ -73,7 +73,9 @@
     @include mdc-text-field-textarea-fill-color_($color);
     // Automatically add label background color the same color as well to ensure
```
acdvorak: Nit: Add a blank line before this comment

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -246,12 +246,20 @@
...
 }
 
 // Show the outline when the label is floating, since the outline has a
 // notch for the label to fit into.
-.mdc-text-field__label--float-above ~ .mdc-text-field__outline {
+.mdc-floating-label--float-above ~ .mdc-text-field__outline {
   opacity: 1;
 }
+
+@include mdc-floating-label-shake-keyframes(box, $mdc-text-field-box-label-position-y);
```
acdvorak: The first parameter of all of these keyframes mixins (`$modifier`) should probably be
prefixed with `text-field-` to avoid future collisions with mdc-select.

E.g., `text-field-box`.

Other references to the keyframe names will also need to be updated.

Issue #1982, opened 2018-01-17
Split up MDCSelect into sub elements


lynnjepsen:
First split MDCSelect into sub elements. This is to help us understand all the logic currently inside MDCSelect.

Follow up PRs will replace the MDCSelect sub elements with public elements, like [line ripple](https://github.com/material-components/material-components-web/issues/1981), [floating label](https://github.com/material-components/material-components-web/issues/2126), and [notched outline](https://github.com/material-components/material-components-web/issues/2128)

lynnjepsen:
Label and bottom line are split into sub-component.

After taking a look at the adapter for select, I think it is also a good idea to split out a select-menu sub-element. 

After that is done we can close this issue

lynnjepsen:
Closing this issue because #2399 is going to remove mdc-select's dependency on mdc-menu, which means we should not bother splitting out the menu code.

PR #2244, opened 2018-02-13
select(refactor): decouple label from select

moog16:
related #1982


test/unit/mdc-select/mdc-select.test.js
```diff
@@ -173,19 +180,11 @@ test('adapter#removeClass removes a class from the root element', () => {
...
 });
 
-test('adapter#addClassToLabel adds a class to the label', () => {
-  const {component, label} = setupTest();
-
-  component.getDefaultFoundation().adapter_.addClassToLabel('foo');
-  assert.isTrue(label.classList.contains('foo'));
-});
-
-test('adapter#removeClassFromLabel removes a class from the label', () => {
```
moog16: this test will be covered in #2244 when MDCSelect uses MDCFloatingLabel

packages/mdc-select/foundation.js
```diff
@@ -153,9 +152,9 @@ export default class MDCSelectFoundation extends MDCFoundation {
     if (this.selectedIndex_ >= 0) {
       selectedTextContent = this.adapter_.getTextForOptionAtIndex(this.selectedIndex_).trim();
       this.adapter_.setAttrForOptionAtIndex(this.selectedIndex_, 'aria-selected', 'true');
-      this.adapter_.addClassToLabel(cssClasses.LABEL_FLOAT_ABOVE);
+      this.adapter_.floatLabel(true);
```
williamernest: We looked up the spec yesterday and saw that this should be moved to the open function, and line 157 should be moved to the closed function.

packages/mdc-select/mdc-select.scss
```diff
@@ -15,16 +15,14 @@
 //
 
 @import "./mixins";
+@import "./variables";
 @import "@material/animation/functions";
-@import "@material/animation/variables";
+// @import "@material/animation/variables";
```
williamernest: Probably don't need the commented out import.

packages/mdc-select/label/_mixins.scss
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "@material/theme/mixins";
+
+@mixin mdc-select-floating-label-color($color, $opacity: 1) {
```
williamernest: Should this be private ?

packages/mdc-select/README.md
```diff
@@ -280,6 +280,26 @@ The `menuFactory` function is passed an `HTMLElement` and is expected to return
 instance attached to that element. This is mostly used for testing purposes, but it's there if you
 need it nonetheless.
 
+#### Instantiating using a custom `MDCSelectLabel` component.
```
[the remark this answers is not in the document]
moog16: Will remove

packages/mdc-select/mdc-select.scss
moog16: Ah! Will remove...

packages/mdc-select/label/_mixins.scss
moog16: In the current implementation in #2237 it is public. It needs to be public for other modules to change the color. Why do you think it needs to be private?

williamernest: I thought public color mixins were supposed to explicitly prevent styling the disabled state 
`:not(.mdc-text-field--disabled)`. 

https://github.com/material-components/material-components-web/blob/3b0977d60fe6c94528bb348bff5b17edf460ff1f/packages/mdc-textfield/label/_mixins.scss#L22

That doesn't really make sense for the separate package though, so maybe it doesn't matter.

moog16: I think it doesn't make sense to have that in the public mixin, but I guess the question then is, should we allow the end developer to style the color in the disabled state?

Also in the current state being in the label module, we shouldn't add in parent classes like .mdc-text-field--disabled or .mdc-select--disabled

moog16:
Also RE: color when menu opened and --float-above, is not currently the primary color in 0.30.0. I think this should be a separate issue as that feature has never been in the code base.

packages/mdc-select/label/_mixins.scss
williamernest: I think we're explicitly preventing them from styling the disabled state. It doesn't make sense to include the parent class, we added proxy mixins from the parent to the line-ripple module to prevent customizing the disabled state.

PR #2249, opened 2018-02-14
fix(checkbox): use background-aware colors instead of hard-coded black

### LINKED ISSUE #2248: Checkboxes aren't dark-background aware

v0.30.0

Chrome 64.0.3282.167 64-bit

### What OS are you using?

Windows 10

1. Add `$mdc-theme-background: #333` (or any dark value) to `demos/common.scss`
2. Run the dev server
3. Visit localhost:8080/checkbox.html
4. Observe black checkboxes

Checkboxes becomes a light color on dark background

Checkboxes are hard-coded to use dark colors.

dpraul:
Not sure if this is a regression from when the dark theme was removed or if this is intentional.

Closes #2248

### DIFF
--- a/demos/theme/theme-dark.scss
+++ b/demos/theme/theme-dark.scss
@@ -28,10 +28,6 @@ $mdc-theme-background: $material-color-grey-900;
 
 // TODO: Call theme mixins for mdc-select
 
-.demo-checkbox {
-  @include mdc-checkbox-container-colors(text-secondary-on-dark);
-}
-
 .demo-radio {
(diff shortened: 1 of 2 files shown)


EskiMojo14:
This change is a good idea, however using text-disabled for disabled checkboxes returns the wrong opacity values.
The spec specifies 26% black on light, and 30% white on dark.
Using text-disabled returns 38% black on light and 50% white on dark.
Personally I'd suggest making a function or mixin that uses the theme functions to compute the correct value, though there may be a better solution.

PR #2277, opened 2018-02-20
feat(chips): Create mixin to customize chip margins

### LINKED ISSUE #2265: Create mixin to customize chip margins
Create a mixin `mdc-chip-set-margins` to make the space between the chips is equal to the padding of the chip set.
image

Also in `mdc-chip-set` use `display: flex; flex-wrap: wrap;` to collapse the whitespace in the markup, because it will give us much better control over spacing (so we can match the spec).

bonniezhou:
Resolves #2265.
According to spec the margins should be 8px all around (in between chips and between edge of chip and edge of chip-set).


bonniezhou:
The mixin is named `mdc-chip-set-margin` but I'm open to suggestions since "margin" made be misleading.

packages/mdc-chips/README.md
```diff
@@ -94,6 +94,7 @@ To customize the colors of any part of the chip, use the following mixins.
 
 Mixin | Description
 --- | ---
+`mdc-chip-set-margin($margin)` | Customizes the amount of space between each chip in the set
```
acdvorak: RE the word "margin": I think the design folks have a specific word for "the amount of space between stuff", but I can't remember what it is... "Gutter" maybe? I would check with them

packages/mdc-chips/chip-set/mdc-chip-set.scss
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+//
+
+@import "../mixins";
+
+.mdc-chip-set {
+  @include mdc-chip-set-margin(8px);
+
+  display: flex;
+  flex-wrap: wrap;
```
acdvorak: Can we add `box-sizing: border-box` while we're here?

packages/mdc-chips/_mixins.scss
```diff
@@ -87,3 +87,11 @@
   padding-left: $horizontal-padding-value;
   border-width: $width;
 }
+
+@mixin mdc-chip-set-margin($margin) {
+  padding: $margin / 2;
```
acdvorak: Nice! :grinning:

packages/mdc-chips/README.md
kfranqueiro: I suspect gutter is more specific to e.g. layout grid and image list, things with a more consistent gap between rows/columns. Chips are inline and can wrap so I'm not sure it'd apply there, so better ask.

bonniezhou: Design suggested "padding" which I think makes more sense than "margin" but could be equally misleading. @acdvorak suggested "spacing", which sounds good to me.

acdvorak: It sounds like design wants "padding", but both "padding" and "margin" have well-established meanings on the Web, so using either of those would likely confuse a lot of our clients.

What about "spacing"?

Issue #1810, opened 2017-12-20
Select with pre-selected option doesn't float label


pgbross:
### What MDC-Web Version are you using?

> 0.27.0

### What browser(s) is this bug affecting?

> Version 64.0.3282.24 (Official Build) beta (64-bit)

### What OS are you using?

> Fedora 27

### What are the steps to reproduce the bug?

> 1. Create mdc-select
> 2. Set one of the options as pre-selected with "aria-selected"
> 3. Observe label is not floated above selection.
> 4. If the select is clicked, then on making a choice the label is correctly floated.
>
> See [https://codepen.io/pgbross/pen/GyqJKg](https://codepen.io/pgbross/pen/GyqJKg) for demo based on MDC select demo.

### What is the expected behavior?

> If an option is pre-selected the label should float above the selection.

### What is the actual behavior?

> The label obscures the selected item.

### Any other information you believe would be useful?
The logic during initialization appears to set the selected index correctly, but the label is only set to float above in the foundation's `open_()` method so you get to see the correct selected option but the label remains in its initial position.

danielweck:
I came across this problem too, so as a workaround I added the `mdc-select__label--float-above` CSS class to the markup where the `aria-selected` attribute is set on an element with class `mdc-list-item`. For example:
```html
<div class="mdc-select__label mdc-select__label--float-above">Options</div>
...
<li class="mdc-list-item" role="option" tabindex="0" aria-selected>Option 1</li>
```

acdvorak:
Thanks for filing this issue!

Text field has the same problem, and its readme recommends exactly what @danielweck suggested: manually adding the appropriate class name to your markup.

We could do this for you in our JS (and IMO we should), but as with text field, you will get a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) if the page is rendered before the JS is loaded (which frequently happens over slow connections).

So it sounds like we need to:

- [ ] Update `MDCSelectFoundation#init()` to automatically add the `mdc-select__label--float-above` class if necessary
- [ ] Update the mdc-select readme to mention this problem, including the markup fix for FOUC

PR #2283, opened 2018-02-21
fix(select): pre-selected option correctly floats label

### LINKED ISSUE #2282: Select with pre-selected option doesn't float label (redux)
#1810 shows its face again, as #2125 appears to have been accidentally regressed by #2244 (see https://github.com/material-components/material-components-web/pull/2244/files#diff-cb9f9b1a189b68df18866209b4dd1cdfL156).

dpraul:
Closes #2282, fixing regression from #2244

### DIFF
--- a/packages/mdc-select/foundation.js
+++ b/packages/mdc-select/foundation.js
@@ -105,9 +105,6 @@ export default class MDCSelectFoundation extends MDCFoundation {
     };
     this.cancelHandler_ = () => {
       this.close_();
-      if (this.selectedIndex_ === -1) {
-        this.adapter_.floatLabel(false);
-      }
     };
   }
 
@@ -155,6 +152,9 @@ export default class MDCSelectFoundation extends MDCFoundation {
     if (this.selectedIndex_ >= 0) {
       selectedTextContent = this.adapter_.getTextForOptionAtIndex(this.selectedIndex_).trim();
       this.adapter_.setAttrForOptionAtIndex(this.selectedIndex_, 'aria-selected', 'true');
+      this.adapter_.floatLabel(true);
+    } else {
+      this.adapter_.floatLabel(false);
     }
     this.adapter_.setSelectedTextContent(selectedTextContent);
   }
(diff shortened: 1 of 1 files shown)


packages/mdc-select/foundation.js
```diff
@@ -155,6 +152,9 @@ export default class MDCSelectFoundation extends MDCFoundation {
     if (this.selectedIndex_ >= 0) {
       selectedTextContent = this.adapter_.getTextForOptionAtIndex(this.selectedIndex_).trim();
       this.adapter_.setAttrForOptionAtIndex(this.selectedIndex_, 'aria-selected', 'true');
+      this.adapter_.floatLabel(true);
```
williamernest: The label should float when the menu is opened, not only when an item is selected.

williamernest:
Thanks for submitting this PR. The label should float when the menu is opened, and not only when an item is selected.

#1810 refers to checking for a selected item in the init class and adding the `float-above` class to it then.

packages/mdc-select/foundation.js
```diff
@@ -105,9 +105,6 @@ export default class MDCSelectFoundation extends MDCFoundation {
     };
     this.cancelHandler_ = () => {
       this.close_();
-      if (this.selectedIndex_ === -1) {
```
williamernest: This is needed to have the floating label class removed when the menu is closed.

packages/mdc-select/foundation.js
```diff
@@ -212,7 +212,6 @@ export default class MDCSelectFoundation extends MDCFoundation {
     const focusIndex = this.selectedIndex_ < 0 ? 0 : this.selectedIndex_;
 
     this.setMenuStylesForOpenAtIndex_(focusIndex);
-    this.adapter_.floatLabel(true);
```
williamernest: This is needed to ensure the label floats when the menu is opened.

packages/mdc-select/foundation.js
```diff
@@ -155,6 +152,9 @@ export default class MDCSelectFoundation extends MDCFoundation {
     if (this.selectedIndex_ >= 0) {
       selectedTextContent = this.adapter_.getTextForOptionAtIndex(this.selectedIndex_).trim();
       this.adapter_.setAttrForOptionAtIndex(this.selectedIndex_, 'aria-selected', 'true');
+      this.adapter_.floatLabel(true);
+    } else {
+      this.adapter_.floatLabel(false);
```
williamernest: This should only called if the menu is not opened, so that the label is always floating if the menu is opened. 
```
if (this.adapter_.isMenuOpen()) {
    this.adapter_.floatLabel(false);
}
```

williamernest:
@dpraul Thanks for submitting this. We got this fixed in #2306 with tests to be sure we don't regress again.

PR #2286, opened 2018-02-21
fix(chips): Don't trigger ripple from trailing icon

### LINKED ISSUE #2266: Interacting with the trailing icon on a chip should not trigger a ripple

bonniezhou:
Resolves #2266.
A trailing icon should never trigger a ripple because its behaviour should always be separate from the chip's.


packages/mdc-chips/chip/foundation.js
```diff
@@ -93,6 +105,15 @@ class MDCChipFoundation extends MDCFoundation {
...
     }
   }
+
+  /**
+   * Handles an interaction event on the trailing icon element. This is used to
+   * prevent the ripple from activating on interaction with the trailing icon.
+   * @param {!Event} evt
+   */
+  handleTrailingIconInteraction_(evt) {
+    evt.stopPropagation();
```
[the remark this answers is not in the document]
bonniezhou: Good call! I updated it so the trailing icon emits a custom event.

In a later PR we'll capture this event in the chip and handle it if it's a remove icon.

bonniezhou:
Eventually having an option in ripple to ignore sub-elements sounds like a good idea.

test/unit/mdc-chips/mdc-chip.test.js
```diff
@@ -120,6 +120,18 @@ test('#adapter.notifyInteraction emits ' +
   td.verify(handler(td.matchers.anything()));
 });
 
+test('#adapter.notifyTrailingIconInteraction emits ' +
+  `${MDCChipFoundation.strings.INTERACTION_EVENT}`, () => {
```
kfranqueiro: Nit: You could omit the template string syntax here since it's already concatenated anyway

test/unit/mdc-chips/mdc-chip.foundation.test.js
```diff
@@ -101,3 +102,22 @@ test('on click, emit custom event', () => {
 
   td.verify(mockAdapter.notifyInteraction());
 });
+
+test('on click in trailing icon, emit custom event', () => {
```
kfranqueiro: You might be able to save some code here by using `test/unit/helpers.captureHandlers`. Example usage within ripple: https://github.com/material-components/material-components-web/blob/master/test/unit/mdc-ripple/foundation-activation.test.js#L27

(You can pass additional properties to the functions it returns, to be assigned to the event object.)

Also, should we actually stub `stopPropagation` and test that it gets called? (I realize right now it needs to exist simply to not throw an error in the first place, but it could also test unintended behavior changes later.)

packages/mdc-chips/README.md
```diff
@@ -141,7 +141,10 @@ Method Signature | Description
 `hasClass(className: string) => boolean` | Returns true if the root element contains the given class
 `registerInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event listener on the root element
 `deregisterInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener on the root element
+`registerTrailingIconInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event listener on the trailing icon element
+`deregisterTrailingIconInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener on the trailing icon element
 `notifyInteraction() => void` | Emits a custom event "MDCChip:interaction" denoting the chip has been interacted with, which bubbles to the parent `mdc-chip-set` element
+`notifyTrailingIconInteraction() => void` | Emits a custom event "MDCChip:trailingIconInteraction" denoting the chip's trailing icon has been interacted with, which bubbles to the parent `mdc-chip-set` element
```
kfranqueiro: Double quotes -> backticks

test/unit/mdc-chips/mdc-chip.test.js
```diff
@@ -90,6 +119,17 @@ test('#adapter.notifyInteraction emits ' +
   td.verify(handler(td.matchers.anything()));
 });
 
+test('#adapter.notifyTrailingIconInteraction emits ' + MDCChipFoundation.strings.INTERACTION_EVENT, () => {
```
kfranqueiro: INTERACTION_EVENT -> TRAILING_ICON_INTERACTION_EVENT

bonniezhou: Ah nice catch!

bonniezhou:
Thanks, fixed!

PR #2287, opened 2018-02-21
Demo index view - modal dialog secondary text

aaronhudon:
Added secondary text to the modal dialog entry


moog16:
@aaronhudon what is the purpose of this change, seeing as its just a text change?

kfranqueiro:
@moog16 this pertains to the index page for the old demo site. We now link to the new catalog everywhere, but these pages still exist for purposes of local development of existing components within MDC Web until we replace them with screenshot test pages.

@aaronhudon thanks for catching this.

Issue #2296, opened 2018-02-23
Use primary color when a chip is selected


bonniezhou:
Because the chip's opacity change is too imperceptible on selection, also include a color change when a chip is selected.
Pending spec from design.

bonniezhou:
Fixed by #2329

Issue #2295, opened 2018-02-23
Change terminology from 'activated' to 'selected' for choice/filter chips


bonniezhou:
After discussing with design, it was decided that referring to the selection state as 'selected' instead of 'activated' makes more sense for choice and filter chips.

PR #2297, opened 2018-02-23
feat(chips): Handle multi-select for filter chips

bonniezhou:
Add multiple selection behavior for filter chips.

Follow-up PRs will address #2295 and #2296.


bonniezhou:
Yep, we also have to make sure that stale chips are removed from the chip-set's `chips` array.

@kfranqueiro I'm maybe rethinking the design of this... what do you think about adding a `selected` field to each chip and a `selectedChoiceChip` field to the chip-set, rather than having the chip-set hold a `selectedChips` array? This means that querying the selected chips would require looping through the entire chip-set's `chips` array, but the flip side is that it saves memory. 
Though I'm not sure how much time/memory either option is really saving, so I think it comes down to what is best practice?

kfranqueiro:
We discussed offline and decided to stick with the selectedChips array, figuring that the number of total chips is more likely to far outnumber the number of selected chips to the point where this may be ultimately more efficient. Moreover, having the separate array may be helpful for eventually providing information on selected chips in events.

PR #2301, opened 2018-02-23
Allow toolbar variables to be customized

aaronhudon:
Unless I'm WAY off base with SCSS, this is needed for me to customize the toolbar heights when using SCSS compilation.


williamernest:
Thanks for submitting this PR. We've decided to deprecate the current toolbar implementation and recommend that developers switch to using the top-app-bar component, which is very similar.

PR #2325, opened 2018-02-28
chore: Move demo Webpack config to separate file

### LINKED ISSUE #1994: Extract webpack.demo.js from webpack.config.js

acdvorak:
Fixes #1994


package.json
```diff
@@ -9,11 +9,11 @@
     "clean": "del-cli build/** build .closure-tmp/** .closure-tmp",
     "commitmsg": "validate-commit-msg",
     "dist": "npm run build && npm run build:min",
-    "dev": "npm run clean && cross-env MDC_ENV=development webpack-dev-server --progress --content-base demos --inline --hot --host 0.0.0.0",
-    "fix:js": "eslint --fix packages test webpack.config.js karma.conf.js",
+    "dev": "npm run clean && cross-env MDC_ENV=development webpack-dev-server --progress --content-base demos --inline --hot --host 0.0.0.0 --config=demos/webpack.config.js",
```
bonniezhou: When do we set the config to `/webpack.config.js` for production?

acdvorak: The `build`, `build:min`, and `dist` commands are used for production builds, so they remain unchanged. If you don't specify a config file, Webpack implicitly uses `/webpack.config.js` by default.

`dev` is only used for demos :slightly_smiling_face:

webpack.config.js
```diff
@@ -29,80 +29,16 @@ env.setBabelEnv();
...
-const copyrightBannerPlugin = pluginFactory.createCopyrightBannerPlugin();
 const cssBundleFactory = new CssBundleFactory({env, pathResolver, globber, pluginFactory});
 const jsBundleFactory = new JsBundleFactory({env, pathResolver, globber, pluginFactory});
 
 const OUTPUT = {
   fsDirAbsolutePath: pathResolver.getAbsolutePath('./build'),
-  httpDirAbsolutePath: '/assets/',
 };
 
-if (env.isDev()) {
```
kfranqueiro: FYI this got lost in translation and I suspect this may be the real cause of what we attempted to address in #2437. I think we should revert that and re-integrate this check.

Currently in master, `isDev()` is never checked whatsoever, #2437 causes `npm run build`  to unconditionally always build both demo files and dist files, and the fact that we patched the `build` target rather than the webpack configs means this regression slipped by `test:buildconfig` unnoticed.

PR #2329, opened 2018-03-01
feat(chips): Change chip color when selected

### LINKED ISSUE #2295: Change terminology from 'activated' to 'selected' for choice/filter chips

bonniezhou:
For filter and choice chips, change its fill and ink (and ripple) color to the primary color upon selection.

Also change selection terminology for both choice and filter chips to "selected" instead of "activated", and use the regular selection ripple opacity of 8%.

Fixes #2295 and #2296.


packages/mdc-chips/_mixins.scss
```diff
@@ -42,23 +42,24 @@
...
-  @include mdc-chip-activated-ink-color($color);
   @include mdc-theme-prop(color, $color);
 
   &:hover {
     @include mdc-theme-prop(color, $color);
   }
 }
 
-@mixin mdc-chip-activated-ink-color($activated-ink-color) {
+@mixin mdc-chip-selected-color($color) {
```
acdvorak: `mdc-chip-selected-ink-color`

PR #2330, opened 2018-03-01
chore(select): Remove unused ID values from demo

williamernest:
Remove unused (and some duplicate) ID values. Fix parentheses around option in RTL mode.


moog16:
For reference the `&#x200E;` char is based off of this post: http://www.i18nguy.com/markup/right-to-left.html. It fixes parens in rtl.

Issue #2275, opened 2018-02-20
MDCSelect's options do not display


FirebertNY:
When copying the HTML from the select's documentation page and putting it into a page, clicking the select does not display the menu options. I'm uncertain if I'm using the component incorrectly or if there's an actual bug.

### What MDC Web Version are you using?

0.30.0

### What browser(s) is this bug affecting?

Chrome, Firefox, IE

### What OS are you using?

Win 7

### What are the steps to reproduce the bug?

Take a look at this codepen: https://codepen.io/anon/pen/YeaOXv
I've simply copied the HTML from the select example in the documentation and set the width on it. Clicking the select does not display the menu.

### What is the expected behavior?

Clicking the select opens the menu showing all of the options.

### What is the actual behavior?

Clicking the select does not open the menu.

mdunisch:
The documentation is really patchy here:
You can't use
`new MDCSelect` when you include the hole js-package (this syntax is only if you import each component with import-syntax)

To make this work just use 
``new mdc.select.MDCSelec`` instead.

See updated codepen: https://codepen.io/anon/pen/yvjMja

FirebertNY:
That fixed it, thank you. I agree the docs should be updated to be more clear about this.

kfranqueiro:
It looks like we're missing a link to "Importing the JS Component" the way we have in e.g. [text field](https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#mdctextfield).

I suspect our MDC Select readme eventually needs more tweaking to bring it in line with other more recently updated readmes, but at the very least we could add that line below the script example (in fact we could probably replace the note about UMD bundle with this line instead).

PR #2332, opened 2018-03-01
chore(select): Fix JS example to actually work

williamernest:
Refs: #2275 
Doesn't fix the entire README, which needs to be redone, but it does fix our JS example to actually run until we get to it.

### DIFF
--- a/packages/mdc-select/README.md
+++ b/packages/mdc-select/README.md
@@ -78,14 +78,14 @@ npm install --save @material/select
 Then with JS
 
 ```js
-import {MDCSelect} from '@material/select';
-
 const select = new mdc.select.MDCSelect(document.querySelector('.mdc-select'));
 select.listen('MDCSelect:change', () => {
   alert(`Selected "${select.selectedOptions[0].textContent}" at index ${select.selectedIndex} ` +
         `with value "${select.value}"`);
 });
 ```
+
+See [Importing the JS component](../../docs/importing-js.md) for more information on how to import JavaScript.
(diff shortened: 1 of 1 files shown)


packages/mdc-select/README.md
```diff
@@ -80,7 +80,7 @@ Then with JS
 ```js
 import {MDCSelect} from '@material/select';
```
moog16: should we also then remove this?

moog16: should we also then remove this?

williamernest: Yes, but this readme doesn't ever reference the `Importing the JS component` section, so I'll add that under this code block.

williamernest: Yes, but this readme doesn't ever reference the Importing the JS component section, so I'll add that under this code block.

packages/mdc-floating-label/mdc-floating-label.scss
```diff
@@ -29,6 +29,7 @@
   transform-origin: left top;
   transition: mdc-floating-label-transition(transform), mdc-floating-label-transition(color);
   cursor: text;
+  will-change: transform;
```
moog16: does this fix the jitter?

kfranqueiro:
This seems reasonable given that we have other READMEs that do the same, but we currently have inconsistency in whether we reference the globals or use syntax that assumes ES6 imports. We should standardize on one or the other eventually.

PR #2344, opened 2018-03-05
fix(rtl): Adding noflip annotations to fix downstream rtl issues

aprigogin:
Fixing rtl for button and ripples.


packages/mdc-button/_mixins.scss
```diff
@@ -63,8 +64,8 @@
   // Note: Adjust padding to maintain consistent width with non-stroked buttons
   $padding-value: max($mdc-button-horizontal-padding - $stroke-width, 0);
 
-  padding-right: $padding-value;
-  padding-left: $padding-value;
+  @include mdc-rtl-reflexive-property(padding, $padding-value, $padding-value);
```
kfranqueiro: Why are we doing this? This ends up adding more CSS for no gain in cases where both left and right have the same value.

packages/mdc-ripple/_mixins.scss
```diff
@@ -190,6 +196,8 @@
   &::before,
   &::after {
     top: calc(50% - #{$radius / 2});
+
```
kfranqueiro: The whitespace here fails the linter

PR #2346, opened 2018-03-05
fix(button): icon in rtl should have margin right flipped.

aprigogin:


packages/mdc-button/_mixins.scss
```diff
@@ -16,6 +16,7 @@
 
 @import "@material/elevation/mixins";
 @import "@material/ripple/mixins";
+@import "@material/rtl/mixins";
```
acdvorak: Add `rtl` to dependencies in button's `package.json`

packages/mdc-button/_mixins.scss
```diff
@@ -127,6 +128,8 @@
 }
 
 @mixin mdc-button__icon_ {
+  @include mdc-rtl-reflexive-box(margin, right, 8px);
```
acdvorak: We should probably remove the `margin-right: 8px` line below (line 136)

aprigogin: sorry, my bad.

packages/mdc-button/_mixins.scss
aprigogin: done

packages/mdc-button/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@material/elevation": "^0.28.0",
     "@material/ripple": "^0.32.0",
+    "@material/rtl": "^0.32.0",
```
kfranqueiro: Have you tried `npm install`ing this? I think the latest version of mdc-rtl is still 0.30.0 due to no changes since then.

aprigogin: I have not tried it, like i have not tried gazillion other things I could have tried... But I should have checked what rtl version there is. it is 30 not 32.

Issue #50, opened 2016-12-05
Full width search (as in MD spec)


traviskaufman:
The MDL search should behave as desceribed in the MD styleguide - please see [examples here](https://material.google.com/patterns/search.html).

I've already seen the issue #192 which has been closed. The "layout demo page" mentioned there is not available (404). I don't see there also any solution. I think we should consider reopening it.

Such component like examples shown in the MD spec would be not only "nice to have" but rather an important feature. For now an existing one is buggy (sometimes overlaps other objects when width is small). It's also limited - it doesn't give a nice user experience when write long phrases inside. My comments are about the trunk version.

It would be really great to have something like [this example](http://assets.materialup.com/uploads/01f6b92a-b1d3-4609-8325-fab8580283bd/header.gif) (an animated GIF demo).

In meantime could you please advice how to make the existing solution 100% width (I mean "search" as in examples [here](https://getmdl.io/components/index.html#layout-section))? I tried simple CSS modifications, but I'm not sure that there is some easy way to do it.

The issue #192 has been closed with annotation that it could be reopened. Could you please back to this topic?

> Copied from https://github.com/google/material-design-lite/issues/4554

traviskaufman:
Particular requests in this issue (choreographic animation between icons and search) are out of scope for us. 

Users wanting a "Full-width search" can use our full-width textfield.

PR #2348, opened 2018-03-06
Fix demos’ RTL ↔ LTR toggling in Safari

swashcap:
* **Problem:** The right-to-left ↔ left-to-right toggling on [the top-app-bar demo](http://material-components-web.appspot.com/top-app-bar.html ) is broken in Safari 11.0.3 (on macOS 10.13.3)
* **Solution:** Don’t remove the `dir` attribute on `body` in the demo: change its value to `ltr` when the “RTL” checkbox is unchecked
* **Testing:**

    1. Get the fix branch set up:
        ```shell
        git clone git@github.com:swashcap/material-components-web.git mcw2
        cd mcw2
        npm install
        npm run dev
        ```
    2. Open http://localhost:8080/top-app-bar.html in Safari
    3. Confirm the “RTL” checkbox works after checked and unchecked

### DIFF
--- a/demos/card.html
+++ b/demos/card.html
@@ -238,11 +238,7 @@
         var rtlToggle = document.getElementById('toggle-rtl');
         rtlToggle.addEventListener('change', function() {
           var docEl = document.documentElement;
-          if (rtlToggle.checked) {
-            docEl.setAttribute('dir', 'rtl');
-          } else {
-            docEl.removeAttribute('dir');
-          }
+          docEl.setAttribute('dir', rtlToggle.checked ? 'rtl' : 'ltr');
         });
 
         [].forEach.call(document.querySelectorAll('.mdc-button'), function(surface) {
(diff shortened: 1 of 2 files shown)


swashcap:
Just to elaborate, here’s the current behavior I see:

current-behavior

swashcap:
Confirmed behavior on iOS 11.2.6. Here’s a reproduction via the iPhone SE simulator:

ios-sim-whoa

swashcap:
Confirmed this still exists in Safari Tech Preview #50 (for macOS 10.13 build). Wonder if there’s a ticket for a browser bug…

swashcap:
Ran through all the demos with RTL checkboxes and found that the card demo also has the s bug. 32d6283e21ecb4f09c1271f49b6d515540f8b5e3 addresses the issue.

Before:

before

After:

after

kfranqueiro:
Thanks so much for the detailed explanation and checking through the rest of the demo pages to find other instances of the problem! This one's definitely a trip; I was able to reproduce it in isolation, too: https://codepen.io/kfranqueiro/pen/mXZNew

PR #2349, opened 2018-03-06
feat(top-app-bar): Add prominent style

### LINKED ISSUE #2378: add href attribute for A element in top-app-bar demo
in your demo here https://material-components-web.appspot.com/top-app-bar.html
```html
 <a id="right-item" class="material-icons mdc-top-app-bar__action-item" aria-label="Download" alt="Download">file_download</a>
```
to somewhat of:
```html
 <a href="javascript:void(0);" ...</a>
```
or:
```html
 <a href="#" ...</a>
```

williamernest:
fixes #2378 
Adds prominent style.


demos/top-app-bar.html
```diff
@@ -227,9 +243,13 @@ <h3>Demo Controls</h3>
             appBarEl.classList.remove('mdc-top-app-bar--short-collapsed');
             appBarEl.classList.remove('mdc-top-app-bar--short-has-action-item');
             alwaysCollapsedCheckbox.checked = false;
+          } else {
```
kfranqueiro: Nit: reverse the if/else so the if is the positive condition

demos/top-app-bar.html
```diff
@@ -238,6 +258,10 @@ <h3>Demo Controls</h3>
           appBar.destroy();
           appBar = mdc.topAppBar.MDCTopAppBar.attachTo(appBarEl);
         });
+
+        prominentCheckbox.addEventListener('change', function()  {
+          appBarEl.classList[this.checked ? 'add' : 'remove']('mdc-top-app-bar--prominent');
```
kfranqueiro: I'd point out that we're disabling the prominent checkbox when short is checked, but not vice versa... do we want these to consistently either toggle each other or disable each other? (If we make it so that prominent disables short, we shouldn't need to uncheck prominent in the short checkbox handler since it would need to be unchecked manually first)

demos/top-app-bar.scss
```diff
@@ -55,3 +55,21 @@
...
 }
+
+
+.demo-col {
+  display: flex;
+  flex: 1 1 auto;
+  flex-direction: column;
+  margin: 0 20px 20px 20px;
+
+  & > span {
```
kfranqueiro: We should really avoid element selectors... Could we use something like `<h4 class="mdc-typography--subheading2 mdc-typography--adjust-margin` for these elements instead? (And maybe use mdc-typography--title on the Demo Controls heading?)

package-lock.json
```diff
@@ -2,6 +2,16 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
+    "JSONStream": {
```
kfranqueiro: I don't think package-lock needs to be in this PR...

packages/mdc-top-app-bar/mdc-top-app-bar.scss
```diff
@@ -130,5 +130,24 @@
   }
 }
 
+.mdc-top-app-bar--prominent {
+  .mdc-top-app-bar__row {
+    min-height: $mdc-top-app-bar-prominent-row-height;
```
kfranqueiro: Why is this using min-height rather than height? Would something cause this to be even taller? And would we not be eventually interested in this collapsing to one line's height on scroll?

packages/mdc-top-app-bar/mdc-top-app-bar.scss
```diff
@@ -130,5 +130,24 @@
...
+  }
+
+  .mdc-top-app-bar__title {
+    align-self: flex-end;
+    padding-bottom: $mdc-top-app-bar-prominent-title-bottom-padding;
+  }
+
+  .mdc-top-app-bar__action-item {
```
kfranqueiro: Combine this and the next selector into one block?

demos/top-app-bar.html
williamernest: Fixed. I changed it so the prominent disabled the short checkbox and vice versa.

demos/top-app-bar.scss
```diff
@@ -55,3 +56,16 @@
 input[type="checkbox"]:disabled + label {
   opacity: .4;
 }
+
+.demo-col {
+  display: flex;
+  flex: 1 1 auto;
+  flex-direction: column;
+  margin: 0 20px 20px 20px;
```
kfranqueiro: Nit: last 20px can be omitted

packages/mdc-top-app-bar/_mixins.scss
```diff
@@ -80,6 +80,16 @@
         @include mdc-rtl-reflexive-box(padding, right, $mdc-top-app-bar-short-collapsed-right-icon-padding);
       }
     }
+
+    .mdc-top-app-bar--prominent {
+      .mdc-top-app-bar__row {
+        height: $mdc-top-app-bar-prominent-row-height;
```
kfranqueiro: So, RE my earlier comment RE min-height, I just realized that you're using `min-height` everywhere that you're setting height for `__row`, so I guess my question applies to all of them. I imagine in these cases you're intentionally overriding those, so it makes sense for them to match, I'm just not sure offhand why they use `min-height`.

kfranqueiro: Also, shouldn't this line not even be necessary if you decrease the specificity of `.mdc-top-app-bar .mdc-top-app-bar__row` to just the latter class, further above in this mixin? I don't think those first 3 selectors should need 2 classes each, since they don't have 2 classes each in the default styles.

williamernest: The first three selectors do have 2 classes each in the default styles.

packages/mdc-top-app-bar/_mixins.scss
```diff
@@ -51,25 +51,23 @@
...
+    .mdc-top-app-bar__title {
+      @include mdc-rtl-reflexive-box(padding, left, $mdc-top-app-bar-mobile-title-left-padding);
     }
 
     .mdc-top-app-bar--short {
       transition: width 200ms $mdc-animation-standard-curve-timing-function;
 
       .mdc-top-app-bar__row {
-        min-height: $mdc-top-app-bar-mobile-row-height;
+        height: $mdc-top-app-bar-mobile-row-height;
```
kfranqueiro: I suspect this rule is also not needed? Both the non-mobile short row rule and the default mobile breakpoint rule already use this value for height.

PR #2350, opened 2018-03-06
fix(list): remove pointer events from list-item child elements

### LINKED ISSUE #1845: MDCSimpleMenu has no support for menu item icons

lastest (codepen)

All

Go to codepen, click on the icon of the menu item
https://codepen.io/kmmbvnr/pen/eyReGv

Menu closed, 'MDCSimpleMenu:selected' event fired

The menu stays open, no event fired.

The issue that the foundation class assumes that menu item could contain only text - https://github.com/material-components/material-components-web/blob/a9e48682bf28f7ccfde9fdf79a97a2ecef8affbc/packages/mdc-menu/simple/foundation.js#L387L389

dpraul:
JS usage of mdc-list (in particular, in mdc-menu) relies on `MouseEvent.target` pointing to an `.mdc-list-item` child-element of the list to correctly get the selected index (see [the mdc-menu foundation](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/foundation.js#L359)). 

When decorative/positional child-elements of `.mdc-list-item` are used (i.e. `.mdc-list-item__graphic`, `.mdc-list-item__meta`, `.mdc-list-item__text`, and `.mdc-list-item__secondary-text`), they can instead become the target of `MouseEvent`, breaking the current selection logic.

Without creating more complicated selection logic, we can disable pointer-events on the child-elements of `.mdc-list-item`, ensuring `MouseEvent.target` points to the `.mdc-list-item` element, even when child-elements are used.

This also ensures ripples on `.mdc-list-item` correctly trigger when checkboxes (or other selectable items) are used within `.mdc-list-item__meta` or `.mdc-list-item__graphic`.

Fixes #1845


moog16:
@dpraul thanks for opening the PR, but it SimpleMenu is not longer a component. Also regarding issue #1845, the issue is fixed with the current Menu. Closing this PR.

PR #2351, opened 2018-03-06
added clear select via js to readme

rhroyston:
Added this to Select Readme:

#### Resetting

Select elements are cleared on page reload. To reset or clear selects via JavaScript, 
remove the `mdc-select__label--float-above` class from the select label, 
clear the text node from `mdc-select__selected-text`, 
and remove the `aria-selected` attribute from the selected option. 

```
var mySelect = document.querySelector(".mdc-select");
mySelect.querySelector(".mdc-select__label").classList.remove("mdc-select__label--float-above");
mySelect.querySelector(".mdc-select__selected-text").innerHTML = "";
mySelect.querySelector('[aria-selected="true"]').removeAttribute("aria-selected");
```

### DIFF
--- a/packages/mdc-select/README.md
+++ b/packages/mdc-select/README.md
@@ -205,6 +205,20 @@ To disable a list item, set `aria-disabled` to `"true"`, and set `tabindex` to `
 </div>
 ```
 
+#### Resetting
+
+Select elements are cleared on page reload. To reset or clear selects via JavaScript, 
+remove the `mdc-select__label--float-above` class from the select label, 
+clear the text node from `mdc-select__selected-text`, 
+and remove the `aria-selected` attribute from the selected option. 
+
+```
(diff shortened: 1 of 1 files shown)


packages/mdc-select/README.md
```diff
@@ -205,6 +205,20 @@ To disable a list item, set `aria-disabled` to `"true"`, and set `tabindex` to `
 </div>
 ```
 
+#### Resetting
```
williamernest: Thanks for submitting this PR. As of the most recent release (yesterday), you can reset the select by calling `setSelectedIndex(-1)` on the component. I overlooked adding that to the readme, but the process explained here is no longer required. 

We could change this to 
"Select elements can be reset, or cleared, by calling `setSelectedIndex(-1)`."

williamernest:
Closed due to not being required.

PR #2352, opened 2018-03-06
feat(button): Add support for SVG icons

williamernest:


packages/mdc-button/_mixins.scss
```diff
@@ -124,6 +124,16 @@
...
+
+  // stylelint-disable-next-line selector-max-type
+  svg {
+    position: relative;
+    top: 4px;
+    width: 18px;
+    height: 18px;
+    margin-right: 8px;
+    fill: currentColor;
+  }
```
moog16: reenable stylelint after the svg selector

demos/button.html
```diff
@@ -125,6 +125,10 @@ <h1 class="mdc-typography--display2">Ripple Enabled</h1>
...
...0-33 15.15-33 33.84 0 14.95 9.45 27.63 22.57 32.11 1.65.31 2.25-.74 2.25-1.63 0-.8-.03-2.93-.04-5.75-9.18 2.04-11.12-4.54-11.12-4.54-1.5-3.91-3.66-4.95-3.66-4.95-3-2.1.23-2.06.23-2.06 3.31.24 5.05 3.49 5.05 3.49 2.94 5.17 7.72 3.68 9.6 2.81.3-2.19 1.15-3.68 2.09-4.52-7.33-.85-15.03-3.76-15.03-16.72 0-3.69 1.29-6.71 3.4-9.08-.34-.86-1.47-4.3.32-8.95 0 0 2.77-.91 9.08 3.47 2.63-.75 5.46-1.13 8.26-1.14 2.8.01 5.63.39 8.26 1.14 6.3-4.38 9.07-3.47 9.07-3.47 1.8 4.66.67 8.1.33 8.95 2.11 2.37 3.39 5.39 3.39 9.08 0 13-7.72 15.86-15.07 16.7 1.19 1.04 2.24 3.11 2.24 6.27 0 4.52-.04 8.17-.04 9.28 0 .91.59 1.96 2.27 1.63 13.1-4.48 22.55-17.16 22.55-32.1.01-18.71-14.77-33.86-33-33.86"/></defs><clipPath id="b"><use xlink:href="#a" overflow="visible"/></clipPath><g clip-path="url(#b)"><defs><path id="c" d="M-1121.33-2471.17h1680v3369h-1680z"/></defs><clipPath id="d"><use xlink:href="#c" overflow="visible"/></clipPath><path clip-path="url(#d)" d="M-278.33-22.17h76v76h-76z"/></g><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm0 18c-4.41 0-8-3.59-8-8s3.59-8 8-8 8 3.59 8 8-3.59 8-8 8z"/><path d="M13 7h-2v4H7v2h4v4h2v-4h4v-2h-4z"/><path fill="none" d="M0 0h24v24H0z"/></svg>
```
moog16: this tab is off by 2 spaces

packages/mdc-button/_mixins.scss
williamernest: This is `disable-stylelint-next-line`.

williamernest:
We could use an `img` tag to reference the svg, but we can't change the color unless it's inline.

moog16:
I was thinking since you added an `img` selector. Should we have at least one instance of that in the demo?

williamernest:
@moog16 I could, but it would be black. I'm also not entirely sure we want to cover `img` in the css. I'm thinking we should just document that you should use the inline `svg` or a font.

moog16:
True...so if you document it, just remove the `img` selector from the rule?

williamernest:
Yeah, I'll get to it tomorrow morning.

packages/mdc-button/README.md
```diff
@@ -68,6 +68,11 @@ We recommend you load [Material Icons](https://material.io/icons/) from Google F
...
...0-33 15.15-33 33.84 0 14.95 9.45 27.63 22.57 32.11 1.65.31 2.25-.74 2.25-1.63 0-.8-.03-2.93-.04-5.75-9.18 2.04-11.12-4.54-11.12-4.54-1.5-3.91-3.66-4.95-3.66-4.95-3-2.1.23-2.06.23-2.06 3.31.24 5.05 3.49 5.05 3.49 2.94 5.17 7.72 3.68 9.6 2.81.3-2.19 1.15-3.68 2.09-4.52-7.33-.85-15.03-3.76-15.03-16.72 0-3.69 1.29-6.71 3.4-9.08-.34-.86-1.47-4.3.32-8.95 0 0 2.77-.91 9.08 3.47 2.63-.75 5.46-1.13 8.26-1.14 2.8.01 5.63.39 8.26 1.14 6.3-4.38 9.07-3.47 9.07-3.47 1.8 4.66.67 8.1.33 8.95 2.11 2.37 3.39 5.39 3.39 9.08 0 13-7.72 15.86-15.07 16.7 1.19 1.04 2.24 3.11 2.24 6.27 0 4.52-.04 8.17-.04 9.28 0 .91.59 1.96 2.27 1.63 13.1-4.48 22.55-17.16 22.55-32.1.01-18.71-14.77-33.86-33-33.86"/></defs><clipPath id="b"><use xlink:href="#a" overflow="visible"/></clipPath><g clip-path="url(#b)"><defs><path id="c" d="M-1121.33-2471.17h1680v3369h-1680z"/></defs><clipPath id="d"><use xlink:href="#c" overflow="visible"/></clipPath><path clip-path="url(#d)" d="M-278.33-22.17h76v76h-76z"/></g><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm0 18c-4.41 0-8-3.59-8-8s3.59-8 8-8 8 3.59 8 8-3.59 8-8 8z"/><path d="M13 7h-2v4H7v2h4v4h2v-4h4v-2h-4z"/><path fill="none" d="M0 0h24v24H0z"/></svg>
```
moog16: I know people like to copy paste the README, but do you think its necessary to have this specific one? Seems large. If Ken is ok with it, then I'm cool with it too.

packages/mdc-button/README.md
```diff
@@ -55,7 +55,7 @@ semantically correct.
 ### Adding Icon
 
 Users can nest `mdc-button__icon` inside the button element to add an icon. The icon in button
-is set to 18px to meet legibility requirements.
+is set to 18px to meet legibility requirements.
```
moog16: Looks like this can be removed

demos/button.html
```diff
@@ -96,7 +96,7 @@
           <div class="mdc-checkbox">
             <input type="checkbox" class="mdc-checkbox__native-control" id="toggle-disabled" aria-labelledby="toggle-disabled-label">
             <div class="mdc-checkbox__background">
-              <svg class="mdc-checkbox__checkmark" viewBox="0 0 24 24">
+              <svg class="mdc-button__icon" class="mdc-checkbox__checkmark" viewBox="0 0 24 24">
```
moog16: You have 2 class attributes on this svg.

packages/mdc-button/README.md
williamernest: How about this one https://github.com/material-components/material-components-web/blob/master/demos/images/material.svg ?

moog16: Ya seems more reasonable. I can't believe that one is a smaller path than the plus sign.

packages/mdc-button/mdc-button.scss
```diff
@@ -34,6 +34,10 @@
   .mdc-button__icon {
     @include mdc-button__icon_;
   }
+  // stylelint-disable-next-line selector-no-qualifying-type
```
kfranqueiro: Add newline before this line

packages/mdc-button/README.md
```diff
@@ -68,6 +68,14 @@ We recommend you load [Material Icons](https://material.io/icons/) from Google F
   <i class="material-icons mdc-button__icon">favorite</i>
   Button
 </button>
+
```
kfranqueiro: I would split this out to its own code block and introduce it with a paragraph e.g. "It's also possible to use an SVG icon:"

Also, RE the actual SVG code, we could potentially hollow it out to only the parts that matter for purposes of the README, e.g.:

```html
<button class="mdc-button">
  <svg class="mdc-button__icon" xmlns="http://www.w3.org/2000/svg" viewBox="...">
    ...
  </svg>
  Button
</button>
```

(In particular, I'm not sure why `fill="#000000"` is in there right now, wouldn't that override our CSS??)

williamernest: Will do. 

`fill=...` is an attribute style, and not inline styling (`style="fill:...`). CSS styles always override attribute styles, because `fill=...` is converted to a css style with a specificity of 0.

packages/mdc-button/README.md
```diff
@@ -70,6 +70,17 @@ We recommend you load [Material Icons](https://material.io/icons/) from Google F
 </button>
 ```
 
+It's also possible to use an SVG icon:
+
+```html
+<button class="mdc-button">
+  <svg class="mdc-button__icon" xmlns="http://www.w3.org/2000/svg" viewBox="...">
```
moog16: I think the viewBox attr is pretty important. Can you add something to highlight that? If it isn't `0 0 24 24` then people need to go to the demo and figure that out.

kfranqueiro: IIUC, viewBox is dependent on the particular SVG, which is why I specifically suggested leaving it in, but not prescribing what its values are. The SVGs provided by Material Icons might all be in a 24x24 space, but that's not necessarily true of someone's custom icons.

moog16: Sorry - this will actually be different in every svg. Please ignore

Issue #2015, opened 2018-01-20
Text Field floating label is not clickable on standard variant


acdvorak:
### What MDC Web Version are you using?

0.28.0

### What browser(s) is this bug affecting?

All (tested in Chrome)

### What OS are you using?

macOS High Sierra 10.13

### What are the steps to reproduce the bug?

1. View the [text field demo page](https://material-components-web.appspot.com/text-field.html) in any browser
2. Enter some text in a standard text field (i.e., not `--box` or `--outlined`)
3. Click on the floating label

### What is the expected behavior?

The text field input should become focused when the floating label is clicked.

### What is the actual behavior?

Clicking on the floating label does not focus the input. If the input was previously focused, clicking on the label will blur it.

![Screen capture](https://user-images.githubusercontent.com/409245/35181153-7827fb16-fd71-11e7-9670-aa462ed81c6a.gif)

### Any other information you believe would be useful?

The bug appears to be caused by the fact that the `__native-input` element is not the full height of the text field (unlike `--box` and `--outlined`):

![Screenshot of box model](https://user-images.githubusercontent.com/409245/35181225-9af9a6f2-fd72-11e7-9ee9-3bdf47ddcc84.png)

theenoahmason:
@acdvorak I can confirm problem and cause.

An example to fix this would be to add a `padding-top` of 20-24px to the input. The reason is that currently, on the floating label, there is a rule of `pointer-events: none;` which is killing the default  behavior of the HTML `for` attribute on the label. This means that clicking the label is the same as clicking away from the input. If `pointer-events: none` is kept, then a new click handler for the label must be added, **or**  the padding at the top of the input _must reach the full height of the label_.

bonniezhou:
Our designers confirmed that the label should indeed be clickable. It's added to the tracker.

PR #2353, opened 2018-03-06
fix(text-field): Clicking label should focus textfield

williamernest:
Fixes : #2015 
Makes the label work like a standard label. If the `for` attribute is specified, clicking the label will focus the corresponding `input`.

### DIFF
--- a/packages/mdc-textfield/mdc-text-field.scss
+++ b/packages/mdc-textfield/mdc-text-field.scss
@@ -60,7 +60,7 @@
   }
 
   width: 100%;
-  padding: 0 0 8px;
+  padding: 20px 0 8px;
   transition: mdc-text-field-transition(opacity);
   border: none;
   border-bottom: 1px solid;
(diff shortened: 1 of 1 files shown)


kfranqueiro:
FYI this will probably make press ripples seem flaky but we're planning to remove that from text field anyway; I had neglected to file an issue before but I just entered #2354 now.

williamernest:
I don't think they're flaky. There's really no discernible difference that I can tell other than clicking the top half of the outlined label doesn't cause a ripple. It behaves the same way that tabbing into the field does.

kfranqueiro:
Huh. I was thinking the label would block the input (which the ripple listens on) from receiving pointer events since it's a sibling, but I guess it doesn't?

williamernest:
The original click event isn't blocked by the label.

moog16:
Yes I can confirm that behavior, which also works on FF. After chatting with Will, that behavior isn't documented in the HTML spec. Change looks good to me.

Issue #2335, opened 2018-03-02
Customize the color of the MDC button text label separately from the icon


lynnjepsen:
There should be support for customizing the either text label or icon. Right now there is just one `mdc-button-ink-color` mixin which changes both the icon and text label. 

- [ ] Create a new `mdc-button-icon-color` mixin
- [ ] Create a new `mdc-button-text-label-color` mixin

Note (by Ken): leave `mdc-button-ink-color` alone for the most common case where both colors match; it will result in more concise CSS as-is than if it were to use the two more specifically-tailored mixins.

williamernest:
I think only one mixin is needed. The icons inherit the color to begin with, so the original `mdc-button-ink-color` works for both cases. Since we don't wrap the text in it's own element, we don't have a good way of just selecting the text without also changing the color of the icon.

kfranqueiro:
Just to clarify, do you mean only one additional mixin is needed (for styling the icon specifically)? That makes sense.

williamernest:
Exactly, one additional mixin.

PR #2362, opened 2018-03-08
feat(button): Add ability to color icons separately from the text

williamernest:
Fixes: #2335
I added the `mdc-button-icon-color` mixin. The other mixin wasn't needed, since the icons use the text colors by default.


demos/button.html
```diff
@@ -306,6 +306,20 @@ <h1 class="mdc-typography--display2">CSS Only</h1>
             </button>
           </div>
         </fieldset>
+
+        <fieldset>
+          <legend class="mdc-typography--title">Different Color Icons/Text</legend>
+          <div>
+            <button class="mdc-button mdc-button--unelevated demo-text-color" data-demo-no-js>
+              <i class="material-icons mdc-button__icon">favorite</i>
+              Big Corner Radius
```
kfranqueiro: Update the labels on these buttons; these labels were copied from the previous two examples but don't make sense here.

demos/button.html
```diff
@@ -306,6 +306,20 @@ <h1 class="mdc-typography--display2">CSS Only</h1>
             </button>
           </div>
         </fieldset>
+
+        <fieldset>
+          <legend class="mdc-typography--title">Different Color Icons/Text</legend>
+          <div>
+            <button class="mdc-button mdc-button--unelevated demo-text-color" data-demo-no-js>
```
bonniezhou: Should we include a demo for JS-upgraded buttons too?

demos/button.scss
```diff
@@ -44,3 +44,12 @@
 .mdc-button.thick-stroke-button {
   @include mdc-button-stroke-width(4px);
 }
+
+.demo-text-color {
+  @include mdc-button-icon-color(white);
+  @include mdc-button-ink-color(orange);
```
kfranqueiro: Should we also be applying mdc-states(orange) here? (I'm assuming that should still prioritize matching text over matching icon? Maybe we should ask design, if text/icon can be different colors, which color the "ink" and therefore the ripple is?)

demos/button.html
```diff
@@ -234,6 +234,20 @@ <h1 class="mdc-typography--display2">Ripple Enabled</h1>
           </div>
         </fieldset>
 
+        <fieldset>
+          <legend class="mdc-typography--title">Different Color Icons/Text</legend>
```
kfranqueiro: Also replace Text w/ Ink here

demos/button.html
```diff
@@ -363,6 +377,20 @@ <h1 class="mdc-typography--display2">CSS Only</h1>
             </button>
           </div>
         </fieldset>
+
+        <fieldset>
+          <legend class="mdc-typography--title">Different Color Icons/Text</legend>
```
kfranqueiro: and here

PR #2367, opened 2018-03-08
feat(image-list): Add base styles and mixins for Standard Image List

### LINKED ISSUE #2314: Implement non-actionable Standard Image List in a new mdc-image-list
- [ ] new Node module: mdc-image-list
- [ ] Document and handle HTML structure for an image-list-item without `<a>`
- [ ] Document and handle HTML structure for text label underneath image
- [ ] Document and handle HTML structure for text label overlaid on top of image
- [ ] Add `mdc-image-list-aspect` mixin for maintaining aspect ratio
- [ ] Add `mdc-image-list-columns` mixin, which styles a Standard Image List to scale its images based on the natural width of the containing element

Follow up PRs will add ripples to the actionable parts of an mdc-image-list

kfranqueiro:
Resolves #2314.

Work on actionable images and iconography will be done separately.


packages/mdc-image-list/mdc-image-list.scss
```diff
@@ -0,0 +1,85 @@
...
+
+@import "./mixins";
+@import "./variables";
+@import "@material/theme/mixins";
+@import "@material/typography/mixins";
+
+.mdc-image-list {
+  display: flex;
+  flex-wrap: wrap;
+  margin: 0 auto; // UA override
```
bonniezhou: What's UA override?

packages/mdc-image-list/mdc-image-list.scss
```diff
@@ -0,0 +1,85 @@
...
+  left: 0;
+  height: 100%;
+  // Background styles to support div instead of img
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: cover;
+}
+
+@at-root {
+  @include mdc-image-list-aspect(1);
```
bonniezhou: Why is this under `@at-root` instead of under `.mdc-image-list`?

packages/mdc-image-list/README.md
```diff
@@ -0,0 +1,138 @@
...
+### CSS Classes
+
+CSS Class | Description
+--- | ---
+`mdc-image-list` | Mandatory. Indicates the root Image List element.
+`mdc-image-list__item` | Mandatory. Indicates each item in an Image List.
+`mdc-image-list__image-aspect-container` | Optional. Parent of each item's image element, responsible for constraining aspect ratio. This element may be omitted entirely if images are alreadysized to the correct aspect ratio.
+`mdc-image-list__image` | Mandatory. Indicates the image element in each item.
+`mdc-image-list__supporting` | Optional. Indicates the area within each item containing the supporting text label, if the Image List contains text labels.
+`mdc-image-list__label` | Optional. Indicates the text label in each item, if the Image List contains text labels.
```
bonniezhou: Add row for `mdc-image-list--protected`.

Also I think we should either rename it to `mdc-image-list--protected-text`, or rename it to `mdc-image-list__supporting--protected` and apply it on the `mdc-image-list__supporting` element.

packages/mdc-image-list/mdc-image-list.scss
```diff
@@ -0,0 +1,85 @@
...
+
+// Modifier for labels/icons with text protection, overlaying images.
+
+.mdc-image-list--protected .mdc-image-list__supporting {
+  position: absolute;
+  bottom: 0;
+  width: 100%;
+  height: 48px;
+  padding: 0 16px;
+  background: rgba(0, 0, 0, .6);
```
bonniezhou: Should we make this a constant if it's the baseline default?

packages/mdc-image-list/mdc-image-list.scss
kfranqueiro: User agent override. As in, this is overriding user agent styles. I can make this comment clearer, it was carried over from my prototype.

packages/mdc-image-list/mdc-image-list.scss
kfranqueiro: Because it already creates a 1-class selector and there's no reason to make it a 2-class selector. Similar to instances where we moved mixin invocations in other packages' base styles to `@at-root` to decrease specificity and take advantage of BEM.

packages/mdc-image-list/README.md
kfranqueiro: Haha, whoops, yeah. I ended up realizing `protected` was missing and adding it before I saw this comment, which probably came in around the same time.

packages/mdc-image-list/mdc-image-list.scss
kfranqueiro: Yeah, I realized a couple of things could probably be constants, so I'll get on that.

Issue #2371, opened 2018-03-09
look how ugly the hamburger button sticks to the title in drawer demo


englishextra:
I have to add that in a few recent days something went wrong with the CSS
look how ugly the hamburger button sticks to the title in drawer demo
```html
<div class="mdc-toolbar__row">
  <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
    <button class="demo-menu material-icons mdc-toolbar__menu-icon">menu</button>
    <span class="mdc-toolbar__title catalog-title">Temporary Drawer</span>
  </section>
</div>
```
https://material-components-web.appspot.com/drawer/temporary-drawer.html
Mozilla/5.0 (Windows NT 10.0; WOW64; rv:51.0) Gecko/20100101 Firefox/51.0
Windows 10

englishextra:
BTW 
I tested https://www.jsdelivr.com/package/npm/material-components-web?path=dist
version 0.31.0 - looks nice
version 0.32.0 - looks bad

@kfranqueiro Oh yes. I confirm that in 0.31.0 button looks nice but changing to A element the gap becomes larger
in version 0.32.0 keeping BUTTON elemnt as menu button makes no gap at all
and changing BUTTON to A - creates nice single default gap

PR #2372, opened 2018-03-09
fix(drawer): Update menu icon to be anchor element

williamernest:
Fixes: #2371


demos/drawer/persistent-drawer.html
```diff
@@ -141,6 +141,7 @@ <h1 class="mdc-typography--display1">Persistent Drawer</h1>
         <button type="button" class="mdc-button mdc-button--stroked mdc-button--dense demo-toolbar-example-heading__rtl-toggle-button">Toggle RTL</button>
       </main>
 
+      <script src="/assets/common.js" async></script>
```
kfranqueiro: Nit: I think we'd been standardizing on putting common.js after mcw.js (I realize now I missed this in top-app-bar).

englishextra:
Thanks

PR #2389, opened 2018-03-12
fix(checkbox): added aria-checked=mixed to indeterminate state

aprigogin:


packages/mdc-checkbox/foundation.js
```diff
@@ -203,6 +211,10 @@ class MDCCheckboxFoundation extends MDCFoundation {
       return;
     }
 
+    // Remove aria-checked (needed for initial inditerminate state) - screen
+    // readers will pick the right state from input.
+    this.adapter_.removeNativeControlAttr(strings.ARIA_CHECKED_ATTR);
```
patrickrodee: Is this a possible race condition between when the [property change callback gets called](https://github.com/material-components/material-components-web/pull/2389/files#diff-6b00952d4ec9f05fa873ace8d7b274a2R112) and when the [attribute gets added](https://github.com/material-components/material-components-web/pull/2389/files#diff-6b00952d4ec9f05fa873ace8d7b274a2R114)?

If so, the attr settter could be moved inside this (`transitionCheckState_`) method with a check made to `newState` to determine whether to add or remove.

PR #2397, opened 2018-03-14
refactor(chips): Manage chip foundations instead of chips in the chip set foundation

bonniezhou:
`MDCChipSetFoundation` should not touch `MDCChip` objects in case `MDCChip` is wrapped by a framework. See discussion on #2391.

We're also ensuring that custom events are emitted with the chip foundation instead of the chip component. There should be no logic in the chip component.


packages/mdc-chips/chip-set/foundation.js
```diff
@@ -57,7 +56,7 @@ class MDCChipSetFoundation extends MDCFoundation {
 
     /**
      * The selected chips in the set. Only used for choice chip set or filter chip set.
-     * @private {!Array<!MDCChip>}
+     * @private {!Array<!MDCChipFoundation>}
      */
     this.selectedChips_ = [];
```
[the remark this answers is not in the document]
bonniezhou: Changed the name of the event detail to make it more obvious that it's passing along the chip foundation, not the chip.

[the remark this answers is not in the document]
kfranqueiro: Is there any way for a user to get meaningful information as to which chips are selected?

i.e. if someone is instantiating components, and they get foundation instances back, they don't really have a simple way of tying it back to the components they actually have a handle on, do they?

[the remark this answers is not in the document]
bonniezhou: Discussed offline: for the purpose of end-users using the events from `notifyInteraction()` we should include the full chip component in the event details, but we'll expose a `get foundation()` method in the chip so `handleChipInteraction_()` can track the chip foundations.

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -82,24 +81,25 @@ class MDCChipSetFoundation extends MDCFoundation {
    */
   handleChipInteraction_(evt) {
     const {chip} = evt.detail;
+    const chipFoundation = chip.foundation;
```
acdvorak: Should we document this requirement in the README?

Wrapper libraries will probably need to know about it.

packages/mdc-chips/README.md
```diff
@@ -182,8 +182,10 @@ Method Signature | Description
...
 `registerTrailingIconInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event listener on the trailing icon element
 `deregisterTrailingIconInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener on the trailing icon element
-`notifyInteraction() => void` | Emits a custom event `MDCChip:interaction` denoting the chip has been interacted with, which bubbles to the parent `mdc-chip-set` element
-`notifyTrailingIconInteraction() => void` | Emits a custom event `MDCChip:trailingIconInteraction` denoting the chip's trailing icon has been interacted with, which bubbles to the parent `mdc-chip-set` element
+`notifyInteraction() => void` | Emits a custom event `MDCChip:interaction` denoting the chip has been interacted with
+`notifyTrailingIconInteraction() => void` | Emits a custom event `MDCChip:trailingIconInteraction` denoting the chip's trailing icon has been interacted with
+
+> _NOTE_: The custom events emitted by `notifyInteraction` and `notifyTrailingIconInteraction` must pass along the target chip in its event details, as well as bubble to the parent `mdc-chip-set` element.
```
acdvorak: Nit:

```
... in its event `detail`, as well as ...
```

packages/mdc-chips/chip-set/foundation.js
kfranqueiro: Reduce these two lines to `const chipFoundation = evt.detail.chip.foundation;` ?

packages/mdc-chips/chip/index.js
```diff
@@ -92,7 +99,8 @@ class MDCChip extends MDCComponent {
           trailingIconEl.removeEventListener(evtType, handler);
         }
       },
-      notifyInteraction: () => this.emit(strings.INTERACTION_EVENT, {chip: this}, true /* shouldBubble */),
+      notifyInteraction: () => this.emit(
```
kfranqueiro: Did these two lines not actually change, ultimately? If so, revert to one line?

PR #2406, opened 2018-03-14
refactor(ripple): Add optional width/height params to layout

### LINKED ISSUE #2334: Add optional width/height parameters to MDCRipple.layout()
For JS-upgraded components, properly size the ripple when its surface width/height is changing in the middle of a ripple activation. 

This is required for the ripple during a chip's width animation (#2320).

bonniezhou:
Pass in width and/or height to `layout()` for ripple surfaces that are JS-upgraded and change size during runtime, so we can update `--mdc-ripple-fg-size` in JS. 

The specific use case this is needed for is in filter chips, but I could extract that out into a separate PR.

Fixes #2334.

### DIFF
--- a/packages/mdc-chips/chip/index.js
+++ b/packages/mdc-chips/chip/index.js
@@ -51,15 +51,15 @@ class MDCChip extends MDCComponent {
     this.leadingIcon_ = this.root_.querySelector(strings.LEADING_ICON_SELECTOR);
     this.ripple_ = new MDCRipple(this.root_);
 
-    // Adjust ripple size for chips with animated growing width.This applies when filter chips without
+    // Adjust ripple size for chips with animated growing width. This applies when filter chips without
     // a leading icon are selected, and a leading checkmark will cause the chip width to expand.
     const checkmarkEl = this.root_.querySelector(strings.CHECKMARK_SELECTOR);
     if (!!checkmarkEl && !this.leadingIcon_) {
-      // The checkmark width is initially set to 0, so use the chip's height as a proxy, since the
+      const height = this.root_.getBoundingClientRect().height;
+      // The checkmark width is initially set to 0, so use the chip's height as a proxy since the
       // checkmark should always be square.
-      const checkmarkWidthProxy = this.root_.getBoundingClientRect().height;
-      const width = this.root_.getBoundingClientRect().width + checkmarkWidthProxy;
-      this.ripple_.layout(width);
+      const width = this.root_.getBoundingClientRect().width + height;
+      this.ripple_.layout({height: height, width: width});
     }
   }
 
--- a/packages/mdc-chips/chip/index.js
+++ b/packages/mdc-chips/chip/index.js
@@ -34,9 +34,9 @@ class MDCChip extends MDCComponent {
     super(...args);
 
     /** @private {?Element} */
-    this.leadingIcon_ = this.root_.querySelector(strings.LEADING_ICON_SELECTOR);
+    this.leadingIcon_;
     /** @private {!MDCRipple} */
-    this.ripple_ = new MDCRipple(this.root_);
+    this.ripple_;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/mdc-ripple/foundation.js
```diff
@@ -488,20 +488,31 @@ class MDCRippleFoundation extends MDCFoundation {
...
   }
 
-  /** @private */
-  layoutInternal_() {
+  /**
+   * @param {number=} width
+   * @param {number=} height
+   * @private
+   */
+  layoutInternal_(width, height) {
```
kfranqueiro: I suspect that we should accept a single object that includes a width and height, and assign that to `this.frame_` when present instead of calling `computeBoundingRect`.

* `this.frame_` is also used in `updateLayoutCssVars_` for unbounded ripples, and in `getFgTranslationCoordinates_` for ripples activated via keyboard, and we would want the updated values to be used in those functions as well. (This also suggests to me that we need more tests to verify that the correct values are used when layout is called with an argument...)
* `layout` should really be called either with both width and height, or neither, and in the case where the argument is passed, we can save the browser from needing to layout/reflow.

packages/mdc-chips/chip/index.js
```diff
@@ -47,6 +47,22 @@ class MDCChip extends MDCComponent {
...
+
+    // Adjust ripple size for chips with animated growing width. This applies when filter chips without
+    // a leading icon are selected, and a leading checkmark will cause the chip width to expand.
+    const checkmarkEl = this.root_.querySelector(strings.CHECKMARK_SELECTOR);
+    if (!!checkmarkEl && !this.leadingIcon_) {
+      const height = this.root_.getBoundingClientRect().height;
+      // The checkmark width is initially set to 0, so use the chip's height as a proxy since the
+      // checkmark should always be square.
+      const width = this.root_.getBoundingClientRect().width + height;
+      this.ripple_.layout({height: height, width: width});
```
kfranqueiro: I think this can be abbreviated to `{height, width}`

packages/mdc-chips/chip/index.js
```diff
@@ -47,6 +47,22 @@ class MDCChip extends MDCComponent {
...
+  initialize() {
+    this.leadingIcon_ = this.root_.querySelector(strings.LEADING_ICON_SELECTOR);
+    this.ripple_ = new MDCRipple(this.root_);
+
+    // Adjust ripple size for chips with animated growing width. This applies when filter chips without
+    // a leading icon are selected, and a leading checkmark will cause the chip width to expand.
+    const checkmarkEl = this.root_.querySelector(strings.CHECKMARK_SELECTOR);
+    if (!!checkmarkEl && !this.leadingIcon_) {
+      const height = this.root_.getBoundingClientRect().height;
+      // The checkmark width is initially set to 0, so use the chip's height as a proxy since the
```
kfranqueiro: If only the checkmark's width is set to 0, shouldn't this use the checkmark's height rather than the chip's? Wouldn't the chip be taller than the checkmark?

test/unit/mdc-ripple/mdc-ripple.test.js
```diff
@@ -92,7 +92,14 @@ test('layout() delegates to the foundation', () => {
...
   component.foundation_.layout = td.function();
   component.layout();
-  td.verify(component.foundation_.layout());
+  td.verify(component.foundation_.layout(undefined));
+});
+
+test('layout() passes value of dimensions to the foundation when set', () => {
+  const {component} = setupTest();
+  component.foundation_.layout = td.function();
+  component.layout({height: '10px', width: '10px'});
```
kfranqueiro: These would be called with numbers, not strings

kfranqueiro:
Oh. Also, this is technically a `feat` I guess, it's adding parameters to a public API.

bonniezhou:
Closing in favour of an alternative solution in #2423.

PR #2408, opened 2018-03-15
chore: update node-sass, now that it works with node 8

lynnjepsen:


package.json
```diff
@@ -73,7 +73,7 @@
     "lolex": "2.1.3",
     "mkdirp": "^0.5.1",
     "mocha": "^5.0.0",
-    "node-sass": "^4.0.0",
+    "node-sass": "^4.7.2",
```
kfranqueiro: Update package-lock?

Also, is there a reason this is set at 4.7.2 rather than the latest it will match (4.8.2)?

patrickrodee: I updated package-lock. 4.7.2 is the most recent release of node-sass on NPM

kfranqueiro: Oh. Yeah just double-checked and 4.8.0 and 4.8.2 on github are marked as pre-release. 4.8.1 oddly isn't. Dunno what's up there.

patrickrodee: All good! Thanks man 👍

PR #2412, opened 2018-03-15
feat(top-app-bar): Switch to use variant specific foundations

williamernest:


packages/mdc-top-app-bar/index.js
```diff
@@ -70,33 +71,41 @@ class MDCTopAppBar extends MDCComponent {
...
+      registerScrollHandler: (handler) => window.addEventListener('scroll', handler, util.applyPassive()),
+      deregisterScrollHandler: (handler) => window.removeEventListener('scroll', handler),
+      getViewportScrollY: () => window.pageYOffset,
+      getTotalActionItems: () =>
+        this.root_.querySelectorAll(strings.ACTION_ITEM_SELECTOR).length,
+    })
     );
+
+    let foundation;
+    if (this.root_.classList.contains(cssClasses.SHORT_CLASS)) {
```
kfranqueiro: Have you run this by Lynn? Given our desire to avoid putting business logic on the component level, this seems to go against that, but I suppose having different components for each type would be overkill...

williamernest: We talked a bit about it over the past couple days. She asked me to create this PR so she could go over how I broke it up and start the discussion. I'm not sure where this block of code would really belong we don't want it in the Component.

lynnjepsen: Actually I think this might be exactly the kind of logic we want to change per web platform

- Vanilla JavaScript: Read the DOM to decide which type of foundation to use
- React: Read a React prop to decide which type of foundation to use
- Internal Google JavaScript: Read some internal property available at initiate-time

packages/mdc-top-app-bar/short/foundation.js
```diff
@@ -0,0 +1,77 @@
...
+  constructor(adapter) {
+    super(adapter);
+    // State variable for the current top app bar state
+    this.isCollapsed = false;
+
+    this.scrollHandler_ = () => this.shortAppBarScrollHandler_();
+  }
+
+  init() {
+    const isAlwaysCollapsed = this.adapter_.hasClass(cssClasses.SHORT_COLLAPSED_CLASS);
```
lynnjepsen: I think you could call super.initi() and then not have to register the scroll handler

packages/mdc-top-app-bar/short/foundation.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    if (!isAlwaysCollapsed) {
+      this.adapter_.registerScrollHandler(this.scrollHandler_);
+      this.shortAppBarScrollHandler_();
+    }
+  }
+
+  destroy() {
+    this.adapter_.deregisterScrollHandler(this.scrollHandler_);
```
lynnjepsen: super.destroy();

PR #2413, opened 2018-03-15
fix(button): add padding mixin, use it, updated demo.

aprigogin:


packages/mdc-button/_mixins.scss
```diff
@@ -60,6 +60,10 @@
   border-radius: $corner-radius;
 }
 
+@mixin mdc-button-horizontal-padding($left, $right: $left) {
+  @include mdc-rtl-reflexive-property(padding, $left, $right);
```
kfranqueiro: Don't use reflexive-property if both values are the same, it creates excess styles.

demos/button.html
```diff
@@ -118,20 +118,16 @@ <h1 class="mdc-typography--display2">Ripple Enabled</h1>
...
+              <i class="material-icons mdc-button__icon">favorite</i>Icon
             </button>
-            <button class="mdc-button">
+            <button class="mdc-button mdc-button--with-icon">
               <svg class="mdc-button__icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="#000000">
                 <path fill="none" d="M0 0h24v24H0z"/>
                 <path d="M23 12c0-6.07-4.93-11-11-11S1 5.93 1 12s4.93 11 11 11 11-4.93 11-11zM5 17.64C3.75 16.1 3 14.14 3 12c0-2.13.76-4.08 2-5.63v11.27zM17.64 5H6.36C7.9 3.75 9.86 3 12 3s4.1.75 5.64 2zM12 14.53L8.24 7h7.53L12 14.53zM17 9v8h-4l4-8zm-6 8H7V9l4 8zm6.64 2c-1.55 1.25-3.51 2-5.64 2s-4.1-.75-5.64-2h11.28zM21 12c0 2.14-.75 4.1-2 5.64V6.37c1.24 1.55 2 3.5 2 5.63z"/>
-              </svg>
-              SVG Icon
+              </svg>SVG Icon
```
kfranqueiro: Revert the SVG icon changes; if there's an issue here we should log it and investigate possible solutions (I'd hope to avoid needing to cram the markup like this), but it's separate from the side padding.

demos/button.html
```diff
@@ -118,20 +118,16 @@ <h1 class="mdc-typography--display2">Ripple Enabled</h1>
...
             </button>
-            <a href="javascript:void(0)" class="mdc-button">
-              Link
-            </a>
+            <a href="javascript:void(0)" class="mdc-button">Link</a>
```
kfranqueiro: Is this fixing anything?

packages/mdc-button/_mixins.scss
```diff
@@ -60,6 +60,10 @@
   border-radius: $corner-radius;
 }
 
+@mixin mdc-button-horizontal-padding($left, $right: $left) {
```
kfranqueiro: This mixin should only need to be passed a single value that will be used for both margins.

The current implementation optimizes for certain specific icons, but causes other icons to become off-center in relation to the button's padding (such as both icons used in the demo page!). Icons are assumed to occupy a 18x18 space; whether the actual stroke/fill of the icon does is beyond our control.

If we simplify this, we also should not need the new `--with-icon` modifier class at all, which means we don't need breaking changes to the component.

aprigogin: I got the feeling that this cannot be simplified any further.

aprigogin:
Closing in favor of #2420.

PR #2420, opened 2018-03-16
feat(button): Add padding mixin, adjust icon margin

### LINKED ISSUE #2400: Fix padding on buttons
Currently horizontal padding on buttons is not customizable, that needs to change.

kfranqueiro:
This addresses the following:

* Adds `mdc-button-horizontal-padding` mixin to parameterize left and right padding (one value applies to both)
* Updates `mdc-button-stroke-width` mixin to accept an optional padding parameter, since this mixin adjusts the button padding to maintain the same size when considering the added stroke
* Reduces effective padding between icon and nearest edge of button (left for LTR, right for RTL)
  * This is done via negative margin on the icon element rather than adjusting padding on the button; the latter would have required adding a class to button when using an icon, and would have caused complexity for both the padding and stroke-width mixins to accommodate different padding values for buttons with vs. without icons

Resolves #2400.


aprigogin:
Can you update this CL to handle the MD cases we need to support (like text icon button). Alternatively, you can leave additional comments on #2413.

kfranqueiro:
I'm totally up for updating this PR, though might need to wait until tomorrow... Maybe just as well since we need to figure out what to do about the text button + icon case anyway.

Pasting the list from #2400 to reflect status on this PR:

* [x] The icon-less text button (not contained) needs 8px padding on each side (currently it is 16px)
* The contained button with an icon should have:
  * [x] 12px padding to the left of the icon
  * [ ] 8px padding between the icon and the text
    * Need to suppress space from between icon and label (inline-flex might work), or revise padding to assume that space will exist
  * [x] 16px padding to the right of the text
* [x] The contained button with an icon should handle RTL by flipping the icon to the right side, and all appropriate padding should also flip
* [ ] The text button with an icon needs an audit for its padding (not sure if it should be closer to the icon-less text button, or closer to the contained button with an icon)
* [ ] The text button with an icon should also handle RTL

To me the real kicker is what we expect text buttons with icons to do. Everything else in this PR is really simple so far, and accomplishes this by adding a padding mixin and with minimal impact to the existing stroke mixin.

packages/mdc-button/mdc-button.scss
```diff
@@ -41,6 +41,16 @@
   }
 }
 
+.mdc-button--raised,
+.mdc-button--unelevated,
+.mdc-button--raised,
```
aprigogin: should --raised be listed twice?

packages/mdc-button/_mixins.scss
```diff
@@ -66,9 +66,14 @@
   border-radius: $corner-radius;
 }
 
-@mixin mdc-button-stroke-width($stroke-width) {
+@mixin mdc-button-horizontal-padding($padding) {
+  padding-right: $padding;
```
aprigogin: strictly speaking we should be adding /* @noflip */ on these. At the same time, I understand the argument about them being the same. Lets hope that never changes.

packages/mdc-button/mdc-button.scss
kfranqueiro: Whoops... clearly I misread a bit when I was looking at the selectors below to apply this to.

packages/mdc-button/_mixins.scss
kfranqueiro: Hm, yeah, I didn't think about it here since I know we handle `@noflip` within our RTL stuff now, and these styles are equal and not RTL-dependent. If these ever did change to being asymmetrical, we'd absolutely want to add `@noflip` (probably automatically via use of `mdc-rtl` mixins).

I think if we wanted to take that strict of a stance about `@noflip` we'd have several instances across the repo where left and right are equal that we're not flagging right now.

PR #2423, opened 2018-03-16
fix(chips): Extend ripple to fill the chip when animating width

### LINKED ISSUE #2334: Add optional width/height parameters to MDCRipple.layout()

bonniezhou:
Fixes #2334.


kfranqueiro:
For context, this is an alternative to #2406 which seeks to resolve the issue entirely within mdc-chips rather than needing to add an optional parameter to MDC Ripple.

While this avoids affecting Ripple APIs and is less code overall than #2406, it requires adding more logic to a component to effectively patch another component's adapter, and if any other component were to need special treatment, they'd have to repeat this logic. I'm potentially still leaning towards #2406, but I'll see what others think.

bonniezhou:
Discussed offline with @kfranqueiro.

There are now 6 components (including this one) that override ripple adapter methods, so it may be worth parameterizing the ripple's constructor, not just the `layout()` method. This would require a more thorough investigation and design, so #2406 won't cut it. Worth looking into a full refactor of ripple.

In the meantime this PR is a fix adhering to the pattern in the other components.

PR #2440, opened 2018-03-20
style: Fix IDE code completion for build config JS

acdvorak:
Editors like IntelliJ provide code completion, navigation, and refactoring features for JavaScript, IFF you use the "right" style of exports.


test/build/mock-env.js
```diff
@@ -17,7 +17,7 @@
 /**
  * Temporarily overwrites environment variables and restores their original values.
  */
-module.exports = class {
+class MockEnv {
```
kfranqueiro: Add `@fileoverview`?

acdvorak:
`test/build/mock-env.js` was deleted in #2450, so we don't need to add `@fileoverview` to it anymore.

PR #2441, opened 2018-03-20
fix: Fix for MDC Textfield / Icon / Tab Index changed to 0 by disable…

### LINKED ISSUE #2156: Bug: MDC Textfield / Icon / Tab Index changed to 0 by disabled call

Latest - 0.29.0

All

See codepen here:
https://codepen.io/MrChriZ/pen/MQwVrz

1. In demo insert text focus on textfield, then press tab.  Text focus moves away from the textfield as expected.
2.  Uncomment out the line:
myTextField.disabled=false;
3. Repeat text focus test.  This time the icon will be focused.

The icon's tabindex is left alone/restored to it's original value allowing focus to move away from the textfield.

The icon's tabindex is left alone/restored to it's original value

Dspite the fact that the icon is created with tabindex -1 the tabindex is being changed to 0 by the textfield/icon js. 
This occurs in textfield/icon/foundation.js javascript which changes the tabindex to 0 when disabled is set to false.

MrChriZ:
Call #2156

Used mdc-icon-toggle as guide to add a number of tweaks to allow saving of TabIndex prior to
disabled call, and then restoration when disabled is turned off.  Also adds new tests to check additional Method calls work.

fix #2156


MrChriZ:
I signed it!

MrChriZ:
Still signed :/

MrChriZ:
I signed it!

MrChriZ:
I signed it abrakadabra?

MrChriZ:
I signed it - Open Sesame

MrChriZ:
Hmm travis check has disappeared... wonder what I've done :S

kfranqueiro:
Thanks for working on this. Some other changes were causing this branch to conflict with master now, so rather than force you to deal with that, I applied a similar solution against master in #2600.

PR #2442, opened 2018-03-20
style: Fix linter errors in package subdirectories

acdvorak:
Our linter commands are missing double quotes around the glob pattern. As a result, we're currently not linting _any_ files in subdirectories of our packages.

E.g.:

* `packages/mdc-chips/chip/mdc-chip.scss` does _not_ get linted
* `packages/mdc-chips/mdc-chips.scss` _does_ get linted

From the [stylelint CLI docs](https://github.com/stylelint/stylelint/blob/a9aad2945976db4bee16805b2df3ed2a44c5debd/docs/user-guide/cli.md#examples):

> When you run commands similar to the examples below, be sure to include the quotation marks around file globs. This ensures that you can use the powers of node-glob (like the `**` globstar) regardless of your shell.

In short, [shells don't support the `**` syntax](https://en.wikipedia.org/wiki/Glob_(programming)), so we need to wrap the pattern in double quotes to prevent the shell from trying to parse it.

This ensures that the pattern gets passed verbatim to stylelint, which then passes it to node-glob for parsing (and node-glob _does_ support the `**` syntax).


acdvorak:
Travis CI test failures are expected because a bunch of linter violations are already merged into master.

PR #2428, opened 2018-03-18
Add CSS import syntax for card

joshunger:

PR #2443, opened 2018-03-21
Fix eslint warning on escaping entities

joshunger:


joshunger:
@kfranqueiro good question looks like I was using the eslint-plugin-react rules which flagged it. https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md do you want this change or do you want to leave it as is?

kfranqueiro:
I think I'd be inclined to leave it as-is for this example. Thanks though!

joshunger:
@kfranqueiro closed; do you want #2428 ?

PR #2449, opened 2018-03-22
fix(button): Suppress whitespace between icon and text label

kfranqueiro:
Opening this for potential discussion. This is a follow-on to #2420 / #2400.

This provides a solution to the problem where whitespace between the icon and text label (which is common since the icon tends to be on its own line in HTML source) causes the spacing between icon and text to be larger than intended by design.

This fixes it by changing button's display property from `inline-block` to `inline-flex`. This doesn't have any effects on usage of button in terms of its flow among other external content.

However, it does have one implication for the button's text label content: if the text label includes any child elements, then the entire label needs to be nested in a child element to avoid inline-flex collapsing the space between text nodes and/or child elements. This is illustrated in several updates to demo pages in this PR to keep them consistent with before.

This ultimately comes down to how common/supported we expect the use case of child elements within text labels to be. We could add a caveat note to the readme if we expect users to run across it.


packages/mdc-button/_mixins.scss
```diff
@@ -95,7 +95,7 @@
   @include mdc-ripple-radius-bounded;
   @include mdc-button-horizontal-padding($mdc-button-horizontal-padding);
 
-  display: inline-block;
+  display: inline-flex;
```
aprigogin: I think you might want to change 
text-align: center; to justify-content: center;

packages/mdc-button/_mixins.scss
```diff
@@ -143,9 +143,8 @@
 }
 
 @mixin mdc-button__icon-svg_ {
-  margin-top: -2px;
   fill: currentColor;
-  vertical-align: middle;
+  align-self: center;
```
aprigogin: maybe align-items: center on button_base?

PR #2453, opened 2018-03-22
chore: Split build task into separate tasks for dist/demo

kfranqueiro:
As of #2437, `npm run build` now builds 2 sets of files at once: demo css/js, and dist css/js. These used to be split via the `MDC_ENV` variable (see https://github.com/material-components/material-components-web/pull/2325/files#r176188331).

The problem with this is now `npm run build` builds a lot more than it used to, and this causes our `cp-pkgs` script to fail due to the presence of files not matching the pattern it expects; this is something that we would've quickly tripped over upon next release if it weren't found now.

In retrospect, Andy and I agree it's odd that we had `npm run build` building two almost completely different sets of files based on an environment variable; therefore this PR splits them out. Building the demos is only ever used for catalog deployment upon release, so I updated the command listed in our release docs accordingly. `npm run build` will now proceed to act as it did before for building dist files.


docs/open_source/release.md
```diff
@@ -82,6 +82,6 @@ This will ensure the commits *and* tags are pushed to the remote git repository.
 
 ### Deploy Catalog Server
 
-`MDC_ENV=development npm run build && gcloud app deploy`
+`MDC_ENV=development npm run build:demos && gcloud app deploy`
```
acdvorak: We also need to update the "Appengine development server" command in [CONTRIBUTING.md](https://github.com/material-components/material-components-web/blob/6078784954b484df575fd54bd732445add358a96/CONTRIBUTING.md#appengine-development-server)

Issue #2373, opened 2018-03-09
Top App Bar/Toolbar button icon not spaced relative to toolbar title and has 0 size.


dprentiss:
### What MDC Web Version are you using?

0.32.0

### What browser(s) is this bug affecting?

Google Chrome Version 64.0.3282.186 (Official Build) (64-bit)

Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.186 Safari/537.36

### What OS are you using?

Linux 4.13.0-36-generic #40-Ubuntu SMP Fri Feb 16 20:07:48 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### What are the steps to reproduce the bug?

Bug occurs when a `<button>` element is used to contain the `mdc-toolbar__menu-icon` in a temporary or persistent drawer.

From [Drawer - Material Components Catalog](https://material-components-web.appspot.com/drawer/index.html):

```html
   <div class="mdc-toolbar mdc-toolbar--fixed">
      <div class="mdc-toolbar__row">
        <section class="mdc-toolbar__section mdc-toolbar__section--align-start">
          <button class="demo-menu material-icons mdc-toolbar__menu-icon">menu</button>
          <span class="mdc-toolbar__title catalog-title">Temporary Drawer</span>
        </section>
      </div>
    </div>
```
### What is the expected behavior?

Icon should be spaced properly relative to `mdc-toolbar_title`. 

### What is the actual behavior?

Icon is placed with no spacing and computed size of 0x0px.

### Any other information you believe would be useful?
This appears to be caused by the user stylesheet applying `box-sizing: border-box`. 

### Workaround
Override user stylesheet with `box-sizing: content-box`.

patrickrodee:
Thanks for submitting an issue. We've added this to our internal tracker.

williamernest:
Steps to reproduce: 
1. Change the menu icon from an anchor element to a `button` element.

PR #2454, opened 2018-03-24
fix(toolbar): Fix toolbar/top-app-bar button icons

williamernest:
Fixes: #2373 
Changed the icons to use `box-sizing: border-box` and adjusted the `width`/`height` values.


packages/mdc-toolbar/_mixins.scss
```diff
@@ -60,8 +60,9 @@
   position: relative;
   align-items: start;
   justify-content: center;
-  width: $mdc-toolbar-icon-size;
-  height: $mdc-toolbar-icon-size;
+  box-sizing: border-box;
+  width: $mdc-toolbar-icon-size * 2;
```
bonniezhou: Let's change this (and below) to `$mdc-toolbar-icon-size + $mdc-toolbar-icon-padding * 2` to be more clear?

PR #2462, opened 2018-03-26
refactor(select): replace menu with native html select

### LINKED ISSUE #2399: Revert `mdc-select` to use the `<select>` element
We initially removed this because we we're trying to style `mdc-select` so it used `mdc-menu`. But we discovered that we lost a lot of native functionality with the `<select>` element. So we will remove `mdc-select`'s dependency on `mdc-menu`, and instead use the native `<select>` element

moog16:
fixes #2399


packages/mdc-select/index.js
```diff
@@ -51,29 +54,28 @@ export class MDCSelect extends MDCComponent {
...
-    this.foundation_.setDisabled(disabled);
+    this.foundation_.disabled = disabled;
   }
 
   item(index) {
-    return this.options[index] || null;
+    return this.surface_.item(index);
   }
 
-  nameditem(key) {
```
moog16: may want to reimplement namedItem

packages/mdc-select/README.md
```diff
@@ -44,164 +40,124 @@ npm install @material/select
...
+    <option class="mdc-select__option" value="fats">
+      Fats, Oils, and Sweets
+    </option>
+  </select>
+  <div class="mdc-select__label">Pick a Food Group</div>
+  <div class="mdc-select__bottom-line"></div>
 </div>
 ```
 
+Use the `<option disabled selected/>` element as a way to display the select having no value.
```
lynnjepsen: Say what?

packages/mdc-select/README.md
```diff
@@ -242,113 +195,49 @@ is outlined below.
...
 | `disabled` | `boolean` | Whether or not the component is disabled. Settings this sets the disabled state on the component. |
 
-#### Methods
+### `MDCSelect`
 
 | Method Signature | Description |
 | --- | --- |
 | `item(index: number) => HTMLElement?` | Analogous to `HTMLSelectElement.prototype.item`. Returns the option at the specified index, or `null` if the index is out of bounds. }
 | `nameditem(key: string) => HTMLElement?` | Analogous to `HTMLSelectElement.prototype.nameditem`. Returns the options either whose `id` equals the given `key`, or whose `name` attribute equals the given `key`. Returns `null` if no item with an `id` or `name` attribute of the specified key is found. |
+| `initialSyncWithDOM() => void` | Syncs the component with the current state of the HTML markup. |
```
lynnjepsen: This is a method on all components, you don't need to document it

packages/mdc-select/README.md
```diff
@@ -211,11 +167,7 @@ To disable a list item, set `aria-disabled` to `"true"`, and set `tabindex` to `
 | ------------------------ | ----------------------------------------------- |
 | `mdc-select`             | Mandatory.                                      |
 | `mdc-select--box`        | Styles the select as a box select.              |
-| `mdc-list-group`         | A group of options.                             |
-| `mdc-list-item`          | A list item.                                    |
-| `mdc-list-divider`       | A divider.                                      |
-
-It is advised that dividers also set `role="presentation"` to disable selection and not cloud accessibility.
+| `mdc-select__option`     | A select option.                                |
```
lynnjepsen: Is this a necessary class?

packages/mdc-select/index.js
```diff
@@ -31,49 +30,56 @@ export class MDCSelect extends MDCComponent {
...
   get options() {
-    return this.menu_.items;
+    return [].slice.call(this.surface_.options);
   }
 
   get selectedOptions() {
-    return this.root_.querySelectorAll('[aria-selected]');
+    // In an array because HTML spec also returns an array.
+    // Cannot access HTMLSelectElement.selectedIndex since it is not supported by IE11.
+    // https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex
```
williamernest: `selectedIndex` -> `selectedOptions`
`https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex` -> `https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedOptions`

kfranqueiro:
I'll get to this review in a bit, but FYI I'm immediately noticing that the arrow doesn't look like it's vertically aligned correctly for the base MDC Select...

image

packages/mdc-select/mdc-select.scss
```diff
@@ -114,14 +109,15 @@
     background-position: left 10px center;
   }
 
-  .mdc-select__surface {
+  .mdc-select__surface,
```
kfranqueiro: Since this element is no longer the ripple surface, we might want to consider renaming it... e.g. checkbox has `__native-control` for this.

(We'd want to change variable names in JS too)

packages/mdc-select/mdc-select.scss
```diff
@@ -27,12 +28,15 @@
 // postcss-bem-linter: define select
 .mdc-select {
   @include mdc-typography-base;
+  @include mdc-states;
```
kfranqueiro: This should be in `--box` styles where the other ripple mixins are used. It doesn't make sense for this to apply to anything without those styles.

When you move it, change it to: `@include mdc-states(text-primary-on-light, $has-nested-focusable-element: true);`

This will make the ripple match the ink color, and make focus shade work properly given the nested select element, even when ripple JS isn't used (I know, right now we require JS, but this seems like the more technically correct thing to do esp. if there is ever hope of CSS-only down the road)

packages/mdc-select/mdc-select.scss
```diff
@@ -27,12 +28,15 @@
...
+
```
kfranqueiro: We might want to go ahead and remove the following `@each` which is applying letter-spacing typography styles, since IIUC we're probably removing that property from typography anyway, and this might be hard to catch when we do a sweep later... @lynnjepsen what do you think?

packages/mdc-select/README.md
```diff
@@ -44,164 +40,117 @@ npm install @material/select
...
+    </option>
+  </select>
+  <div class="mdc-select__label mdc-select__label--float-above">Pick a Food Group</div>
+  <div class="mdc-select__bottom-line"></div>
 </div>
 ```
 
+#### Select with floating label as the placeholder
+
+Use the `<option disabled selected hidden/>` element as a way to display the select having no value.
```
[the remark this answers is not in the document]
moog16: I was saying to add float-above if you want a placeholder instead of using the label, but the way I worded it is confusing and this might not be the place so I'll remove. Should I also remove `hidden` from the demos as well?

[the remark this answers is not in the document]
kfranqueiro: Yes, I think I commented on the line in the demo page too.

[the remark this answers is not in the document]
kfranqueiro: Also, not sure what you mean regarding wanting a placeholder instead of using the label... the `placeholder` attribute is not a thing for native `select` (since `select` normally always has an option selected)

packages/mdc-select/README.md
```diff
@@ -242,113 +188,45 @@ is outlined below.
...
 | Method Signature | Description |
 | --- | --- |
-| `item(index: number) => HTMLElement?` | Analogous to `HTMLSelectElement.prototype.item`. Returns the option at the specified index, or `null` if the index is out of bounds. }
-| `nameditem(key: string) => HTMLElement?` | Analogous to `HTMLSelectElement.prototype.nameditem`. Returns the options either whose `id` equals the given `key`, or whose `name` attribute equals the given `key`. Returns `null` if no item with an `id` or `name` attribute of the specified key is found. |
+| `item(index: number) => HTMLElement?` | Analogous to `HTMLSelectElement.prototype.item`. Returns the option at the specified index, or `null` if the index is out of bounds. |
 
 #### Events
 
-The MDC Select JS component emits an `MDCSelect:change` event when the selected option changes as
+The MDC Select JS component emits an `change` event when the selected option changes as
```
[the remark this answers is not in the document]
moog16: Yes, we're using the native selects `change` event.

packages/mdc-select/_mixins.scss
```diff
@@ -34,8 +34,7 @@
 
 @mixin mdc-select-focused-label-color($color, $opacity: .87) {
   &:not(.mdc-select--disabled) {
-    .mdc-select__surface:focus,
-    &.mdc-select--open {
+    .mdc-select__surface:focus ~ {
```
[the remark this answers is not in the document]
moog16: Yup it does work. It is selecting any sibling appearing after itself. In this case, it is the label.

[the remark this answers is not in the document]
kfranqueiro: Oh wait. It's working because the floating-label-color mixin is appending the correct class, which makes the combinator valid.

That's pretty confusing without also looking at the mixin... might want to add a comment.

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -53,19 +37,18 @@ class FakeBottomLine {
...
-      <div class="mdc-menu mdc-select__menu">
-        <ul class="mdc-list mdc-menu__items">
-          <li class="mdc-list-item" role="option" aria-disabled="true">
-            Pick a food group
-          </li>
-        </ul>
-      </div>
+    <div class="mdc-select">
+      <select class="mdc-select__native-control">
+        <option value="" disabled selected hidden></option>
```
kfranqueiro: Remove `hidden`

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -115,77 +82,153 @@ test('#get/setDisabled', () => {
   const {component} = setupTest();
   assert.equal(component.disabled, false);
   component.disabled = true;
-  assert.isOk(component.disabled);
+  assert.isTrue(component.disabled);
```
kfranqueiro: Change line 83 (2 lines before this) to `assert.isFalse(component.disabled)`

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -115,77 +82,153 @@ test('#get/setDisabled', () => {
...
+  assert.equal(component.value, 'orange');
+});
+
+test('#set value sets the value on the <select>', () => {
+  const {component, nativeControl} = setupTest();
+  component.value = 'orange';
+  assert.equal(nativeControl.value, 'orange');
+});
+
+test('#set value sets non value to empty string on <select>', () => {
```
kfranqueiro: This is testing native browser behavior; remove it

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -115,77 +82,153 @@ test('#get/setDisabled', () => {
...
 
 test('adapter#removeClass removes a class from the root element', () => {
   const {component, fixture} = setupTest();
   fixture.classList.add('foo');
   component.getDefaultFoundation().adapter_.removeClass('foo');
-  assert.isNotOk(fixture.classList.contains('foo'));
+  assert.isFalse(fixture.classList.contains('foo'));
+});
+
+test('does not create label', () => {
```
kfranqueiro: We would never expect a component or adapter to create elements; we should instead test that `adapter_.floatLabel` doesn't throw.

There are some examples of this e.g. for helperText in mdc-text-field.test.js.

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -115,77 +82,153 @@ test('#get/setDisabled', () => {
...
+        </option>
+      </select>
+    </div>
+  `;
+  const component = new MDCSelect(fixture);
+  component.getDefaultFoundation().adapter_.floatLabel('foo');
+  assert.equal(component.label_, undefined);
+});
+
+test('does not create bottomLine element', () => {
```
kfranqueiro: See previous comment

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -115,77 +82,153 @@ test('#get/setDisabled', () => {
...
+        </option>
+      </select>
+    </div>
+  `;
+  const component = new MDCSelect(fixture);
+  assert.equal(component.bottomLine_, undefined);
+});
+
+test(`does not create ripple element if ${cssClasses.BOX}` +
+  'class is present', () => {
```
kfranqueiro: is not present

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -115,77 +82,153 @@ test('#get/setDisabled', () => {
...
+          Orange
+        </option>
+        <option value="apple" selected>
+          Apple
+        </option>
+      </select>
+    </div>
+  `;
+  const component = new MDCSelect(fixture);
+  assert.equal(component.ripple, undefined);
```
kfranqueiro: `assert.isUndefined(component.ripple)`

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -209,232 +252,55 @@ test('adapter#deactivateBottomLine removes active class from the bottom line', (
...
-  document.body.appendChild(fixture);
-
-  component.getDefaultFoundation().adapter_.focus();
-  assert.equal(document.activeElement, surface);
-
-  document.body.removeChild(fixture);
+test('adapter#setValue sets the value of the select to the correct option', () => {
+  const {nativeControl, component} = setupTest();
+  component.getDefaultFoundation().adapter_.setValue('orange');
+  assert.equal(nativeControl.selectedIndex, 1);
```
kfranqueiro: Why not verify against .value in the setValue test?

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -209,232 +252,55 @@ test('adapter#deactivateBottomLine removes active class from the bottom line', (
...
 });
 
-test('adapter#makeTabbable sets the surface element\'s tabindex to 0', () => {
-  const {component, surface} = setupTest();
-  surface.tabIndex = -1;
-  component.getDefaultFoundation().adapter_.makeTabbable();
-  assert.equal(surface.tabIndex, 0);
+test('adapter#setValue sets the selectedIndex of the select to -1 if there is ' +
```
kfranqueiro: This is testing native browser behavior; remove it

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -209,232 +252,55 @@ test('adapter#deactivateBottomLine removes active class from the bottom line', (
...
-  const {component, fixture} = setupTest();
-  component.getDefaultFoundation().adapter_.setSelectedTextContent('content');
-  assert.equal(fixture.querySelector('.mdc-select__selected-text').textContent, 'content');
-});
-
-test('adapter#getNumberOfOptions returns the length of the component\'s options property', () => {
+test('adapter#getValue returns the nativeControl value', () => {
   const {component} = setupTest();
-  assert.equal(component.getDefaultFoundation().adapter_.getNumberOfOptions(), component.options.length);
+  component.getDefaultFoundation().adapter_.setValue('orange');
```
kfranqueiro: Replace this with `nativeControl.value = 'orange'` to ensure we're actually testing what we say we're testing. This would currently pass if the set/getValue APIs match even if neither one touches the native input.

(You're doing the right thing below to test getSelectedIndex, so not sure why this one was done like this.)

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -209,232 +252,55 @@ test('adapter#deactivateBottomLine removes active class from the bottom line', (
...
+  assert.equal(component.getDefaultFoundation().adapter_.getValue(), 'orange');
 });
 
-test('adapter#getTextForOptionAtIndex gets the text content for the option at the given index', () => {
+test('adapter#getValue returns empty string if nothing is selected', () => {
```
kfranqueiro: This is testing native browser behavior; remove it

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -209,232 +252,55 @@ test('adapter#deactivateBottomLine removes active class from the bottom line', (
...
-test('adapter#addBodyClass adds a class to the body', () => {
-  const {component} = setupTest();
-  component.getDefaultFoundation().adapter_.addBodyClass('mdc-select-scroll-lock');
-  assert.isOk(document.querySelector('body').classList.contains('mdc-select-scroll-lock'));
+  component.selectedIndex = 1;
+  assert.equal(component.getDefaultFoundation().adapter_.getSelectedIndex(), 1);
 });
 
-test('adapter#removeBodyClass remove a class from the body', () => {
+test('adapter#getSelectedIndex returns -1 if selected index is out of range', () => {
```
kfranqueiro: This is testing native browser behavior; remove it

demos/select.scss
```diff
@@ -15,6 +15,7 @@
 //
 
 @import "./common";
+@import "../packages/mdc-button/mdc-button";
```
kfranqueiro: We should be able to remove mdc-list and mdc-menu imports here, right?

Also, we should be able to remove mdc-menu and mdc-list from mdc-select's package.json, which hasn't been touched in this PR.

packages/mdc-select/README.md
```diff
@@ -242,113 +183,36 @@ is outlined below.
...
-
-The select's bottom border is set to the current theme's primary color when focused.
+| `floatLabel(value: boolean) => void` | Floats or defloats label as necessary. |
+| `activateBottomLine() => void` | Activates the bottom line component. |
+| `deactivateBottomLine() => void` | Deactivates the bottom line component. |
+| `setDisabled(disabled: boolean) => void` | Sets the `disabled` property of the `<select>` element. |
+| `registerInteractionHandler(type: string, handler: EventListener) => void` | Adds an event listener `handler` for event type `type` on the `<select>` element. |
+| `deregisterInteractionHandler(type: string, handler: EventListener) => void` | Removes an event listener `handler` for event type `type` on the `<select>` element. |
+| `getSelectedIndex() => number` | Returns the selected index of the `<select>` element. |
+| `setSelectedIndex(index: number) => void` | Sets the select's selectedValue to the option found at the provided index. If the index is out of the select's range, it will default to -1. |
```
kfranqueiro: Fix description: Sets the selected index of the `<select>` element.

packages/mdc-select/README.md
```diff
@@ -242,113 +183,36 @@ is outlined below.
...
-Returns whether or not the select is disabled.
-
-#### MDCSelectFoundation.setDisabled(disabled: boolean) => void
-
-Enables/disables the select.
-
-## Theming
-
-The select's bottom border is set to the current theme's primary color when focused.
+| `floatLabel(value: boolean) => void` | Floats or defloats label as necessary. |
```
kfranqueiro: This is a simple on/off with no in-grained logic now, right? I'd remove "as necessary"

packages/mdc-select/README.md
```diff
@@ -242,113 +183,36 @@ is outlined below.
...
+| `activateBottomLine() => void` | Activates the bottom line component. |
+| `deactivateBottomLine() => void` | Deactivates the bottom line component. |
+| `setDisabled(disabled: boolean) => void` | Sets the `disabled` property of the `<select>` element. |
+| `registerInteractionHandler(type: string, handler: EventListener) => void` | Adds an event listener `handler` for event type `type` on the `<select>` element. |
+| `deregisterInteractionHandler(type: string, handler: EventListener) => void` | Removes an event listener `handler` for event type `type` on the `<select>` element. |
+| `getSelectedIndex() => number` | Returns the selected index of the `<select>` element. |
+| `setSelectedIndex(index: number) => void` | Sets the select's selectedValue to the option found at the provided index. If the index is out of the select's range, it will default to -1. |
+| `getValue() => string` | Returns the value selected on the `<select>` element. |
+| `setValue(value: string) => void` | Sets the select's value. If no option has the provided value, it sets `value` to empty string. |
```
kfranqueiro: Reduce this to: Sets the value of the `<select>` element.

(The rest is re-documenting native behavior.)

packages/mdc-select/README.md
```diff
@@ -242,113 +183,36 @@ is outlined below.
...
+
+### `MDCSelectFoundation`
+
+| Method Signature | Description |
+| --- | --- |
+| `setValue(value: string) => void` | Sets the select's value. If value is not found within list of options, it will set it to empty string. |
```
kfranqueiro: Similar to above, reduce description: Sets the value of the component.

packages/mdc-select/README.md
```diff
@@ -44,164 +40,114 @@ npm install @material/select
...
+      Milk, Yogurt, and Cheese
+    </option>
+  </select>
+  <div class="mdc-select__label mdc-select__label--float-above">Pick a Food Group</div>
+  <div class="mdc-select__bottom-line"></div>
 </div>
 ```
 
+#### Select with floating label as the placeholder
+
```
kfranqueiro: Extra newlines here and before the next heading? Just need one empty line

packages/mdc-select/mdc-select.scss
```diff
@@ -44,54 +48,36 @@
...
-  }
-
   &__surface {
     @include mdc-rtl-reflexive-property(padding, 0, $mdc-select-arrow-padding);
-    @include mdc-states;
 
-    display: flex;
-    position: relative;
-    flex-grow: 1;
+    &::-ms-expand {
```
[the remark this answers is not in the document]
kfranqueiro: This still needs to be resolved; non-box selects are still using the browser's default input font

packages/mdc-select/index.js
```diff
@@ -85,11 +65,10 @@ export class MDCSelect extends MDCComponent {
     if (bottomLineElement) {
       this.bottomLine_ = bottomLineFactory(bottomLineElement);
     }
-    this.selectedText_ = this.root_.querySelector(strings.SELECTED_TEXT_SELECTOR);
-    this.menuEl_ = this.root_.querySelector(strings.MENU_SELECTOR);
-    this.menu_ = menuFactory(this.menuEl_);
 
-    this.ripple = new MDCRipple(this.surface_);
+    if (this.root_.classList.contains(cssClasses.BOX)) {
+      this.ripple = new MDCRipple(this.root_);
```
kfranqueiro: Only one of my two previous comments was resolved. Here's the other one:

We probably want to do something like what we do with checkbox/radio/text-field to forward the ripple component's event handlers to the select element. This should make activation work when pressing space, and should make focus styles work via ripple JS.

Specifically, I'm talking about e.g. this: https://github.com/material-components/material-components-web/blob/3f238212915cb781a1ca4cb4cad9da392a52bf22/packages/mdc-checkbox/index.js#L64-L65

test/unit/mdc-select/mdc-select.test.js
moog16: I'm going to wrap the adapter methods like so:
```
activateBottomLine: () => {
        if (this.bottomLine_) {
          this.bottomLine_.activate();
        }
      },
      deactivateBottomLine: () => {
        if (this.bottomLine_) {
          this.bottomLine_.deactivate();
        }
      },```

File: demos/select.scss (hunk 24 above)
Comment (moog16) [reply demos/select.scss]: Good call!

File: packages/mdc-select/mdc-select.scss [hunk 32]
```diff
@@ -44,55 +40,38 @@
...
-    bottom: 8px;
-    align-items: flex-end;
-    transition:
-      mdc-animation-exit-temporary(opacity, 125ms),
-      mdc-animation-exit-temporary(transform, 125ms);
+    background-color: transparent;
     white-space: nowrap;
-    overflow: hidden;
+    cursor: pointer;
+    appearance: none;
```
bcardiff: FYI, across the code base the prefixed version of appearance is used

```
-moz-appearance: none;
-webkit-appearance: none;
```

According to https://caniuse.com/#search=appearance the unprefixed version has very little support currently

kfranqueiro: The vendor prefixes are added automatically by autoprefixer when we build releases (and you are encouraged to do the same if you have your own Sass workflow). I have noticed that we have a few components using the vendor prefixes directly, which we should probably fix to use the non-prefixed version that will be picked up by autoprefixer.

PR #2464, opened 2018-03-27
chore(typography): Remove the adjust margin feature

### LINKED ISSUE #2463: Remove the adjust margin feature in mdc-typography
Remove the adjust margin feature in mdc-typography. It is not in MD guidelines.

lynnjepsen:
Fixes #2463


demos/top-app-bar.html
```diff
@@ -172,14 +172,14 @@ <h4 class="mdc-typography--subheading2 mdc-typography--adjust-margin">Common Opt
...
-          <h4 class="mdc-typography--subheading2 mdc-typography--adjust-margin">Top App Bar Specific Options</h4>
+          <span class="mdc-typography--subheading2 demo-options-heading">Top App Bar Specific Options</span>
           <div>
             <input type="checkbox" id="prominent-checkbox"/>
             <label for="prominent-checkbox">Prominent</label>
           </div>
         </div>
         <div class="demo-col">
-          <h4 class="mdc-typography--subheading2 mdc-typography--adjust-margin">Short Top App Bar Specific Options</h4>
+          <span class="mdc-typography--subheading2 demo-options-heading">Short Top App Bar Specific Options</span>
```
kfranqueiro: This change makes me realize something, related to this PR as well as the task to redefine the typography scales...

If we're going to be introducing typography scales for Headline 1-6, I would assume they're naturally going to often align with `h1` to `h6`. These elements commonly have default margins set by the user agent, which might not be appropriate when combined with our styles. Are we sure we don't want to override those with saner defaults? Or do we think we'd override those in all cases they're used, perhaps, thus obviating the need for adjust-margin anyway?

lynnjepsen: I think typography should only handle font properties (font-family, font-size, letter-spacing, etc) which can be displayed in any HTML element. We should not assume that `@mdc-typography` mixin is applied to a `h3` element.

packages/mdc-typography/README.md
```diff
@@ -90,26 +90,20 @@ CSS Class | Description
...
 `mdc-typography-overflow-ellipsis` | Truncates overflow text to one line with an ellipsis
 
-> **A note about `mdc-typography-adjust-margin`**, `mdc-typography-adjust-margin` will change the margin properties of the element it is applied to, to align text correctly. `mdc-typography-adjust-margin` should only be used in a text context; using this property on UI elements such as buttons may cause them to be positioned incorrectly.
-
 > **A note about `mdc-typography-overflow-ellipsis`**, `mdc-typography-overflow-ellipsis` should only be used if the element is `display: block` or `display: inline-block`.
 
 #### `$style` Values
 
-These styles can be used as the `$style` argument for `mdc-typography` and `mdc-typography-adjust-margin` mixins.
+These styles can be used as the `$style` argument for `mdc-typography` mixins.
```
kfranqueiro: Fix wording: These styles can be used as the `$style` argument for the `mdc-typography` mixin.

PR #2474, opened 2018-03-27
feat(top-app-bar): Add --fixed variant to top app bar

williamernest:
Fixes: #2425


packages/mdc-top-app-bar/constants.js
```diff
@@ -26,6 +26,8 @@ const strings = {
 
 /** @enum {string} */
 const cssClasses = {
+  FIXED_CLASS: 'mdc-top-app-bar--fixed',
+  FIXED_SCROLLED_CLASS: 'mdc-top-app-bar--fixed__scrolled',
```
acdvorak: In BEM, `__` indicates a sub-element, but this appears to be a modifier class (for a modifier class - yo dawg).

We normally handle that by using a single `-` instead of `__`.

E.g.: `mdc-top-app-bar--fixed-scrolled`

packages/mdc-top-app-bar/fixed/foundation.js
```diff
@@ -0,0 +1,69 @@
...
+ * @extends {MDCTopAppBarFoundation<!MDCFixedTopAppBarFoundation>}
+ * @final
+ */
+class MDCFixedTopAppBarFoundation extends MDCTopAppBarFoundation {
+  /**
+   * @param {!MDCTopAppBarAdapter} adapter
+   */
+  constructor(adapter) {
+    super(adapter);
+    // State variable for the previous top app bar state
```
acdvorak: Pro tip: Always use `/**` comments to document fields and methods.

Editors like IntelliJ will parse the JSDoc `/**` comments and display them inline when you hit <kbd>Ctrl</kbd>+<kbd>Q</kbd>.

packages/mdc-top-app-bar/mdc-top-app-bar.scss
```diff
@@ -145,5 +145,13 @@
   }
 }
 
+.mdc-top-app-bar--fixed {
+  position: fixed;
```
acdvorak: We probably want some kind of `transition` for `box-shadow` to make the transition between `--fixed` and `--fixed-scrolled` a little smoother.

demos/top-app-bar.html
```diff
@@ -232,6 +238,23 @@ <h4 class="mdc-typography--subheading2 mdc-typography--adjust-margin">Short Top
...
 
+        // Top App Bar Specific Options
+        prominentCheckbox.addEventListener('change', function()  {
+          appBarEl.classList[this.checked ? 'add' : 'remove']('mdc-top-app-bar--prominent');
+
+          shortCheckbox.disabled = this.checked;
+        });
+
+        fixedCheckbox.addEventListener('change', function()  {
+          appBarEl.classList[this.checked ? 'add' : 'remove']('mdc-top-app-bar--fixed');
```
acdvorak: Nit: You might also want to toggle/remove `mdc-top-app-bar--fixed-scrolled` here, if it's not too hard to do.

PR #2475, opened 2018-03-27
feat(top-app-bar): Add dense

williamernest:
closes: #2426 
Adds the `--dense` style to the top app bar.


packages/mdc-top-app-bar/mdc-top-app-bar.scss
```diff
@@ -145,5 +145,36 @@
...
+}
+
+// Specific styles for prominent and dense styled top app bar
+.mdc-top-app-bar--dense.mdc-top-app-bar--prominent {
+  .mdc-top-app-bar__row {
+    height: $mdc-top-app-bar-dense-row-height * 2;
+  }
+
+  .mdc-top-app-bar__section {
+    padding: 0 $mdc-top-app-bar-dense-section-horizontal-padding;
```
moog16: I don't think you need this css rule. Isn't it covered by line 154?

demos/top-app-bar.html
```diff
@@ -256,10 +261,16 @@ <h3 class="mdc-typography--title">Demo Controls</h3>
           appBar = mdc.topAppBar.MDCTopAppBar.attachTo(appBarEl);
         });
 
+        denseCheckbox.addEventListener('change', function()  {
+          appBarEl.classList[this.checked ? 'add' : 'remove']('mdc-top-app-bar--dense');
+
+          shortCheckbox.disabled = this.checked || prominentCheckbox.checked;
```
kfranqueiro: Dense disables short, but short does not disable dense

packages/mdc-top-app-bar/mdc-top-app-bar.scss
```diff
@@ -145,5 +159,24 @@
...
+  }
+
+  .mdc-top-app-bar__section {
+    padding: 0 $mdc-top-app-bar-section-horizontal-padding;
+  }
+
+  .mdc-top-app-bar__title {
+    @include mdc-rtl-reflexive-box(padding, left, $mdc-top-app-bar-title-left-padding);
+
+    padding-bottom: $mdc-top-app-bar-prominent-dense-title-bottom-padding;
```
kfranqueiro: This is also overriding mobile-specific title bottom padding for prominent (not dense). Is that intentional?

(TBH I find it hard to follow the plethora of different title paddings but I get the impression it's to counteract outer padding...)

williamernest: It's intentional. The `--dense` style is desktop only, so any styles it applies should override any other styles from the mobile media queries.

PR #2476, opened 2018-03-28
docs: Update README template and standards

bonniezhou:
Re-structure our READMEs to make our API documentation more readable.


docs/code/readme_standards.md
```diff
@@ -108,20 +143,18 @@ All tables should follow this pattern:
...
-difficult to read.
+~~~
+>_NOTE_: This is the long description.
+~~~
 
-If a table feels too long, and there is a pattern to the rows, you can use an
-uppercase angled brackets, e.g. `<FOO>` format to explain the pattern. Add a
-note above the table detailing what the uppercase angled brackets means,
-including an example of a valid pattern match.
+Do not add multiple sentences to the tabular format, as this will make the table more difficult to read.
```
kfranqueiro: I don't 100% agree with this...

Firstly, "Mandatory." or "Optional." is sometimes the first sentence :) (though for that matter maybe we should standardize on only pointing out mandatory and implying everything else is optional, 'cause there tend to be far more optional things than mandatory ones)

Secondly, we sometimes add a second sentence to explain the meaning of parameters (especially optional ones), and I'm not sure having separate NOTEs for a bunch of mixins would scale for that purpose (unless multiple mixins follow a similar pattern).

docs/code/readme_standards.md
```diff
@@ -3,100 +3,135 @@
...
+
+```html
+<button class="mdc-button mdc-button--foo">Foo</button>
 ```
-Variable | Description
+~~~
+
+## Style Customization
+
+CSS Classes are documented in a [tabular format](#tabular-format).
```
kfranqueiro: We've got a mix of "in a tabular format" and "in tabular format" across this file. Pick one?

docs/code/readme_template.md
```diff
@@ -61,7 +64,25 @@ path: /catalog/<COMPONENT_ID>/
...
+
+<STYLE_CUSTOMIZATION_SECTION>
+
+## `<JS_COMPONENT_NAME>` Properties and Methods
+
+<MDC_COMPONENT_SECTION>
+
+## Usage within Web Frameworks
+
+If you are using a JavaScript framework, such as React or Angular2, you can create a <COMPONENT_NAME> for your framework by wrapping <JS_COMPONENT_NAME>. Depending on your framework, you can use the _Simple Approach: Wrapping MDC Web Vanilla Components_, or the _Advanced Approach: Using Foundations and Adapters_. Please follow the instructions [here](https://github.com/material-components/material-components-web/blob/master/docs/integrating-into-frameworks.md).
```
kfranqueiro: I might remove "by wrapping <JS_COMPONENT_NAME>" from the first sentence, since that sort of leads in one of the two possible directions whereas the second sentence explains both options. (Maybe also re-word "Depending on your framework" to "Depending on your needs" to reduce redundancy since the first sentence would also end with "framework" 3 words prior.)

Also, replace absolute link with a relative one i.e. `../integrating-into-frameworks.md`

docs/code/readme_standards.md
```diff
@@ -3,100 +3,135 @@
...
-```
+`mdc-foo` | Indicates the foo
+~~~
 
-Sass mixins, including their method signature, are documented in a
-[tabular format](#tabular-format).
+Sass mixins are documented in [tabular format](#tabular-format).
+
+~~~
+### Sass Mixins
```
kfranqueiro: What about optional sections/tables for:

* Sass Functions
* Sass Variables (esp. for subsystem packages such as theme and typography which have or will have overridable global variables)
* CSS Variables (probably rare, but e.g. for theme; maybe this can be relegated to / summarized in additional information)

docs/code/readme_standards.md
```diff
@@ -3,100 +3,135 @@
...
 
-### JavaScript
+Add a section for additional information if applicable.
 
-First have a section for the Vanilla Component. The title of the
-section should be the name of the class, formatted as inline code, aka with the \` marks.
-At the top of the section you should reference our generic documentation for
-importing a JavaScript component.
+~~~
+### Additional Information
```
kfranqueiro: How would we handle stuff like the Tips and Tricks, util API, and Caveats sections in Ripple? Would those all end up crammed under here now? Those would get into really tiny subheadings if they had to be nested starting at h4... (See https://github.com/material-components/material-components-web/tree/master/packages/mdc-ripple#tipstricks and onward in the Ripple readme)

docs/code/readme_standards.md
bonniezhou: Added. I put all the Sass mixins/functions/variables under one subsection, but gave CSS custom properties its own subsection since looking at theme it seems to be quite important. Does this make sense?

docs/code/readme_standards.md
bonniezhou: Those sections can replace 'Additional Information'. Added a note.

docs/code/readme_standards.md
bonniezhou: Yeah good point about Mandatory/Optional. Changed "sentence" to "line", hopefully that gets the idea across.
The idea here is to keep the table so users can skim it to find what they're looking for, and then go into details later. It doesn't always fit on one line but that's the goal, and authors can use their own discretion. The Ripple README is an example of separating details out into notes.

docs/code/readme_standards.md
```diff
@@ -3,100 +3,153 @@
...
+~~~
+### JavaScript Instantiation
+
+```js
+import {MDCButton} from '@material/button';
+
+const button = new MDCButton(document.querySelector('.mdc-button'));
+```
+
+>See [Importing the JS component](../../docs/importing-js.md) for more information on how to import JavaScript.
```
kfranqueiro: Is this intentionally a blockquote? It isn't where we currently use it in docs.

If we want it to be a blockquote, nit: space after `>`

bonniezhou: I'm gonna make it a blockquote since it's less important now that we explicitly list the import statements here.

PR #2483, opened 2018-03-28
feat(theme): Add new mdc-theme-on-primary global variable

### LINKED ISSUE #2481: Add new on-primary and on-secondary variables to mdc-theme
`mdc-theme-on-primary` is global variable clients can set to a hex value before importing any MD Web components. Various MD Web components will look to `mdc-theme-on-primary` for color, e.g. button, and top app bar.

This will replace the "text-TEXT STYLE-on-primary". Only the "primary" text style was actually used in any MD Web components, and always had 100% opacity.

We will demonstrate this `mdc-theme-on-primary` with a new "shrine" custom theme on the theme catalog page. Shrine has a pink color for primary, and a dark brown for on-primary
image

image

lynnjepsen:
Fixes #2481

BREAKING CHANGE: Removes the `--mdc-theme-text-<TEXT_STYLE>-on-<THEME_COLOR>` CSS custom properties, and the `mdc-theme--text-<TEXT_STYLE>-on-<THEME_COLOR>` CSS classes


packages/mdc-theme/_variables.scss
```diff
@@ -23,11 +23,13 @@
 //
 
 $mdc-theme-primary: #6200ee !default; // baseline purple, 500 tone
+$mdc-theme-on-primary: #fff !default;
```
kfranqueiro: Shouldn't these still be computed based on `$mdc-theme-primary` or `$mdc-theme-secondary`?

i.e. these values should adjust to remain correct when the corresponding variable is changed, which we were doing before and should still easily be able to do by calling a mixin here.

Also, are we really doing away with semitransparent text OOTB? I thought that was in spec. I thought on-primary/secondary were additional slots, not replacing text?

lynnjepsen: mdc-theme should not compute `$mdc-theme-on-primary` based on `$mdc-theme-primary` because then `$mdc-theme-on-primary` is always white/black. Mdc-theme should allow users to customize `$mdc-theme-on-primary` to any color, like the brown color used in the shrine example.

kfranqueiro: Sorry if I worded poorly; my suggestion wouldn't prevent users from customizing it, it would just provide a sane default.

We'd still have `!default` at the end to allow overriding it.

lynnjepsen: I cant actually figure how to build this functionality (without probably restructuring all our mixins in mdc-theme). I tried to do this

```
$mdc-theme-on-primary: #fff !default;
@if mdc-theme-contrast-tone($mdc-theme-primary) == "dark" {
  $mdc-theme-on-primary: #000 !default;
}
```

and commented out the `$mdc-theme-on-primary` in `theme-shrine.scss`...but it keeps setting the -on-primary color the white

kfranqueiro: That will fail because the earlier assignment will always happen, and then the latter one inside the if will never happen because `!default` makes the assignment only happen if that variable doesn't already have an assigned value. This is the exact behavior that makes redefining our globals before loading our styles work.

http://sass-lang.com/documentation/file.SASS_REFERENCE.html#variable_defaults_default

There is an `if` function that can be used akin to a ternary operator in other languages which you can use to set the value in one statement:

```scss
$mdc-theme-on-primary: if(mdc-theme-contrast-tone($mdc-theme-primary) == "dark", #000, #fff) !default;
```

Or, you can populate a map with `light` and `dark` entries and simply reference that with the value returned from contrast-tone. That is what the logic that you're replacing is _already doing_, which makes it hard for me to understand why we're removing it.

https://github.com/material-components/material-components-web/blob/v0.33.0/packages/mdc-theme/_variables.scss#L42-L63

lynnjepsen: How would users override the -on-primary with Sass without the changes in this PR?

kfranqueiro: IIUC your point is right now it's impossible to directly override, say, text-primary-on-primary, which is true. If enabling that is the goal, then I can see why this PR is changing how the variables are defined, so they can each be _directly_ overridden.

I was specifically concerned about the existing dark/light logic that was being taken out of the picture in the process of this change, and was a little confused since in this thread you seemed to be having trouble replicating the logic that already exists.

How many text color levels are we going to end up with after this is all done (compared to the current primary/secondary/hint/disabled/icon)? Would it be cleaner to reuse the existing functions that pick the defaults out of the `mdc-theme-text-colors` map, versus writing ternaries with hard-coded values on every variable we declare?

Either way, we should be able to ensure sane behavior if someone doesn't touch the on-x variables, but _does_ touch the primary/secondary values.

kfranqueiro:
Removing the entries from the map causes CSS classes to no longer be generated here:

https://github.com/material-components/material-components-web/blob/v0.33.0/packages/mdc-theme/mdc-theme.scss#L30-L36

Which causes regressions in demo pages that use `mdc-theme--x` classes to apply color.

Are we expecting to maintain classes for this as before (e.g. now we would have `mdc-theme--on-primary` and `--on-secondary`)? We'll need to enumerate the new classes to generate in mdc-theme.scss.

Here's an example of a regression, in the theme page (the 3rd and 4th buttons should have white text):

image

packages/mdc-theme/README.md
```diff
@@ -53,58 +37,58 @@ npm install @material/theme
...
+@import "@material/button/mdc-button";
 ```
 
-> **A note about `$mdc-theme-secondary`**: This variable was previously named `$mdc-theme-accent`.
-> For backward compatibility, `$mdc-theme-accent` still exists, but it is **deprecated**.
-> Apps that previously customized `$mdc-theme-accent` will continue to work, but new apps should use
-> `$mdc-theme-secondary` instead.
-
-The text color, for text placed on top of these selected theme colors, is programmatically computed based on color contrast. We follow the Web Content Accessibility Guidelines 2.0.
+We suggest you follow the Web Content Accessibility Guidelines 2.0 we picking the values for on-primary, on-secondary, etc. These values should be accessible on top of the corresponding value, e.g. primary and secondar.
```
kfranqueiro: Typos:
"we picking" -> "when picking"
"secondar." -> "secondary."

packages/mdc-theme/README.md
```diff
@@ -53,58 +37,58 @@ npm install @material/theme
...
 
 https://www.w3.org/TR/WCAG20
 
+### Advanced customization
+
+Color scheme will only get your app 80% of the way to a well designed app. Inevitably there will be some components that do not work "out of the box". To fix problems with accessbility and design, we suggest you use our SASS mixins, such as `mdc-button-filled-accessible`. For more information, consult the documentation for each component.
```
kfranqueiro: "your app" -> "you" (it's currently redundant of "well designed app" later in the sentence)
"well designed" -> "well-designed"
Typo: accessbility -> accessibility
SASS -> Sass (it's how they themselves capitalize it)

packages/mdc-theme/README.md
```diff
@@ -53,58 +37,58 @@ npm install @material/theme
...
+
+### Non-Sass customization
+
+Only a very limited number of Material Design color customization features are supported for non-Sass clients. They are a set of CSS custom properties, and a set of CSS classes
```
kfranqueiro: Add period at end of sentence

packages/mdc-theme/_variables.scss
```diff
@@ -69,18 +70,8 @@ $mdc-theme-property-values: (
...
-  text-hint-on-primary: mdc-theme-ink-color-for-fill_(hint, $mdc-theme-primary),
-  text-disabled-on-primary: mdc-theme-ink-color-for-fill_(disabled, $mdc-theme-primary),
-  text-icon-on-primary: mdc-theme-ink-color-for-fill_(icon, $mdc-theme-primary),
-  // Text-primary on "secondary" background
-  text-primary-on-secondary: mdc-theme-ink-color-for-fill_(primary, $mdc-theme-secondary),
-  text-secondary-on-secondary: mdc-theme-ink-color-for-fill_(secondary, $mdc-theme-secondary),
-  text-hint-on-secondary: mdc-theme-ink-color-for-fill_(hint, $mdc-theme-secondary),
-  text-disabled-on-secondary: mdc-theme-ink-color-for-fill_(disabled, $mdc-theme-secondary),
-  text-icon-on-secondary: mdc-theme-ink-color-for-fill_(icon, $mdc-theme-secondary),
+  on-primary: $mdc-theme-on-primary,
```
kfranqueiro: Just to confirm... are we expecting to update/prune the rest of this map (and the readme) in other imminent PRs?

demos/theme/theme-shrine.scss
```diff
@@ -16,8 +16,9 @@
 
 @import "../../packages/mdc-theme/color-palette";
 
-// TODO: Use theme mixin(s) instead
-$mdc-theme-primary: #fff;
-$mdc-theme-secondary: $material-color-orange-500;
+$mdc-theme-primary: #FCB8AB;
```
kfranqueiro: Nit: uppercase hex to lowercase hex wherever these are used, to match other styles

PR #2486, opened 2018-03-29
docs(toolbar): Add deprecation notice to readme

williamernest:
Add deprecation notice to the top of the Toolbar readme.


williamernest:
@bonniezhou I don't think the change you requested got posted.

PR #2503, opened 2018-04-04
Missing webkit-appearance

jvalst:
Missing webkit-appearance, in the demo this line is present.


kfranqueiro:
Thanks for the PR. We use autoprefixer to add vendor-prefixed styles such as this one during our build process, so its absence in the source `scss` file is intentional. If you look at https://unpkg.com/@material/select@0.34.1/dist/mdc.select.css, the vendor-prefixed properties are also present.

PR #2522, opened 2018-04-05
refactor(select): removed label and replaced with floating-label

### LINKED ISSUE #2130: Update Select to use floating label
Depends on #2126

moog16:
fixes #2130


karma.conf.js
```diff
@@ -92,6 +92,16 @@ const SL_LAUNCHERS = {
   // },
 };
 
+// Added Chrome --no-sandbox option because
+// https://github.com/travis-ci/docs-travis-ci-com/blob/
```
kfranqueiro: Shorten link to `https://github.com/travis-ci/docs-travis-ci-com/blob/c1da4af0/user/chrome.md#sandboxing` so you can fit it on one line :)

Also, I think I'm a bit confused... Doesn't Travis run our unit tests via Sauce Labs? This config affects the not-Sauce-Labs case, so how does it end up fixing it?

packages/mdc-select/README.md
```diff
@@ -164,12 +164,12 @@ Mixin | Description
...
 `mdc-select-ink-color($color)` | Customizes the color of the selected item displayed in the select.
 `mdc-select-container-fill-color($color)` | Customizes the background color of the select.
-`mdc-select-label-color($color)` | Customizes the label color of the select in the unfocused state.
+`mdc-select-label-color($color, $opacity: 1)` | Customizes the label color of the select in the unfocused state.
 `mdc-select-focused-label-color($color, $opacity: 0.87)` | Customizes the label color of the select when focused. Changing opacity for the label when floating is optional.
 `mdc-select-bottom-line-color($color)` | Customizes the color of the default bottom line of the select.
 `mdc-select-focused-bottom-line-color($color)` | Customizes the color of the bottom line of the select when focused.
 
-> NOTE: To customize label color please see the [label readme](./label/README.md).
+> NOTE: To customize label color please see the [floating label readme](./../mdc-floating-label/README.md).
```
kfranqueiro: I'm a little confused here... we say to see a separate readme for customizing label color, but there are two mixins directly above that seem to cover this?

moog16: Maybe it would be word better as:

`NOTE: To further customize label color please see...`

There is fill-color that we do not cover in the mixins above.

PR #2523, opened 2018-04-05
fix(floating-label): achieved 100% test coverage

moog16:
It has been bothering me that I didn't get to 100% test coverage before, so here we are!

### DIFF
--- a/test/unit/mdc-floating-label/mdc-floating-label.test.js
+++ b/test/unit/mdc-floating-label/mdc-floating-label.test.js
@@ -40,7 +40,7 @@ test('#shake calls the foundation shake method', () => {
   const {component} = setupTest();
   component.foundation_.shake = td.func();
   component.shake(true);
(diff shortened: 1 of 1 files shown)


test/unit/mdc-floating-label/mdc-floating-label.test.js
```diff
@@ -35,6 +36,27 @@ function setupTest() {
   return {root, component};
 }
 
+test('#shake calls the foundation shake method', () => {
+  const {component} = setupTest();
+  component.foundation_.shake = td.func();
+  component.shake(true);
+  td.verify(component.shake(true), {times: 1});
```
kfranqueiro: This is literally verifying the thing you just called. I think this is supposed to be foundation?

moog16: Ah! Yes...I'm supposed to be testing the foundation like in the subsequent tests. Thanks for catching.

PR #2539, opened 2018-04-09
feat(button): Add shape mixin to auto-adjust padding

### LINKED ISSUE #2537: Button / Shapes: Add mixin to auto-adjust padding based on angled corner size
MDC Shapes (added in #2506) adds the ability to apply masks to achieve angled corners on any unelevated component.

For MDC Button, with large enough angled corners, the text label can become a bit cramped...

image

It should be feasible to add a mixin to the `mdc-button` package to assist with angled corners, to automatically adjust padding in cases where the space gets too tight.

kfranqueiro:
Resolves #2537.

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -17,6 +17,7 @@
 @import "@material/elevation/mixins";
 @import "@material/ripple/mixins";
 @import "@material/rtl/mixins";
+@import "@material/shape/mixins";
 @import "@material/theme/functions";
 @import "@material/theme/mixins";
 @import "@material/typography/mixins";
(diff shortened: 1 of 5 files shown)


packages/mdc-button/_mixins.scss
```diff
@@ -66,6 +67,24 @@
   border-radius: $corner-radius;
 }
 
+@mixin mdc-button-angled-corner($background-color, $top-left-size, $top-right-size: $top-left-size, $bottom-right-size: $top-left-size, $bottom-left-size: $top-right-size, $stroke-width: 0) {
+  $threshold: $mdc-button-contained-horizontal-padding * 3 / 4;
```
moog16: How did you arrive at 3/4? It looks better when you have that there. What that more of a guess/check?

kfranqueiro: It was a bit of a guess / check. I started with `/ 2`, but that made the buttons look too wide. I wanted the value to be relative to the overall default padding, though.

...although this makes me realize that this should probably also depend on whether the horizontal-padding mixin is being used, which means yet another parameter to add. Not sure how I feel about this...

moog16: Gotcha...I tried a few different values for `$mdc-button-contained-horizontal-padding` and overall looks good. 

RE: more arguments: I know with JS we can pass an object. Can we do the same in Sass?

kfranqueiro: We can, and OTOH Sass also supports named parameters. My concern was just parameter overload either way.

kfranqueiro:
I've decided to abandon this for the time being because I'm concerned this mixin might be more confusing than it is useful:

* There are a lot of parameters to keep track of, due to the need to also account for `$stroke-width` (matching `mdc-button-stroke-width` if used) and potentially `$padding` (matching `mdc-button-horizontal-padding` if used), in addition to the mixin signature inherited from MDC Shape
* The auto-adjustment of padding for the extreme cases results in a button with the same content being a different size than other cases
* In design guidance, the only use cases for buttons with cuts large enough to warrant adjustment of padding are listed as "caution" or "don't" examples, so the use case is exceedingly rare and might be better off handled case-by-case rather than increasing our API surface.

PR #2542, opened 2018-04-09
fix(ripple): Re-flow logic to avoid crashing Edge

### LINKED ISSUE #2241: Edge crashes when clicking form button after filling text field

I have tested in the latest version (0.30.0) as well as the version we are currently using (0.22.0), and both versions have the issue

Only Edge is affected. It works fine in Chrome, Firefox and IE.

Microsoft Edge 41.16299.15.0
Microsoft EdgeHTML 16.16299

### What OS are you using?

Windows 10

1: Follow the first two steps of this guide to create a form: https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md

Here is the one I created from that guide:
https://codepen.io/anon/pen/MQoZRw

2: Open the pen in Edge
3: Type something in one of the of the input fields and then click the button

No crash

When clicking the button, Edge freezes for a few seconds, and then the page reloads.

Open Windows Event Viewer, in Windows Logs > Application, this error is logged:

Faulting application name: MicrosoftEdgeCP.exe, version: 11.0.16299.15, time stamp: 0x59cda7cd
Faulting module name: edgehtml.dll, version: 11.0.16299.192, time stamp: 0xf93e5331
Exception code: 0xc0000005
Fault offset: 0x00000000004451a8
Faulting process id: 0x21b0
Faulting application start time: 0x01d3a4a9b92eb23f
Faulting application path: C:\Windows\SystemApps\Microsoft.MicrosoftEdge_8wekyb3d8bbwe\MicrosoftEdgeCP.exe
Faulting module path: C:\WINDOWS\SYSTEM32\edgehtml.dll
Report Id: 6b984911-9984-4e1f-8dea-379e6e5e85b1
Faulting package full name: Microsoft.MicrosoftEdge_41.16299.15.0_neutral__8wekyb3d8bbwe
Faulting package-relative application ID: ContentProcess

If I remove data-mdc-auto-init="MDCRipple" from the button, the problem does not happen.

kfranqueiro:
Fixes #2241.

We currently have a `requestAnimationFrame` call to delay a frame before checking whether the element matches the `:active` selector, to work around issues in Chrome and Firefox. This delay causes the crash in Edge when a submit button is interacted with.

Given that the two problems are mutually exclusive in terms of affected browsers, this means we can forego the `requestAnimationFrame` delay in Edge, and avoid the crash.

### DIFF
--- a/packages/mdc-ripple/foundation.js
+++ b/packages/mdc-ripple/foundation.js
@@ -330,23 +330,41 @@ class MDCRippleFoundation extends MDCFoundation {
       this.registerDeactivationHandlers_(e);
     }
 
+    activationState.wasElementMadeActive = this.checkElementMadeActive_(e);
+    if (activationState.wasElementMadeActive) {
+      this.animateActivation_();
+    }
+
     requestAnimationFrame(() => {
-      // This needs to be wrapped in an rAF call b/c web browsers
-      // report active states inconsistently when they're called within
-      // event handling code:
-      // - https://bugs.chromium.org/p/chromium/issues/detail?id=635971
-      // - https://bugzilla.mozilla.org/show_bug.cgi?id=1293741
-      activationState.wasElementMadeActive = (e && e.type === 'keydown') ? this.adapter_.isSurfaceActive() : true;
-      if (activationState.wasElementMadeActive) {
-        this.animateActivation_();
-      } else {
+      // Reset array on next frame after the current event has had a chance to bubble to prevent ancestor ripples
+      activatedTargets = [];
+
+      if (!activationState.wasElementMadeActive && e && e.type === 'keydown' && (e.key === ' ' || e.keyCode === 32)) {
+        // If space was pressed, try again within an rAF call to detect :active, because different UAs report
+        // active states inconsistently when they're called within event handling code:
+        // - https://bugs.chromium.org/p/chromium/issues/detail?id=635971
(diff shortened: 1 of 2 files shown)


packages/mdc-ripple/foundation.js
```diff
@@ -330,25 +330,43 @@ class MDCRippleFoundation extends MDCFoundation {
...
-      // - https://bugs.chromium.org/p/chromium/issues/detail?id=635971
-      // - https://bugzilla.mozilla.org/show_bug.cgi?id=1293741
-      activationState.wasElementMadeActive = (e && e.type === 'keydown') ? this.adapter_.isSurfaceActive() : true;
-      if (activationState.wasElementMadeActive) {
-        this.animateActivation_();
-      } else {
+      // Reset array on next frame after the current event has had a chance to bubble to prevent ancestor ripples
+      activatedTargets = [];
+
+      if (!activationState.wasElementMadeActive && e && e.type === 'keydown' && (e.key === ' ' || e.keyCode === 32)) {
```
williamernest: I think we can remove `e && e.type === 'keydown'` from this statement. The only way `activationState.wasElementMadeActive ===  false` is because `(e && e.type == 'keydown') === true`.

kfranqueiro: I was a bit reluctant to assume `e` is defined here, but confirmed our unit tests will catch if this assumption ever becomes false.

PR #2543, opened 2018-04-09
feat(chips): Add animation for entry chips

bonniezhou:
Add grow/fade in animation for entry chips.


bonniezhou:
1. Whoops, the keydown event was meant for the input. Fixed.
2. Added a test for `addChip` in the component. Seems unnecessary in the foundation since it's just proxying to the adapter.

demos/chips.html
```diff
@@ -254,9 +254,8 @@ <h2>Custom theme</h2>
             entryInput.value = '';
           }
         };
-        ['click', 'keydown'].forEach(function(evtType) {
-          entryButton.addEventListener(evtType, addChip);
-        });
+        entryButton.addEventListener('click', addChip);
+        entryInput.addEventListener('keydown', addChip);
```
kfranqueiro: Oh... What you really want is a `form` containing the text field and a submit button, which would make this Just Work without necessitating any keydown handling (though you'd want to `preventDefault` on form submit instead).

I won't insist on that being done in this PR, at least this does what was originally intended.

test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -66,6 +66,15 @@ test('#destroy cleans up child chip components', () => {
   td.verify(component.chips[2].destroy());
 });
 
+test('#addChip creates and adds a new chip to the DOM', () => {
+  const root = getFixture();
+  const component = new MDCChipSet(root, undefined, (el) => new FakeChip(el));
+  component.initialSyncWithDOM();
+  component.addChip('hello world');
+  td.verify(component.foundation_.addChip('hello world'));
+  assert.isOk(component.chips.length === 4 && component.chips[3] instanceof FakeChip);
```
kfranqueiro: This could be broken up into an `assert.equal` and an `assert.instanceOf`

demos/chips.html
bonniezhou: Noted. I've been meaning to talk to a designer about making this input look better anyways, so I'll change it then.

PR #2544, opened 2018-04-09
refactor(select): use line ripple

### LINKED ISSUE #2129: Update Select to use line ripple
Depends on #1981

moog16:
fixes #2129


packages/mdc-select/_mixins.scss
```diff
@@ -76,18 +82,26 @@
 }
 
 @mixin mdc-select-bottom-line-color_($color) {
-  .mdc-select__bottom-line {
-    @include mdc-select-bottom-line-fill-color($color);
+  .mdc-line-ripple {
```
lynnjepsen: I dont think you need to set the mdc-line-ripple here

packages/mdc-select/_mixins.scss
```diff
@@ -76,18 +82,26 @@
...
+}
+
+@mixin mdc-select-focused-line-ripple-color_($color) {
+  @include mdc-select-focused-line-ripple_ {
+    @include mdc-line-ripple-color($color);
   }
 
-  .mdc-select__bottom-line::after {
-    @include mdc-select-bottom-line-fill-color($color);
+  .mdc-line-ripple::after {
```
lynnjepsen: ::after?? whats this for?

packages/mdc-select/constants.js
```diff
@@ -26,7 +26,7 @@ export const numbers = {
 
 export const strings = {
   CHANGE_EVENT: 'MDCSelect:change',
-  BOTTOM_LINE_SELECTOR: '.mdc-select__bottom-line',
+  BOTTOM_LINE_SELECTOR: '.mdc-line-ripple',
```
lynnjepsen: rename this to LINE_RIPPLE_SELECTOR?

test/unit/mdc-floating-label/mdc-floating-label.test.js
```diff
@@ -21,7 +21,7 @@ import {assert} from 'chai';
 import {MDCFloatingLabel} from '../../../packages/mdc-floating-label/index';
 
 const getFixture = () => bel`
-  <div class="mdc-floating-label"></div>
+  <label class="mdc-floating-label"></label>
```
lynnjepsen: I think this change snuck in here

packages/mdc-select/_mixins.scss
moog16: Not sure what it was even doing before. Removing the rule.

test/unit/mdc-floating-label/mdc-floating-label.test.js
moog16: This is done on purpose. You're right in that it is more related to the floating label PR, but it wasn't caught until doing changes with line-ripple.

PR #2548, opened 2018-04-10
fix(floating-label): Add `font-family: Roboto`

acdvorak:


kfranqueiro:
Is there a specific case this is fixing? Would/should this not be inherited from an ancestor?

williamernest:
I'm going to update this in my typography styles PR, so I can set the floating label to the new style `subtitle1`.

Issue #2482, opened 2018-03-28
Add new surface and on-surface variables to mdc-theme


lynnjepsen:
`mdc-theme-surface` and `mdc-theme-on-surface` are global variable clients can set to a hex value before importing any MD Web components. Various MD Web components will look to mdc-theme-surface and mdc-theme-on-surface for color, e.g. cards and chips.

We will demonstrate this `mdc-theme-surface` with a new dark theme for "shrine" on the theme catalog page.

acdvorak:
- [ ] Automatically calculate `on-surface` as `#ffffff` or `#000000`

Issue #2479, opened 2018-03-28
mdc-chip should not be translucent


lynnjepsen:
Instead calculate the background color by mixing 8% black with white.

PR #2556, opened 2018-04-11
feat(color): Add on-surface and surface to theme. 

williamernest:
Fixes: #2482


packages/mdc-card/mdc-card.scss
```diff
@@ -23,15 +23,15 @@
...
+  @include mdc-card-fill-color(surface);
   @include mdc-card-corner-radius(2px);
   @include mdc-elevation(2);
   @include mdc-card-container-layout_;
 }
 
 .mdc-card--stroked {
   @include mdc-elevation(0);
-  @include mdc-card-stroke(#dbdbdb);
+  @include mdc-card-stroke(rgba(mdc-theme-prop-value(on-surface), .12));
```
bonniezhou: This will make the stroke translucent, use the Sass `mix()` function instead:
```
mix(mdc-theme-prop-value(on-surface), mdc-theme-prop-value(surface), 12%)
```
Note: the `mix()` function takes percentages so make sure to use 12%, not .12

packages/mdc-chips/_variables.scss
```diff
@@ -14,16 +14,18 @@
 // limitations under the License.
 //
 
+@import "@material/theme/variables";
+
 $mdc-chip-border-radius-default: 16px;
-$mdc-chip-fill-color-default: rgba(black, .08);
-$mdc-chip-ink-color-default: rgba(black, .6);
+$mdc-chip-fill-color-default: rgba(mdc-theme-prop-value(on-surface), .12);
```
bonniezhou: These should all use the `mix()` function too, but that can go in another PR (to fix #2479)

packages/mdc-theme/mdc-theme.scss
```diff
@@ -24,7 +24,7 @@
 
 // Special case, so that .mdc-theme--background changes background color, not text color.
 .mdc-theme--background {
```
bonniezhou: This doesn't look like it's being used or documented anywhere so we can probably get rid of it?

williamernest: We're keeping background. I reverted changes that I made to the components other than chips/card.

packages/mdc-theme/_variables.scss
```diff
@@ -32,6 +32,9 @@ $mdc-theme-secondary: $mdc-theme-accent !default;
 $mdc-theme-on-secondary: if(mdc-theme-contrast-tone($mdc-theme-secondary) == "dark", #000, #fff) !default;
 $mdc-theme-background: #fff !default; // White
 
+$mdc-theme-surface: white !default;
```
kfranqueiro: Nit: change to #fff since we changed all the others to #fff

packages/mdc-card/mdc-card.scss
```diff
@@ -24,15 +24,15 @@
...
+  @include mdc-card-fill-color(surface);
   @include mdc-card-corner-radius(2px);
   @include mdc-elevation(2);
   @include mdc-card-container-layout_;
 }
 
 .mdc-card--stroked {
   @include mdc-elevation(0);
-  @include mdc-card-stroke($mdc-card-stroke-color);
+  @include mdc-card-stroke(mix(mdc-theme-prop-value(on-surface), mdc-theme-prop-value(surface), 12%));
```
kfranqueiro: Can we update the variable instead of replacing it here? (IIRC I added that variable for reuse with shape, at least in demos, but potentially of use to users for similar purposes as well)

Also FYI this results in changing from `#dbdbdb` to `#e0e0e0`

PR #2563, opened 2018-04-13
feat(infrastructure): Add MDC version number to compiled JS/CSS files

acdvorak:
### What?

This PR injects the MDC repo version number from `lerna.json` into the copyright header at the top of all compiled output files:

```js
/*!
 Material Components for the Web v0.35.0
 Copyright (c) 2018 Google Inc.
 License: Apache-2.0
*/
```

In order to do this, some of the release scripts needed to be refactored, and the release process has changed slightly (see below).

The new release process should be simpler and less error prone. Packages are now only published to NPM if _everything else_ succeeds, including generating the changelog and pushing changes to GitHub.

Previously, it was possible to publish packages to NPM and then have subsequent steps fail, leaving NPM and GitHub out of sync.

### How to test

1. Apply a temporary patch to prevent the scripts from _actually_ publishing anything:
    ```bash
    PATCH_URL='https://gist.githubusercontent.com/acdvorak/835a2f5f9535476def454a501bd6a117/raw/a118cf0487157ffdb50de798cfd03cbf13739366/mdc-web-pr-2563.patch'
    curl "$PATCH_URL" | git apply
    ```
2. Run:
    ```bash
    ./scripts/pre-release.sh && ./scripts/release.sh && ./scripts/post-release.sh
    ```
3. When prompted to continue, type `Y` <kbd>Enter</kbd>
4. After all 3 scripts finish, view any `.css` or `.js` file in the `build/` directory:
    ```bash
    head build/mdc.checkbox.{js,css}
    ```
5. The copyright header at the top of the file should look like this:
    ```js
    /*!
     Material Components for the Web v0.35.0
     Copyright (c) 2018 Google Inc.
     License: Apache-2.0
    */
    ```

### Why?

So you can look at any compiled `.js` or `.css` file and know exactly which version of MDC it came from.

Otherwise, there's no way to know which version you're looking at.

For example, which version of MDC is running on our [catalog site](https://material-components-web.appspot.com/)? Who knows!


scripts/webpack/plugin-factory.js
```diff
@@ -47,6 +48,11 @@ class PluginFactory {
   createCssExtractorPlugin(outputFilenamePattern) {
     return new ExtractTextPlugin(outputFilenamePattern);
   }
+
+  getProjectVersion_() {
+    const lernaJson = require('../../lerna.json');
```
kfranqueiro: Is this going to work as expected? Don't we run lerna publish which revs things _after_ running pre-release.sh which would run this code?

acdvorak: Good catch, Ken. I updated the PR to fix that by refactoring the release scripts a bit.

acdvorak:
Putting this on the back burner for now. We'll revisit it at a later date.

acdvorak:
I need to double-check that this publishes all npm packages correctly - e.g., skip `private: true`.

acdvorak:
Closing this PR. It's too risky and difficult to test. Not worth the time and effort right now.

PR #2564, opened 2018-04-13
fix(card): updated demo to have separate block element

moog16:
Moved `mdc-card__action-icons` to a separate block element, since it breaks the rules of BEM, which states you shouldn't have 2 block selectors on the same element.

### DIFF
--- a/demos/card.html
+++ b/demos/card.html
@@ -151,32 +151,34 @@
               </div>
             </div>
           </a>
-          <div class="mdc-card__actions mdc-card__action-icons">
-            <i class="mdc-icon-toggle material-icons mdc-card__action mdc-card__action--icon"
-               tabindex="0"
-               role="button"
-               aria-pressed="false"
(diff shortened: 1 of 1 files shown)


kfranqueiro:
Out of curiosity did you find a place where the BEM guidelines actually stated this? I was looking for any sign of it a day or two ago.

moog16:
I remember seeing it in a video, but I found this online: http://getbem.com/faq/#ask-your-question

```
div class="block1 block2"></div>
But the important thing about it is that both block1 and block2 should be standalone blocks. This is slightly different from what people usually mean by "global modifiers", as modifiers do not have any sense on their own and are just a set of properties to change.

<div class="block globalmod"></div>```

moog16:
After reviewing some of the guidelines and documentation, we couldn't find any thing directly saying that it was _not_ ok to mix element and block classes on the same element. So I'm closing this PR.

PR #2565, opened 2018-04-13
feat(tabs): Add tab indicator inside tab

### LINKED ISSUE #2550: Add new `MDCTab.activate`, which calls `MDCTabIndicator.activate`
...and the deactivate equivalent.

patrickrodee:
Closes #2550


packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -27,6 +27,12 @@
   }
 }
 
+@mixin mdc-tab-indicator-not-upgraded-active_ {
```
lynnjepsen: The `mdc-tab-indicator-active_` mixin is only called within the mdc-tab-indicator.scss file right? Why does the `mdc-tab-indicator-not-upgraded-active_` mixin called within the mdc-tab.scss file?

packages/mdc-tab/foundation.js
```diff
@@ -84,16 +96,20 @@ class MDCTabFoundation extends MDCFoundation {
 
   /**
    * Activates the Tab
+   * @param {!ClientRect=} previousIndicatorClientRect
```
lynnjepsen: I forget my closure syntax...but can previousIndicatorClientRect be undefined?

packages/mdc-tab/foundation.js
```diff
@@ -84,16 +96,20 @@ class MDCTabFoundation extends MDCFoundation {
...
     // Early exit
     if (this.isActive()) {
       return;
     }
+
     this.adapter_.registerEventHandler('transitionend', this.handleTransitionEnd_);
     this.adapter_.addClass(cssClasses.ANIMATING_ACTIVATE);
     this.adapter_.addClass(cssClasses.ACTIVE);
     this.adapter_.setAttr(strings.ARIA_SELECTED, 'true');
+    this.adapter_.setAttr(strings.TABINDEX, '0');
```
lynnjepsen: Why do you need to add tab-index to the tab when it is activated? Shouldnt tabs always have tab-index?

packages/mdc-tab/foundation.js
```diff
@@ -104,10 +120,21 @@ class MDCTabFoundation extends MDCFoundation {
     if (!this.isActive()) {
       return;
     }
+
     this.adapter_.registerEventHandler('transitionend', this.handleTransitionEnd_);
     this.adapter_.addClass(cssClasses.ANIMATING_DEACTIVATE);
     this.adapter_.removeClass(cssClasses.ACTIVE);
     this.adapter_.setAttr(strings.ARIA_SELECTED, 'false');
+    this.adapter_.setAttr(strings.TABINDEX, '-1');
```
lynnjepsen: -1 for deactivate? See question above.

packages/mdc-tab/index.js
```diff
@@ -53,32 +74,62 @@ class MDCTab extends MDCComponent {
...
+  }
+
+  /**
+   * @return {!{
+   *   activateIndicator: function(!ClientRect): undefined,
+   *   deactivateIndicator: function(): undefined,
+   *   computeIndicatorClientRect: function(): !ClientRect,
+   * }}
+   */
+  getIndicatorAdapterMethods_() {
```
lynnjepsen: I dont think you need to make this a separate method. Just move these three methods into the adapter class above.

packages/mdc-tab/mdc-tab.scss
```diff
@@ -97,3 +123,11 @@
   @include mdc-tab-text-label-color(primary);
   @include mdc-tab-icon-color(primary);
 }
+
+.mdc-tab--active .mdc-tab-indicator {
+  @include mdc-tab-indicator-not-upgraded-active_;
+}
+
+.mdc-tab:not(.mdc-tab--stacked) .mdc-tab__icon + .mdc-tab__text-label {
```
lynnjepsen: this CSS selector is so complicated! whats it doing?

packages/mdc-tab-indicator/_mixins.scss
patrickrodee: `mdc-tab-indicator-not-upgraded-active` makes the active tab indicator visible until the tab indicator is upgraded and initialized. I'll make it non-private and rename it while also removing `mdc-tab-indicator-active` as it serves no real purpose.

packages/mdc-tab/foundation.js
patrickrodee: It's optional. The `=` makes it optional (undefined) where as the `?` makes it nullable. We want it to be optional in this case.

packages/mdc-tab/mdc-tab.scss
patrickrodee: I simplified this.

PR #2571, opened 2018-04-16
feat(chips): Allow close icon and exit animation

bonniezhou:


packages/mdc-chips/chip/index.js
```diff
@@ -125,6 +125,9 @@ class MDCChip extends MDCComponent {
       notifyInteraction: () => this.emit(strings.INTERACTION_EVENT, {chip: this}, true /* shouldBubble */),
       notifyTrailingIconInteraction: () => this.emit(
         strings.TRAILING_ICON_INTERACTION_EVENT, {chip: this}, true /* shouldBubble */),
+      getComputedStyleValue: (propertyName) => window.getComputedStyle(this.root_).getPropertyValue(propertyName),
+      setStyleProperty: (propertyName, value) => this.root_.style.setProperty(propertyName, value),
+      removeFromDOM: () => this.root_.remove(),
```
kfranqueiro: This won't work in IE. You're going to have to do `this.root_.parentNode.removeChild(this.root_)`

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -44,6 +44,15 @@
   }
 }
 
+.mdc-chip--exit {
+  transition:
+    opacity 75ms $mdc-animation-standard-curve-timing-function,
+    width 150ms $mdc-animation-deceleration-curve-timing-function,
```
kfranqueiro: Does this need to be width/padding/margin or can we do a scale transform instead, to take advantage of GPU compositing and eliminate potential for text truncation/wrapping issues?

I'm also a tad confused why we have the non-opacity transitions taking longer than the opacity (i.e. the element will already be invisible at that point).

bonniezhou: Discussed offline - transitioning the transform would be extremely complicated due to chips being wrapped to the next line.

packages/mdc-chips/chip/foundation.js
```diff
@@ -121,11 +124,36 @@ class MDCChipFoundation extends MDCFoundation {
 
   /**
    * Handles a transition end event on the root element.
-   * This is a proxy for handling a transition end event on the leading icon or checkmark,
-   * since the transition end event bubbles.
    * @param {!Event} evt
    */
   handleTransitionEnd_(evt) {
+    // Handle transition end event on the chip when the it is about to be removed.
```
kfranqueiro: "when the it" -> "when it"

packages/mdc-chips/chip/foundation.js
```diff
@@ -121,11 +124,36 @@ class MDCChipFoundation extends MDCFoundation {
...
+    if (this.adapter_.eventTargetHasClass(/** @type {!EventTarget} */ (evt.target), cssClasses.CHIP_EXIT)) {
+      if (evt.propertyName === 'width') {
+        this.adapter_.removeFromDOM();
```
kfranqueiro: Should we also be calling `this.destroy()` here? AFAIK we're never destroying these chips even though they're effectively removed.

Also, just checking, chip set isn't holding any references to entry chips, is it? (I think it only holds references to selected chips for filter/choice.)

bonniezhou: Oh that's a really good point, the chip set does hold a reference to all its chips, regardless of the type of chip set.

I refactored it so that the chip set is in charge of removing the chip object from its reference, and also calls `remove` on the chip, which will destroy and remove itself from the DOM.

Note that this means the chip set foundation has to pass a chip object to its adapter, and the chip object needs to have an explicit `remove` method.

packages/mdc-chips/README.md
```diff
@@ -217,6 +219,9 @@ Method Signature | Description
 `deregisterTrailingIconInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener on the trailing icon element
 `notifyInteraction() => void` | Emits a custom event `MDCChip:interaction` denoting the chip has been interacted with
 `notifyTrailingIconInteraction() => void` | Emits a custom event `MDCChip:trailingIconInteraction` denoting the chip's trailing icon has been interacted with
+`notifyRemoval() => void` | Emits a custom event `MDCChip:removal` denoting the chip will be removed
```
kfranqueiro: Is there a reason you preferred removal over remove (or removed) here?

packages/mdc-chips/README.md
```diff
@@ -229,6 +234,7 @@ Method Signature | Description
 `deregisterInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event handler on the root element for a given event
 `createChipElement(text: string, leadingIcon: Element, trailingIcon: Element) => Element` | Returns a chip element with the given text, leading icon, and trailing icon
 `appendChild(el: Element) => void` | Appends the given element as a child of the root element
+`removeChip(chip: MDCChip) => void` | Removes the chip object from the chip set
```
kfranqueiro: Do you think it would make sense to combine `createChipElement` and `appendChild` into a single `appendChip` API, to mirror `removeChip`? (I think those other two are only ever used together back to back anyway?)

test/unit/mdc-chips/mdc-chip.test.js
```diff
@@ -49,6 +49,16 @@ test('get ripple returns MDCRipple instance', () => {
   assert.isTrue(component.ripple instanceof MDCRipple);
 });
 
+test('#remove removes the root element from the DOM', () => {
```
kfranqueiro: Is there a way for us to also test that `destroy` is run?

test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -139,3 +140,11 @@ test('#adapter.appendChild adds a child to the chip set element', () => {
   component.getDefaultFoundation().adapter_.appendChild(chipEl);
   assert.equal(root.childNodes[3], chipEl);
 });
+
+test('#adapter.removeChip removes the chip object from the chip set', () => {
+  const root = getFixture();
+  const component = new MDCChipSet(root, undefined, (el) => new FakeChip(el));
+  const chip = component.chips[0];
+  component.getDefaultFoundation().adapter_.removeChip(chip);
+  td.verify(chip.remove());
```
kfranqueiro: Also veriy that component.chips' length is 0?

packages/mdc-chips/README.md
bonniezhou: Yeah that's a really good idea

packages/mdc-chips/README.md
bonniezhou: Just using a noun to mirror "interaction". Is "remove" more understandable?

test/unit/mdc-chips/mdc-chip-set.test.js
bonniezhou: Done. (It should have length 2 since we're only removing 1 chip)

packages/mdc-chips/README.md
kfranqueiro: Oh, I guess that's true. I was thinking of e.g. "selected" on some other components. I guess removal works here.

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -83,8 +91,7 @@ class MDCChipSetFoundation extends MDCFoundation {
    * @return {!Element}
    */
   addChip(text, leadingIcon, trailingIcon) {
-    const chipEl = this.adapter_.createChipElement(text, leadingIcon, trailingIcon);
-    this.adapter_.appendChild(chipEl);
+    const chipEl = this.adapter_.appendChip(text, leadingIcon, trailingIcon);
```
kfranqueiro: Do these still need to return the chip element? Would it make more sense to return the chip itself, or nothing at all?

kfranqueiro: Oops, nevermind, forgot the component needs it.

bonniezhou: The foundation needs to return the chip element so the component can push it to its array of chip references.

test/unit/mdc-chips/mdc-chip-set.foundation.test.js
```diff
@@ -184,3 +185,20 @@ test('in filter chips, on custom MDCChip:interaction event deselects selected ch
   td.verify(chipA.foundation.setSelected(false));
   assert.equal(foundation.selectedChips_.length, 0);
 });
+
+test('on custom MDCChip:removal event removes chip', () => {
```
kfranqueiro: Do we not have a test in this file for addChip? (We're testing it at the component level, but IIUC that's mocking the foundation, so we should be testing it here too.)

test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -125,17 +126,20 @@ test('#adapter.deregisterInteractionHandler removes a handler from the root elem
   td.verify(handler(td.matchers.anything()), {times: 0});
 });
 
-test('#adapter.createChipElement returns a new chip element', () => {
-  const {component} = setupTest();
-  const chipEl = component.getDefaultFoundation().adapter_.createChipElement('hello world');
+test('#adapter.appendChip add a new chip to the chip set element', () => {
```
kfranqueiro: add -> adds

PR #2572, opened 2018-04-16
chore(package): update google-closure-compiler to version 20180402.0.0

acdvorak:
Closes #2480


acdvorak:
CLA bot doesn't like commits from greenkeeper... So I'll create a fresh PR for this.

PR #2574, opened 2018-04-16
fix(select): Override floating label properties in select box

williamernest:
Fixes an issue with typography and label behavior using the new styles and new floating label.

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -67,8 +67,8 @@
     }
 
     width: 100%;
-    padding-top: 22px;
-    padding-bottom: 8px;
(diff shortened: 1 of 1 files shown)


packages/mdc-select/mdc-select.scss
```diff
@@ -114,12 +114,8 @@
     background-position: left 10px center;
   }
 
-  .mdc-select__native-control,
-  .mdc-select__label {
-    @include mdc-typography(subtitle1);
-  }
-
   .mdc-select__native-control {
+    @include mdc-typography(subtitle1);
```
kfranqueiro: I'm guessing this was removed from label to fix line-height, but do we not want other properties of this style?

PR #2589, opened 2018-04-18
docs(select): Update README to new format

kfranqueiro:
This also updates the README to follow our new structure, since I needed to document the necessary imports now anyway.

BREAKING CHANGE: mdc-floating-label and mdc-line-ripple styles must now be explicitly imported


kfranqueiro:
Disregard the branch name; since we decided to move forward with #2578 instead, I've repurposed this PR to simply update the Select README to the new format, since I had done it in this PR anyway.

packages/mdc-select/README.md
```diff
@@ -58,7 +58,19 @@ npm install @material/select
 </div>
 ```
 
-Then with JS
+### Styles
+
+```scss
+@import "@material/select/mdc-select";
+
+// Invoke mixins in the context of a custom class applied to the component's root element:
```
bonniezhou: Is this necessary to get the default select? If not, it doesn't need to go under Basic Usage.

packages/mdc-select/README.md
```diff
@@ -148,6 +171,8 @@ Since MDC Select uses native `<select>` and `<option>` elements, simply add the
 </div>
 ```
 
+## Style Customization
+
 #### CSS Classes
 
 | Class                    | Description                                     |
```
bonniezhou: I think these were missing from before: `mdc-select--disabled`, `mdc-select__native-control`

packages/mdc-select/README.md
kfranqueiro: I had done this as a replacement for the remark about including the mixins within a `.foo-select` custom class selector in the Sass mixins section, but I can put that back.

PR #2590, opened 2018-04-18
fix(floating-label): Update transition durations

### LINKED ISSUE #2561: Incorrect transition timing for floating label scale/translate
One of our designers noticed that some of our text fields looked like they were animating their labels slower than others.

Currently, floating-label defaults to 180ms duration, and we're overriding it for outlined text field to 260ms.

All three types of text fields should be transitioning at 150ms for each direction.

kfranqueiro:
Fixes #2561.

Also fixes notched-outline transition durations to match, since both occur at the same time in outlined text fields.

BREAKING CHANGE: Removes the (undocumented) mdc-floating-label-transition function


packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -53,11 +55,9 @@
 .mdc-notched-outline__path {
```
bonniezhou: Not sure if you want to change it here or another PR, but line 50 needs to be updated: change `opacity 100ms ease` to use the new duration variable and standard curve timing.
Also include a transition for `border-color` for the transition on hover.

bonniezhou: In the `.mdc-notched-outline__idle` selector (GitHub won't let me comment there)

kfranqueiro: I updated the branch with all of the suggested changes. Given the brevity of the overall duration, this seems like it might actually be fine (and even if I slow it down in dev tools it still doesn't seem like the half-faded line and the text don't end up intersecting too badly).

As much as I'd like to avoid transitioning on non-compositable properties, adding border-color fixes another issue I'd noticed, so thanks for pointing that out.

PR #2619, opened 2018-04-20
fix(chips): Rename all entry chips to input chips

bonniezhou:
Design decision: change name from "entry" to "input" for chips which enable user input by converting text into chips.

### DIFF
--- a/demos/chips.html
+++ b/demos/chips.html
@@ -54,16 +54,16 @@
       </section>
 
       <section class="example">
-        <div id="entry-chip-icon-clones">
-            <i class="entry-leading-icon material-icons mdc-chip__icon mdc-chip__icon--leading">face</i>
-            <i class="entry-trailing-icon material-icons mdc-chip__icon mdc-chip__icon--trailing" tabindex="0" role="button">cancel</i>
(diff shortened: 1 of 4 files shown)


demos/chips.html
```diff
@@ -235,26 +235,26 @@ <h2>Custom theme</h2>
     <script src="/assets/material-components-web.js" async></script>
     <script>
       demoReady(function() {
-        var entryChipSet = document.getElementById('entry-chip-set');
-        var chipSets = document.querySelectorAll('.mdc-chip-set:not(#entry-chip-set)');
-        var entryInput = document.getElementById('entry-chip-set-input');
-        var entryButton = document.getElementById('entry-chip-set-button');
+        var inputChipSet = document.getElementById('input-chip-set');
+        var chipSets = document.querySelectorAll('.mdc-chip-set:not(#input-chip-set)');
+        var entryInput = document.getElementById('input-chip-set-input');
```
moog16: I think rename this to just `input`

bonniezhou:
The keyframe doesn't need to be renamed because it refers to the animation when the chip "enters" the DOM. It just so happens that the animation is only used on input chips.

PR #2632, opened 2018-04-23
fix(button): Rename stroke to outline

williamernest:
BREAKING CHANGE: Renames variant, classes and mixins containing `stroke` to use `outline`.

### DIFF
--- a/demos/shape.html
+++ b/demos/shape.html
@@ -46,7 +46,7 @@
           <div class="mdc-shape-container__corner mdc-shape-container__corner--bottom-left"></div>
         </div>
         <div class="mdc-shape-container four-corner-container four-corner-container--stroked">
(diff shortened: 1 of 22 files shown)


demos/shape.html
```diff
@@ -46,7 +46,7 @@
           <div class="mdc-shape-container__corner mdc-shape-container__corner--bottom-left"></div>
         </div>
         <div class="mdc-shape-container four-corner-container four-corner-container--stroked">
-          <button class="mdc-button mdc-button--stroked">Stroked Button</button>
+          <button class="mdc-button mdc-button--outlined">Stroked Button</button>
```
kfranqueiro: The button caption here still says Stroked

williamernest: Yes, this was updated in the shape PR.

PR #2635, opened 2018-04-23
fix(chips): Rename stroke to outline

williamernest:
BREAKING CHANGE: Renames variant, classes and mixins containing the word stroke to use the word outline.

### DIFF
--- a/packages/mdc-chips/README.md
+++ b/packages/mdc-chips/README.md
@@ -163,10 +163,10 @@ Mixin | Description
 `mdc-chip-fill-color($color)` | Customizes the background fill color for a chip
 `mdc-chip-ink-color($color)` | Customizes the text ink color for a chip, and updates the chip's ripple color to match
 `mdc-chip-selected-ink-color($color)` | Customizes text ink and ripple color of a chip in the _selected_ state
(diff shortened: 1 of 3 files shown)


packages/mdc-chips/README.md
```diff
@@ -163,10 +163,10 @@ Mixin | Description
 `mdc-chip-fill-color($color)` | Customizes the background fill color for a chip
 `mdc-chip-ink-color($color)` | Customizes the text ink color for a chip, and updates the chip's ripple color to match
 `mdc-chip-selected-ink-color($color)` | Customizes text ink and ripple color of a chip in the _selected_ state
-`mdc-chip-stroke($width, $style, $color)` | Customizes the border stroke properties for a chip
-`mdc-chip-stroke-width($width)` | Customizes the border stroke width for a chip
-`mdc-chip-stroke-style($style)` | Customizes the border stroke style for a chip
-`mdc-chip-stroke-color($color)` | Customizes the border stroke color for a chip
+`mdc-chip-outline($width, $style, $color)` | Customizes the border outline properties for a chip
```
kfranqueiro: This sounds weird saying "border outline" everywhere...  And I'm not sure how I feel about saying just "outline" because that is _not_ the CSS property this is controlling, but we should definitely pick one.

PR #2643, opened 2018-04-25
fix(top-app-bar): Change margin-top to padding-top to prevent margin collapse

williamernest:
Fixes an issue with the `fixed-adjust` classes being susceptible to margin collapsing by changing `margin-top` to `padding-top`


kfranqueiro:
Is the idea behind this to effectively "carry over" e.g. body margin so that it applies after / in addition to the adjustment? It took me a few minutes to understand why we would not want margin collapsing.

williamernest:
Margin collapsing causes the top app bar to initially render 64px from the top of the page when using the `mdc-top-app-bar--fixed-adjust` class on a sibling element unless you use `&nbsp;` to separate them.

cheng93:
Media queries also need to changed

https://github.com/material-components/material-components-web/blob/master/packages/mdc-top-app-bar/_mixins.scss#L81

PR #2646, opened 2018-04-25
feat(infrastructure): Limit # of parallel screenshot requests to 5

acdvorak:
### How to test

```bash
export CBT_USERNAME='...'
export CBT_AUTHKEY='...'

npm run screenshot:build
npm run screenshot:test
```

### What it does

Avoids sending unnecessary traffic to CrossBrowserTesting.com by limiting the number of concurrent screenshot requests to 5 (the max allowed by our plan). This prevents us from making pointless API requests that we know will fail.

However, note that we will still hit the plan limit when two or more developers run screenshot tests at the same time. That use case is handled by `reachedParallelExecutionLimit()` in `test/screenshot/lib/screenshot.js`, which periodically retries the screenshot request until it either succeeds or times out.


test/screenshot/lib/screenshot.js
```diff
@@ -48,18 +52,33 @@ const BROWSERS = [
...
-    .then(
-      async (captureResponseBody) => handleCaptureResponse(testPageUrl, captureResponseBody),
-      async (err) => rejectWithError('captureOne', testPageUrl, err)
-    );
+  return requestQueue.enqueue(testPageUrl)
+    .then(() => {
+      return sendCaptureRequest(testPageUrl)
+        .then(
+          (captureResponseBody) => handleCaptureResponse(testPageUrl, captureResponseBody),
+          (err) => rejectWithError('captureOne', testPageUrl, err)
```
patrickrodee: Should probably dequeue the capture request in the error handlers unless the whole process exits on failure.

acdvorak: Good catch! I was totally planning to do that, but clearly I forgot 😆 Thanks!!

PR #2672, opened 2018-04-27
fix(checkbox): Fix visibility in Windows high-contrast mode

patrickrodee:


packages/mdc-radio/mdc-radio.scss
```diff
@@ -89,6 +89,8 @@
     height: 100%;
     transform: scale(0, 0);
     transition: mdc-radio-exit(transform), mdc-radio-exit(background-color);
+    border-width: 10px;
```
kfranqueiro: Did you experiment with updating the transition above this to use border-color rather than background-color?

PR #2676, opened 2018-04-27
docs(form-field): Update README

bonniezhou:


packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,95 @@ path: /catalog/input-controls/form-fields/
 
 # Form Fields
 
-MDC Form Field provides an `mdc-form-field` helper class for easily making theme-aware, RTL-aware
-form field + label combos. It also provides an `MDCFormField` class for easily making input ripples
-respond to label events.
+MDC Form Field aligns a form field (for example, a checkbox) with its label, and makes it RTL-aware and [theme](../mdc-theme)-aware.
```
kfranqueiro: What does form-field do with regards to mdc-theme? I see it applying typography styles but not theme...

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,95 @@ path: /catalog/input-controls/form-fields/
...
-      <svg class="mdc-checkbox__checkmark"
-           viewBox="0 0 24 24">
-        <path class="mdc-checkbox__checkmark-path"
-              fill="none"
-              stroke="white"
-              d="M1.73,12.91 8.1,19.28 22.79,4.59"/>
-      </svg>
-      <div class="mdc-checkbox__mixedmark"></div>
-    </div>
+    ...
```
kfranqueiro: I'd argue the inner content of the checkbox is somewhat important for correlating the `id` attribute of the input with the `for` atatribute of the label?

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,95 @@ path: /catalog/input-controls/form-fields/
...
-```
+### JavaScript Instantiation
 
-### Instantiation
+If you are using MDC Form Field with an MDC Web component that has a [ripple](../mdc-ripple) effect, you can instantiate `MDCFormField` and set its [`input` property](#MDCFormField-properties-and-methods) to activate the ripple effect upon interacting with the label. Here is an example using [MDC Radio](../mdc-radio):
 
-```javascript
+```js
 import {MDCFormField} from '@material/form-field';
+import {MDCFormField} from '@material/radio';
```
kfranqueiro: MDCFormField -> MDCRadio

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,95 @@ path: /catalog/input-controls/form-fields/
...
+### Label position
 
-Read-write property that works with an instance of an MDC Web input element.
+By default, the input will be positioned before the label. You can position the input after the label by adding the `mdc-form-field--align-end` class:
 
-In order for the label ripple integration to work correctly, this property needs to be set to a
-valid instance of an MDC Web input element which exposes a `ripple` getter.
+```html
+<div class="mdc-form-field mdc-form-field--align-end">
+  <input type="checkbox" id="input">
```
kfranqueiro: Use MDC Checkbox markup?

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,95 @@ path: /catalog/input-controls/form-fields/
...
-const radio = new MDCRadio(document.querySelector('.mdc-radio'));
+## `MDCFormField` Properties and Methods
 
-formField.input = radio;
-```
+Property | Value Type | Description
+--- | --- | ---
+`input` | String | Gets and sets the form field input. 
+
+In order for the label ripple integration to work correctly, this property needs to be set to a valid instance of an MDC Web input element which exposes a `ripple` getter. No action is taken if the `input` property is not set or the input instance doesn't expose a `ripple` getter.
```
kfranqueiro: "this property" -> "the `input` property" (this sentence used to be in the context of its heading but is now in the context of the table above)

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,95 @@ path: /catalog/input-controls/form-fields/
...
 
-No action is taken if the `input` property is not set or the input instance doesn't expose a
-`ripple` getter.
+## Usage within Web Frameworks
 
+If you are using a JavaScript framework, such as React or Angular, you can create a Form Field for your framework. Depending on your needs, you can use the _Simple Approach: Wrapping MDC Web Vanilla Components_, or the _Advanced Approach: Using Foundations and Adapters_. Please follow the instructions [here](../../docs/integrating-into-frameworks.md).
 
-### Adapter
+### `MDCFormFieldAdapter`
```
kfranqueiro: I'm not sure why the table below is empty, but form field's adapter actually has 4 APIs. Can you fill this in?

bonniezhou: Oh there's stuff at the bottom, it's just collapsed on the GitHub PR diff view because nothing changed.

kfranqueiro: Oh, whoops, dunno how I didn't notice the fold.

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,103 @@ path: /catalog/input-controls/form-fields/
...
-By default, this will position the label after the input. You can change this behavior using the
-`align-end` modifier class.
+### HTML Structure
+
+The `mdc-form-field` class is used as a parent element, with any combination of adjacent `input` and `label` elements as
+immediate children:
 
 ```html
-<div class="mdc-form-field mdc-form-field--align-end">
+<div class="mdc-form-field">
```
kfranqueiro: I apparently failed to notice this before... Should we just remove this first example and combine the first and next paragraph, jumping straight to the part that talks about using it with MDC Checkbox and Radio?

"MDC Form Field operates as a parent element and works with any type of immediate child element whose successive sibling is a `label` element. This means ..."

packages/mdc-form-field/README.md
```diff
@@ -7,141 +7,91 @@ path: /catalog/input-controls/form-fields/
...
+> _NOTE_: MDC Form Field is **not** intended for cases where a label and input are already handled together in a component's styles and logic. For example, [MDC Text Field](../mdc-textfield) already manages a label and input together under its own root element.
 
-```javascript
-const MDCFormField = mdc.formField.MDCFormField;
-const MDCFormFieldFoundation = mdc.formField.MDCFormFieldFoundation;
-```
+### JavaScript Instantiation
 
-### Instantiation
+If you are using MDC Form Field with an MDC Web component that has a [ripple](../mdc-ripple) effect, you can instantiate `MDCFormField` and set its [`input` property](#MDCFormField-properties-and-methods) to activate the ripple effect upon interacting with the label. Here is an example with [MDC Checkbox](../mdc-checkbox):
```
kfranqueiro: Change the anchor on the input property link to all-lowercase (`#mdcformfield-properties-and-methods`) to make it work

PR #2685, opened 2018-04-30
chore(infrastructure): Allow testing site generator execution

kfranqueiro:
Since we're doing quite a few README updates recently, it makes sense to flag any obvious errors in CI.


kfranqueiro:
This is currently failing because it successfully caught #2686. I'll update once that's merged.

PR #2690, opened 2018-05-01
fix(theme): Add error color. Update text-field to use new error color

williamernest:
Adds a new color to the theme for `error` and `on-error`. Update text-field variable to use it.


packages/mdc-theme/README.md
```diff
@@ -85,8 +89,12 @@ CSS Class | Description
 `mdc-theme--primary` | Sets the text color to the theme primary color
 `mdc-theme--secondary` | Sets the text color to the theme secondary color
 `mdc-theme--background` | Sets the background color to the theme background color
+`mdc-theme--error` | Sets the error color to the theme error color
```
kfranqueiro: You have surface before error above; want to do the same here and in instances below?

PR #2692, opened 2018-05-02
fix(fab): Change fab to inline-block. Move material-icons class to icon

williamernest:
Fixes icon alignment issues with IE11. The icon was rendered on the right side of the fab. 

Also, the demos were putting the `material-icons` class on the button instead of on the icon class, which causes IE11 to align the icons higher than it should be.

### DIFF
--- a/demos/fab.html
+++ b/demos/fab.html
@@ -103,8 +103,8 @@
     <main>
       <div class="mdc-toolbar-fixed-adjust"></div>
       <section class="hero">
-        <button class="mdc-fab material-icons" aria-label="Favorite">
-          <span class="mdc-fab__icon">
+        <button class="mdc-fab " aria-label="Favorite">
+          <span class="material-icons mdc-fab__icon">
             favorite_border
           </span>
         </button>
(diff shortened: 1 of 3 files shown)


demos/fab.html
```diff
@@ -143,8 +143,8 @@
             </figcaption>
           </figure>
           <figure>
-            <button class="mdc-fab material-icons" aria-label="Favorite">
-              <svg xmlns="http://www.w3.org/2000/svg" class="mdc-fab__icon" viewBox="0 0 24 24" fill="#000000">
+            <button class="mdc-fab" aria-label="Favorite">
+              <svg xmlns="http://www.w3.org/2000/svg" class="material-icons mdc-fab__icon" viewBox="0 0 24 24" fill="#000000">
```
kfranqueiro: The SVG variants shouldn't even need the material-icons class, right?

packages/mdc-fab/_mixins.scss
```diff
@@ -58,7 +58,7 @@ $mdc-fab-icon-enter-duration_: 180ms;
   @include mdc-ripple-radius-bounded;
   @include mdc-elevation(6);
 
-  display: inline-flex;
+  display: inline-block;
```
kfranqueiro: This seems like a heavy-handed change just to fix IE 11. Does Button not have a similar issue? That uses inline-flex as well.

If we change this, `justify-content` no longer has any meaning and I suspect we should add `text-align: center`, but this further confuses me as to how this is fixing something without breaking anything else.

kfranqueiro:
This change causes the icon to be vertically misaligned in Chrome (and vertical alignment is most likely the reason we were using inline-flex all along):

fab html

PR #2696, opened 2018-05-03
fix(floating-label): Add @noflip annotation to floating label

### LINKED ISSUE #2695: mdc-floating-label does not work with RTL post processors
** PLEASE READ THIS BEFORE FILING AN ISSUE **

## Bugs

Follow the template below **and include a codepen** to ensure the quickest and most accurate response to your issue.

HEAD

All

### What OS are you using?

Mac

Use mdc-floating-label.scss with an RTL post processor like CSS Janus, etc.
Load the input with dir="rtl" in the body.

Label will look correct in RTL mode.

Label is slightly out of place like so: 
image

I have a pull request for this.

TedSander:
Add @noflip annotation so that post RTL CSS processors do not change the values since they are already flipped for RTL.

Closes #2695


packages/mdc-floating-label/mdc-floating-label.scss
```diff
@@ -40,8 +40,11 @@
     cursor: text;
 
     @include mdc-rtl {
+      /* @noflip */
       right: 0;
+      /* @noflip */
```
aprigogin: extra line above

packages/mdc-floating-label/mdc-floating-label.scss
```diff
@@ -40,8 +40,11 @@
...
       left: auto;
+      /* @noflip */
```
aprigogin: extra line above

packages/mdc-floating-label/mdc-floating-label.scss
TedSander: Done

packages/mdc-floating-label/mdc-floating-label.scss
TedSander: Done

TedSander:
Ping is the tree open again?

PR #2704, opened 2018-05-04
fix(top-app-bar): Replace margin-top in media query with padding-top

### LINKED ISSUE #2712: Top App Bar spacing issue
mdc-top-app-bar--fixed-adjust and all other related classes should be margin-top not padding-top. It is conflicting with the media queries.

williamernest:
Missed a change to mobile `fixed-adjust` class with #2643 

Resolves #2712.


alex-page:
If anyone is experiencing this issue and needs a temporary patch they can do this:

```
@media (max-width: 599px){
	.mdc-top-app-bar--fixed-adjust {
		margin-top: 0;
		padding-top: $mdc-top-app-bar-mobile-row-height;
	}
}
```

PR #2715, opened 2018-05-08
fix(fab): Restore horizontal alignment in IE11

kfranqueiro:
Also fixes markup to move `material-icons` to the icon element, as pointed out by Will in #2692.


williamernest:
It's better than it was, and addresses the issue with GM regarding the icon being on the right, but it's still off in IE11. 

We might want to just accept these changes as they are now, to fix the immediate regression, but file a bug to fix this in IE11 once and for all. 

Before
before

After: 
after

The icon on the fab in the bottom right was centered in the second picture.

kfranqueiro:
I can file a bug for the vertical alignment. AFAICT, that vertical alignment was a thing (at least according to the in-repo demos and screenshot tests) even prior to the changes to support SVGs in FABs.

The baffling part, though, is that if I throw FABs into a [codepen](https://codepen.io/kfranqueiro/pen/NMLRgv) running 0.34.1, they align fine in IE 11...  but in our demos on 0.34.1, they don't. Removing normalize.css makes no difference.

PR #2719, opened 2018-05-09
feat(infrastructure): More reliable auto-cropping of screenshot images

acdvorak:
### How to test

```bash
export CBT_USERNAME='...'
export CBT_AUTHKEY='...'

npm run screenshot:build
npm run screenshot:test
```

### What it does

* Changes the layout of test pages
    - There are now only 2 columns of buttons
    - This makes it easier to compare desktop vs. mobile rendering
* Gives each "cell" a specific height/width
    - If a button's size or margin changes, it will no longer affect the positioning of other buttons
    - This makes diffs less noisy and easier to review
* Adds a background color of `#333333` to the `<html>` element
    - The `<main>` element still has a background of `#ffffff`
    - This makes it easier to automatically crop the image
* Replaces [Jimp's unreliable `autocrop()`](https://github.com/oliver-moran/jimp#basic-methods) with custom auto-cropping logic
    - The top and bottom are cropped first, followed by left and right
    - Starting from the outer edges of the image, rows and columns in which at least 95% of the pixels are the page background color (`#333333`) get removed
    - The right side is cropped more loosely (67% instead of 95%) to remove [Edge's promotional notification popovers](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/08/20_40_45_142/c6cc25f87/mdc-button/classes/baseline.html.win10e17_edge17.png)
* Changes the directory structure and URL format of files stored in GCS
    - Instead of two separate directories for `assets` (HTML/CSS/JS) and `screenshots` (PNGs), captured images have the same path as the HTML file they were captured from, but with a suffix appended
    - Before:
        * [`.../<commit>/assets/mdc-button/classes/baseline.html`](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/07/22_52_10_643/1de2aaf96/assets/mdc-button/classes/baseline.html)
        * [`.../<commit>/screenshots/mdc-button/classes/baseline.html/win10e15_chrome66x64.png`](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/07/22_52_10_643/1de2aaf96/screenshots/mdc-button/classes/baseline.html/win10e15_chrome66x64.png)
    - After:
        * [`.../<commit>/mdc-button/classes/baseline.html`](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/08/23_53_42_405/72d9999e9/mdc-button/classes/baseline.html)
        * [`.../<commit>/mdc-button/classes/baseline.html.win10e15_chrome66x64.png`](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/08/23_53_42_405/72d9999e9/mdc-button/classes/baseline.html.win10e15_chrome66x64.png)
    - This makes it easier to view the page that a screenshot was generated from (you simply remove everything after `.html` from the URL)
* Reorders the buttons with font icons and SVG icons
    - This makes it easier to compare the positioning of font vs. SVG icons

#### Before

image

#### After

image


test/screenshot/lib/controller.js
```diff
@@ -252,17 +252,109 @@ class Controller {
...
-              });
+              })
+            ;
           });
         },
         (err) => Promise.reject(err)
       );
   }
 
+  getCropMatches_(image) {
```
patrickrodee: Add a function annotation here so it's clear that `image` is a Jimp image object

test/screenshot/lib/controller.js
```diff
@@ -252,17 +252,109 @@ class Controller {
...
+
+    for (const row of rows) {
+      if (matchPercentage(row) > HIGH_MATCH_PERCENTAGE) {
+        amounts.top++;
+      } else {
+        break;
+      }
+    }
+
+    for (const row of rows.concat().reverse()) {
```
patrickrodee: Are you using `.concat()` as a cheap array cloning method since `.reverse()` modifies the array in-place?

test/screenshot/lib/controller.js
```diff
@@ -252,17 +252,109 @@ class Controller {
...
+      }
+      const isMatch = trimColors.includes(image.getPixelColor(x, y));
+      rows[y][x] = isMatch;
+      cols[x][y] = isMatch;
+    });
+
+    return {rows, cols};
+  }
+
+  getCropRect_({rows, cols}) {
```
patrickrodee: This is pretty clever. However, I wonder if something similar could be accomplished with more certainty if we had a right and bottom border that's 1px solid #abc123 or something similarly obvious and outside of the Material scope. Cropping could then rely on finding the column with that color and making sure the subsequent rows at that column have that color. If we find one that doesn't, we can check the previous row and move left from there.

Or, could use a binary search and choose the image midpoint and move left from there until we find that color. If we find it, we're on the axis with that color. Then binary search downwards from there to find the bottom right "corner" where the pixel above it is the same color, below it is not, to the left is the same color, and to the right is not.

Eh, that gets hairy in RTL because you have to reverse where you search from.

Let's stick with this for now. I can see it potentially raising problems if we introduce tests with the background color we're checking for but that's a big "if".

This is pretty dope as-is.

patrickrodee: Another option, though, is to provide a crop-width argument in the screenshot API. Since the fixture.scss sets the width of each column to 171px with 10px margins, you know the exact width of the content. It's a possibility to consider.

test/screenshot/lib/controller.js
acdvorak: Yes, exactly! Should I add a comment explaining that?

test/screenshot/lib/controller.js
acdvorak: I originally planned to use a border, but background-color sounded easier to program 😛 

Is it OK if I create a separate follow-up PR to crop based on border instead of background-color?

patrickrodee: 100%!! This works as-is which is great. I was mostly brainstorming.

test/screenshot/lib/controller.js
patrickrodee: Yeah, a one-liner is fine.

PR #2729, opened 2018-05-10
fix(infrastructure): Ensure grid pattern renders correctly in IE

acdvorak:
### How to test

```bash
export CBT_USERNAME='...'
export CBT_AUTHKEY='...'

npm run screenshot:build
npm run screenshot:test

echo '.mdc-button:disabled{margin-left:10px}' >> packages/mdc-button/mdc-button.scss
npm run screenshot:build
npm run screenshot:test -- --mdc-include-url=button/classes/ --mdc-include-browser=ie@11
```

### What it does

- Uses PNG `background-image` instead of `linear-gradient`
- Adds Webpack loader `base64-inline-loader` to base64-encode `data:` URIs from background images in Sass

IE occasionally renders the `linear-gradient` version incorrectly, shifting it 1px to the left, which generates noisy, flaky diffs.


test/screenshot/fixture.scss
```diff
@@ -52,13 +51,8 @@ $test-button-columns: 2;
...
+  padding: $test-button-cell-padding;
 
   // Ruler grid pattern
-  // https://stackoverflow.com/a/32861765/467582
-  background-image:
-    linear-gradient(to right, #{$test-ruler-color} 1px, transparent 1.01px),  // fraction for IE 11
-    linear-gradient(to bottom, #{$test-ruler-color} 1px, transparent 1.01px); // fraction for IE 11
-
-  background-size: #{$test-ruler-granularity} #{$test-ruler-granularity};
+  background-image: url("images/ruler-grid.png");
```
patrickrodee: Inline the base64 image so you can drop the b64 webpack dependency

acdvorak: Done - good call 😄

PR #2740, opened 2018-05-14
fix(text-field): Add missing import to textfield _mixins.scss

### LINKED ISSUE #2739: textfield mixins file uses imports included in the file

> HEAD

> N/A

### What OS are you using?

> Linux

> Import mdc-textfiled/mixins without importing any other files from that directory.
> Use mdc-text-field-outlined-dense_ or any of the other dense mixins

> Scss file will compile without build error

> Error not finding $mdc-text-field-box-dense-label-position-y which is in the variables import.

TedSander:
Closes #2739


TedSander:
@aprigogin would you mind reviewing this?

packages/mdc-textfield/_mixins.scss
```diff
@@ -21,6 +21,7 @@
 @import "helper-text/mixins";
 @import "icon/mixins";
 @import "icon/variables";
+@import "variables";
```
aprigogin: should this be @import "./variables";?

TedSander: Done

PR #2751, opened 2018-05-17
fix(top-app-bar): Fix JS error when navigation icon is not present.

### LINKED ISSUE #2721: Top App Bar doesn't initialize without a navigation icon
If you don't have a navigation icon, the top app bar fails to instantiate. The permanent drawer example is a good use case for having a top app bar without a menu/navigation icon.

abhiomkar:
Fixes Top App Bar's initialize method to add ripple to nav icon only when it is present.

Fixes #2721

### DIFF
--- a/test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
+++ b/test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
@@ -66,6 +66,11 @@ function getFixture(removeIcon) {
   return html;
 }
 
+function getIconsCount(root) {
+  const selector = strings.ACTION_ITEM_SELECTOR + ',' + strings.NAVIGATION_ICON_SELECTOR;
+  return root.querySelectorAll(selector).length;
+}
+
 class FakeRipple {
   constructor(root) {
     this.root = root;
@@ -74,11 +79,11 @@ class FakeRipple {
   }
 }
 
-function setupTest(removeIcon = false) {
+function setupTest(removeIcon = false, rippleFactory = (el) => new FakeRipple(el)) {
   const fixture = getFixture(removeIcon);
   const root = fixture.querySelector(strings.ROOT_SELECTOR);
   const icon = root.querySelector(strings.NAVIGATION_ICON_SELECTOR);
-  const component = new MDCTopAppBar(root, undefined, (el) => new FakeRipple(el));
+  const component = new MDCTopAppBar(root, undefined, rippleFactory);
 
   return {root, component, icon};
 }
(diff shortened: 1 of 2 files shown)


abhiomkar:
Yes, the use of FakeRipple in unit test suppresses this bug.

I've added a new test case to cover this case also updated other test case to make use of stubs instead of asserting on private properties. PTAL.

test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
```diff
@@ -89,12 +94,22 @@ test('attachTo initializes and returns an MDCTopAppBar instance', () => {
...
+test('constructor instantiates icon ripples for all icons', () => {
+  const rippleFactory = td.function();
+  td.when(rippleFactory(td.matchers.anything())).thenReturn((el) => new FakeRipple(el));
+  const {root} = setupTest(/** removeIcon */ false, rippleFactory);
+
+  const totalIcons = getIconsCount(root);
+  td.verify(rippleFactory(td.matchers.anything()), {times: totalIcons});
+});
+
+test('constructor instantiates icon ripples except for nav icon', () => {
```
kfranqueiro: Nit: Rephrase to `'constructor does not instantiate ripple for nav icon when not present'`

abhiomkar: Done. Thanks!

test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
```diff
@@ -89,12 +94,22 @@ test('attachTo initializes and returns an MDCTopAppBar instance', () => {
...
-  const {root, component} = setupTest();
-  const selector = strings.ACTION_ITEM_SELECTOR + ',' + strings.NAVIGATION_ICON_SELECTOR;
-  const totalIcons = root.querySelectorAll(selector).length;
+test('constructor instantiates icon ripples for all icons', () => {
+  const rippleFactory = td.function();
+  td.when(rippleFactory(td.matchers.anything())).thenReturn((el) => new FakeRipple(el));
+  const {root} = setupTest(/** removeIcon */ false, rippleFactory);
+
+  const totalIcons = getIconsCount(root);
+  td.verify(rippleFactory(td.matchers.anything()), {times: totalIcons});
```
williamernest: This test and the test are causing warnings in the test runner and CI. Line 99 is mocking this object, and then it's testing that the mocked object got called. 
```
WARN: 'Warning: testdouble.js - td.verify - test double was both stubbed and verified with arguments (anything()), which is redundant and probably unnecessary. (see: https://github.com/testdouble/testdouble.js/blob/master/docs/B-frequently-asked-questions.md#why-shouldnt-i-call-both-tdwhen-and-tdverify-for-a-single-interaction-with-a-test-double )'
```

kfranqueiro: Whoops, good catch. I suppose since we're stubbing the function, we can increment a variable that's local to the test function, and just assert that it has the value we expect afterwards?

PR #2771, opened 2018-05-18
feat(text-field): Add methods to set text field icon aria-label and content

rlfriedman:
Add methods to set text field icon aria-label and content.

BREAKING CHANGE: Adds setContent adapter API to text field icon


kfranqueiro:
What use case does programmatically updating only the aria-label fulfill?

rlfriedman:
As an example, if there was something like a password visibility icon the label would need to change after press depending on the state of the text field ('Hide Password' -> 'Show Password').

kfranqueiro:
Would the icon itself not also need to change, then? Do we need to have a `setIconContent` API or something (at least under the assumption of using Material icons, where the icon is determined purely by text content)?

PR #2773, opened 2018-05-19
chore(infrastructure): Always build JS/CSS before running `screenshot:test`

### LINKED ISSUE #2753: screenshot:test should run screenshot:build first
Currently, the screenshot test script assumes `screenshot:build` was just run (or that `npm start` is running). This can cause two very confusing situations:

* If you don't have an `out` folder at all due to never running the build, `screenshot:test` will upload no JS/CSS, resulting in a ton of noisy screen diffs between goldens with styles and snapshots without
* If you have an `out` folder from a previous build on another branch and forget to rebuild, `screenshot:test` will upload the last JS/CSS you built, resulting in diffing screenshots that totally aren't from the current branch you're working on, and probably confusing the living daylights out of you (or giving you a false sense of security if there were no diffs but there should've been)

I just burned 15 minutes this morning scratching my head due to the first bullet point. I imagine the second would be far more confusing.

We shouldn't optimize for the case of the developer already having run something else; moreover, if we integrate this into CI, that will need to run the build first anyway. Building CSS/JS takes far less time than uploading and running CBT does; I'd rather take a few extra seconds to run a build than take several minutes to re-run CBT after realizing I forgot to build manually.

acdvorak:
Fixes #2753.

### How to test

```bash
export CBT_USERNAME='...'
export CBT_AUTHKEY='...'
```

#### Compile JS/CSS by default

```bash
npm run screenshot:test
```

#### Manually disable compilation

```bash
npm run screenshot:test -- --mdc-skip-build
```

#### Automatically disable compilation if local dev server is already running

```bash
# Terminal 1:
npm run screenshot:watch

# Terminal 2:
npm run screenshot:test
```

### What it does

* Builds JS/CSS files by default when you run `npm run screenshot:test`
* Automatically disables compilation if `npm run screenshot:watch` or `npm run screenshot:build` are already running in another terminal
* Allows the user to manually disable compilation with a `--mdc-skip-build` CLI flag
* Adds a new `npm run screenshot:clean` command
* Removes `test/screenshot/out/` from `npm run clean`


package.json
```diff
@@ -23,12 +23,13 @@
     "pretest": "npm run lint",
     "test": "npm run test:unit && npm run test:closure && npm run test:dependency && npm run build && npm run clean",
     "posttest": "istanbul report --root coverage text-summary && istanbul check-coverage --lines 95 --statements 95 --branches 95 --functions 95",
-    "screenshot:build": "npm run clean && mkdirp test/screenshot/out/ && webpack --config=test/screenshot/webpack.config.js --progress",
+    "screenshot:build": "npm run screenshot:clean && mkdirp test/screenshot/out/ && webpack --config=test/screenshot/webpack.config.js --progress",
+    "screenshot:clean": "del-cli test/screenshot/out/",
```
kfranqueiro: I suspect `start` should now run `screenshot:clean` instead of `clean`?

acdvorak: Done! Good catch, Ken 😁

test/screenshot/lib/controller.js
```diff
@@ -16,6 +16,7 @@
 
 'use strict';
 
+const child_process = require('child_process'); // eslint-disable-line
```
patrickrodee: Why do you need `// eslint-disable-line` here?

test/screenshot/lib/cli-arg-parser.js
```diff
@@ -162,6 +175,11 @@ E.g., 'origin/master' (default), 'HEAD', 'feat/foo/bar', 'fad7ed3:path/to/golden
     return this.args_['mdc_diff_base'];
   }
 
+  /** @return {boolean} */
+  get skipBuild() {
+    return this.args_['mdc_skip_build'];
```
patrickrodee: Silly question but how/where is `mdc_skip_build` defined as a boolean?

test/screenshot/lib/controller.js
acdvorak: Oops! Forgot to include the specific rule I'm disabling: `camelcase`. Added!

I suppose I could rename the variable to `childProcess`, but that felt kinda weird. Every example I've seen calls it `child_process`. 🤷‍♂️ 

WDYT?

test/screenshot/lib/cli-arg-parser.js
acdvorak: Not a silly question at all!

The `argparse` API doesn't make it very clear, but the param is defined as a boolean here:

https://github.com/material-components/material-components-web/pull/2773/files/c2f2b6b5a6871b5d5c3ca84eec51a493c9a3b220#diff-d2bba4088f19b97ee4fe618d0353157aR131

The `action: 'storeTrue'` line the argument config sets the value to `true` if the argument is present, or `false` if it's omitted.

Is there a way I could make this clearer?

test/screenshot/lib/controller.js
patrickrodee: Unless you need the whole `child_process` module, you could could destructure and just load `spawnSync`.

That being said, I'd just camel case it.

test/screenshot/lib/cli-arg-parser.js
patrickrodee: Probably, but that's mostly on `argparse` for having an unclear API. I'd add a comment explaining that it casts as a boolean.

acdvorak: Done!

PR #2776, opened 2018-05-20
fix(dialog): higher '.mdc-dialog__action' specificity

### LINKED ISSUE #2243: Dialog: '.mdc-dialog__action' not highest specificity and being overwritten

0.30.0

This should be happening in all browsers

### What OS are you using?

Ubuntu 16.04 but I don't

Add an `mdc-dialog__action` class to one of the footer action `<button>`s.
You can view a simple codepen here: https://codepen.io/hvolschenk/pen/ddzRpL
This is also an issue on the demo site: https://material-components-web.appspot.com/dialog.html

The `<button>` should have a `text-color` that is the secondary color

The correct style is being applied, although it is being overwritten by the `.mdc-button:not(:disabled)` style (Which is a part of the **mdc-button** component).

That should be it.

hvolschenk:
Closes #2243


hvolschenk:
I don't quite get why this test isn't running: https://travis-ci.org/material-components/material-components-web/jobs/381379433

It doesn't have anything to do with the changes I made here. I also don't get a _retry_ option on the build, don't really want to push up something else to retry it. :confused:

packages/mdc-dialog/mdc-dialog.scss
```diff
@@ -122,7 +122,8 @@
   }
 }
 
-.mdc-dialog__action {
+.mdc-dialog__action,
```
patrickrodee: You can safely remove this line and the change will work as intended.

hvolschenk:
Man, again: https://travis-ci.org/material-components/material-components-web/jobs/383029803

```
ERROR [launcher]: Cannot start Chrome
```

Anything I can do from my side to spur 'ol Travis along?

kfranqueiro:
Don't worry about the Travis failure; it's presently unavoidable because the unit tests run through Sauce Labs and use keys configured within Travis, which aren't exposed to forks.

PR #2777, opened 2018-05-21
chore(infrastructure): Only update changed screenshots in `golden.json`

### LINKED ISSUE #2754: screenshot:update-goldens nukes excluded URLs
The update-goldens script understands the `--mdc-include-url`/etc. filter arguments, but when it generates the new `golden.json`, it ends up removing any URLs that were filtered out and only contains the URLs that were included.

Should we restrict those options to only the `screenshot:test` script?

It also occurs to me that if you were to run `screenshot:test` on a subset of URLs, then run `screenshot:update-goldens` on the full suite, you may end up unexpectedly changing more goldens than you originally tested. Should we enforce that at least one full unfiltered `screenshot:test` run happens for every PR that updates goldens?

acdvorak:
Fixes #2754.

### How to test

```bash
export CBT_USERNAME='...'
export CBT_AUTHKEY='...'

# No diffs
npm run screenshot:update-goldens

# Lots of diffs (without CLI filters)
npm run screenshot:update-goldens

# Lots of diffs (with CLI filters)
echo '.mdc-fab:not(:disabled){color:red}' >> packages/mdc-fab/mdc-fab.scss
npm run screenshot:update-goldens -- --mdc-include-url=mdc-fab/classes/baseline --mdc-include-browser=ie@11
```

#### Old behavior

Previously, running `npm run screenshot:update-goldens` would blindly update all screenshot URLs in `golden.json` - even if a given screenshot did not have any diffs.

This makes PRs harder to review because the git diffs for `golden.json` are really noisy, making it look like every single screenshot changed.

#### New behavior (without CLI filters)

Now, when you run `update-goldens` without any CLI filters (e.g., `--mdc-include-url=mdc-fab`), it **generates a brand new `golden.json`** file but **keeps all diff-less screenshot URLs** from the old `golden.json`.

This ensures that git diffs for `golden.json` only contain screenshots that actually changed, while still allowing deleted HTML files to be purged from the newly-generated JSON.

#### New behavior (with CLI filters)

If you pass one or more CLI filters to `update-goldens` (e.g., `--mdc-include-browser=ie@11`), it **keeps the old `golden.json`** file and **only updates screenshot URLs that have diffs**.

This ensures that we don't accidentally delete screenshots for pages/browsers that were excluded by the filters.

### Example output

[Diff Report](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/21/16_57_34_771/49282355a/report.html)


test/screenshot/lib/controller.js
```diff
@@ -299,7 +307,7 @@ class Controller {
           console.log('\n\nDONE diffing screenshot images!\n\n');
           console.log(diffs);
           console.log(`\n\nFound ${diffs.length} screenshot diffs!\n\n`);
-          return {diffs, testCases};
+          return {testCases, diffs};
```
kfranqueiro: Nit: not sure why this line changed?

test/screenshot/lib/controller.js
```diff
@@ -84,6 +91,7 @@ class Controller {
 
   async initialize() {
     this.baseUploadDir_ = await this.storage_.generateUniqueUploadDir();
+    await this.gitRepo_.fetch();
```
kfranqueiro: Are there any implications of this being run here rather than `parseDiffBase` in `cli-arg-parser` now?

kfranqueiro: Also FYI this is going to conflict with the `shouldBuild` code that was merged.

test/screenshot/lib/controller.js
acdvorak: Just for consistency with the other `return` statements, to satisfy my OCD 😛

test/screenshot/lib/controller.js
acdvorak: AFAIK the only differences are:

1. It only runs once, instead of every time `parseDiffBase()` is called
    (which is why I moved it here)
2. If fetching fails for some reason, it'll happen immediately instead of after a minute of uploading files to GCS (which is probably better devex)

acdvorak: Fixed merge conflict. Thanks for the heads up!

test/screenshot/lib/snapshot-store.js
```diff
@@ -117,6 +131,97 @@ class SnapshotStore {
...
+
+    /** @type {!Array<[string, !SnapshotPageJson]>} */
+    const newMatchingPageEntries = Object.entries(newJsonData).filter(existsInOldJsonData);
+
+    for (const [htmlFilePath, newPage] of newMatchingPageEntries) {
+      let pageHasChanges = false;
+
+      for (const browserKey of Object.keys(newPage.screenshots)) {
+        const oldUrl = oldJsonData[htmlFilePath].screenshots[browserKey];
+        const matchingDiff = diffs.find((diff) => diff.htmlFilePath === htmlFilePath && diff.browserKey === browserKey);
```
patrickrodee: Nit: rename to `hasMatchingDiff`

test/screenshot/lib/snapshot-store.js
```diff
@@ -117,6 +131,97 @@ class SnapshotStore {
...
+          pageHasChanges = true;
+        }
+      }
+
+      if (!pageHasChanges) {
+        newPage.publicUrl = oldJsonData[htmlFilePath].publicUrl;
+      }
+    }
+
+    return newJsonData;
```
patrickrodee: Maybe I'm missing something but where does `newJsonData` get modified in the above loops?

test/screenshot/lib/snapshot-store.js
acdvorak: Done

test/screenshot/lib/snapshot-store.js
acdvorak: It might not be 100% obvious, but the `for` loops iterate over `newJsonData` and change some of its properties:

```js
newPage.screenshots[browserKey] = oldUrl;
...
newPage.publicUrl = oldJsonData[htmlFilePath].publicUrl;
```

I tried to name all the variables `newPage` and `newMatchingPageEntries` to hint that they're from `newJsonData`, but I maybe it's still not totally clear 😕

Any advice?

patrickrodee: How is it changing the properties? The variable you're updating is `newPage`. Is `newPage` a copy of `newJsonData`?

acdvorak: `newPage` is a pointer to a sub-object within `newJsonData`.

I basically loop over the pages and screenshots in `newJsonData`, and if I find a screenshot that hasn't changed, I copy the URL from `oldJsonData` and set it on `newPage` (which is just a reference to an object in `newJsonData`).

E.g.:

Here's `newJsonData`:
```json
{
  "page.html": { // <-- `newPage`
    "publicUrl": "NEW/page.html",
    "screenshots": {
      "desktop_windows_ie@11": "NEW/page.html.ie11.png"
    }
  }
}
```

Here's `oldJsonData`:
```json
{
  "page.html": {
    "publicUrl": "OLD/page.html",
    "screenshots": {
      "desktop_windows_ie@11": "OLD/page.html.ie11.png"
    }
  }
}
```

Does that make sense at all?

(Sorry for the confusion... I knew this method might be kinda hard to understand, but I wanted to get something working ASAP.)

patrickrodee: Ah, I see. Yeah, that's not super clear. Since you've explained it, I get it now, so I'll stamp. That being said, we should make an issue to clear this up soon. This is one of the more important parts of screenshot testing and should be pretty clear.

acdvorak: K, I added a TODO to refactor this method. Thanks!

PR #2781, opened 2018-05-21
fix(text-field): Update floating-label to work properly for number fields

### LINKED ISSUE #2744: textfield label doesn't float while invalid text is present (e.g. in number fields)
### Conditions for the bug
- The `type` attribute of the `input` element should be set to `number`
- The text entered should of the form number-number. Example: 12-345

### Steps to reproduce the bug
1. Open https://codepen.io/namannehra/pen/aGaNZv
2. Enter "1-1" in the textfield
3. Unfocus the textfield
4. Look at the label

### Expected behavior
The label should be floating after losing focus

### Actual behavior?
The label is **not** floating after losing focus

### Browsers affected
Chrome/66.0.3359.139
Edge/17.17134
Firefox/59.0

namannehra:
Invalid input in number fields reports its value as an empty string. This value is used to check if label should float. This results in label not floating when invalid input is present. This commit changes shouldFloat to check for badinput too. badinput is true if invalid input present so the label floats correctly.


namannehra:
I signed it!

merlinnot:
Tests? :)

namannehra:
@merlinnot Are the scripts to build and run individual tests documented somewhere? I have found a few more bugs that I should like to fix but dont know how to run the test.

merlinnot:
I don't really know, I'm just a regular guy, not a team member. But looking at a package.json I see `test:unit` and `screenshot:test` scripts, the former being Karma, the latter some custom script.

namannehra:
But how do test only a single package (textfield in this case)?

merlinnot:
They are using [Lerna](https://github.com/lerna/lerna), so the command would be sth like `lerna run --scope mdc-textfield test:unit`

kfranqueiro:
The `test/unit` folder has subfolders for each package. There is always a `mdc-foo.test.js` for the component logic, and a `foundation.test.js` for the foundation logic.

To filter which tests to run, you can also run `npm run test:watch`, then click the "debug" button in the browser that opens, and add e.g. `?grep=MDCTextField` to the debug page's URL. (You can then also paste that URL into other browsers.)

namannehra:
I wrote a new test which fails before this fix and is now passing. All other tests are passing. Is this ok to merge now?

namannehra:
@williamernest
This now works not just for `text` and `number` but for the following: `email`, `number`, `password`, `search`, `tel`, `text`, `url`

It doesnt support `color`, `date`, `datetime-local`, `file`, `month`, `time`, `week`

Support for these will require bigger changes. I'll try to do these once this one is approved merged.

PR #2785, opened 2018-05-22
chore(infrastructure): Improve screenshot auto-cropping

acdvorak:
### How to test

```bash
export CBT_USERNAME='...'
export CBT_AUTHKEY='...'

# No diffs
npm run screenshot:build
npm run screenshot:update-goldens

# Lots of diffs (without CLI filters)
npm run screenshot:build
npm run screenshot:update-goldens

# Lots of diffs (with CLI filters)
echo '.mdc-fab:not(:disabled){color:red}' >> packages/mdc-fab/mdc-fab.scss
npm run screenshot:build
npm run screenshot:update-goldens -- --mdc-include-url=mdc-fab/classes/baseline --mdc-include-browser=ie@11
```

### What it does

* **Fixes cropping in IE 11 💩**:
    - Adds 1px of padding to the `body` element
    - Adds trim borders (`#abc123`) on the top and left sides of the `main` element
    - IE or CBT (not sure which) occasionally shifts the entire page 1px to the left. Adding 1px of padding to the `body` should (hopefully) prevent the left side of the screenshot from getting cut off, and thus avoid a lot of noisy diffs.
* **Fixes cropping in Android Chrome**:
    - Adds a small tolerance to trim border detection: Each channel (R, G, and B) must be within +/- 20 units (out of 255) of the corresponding channel of the trim color (`#abc123`)
    - Some Android devices render a shadow at the top of the page, which darkens the trim border
    - Some displays render slightly different colors than what's authored in the CSS. E.g., I've seen `#abc123` get rendered as `#abc033`, `#abc133`, or `#abc217`, depending on the display.
* **Fixes fuzzy ruler grid lines** on high-DPI devices (retina MacBooks, iPhones, Android Galaxy S7)
    - Reverts back to `linear-gradient` instead of a PNG image

### Example output

* [Full diff report with 2 flakes](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/22/16_23_58_618/af3d7271f/report.html)
* [Partial diff report for 2 flakes](https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/05/22/17_07_39_456/af3d7271f/report.html)


test/screenshot/fixture.scss
```diff
@@ -18,8 +18,9 @@
 @import url("https://fonts.googleapis.com/icon?family=Material+Icons");
 
 $test-trim-color: #abc123; // Value must match `TRIM_COLOR_CSS_VALUE` in `image-cropper.js`
+$test-grid-color: #dddddd;
+$test-fixture-width: 348px;
```
patrickrodee: Where is 348px coming from?

test/screenshot/lib/image-cropper.js
```diff
@@ -94,21 +108,74 @@ class ImageCropper {
...
   }
 
   /**
+   * @param {!RGBA} pixelColorRGB
+   * @return {boolean}
+   * @private
+   */
+  isTrimColor_(pixelColorRGB) {
+    return (
+      Math.abs(pixelColorRGB.r - this.trimColorRGB_.r) < TRIM_COLOR_CHANNEL_DISTANCE &&
```
patrickrodee: If the only colors that change value across browser (based on your comment) are green and blue, do you need the red color distance check?

test/screenshot/fixture.scss
acdvorak: It's wide enough to fit 2 columns of buttons, and narrow enough to fit in a mobile Chrome viewport without having extra whitespace on the right side of the cropped image.

I just played around in the Chrome Dev Tools until I found the best value 😛 

The width of the cells is also hard-coded. They're wide enough to fit most of our buttons, narrow enough to fit in two columns in mobile Chrome, and they show the right-most and bottom-most ruler line instead of cutting off.

I originally calculated the widths, but it was such a pain in the ass to update the formulas every time I changed the borders/margins of the grid cells, so I figured it would be simpler to just hard-code the value for now.

test/screenshot/lib/image-cropper.js
acdvorak: I added it mainly out of paranoia - I don't trust the `red` color value to be the same on every device 😛

PR #2789, opened 2018-05-22
Fix broken menu/constants.js link

iansan5653:


kfranqueiro:
Thanks for catching this. Can we simply peg this link at a specific tag instead of a commit to make the link more meaningful as well? e.g. replace the commit hash with `v0.35.2`.

iansan5653:
@kfranqueiro That makes a lot more sense, actually. Change has been made.

kfranqueiro:
Thanks!

PR #2791, opened 2018-05-22
refactor(chips): Stop handling DOM manipulation in input chips

### LINKED ISSUE #2761: Investigate how input chips foundation will work with wrapping libraries
We had logic in our `MDCChipSetFoundation` to add and remove HTML from the DOM. This is difficult for MDC Web to maintain. So we need to remove that logic...but we need our wrapping libraries (like React) to handle adding/removing HTML from the DOM. So our Foundation needs to be workable with wrapping libraries

bonniezhou:
The main functionality behind input chips is the ability to [convert text into a chip](https://material.io/design/components/chips.html#input-chips). The MDC Chips component currently handles the adding/removing of a chip element to/from the DOM. Wrapping this in other web frameworks, such as React, becomes much more complicated if it handles DOM manipulation differently.

For this reason, part of MDC Web’s philosophy is that we should not manage the adding/removing of elements to the DOM at all. For input chips, we should let the user handle adding/removing chip DOM elements.

Fixes #2761


test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -126,20 +131,10 @@ test('#adapter.deregisterInteractionHandler removes a handler from the root elem
...
-  assert.equal(root.childNodes[3], chipEl);
-});
-
 test('#adapter.removeChip removes the chip object from the chip set', () => {
   const root = getFixture();
   const component = new MDCChipSet(root, undefined, (el) => new FakeChip(el));
   const chip = component.chips[0];
   component.getDefaultFoundation().adapter_.removeChip(chip);
   assert.equal(component.chips.length, 2);
-  td.verify(chip.remove());
```
kfranqueiro: Replace this with verifying destroy?

demos/chips.html
```diff
@@ -235,27 +235,55 @@ <h2>Custom theme</h2>
...
+            inputChipSetComponent.addChip(newChipEl);
+
             input.value = '';
           }
         };
-        inputButton.addEventListener('click', addChip);
-        input.addEventListener('keydown', addChip);
+
+        function removeChip(evt) {
+          const {root} = evt.detail;
```
kfranqueiro: Change to `const root = evt.detail.root` so the demo page works in IE.

PR #2792, opened 2018-05-22
chore(infrastructure): Add quick start and usage info to screenshot test README

acdvorak:
Adds the following sections to the README:

- Quick start
    * API credentials from CBT and GCP
    * Testing changes with `npm run screenshot:test`
- Basic usage
    * Updating `golden.json` with `npm run screenshot:update-goldens`
    * Filtering tests by browser/URL with `--mdc-include-url`, `--mdc-exclude-url`, `--mdc-include-browser`, `--mdc-exclude-browser` CLI flags
- Advanced usage
    * `--help` CLI flag
    * Creating public demos with `npm run screenshot:upload-assets`
    * Diffing against other branches/files with `--mdc-diff-base` CLI flag


test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
...
+
+If a single screenshot fails to render, you can rerun it without running the entire suite:
+
+```bash
+npm run screenshot:update-goldens -- \
+  --mdc-include-browser=ie@11 \
+  --mdc-include-url=mdc-button/classes/dense
+```
+
+**IMPORTANT:** Note the `--` between `screenshot:test` and `--help`. This is required by `npm`.
```
kfranqueiro: The example above uses `update-goldens` but the strings here assume `screenshot:test -- --help`. These should be consistent.

test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
...
+### Updating "golden" screenshots
+
+```bash
+npm run screenshot:update-goldens
+```
+
+In addition to running the tests, this command will also update your local `test/screenshot/golden.json` file with the 
+newly captured screenshots.
+
+### Rerunning flaky tests
```
kfranqueiro: Given we're now auto-retrying, I'd re-word this to "Rerunning a subset of tests" and re-word below to "You can rerun a subset of tests without running the entire suite, filtering by browser and/or URL:"

I'm also wondering if it's better to show this example with `screenshot:test`, though it does work with both... ostensibly we should be running the entire suite on a branch at least once though, right?

test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
...
+**IMPORTANT:** Note the `--` between `screenshot:test` and `--help`. This is required by `npm`.
+
+### Diffing against a local `golden.json` file
+
+By default, screenshots are diffed against `origin/master:test/screenshot/golden.json`.
+
+To diff against a local `golden.json` file, run:
+
+```bash
+npm run screenshot:update-goldens -- --mdc-diff-base=test/screenshot/golden.json
```
kfranqueiro: I'm slightly confused by this usage, given that diff-base can also specify a URL to a golden.json, or an entire branch name. It feels like this argument is potentially overloaded. Is it actually looking at the golden.json within a branch when you specify just a branch name anyway?

test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
...
+### Diffing against another branch
+
+By default, screenshots are diffed against `origin/master:test/screenshot/golden.json`.
+
+To diff against a different branch, run:
+
+```bash
+npm run screenshot:update-goldens -- --mdc-diff-base=fix/fab/icon-alignment-ie11
+```
+### Public demos
```
kfranqueiro: This title seems misleading since we're still talking about the screenshot test pages. I assume `upload-assets` is just the thing that `test` and `update-goldens` runs after `build`?

test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
...
+
+```bash
+npm run screenshot:upload-assets
+```
+
+This will upload all test assets (HTML/CSS/JS files) to a public URL and print the URL in the terminal.
```
kfranqueiro: all **currently-built** test assets, I assume... (i.e. you need to run `screenshot:build` first)

test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
...
+  --mdc-include-browser=ie@11 \
+  --mdc-include-browser=android_chrome \
+  --mdc-include-url=mdc-button/classes/dense \
+  --mdc-include-url=mdc-fab/classes/mini
+```
+
+Partial matches are allowed, so `ie@11` is equivalent to `desktop_windows_ie@11`, and `chrome` is equivalent to 
+`desktop_windows_chrome@latest` and `mobile_android_chrome@latest`.
+
+See `test/screenshot/browser.json` for the full list of supported browsers.
```
kfranqueiro: Add note about `exclude` counterparts?

test/screenshot/README.md
```diff
@@ -1,22 +1,144 @@
 # Screenshot Tests
 
-## Local dev server
+Prevent visual regressions by running screenshot tests on every PR.
 
-Run:
+## Quick start
+
+Add the following to your `~/.bash_profile` (Mac) or `~/.bashrc` (Linux):
```
kfranqueiro: Both of these files work on both OSes, so I'm not sure these parentheticals are warranted... bash_profile vs. bashrc is a login-shell vs. interactive-non-login-shell thing. I think most terminal apps tend to execute login shells?

test/screenshot/README.md
kfranqueiro: Tangentially, I've been wondering if we have suggestions on recommended workflow that can be documented. Like, screenshot:test and screenshot:update-goldens are both going to run the tests... and presumably we should run the tests once to confirm the results before updating goldens... but then we have to wait for all of the tests to run AGAIN when we run update-goldens. I guess if you haven't changed anything you can then filter update-goldens based on what was actually found in test, but that's subject to user error.

As much as I don't want to add more work, it might be nice to eventually repurpose update-goldens to hand it an existing report and update the JSON based on that rather than needing to run all of the tests all over again?

test/screenshot/README.md
acdvorak: All of the `npm run screenshot:*` commands accept the same CLI flags.

How can I make that clearer?

test/screenshot/README.md
acdvorak: Yeah, all CLI flags work with both `screenshot:test` and `screenshot:update-goldens`.

As far as workflow goes, I was planning to eventually add an "Approve" button on the diff report page. It would copy a CLI command to the clipboard that updates `golden.json` without rerunning the tests.

To do that, I would change `update-goldens` so that it doesn't run any tests, but instead accepts a URL or file path of a `snapshot.json` file, which would be automatically generated by `screenshot:test` and uploaded to GCS.

So the workflow would go like this:

```bash
# ... hack hack hack ...

npm run screenshot:test

# ... user views the report and clicks the "Approve" button ...

npm run screenshot:update -- https://storage.../snapshot.json

git add test/screenshot/golden.json
git commit -m 'WIP: Update goldens'
```

This should be pretty quick and easy to do.

test/screenshot/README.md
kfranqueiro: Add another note under advanced CLI usage to say the flags work for both test and update-goldens.

(I'm assuming most of the options aren't _relevant_ for the other commands, like build and upload...)

test/screenshot/README.md
acdvorak: Yeah, it's a bit overloaded, and it's not entirely obvious how it works. I definitely get that.

I had a feeling that would be the case, but it seemed better than the alternative of having a bunch of different CLI flags...
E.g.: `--mdc-diff-base-url`, `--mdc-diff-base-file`, `--mdc-diff-base-commit`, `--mdc-diff-base-branch`, `--mdc-diff-base-tag`.

If you only specify a branch name, the `golden.json` file path is assumed to be
`--mdc-golden-path`.

However, the primary purpose of `--mdc-golden-path` is to determine where to write the updated `golden.json` file when the user runs `screenshot:update-goldens`.

If we ever decide to move or rename `golden.json` in our repo, you can tell it to use the old repo path like this:

```
--mdc-diff-base=some/OLD/branch:OLD/path/to/golden.json
--mdc-golden-path=NEW/path/to/golden.json
```

I figure most people won't need this feature right away, but it's a really useful tool to have in your toolbox when you're investigating a regression that was accidentally merged into `master` some time ago.

test/screenshot/README.md
acdvorak: Yeah exactly: `test` and `update-goldens` run `upload-assets` after building.

I was trying to give it a title that members of the team would understand immediately.

Any suggestions for a better title?

test/screenshot/README.md
acdvorak: Now that #2773 is merged, `test` and `update-goldens` automatically run `build` first, so "currently-built" seems a bit redundant; the assets are always built.

test/screenshot/README.md
acdvorak: Done

test/screenshot/README.md
acdvorak: Done

test/screenshot/README.md
acdvorak: Ahh, I see. It exists because we've all tried to share a demo with a designer at some point and discovered that it's surprisingly difficult due to network security policies.

What's more, demos running on `localhost:8080` are hard to access on non-corp devices and non-corp networks.

So I'm aiming for a **non-screenshot** use case where you just want to show somebody your local demo.

Does that make sense? How can I make that clearer?

test/screenshot/README.md
acdvorak: Yes, `upload-assets` also goes through the `shouldBuild` flow 😀 

All of the `run-*.js` files call `Controller.initialize()`, which automatically builds the assets (unless you pass `--mdc-skip-build` or have `npm start` running in another terminal).

test/screenshot/README.md
acdvorak: Done

test/screenshot/README.md
kfranqueiro: That's what I suspected the use case was, OK. So then, back to my original point, rephrasing my first question:

If you've never run screenshot tests before (or haven't run them yet against your current branch), and immediately run _just_ the upload-assets script, are you going to have A Bad Time? Or is it going to build first?

test/screenshot/README.md
```diff
@@ -1,22 +1,160 @@
...
+
+You can rerun a subset of the tests without running the entire suite, filtering by browser and/or URL:
+
+```bash
+npm run screenshot:update-goldens -- \
+  --mdc-include-url=mdc-button/classes/dense \
+  --mdc-include-browser=ie@11
+```
+
+**IMPORTANT:** Note the `--` between `screenshot:test` and `--help`. This is required by `npm`.
```
kfranqueiro: This doesn't match the invocation immediately above. Maybe genericize this remark to "between the script name and its arguments"

acdvorak: Done

PR #2794, opened 2018-05-23
fix(text-field): Made handleValidationAttributeMutation to accept attribute list

### LINKED ISSUE #2717: Textfield.foundation.handleValidationAttributeMutation_ should accept an array of attributes

> 0.35.1

> all browsers

### What OS are you using?

> Mac OSx 10.13.3

Textfield.foundation.handleValidationAttributeMutation_ is expecting a MutationObject from a MutationObserver. This code should not be in the foundation, since React and other frameworks may not need MutationObservers.

The method should instead accept an array of attributes.

abhiomkar:
Fixes #2717

### DIFF
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -123,7 +123,9 @@ class MDCTextFieldFoundation extends MDCFoundation {
     /** @private {function(!Event): undefined} */
     this.textFieldInteractionHandler_ = () => this.handleTextFieldInteraction();
     /** @private {function(!Array): undefined} */
-    this.validationAttributeChangeHandler_ = (mutations) => this.handleValidationAttributeMutation(mutations);
+    this.validationAttributeChangeHandler_ =
+        (mutations) => this.handleValidationAttributeMutation(this.getMutationAttributeNames_(mutations));
+
     /** @private {!MutationObserver} */
     this.validationObserver_;
   }
@@ -149,8 +151,8 @@ class MDCTextFieldFoundation extends MDCFoundation {
     ['click', 'keydown'].forEach((evtType) => {
       this.adapter_.registerTextFieldInteractionHandler(evtType, this.textFieldInteractionHandler_);
     });
-    this.validationObserver_ = this.adapter_.registerValidationAttributeChangeHandler(
-      this.validationAttributeChangeHandler_);
+    this.validationObserver_ =
+        this.adapter_.registerValidationAttributeChangeHandler(this.validationAttributeChangeHandler_);
   }
 
   destroy() {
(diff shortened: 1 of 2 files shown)


moog16:
Please fix lint issues

packages/mdc-textfield/foundation.js
```diff
@@ -420,6 +422,14 @@ class MDCTextFieldFoundation extends MDCFoundation {
       },
     });
   }
+
+  /**
+   * @param {!Array<!MutationRecord>} mutationsList 
+   * @return {!Array<string>} Returns list of attribute names.
+   */
+  getMutationAttributeNames_(mutationsList) {
```
moog16: This should be an adapter method. We should be removing the MutationRecord from the foundation, since its environment specific.

kfranqueiro:
Discussed this with Abhinay, and we realized we should be able to solve this with a 1-2 line change in the vanilla DOM adapter implementation for `registerValidationAttributeChangeHandler` and a small change to the foundation's `handleValidationAttributeMutation`.

The adapter will be changed to expect the handler to handle an array of strings rather than an array of mutation objects, so it'll require a README update and a BREAKING CHANGE note in the commit description.

The foundation method will also change to require an array of strings, which also needs a doc update but isn't a breaking change itself since we only just made the method public after the last release. (Should we change the name to `Change` instead of `Mutation`?)

This way, no additional adapter methods need to be implemented.

In MDC React, the logic should change to look at `Object.keys(nextProps)`, filter to only include properties that actually changed, then pass the resulting filtered array of attributes to `handleValidationAttributeMutation`. MDC React should not need to mirror the attribute whitelist array; the foundation is already doing that work.

packages/mdc-textfield/foundation.js
abhiomkar: This method is no longer required here since this part is moved to index.js (registerValidationAttributeChangeHandler's implementation) to separate the concerns of MutationRecord logic. PTAL.

moog16:
The function name change sounds good to me. 

w/ regards to the React, if we don't have the list of attributes to look for, we will be calling the handleValidationChange every time a prop changes.

kfranqueiro:
Hm. If we move MDC Web's whitelist into constants.js, can MDC React reference it there rather than needing to redefine it?

moog16:
Ya that's a good idea 👍

packages/mdc-textfield/adapter.js
```diff
@@ -102,7 +102,8 @@ class MDCTextFieldAdapter {
 
   /**
    * Registers a validation attribute change listener on the input element.
-   * @param {function(!Array): undefined} handler
+   * Handler accepts list of attribute changes.
```
kfranqueiro: changes -> names?

test/unit/mdc-textfield/mdc-text-field.test.js
```diff
@@ -319,7 +319,7 @@ test('#adapter.deregisterTextFieldInteractionHandler removes an event handler fo
 test('#adapter.registerValidationAttributeChangeHandler creates a working mutation observer', (done) => {
   const {root, component} = setupTest();
   const handler = td.func('ValidationAttributeChangeHandler');
-  td.when(handler(td.matchers.anything(), td.matchers.anything())).thenDo(() => {
+  td.when(handler(td.matchers.anything())).thenDo(() => {
```
kfranqueiro: `anything` -> `contains('required')`?

PR #2795, opened 2018-05-23
fix(ripple): Fix missing dependency

### LINKED ISSUE #2713: mdc-ripple missing mdc-animation dependency
## Bugs

The `mdc-ripple` package appears to be missing a dependency of `mdc-animation`.

    "@material/list": "^0.35.0"

### What OS are you using?

> macOS 10.13.4

1. `npm install @material/list`
2. `@import '@material/list/mdc-list';` in a Sass file
3. Attempt to compile Sass

Sass should compile without errors.

> Error:  File to import not found or unreadable: `@material/animation/variables`

I believe that adding `mdc-animation` to `dependencies` in `mdc-ripple`'s `package.json` should fix this issue.

abhiomkar:
Fixes #2713


packages/mdc-ripple/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@material/base": "^0.35.0",
-    "@material/theme": "^0.35.0"
+    "@material/theme": "^0.35.0",
+    "@material/animation": "^0.35.0"
```
patrickrodee: This should be `0.34.0`. We haven't released `0.35.0` of animation yet.
https://github.com/material-components/material-components-web/blob/e6b787c4dfdd0b6810534d929b3aad2011ab19d3/packages/mdc-animation/package.json#L4

patrickrodee: That's why the CI tests were failing, too.

abhiomkar: Done. Thanks!

PR #2804, opened 2018-05-24
fix(chips): Add delay to filter chip checkmark

EstebanG23:
Add delay to transition animation so checkmark does not appear behind text. Add speed up to leading icon to brige gap between icons.

### DIFF
--- a/packages/mdc-chips/chip/mdc-chip.scss
+++ b/packages/mdc-chips/chip/mdc-chip.scss
@@ -118,10 +118,12 @@
 .mdc-chip__icon--leading {
   transition: opacity $mdc-chip-opacity-animation-duration linear;
   opacity: 1;
+  transition-delay: -50ms; 
 
   + .mdc-chip__checkmark {
     transition: opacity $mdc-chip-opacity-animation-duration linear;
     opacity: 0;
+    transition-delay: 80ms;
 
     .mdc-chip__checkmark-svg {
       transition: width 0ms;


packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -118,10 +118,12 @@
 .mdc-chip__icon--leading {
   transition: opacity $mdc-chip-opacity-animation-duration linear;
   opacity: 1;
+  transition-delay: -50ms;
```
bonniezhou: Add a comment here to explain what's this line is for

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -117,10 +117,16 @@
 
 .mdc-chip__icon--leading {
   transition: opacity $mdc-chip-opacity-animation-duration linear;
+
+  // Speed up delay to bridge gap between leading icon and checkmark transition.
+  transition-delay: -50ms;
```
patrickrodee: Why is this negative?

EstebanG23: By delaying the transition of the checkmark it also delayed the appearance of the leading icon, and so I used the transition-delay feature to actually speed it up to bridge the gap from checkmark fading out and leading icon fading in. Helps the transition feel/look more natural.

packages/mdc-chips/chip/mdc-chip.scss
EstebanG23: Done!

packages/mdc-chips/chip/mdc-chip.scss
patrickrodee: Ah my bad! I thought this was the `transition-duration`. 

Since these numbers are important, I'd move them to the `_variables.scss` file.

Once that's in, I'll LGTM it.

EstebanG23: Done!

packages/mdc-chips/_variables.scss
```diff
@@ -28,6 +28,8 @@ $mdc-chip-trailing-icon-hover-opacity: .62;
 $mdc-chip-trailing-icon-focus-opacity: .87;
 $mdc-chip-leading-icon-size: 20px;
 $mdc-chip-trailing-icon-size: 18px;
+$mdc-chip-leading-icon-speedup: -50ms;
```
patrickrodee: Change this name to `$mdc-chip-leading-icon-delay` and move the comment about why it's negative to this file.

EstebanG23: Done!

PR #2808, opened 2018-05-24
fix(text-field): Moved VALIDATION_ATTR_WHITELIST to constants.

abhiomkar:


abhiomkar:
Yes, I'm doing React fix followed by this change. We don't need to use this constant in React anymore because the new `handleAttributeChange` change will do that for us. :)

Issue #2820, opened 2018-05-25
[Dialog] uses legecy Typography


yezhi780625:
** PLEASE READ THIS BEFORE FILING AN ISSUE **

## Bugs

Dialog uses legacy ` @material/typography ` which is incompatible with other material components web.

### What MDC Web Version are you using?

0.35.2

david-j-davis:
I'm also getting a similar error when building the dialog:

Error: Invalid style specified! Choose one of display4, display3, display2, display1, headline, title, subheading2, subheading1, body2, body1, caption, button
        on line 29 of node_modules/@material/typography/_mixins.scss, in mixin `mdc-typography`
        from line 86 of node_modules/@material/dialog/mdc-dialog.scss

kfranqueiro:
Yeah, that's the error that occurs as a result of this. Manually also installing `@material/typography@0.35` should resolve the error.

We've merged the PR to fix the dependency, so this error won't occur in our next release.

PR #2821, opened 2018-05-25
fix(dialog): Fix Typography version

yezhi780625:
It uses legacy @material/typography which is incompatible with
other material components.

#2820


yezhi780625:
I signed it!

yezhi780625:
I signed it!

kfranqueiro:
Yikes. Thanks for catching this! I was indeed able to repro failure compiling Sass when installing `@material/dialog` and its listed dependencies alone, and it works when updating typography.

Issue #2392, opened 2018-03-14
Label in TextArea should be full width, and not show any TextArea text above the label


sfdexter:
## Bugs

Follow the template below to ensure the quickest and most accurate response to your issue.

### What MDC Web Version are you using?
0.32.0 

### What browser(s) is this bug affecting?
Chome
> Please include the browser version. A user-agent string is also quite helpful.
Chrome  64.0.3282.186

### What OS are you using?
OS X

### What are the steps to reproduce the bug?
1) Navigate to demos page : https://material-components-web.appspot.com/text-field.html
2) Scroll down to "TextArea"
3) Enter "8" and then [Enter] 10 times

### What is the expected behavior?
Should not be able to see any characters behind the Text Area Label

> Please describe what the component/code should be doing that it's not.

### What is the actual behavior?

Can see a little bit of one of the "8"s above "Textarea Label"

kfranqueiro:
I checked with Alex and the reason for this is that we adjusted the label as much as we could without risking it overlapping (and "eating away at") the borders of the textarea at various zoom levels.

patrickrodee:
Good to know, thanks Ken. I'll discuss with the designers and see what they have to say about this.

williamernest:
Design guidance is that the label should be full width and not show any text at the top. Add to tracker.

PR #2823, opened 2018-05-25
Make mdc-rtl-reflexive sass mixin public.

abhiomkar:
Changes required for #2392 

Also updated the docs and usage in top app bar.


packages/mdc-ripple/package.json
```diff
@@ -15,7 +15,6 @@
   },
   "dependencies": {
     "@material/base": "^0.35.0",
-    "@material/theme": "^0.35.0",
-    "@material/animation": "^0.34.0"
```
williamernest: `@material/animation` is required for the `$mdc-animation-standard-curve-timing-function` variable in `_mixins.scss`

abhiomkar: Looks like I messed up with resetting the commits. This change is not supposed to be here. Reverting these changes.

abhiomkar: Reverted. PTAL.

packages/mdc-rtl/_mixins.scss
```diff
@@ -209,10 +209,34 @@
...
+//    left: 2px;
+//    right: 5px;
+//
+//    @include mdc-rtl {
+//      right: 2px;
+//      left: 5px;
+//    }
+//  }
+//  ```
+// An optional third $root-selector argument may also be given, which is passed to `mdc-rtl`.
```
kfranqueiro: third -> fifth

abhiomkar: Done.

PR #2840, opened 2018-05-29
docs(layout-grid): update README.md

### LINKED ISSUE #2836: mdc-layout-grid naming devices
0.35.2
Chrome
### What OS are you using?
Windows 10
- create an grid with the classes `mdc-layout-grid__cell--span-1`.
- add an extra class for devices (desktop, tablet & mobile), like how it's mentioned in [the documentation](https://github.com/material-components/material-components-web/tree/master/packages/mdc-layout-grid)
- desktop and tablet work like expected, but mobile does not.

the classname `mdc-layout-grid__cell--span-1-mobile` doesn't work because it should be `mdc-layout-grid__cell--span-1-phone`
I expected that the class `mdc-layout-grid__cell--span-1-mobile` would have worked. That being said, `mdc-layout-grid__cell--span-1-phone` works, so it's problem is within the naming of the classes.
`mdc-layout-grid__cell--span-1-mobile` won't work because it is the wrong classname
No extra information, just the documentation should be edited

robbyrice:
Rename 'mobile' to 'phone' for consistency.

Fixes #2836.

### DIFF
--- a/packages/mdc-layout-grid/README.md
+++ b/packages/mdc-layout-grid/README.md
@@ -81,7 +81,7 @@ CSS Class | Description
 `mdc-layout-grid__inner` | Mandatory, for wrapping grid cell
 `mdc-layout-grid__cell` | Mandatory, for the layout grid cell
 `mdc-layout-grid__cell--span-<NUMBER_OF_COLUMNS>` | Optional, specifies the number of columns the cell spans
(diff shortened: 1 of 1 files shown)


robbyrice:
CLA signed.

kfranqueiro:
Thanks for the PR!

robbyrice:
You're welcome! Glad to be able to contribute.

PR #2845, opened 2018-05-30
feat(chips): Expose method to begin chip exit animation

bonniezhou:
This is necessary for users to allow deleting a chip without interacting with the remove icon. A common use case of this is deleting the last chip by pressing the backspace key in an input.

### DIFF
--- a/demos/chips.html
+++ b/demos/chips.html
@@ -63,6 +63,9 @@
         <button id="input-chip-set-button" class="mdc-button mdc-button--dense">
           Add Input Chip
         </button>
+        <button id="input-chip-set-delete-button" class="mdc-button mdc-button--dense">
+          Delete Last Chip
+        </button>
         <div id="input-chip-set" class="mdc-chip-set mdc-chip-set--input">
           <div class="demo-chip mdc-chip" tabindex="0">
             <i class="material-icons mdc-chip__icon mdc-chip__icon--leading">face</i>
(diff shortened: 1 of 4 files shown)


test/unit/mdc-chips/mdc-chip.test.js
```diff
@@ -202,3 +202,9 @@ test('#isSelected proxies to foundation', () => {
   component.isSelected();
   td.verify(mockFoundation.isSelected());
 });
+
+test('#beginExit adds `mdc-chip--exit` class', () => {
```
kfranqueiro: Reuse constant in test description and assertion

demos/chips.html
```diff
@@ -281,8 +285,17 @@ <h2>Custom theme</h2>
           root && inputChipSetEl.removeChild(root);
         }
 
+        function deleteLastChip(evt) {
+          if (evt.type === 'click' || evt.key === 'Enter' || evt.keyCode === 13) {
```
kfranqueiro: This `if` is unnecessary; this function is only used in response to `click` which already triggers on enter/space for buttons.

demos/chips.html
```diff
@@ -281,8 +285,15 @@ <h2>Custom theme</h2>
           root && inputChipSetEl.removeChild(root);
         }
 
+        function deleteLastChip(evt) {
```
kfranqueiro: Nit: `evt` is unused and can be omitted

PR #2846, opened 2018-05-31
docs(mdc-layout-grid): Correct wrong reference to Sass variable name

frebro:


kfranqueiro:
Thanks for catching this!

PR #2847, opened 2018-05-31
docs: Add docs for releases/branches; rename releases.md more clearly

kfranqueiro:


docs/open_source/releases-and-branches.md
```diff
@@ -0,0 +1,73 @@
...
+MDC Web Catalog and MDC React both depend on MDC Web via npm dependencies. In order to be able to integrate
+breaking changes for the upcoming release ahead of time - preferably while the necessary changes are fresh in someone's
+mind - we will cut pre-releases of MDC Web when breaking changes are merged which warrant updates in the other
+repositories. These will be cut on an as-needed basis, with no set schedule.
+
+The Catalog and React repositories will each have a `next` branch, which will be updated to point to the pre-release.
+Pull requests should be filed against this branch for changes needed for the upcoming MDC Web release.
+
+Later, when the new version of MDC Web is released, the `next` branch will again be updated to point to the new release,
+and we will open a PR to squash & merge its contents onto `master`. Once this is done, the `next` branch will be reset
```
moog16: "squash & merge" - so are you saying we will be losing all the "next" branch commits?

docs/open_source/releases-and-branches.md
```diff
@@ -0,0 +1,73 @@
...
+against `master` to accommodate the subsequent release.
+
+The release process for pre-releases should be largely similar to for regular releases, with minor changes
```
moog16: "should be largely similar to regular releases"

docs/open_source/releases-and-branches.md
```diff
@@ -0,0 +1,73 @@
...
+
+1. `git checkout v0.36.0` (or e.g. `v0.36.1` if this is the second bugfix release since the last breaking-change release)
+2. `git cherry-pick -x <commits>`
+3. Follow the [release process](./release-process.md), choosing "patch" when lerna prompts for version
+4. Cherry-pick the publish and CHANGELOG commits onto `master` to ensure that bugfix releases are included in the overall history
+
+A branch can be created in/after step 1 if desired, but should be temporary - the bugfix release can be referenced via
+its git tag once the release process is done.
+
+We're planning to investigate scripting the process of finding non-breaking chores/fixes and cherry-picking them.
```
moog16: I think refactor should also go in the list. As long as there are no breaking changes no functionality should be changing.

docs/open_source/releases-and-branches.md
kfranqueiro: Oops. I accidentally some words in this sentence.

docs/open_source/releases-and-branches.md
kfranqueiro: I was mostly thinking to preserve the normal PR+review+squash/merge process for master rather than needing to temporarily override it. Otherwise we would need to temporarily disable some of the master branch protections in order to cherry-pick and push commits into master. On the other hand, this does mean it'd need to get reviewed quickly, preferably, but everything on `next` should have already been reviewed once in detail. It also hopefully means only resolving conflicts once, vs. potentially multiple times for cherry-picks.

Do you think there's a strong argument for keeping the commits separate?

kfranqueiro: I'm saying both.

Each change targeting `next` should go through the PR process and be squashed+merged into it. This may result in multiple commits to `next`.

When it's time to bring everything over into `master`, send a PR and squash+merge `next` into `master` (i.e. you end up with one "update for breaking changes" commit in master). The alternative is temporarily unprotecting master to just push either a squashed commit or cherry-picks straight through, but I'd prefer the PR approach as one final validation in case anything skewed between `master` and `next` since `next` was branched.

moog16: If we squash everything from next --> master, we would lose those commits for the Changelog, no?

kfranqueiro: That's a good point for MDC React, if that will involve individual breaking changes to it as well. I already followed the squash+merge process once for Catalog, where changelog isn't a consideration.

kfranqueiro:
Discussed with Matt; going to do some further thinking on how to streamline the process for MDC React, but will merge the docs as-is for now. MDC React isn't yet at the point of pre-emptively incorporating pre-release changes.

PR #2848, opened 2018-05-31
fix(checkbox): make checkmark in high contrast mode on IE visible.

aprigogin:


aprigogin:
not sure what travis complaining...

PR #2854, opened 2018-05-31
feat(text-field): Support for types- color, date, datetime-local, etc

namannehra:
Label will always float for the following types as they show some UI even if value is empty
- color
- date
- datetime-local
- month
- range
- time
- week

Added `getType` and `setType` methods to foundation. `setType` must be used to update type after initialisation.


namannehra:
Is this fine? Do I need fix something or make any changes?

packages/mdc-textfield/foundation.js
```diff
@@ -49,9 +48,15 @@ class MDCTextFieldFoundation extends MDCFoundation {
     return !this.isValid() && !this.isFocused_;
   }
 
+  /** @return {boolean} */
```
williamernest: Change to `@private` and update the name to `shouldAlwaysFloat_`. 
```
  /**
   * @return {boolean}
   * @private
   */
  get shouldAlwaysFloat_() {
    const type = this.getNativeInput_().type;
    return ALWAYS_FLOAT_TYPES.indexOf(type) >= 0;
  }

  /** @return {boolean} */
  get shouldFloat() {
    return this.shouldAlwaysFloat_ || this.isFocused_ || !!this.getValue() || this.isBadInput_();
  }
```

packages/mdc-textfield/foundation.js
```diff
@@ -250,21 +253,40 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
+   * @return {string} The type of the input Element.
+   */
+  getType() {
+    return this.getNativeInput_().type;
+  }
+
+  /**
+   * @param {string} type The type to set on the input Element.
+   */
+  setType(type) {
```
williamernest: Dynamically changing type isn't a common use case. This should be removed.

packages/mdc-textfield/foundation.js
```diff
@@ -250,21 +253,40 @@ class MDCTextFieldFoundation extends MDCFoundation {
...
-    if (shouldRemoveLabelFloat) {
+    if (!this.shouldFloat) {
       this.receivedUserInput_ = false;
     }
   }
 
+  /**
+   * @return {string} The type of the input Element.
+   */
+  getType() {
```
williamernest: This can be removed.

williamernest:
@namannehra Sorry for the delay. Thanks for fixing this for us!

PR #2858, opened 2018-06-01
feat(fab): Add Extended FAB

### LINKED ISSUE #2043: Add a mdc-fab--extended
This will include text in the FAB

abhiomkar:
This change implements only the basic version of Extended FAB (without transform animation and speed dial).

Fixes #2043


demos/fab.scss
```diff
@@ -26,3 +26,32 @@
...
+
+.demo-figure-fab-extended {
+  width: 100%;
+  padding: 0;
+  margin: 1em 40px;
+  box-sizing: border-box;
+}
+
+.demo-fab-extended-fluid {
+  @include mdc-fab-extended-fluid();
```
moog16: `@include  mdc-fab-extended-fluid` - don't need the parens if there are no args

demos/fab.scss
```diff
@@ -26,3 +26,32 @@
...
+
+@media (max-width: 700px) {
+  .demo-figure-fab-extended {
+    padding: 0 16px;
+    margin: 0;
+  }
+
+  .demo-fab-extended-fluid-responsive {
+
+    @include mdc-fab-extended-fluid();
```
moog16: remove parens

package-lock.json
```diff
@@ -8,24 +8,24 @@
...
-        "array-uniq": "^1.0.3",
-        "arrify": "^1.0.1",
-        "concat-stream": "^1.6.0",
-        "create-error-class": "^3.0.2",
-        "duplexify": "^3.5.0",
-        "ent": "^2.2.0",
-        "extend": "^3.0.1",
-        "google-auto-auth": "^0.9.0",
-        "is": "^3.2.0",
+        "array-uniq": "1.0.3",
```
moog16: A lot of changes in package-lock. Your changes shouldn't affect the packages. Are you on npm 6?

packages/mdc-fab/README.md
```diff
@@ -99,6 +101,7 @@ MDC FAB uses [MDC Theme](../mdc-theme)'s `secondary` color by default. Use the f
 Mixin | Description
 --- | ---
 `mdc-fab-accessible($container-color)` | Changes the FAB's container color to the given color, and updates the FAB's ink and ripple color to meet accessibility standards.
+`mdc-fab-extended-fluid()` | Makes the Extended FAB fluid to container, such as screen width or the layout grid.
```
moog16: Where'd you get the word fluid from? I understand its for responsive layouts, but all the mixin does is make the fab full-width.

packages/mdc-fab/mdc-fab.scss
```diff
@@ -31,6 +34,26 @@
...
+  align-items: center;
+  justify-content: center;
+  white-space: nowrap;
+  border-radius: $mdc-fab-extended-height_ / 2;
+}
+
+.mdc-fab__label {
+  padding-left: 8px;
+  justify-content: flex-start;
+  text-overflow: ellipsis;
```
moog16: I don't think the ellipsis will ever show. I think you need a max-width or something.
screen shot 2018-06-01 at 10 22 06

packages/mdc-fab/package-lock.json
```diff
@@ -0,0 +1,58 @@
+{
```
moog16: we don't have package-lock.json files in packages.

demos/fab.scss
```diff
@@ -26,3 +26,32 @@
...
+
+    &.mdc-fab {
```
moog16: This is overly specific. The rule should only be:

```css
.demo-fab-extended-fluid-responsive {
  @include mdc-fab-extended-fluid;
  margin: 16px 0;
}
```

demos/fab.scss
abhiomkar: Done.

demos/fab.scss
abhiomkar: Done.

package-lock.json
abhiomkar: This needs to be reverted.

packages/mdc-fab/README.md
abhiomkar: Yah, that is true. I got this term from Spec

https://material.io/design/components/buttons-floating-action-button.html#extended-fab

packages/mdc-fab/mdc-fab.scss
abhiomkar: This was required for fluid Extended FAB. But is is probably a good idea to add max-width to 100%.

packages/mdc-fab/package-lock.json
abhiomkar: Removed.

demos/fab.scss
abhiomkar: Added this to improve specificity. We can remove this extra specificity since I've moved all styles from html file to sass file.

packages/mdc-fab/_mixins.scss
```diff
@@ -50,6 +50,10 @@
   @include mdc-theme-prop(color, $color);
 }
 
+@mixin mdc-fab-extended-fluid {
```
kfranqueiro: What's the thought process behind this being a mixin rather than just another modifier class? Seems like this could be done without requiring Sass.

packages/mdc-fab/mdc-fab.scss
```diff
@@ -31,6 +34,27 @@
   @include mdc-fab--mini_;
 }
 
+.mdc-fab--extended {
+  @include mdc-typography(button);
+
+  display: inline-flex;
```
kfranqueiro: Shouldn't these flex-related styles already be present on `.mdc-fab` (via the `base_` mixin)? They shouldn't need to be repeated here.

packages/mdc-fab/_mixins.scss
abhiomkar: We want user to be able to use this mixin within CSS media query.

packages/mdc-fab/mdc-fab.scss
abhiomkar: Good point. Removed redundant styles from here.

abhiomkar:
Created [material-components-web-catalog/issues/164](https://github.com/material-components/material-components-web/pull/2858) to update catalog with Extended FAB demo.

abhiomkar:
Added golden screenshot URLs.

packages/mdc-fab/mdc-fab.scss
```diff
@@ -31,6 +34,24 @@
   @include mdc-fab--mini_;
 }
 
+.mdc-fab--extended {
```
kfranqueiro: Since MDC FAB uses private mixins for most of these selectors' styles, can we do the same for the new ones? It'll result in most of the actual styles being in the same file.

test/screenshot/mdc-fab/classes/extended.html
```diff
@@ -0,0 +1,39 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="../../out/fixture.css">
+    <link rel="stylesheet" href="../../out/mdc.fab.css">
+    <link rel="stylesheet" href="../../out/mdc-fab/fixture.css">
+  </head>
+
+  <body class="test-body">
+    <main class="test-main test-main--fab">
+      <div class="test-grid">
+        <div class="test-cell">
```
kfranqueiro: We should be explicitly specifying a width for `test-cell`; see other screenshot tests for examples. (Usually it's a multiple of 10 + 1 to show the grid lines on both ends, e.g. 81px.)

abhiomkar:
As discussed with Ken, seems like Roboto & Google Sans fonts render slightly differently on Windows Chrome browser because of inconsistent [vertical metrics](https://glyphsapp.com/tutorials/vertical-metrics). I tested this with Open Sans font and Arial font both has consistent base line across browsers.

roboto google sans font rendering

test/screenshot/mdc-fab/fixture.scss
```diff
@@ -22,3 +22,7 @@
   width: 81px;
   height: 81px;
 }
+
+.test-cell--extended {
+  width: 191px;
```
kfranqueiro: Also add a height (also ending in 1)

Don't forget to regenerate goldens (it looks like the ones currently in this PR aren't current with this width update)

packages/mdc-fab/_mixins.scss
```diff
@@ -122,12 +127,28 @@ $mdc-fab-icon-enter-duration_: 180ms;
   height: 40px;
 }
 
+@mixin mdc-fab--extended_ {
+  width: auto;
+  max-width: 100%;
+  height: $mdc-fab-extended-height_;
+  padding: 0 20px 0 16px;
+  border-radius: $mdc-fab-extended-height_ / 2;
+  white-space: nowrap;
```
kfranqueiro: Would this style make more sense specifically on the label element? (Tangentially, flex-wrap is already nowrap by default)

packages/mdc-fab/mdc-fab.scss
```diff
@@ -31,6 +32,15 @@
...
+  @include mdc-typography(button);
```
kfranqueiro: I would move this include into the extended_ mixin as well, thinking forward to if we end up with a reason to expose it. (We do the same with e.g. `mdc-button-base_`)

PR #2870, opened 2018-06-04
chore(elevation): Update to use latest theme version

kfranqueiro:
I noticed while testing installing 0.36.0 that elevation ends up with a nested `node_modules/@material/theme` because it is still depending on a really old version. This fixes that, and takes advantage of the `mdc-theme-prop-value` function that now exists, which elevation was pretty much repeating code from (with a potentially less-helpful error message).

### DIFF
--- a/packages/mdc-elevation/package.json
+++ b/packages/mdc-elevation/package.json
@@ -14,6 +14,6 @@
   },
   "dependencies": {
     "@material/animation": "^0.34.0",
-    "@material/theme": "^0.4.0"
+    "@material/theme": "^0.35.0"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/mdc-elevation/package.json
```diff
@@ -14,6 +14,6 @@
   },
   "dependencies": {
     "@material/animation": "^0.34.0",
-    "@material/theme": "^0.4.0"
+    "@material/theme": "^0.35.0"
```
patrickrodee: Is 0.35.0 the most recent version of the theme packages?

kfranqueiro: [Yes](https://github.com/material-components/material-components-web/blob/v0.36.0/packages/mdc-theme/package.json#L4)

PR #2881, opened 2018-06-04
chore: Create script to cherry-pick commits for patch releases

kfranqueiro:
To test:

* Checkout this branch
* Run `node scripts/cherry-pick-patch-commits`

It should check out the latest tag, cherry-pick commits, and run the build and unit tests locally, then provide output summarizing what was cherry-picked, what was skipped, and what was conflicted, and whether the build and tests succeeded. You will then be on a detached HEAD with cherry-picked commits on top of the tag you selected. None of these cherry-picked commits should be features or breaking changes.

For more of a stress-test, you can also pass it a previous release tag, e.g. `node scripts/cherry-pick-patch-commits v0.35.2`. (This will fail the build, which is expected, because there is a fix that depends on a feature; the fix was cherry-picked but the feature wasn't.)

### DIFF
--- a/scripts/cherry-pick-patch-commits.js
+++ b/scripts/cherry-pick-patch-commits.js
@@ -1,0 +1,164 @@
+/**
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Identifies and cherry-picks commits that are not features and do not contain breaking changes.
+ *
+ * Defaults to operating against the latest non-pre-release tag, but a tag can be specified via command-line argument.
+ * This will automatically attempt to cherry-pick appropriate commits since the tag, but will abort and skip any
+ * cherry-picks that result in conflicts.
+ *
+ * Note that this does not create a branch - you will be in detached HEAD state. You can create a branch afterwards
+ * if you want, via `git checkout -b <branchname>`.
+ */
+
+const args = process.argv.slice(2);
+
+const parser = require('conventional-commits-parser');
+const parserOpts = require('conventional-changelog-angular/parser-opts');
+const simpleGit = require('simple-git/promise')();
+const {spawnSync} = require('child_process');
+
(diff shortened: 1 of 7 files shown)


scripts/cherry-pick-patch-commits.js
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  return results;
+}
+
+// Given a string containing a command and arguments, runs the command and returns true if it exits successfully.
+// The command's I/O is piped through to the parent process intentionally to be visible in the console.
+function checkSpawnSuccess(command) {
+  const parts = command.split(' ');
+  return spawnSync(parts[0], parts.slice(1), {stdio: 'inherit'}).status === 0;
```
patrickrodee: The `parts.slice(1)` is because `spawnSync` takes an array?

scripts/cherry-pick-patch-commits.js
```diff
@@ -0,0 +1,164 @@
...
+    if (shouldSkipCommit(logLine)) {
+      results.skipped.push(logLine);
+      continue;
+    }
+
+    try {
+      await simpleGit.raw(['cherry-pick', '-x', logLine.hash]);
+      results.successful.push(logLine);
+    } catch (e) {
+      if (e.message.includes(CONFLICT_MESSAGE)) {
```
patrickrodee: This part confuses me. Where does this error message come from? Is it from the git cherry-pick command?

scripts/cherry-pick-patch-commits.js
kfranqueiro: Yeah, it takes a command + an array of arguments to that command.

scripts/cherry-pick-patch-commits.js
kfranqueiro: Yes, it's from the command.

> If the process times out or has a non-zero exit code, this method will throw an Error that will include the full result of the underlying child_process.spawnSync().

scripts/cherry-pick-patch-commits.js
```diff
@@ -0,0 +1,166 @@
...
+    to: 'origin/master',
+    // simple-git defaults to only including subject + tag/branch (%s%d) in message, which adds noise and is not useful
+    // for finding BREAKING CHANGEs (which are in body, %b).
+    format: {
+      hash: '%H', // Contains full hash only, for cherry-picking
+      message: '%s\n\n%b', // Contains subject + body for passing to conventional-commits-parser
+      oneline: '%h %s' // Contains abbreviated hash + subject for summary lists
+    },
+    // Use a splitter that is very unlikely to be included in commit descriptions
+    splitter: ';;;'
```
patrickrodee: Where does this splitter get removed or modified, or is it coming from one of the git commands?

kfranqueiro: It's used within `simple-git`. It defaults to `;` which I figure is risky if I'm including the body of commit messages in what gets split.

kfranqueiro: See https://github.com/steveukx/git-js/blob/master/src/git.js#L1228

patrickrodee: Yeah, good call. That’s a safer choice.

PR #2892, opened 2018-06-06
fix(typography): Update letter-spacing units from em to rem

williamernest:
All other sizes for typography are in `rem`, so the letter-spacing function should also return a `rem` value.


kfranqueiro:
Pat and I discussed this last week, and this has many unwanted side effects. Moreover, `em` was always used for letter-spacing, even prior to the 0.35.0 typography updates.

PR #2893, opened 2018-06-06
feat(chips): Make chip exit on trailing icon click optional

### LINKED ISSUE #2869: Make chip exit on trailing icon interaction optional
## Feature Request

As of 0.35.2, it is not possible to prevent a chip from exiting when the trailing icon is interacted with. In my case, I'd like to remove the chip only after manual confirmation. The `.mdc-chip--exit` class starting the exit sequence of the chip is immediately applied on handling the trailing icon interaction, making it impossible to abort the process unless overwriting the transition values for the class to nothing.

bonniezhou:
Fixes #2869.

Taking name suggestions to shorten `shouldRemoveOnTrailingIconClick`....


demos/chips.html
```diff
@@ -243,12 +256,34 @@ <h2>Custom theme</h2>
...
+        inputButton.addEventListener('click', addInputChip);
+        input.addEventListener('keydown', addInputChip);
+        deleteButton.addEventListener('click', deleteLastChip);
+        inputChipSetEl.addEventListener('MDCChip:removal', removeChip);
+        trailingIconToggle.addEventListener('change', resetInputChipSet);
+
+        function resetInputChipSet() {
+          inputChipSetComponent.destroy();
+          if (this.checked) {
+            inputChipSetComponent = new mdc.chips.MDCChipSet(inputChipSetEl, undefined, (chipEl) => {
```
kfranqueiro: Use `function`, not `=>` (IE 11)

bonniezhou: argh, nice catch

packages/mdc-chips/README.md
```diff
@@ -211,12 +211,13 @@ To use the `MDCChip` and `MDCChipSet` classes, [import](../../docs/importing-js.
...
 Method Signature | Description
 --- | ---
-`get foundation() => MDCChipFoundation` | Returns the foundation
 `isSelected() => boolean` | Proxies to the foundation's `isSelected` method
 `beginExit() => void` | Begins the exit animation which leads to removal of the chip
 
 Property | Value Type | Description
 --- | --- | ---
+`foundation` | MDCChipFoundation | The foundation
+`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `shouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods
```
kfranqueiro: Need to add `get` to the foundation method name in the description.

Also, an additional thought: add a sentence here explaining that if this is set to `false`, manually call `beginExit` on the chip component to control when to remove it.

packages/mdc-chips/chip/foundation.js
```diff
@@ -175,7 +194,10 @@ class MDCChipFoundation extends MDCFoundation {
     evt.stopPropagation();
     if (evt.type === 'click' || evt.key === 'Enter' || evt.keyCode === 13) {
       this.adapter_.notifyTrailingIconInteraction();
-      this.adapter_.addClass(cssClasses.CHIP_EXIT);
+      if (this.shouldRemoveOnTrailingIconClick_) {
+        // Begins the exit animation which leads to removal of the chip
+        this.adapter_.addClass(cssClasses.CHIP_EXIT);
```
kfranqueiro: It occurs to me that this is effectively doing the same thing that `beginExit` on the component is doing. Do you think we should create a `beginExit` foundation method and proxy to it from the component's method, so that both can reuse the same code?

packages/mdc-chips/README.md
bonniezhou: Sounds good, added as a `NOTE` to stick with having one line per table entry

packages/mdc-chips/README.md
```diff
@@ -211,13 +211,15 @@ To use the `MDCChip` and `MDCChipSet` classes, [import](../../docs/importing-js.
...
-`get foundation() => MDCChipFoundation` | Returns the foundation
 `isSelected() => boolean` | Proxies to the foundation's `isSelected` method
 `beginExit() => void` | Begins the exit animation which leads to removal of the chip
 
 Property | Value Type | Description
 --- | --- | ---
+`foundation` | MDCChipFoundation | The foundation
+`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods
 `ripple` | `MDCRipple` | The `MDCRipple` instance for the root element that `MDCChip` initializes
+>_NOTE_: If `shouldRemoveOnTrailingIconClick` is set to false, you must manually call `beginExit()` on the chip to remove it.
```
kfranqueiro: Add empty line between table and blockquote

packages/mdc-chips/chip/foundation.js
bonniezhou: I thought about that but realized that's actually more code?
...But I guess it makes sense pattern-wise. Ok I'll update.

kfranqueiro: It probably is more code ultimately, but in this case the trade-off is avoiding repeating the same logic across modules which risks skew later.

trimox:
@bonniezhou Offering name suggestion, `removable`.

bonniezhou:
I think that could be misinterpreted as whether you can remove the chip at all, not whether it will be removed automatically upon click. :(

Issue #2819, opened 2018-05-25
Text field caret color does not match as per spec


cubetastic33:
### What MDC Web Version are you using?

0.35.2

### What browser(s) is this bug affecting?

I tested it on the following browsers, and the bug was consistent across them:
Firefox Quantum 60.0.1 (64-bit)
Google Chrome Version 66.0.3359.181 (Official Build) (64-bit)
Opera 53.0.2907.68 Stable

### What OS are you using?

Ubuntu 18.04 LTS

### What are the steps to reproduce the bug?
1. Open the codepen [here](https://codepen.io/cubetastic/pen/ELBZJO).
2. Hover on the text field. The border color should turn black.
3. Click on it and start typing. The border color might seem blurry, but it will be purple, and the caret color is still black.
4. Click outside to deactivate, and then hover on top of it. It won't turn full black, rather, you'll see somewhat of a double border.

### What is the expected behavior?

The border radius should turn black on hover even after it has text in it, and the caret color should be the same color as the border.
This is in the Material Docs for Text field:
![image](https://user-images.githubusercontent.com/33519300/40524858-969a21dc-5ffa-11e8-9b78-c7b1ec46bdd5.png)

### What is the actual behavior?

The caret color remains black, and the border appears double when you hover after typing in text:
![image](https://user-images.githubusercontent.com/33519300/40525069-b2180cf2-5ffb-11e8-9a13-3b2688d03a0a.png)
 as compared to
![image](https://user-images.githubusercontent.com/33519300/40525050-7fcd39f2-5ffb-11e8-9a55-e49c571344a6.png)

cubetastic33:
@kfranqueiro  I made a [codepen](https://codepen.io/cubetastic/pen/bKGzro), which seems to address the outline issue properly, and the caret color is changing in Google Chrome, Firefox, and Opera. This is also what I was using before I knew about MDC Web.

PR #2894, opened 2018-06-06
fix(text-field): Update caret color to match spec

williamernest:
fixes: #2819 
Added a mixin to support modifying the color.


packages/mdc-textfield/_mixins.scss
```diff
@@ -124,6 +124,12 @@
   }
 }
 
+@mixin mdc-text-field-caret-color($color) {
+  &:not(.mdc-text-field--disabled) {
```
bonniezhou: I think we can remove the `not` selector and the private mixin, since disabled text fields don't have a caret.

Issue #2872, opened 2018-06-04
New `mdc-drawer`


lynnjepsen:
- [x] Modeless
- [x] Shows the icons and the text
- [x] Uses mdc-list (see #2851 )
- [x] New Sass mixins for customizing colors/typography of the mdc-list within the drawer
- [x] New `mdc-drawer-surface-fill-color` Sass mixin
- [x] New `mdc-drawer-title-ink-color` Sass mixin
- [x] New `mdc-drawer-active-overlay-color` Sass mixin

moog16:
A few things will be left out of my changes. This is because list is missing some mixins. I've created https://github.com/material-components/material-components-web/issues/2903.

moog16:
Also added this bug https://github.com/material-components/material-components-web/issues/2906, which will be worked on as a separate issue to be checked into feat/navigation-drawer feature branch.

abhiomkar:
Fixed in #3417

PR #2904, opened 2018-06-07
feat(drawer): Add styles and mixins for standard drawer

moog16:
fixes https://github.com/material-components/material-components-web/issues/2872


packages/mdc-drawer/standard/mdc-standard-drawer.scss
```diff
@@ -0,0 +1,99 @@
...
+  a.mdc-list-item {
+    @include mdc-typography(body2);
+
+    height: calc(48px - 2 * #{$mdc-drawer-list-item-spacing});
+    margin: $mdc-drawer-list-item-spacing 0;
+    padding: 0 $mdc-drawer-surface-padding/2;
+  }
+  // stylelint-enable selector-no-qualifying-type
+
+  .mdc-list-group__subheader {
```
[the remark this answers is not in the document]
kfranqueiro: Also, this should really be called subtitle, not subheader, according to the design spec. Maybe subtitle within header is just a specialization of this subtitle?

kfranqueiro:
"Labels" seems like it needs margin adjustment...

Also, are we sure we're using the correct typography scale for subtitle/subheader? It seems heavier than what's pictured in the spec.

image

packages/mdc-drawer/standard/mdc-standard-drawer.scss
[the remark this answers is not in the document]
moog16: Spec has 13dp for the label, but the params spreadsheet state 14dp (both subtitle2 and body2). I changed from subtitle2 --> body2 and looks closer. material.io spec doesn't explicitly call subtitle or subheader out. The "subtitle" that is called out in spec is referring to the primary text of the list item.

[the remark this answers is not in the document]
moog16: RE list vs drawer stylesheet: I had the same question to design. They stated nav drawer list styles will only apply to nav drawer. Since its not specifically called out in list's spec, I believe its left open to the end designer/developer.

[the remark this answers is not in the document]
kfranqueiro: I was referring to this wording:

image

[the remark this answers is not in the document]
moog16: Ah gotcha - I do have every reference in drawer as 'subtitle'. I can change subheader --> subtitle in list. It will be a breaking change.

packages/mdc-drawer/standard/mdc-drawer-standard.scss
```diff
@@ -0,0 +1,99 @@
...
+  @include mdc-drawer-group-header-text-color($mdc-drawer-subtitle-ink-color, $mdc-drawer-subtitle-ink-opacity);
+  @include mdc-drawer-surface-fill-color($mdc-drawer-surface-fill-color, $mdc-drawer-surface-fill-opacity);
+  @include mdc-drawer-icon-ink-color($mdc-drawer-icon-ink-color, $mdc-drawer-icon-ink-opacity);
+  @include mdc-drawer-meta-ink-color($mdc-drawer-icon-ink-color, $mdc-drawer-icon-ink-opacity);
+  @include mdc-drawer-item-text-color($mdc-drawer-inactive-ink-color, $mdc-drawer-inactive-ink-opacity);
+  @include mdc-drawer-item-activated-text-color($mdc-drawer-activated-ink-color, $mdc-drawer-activated-ink-opacity);
+
+  width: $mdc-drawer-width;
+  min-width: $mdc-drawer-width;
+  border-right-width: 1px;
```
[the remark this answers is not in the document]
moog16: If I use `border-right`, the sass color mixin doesn't work `mdc-drawer-border-color`

packages/mdc-drawer/_mixins.scss
```diff
@@ -14,98 +14,89 @@
...
-    border-bottom: 1px solid rgba(0, 0, 0, .12);
-
-    // TODO(sgomes): replace with global breakpoints when we have them
-    @media (min-width: 600px) {
-      height: 64px;
-    }
+@import "@material/list/mixins";
+@import "@material/ripple/mixins";
+
+@mixin mdc-drawer-active-overlay-color($color) {
```
kfranqueiro: Suggestions:

* Rename this `activated-overlay-color`
* Remove `mdc-states-selected`, since IIUC drawer items are only ever activated, not selected

The original styles in list combined both to accommodate both drawer (activated) and menu (selected) contexts.

packages/mdc-drawer/_mixins.scss
```diff
@@ -14,98 +14,89 @@
...
+
+  @include mdc-list-item-graphic-ink-color($value);
 }
 
-@mixin mdc-drawer-ink-color($color) {
-  @include mdc-theme-prop(color, $color);
+@mixin mdc-drawer-item-activated-text-color($color, $opacity) {
+  $value: rgba(mdc-theme-prop-value($color), $opacity);
+  // stylelint-disable selector-no-qualifying-type
+  .mdc-list-item--selected,
```
kfranqueiro: Same comment as above RE removing `--selected` assuming drawer items don't have selected state

packages/mdc-drawer/_mixins.scss
```diff
@@ -14,98 +14,89 @@
...
-    top: 0;
-    right: 0;
-    bottom: 0;
-    left: 0;
-    align-items: flex-end;
-    box-sizing: border-box;
-    padding: 16px;
-  }
-  // stylelint-enable selector-max-type
+@mixin mdc-drawer-divider-color($color, $opacity) {
```
kfranqueiro: OOC, is there a reason why we're wrapping a bunch of list mixins here rather than making the list mixins support opacity?

packages/mdc-drawer/mdc-drawer.scss
```diff
@@ -14,6 +14,4 @@
 // limitations under the License.
 //
 
-@import "./persistent/mdc-persistent-drawer";
-@import "./permanent/mdc-permanent-drawer";
-@import "./temporary/mdc-temporary-drawer";
+@import "./standard/mdc-drawer-standard";
```
kfranqueiro: FYI for future PRs: We should figure out what common subset of styles applies across all drawers, and provide those in the context of `.mdc-drawer` here. We should avoid repeating the mistake of the current `mdc-drawer` which effectively repeated a bunch of the same styles for 3 separate selectors.

packages/mdc-drawer/standard/mdc-drawer-standard.scss
```diff
@@ -0,0 +1,99 @@
...
+
+@import "@material/rtl/mixins";
+@import "@material/typography/mixins";
+@import "@material/typography/variables";
+@import "@material/theme/variables";
+@import "@material/theme/mixins";
+@import "../variables";
+@import "../mixins";
+
+$mdc-drawer-width: 256px;
```
kfranqueiro: Why are these here rather than in `variables`?

packages/mdc-drawer/standard/mdc-drawer-standard.scss
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  .mdc-list,
+  .mdc-list-group .mdc-list {
+    padding: $mdc-drawer-surface-padding/2;
+    padding-top: 0;
+  }
+
+  // stylelint-disable selector-no-qualifying-type
+  .mdc-list-item,
```
kfranqueiro: Is there supposed to be a default corner radius on list items? The design guidelines seem to show one.

packages/mdc-drawer/_mixins.scss
moog16: I think I did this after reading the opacity doc. I was thinking that we would eventually do list and other components in a component wide change. I think soon after I did it this way designers were going to take it back to the drawing board.

packages/mdc-drawer/mdc-drawer.scss
moog16: Yup that makes sense. I think the next PR should give me a better idea of what can be in `.mdc-drawer`.

packages/mdc-drawer/standard/mdc-drawer-standard.scss
moog16: I didn't use that variable anywhere else, but perhaps a user would. I'll move it.

moog16:
@kfranqueiro updated and responded to comments

packages/mdc-drawer/_mixins.scss
kfranqueiro: So I guess part of this was to localize the changes to drawer rather than affecting list right away... But I'm wondering whether some of these mixins even need to exist, rather than simply using list mixins directly, if they're doing nothing but calling the list mixins in the long run?

Could we just cross-link to the list mixins in the drawer documentation and cut a bunch of these mixins, presuming we eventually update the list mixins appropriately once we know how we want to handle opacity?

demos/drawer/drawer.scss
```diff
@@ -59,36 +60,48 @@
...
+  @include mdc-drawer-border-color(lightblue, $mdc-drawer-divider-opacity);
+  @include mdc-drawer-group-header-text-color(darkolivegreen, $mdc-drawer-subtitle-ink-opacity);
+  @include mdc-drawer-surface-fill-color(pink, $mdc-drawer-surface-fill-opacity);
+  @include mdc-drawer-icon-ink-color(darkolivegreen, $mdc-drawer-icon-ink-opacity);
+  @include mdc-drawer-title-ink-color(darkolivegreen, 1);
+  @include mdc-drawer-activated-overlay-color(darkcyan);
+  @include mdc-drawer-item-text-color(darkolivegreen, $mdc-drawer-inactive-ink-opacity);
+  @include mdc-drawer-item-activated-text-color(darkcyan, 1);
+
+  .mdc-list-item--selected {
```
kfranqueiro: Putting aside that this should be activated rather than selected, why do we have a mixin for activated text color but not activated icon color? (See also question RE having a general item mixin and an icon mixin)

demos/drawer/drawer.scss
```diff
@@ -59,36 +60,48 @@
...
-.demo-form-field {
-  margin: 16px 0;
+.demo-drawer--custom {
+  @include mdc-drawer-border-color(lightblue, $mdc-drawer-divider-opacity);
+  @include mdc-drawer-group-header-text-color(darkolivegreen, $mdc-drawer-subtitle-ink-opacity);
+  @include mdc-drawer-surface-fill-color(pink, $mdc-drawer-surface-fill-opacity);
+  @include mdc-drawer-icon-ink-color(darkolivegreen, $mdc-drawer-icon-ink-opacity);
+  @include mdc-drawer-title-ink-color(darkolivegreen, 1);
+  @include mdc-drawer-activated-overlay-color(darkcyan);
+  @include mdc-drawer-item-text-color(darkolivegreen, $mdc-drawer-inactive-ink-opacity);
```
kfranqueiro: Is there a reason we seem to have mixins explicitly targeting _only_ the text and _only_ the icon?

Shouldn't there be plenty of cases where both have the same color? Can we target both text and icon in one mixin for the simple case, then have another specifically for overriding just the icon? That way if you want to customize both to the same color, you only have to call one mixin.

(Technically the `mdc-list-item-primary-text-ink-color` mixin doesn't restrict its selector, so it can be used for this purpose... maybe we should eventually rename it to reflect that. I guess this whole comment also ties back to the duplication of mixins between list and drawer...)

packages/mdc-drawer/README.md
```diff
@@ -2,23 +2,20 @@
...
-It implements permanent, persistent, and temporary drawers. Permanent drawers are CSS-only and require no JavaScript, whereas persistent and temporary drawers require JavaScript to function, in order to respond to
-user interaction.
+The MDC Navigation Drawer is used to organize access to destinations and other functionality on an app.
 
 ## Design & API Documentation
 
 <ul class="icon-list">
   <li class="icon-list-item icon-list-item--spec">
-    <a href="https://material.io/go/design-navigation-drawer">Material Design guidelines: Navigation drawer</a>
+    <a href="https://material.io/design/components/navigation-drawer.html">Material Design guidelines: Navigation drawer</a>
```
kfranqueiro: I think we'd prefer to keep the go link here, so the site team has flexibility to change things without these links breaking.

packages/mdc-drawer/_mixins.scss
```diff
@@ -14,98 +14,86 @@
...
+  $value: rgba(mdc-theme-prop-value($color), $opacity);
+  // stylelint-disable selector-no-qualifying-type
+  .mdc-list-item,
+  a.mdc-list-item {
+    @include mdc-list-item-primary-text-ink-color($value);
   }
+  // stylelint-enable selector-no-qualifying-type
 }
 
-@mixin mdc-drawer-fill-color-accessible($color) {
```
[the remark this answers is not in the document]
moog16: I just didn't list it off in my design doc, and didn't think this was still a pattern with other packages - although I see it now in top-app-bar. Adding it back.

packages/mdc-drawer/_mixins.scss
moog16: So remove all the drawer/list mixins? I think most of the mixins are specific to drawer w/ the exception of the `divider`, `meta`, and `graphic`. It would be strange to remove divider, meta, and graphic, but still have the others proxied from the drawer mixin files.

I do however like the primary-ink-color mixin changing both meta and graphic, but then having the other 2 mixins to override the meta and grahpic.

demos/drawer/standard-drawer.html
```diff
@@ -156,50 +127,51 @@ <h1 class="mdc-typography--headline4">Permanent Drawer</h1>
...
+        document.querySelector('#toggle-wide').addEventListener('click', function() {
+          extraWide.style.display = extraWide.style.display ? '' : 'none';
+        });
+        var extraTall = document.querySelector('#extra-tall-content');
+        extraTall.style.display = 'none';
+        document.querySelector('#toggle-tall').addEventListener('click', function() {
+          extraTall.style.display = extraTall.style.display ? '' : 'none';
+        });
+
+        var activatedClass = 'mdc-list-item--selected';
```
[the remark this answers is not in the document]
moog16: got it - selected is more for components like chips, while activated is more permanent like tabs or drawer.

packages/mdc-drawer/_mixins.scss
kfranqueiro: I was wondering if that was the angle you were coming from. I guess that's fair from a perspective of having all relevant customizations in one place, and the fact they really just proxy is an implementation detail.

I suppose the upside to this is it gives us the freedom to tweak it uniquely from list if we need to, while the downside is if we ever need to make breaking changes to list APIs, we'll need to be careful to update drawer and menu as well. (Hopefully we can add screenshot tests for these revamped packages SRTL to help mitigate this.)

demos/drawer/drawer.scss
moog16: I like this suggestion - I will update the text-ink-color to include graphic and meta mixins

kfranqueiro: Does it actually need those in order to work? I was hoping they would Just Work based on simply setting the color of the list item as a whole. The point was to both reduce excess Sass _and_ output CSS.

moog16: I thought we were going to do this change so the user didn't need 2 mixins (`mdc-drawer-item-text-color` and `mdc-drawer-icon-ink-color`), but instead just 1 which incapsulates both the text and icon colors.

moog16: Spoke offline - because list sets graphic and meta colors in baseline styles we can't just use the primary-text-color mixin for all list item colors. We are keeping them as separate mixins.

demos/drawer/drawer.scss
```diff
@@ -59,36 +59,53 @@
...
+  @include mdc-drawer-item-activated-text-color(darkcyan, 1);
+
+  .mdc-list-item--activated {
+    @include mdc-drawer-icon-ink-color(darkcyan, 1);
+  }
 }
+
+.demo-drawer--accessible {
+  @include mdc-drawer-fill-color-accessible(darkcyan, 1);
+  @include mdc-drawer-activated-overlay-color(pink);
```
kfranqueiro: Should probably set activated text color to this as well?

demos/drawer/standard-drawer.html
```diff
@@ -138,7 +119,7 @@ <h1 class="mdc-typography--headline4">Permanent Drawer</h1>
                 <div class="mdc-radio__inner-circle"></div>
               </div>
             </div>
-            <label for="theme-radio-accessible">Accessible Theme</label>
+            <label for="theme-radio-custom">Accessible Theme</label>
```
kfranqueiro: `for` points to the wrong input

moog16:
Per convo with design: added mdc-drawer-icon-activated-ink-color, but not for `meta` element since meta should only appear in cascading navigation which should never have the activated state.

PR #2912, opened 2018-06-08
feat(list): new sass mixins

### LINKED ISSUE #2903: MDC List: Add missing sass mixins for MDC Drawer Support
The new implementation of drawer (https://github.com/material-components/material-components-web/issues/2872), requires a new list item border radius mixin seen in https://material.io/design/components/navigation-drawer.html#specs. So I propose to add the following and add the functionality to complete #2872:

1. `mdc-list-item-border-radius($radius)`

moog16:
fixes #2903


moog16:
Thanks will merge after feat/standard-drawer is merged

PR #2913, opened 2018-06-08
chore(list): removed unused demo list sass

moog16:


kfranqueiro:
Thanks for catching this!

PR #2927, opened 2018-06-12
fix(checkbox): support high contrast mode in Firefox on Windows

aprigogin:
**NOTE:** The `stroke="white"` SVG attribute has been removed as it is no longer needed.


packages/mdc-checkbox/_mixins.scss
```diff
@@ -69,10 +69,8 @@
 }
 
 @mixin mdc-checkbox-ink-color($color) {
-  .mdc-checkbox__checkmark-path {
-    // !important is needed here because a stroke must be set as an attribute on the SVG in order
```
aprigogin: I did not find this comment being true at all. In fact I've moved stroke color definition from html to css.

PR #2930, opened 2018-06-13
docs(mdc-icon-button): Fix event description

matsp:
The docs are just missing the right event.


kfranqueiro:
Thanks for catching this!

PR #2945, opened 2018-06-18
Fix: Same margin than text-field in boxed style

vokimon:
Using boxed style, Text-fields have a 16px margin than selects haven't. If you layout them horizontally (in the same row), they shown distinct vertical positions. If you layout them vertically, the spacing between fields is irregular. This adds the same margin-top to the selects than it is defined for the textfields.

### DIFF
--- a/test/screenshot/commands/build.js
+++ b/test/screenshot/commands/build.js
@@ -26,6 +26,9 @@ const processManager = new ProcessManager();
 
 module.exports = {
   async runAsync() {
+    // Travis sometimes forgets to emit this
+    logger.foldEnd('install.npm');
+
     const webpackArgs = [];
     const shouldBuild = await this.shouldBuild_();
     const shouldWatch = await this.shouldWatch_();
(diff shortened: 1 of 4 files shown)


patrickrodee:
Thanks for the PR, @vokimon! We always appreciate community contributions.

After discussing this, we'd like to actually do the inverse of your PR. Let's remove the margin-top (and margin-bottom) from the MDC Text Field component. I'll explain why below.

Having margin on the top-level element of a component may make it look nice in general, but quickly becomes problematic when used in any custom context. The margin can easily cause layout problems where a component can't be vertically lined up next to a horizontally adjacent item. That's no good as it makes assumptions about the context in which the component is used.

By removing the margin from the top-level elements, we make no assumptions about the surrounding context of the component. Instead, the user can apply margin as they need it.

If you could update your PR with those changes (remove `margin-top` and `margin-bottom` from MDC Text Field and revert the `margin-top` addition to MDC Select), that would be great!

Also, make sure you sign the CLA 👍

vokimon:
Indeed i agree a lot with your proposal. On it.

vokimon:
I just removed the margin-top. Not sure if there is any margin-bottom. Besides the examples i am running just exercise the boxed style.

Issue #2918, opened 2018-06-12
mdc-icon-button: documentation inconsistencies and dead code, unused adapter methods


gjdev:
The mdc-icon-button seems to be an inconsistent copy of the deprecated mdc-icon-toggle. Some code that is not used anymore is left in mdc-icon-button. A big part of the implemented functionality isn't documented anymore.

* The mdc-icon-toggle doesn't change the tabIndex of it's component. This seems right as it's documented to be used on anchor and button tags, that already have the proper tabIndex applied by default. However it still exposes the getTabIndex/setTabIndex adapter methods, that are now useless. setTabIndex is never called. getTabIndex is called, but never really used. Both adapter methods should be removed from the code and the documentation.
* The adapter method removeAttr is never called, and should be removed from code and documentation.
* The toggled variant has styling for a disabled button: mdc-icon-button--disabled. This is not documented. It's also weird. More appropriate would be to apply the disabled styling based on the availability of a :disabled selector. (Especially since the foundation/component never use the mdc-icon-button--disabled class.)
* The 'normal' mdc-icon-button doesn't have styling for a disabled button. (This should also be provided based on a disabled selector).
* Just as the mdc-icon-toggle, the toggled variant of the mdc-icon-button allows a child element to be the actual icon, and receive the icon-font classnames or ligatures (to work around issues with pseudo element conflicts between the ripple and the icon font classes). However, most of the documentation for that feature has been stripped, making it a very confusing read. It does still work, so I think the documentation should be fixed.

Furthermore as a separate thought for this component: I don't really understand why the toggled variant works the way it does, with iconOn/iconOff properties that set an icon-font class or ligature. This needs the earlier workarounds for pseudo elements that are difficult to document/grasp. It also excludes the use of SVG icons. Imho a much better approach would be to have an icon-button-toggle DOM with two embedded icons, one marked as the 'on' icon, one as the 'off' icon. This would work with all types of icon-fonts without workarounds for pseudo elements, and would also work with SVG icons. It's also much easier to document for all those different use cases, as the usage would be exactly the same for any kind of icon.

kfranqueiro:
For purposes of clarity, I'm going to lift the final paragraph of this bug to its own issue. I believe we've resolved the documentation issues here.

Thanks @gjdev for all of your feedback!

PR #2957, opened 2018-06-18
fix(icon-button): Remove unused styles and update docs

williamernest:
Address: #2918 
Also, fixes a bug where the foundation was always initializing as 'off'.

### DIFF
--- a/packages/mdc-icon-button/mdc-icon-button.scss
+++ b/packages/mdc-icon-button/mdc-icon-button.scss
@@ -23,10 +23,4 @@
   @include mdc-states;
 }
 
-.mdc-icon-button--disabled {
-  @include mdc-theme-prop(color, text-disabled-on-light);
-
-  pointer-events: none;
-}
-
 // postcss-bem-linter: end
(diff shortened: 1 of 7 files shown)


packages/mdc-icon-button/mdc-icon-button.scss
```diff
@@ -23,10 +23,4 @@
   @include mdc-states;
 }
 
-.mdc-icon-button--disabled {
```
patrickrodee: Should this be replaced with a `:disabled` selector instead of removed completely?

williamernest: The `:disabled` selector is in the mixins file. This was left over from `icon-toggle` and was used when the `icon-toggle` was on an element other than a `button` (`span`, `div`, etc). We dropped support for everything except for `button` and `a` tags, so this is no longer needed at all. Like `mdc-button` you can use `:disabled` for `button` elements to disable them. Anchor tags don't support disabled at all.

patrickrodee: Sweet, thanks for explaining!

PR #2959, opened 2018-06-18
feat(fab): Enable padding customization

### LINKED ISSUE #2900: Extended FAB: Enable padding customizations
After discussing with designers, the baseline extended FAB will consist of the following:

* 12px icon-side padding
* 24px icon
* 12px padding between icon and label
* 20px label-side padding

The baseline for extended FAB with no label will have 20px padding on each side.

We need to be able to customize the following:

* The basis for the icon-side padding and padding between icon and label (both of these should equal each other)
* The label-side padding

Additional thoughts:

* It might be nice if we could bundle icon size and padding adjustment to compute the padding automatically depending on the size, but this might warrant prototyping to see whether "smarter" is actually easier to use
* We'd still need to keep in mind that we don't need to adjust padding for the default collapsed FAB (we simply center the icon, no matter the size)
* We also need to properly style icon-less extended FAB to have equal padding on both sides (using the label-side padding setting)

abhiomkar:
Fixes #2900 

This change enables padding customization for Extended FAB including icon & label side padding. This approach uses Sass variables instead of mixins since the icons side padding needs to be calculated based on the label side padding. It avoids creating additional class name for Extended FAB without icon.


kfranqueiro:
Discussed with @abhiomkar offline. Instead of using global variables, we'll create mixins, to allow for the potential of different customizations for FABs on different views (e.g. a primary/front view vs. secondary/subsidiary views).

I had initially pondered encoding the relationship between icon size and padding all within the icon-size mixin, but Abhinay pointed out that it would probably require significantly more mental overhead for the simple case where someone has redlines they want to apply. So instead, we'll expose mixins allowing direct adjustment of the icon and label paddings.

Otherwise, the approach in this PR is nice, since the negative margin automatically accommodates the intended difference between label-side and icon-side edge padding.

packages/mdc-fab/_mixins.scss
```diff
@@ -56,6 +56,30 @@
...
+  .mdc-fab__label + .mdc-fab__icon {
+    @include mdc-rtl-reflexive-property(
+      margin,
+      $mdc-fab-extended-icon-padding,
+      $mdc-fab-extended-icon-padding - $mdc-fab-extended-label-padding);
+  }
+}
+
+@mixin mdc-fab-extended-label-padding($label-padding) {
+  padding: 0 $mdc-fab-extended-label-padding;
```
kfranqueiro: `$label-padding`

packages/mdc-fab/_mixins.scss
```diff
@@ -56,6 +56,30 @@
...
 }
 
+@mixin mdc-fab-extended-padding($icon-padding, $label-padding) {
+  @include mdc-fab-extended-label-padding($label-padding);
+
+  // Offsets the outer label padding by deducting it from icon padding.
+  .mdc-fab__icon {
+    @include mdc-rtl-reflexive-property(
+      margin,
+      $mdc-fab-extended-icon-padding - $mdc-fab-extended-label-padding,
```
kfranqueiro: The variable names throughout these mixins need to be updated to reference the mixin parameters, otherwise they will never actually customize anything.

We should ideally have an additional screenshot test page under `mdc-fab/mixins` to test these.

packages/mdc-fab/README.md
```diff
@@ -112,6 +112,8 @@ Mixin | Description
 `mdc-fab-container-color($color)` | Sets the container color to the given color
 `mdc-fab-icon-size($width, $height)` | Sets the icon `width`, `height`, and `font-size` properties to the specified `width` and `height`. `$height` is optional and will default to `$width` if omitted. The `font-size` will be set to the provided `$width` value.
 `mdc-fab-ink-color($color)` | Sets the ink color to the given color
+`mdc-fab-extended-padding($icon-padding, $label-padding)` | Sets the padding of icon sides (Left side and gutter space) and label side for Extended FAB. The left & right padding of the icon are equal.
```
kfranqueiro: I'm suggesting some rephrasings for each of these, to try to simplify and clarify when to use each.

"Sets the padding on both sides of the icon, and between the label and the edge of the FAB. In cases where there is no icon, `$label-padding` will apply to both sides."

packages/mdc-fab/README.md
```diff
@@ -112,6 +112,8 @@ Mixin | Description
...
+`mdc-fab-extended-label-padding($label-padding)` | Sets the label side padding for Extended FAB. This padding size will be applied to both sides when Extended FAB is rendered without icon.
```
kfranqueiro: Change the second sentence to "Useful in lieu of `mdc-fab-extended-padding` when styling an Extended FAB with no icon."

packages/mdc-fab/README.md
```diff
@@ -112,8 +112,8 @@ Mixin | Description
 `mdc-fab-container-color($color)` | Sets the container color to the given color
 `mdc-fab-icon-size($width, $height)` | Sets the icon `width`, `height`, and `font-size` properties to the specified `width` and `height`. `$height` is optional and will default to `$width` if omitted. The `font-size` will be set to the provided `$width` value.
 `mdc-fab-ink-color($color)` | Sets the ink color to the given color
-`mdc-fab-extended-padding($icon-padding, $label-padding)` | Sets the padding of icon sides (Left side and gutter space) and label side for Extended FAB. The left & right padding of the icon are equal.
-`mdc-fab-extended-label-padding($label-padding)` | Sets the label side padding for Extended FAB. This padding size will be applied to both sides when Extended FAB is rendered without icon.
+`mdc-fab-extended-padding($icon-padding, $label-padding)` | Sets the padding on both sides of the icon, and between the label and the edge of the FAB. In cases where there is no icon, $label-padding will apply to both sides.
```
kfranqueiro: Nit: backticks around $label-padding

test/screenshot/golden.json
```diff
@@ -125,6 +125,15 @@
       "desktop_windows_ie@11": "https://storage.googleapis.com/mdc-web-screenshot-tests/kfranqueiro/2018/05/25/21_14_58_299/6b0f3e00/mdc-fab/classes/mini.html.win10_ie11.png"
     }
   },
+  "mdc-fab/mixins/mdc-fab-extended-label-adding.html": {
```
kfranqueiro: Typo in filename? adding -> padding?

packages/mdc-fab/package.json
```diff
@@ -18,6 +18,7 @@
     "@material/elevation": "^0.36.1",
     "@material/ripple": "^0.36.0",
     "@material/theme": "^0.35.0",
-    "@material/typography": "^0.35.0"
+    "@material/typography": "^0.35.0",
+    "@material/rtl": "^0.36.0"
```
kfranqueiro: Nit: put rtl after ripple

PR #2965, opened 2018-06-20
fix(chips): Add an event typedef for chip interaction events

rlfriedman:


packages/mdc-chips/chip-set/adapter.js
```diff
@@ -15,7 +15,8 @@
  * limitations under the License.
  */
 
-/* eslint no-unused-vars: [2, {"args": "none"}] */
+/* eslint-disable no-unused-vars */
+import {MDCChipInteractionEventType} from '../chip/foundation';
```
bonniezhou: I think we want to keep the `no-unused-vars: [2, {"args": "none"}]` rule for the rest of the file, so you can do something like this:
```
// eslint-disable-next-line no-unused-vars
import {MDCChipInteractionEventType} from '../chip/foundation';

/* eslint no-unused-vars: [2, {"args": "none"}] */
...
```

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -17,7 +17,8 @@
 
 import MDCFoundation from '@material/base/foundation';
 import MDCChipSetAdapter from './adapter';
-import MDCChipFoundation from '../chip/foundation';
+/* eslint-disable no-unused-vars */
```
bonniezhou: Same here, use `// eslint-disable-next-line no-unused-vars`

packages/mdc-chips/chip-set/adapter.js
rlfriedman: Cool, done!

packages/mdc-chips/chip-set/foundation.js
rlfriedman: Done

PR #2970, opened 2018-06-20
feat(list): Add single selection

williamernest:


package-lock.json
```diff
@@ -6176,6 +6176,7 @@
           "version": "0.0.9",
           "bundled": true,
           "dev": true,
+          "optional": true,
```
patrickrodee: Was this file supposed to change?

packages/mdc-list/README.md
```diff
@@ -212,7 +254,11 @@ Method Signature | Description
 `getListItemCount() => Number` | Returns the total number of list items (elements with `mdc-list-item` class) that are direct children of the `root_` element.
 `getFocusedElementIndex() => Number` | Returns the `index` value of the currently focused element.
 `getListItemIndex(ele: Element) => Number` | Returns the `index` value of the provided `ele` element.
+`setAttributeForElementIndex(ndx: Number, attr: String, value: String) => void` | Sets the `attr` attribute to the value of `value` for the list item at `ndx`.
```
patrickrodee: I'm not in love with the variable name `ndx` but it's already in here so I'm not gonna request a change.

packages/mdc-list/foundation.js
```diff
@@ -63,6 +79,43 @@ class MDCListFoundation extends MDCFoundation {
...
+  setSingleSelection(value) {
+    this.isSingleSelectionList_ = value;
+  }
+
+  /** @param {number} ndx */
+  setSelectedIndex(ndx) {
+    if (ndx === this.selectedIndex_) {
+      this.adapter_.setAttributeForElementIndex(this.selectedIndex_, strings.ARIA_SELECTED, false);
+      this.adapter_.removeClassForElementIndex(this.selectedIndex_, cssClasses.LIST_SELECTED_CLASS);
+      if (this.selectedIndex_ > 0) {
```
patrickrodee: Why does this only set negative tabindex for items 1...n instead of 0...n?

packages/mdc-list/foundation.js
```diff
@@ -63,6 +79,43 @@ class MDCListFoundation extends MDCFoundation {
...
+    }
+
+    if (this.selectedIndex_ >= 0) {
+      this.adapter_.setAttributeForElementIndex(this.selectedIndex_, strings.ARIA_SELECTED, false);
+      this.adapter_.removeClassForElementIndex(this.selectedIndex_, cssClasses.LIST_SELECTED_CLASS);
+      this.adapter_.setAttributeForElementIndex(this.selectedIndex_, 'tabindex', -1);
+    }
+
+    if (ndx >= 0) {
+      if (this.adapter_.getListItemCount() > ndx) {
```
patrickrodee: Collapse these into one if statement

packages/mdc-list/foundation.js
```diff
@@ -120,7 +176,37 @@ class MDCListFoundation extends MDCFoundation {
     } else if (isEnd) {
       this.preventDefaultEvent_(evt);
       this.focusLastElement();
+    } else if (this.isSingleSelectionList_ && (isEnter || isSpace)) {
+      this.preventDefaultEvent_(evt);
+      // Check if the space key was pressed on the list item or a child element.
+      if (evt.target.classList.contains(cssClasses.LIST_ITEM_CLASS)) {
```
patrickrodee: This should be an adapter method (`hasClass` or something similar)

packages/mdc-list/foundation.js
```diff
@@ -120,7 +176,37 @@ class MDCListFoundation extends MDCFoundation {
...
+
+    if (currentIndex === -1) {
+      const listItem = this.getListItem_(evt.target);
+      if (this.adapter_.isElementFocusable(evt.target)) {
+        return;
+      }
+
+      currentIndex = this.adapter_.getListItemIndex(listItem);
+
+      if (currentIndex < 0) {
```
patrickrodee: This if statement can be removed since it always evaluates to true.

packages/mdc-list/index.js
```diff
@@ -89,13 +94,47 @@ export class MDCList extends MDCComponent {
...
+  /** @param {boolean} value */
+  set singleSelection(value) {
+    if (value) {
+      this.root_.addEventListener('click', this.handleClick_);
+    } else {
+      this.root_.removeEventListener('click', this.handleClick_);
+    }
+
+    this.foundation_.setSingleSelection(value);
+    const selectedElement = this.root_.querySelector('.mdc-list-item[aria-selected="true"]');
```
patrickrodee: Could you provide some explanation about why this uses the aria-selected attribute instead of the `mdc-list-item--selected` class?

packages/mdc-list/index.js
```diff
@@ -89,13 +94,47 @@ export class MDCList extends MDCComponent {
...
+    [].slice.call(this.root_.querySelectorAll('.mdc-list-item:not([aria-selected="true"])'))
```
patrickrodee: Could you provide some explanation about why this uses the aria-selected attribute instead of the `mdc-list-item--selected` class?

packages/mdc-list/index.js
```diff
@@ -89,13 +94,47 @@ export class MDCList extends MDCComponent {
...
+      addClassForElementIndex: (ndx, className) => this.listElements_[ndx].classList.add(className),
+      removeClassForElementIndex: (ndx, className) => this.listElements_[ndx].classList.remove(className),
       focusItemAtIndex: (ndx) => this.listElements_[ndx].focus(),
+      isElementFocusable: (ele) => {
+        if (!ele) return false;
+        let matches = Element.prototype.matches;
+        if (!matches) { // IE uses a different name for the same functionality
+          matches = Element.prototype.msMatchesSelector;
+        }
+        return matches.call(ele, strings.FOCUSABLE_CHILD_ELEMENTS);
```
patrickrodee: Clever!!!

packages/mdc-list/README.md
```diff
@@ -135,6 +135,44 @@ OR
...
+```js
+var listEle = document.getElementById('my-list');
+var list = new mdc.list.MDCList(listEle);
+list.singleSelection = true;
+```
+
+#### Pre-selected list item
+
+When rendering the list with a pre-selected list item, the list item that needs to be selected should contain
+the `mdc-list-item--selected` class before creating the list.
```
patrickrodee: Should the list item also have the `aria-selected="true"` attribute?

williamernest: No, it gets added automatically when the list component is created. We add the `--selected` class because we use that as our selector.

packages/mdc-list/index.js
```diff
@@ -89,13 +94,48 @@ export class MDCList extends MDCComponent {
     this.foundation_.setWrapFocus(value);
   }
 
+  /** @param {boolean} value */
+  set singleSelection(value) {
```
patrickrodee: Use a more descriptive variable name than `value` like `isSelected` or something similar.

packages/mdc-list/README.md
patrickrodee: Gotcha. Can you show me the line number where that occurs?

williamernest: https://github.com/material-components/material-components-web/blob/1d5f6106e02a3755ceb9eccaa2a0bf39444d08e8/packages/mdc-list/index.js#L106

patrickrodee: I don't see where `aria-selected="true"` gets added though. Could you point me to that?

williamernest: That's in the foundations `setSelectedIndex`.

williamernest: If the `selectedElement` is found it sets the `selectedIndex` to the `selectedElement` index. The `selectedIndex` proxies to the foundation `setSelectedIndex`, which sets the attributes on the list element. 

https://github.com/material-components/material-components-web/blob/cbbd3948ee793abddb5bdbdaa8208385f5e7addd/packages/mdc-list/index.js#L111

https://github.com/material-components/material-components-web/blob/cbbd3948ee793abddb5bdbdaa8208385f5e7addd/packages/mdc-list/index.js#L117

https://github.com/material-components/material-components-web/blob/cbbd3948ee793abddb5bdbdaa8208385f5e7addd/packages/mdc-list/foundation.js#L115

patrickrodee: Awesome, thanks for the patience and guiding me through. Since that means the `aria-selected` attribute is only set by the MDC component index.js implementation during initialization, let's add a note to include `aria-selected="true"` for the pre-selected list item. That will help users who wrap our components and our users in general.

williamernest: ? `aria-selected` is set by the foundation

williamernest: If you wrap the component and want to use the single selection list, you have to call `setSelectedIndex` on the foundation so that it can initialize the initial selected value. `setSelectedIndex` also sets `aria-selected` on the list element that is selected.

packages/mdc-list/adapter.js
```diff
@@ -41,11 +41,48 @@ class MDCListAdapter {
   /** @param {Element} node */
   getListItemIndex(node) {}
 
+  /**
+   * @param {Number} index
```
patrickrodee: These basic types (number, string, boolean, etc) should be lowercase.

packages/mdc-list/constants.js
```diff
@@ -18,14 +18,16 @@
...
 };
 
 /** @enum {string} */
 const strings = {
   ARIA_ORIENTATION: 'aria-orientation',
   ARIA_ORIENTATION_VERTICAL: 'vertical',
+  ARIA_SELECTED: 'aria-selected',
   FOCUSABLE_CHILD_ELEMENTS: 'button:not(:disabled), a',
-  ITEMS_SELECTOR: '.mdc-list-item',
+  ITEMS_SELECTOR: '.mdc-list-item:not(.mdc-list-item--disabled)',
```
patrickrodee: Why did you add the `:not(.mdc-list-item--disabled)` selector?

packages/mdc-list/index.js
```diff
@@ -80,22 +85,55 @@ export class MDCList extends MDCComponent {
...
 
   /** @param {boolean} value */
   set wrapFocus(value) {
     this.foundation_.setWrapFocus(value);
   }
 
+  /** @param {boolean} isSingleSelectionList */
+  set singleSelection(isSingleSelectionList) {
+    if (isSingleSelectionList) {
+      this.root_.addEventListener('click', this.handleClick_);
```
patrickrodee: This should be a foundation method.

packages/mdc-list/constants.js
williamernest: Disabled items should not be receive focus, so the items selector should ignore them.

packages/mdc-list/index.js
williamernest: ? Aren't we moving toward adding event listeners in the index.js ? See `initialSyncWithDOM` above.

patrickrodee: True, we're exploring it. I guess we can try it here and see how it goes.

Will `handleClick` cause problems if it's triggered by a non-single selection list item? If so, there should be a check inside the `handleClick` method that ensures it won't cause problems if it's called.

williamernest: `handleClick` has logic to determine if you did not click on a list item (it looks for an ancestor `mdc-list-item`) and if doesn't find one it returns early.

trimox:
Hey @williamernest I opened PR #3029 to export `MDCListFoundation`. Would you prefer to add those exports here? If so, I'll close #3029.

packages/mdc-list/README.md
```diff
@@ -200,30 +239,76 @@ Key | Action
...
+`setSelectedIndex()` should be called with the initially selected element immediately after the foundation is 
+instantiated.
+
+```html
+<ul id="my-list" class="mdc-list" aria-orientation="vertical">
+  <li class="mdc-list-item" tabindex="-1">Single-line item<button tabindex="-1"></button></li>
+  <li class="mdc-list-item mdc-list-item--selected" aria-selected="true" tabindex="0">Single-line item</li>
+  <li class="mdc-list-item" tabindex="-1">Single-line item</li>
+</ul>
+```
```
moog16: I think adding the JS here showing to call `setSelectedIndex` will complete the documentation....if I'm reading that correctly.

packages/mdc-list/constants.js
```diff
@@ -18,14 +18,16 @@
 /** @enum {string} */
 const cssClasses = {
   LIST_ITEM_CLASS: 'mdc-list-item',
+  LIST_SELECTED_CLASS: 'mdc-list-item--selected',
```
moog16: LIST_ITEM_SELECTED_CLASS is more consistent with the above class

packages/mdc-list/constants.js
moog16: Changing the key to `ENABLED_ITEMS_SELECTOR` is more semantically correct

packages/mdc-list/foundation.js
```diff
@@ -120,9 +190,26 @@ class MDCListFoundation extends MDCFoundation {
     } else if (isEnd) {
       this.preventDefaultEvent_(evt);
       this.focusLastElement();
+    } else if (this.isSingleSelectionList_ && (isEnter || isSpace)) {
```
moog16: I don't see a demo for this, and this seems like an important thing to demo.

packages/mdc-list/index.js
```diff
@@ -80,27 +85,62 @@ export class MDCList extends MDCComponent {
...
-    return new MDCListFoundation(/** @type {!MDCListAdapter} */{
+    return new MDCListFoundation(/** @type {!MDCListAdapter} */ (Object.assign({
       getListItemCount: () => this.listElements_.length,
       getFocusedElementIndex: () => this.listElements_.indexOf(document.activeElement),
       getListItemIndex: (node) => this.listElements_.indexOf(node),
-      focusItemAtIndex: (ndx) => this.listElements_[ndx].focus(),
+      setAttributeForElementIndex: (index, attr, value) => this.listElements_[index].setAttribute(attr, value),
+      removeAttributeForElementIndex: (index, attr) => this.listElements_[index].removeAttribute(attr),
+      addClassForElementIndex: (index, className) => this.listElements_[index].classList.add(className),
+      removeClassForElementIndex: (index, className) => this.listElements_[index].classList.remove(className),
```
moog16: I checked the foundation and it looks like you have null check code there (eg. this.selectedIndex >= 0), but is there a possibility that `this.listElements_[index]` could be undefined? Just wanna check

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -72,6 +78,43 @@ test('#adapter.getListItemIndex returns the index of the element specified', ()
...
+  assert.isTrue(selectedNode.classList.contains('foo'));
+  document.body.removeChild(root);
+});
+
+test('#adapter.removeClassForElementIndex removes the class from the list element at index specified', () => {
+  const {root, component} = setupTest();
+  document.body.appendChild(root);
+  const selectedNode = root.querySelectorAll('.mdc-list-item')[1];
+  component.getDefaultFoundation().adapter_.removeClassForElementIndex(1, 'foo');
+  assert.isFalse(selectedNode.classList.contains('foo'));
```
moog16: when does the element get the foo class? I ran this and it looks like if you remove line 113 it still passes

test/unit/mdc-list/foundation.test.js
```diff
@@ -383,7 +392,197 @@ test('#handleKeydown does not find ancestor with mdc-list-item so returns early'
...
+  const event = {key: 'Space', target, preventDefault};
+
+  td.when(mockAdapter.getFocusedElementIndex()).thenReturn(0);
+  td.when(mockAdapter.getListItemCount()).thenReturn(3);
+  td.when(mockAdapter.isListItem(target)).thenReturn(true);
+  foundation.setSingleSelection(true);
+  foundation.handleKeydown(event);
+
+  td.verify(preventDefault(), {times: 1});
+  td.verify(mockAdapter.setAttributeForElementIndex(0, strings.ARIA_SELECTED, true));
```
moog16: do you want a {times: 1} here?

test/unit/mdc-list/foundation.test.js
```diff
@@ -383,7 +392,197 @@ test('#handleKeydown does not find ancestor with mdc-list-item so returns early'
...
+
+  td.when(mockAdapter.getFocusedElementIndex()).thenReturn(0);
+  td.when(mockAdapter.getListItemCount()).thenReturn(3);
+  td.when(mockAdapter.isListItem(target)).thenReturn(true);
+  foundation.setSingleSelection(true);
+  foundation.handleKeydown(event);
+  foundation.handleKeydown(event);
+
+  td.verify(preventDefault(), {times: 2});
+  td.verify(mockAdapter.removeAttributeForElementIndex(0, strings.ARIA_SELECTED));
```
moog16: here too {times: 1}. Few places below as well.

test/unit/mdc-list/foundation.test.js
```diff
@@ -383,7 +392,197 @@ test('#handleKeydown does not find ancestor with mdc-list-item so returns early'
...
+  td.verify(preventDefault(), {times: 2});
+  td.verify(mockAdapter.setAttributeForElementIndex(0, 'tabindex', 0));
+  td.verify(mockAdapter.setAttributeForElementIndex(1, 'tabindex', -1));
+});
+
+test('#handleKeydown space key is triggered and focused is moved to a different element', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const preventDefault = td.func('preventDefault');
+  const target = {classList: ['mdc-list-item']};
+  const event = {key: 'Space', target, preventDefault};
```
moog16: You don't have any tests events with keyCodes for space and enter. Will that increase coverage?

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -115,6 +190,33 @@ test('wrapFocus calls setWrapFocus on foundation', () => {
...
+  td.verify(mockFoundation.handleClick(td.matchers.anything()), {times: 0});
+});
+
+test('singleSelection calls foundation setSingleSelection with the provided value', () => {
+  const {component, mockFoundation} = setupTest();
+  component.singleSelection = true;
+  td.verify(mockFoundation.setSingleSelection(true), {times: 1});
+});
+
+test('selectedIndex calls setSelectedIndex on foundation', () => {
```
moog16: setSelectedIndex is a pretty complex function, but only has this one test. Might want to add more

test/unit/mdc-list/foundation.test.js
williamernest: `{times: 1}` is the default behavior.

packages/mdc-list/README.md
```diff
@@ -200,30 +239,76 @@ Key | Action
...
+
+The `MDCListFoundation` expects the HTML to be setup a certain way before being used. This setup is a part of the `layout()` and `singleSelection()` functions within the `index.js`.
+
+#### Setup in `layout()`
+
+The default component requires that every list item receives a `tabindex` value so that it can receive focus
+(`li` elements cannot receive focus at all without a `tabindex` value). Any element not already containing a
+`tabindex` attribute will receive `tabindex=-1`. The first list item should have `tabindex="0"` so that it the
+user can find the first element using the `tab` key, but subsequent `tab` keys strokes will cause focus to
+skip over the entire list. If the list items contain sub-elements that are focusable (`button` or `a` elements),
```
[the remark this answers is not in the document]
moog16: oh sorry - i meant to remove this comment.

test/unit/mdc-list/foundation.test.js
kfranqueiro: I'm pretty sure that is not correct. The default behavior is any non-zero number of invocations is valid.

test/unit/mdc-list/foundation.test.js
williamernest: It doesn't doesn't increase coverage since that statement is already covered by the `key` parameter.

test/unit/mdc-list/mdc-list.test.js
williamernest: `setSelectedIndex` has nearly 100% coverage through the `click`/`key` handler tests.

packages/mdc-list/README.md
williamernest: The documentation shows the JS in above examples. This example is discussing how the html should be rendered when creating your own component.

test/unit/mdc-list/foundation.test.js
```diff
@@ -383,7 +419,230 @@ test('#handleKeydown does not find ancestor with mdc-list-item so returns early'
...
+
+  td.when(mockAdapter.getFocusedElementIndex()).thenReturn(0);
+  td.when(mockAdapter.getListItemCount()).thenReturn(3);
+  foundation.setSingleSelection(true);
+  foundation.handleKeydown(event);
+
+  td.verify(preventDefault(), {times: 1});
+});
+
+test('#handleKeydown enter key causes preventDefault to be called on the event', () => {
```
moog16: I think you have the titles of this and the next tests reversed with regards to `if singleSelection=true`

test/unit/mdc-list/foundation.test.js
```diff
@@ -383,7 +419,230 @@ test('#handleKeydown does not find ancestor with mdc-list-item so returns early'
...
+  const event = {key: 'Space', target, preventDefault};
+
+  td.when(mockAdapter.getFocusedElementIndex()).thenReturn(0);
+  td.when(mockAdapter.getListItemCount()).thenReturn(3);
+  foundation.handleKeydown(event);
+
+  td.verify(preventDefault(), {times: 0});
+});
+
+test('#handleKeydown enter key does not causes preventDefault to be called if singleSelection=true', () => {
```
moog16: same here - titles reversed

PR #2974, opened 2018-06-21
Adding prefix for appearance

arpontes:
In Chrome, the propery appearance is not working.
The result is a select displaying two arrows: 
image

It's necessary to add -webkit- to the property.

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -75,7 +75,9 @@
     background-color: transparent;
     white-space: nowrap;
     cursor: pointer;
-    appearance: none;
+    -webkit-appearance: none;
+       -moz-appearance: none;
+            appearance: none;
   }
 
   // stylelint-disable-next-line plugin/selector-bem-pattern
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -76,8 +76,8 @@
     white-space: nowrap;
     cursor: pointer;
     -webkit-appearance: none;
-       -moz-appearance: none;
-            appearance: none;
+    -moz-appearance: none;
+    appearance: none;
   }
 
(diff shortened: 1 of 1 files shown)


arpontes:
I signed it!

arpontes:
The continuous integration process has failed, but it is non related to my change. It is something with the validator.

21 06 2018 14:29:30.729:ERROR [launcher]: Cannot start Chrome
	[2855:2855:0621/142930.560972:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/google/chrome/chrome-sandbox is owned by root and has mode 4755.

What should I do?

kfranqueiro:
Thanks for filing a PR with us!

These vendor prefixes are already applied via autoprefixer, in [our build process](https://github.com/material-components/material-components-web/blob/v0.36.1/scripts/webpack/css-bundle-factory.js#L195-L199) as well as documented in the [getting started guide](https://github.com/material-components/material-components-web/blob/v0.36.1/docs/getting-started.md#step-2-include-css-for-a-component). As such, I'm going to close this PR.

RE the CI failure, this is currently expected due to private environment variables not being exposed to forks.

arpontes:
@kfranqueiro, thanks for your answer.
In my case, I'm compiling sass on runtime (the colors and some other configurations may be changed in a user-level granularity, so I let all the .scss files in a folder and execute dart-sass when this change occurs, and keep the outcome in memory while the user sessions is alive). Because of this requirement, I'm not using webpack or any other compile-time process.

Is there any tip you could gave of how to handle this situation? I was expecting to avoid including an ad hoc css to resolve this case (by the way, I only made this change after seeing that some components, like Ripple or Slider, have vendor-prefixed properties).

PR #2984, opened 2018-06-22
chore(infrastructure): Remove unnecessary CLI options

acdvorak:
Removes 3 CLI options and moves them to `constants.js`:

- `--test-dir`
- `--golden-path`
- `--gcs-bucket`

Nobody will ever need to override these values.

Having a CLI flag for a constant is confusing because it implies that users can/should override it, but it ends up breaking the tests in non-obvious ways.


test/screenshot/lib/cli-arg-parser.js
```diff
@@ -19,7 +19,11 @@
 const GitRepo = require('./git-repo');
 const argparse = require('argparse');
 const fs = require('mz/fs');
-const path = require('path');
+
```
patrickrodee: Nit: Make these all inline to follow the style of the other imports.

test/screenshot/lib/cli-arg-parser.js
```diff
@@ -195,29 +163,25 @@ The default behavior is to always build assets before running the tests.
   }
 
   initCleanCommand_() {
+    // eslint-disable-next-line no-unused-vars
```
patrickrodee: Why is this necessary?

test/screenshot/lib/cloud-storage.js
```diff
@@ -16,13 +16,16 @@
...
 
-const CliArgParser = require('./cli-arg-parser');
 const Duration = require('./duration');
 const GitRepo = require('./git-repo');
 const GoogleCloudStorage = require('@google-cloud/storage');
 const LocalStorage = require('./local-storage');
 const ProcessManager = require('../lib/process-manager');
-const {ExitCode} = require('../lib/constants');
+
+const {
```
patrickrodee: Nit: Inline these imports like the above comment.

test/screenshot/lib/cli-arg-parser.js
acdvorak: Done

test/screenshot/lib/cloud-storage.js
acdvorak: Done

test/screenshot/lib/cli-arg-parser.js
acdvorak: It's not. Removed!

acdvorak:
I don't think the README needs to be updated, because it doesn't mention any of the CLI options I removed 😀

PR #3024, opened 2018-07-03
fix: hot-patching closure annotations.

aprigogin:


kfranqueiro:
Aren't there still many other foundations that lack `@override` comments? What is this fixing specifically?

PR #3047, opened 2018-07-06
feat(typography): make it compatible with upcoming sass 4

grzuy:
The following is the deprecation warning you get by running typography
with SASS 3.5.6:

```
$ sass -v
Ruby Sass 3.5.6
$ sass --check packages/mdc-typography/mdc-typography.scss
DEPRECATION WARNING on line 57 of packages/mdc-typography/_functions.scss:
To preserve the current behavior, use quotes:

  unquote("mdc-typography-styles-#{$style}")
```

### DIFF
--- a/packages/mdc-typography/_functions.scss
+++ b/packages/mdc-typography/_functions.scss
@@ -54,7 +54,7 @@
     $style-props: map-merge($base-styles, $style-props);
 
     // Merge global overrides onto each style.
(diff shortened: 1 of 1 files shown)


williamernest:
For reference: 
https://github.com/sass/ruby-sass/blob/4.0.0.alpha.1/doc-src/SASS_CHANGELOG.md#backwards-incompatibilities----must-read

PR #3069, opened 2018-07-13
fix(text-field): Adjust the baseline of text field's helper text to match spec

### LINKED ISSUE #2783: mdc-text-field's helper text is too far from text field
The spec shows 16px from bottom of text field to baseline of helper text:
https://material.io/design/components/text-fields.html#spec
image

But our `mdc-text-field` looks like this:
image

abhiomkar:
Fixes #2783 

*BREAKING CHANGE*

Changes included:

- Text field's helper text block element uses typography baseline mixin to set baseline height.
- Removed margin bottom from both text field root and helper text.
- Updated the text field demos.
- Changes to typography baseline mixin to use flexbox to avoid extra white-space on newline.


demos/text-field.html
```diff
@@ -80,8 +80,8 @@ <h2>Full Functionality JS Component (Floating Label, Validation)</h2>
             <div class="mdc-line-ripple"></div>
           </div>
           <p id="my-text-field-helper-text" class="mdc-text-field-helper-text"
-             aria-hidden="true" style="display:none;">
-            Helper Text (possibly validation message)
+             aria-hidden="true" style="display:none;"
+             >Helper Text (possibly validation message)
```
kfranqueiro: Why is the helper text sensitive to leading whitespace now? AFAICT it was not previously?

If this is a limitation of the new baseline styles/mixin, I'd like to see us work it out if at all possible...

packages/mdc-textfield/helper-text/README.md
```diff
@@ -57,9 +57,7 @@ the input element.
   <label for="username" class="mdc-floating-label">Username</label>
   <div class="mdc-line-ripple"></div>
 </div>
-<p id="username-helper-text" class="mdc-text-field-helper-text" aria-hidden="true">
-  This will be displayed on your public profile
-</p>
+<p id="username-helper-text" class="mdc-text-field-helper-text" aria-hidden="true">This will be displayed on your public profile</p>
```
kfranqueiro: I guess this formatting change can be reverted?

Issue #3038, opened 2018-07-06
MDCList: Update typography of two line list items to use baseline alignment


williamernest:
Modify the 2 line list item to support the baseline text alignment using `::before` and `::after` elements. The mixins should be added into the typography package.

PR #3083, opened 2018-07-16
feat(typography): New mixin to set exact baseline height of text elements.

abhiomkar:
Feat required for #2783 & #3038


packages/mdc-typography/_mixins.scss
```diff
@@ -40,3 +40,35 @@
...
+
+  &::before {
+    display: inline-block;
+    width: 0;
+    height: $size;
+    content: "";
+    vertical-align: 0;
+  }
+
+  @if $trailing {
```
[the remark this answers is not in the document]
lynnjepsen: Okay I looked at #3085 and I think i understand (somewhat) why this mixin's logic is more complicated than I expected it to be.

I think that mdc-list has a strange HTML structure for supporting two line lists. I think if we change that HTML markup (so that it is not mixing text elements with span elements), then we can more simply implement the baseline alignment fix.

packages/mdc-typography/_mixins.scss
```diff
@@ -40,3 +40,28 @@
   white-space: nowrap;
   overflow: hidden;
 }
+
+@mixin mdc-typography-baseline-top($distance) {
+  margin-top: 0;
+  line-height: normal;
```
lynnjepsen: I think setting line-height to normal will become unnecessary once we remove line-height from typography scale.

But for now I think it is okay.

abhiomkar: I would keep this for now. Secondary text in list and hint text on text field still has line-heights.

PR #3084, opened 2018-07-16
feat(text-field): New API to enable/disable native input validation for custom validity

### LINKED ISSUE #2988: Text field error state does not change as soon as input becomes valid
0.36.1

Chrome 66

### What OS are you using?
OS X

When a text field contains invalid input, the error state does not change until blur.

e.g. I have a text input with a pattern and maxlength.  When I enter invalid input and press tab then the lable and line change to red.

I expected the label and line to change to red as soon as I typed an invalid input.  The text field documentation says that it uses HTML5 :invalid to provide validity styling.

https://material.io/develop/web/components/input-controls/text-field/

Perhaps it should use the :user-invalid HTML5 pseudo class to show red when I change the input to be invalid.

abhiomkar:
BREAKING CHANGE: Setting the validity state using `setValid` no longer ignores native input validation. New API `useNativeValidation` to enable / disable native validation for custom validity.

Fixes #2988


kfranqueiro:
My other concern with this is that `setError` is easily confusable with the existing `setValid` API, which actually disables native validity checking when you first call it, and from that point onward you must call it any time you want to update the text field's validity state.

So I have a few questions/proposals for this:

* Is the existing `setValid` already useful for solving the problems we're trying to solve with `setError`? (Maybe not, since that forcibly disables native validity checking)
* If not, could we salvage the existing `setValid` API to serve this purpose, e.g. by making it possible to keep native validity checking on?
* Otherwise, can we better distinguish this API from that one since currently it sounds like a polar opposite, but isn't? Maybe even just directly expose it as `styleValidity` on the component as well, since that may make it more obvious that it's just a one-time style change? (This would also have the advantage of keeping the boolean directionality the same as the other APIs we have so far, rather than flipping them such that error is now `true`.)

These aren't necessarily in order of preference; I'm okay with option 3 if we can find a good way to name it to clarify how it's different.

abhiomkar:
That's correct. Setting validity using `setValid` API ignores the native input validations. I like the idea of making changes to `setValid` API to bring back the native input validations when input value is changed. Not sure if there are any other reasons for current implementation to enforce having manual validation.

My rationale of creating a new `setError` API is it encapsulates the internal style updates that is done by `styleValidity()`. We can also improve `setError` API in future by making it accept string to show as error text. Example: `setError('Username ${textField.value} is already taken.')`

WDYT?

kfranqueiro:
When we discussed this offline yesterday, the thought was that changing `setValid` to no longer ignore native HTML5 validation is acceptable since in such a case where you want to ignore native validation, you would simply not set HTML5 validation attributes to begin with; in addition, it removes an irreversible and possibly surprising behavior of the current API.

Looping in @Teoxoy and @DBartsch from #1018 / #1092 which added `setValid`, in case they want to chime in on this thought.

My concern remains that as this PR stands right now, we are exposing two APIs which look like polar opposites but aren't, which is hard to understand (and justify).

I wouldn't want to conflate setting unique error text right now, since that would presumably need to also store helper/error text to be able to potentially flip between the two, and based on comments we've seen from our designers, I'm not sure whether that is a feature we'd want to explicitly encode in MDC Web anyway. The `helperTextContent` setter API already offers a way to do this manually if consumers really want to do it.

abhiomkar:
We may also have a `pattern` attribute along with asynchronous errors which requires custom validation. For example, the username must not contain special characters and should be unique.

`set valid()` and `setError` does different things - they're not the same. `setValidity` is API from foundation not component API. Changing the API name may solve the problem. Maybe `set errorState`

Teoxoy:
IMO both APIs try to accomplish the same thing (setting the validity manually) and this PR doesn't resolve #2988.

Using `setValid` alows you to set the validity state internally (in the `isValid_` variable) while `setError` would only apply a class to the element. I belive that having the validity state internally is the way to go because it can be used for other checks in the foundation class (ex. it's used for `shouldShake`).

A fix for #2988 would be to register an event handler for the `change` event via `registerTextFieldInteractionHandler` and use `setValid` to set the validity. That is in fact why I added the functionality; I had the same usecase.

Regarding this:

> When we discussed this offline yesterday, the thought was that changing setValid to no longer ignore native HTML5 validation is acceptable since in such a case where you want to ignore native validation, you would simply not set HTML5 validation attributes to begin with; in addition, it removes an irreversible and possibly surprising behavior of the current API.

That would not work because if the user inputs something that is not valid and I set it via `setValid`, on blur it will be valid because if you dont specify any HTML5 validation attributes `input.validity.valid` will return true. That is why I had to block the native validation.

kfranqueiro:
Thanks for pointing that out about your use case of avoiding native validation clearing the error state.

This PR doesn't solve #2988 OOTB by design, but its goal is to expose an API allowing that problem to be solved via an `input` event handler.

#2988 still calls for native HTML validation, and can't be solved via a `change` event handler because the problem is they want to clear the invalid state once the input becomes valid, which may happen before the field loses focus and `change` is fired.

How about this alternative: Could we decouple the `useCustomValidityChecking_` flag from `setValid` and expose it separately?

* By default `setValid` would serve the purpose of `setError` in this PR (just inverted)
* We would expose a component flag e.g. `useNativeValidation`, defaulting to `true`, which you could set to `false` to ignore native validation (like `setValid` currently does implicitly and irreversibly)
* `set useNativeValidation` on the component would proxy to a `setUseNativeValidation` foundation method to set a flag which would be checked in `isValid()`

This proposed alternative would mean that in order to get the current behavior of `setValid`, you would first need to set `useNativeValidation` to `false` on the text field component instance. IMO this makes the consequences clearer, and opens up the ability to satisfy both #2988 and #1018.

Teoxoy:
> #2988 still calls for native HTML validation, and can't be solved via a change event handler because the problem is they want to clear the invalid state once the input becomes valid, which may happen before the field loses focus and change is fired.

Yes, I was thinking about the `input` not `change` event, I wrote wrong.

The alternative sounds good.

Teoxoy:
But even if the alternative is implemented, I don't see how it would fix #2988. How would the fix look like if the alternative would get implemented?

abhiomkar:
@Teoxoy You can look at the updated demo in this PR where the new API is being called on `input` which removes the error state. With the proposed implementation it'll be `setValid(true)` instead of `setError`.

PR #3085, opened 2018-07-16
feat(list): Updated two-line list to use typography baseline to match spec.

### LINKED ISSUE #3038: MDCList: Update typography of two line list items to use baseline alignment

abhiomkar:
BREAKING CHANGE

Fixes #3038


packages/mdc-list/mdc-list.scss
```diff
@@ -90,23 +90,44 @@
...
   @include mdc-typography-overflow-ellipsis;
 
+  align-self: flex-start;
+}
+
+$mdc-list-item-primary-text-baseline-height_: 32px;
+$mdc-list-item-secondary-text-baseline-height_: 20px;
+
+$mdc-list-dense-item-primary-text-baseline-height_: 24px;
+$mdc-list-dense-item-secondary-text-baseline-height_: 20px;
```
lynnjepsen: if the dense version is the same as the regular version...then you probably dont need it?

PR #3108, opened 2018-07-17
feat(tab): Update tab color and typography

### LINKED ISSUE #3058: Update new tabs dependencies
>Probably want to update these, but maybe that should be done across tab packages once they're all ready

https://github.com/material-components/material-components-web/pull/2577#discussion_r197538305

patrickrodee:
Update the use of color and typography in MDCTab to use the most-recent mixins. Add variables for icon size, tab height, text label opacity, and icon opacity. Closes #3058

### DIFF
--- a/packages/mdc-tab/mdc-tab.scss
+++ b/packages/mdc-tab/mdc-tab.scss
@@ -73,14 +73,14 @@
   display: inline-block;
   padding: 16px 0 20px;
   line-height: 1;
-  opacity: 0.6;
+  opacity: $mdc-tab-text-label-opacity;
 }
 
 .mdc-tab__icon {
   width: $mdc-tab-icon-size;
   height: $mdc-tab-icon-size;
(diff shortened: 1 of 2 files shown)


packages/mdc-tab/mdc-tab.scss
```diff
@@ -69,22 +70,21 @@
...
 
 .mdc-tab__text-label {
-  @include mdc-typography-base;
-
   display: inline-block;
   padding: 16px 0 20px;
-  font-size: 14px;
-  font-weight: 500;
+  opacity: $mdc-tab-text-label-opacity;
+  line-height: 1;
```
kfranqueiro: Is this overriding `mdc-typography(button)` on `.mdc-tab` above? If so, should we just put this there with a comment to make it clearer?

patrickrodee: Sounds good

PR #3119, opened 2018-07-17
fix(switch): Fix switch toggling on IE and Edge

### LINKED ISSUE #3111: New Switch: IE and Edge don't toggle switch when interacting with thumb
This is reproducible on IE 11 or Edge on the `feat/switch-update` branch.

1. Run `npm run dev`
2. Load http://localhost:8080/switch.html in IE or Edge
3. Attempt to click (or tap) directly on the thumb
   * Nothing happens

Other observations: 
* Clicking/tapping anywhere else on the switch's hit target works
* Notice that the mouse has the default cursor when hovered over the thumb (which doesn't work), but the pointer cursor when hovered over any other part of the switch's hit target (which works)

rlfriedman:
Fixes #3111, tested in Chrome, Firefox, IE, and Edge. The issue seems to have come from some sort of interaction between the display flex on the parent of the thumb and the rest of the styling (for IE and Edge).

### DIFF
--- a/packages/mdc-switch/mdc-switch.scss
+++ b/packages/mdc-switch/mdc-switch.scss
@@ -45,6 +45,7 @@
   margin: 0;
   opacity: 0;
   cursor: pointer;
+  pointer-events: auto;
 }
 
 .mdc-switch__track {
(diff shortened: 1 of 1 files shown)


kfranqueiro:
Oh, because ripple typically acts as an overlay. Did design specifically indicate that the ripple is supposed to appear in back of the thumb for switch?

I'd check with them first to make sure we're doing the right thing, but otherwise I'll approve this if that's correct.

rlfriedman:
Yeah, I confirmed with design that for switch the ripple should be under the thumb.

PR #3120, opened 2018-07-17
chore: water feat/navigation-drawer branch w/ master

moog16:


.travis.yml
```diff
@@ -1,32 +1,41 @@
 language: node_js
 branches:
+  # Only run Travis on:
+  #   A) Commits made directly to the `master` branch (i.e., merged PRs); and
+  #   B) PRs that will eventually be merged into `master`.
+  # This prevents excessive resource usage and CI slowness.
```
kfranqueiro: The lines under `only:` should match the indentation of `only:` to match master... but we'll probably end up checking out master's .travis.yml to merge this branch at the end anyway, so your call whether to do that here

PR #3131, opened 2018-07-18
chore(infrastructure): Reduce noisy Travis job log output

acdvorak:
### What it does

A few recent Travis jobs have had their job logs truncated for being too long. Let's fix that.

- Reduces Travis job log output from ~8k lines to ~2k lines
    - Suppresses noisy log output from `install` step
        - Only runs `npm install` (but not `npm ls`)
        - Travis runs `npm ls` by default, which produces about 4k log lines. This makes the log output harder to read and slower to load in the UI, and it doesn't seem terribly useful.
    - Suppresses noisy log output from gcloud installer's `tar` command
        - Makes the log output easier to read and faster to load in the Travis job log UI
- Colorizes log output for "parallel execution limit reached" warning messages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,11 @@ cache:
   directories:
   - $HOME/.npm
   - $HOME/google-cloud-sdk
-  - node_modules
 before_install:
   # Source the script to run it in the same shell process. This ensures that any environment variables set by the
   # script are visible to subsequent Travis CLI commands.
   # https://superuser.com/a/176788/62792
   - source test/screenshot/infra/commands/travis.sh
+install:
+  - npm ci # Deletes `node_modules` before installing packages
+  #- npm ls # Noisy output, but useful for debugging npm package dependency version issues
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -33,6 +33,11 @@ matrix:
     env:
     - TEST_SUITE=screenshot
     script: npm run screenshot:test -- --no-fetch
+cache:
+  directories:
+  - $HOME/.npm
+  - $HOME/google-cloud-sdk
+  - node_modules
```
kfranqueiro: What effect would this have on package updates?

test/screenshot/infra/commands/travis.sh
```diff
@@ -13,47 +13,53 @@ function exit_if_external_pr() {
   fi
 }
 
+function print_travis_env_vars() {
+  echo
+  env | grep TRAVIS
```
kfranqueiro: This isn't going to output anything that should be hidden, right?

acdvorak: I've checked the output many times and have never seen anything sensitive.

That said, I'm still planning to remove this once the Travis config is stable.

PR #3137, opened 2018-07-18
chore(text-field): Add baseline screenshot test page

kfranqueiro:
Does not include textarea, full-width or icon variants, or mixin use.

Intentionally omits the current default since it is being deprecated.

Original report that generated the new goldens: https://storage.googleapis.com/mdc-web-screenshot-tests/kfranqueiro/2018/07/18/20_19_24_759/report/report.html


test/screenshot/spec/mdc-textfield/fixture.js
```diff
@@ -0,0 +1,3 @@
+[].forEach.call(document.querySelectorAll('.mdc-text-field'), function(el) {
+  const textField = new mdc.textField.MDCTextField(el);
```
acdvorak: Unused variable triggers a linter warning

PR #3143, opened 2018-07-19
fix(switch): Fix showing switch ripple focus state in Edge

rlfriedman:
Realized that having '$has-nested-focusable-element' in mdc-states for switch was causing the focus state to not show up properly in Edge. 

In addition, since we're no longer supporting CSS only switch with the new version that requires JS for styling, it's not necessary to have the focus-within selector for ripple anyway and should be safe to remove.

Tested on Edge, Chrome, and Firefox.


rlfriedman:
My thought was that Edge was unhappy with the focus-within selector and dropping other related styling. We have seen a similar issue with with using the placeholder pseudo element and styles being dropped.

Issue #3066, opened 2018-07-13
TextField does not respect CSS var --mdc-theme-primary


kennylevinsen:
TextField no longer respects --mdc-theme-primary.

Specifically, `.mdc-text-field--focused:not(.mdc-text-field--disabled) .mdc-floating-label` no longer references the CSS variable, and instead use the default purple primary color directly.

### What MDC Web Version are you using?

Bug first occurred in 0.35.0, and persists in 0.37.0.

Last working version was 0.34.1

### What browser(s) is this bug affecting?

Tested in Chrome

### What OS are you using?

Fedora 28

### What are the steps to reproduce the bug?

0.34.1 (working) codepen: https://codepen.io/anon/pen/oMjBPq
0.35.0 (not working) codepen: https://codepen.io/anon/pen/rrOjbm
0.37.0 (not working) codepen: https://codepen.io/anon/pen/vaNgMQ

### What is the expected behavior?

That the primary color is observed like in 0.34.1.

### What is the actual behavior?

The default primary purple color is observed from 0.35.0 and onwards.

### Any other information you believe would be useful?

444f14fc8aa70ebdd0e4a4d5b6d6d1779fe31ad3 looks suspicious, but I have not bisected my way between the releases.

kennylevinsen:
This is likely due to not being able to apply the .87 alpha channel to CSS var that already contains a color.

A solution to this could be to specify the color in a different way:

```css
:root {
    --mdc-theme-primary: 240, 240, 240; /* arbitrary color */
}

.something {
    color: rgba(var(--mdc-theme-primary), .87);
}
```

TANJX:
I have the same issue too; I overrode the original property

abhiomkar:
@kennylevinsen Thanks for filing this issue.

`mdc-theme-prop` mixin currently only supports colors with no alpha value on theme property values (primary, secondary, on-surface etc). But, Text field component has alpha values mixed with the baseline colors which passed onto `mdc-theme-prop` that renders only color style directly without using CSS variables.

Added this to our tracker to make changes to out theme component to support using CSS variables.

kvdheuvel:
This bug is still visible in 0.41.0.

We fixed this locally by setting the color using the mdc theme props:
`.mdc-text-field--focused {
	&:not(.mdc-text-field--disabled) {
		.mdc-floating-label {
			color: $mdc-theme-primary;
			color: var(--mdc-theme-primary, $mdc-theme-primary);
		}
	}
}`

[...]

PR #3145, opened 2018-07-19
fix(theme): Example of runtime TextField theming fix

jamesmfriedman:
This PR is intended to be more of an example than anything, but if the contributors are ok with this approach, I'd be happy to go ahead and apply this to the whole codebase.

## The issue
image

This is an example of setting red as the `mdc-theme-primary` color at runtime, and the resulting `mdc-floating-label` still using the default purple theme color.

## The short version
Anytime the `mdc-theme-prop-value` function is used in SASS, the value is being hardcoded into the CSS which means you can never override it with the runtime css variables. The issue is compounded by using the `rgba` function to add opacity to the color.

## Proposed fixes / Ideas
- Don't use that function
- Handle dynamic opacity using standard opacity OR define the different percent color stops as runtime CSS Vars. Both have their trade offs.
- A slightly crazier idea would be to just define R, G, and B as numeric variables. This would allow you to do something like this during runtime. This still results in a simple mdc-theme-primary variable consumers can use, and allow for dynamic opacity.
image

## Why don't you just recompile the SASS?
I have apps that use themes in a variety of ways
- One app has a light and a dark mode in the UI. Using RMWC ThemeProvider I just set the new variables and everything works like magic
- I have another app where the theme is applied dynamically based on their brand colors.
- The runtime theming stuff is too magical to pass up. I'd love to just get it working across the board as expected.

### DIFF
--- a/packages/mdc-floating-label/_mixins.scss
+++ b/packages/mdc-floating-label/_mixins.scss
@@ -17,8 +17,9 @@
 @import "@material/rtl/mixins";
 @import "./variables";
 
-@mixin mdc-floating-label-ink-color($color) {
+@mixin mdc-floating-label-ink-color($color, $opacity: 1) {
   @include mdc-theme-prop(color, $color);
+  opacity: $opacity;
 }
 
 // Used for textarea in case of scrolling
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -564,7 +564,7 @@
 
 @mixin mdc-text-field-label-ink-color_($color) {
   .mdc-floating-label {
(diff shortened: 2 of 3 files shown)


kfranqueiro:
Cross-linking #3066 which is somewhat related

jamesmfriedman:
I'm going to go ahead and close this out since it was never intended to be merged and was more of an idea exploration.

With that said, the proposal above seems like it would be a marked improvement to the theming framework and allow you to dynamically mix colors at runtime instead of relying on SASS to hardcode them in.

Personally, I have CSS overrides in every single project to correct the above mentioned issues (like textfield label). It's not ideal, but it works and is maintainable.

Issue #2813, opened 2018-05-24
Remove [de]registerEventHandler methods from adapters


patrickrodee:
Our adapters often have `registerEventHandler`/`deregisterEventHandler` methods that are used to add and remove event listeners from DOM nodes. Many frameworks (React, Angular, Vue, etc) prefer events to be bound directly on the DOM elements. This has the benefit of not requiring references to the DOM in the component lifecycle and ensuring that the event handlers are appropriately cleaned up on component destruction.

For many of those frameworks, the `registerEventHandler`/`deregisterEventHandler` adapter methods become no-ops.

We should remove the `registerEventHandler`/`deregisterEventHandler` methods from our adapters and instead expose information about which handlers should be bound to which events. Registering the event handlers will then become the responsibility of those who wrap our components.

Packages that may require updates, as of 5/29/18:

* [x] checkbox
* [x] chips
* [x] dialog
* [x] drawer
* [ ] floating-label
* [ ] form-field
* [ ] grid-list
* [x] icon-button (which replaces icon-toggle)
* [ ] line-ripple
* [x] menu
* [ ] radio
* [ ] ripple
* [x] select
* [ ] slider
* [x] snackbar
* [ ] top-app-bar (which replaces toolbar)
* [x] tab
* [x] tab-indicator
* [ ] text-field and text-field icon

kfranqueiro:
You say "expose information" which I assume means documentation... but I assume we also need to move any existing registrations out of the foundation and into our vanilla components, right?

patrickrodee:
That's correct

kfranqueiro:
Tracking this as one issue for now, but we will undoubtedly want to PR this one package at a time.

aluminick:
Hi, is adapter and foundation the same thing? And can you give me a sample code on integrating MDC with React using adapter/foundation?

kfranqueiro:
Foundation, adapter, and component are briefly described here: https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md#javascript

We have a codelab that walks through wrapping a component for React: https://codelabs.developers.google.com/codelabs/mdc-112-web/index.html

aluminick:
Sir @kfranqueiro regarding the codelab, how to know if a method uses a pub/sub pattern?

moog16:
@abhiomkar I'm going to fix this for top app bar to fix https://github.com/material-components/material-components-web/issues/4527.

PR #3146, opened 2018-07-19
refactor(chips): Register handlers in component instead of foundation

bonniezhou:
Part of #2813


bonniezhou:
Whoops, nice catch @kfranqueiro !

packages/mdc-chips/chip/index.js
```diff
@@ -69,8 +79,35 @@ class MDCChip extends MDCComponent {
     }
   }
 
+  initialSyncWithDOM() {
+    this.handleInteraction_ = (evt) => this.foundation_.handleInteraction(evt);
+    this.handleTransitionEnd_ = (evt) => this.foundation_.handleTransitionEnd(evt);
+    this.handleTrailingIconInteraction_ = (evt) => this.foundation_.handleTrailingIconInteraction(evt);
+
+    ['click', 'keydown'].forEach((evtType) => {
```
moog16: I think making this array and the one below into variables will read better, and then you can share it in the destroy.

bonniezhou: Done! I'm not sure why we omitted touch events from the first array...so I added them in and used the same array for chip interaction and trailing icon interaction.

kfranqueiro: > I'm not sure why we omitted touch events from the first array

Maybe because the code in `handleInteraction` is a no-op for anything besides click or keyboard events anyway? I don't think adding touch events will be harmful but they'll also never do anything.

bonniezhou: Oh good point. The touch events are also a no-op in `handleTrailingIconInteraction`, so I'm just gonna remove them altogether.

PR #3153, opened 2018-07-20
fix(switch): Update switch styling to match the design spec

rlfriedman:
Checked with design and the knob should be further out than it was to make more of the track visible. This change results in 16px of the track being visible in both checked and unchecked positions (not covered by the knob). Before this change only 12px was visible.

Also added a no-op fix to correctly calculate underlay top.


kfranqueiro:
Another thing I just noticed: the click/tap target extends to the left edge of the ripple when the switch is in the off state, but does not extend to the right edge of the ripple when the switch is in the on state.

rlfriedman:
Opened #3161 for the first issue you mentioned. The second issue was from not updating the size of the native control to reflect the now further extending ripple, so will fix that. Thanks!

rlfriedman:
Updated the values to be calculated based on variables. Let me know what you think!

PR #3164, opened 2018-07-21
feat(tab): Give focus to tab when activated

### LINKED ISSUE #3163: Tab should receive focus when activated
- Update Tab adapter to include a `focus` method
- Use the `focus` method in the Tab foundation `activate` method
- Update readme with new adapter method
- Update tests with new adapter method

patrickrodee:
Closes #3163

### DIFF
--- a/packages/mdc-tab/README.md
+++ b/packages/mdc-tab/README.md
@@ -145,7 +145,7 @@ Method Signature | Description
 `getContentOffsetWidth() => number` | Returns the `offsetWidth` value of the content element.
 `notifyInteracted() => void` | Emits the `MDCTab:interacted` event.
 `notifyActivated() => void` | Emits the `MDCTab:activated` event.
-`focus() => void` | Gives focus to the root element.
+`focus() => void` | Applies focus to the root element.
 
 ### `MDCTabFoundation`
 
--- a/packages/mdc-tab/README.md
+++ b/packages/mdc-tab/README.md
@@ -145,6 +145,7 @@ Method Signature | Description
 `getContentOffsetWidth() => number` | Returns the `offsetWidth` value of the content element.
 `notifyInteracted() => void` | Emits the `MDCTab:interacted` event.
 `notifyActivated() => void` | Emits the `MDCTab:activated` event.
+`focus() => void` | Gives focus to the root element.
 
 ### `MDCTabFoundation`
 
--- a/packages/mdc-tab/adapter.js
+++ b/packages/mdc-tab/adapter.js
@@ -124,6 +124,11 @@ class MDCTabAdapter {
    * @return {number}
    */
   getContentOffsetWidth() {}
+
+  /**
+   * Applies focus to the root element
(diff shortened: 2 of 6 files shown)


patrickrodee:
If the tab is programmatically activated, then the content will change and
I’d think the user would want to be aware that something changed via a
focus adjustment.

Focusing on activation via keyboard navigation is part of the W3 spec for
accessible tabs. That was my goal here: to add support for keyboard
navigable Tabs.

https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

kfranqueiro:
I realize this is about keyboard a11y, but that example seems to be showing almost the reverse of what this is doing: the example activates a tab when it receives focus. This PR focuses a tab when it's activated.

There's also another WAI-ARIA example that shows tabs not being activated until space/enter is pressed when they have focus. Our design article seems to show focus states for both activated and unactivated tabs, and our earlier Tabs implementation doesn't activate on focus. Have we talked to a designer about which interaction behavior is intended for Material tabs? I don't see clear guidance on keyboard navigation.

Either way, this PR seems potentially backwards - should we be manipulating focus when arrow keys are used in the tab bar and activating based on focus or space/enter, instead of driving activation off of arrow keys?

mathiasschopmans:
**tl;dr:** Can this behaviour be optional/configurable please?

---

I am implementing MDC components on a onepage website and (mis)using the tabs as a navigation element in a fixed header, which will scroll the user to some sections of the page.

Additionally, I'm using something like Bootstrap's scrollspy-script to detect whenever the user scrolls to a new section of the page and use `instance.activateTab()` to mark the current visible section's link (tab) … and this is causing an issue when I'm trying to use the page by keaboard only:

**Example:**
- Tabbing: Focus on the tabs. ✔️ 
- Tabbing again: Focus is inside a input-element: ✔️ 
- Tabbing again: Focus is inside a input element within the next section below the fold
  - Browser scrolls focused element into viewport ✔️ 
  - Scrollspy script will activate second tab via `.activateTab(1)` ✔️ 
  - Focus isn't inside the input-element anymore. Instead tab has focus. ❌ 

I thought this behaviour could be somehow related to the `useAutomaticActivation` property, should'nt it, @kfranqueiro? Any help / suggestion would be appreciated. 🙇

kfranqueiro:
`useAutomaticActivation` was added in #3303 and is false by default, in which case focusing on tabs shouldn't immediately activate them.

However, it sounds like you have the reverse issue - you're programmatically activating a tab, which gives it focus automatically.

Can you log a separate issue for this? We should probably check with our interaction designers on it.

mathiasschopmans:
Ah, I see. You're right. In this case `useAutomaticActivation` shouldn't be used.

> However, it sounds like you have the reverse issue - you're programmatically activating a tab, which gives it focus automatically.
Jup, that's my issue. :(

> Can you log a separate issue for this? 
I will.

Issue #3198, opened 2018-07-24
Update Select Eventing by removing registerInteractionHandler and deregisterInteractionHandler adapter methods


moog16:
We need to remove the deregisterInteractionHandler and registerInteractionHandler methods from the select adapter, since we are moving away from this pattern. We will instead replace them with public handleEvent foundation methods which will be used from the component.

kfranqueiro:
Related to #2813

PR #3204, opened 2018-07-25
fix(select): reduce adapter apis not used in MDCReact and update events to new pattern

moog16:
fixes https://github.com/material-components/material-components-web/issues/3198


moog16:
Will do!

PR #3220, opened 2018-07-25
feat(tab): Update layout; Add fixed-width mixin; Add min-width class

patrickrodee:
Update the layout to rely more directly on flexbox positioning instead of padding. Add a mixin for setting a tab to a fixed width. Add a class to make a tab shrink to the narrowest possible width.


packages/mdc-tab/README.md
```diff
@@ -104,6 +105,7 @@ Mixin | Description
 `mdc-tab-text-label-color($color)` | Customizes the color of the tab text label.
 `mdc-tab-icon-color($color)` | Customizes the color of the tab icon.
 `mdc-tab-parent-positioning` | Sets the positioning of the MDCTab's parent element so that `MDCTab.computeDimensions()` reports the same values in all browsers.
+`mdc-tab-fixed-width($width)` | Sets the fixed width of the tab. The tab will never be smaller than the given width, although it may be larger if the content of the tab is larger than the given width.
```
acdvorak: Should this be named `mdc-tab-min-width` instead?

packages/mdc-tab/mdc-tab.scss
```diff
@@ -63,15 +67,20 @@
...
+  align-items: center;
   justify-content: center;
   height: inherit;
   pointer-events: none;
   z-index: 2;
 }
 
+.mdc-tab__text-label,
+.mdc-tab__icon {
+  z-index: 2;
```
acdvorak: Why is this needed?

patrickrodee: Checking...

packages/mdc-tab/README.md
patrickrodee: It does set the tab to a fixed width from which it won't grow or shrink. I think min-width isn't quite as accurate.

patrickrodee:
The z-index on the text label and icon that you originally asked about was to fix layering issues when an icon tab indicator was placed inside the tab content. In those cases, the icon tab indicator would appear "on top" of the text label and the icon. By removing the z-index from the tab content and instead applying it directly to the text label and the icon, we can get the original intention and solve the layering problem I originally set out to solve, and with no additional z-index properties. Yay

PR #3221, opened 2018-07-25
feat(menu): New menu

williamernest:
New menu. 

Still working on Readme updates.


packages/mdc-menu-surface/foundation.js
```diff
@@ -124,8 +124,8 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
     this.animationRequestId_ = 0;
     /** @private {!{ width: number, height: number }} */
     this.dimensions_;
-    /** @private {MenuSurfaceCorner} */
-    this.anchorCorner_ = MenuSurfaceCorner.TOP_START;
+    /** @private {Corner} */
```
acdvorak: You might want to double check with @bwobrien if these type annotations need to be marked non-null - e.g., `{!Corner}`.

I realize `Corner` is an `@enum`, which is basically just an alias for `number` which is non-null by default, but I vaguely recall something in the Google JS style guidelines and/or Closure Compiler docs about how you should always mark them explicitly non-null regardless.

Ditto elsewhere in this PR.

packages/mdc-menu/_mixins.scss
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+//
+
+@import "./variables";
+
+@mixin mdc-menu-width($width) {
+  width: $width;
+}
+
+@mixin mdc-menu-width-calc($multiplier) {
```
acdvorak: Maybe rename this to something like:

- `mdc-menu-width-multiplier($multiplier)`

Or, better yet, combine these two functions into one:

```sass
@mixin mdc-menu-width($width) {
  @if unit­les­s($­width) {
    width: $width * $mdc-menu-width-base;
  } @else {
    width: $width;
  }
}

...

@include mdc-menu-width(3); // multiplier
@include mdc-menu-width(300px); // verbatim
@include mdc-menu-width(calc(100% - 50px)); // verbatim
```

packages/mdc-menu/adapter.js
```diff
@@ -37,128 +34,88 @@
...
    */
-  setAttrForOptionAtIndex(index, attr, value) {}
+  getSelectedElementIndex(selectionGroup) {};
 
   /**
-   * @param {number} index
-   * @param {string} attr
+   * Emits an event using the evtData.
+   * @param {{
+ *   index: number
```
acdvorak: Indentation

packages/mdc-menu/adapter.js
```diff
@@ -37,128 +34,88 @@
...
-   * @param {string} className
    */
-  addClassForOptionAtIndex(index, className) {}
+  getCheckbox(index) {};
 
   /**
-   * @param {number} index
-   * @param {string} className
+   * Toggles the checkbox within a list item, if any.
+   * @param {HTMLElement} target
```
acdvorak: Non-primitive types should be explicitly non-null: `{!HTMLElement}`

packages/mdc-menu/adapter.js
```diff
@@ -37,128 +34,88 @@
...
   /**
-   * @param {number} index
-   * @param {string} attr
-   * @param {string} value
+   * Returns the element within the selectionGroup containing the selected element class.
+   * @param {HTMLElement} selectionGroup
+   * @return {number} element within the selectionGroup that contains the selected element class.
    */
-  setAttrForOptionAtIndex(index, attr, value) {}
+  getSelectedElementIndex(selectionGroup) {};
```
acdvorak: Remove semicolons (ditto below)

packages/mdc-menu/foundation.js
```diff
@@ -57,593 +36,186 @@ class MDCMenuFoundation extends MDCFoundation {
...
-      } else {
-        this.adapter_.focusItemAtIndex(focusedItemIndex - 1);
-      }
-    } else if (isArrowDown) {
-      if (focusedItemIndex === lastItemIndex || this.adapter_.isFocused()) {
-        this.adapter_.focusItemAtIndex(0);
-      } else {
-        this.adapter_.focusItemAtIndex(focusedItemIndex + 1);
-      }
+    if (isSpace || isEnter) {
```
acdvorak: This method is really simple and readable. Most keyboard event handling functions are NOT.

Nice work!

packages/mdc-menu/foundation.js
```diff
@@ -57,593 +36,186 @@ class MDCMenuFoundation extends MDCFoundation {
...
+    this.adapter_.addClassToElementAtIndex(index, cssClasses.MENU_SELECTED_LIST_ITEM);
+    this.adapter_.addAttributeToElementAtIndex(index, strings.ARIA_SELECTED_ATTR, 'true');
   }
 
   /**
-   * @param {Corner} corner Origin corner of the menu.
-   * @return {number} Maximum height of the menu, based on available space. 0 indicates should not be set.
+   * Returns the parent selection group of an element or the
+   * @param listItem
+   * @return {HTMLElement|null}
```
acdvorak: Is there a reason you prefer `Type|null` over `?Type`?

packages/mdc-menu/foundation.js
```diff
@@ -57,593 +36,186 @@ class MDCMenuFoundation extends MDCFoundation {
...
-
-    // Clear measures after positioning is complete.
-    this.measures_ = null;
   }
 
   /**
-   * Open the menu.
-   * @param {{focusIndex: ?number}=} options
+   * Find the first ancestor with the mdc-list-item class.
+   * @param {HTMLElement} target
```
acdvorak:
```
@param {!HTMLElement} target
@return {?HTMLElement}
```

packages/mdc-menu/foundation.js
```diff
@@ -57,593 +36,186 @@ class MDCMenuFoundation extends MDCFoundation {
...
-  getSelectedIndex() {
-    return this.selectedIndex_;
+    return target;
   }
 
   /**
-   * @param {number} index Index of the item to set as selected.
+   * Ensures that preventDefault is only called if the containing element doesn't
+   * consume the event, and it will cause an unintended scroll.
+   * @param {Event} evt
```
acdvorak: `{!Event}`

packages/mdc-menu/index.js
```diff
@@ -27,79 +29,105 @@ class MDCMenu extends MDCComponent {
...
-      this.foundation_.close();
+      this.hide();
     }
   }
 
-  /** @param {{focusIndex: ?number}=} options */
-  show({focusIndex = null} = {}) {
-    this.foundation_.open({focusIndex: focusIndex});
+  show() {
+    this.menuSurface_.listen(MDCMenuSurfaceFoundation.strings.OPENED_EVENT, () => {
```
acdvorak: This registers a new event listener every time `show()` is called.

It might be better to make it idempotent, so event handlers don't get called more than once for the same event.

And probably add a unit test that calls `show()` several times in a row.

packages/mdc-menu/index.js
```diff
@@ -111,85 +139,100 @@ class MDCMenu extends MDCComponent {
...
+  }
+
+  /**
+   * Sets the element that the menu-surface is anchored to.
+   * @param {!HTMLElement} element
+   */
+  setAnchorElement(element) {
+    this.menuSurface_.anchorElement = element;
+  }
+
```
acdvorak: Remove extra line

packages/mdc-menu/mdc-menu.scss
```diff
@@ -14,135 +14,58 @@
...
-    // stylelint-disable plugin/selector-bem-pattern
-    > .mdc-list-item {
-      cursor: pointer;
-    }
-    // stylelint-enable plugin/selector-bem-pattern
-
-    .mdc-menu--animating & {
-      overflow-y: hidden;
-    }
+  .mdc-list {
```
acdvorak: Doesn't this violate BEM by creating increasing specificity with descendent selectors?

packages/mdc-menu/adapter.js
williamernest: I wasn't able to remove all non-null types, but I went through and declared everything either nullable/non-nullable.

packages/mdc-menu-surface/foundation.js
```diff
@@ -162,7 +162,7 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
   }
 
   /**
-   * @param {!MenuSurfaceCorner} corner Default anchor corner alignment of top-left menu surface corner.
+   * @param {!Corner} corner Default anchor corner alignment of top-left menu surface corner.
```
kfranqueiro: I see this one `{!Corner}` but otherwise a whole lot of `{Corner}`s still. What was the verdict on Andy's comment about whether enums require explicit `!`?

packages/mdc-menu/README.md
```diff
@@ -108,6 +108,13 @@ CSS Class | Description
 `mdc-menu--open` | Indicates the menu is currently open, or is currently animating open.
 `mdc-menu--animating-closed` | Indicates the menu is currently animating closed. This class is removed once the animation completes.
 
+### Sass Mixins
+
+Mixin | Description
+--- | ---
+`mdc-menu-width($multiplier)` | Used to set the width of the menu. Accepts an integer to multiply the base width of `56px`.
+`mdc-menu-width($value)` | Used to set the width of the menu. Sets the menu width to `$value`.
```
kfranqueiro: Might want to somehow distinguish that this version of the mixin applies to values with units whereas the multiplier version applies to unitless values?

packages/mdc-menu/_mixins.scss
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@import "./variables";
+
+@mixin mdc-menu-width($width) {
+  @if unitless($width) {
+    width: $width * $mdc-menu-width-base;
```
kfranqueiro: I have some questions about this mixin...

1. What is the motive behind supporting passing a unitless multiplier? I didn't seem to see anything in the spec suggesting this...
2. What is the motive for using this mixin at all, rather than letting the menu auto-size based on its content?

packages/mdc-menu/index.js
```diff
@@ -111,85 +138,106 @@ class MDCMenu extends MDCComponent {
...
+        } else {
+          return null;
+        }
       },
-      setMaxHeight: (height) => {
-        this.root_.style.maxHeight = height;
+      toggleCheckbox: (checkBox) => {
+        if (checkBox) {
+          checkBox.checked = !checkBox.checked;
+          const event = document.createEvent('Event');
```
kfranqueiro: Why are we emitting a change event here?

This seems like more than what someone would expect of this adapter API... maybe we should put this logic into a util function and call that here and allow wrappers to reuse it, if this is needed?

packages/mdc-menu/index.js
```diff
@@ -111,85 +138,106 @@ class MDCMenu extends MDCComponent {
...
+        if (list.length > index) {
+          return list[index].querySelector(strings.CHECKBOX_SELECTOR);
+        } else {
+          return null;
+        }
       },
-      setMaxHeight: (height) => {
-        this.root_.style.maxHeight = height;
+      toggleCheckbox: (checkBox) => {
+        if (checkBox) {
```
kfranqueiro: AFAICT this check shouldn't need to be in the adapter impl; the foundation is already guarding against this.

packages/mdc-menu/index.js
```diff
@@ -111,85 +138,106 @@ class MDCMenu extends MDCComponent {
...
-      setPosition: (position) => {
-        this.root_.style.left = 'left' in position ? position.left : null;
-        this.root_.style.right = 'right' in position ? position.right : null;
-        this.root_.style.top = 'top' in position ? position.top : null;
-        this.root_.style.bottom = 'bottom' in position ? position.bottom : null;
+      notifySelected: (evtData) => this.emit(strings.SELECTED_EVENT, {
+        index: evtData.index,
+        item: this.items[evtData.index],
+      }),
+      getCheckbox: (index) => {
```
kfranqueiro: Maybe consider renaming this API `getCheckboxAtIndex` to be more clear and consistent with methods above?

packages/mdc-menu/index.js
```diff
@@ -111,85 +138,106 @@ class MDCMenu extends MDCComponent {
...
-      removeClass: (className) => this.root_.classList.remove(className),
-      hasClass: (className) => this.root_.classList.contains(className),
-      hasNecessaryDom: () => Boolean(this.itemsContainer_),
-      getAttributeForEventTarget: (target, attributeName) => target.getAttribute(attributeName),
-      getInnerDimensions: () => {
-        const {itemsContainer_: itemsContainer} = this;
-        return {width: itemsContainer.offsetWidth, height: itemsContainer.offsetHeight};
+      addClassToElementAtIndex: (index, className) => {
+        const list = this.items;
+        if (list.length > index) {
```
kfranqueiro: Questions about these various adapters that check `length > index`:

1. Should the adapters even need to sanity-check this? Should it even be possible for the foundation to pass an invalid index to these?
2. If the sanity check is needed, should we be checking both boundaries rather than just the high end?

packages/mdc-menu/mdc-menu.scss
```diff
@@ -14,135 +14,56 @@
...
-    }
+@import "@material/rtl/mixins";
+@import "@material/list/mixins";
+@import "@material/ripple/common";
+@import "@material/ripple/mixins";
+@import "./variables";
+@import "./mixins";
+
+// Customize the menu-surface and contained list to match the mdc-menu styles.
+.mdc-menu-surface {
```
kfranqueiro: Wait... we're directly applying styles to another package's selector?

If `mdc-menu` isn't applying its own base selector to distinguish itself, then what's the point of `mdc-menu-surface` being a separate package at all?

packages/mdc-menu/index.js
williamernest: Toggling checked of a checkbox doesn't cause a change event. If someone is listening to the change event of that checkbox to do something with it, it will never get fired if we don't emit the change event.

packages/mdc-menu/mdc-menu.scss
williamernest: Menu surface is a separate package because the auto positioning surface logic can be reused by other components in the future. The `mdc-menu` doesn't have a lot of individual styles it needs other than modifying the surface and list. We could change this to have it's own styles, but it seems a little redundant to have `mdc-menu mdc-menu-surface` on the root element.

packages/mdc-menu/_mixins.scss
williamernest: Spec guidelines recommend the menu width being a multiple of `56px` (min: 2, max: 6). I wanted to have a mixin to make the multiple apparent while also allowing the user to override it with their custom value.

packages/mdc-menu/mdc-menu.scss
kfranqueiro: My question is how would anything else reuse _just_ the menu surface logic if it inevitably ends up dragging whatever CSS is added by `mdc-menu` along with it in any case where both packages are loaded?

williamernest: Good point. I'll change these styles to a `mdc-menu` class selector.

packages/mdc-menu/index.js
williamernest: I'm probably being overly defensive here. I'll remove the check.

packages/mdc-menu-surface/foundation.js
bwobrien: Style guidelines say to always include either `?` or `!` for enums.

packages/mdc-menu/README.md
```diff
@@ -56,13 +53,72 @@ A menu is initially hidden, appearing when opened via the JS API. It is appropri
...
 
-#### Anchor To Parent
+### Styles
+
+```scss
+@import "@material/menu/mdc-list";
+@import "@material/menu/mdc-menu-surface";
+@import "@material/menu/mdc-menu";
+
+.demo-menu {
```
kfranqueiro: Remove `.demo-menu` block

packages/mdc-menu/README.md
```diff
@@ -56,13 +53,72 @@ A menu is initially hidden, appearing when opened via the JS API. It is appropri
...
+}
+
+```
+
+### JavaScript Instantiation
+
+```js
+import {MDCMenu} from '@material/menu';
+
+const menu = new MDCMenu(document.querySelector('.demo-menu'));
```
kfranqueiro: demo-menu -> mdc-menu

packages/mdc-menu/foundation.js
```diff
@@ -57,593 +37,188 @@ class MDCMenuFoundation extends MDCFoundation {
...
 
-    return maxHeight;
-  }
-
-  /** @private */
-  autoPosition_() {
-    if (!this.adapter_.hasAnchor()) {
-      return;
+    // Iterate through ancestors until we find the group or get to the list.
+    while (!isGroup && !this.adapter_.elementContainsClass(parent, MDCListFoundation.cssClass.ROOT)) {
```
kfranqueiro: cssClass -> cssClasses?

test/unit/mdc-menu/mdc-menu.test.js
```diff
@@ -16,422 +16,404 @@
...
 });
 
-test('adapter#isRtl returns false for implicit LTR documents', () => {
-  const anchor = bel`<div class="mdc-menu-anchor"></div>`;
-  const {root, component} = setupTest(true);
-  anchor.appendChild(root);
-  document.body.appendChild(anchor);
-  assert.isNotOk(component.getDefaultFoundation().adapter_.isRtl());
-  document.body.removeChild(anchor);
+test('adapter#closeSurface returns false if the class does not exist on the element', () => {
```
kfranqueiro: Wrong description?

test/unit/mdc-menu/mdc-menu.test.js
```diff
@@ -16,422 +16,404 @@
...
+  assert.equal(0, indexValue);
+});
 
-  component.getDefaultFoundation().adapter_.setTransformOrigin('left top 10px');
-  assert.equal(root.style.getPropertyValue(`${getTransformPropertyName(window)}-origin`), expected);
+test('adapter#getElementIndex returns -1 if the element does not exist in the list', () => {
+  const {component} = setupTest();
+  const firstItem = document.createElement('li');
+  const indexValue = component.getDefaultFoundation().adapter_.getElementIndex(firstItem);
+  assert.equal(-1, indexValue);
```
kfranqueiro: The `assert.equal` API specifies `actual` before `expected`. These tests will probably produce confusing messages if they fail.

test/unit/mdc-menu/mdc-menu.test.js
```diff
@@ -16,422 +16,404 @@
...
-  component.getDefaultFoundation().adapter_.addClassForOptionAtIndex(1, 'test-class');
-  assert.equal(root.querySelector('.test-class'), item1);
-  document.body.removeChild(root);
+test('adapter#getCheckboxAtIndex returns null if a checkbox does not exist in the element at index specified', () => {
+  const {component} = setupTest();
+  assert.isNull(component.getDefaultFoundation().adapter_.getCheckboxAtIndex(0));
 });
 
-test('adapter#rmClassForOptionAtIndex removes a class from the option at the index specified', () => {
+test('adapter#toggleCheckbox toggle a checkbox', () => {
```
kfranqueiro: toggle -> toggles

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
-    td.verify(mockAdapter.deregisterBodyClickHandler(td.matchers.anything()), {times: 0});
-  });
-
-testFoundation('#close adds the animation class to start an animation', ({foundation, mockAdapter}) => {
-  foundation.close();
-  td.verify(mockAdapter.addClass(cssClasses.ANIMATING_CLOSED));
-});
+test('handleKeydown space/enter key causes the menu to close', () => {
+  const {foundation, mockAdapter} = setupTest();
+  const key = {key: 'Space', target: 'My Element', preventDefault: td.func('preventDefault')};
```
kfranqueiro: key -> event?

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
+  td.when(mockAdapter.elementContainsClass(key.target, listClasses.LIST_ITEM_CLASS)).thenReturn(true);
+  td.when(mockAdapter.getElementIndex(key.target)).thenReturn(0);
 
-testFoundation('#close does not add animation class if quickOpen is set to true', ({foundation, mockAdapter}) => {
-  foundation.setQuickOpen(true);
-  foundation.close();
-  td.verify(mockAdapter.addClass(cssClasses.ANIMATING_CLOSED), {times: 0});
-  td.verify(mockAdapter.removeClass(cssClasses.ANIMATING_CLOSED), {times: 0});
-});
+  foundation.handleKeydown(key);
```
kfranqueiro: Should this be running another test with the enter key and checking for 2 calls?

Ditto on the next few tests

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
 
-  foundation.init();
-  handlers.keydown({target, keyCode: 32, preventDefault});
-  handlers.keyup({target, keyCode: 32});
-  clock.tick(numbers.SELECTED_TRIGGER_DELAY);
-  td.verify(mockAdapter.notifySelected({index: 0}));
+  foundation.handleKeydown(key);
+  clock.tick(numbers.TRANSITION_CLOSE_DURATION);
 
+  td.verify(mockAdapter.removeClassFromElementAtIndex(0, cssClasses.MENU_SELECTED_LIST_ITEM), {times: 0});
```
kfranqueiro: Replace `0` first argument to adapter call with `td.matchers.isA(Number)`?

(Same for the next 2 tests in `{times: 0}` checks, to ensure they're not calling it at all with any value)

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
-  raf.flush();
-  td.verify(mockAdapter.focusItemAtIndex(1));
-
-  raf.restore();
-  clock.uninstall();
+  td.verify(key.preventDefault(), {times: 0});
 });
 
-test('on ArrowDown keydown prevents default on the event', () => {
+test('Click event inside a list item causes the preventDefault to be called', () => {
```
kfranqueiro: "causes preventDefault"

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
+// Clicks
 
-test('on enter keyup does works if DOM3 keyboard events are not supported', () => {
+test('Click event causes the menu to close', () => {
   const {foundation, mockAdapter} = setupTest();
-  const handlers = captureHandlers(mockAdapter, 'registerInteractionHandler');
-  const clock = lolex.install();
-  const target = {};
-  td.when(mockAdapter.getIndexForEventTarget(target)).thenReturn(0);
+  const key = {target: 'My Element', preventDefault: td.func('preventDefault')};
```
kfranqueiro: `key` -> `event` in the click event test cases

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
-  raf.flush();
-  td.verify(preventDefault());
-
-  raf.restore();
-  clock.uninstall();
+  td.verify(mockAdapter.notifySelected(td.matchers.anything()), {times: 0});
 });
 
-test('on document click cancels and closes the menu', () => {
+test('Click event not inside of a child of a list item causes the list item to be selected', () => {
```
kfranqueiro: Inaccurate description?

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -80,1165 +51,431 @@ test('exports cssClasses', () => {
...
-  clock.tick(numbers.SELECTED_TRIGGER_DELAY);
-
-  td.verify(mockAdapter.rmClassForOptionAtIndex(expectedIndex, td.matchers.anything()),
-    {times: 0});
-  td.verify(mockAdapter.addClassForOptionAtIndex(expectedIndex, td.matchers.anything()),
-    {times: 1});
+  foundation.handleClick(key);
+  clock.tick(numbers.TRANSITION_CLOSE_DURATION);
 
+  td.verify(mockAdapter.removeClassFromElementAtIndex(0, cssClasses.MENU_SELECTED_LIST_ITEM), {times: 0});
```
kfranqueiro: See earlier comment RE 0 -> td.matchers.isA(Number) for these `{times: 0}` verifications

packages/mdc-menu/index.js
```diff
@@ -111,85 +138,100 @@ class MDCMenu extends MDCComponent {
...
+
+  deregisterListeners_() {
+    this.root_.removeEventListener('keydown', this.handleKeydown_);
+    this.root_.removeEventListener('click', this.handleClick_);
+  }
+
+  handleAfterOpened_() {
+    const list = this.items;
+    if (list.length > 0) {
+      list[0].focus();
```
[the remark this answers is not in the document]
williamernest: Menu doesn't, but List already does. It queries for list items that are enabled `.mdc-list-item:not(.mdc-list-item--disabled)`, which is what the menu uses to get the items.

test/unit/mdc-menu/mdc-menu.test.js
```diff
@@ -16,422 +16,404 @@
...
+import {Corner} from '../../../packages/mdc-menu-surface/constants';
+import {MDCListFoundation} from '../../../packages/mdc-list';
+import {MDCMenuSurfaceFoundation} from '../../../packages/mdc-menu-surface/foundation';
 
 function getFixture(open) {
   return bel`
-    <div class="mdc-menu ${open ? 'mdc-menu--open' : ''}" tabindex="-1">
-      <ul class="mdc-menu__items mdc-list" role="menu">
-        <li class="mdc-list-item" role="menuitem" tabindex="0">Item</a>
+    <div class="mdc-menu-surface ${open ? 'mdc-menu-surface--open' : ''}" tabindex="-1">
```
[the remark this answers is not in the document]
williamernest: Used only for styles but I added it anyways.

test/unit/mdc-menu/menu.foundation.test.js
williamernest: 2 calls of the `closeSurface` ?

kfranqueiro: In this test, yes; 2 calls of notifySelected in the next one, etc.

At least that was my idea, as the shortest way to test both keys.

PR #3222, opened 2018-07-25
feat(tab-scroller): Add scroll content width method for use in tab bar

patrickrodee:
Expose method on foundation and component to get the width of the scroll content element. Add tests for newly exposed scroll content width methods. Add classes for adjusting layout of scroll content child elements. Update test coverage.

### DIFF
--- a/packages/mdc-tab-scroller/README.md
+++ b/packages/mdc-tab-scroller/README.md
@@ -65,6 +65,9 @@ CSS Class | Description
 `mdc-tab-scroller` | Mandatory. Contains the tab scroller content.
 `mdc-tab-scroller__scroll-area` | Mandatory. Denotes the scrolling area.
 `mdc-tab-scroller__scroll-content` | Mandatory. Denotes the scrolling content.
+`mdc-tab-scroller--start-aligned` | Optional. Sets the elements inside the scroll content element to be aligned to the start of the scroll content element.
+`mdc-tab-scroller--end-aligned` | Optional. Sets the elements inside the scroll content element to be aligned to the end of the scroll content element.
+`mdc-tab-scroller--center-aligned` | Optional. Sets the elements inside the scroll content element to be aligned to the center of the scroll content element.
 
 ## `MDCTabScroller` Methods
 
@@ -73,6 +76,7 @@ Method Signature | Description
 `scrollTo(scrollX: number) => void` | Scrolls to the scrollX value.
 `incrementScroll(scrollX: number) => void` | Increments the current scroll value by the scrollX value.
 `getScrollPosition() => number` | Returns the current visual scroll position.
+`getScrollContentWidth() => number` | Returns the width of the scroll content element.
 
 ## Usage within Web Frameworks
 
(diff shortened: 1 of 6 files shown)


packages/mdc-tab-scroller/README.md
```diff
@@ -111,3 +115,4 @@ Method Signature | Description
 `scrollTo(scrollX: number) => void` | Scrolls to the `scrollX` value.
 `incrementScroll(scrollX: number) => void` | Increments the current scroll value by the `scrollX` value.
 `getScrollPosition() => number` | Returns the current visual scroll position.
+`getScrollContentWidth() => number` | Returns the width of the scroll content element.
```
kfranqueiro: Is this actually needed on the foundation, or could we simply write a one-liner referencing the correct node in the component?

patrickrodee:
Could definitely write a one-liner referencing the appropriate node.

packages/mdc-tab-scroller/README.md
```diff
@@ -65,6 +65,9 @@ CSS Class | Description
 `mdc-tab-scroller` | Mandatory. Contains the tab scroller content.
 `mdc-tab-scroller__scroll-area` | Mandatory. Denotes the scrolling area.
 `mdc-tab-scroller__scroll-content` | Mandatory. Denotes the scrolling content.
+`mdc-tab-scroller--start-aligned` | Optional. Sets the elements inside the scroll content element to be aligned to the start of the scroll content element.
```
kfranqueiro: Nit: `--align-start`, `--align-end`, and `--align-center` would...align...with modifier classes we have on a few other components

packages/mdc-tab-scroller/foundation.js
```diff
@@ -167,6 +172,14 @@ class MDCTabScrollerFoundation extends MDCFoundation {
     return this.rtlScrollerInstance_;
   }
 
+  /**
+   * Returns the width of the scroll content
+   * @return {number}
+   */
+  getScrollContentWidth() {
```
moog16: I agree with Ken on this one. I don't think this method is needed, and can just be a one liner in the component.

packages/mdc-tab-scroller/mdc-tab-scroller.scss
```diff
@@ -59,6 +59,18 @@
   will-change: transform;
 }
 
+.mdc-tab-scroller--start-aligned .mdc-tab-scroller__scroll-content {
```
moog16: Instead of having all three (start, end, center), could you just have 2 where 1 is the default?

packages/mdc-tab-scroller/foundation.js
patrickrodee: I don't disagree. Since this is a subcomponent that's required inside of another, we have yet to figure out a pattern for accessing the child elements of a subcomponent.

packages/mdc-tab-scroller/mdc-tab-scroller.scss
patrickrodee: The default is that they fill the horizontal space completely with no specific alignment. These three options are all optional.

moog16: ah gotcha :)

kfranqueiro: The docs were updated but these classes were not

patrickrodee: lololol i'm a fool 😮

PR #3229, opened 2018-07-26
feat(tab-bar): Add tab bar

patrickrodee:


packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+     *    +------------+--------------+-------------------+
+     *    |    Tab     |      Tab     |        Tab        |
+     *    +------------+--------------+-------------------+
+     *        |                                   |
+     *        +-----------------------------------+
+     *
+     * To determine the next adjacent index, we look at the Tab root left and
+     * Tab root right, both relative to the scroll position. If the Tab root
+     * left is less than 0, then we know it's out of view to the left. If the
+     * Tab root right minues the bar width is greater than 0, we know the Tab is
```
blowery: little typo: `minues` -> `minus`

patrickrodee:
Good catch! Thanks Ben, always appreciate another set of eyes on big PRs 😊

blowery:
@patrickrodee ;)

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+    const activeIndex = this.adapter_.getActiveTabIndex();
+    this.scrollIntoView(activeIndex);
+  }
+
+  /**
+   * Activates the tab at the given index
+   * @param {number} index
+   */
+  activateTab(index) {
+    const previousActiveIndex = this.adapter_.getActiveTabIndex();
```
kfranqueiro: `mdc-tab` already has logic to avoid activating if already activated... do we really need to sort of repeat that logic here or is there some way we can make use of the activated event?

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,347 @@
...
+    if (keyboardNavigation.indexOf(evt.key) === -1) {
+      return;
+    }
+
+    evt.preventDefault();
+    this.activateTabFromKeydown_(evt);
+  }
+
+  /**
+   * Handles the MDCTab:interacted event
```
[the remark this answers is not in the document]
kfranqueiro: Pretty sure this is the right event. `interacted` emits when a user interacts with a tab (which might not activate it, e.g. if it's already activated); `activated` emits when it's actually activated.

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+  /**
+   * Determines the index of the next adjacent tab
+   * @param {number} index The index of the tab
+   * @param {!MDCTabDimensions} tabDimensions Whether the next tab is left or right
+   * @param {number} scrollPosition The current scroll position
+   * @param {number} barWidth The width of the tab bar
+   * @return {number}
+   * @private
+   */
+  findTabIndexClosestToEdge_(index, tabDimensions, scrollPosition, barWidth) {
```
kfranqueiro: I'm a little confused since the docs keep talking about the "next adjacent tab" but the method name says "closest to edge". It seems like the definition of "next" is subject to what this method finds? Is there some better way to word the docs to make it more evident what this is used for?

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+   * Determines the index of the next adjacent tab in RTL
+   * @param {number} index The index of the tab
+   * @param {!MDCTabDimensions} tabDimensions Whether the next tab is left or right
+   * @param {number} scrollDistance The distance scrolled
+   * @param {number} scrollWidth The width of the scroller
+   * @param {number} barWidth The width of the bar
+   * @return {number}
+   * @private
+   */
+  findTabIndexClosestToEdgeRTL_(index, tabDimensions, scrollDistance, scrollWidth, barWidth) {
```
kfranqueiro: I'm not sure why some parameter and variable names here are different from the non-RTL counterpart; it seems like the same things are passed in either case. Can we sync these up?

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+    this.adapter_.incrementScroll(scrollIncrement);
+  }
+
+  /**
+   * Scrolls the tab at the given index into view in RTL
+   * @param {number} index The tab index to make visible
+   * @private
+   */
+  scrollIntoViewRTL_(index) {
+    const barWidth = this.adapter_.getOffsetWidth();
```
kfranqueiro: Nit: sync these up with the order of variables in the non-RTL method to make it easier to see the actual differences

packages/mdc-tab-bar/index.js
```diff
@@ -0,0 +1,147 @@
...
+ */
+class MDCTabBar extends MDCComponent {
+  /**
+   * @param {...?} args
+   */
+  constructor(...args) {
+    super(...args);
+
+    /** @type {!Array<!MDCTab>} */
+    this.tabList;
```
kfranqueiro: Is there a reason that `tabList` and `tabScroller` are implicitly public? And should the others use `@private`?

test/unit/mdc-tab-bar/foundation.test.js
```diff
@@ -0,0 +1,543 @@
...
+  td.when(mockAdapter.getActiveTabIndex()).thenReturn(6);
+  td.when(mockAdapter.getTabIndicatorClientRectAtIndex(6)).thenReturn({
+    left: 22, right: 33,
+  });
+  foundation.activateTab(1);
+  td.verify(scrollIntoView(1));
+});
+
+function setupScrollIntoViewTest({
+  activeIndex=0,
```
kfranqueiro: Spaces around `=`s? (Surprised lint didn't complain)

packages/material-components-web/index.js
```diff
@@ -64,6 +65,7 @@ autoInit.register('MDCNotchedOutline', notchedOutline.MDCNotchedOutline);
 autoInit.register('MDCRadio', radio.MDCRadio);
 autoInit.register('MDCSnackbar', snackbar.MDCSnackbar);
 autoInit.register('MDCTab_', tab.MDCTab);
+autoInit.register('MDCTabBar_', tabBar.MDCTabBar);
```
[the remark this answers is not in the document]
patrickrodee: I get collusions if the packages have the same name as the existing ones. I could rename the old ones to `MDCTabBarDeprecated` or something.

packages/mdc-tab-bar/foundation.js
patrickrodee: Good find. I need to bring the docs in-line with the code. I was using the "next adjacent tab" beforehand but that's more ambiguous than "closest to edge". I'll update to use "closest to edge".

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+    this.scrollIntoView(index);
+  }
+
+  /**
+   * Handles the keydown event
+   * @param {!Event} evt
+   */
+  handleKeyDown(evt) {
+    const keyboardNavigation = [
+      'ArrowLeft',
```
kfranqueiro: IE and Edge report `Left` and `Right` instead of `ArrowLeft` and `ArrowRight`, so we need to handle those here and in `activateTabFromKeydown_`.

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,365 @@
...
+
+  /**
+   * Private method for activating a tab from the keydown
+   * @param {!Event} evt The keydown event
+   * @private
+   */
+  activateTabFromKeydown_(evt) {
+    const isRTL = this.isRTL_();
+    const maxTabIndex = this.adapter_.getTabListLength() - 1;
+    const shouldGoToEnd = evt.key === 'End' && !isRTL || evt.key === 'Home' && isRTL;
```
kfranqueiro: I don't think we should be reversing the meaning of home and end in RTL. RTL reverses left and right, but the notion of which tab is first and which tab is last shouldn't change, right?

patrickrodee: Makes sense. Lynn and I had a similar conversation. I'll update this.

packages/mdc-tab-bar/foundation.js
patrickrodee: This is to avoid unknown side-effects when a tab is deactivated then immediately activated again with it's own indicator's client rect.

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,347 @@
...
+    this.adapter_.deactivateTabAtIndex(previousActiveIndex);
+    this.adapter_.activateTabAtIndex(index, this.adapter_.getTabIndicatorClientRectAtIndex(previousActiveIndex));
+    this.scrollIntoView(index);
+  }
+
+  /**
+   * Handles the keydown event
+   * @param {!Event} evt
+   */
+  handleKeyDown(evt) {
```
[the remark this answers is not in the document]
patrickrodee: You have to wire it up by hand, yeah. At least now it works for users that want to use keyboard navigation regardless of screen reader enabled-ness.

packages/mdc-tab-bar/foundation.js
```diff
@@ -0,0 +1,347 @@
...
+    this.scrollIntoView(activeIndex);
+  }
+
+  /**
+   * Activates the tab at the given index
+   * @param {number} index
+   */
+  activateTab(index) {
+    const previousActiveIndex = this.adapter_.getActiveTabIndex();
+    if (!this.indexIsInRange_(index) || index === previousActiveIndex) {
```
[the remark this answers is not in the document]
patrickrodee: I'd rather not throw errors and instead just do nothing. Throwing errors feels wrong for a component.

packages/mdc-tab-bar/foundation.js
patrickrodee: Ah my bad, I misunderstood. I'll update.

packages/material-components-web/index.js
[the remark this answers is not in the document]
patrickrodee: Done!

packages/mdc-tab-bar/foundation.js
patrickrodee: Fixed

PR #3233, opened 2018-07-26
fix(select): add adapter

moog16:


packages/mdc-ripple/adapter.js
```diff
@@ -38,6 +38,7 @@
  *
  * @record
  */
+
```
kfranqueiro: How'd this get into this PR? Is this fixing anything?

packages/mdc-select/README.md
```diff
@@ -242,12 +242,19 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
...
 | `activateBottomLine() => void` | Activates the bottom line component. |
+| `addClass(className: string) => void` | Adds a class to the root element. |
+| `closeOutline() => void` | Switches the notched outline element to its closed state. |
 | `deactivateBottomLine() => void` | Deactivates the bottom line component. |
+| `floatLabel(value: boolean) => void` | Floats or defloats label. |
+| `getLabelWidth() => number` | Returns the offsetWidth of the label element. |
 | `getValue() => string` | Returns the value selected on the `select` element. |
+| `hasClass(className: string) => boolean` | Returns true if the root element has the className in its classList. |
+| `hasLabel() => boolean` | Returns true if the `select` has a label associated with it. |
+| `hasOutlined() => boolean` | Returns true if the `select` has the notched outline element. |
```
kfranqueiro: Typo: hasOutlined -> hasOutline

packages/mdc-select/README.md
```diff
@@ -242,12 +242,19 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
 
 | Method Signature | Description |
 | --- | --- |
-| `addClass(className: string) => void` | Adds a class to the root element. |
-| `removeClass(className: string) => void` | Removes a class from the root element. |
-| `floatLabel(value: boolean) => void` | Floats or defloats label. |
 | `activateBottomLine() => void` | Activates the bottom line component. |
+| `addClass(className: string) => void` | Adds a class to the root element. |
```
kfranqueiro: I think we have typically kept correlated APIs together, so can we move removeClass and hasClass back here, and put APIs for the label / outline each together, etc?

packages/mdc-select/adapter.js
```diff
@@ -0,0 +1,120 @@
...
+
+class MDCSelectAdapter {
+  /**
+   * Adds class to root element.
+   * @param {string} className
+   */
+  addClass(className) {}
+
+  /**
+   * Removes a class from the root Element.
```
kfranqueiro: Element -> element to be consistent with other docs

packages/mdc-select/adapter.js
```diff
@@ -0,0 +1,120 @@
...
+  isRtl() {}
+
+  /**
+   * Returns true if label element exists, false if it doesn't.
+   * @return {boolean}
+   */
+  hasLabel() {}
+
+  /**
+   * Only implement if label exists.
```
kfranqueiro: I would rephrase all of these "Only implement" comments, e.g. "Returns width of label in pixels, if the label exists."

The adapter APIs are always implemented, but ones like this simply return early if the element in question doesn't exist.

packages/mdc-select/foundation.js
```diff
@@ -64,18 +83,27 @@ export default class MDCSelectFoundation extends MDCFoundation {
     }
   }
 
+  /**
+   * Handles value event changes.
```
kfranqueiro: Handles value changes, via change event or programmatic updates.

packages/mdc-ripple/adapter.js
moog16: hmm weird...don't remember that

packages/mdc-select/README.md
moog16: gotcha - i alpha sorted

PR #3239, opened 2018-07-26
chore(infrastructure): Skip unnecessary tests on Travis

acdvorak:
### 1. Skips tests in Travis CI if there are no testable files in the PR

- Cuts Travis build times in half for PRs that only update docs/demos (from ~10 min to ~5 min).

- Runs a shell script on `before_install` that checks for files that need to be tested in the PR.

- Each `TEST_SUITE` has its own criteria for "testable" files:
    - `unit` tests only run if `packages/**/*.(js|scss)` or `test/unit/**/*` files have changed.

    - `site-generator` test only runs if `docs/**/*` or `*.md` files have changed.

    - See `scripts/travis-env-vars.sh` for more details

- If no testable files are found, the script sets a `HAS_TESTABLE_FILES=true` env var that tells subsequent Travis commands to skip tests. This allows most jobs to pass quickly (30-60 seconds) without running any unnecessary commands like `npm install` or `npm run lint`.

    - Screenshot tests always build MDC source code, even if tests are skipped. This acts as a sanity check, and it only takes ~30 seconds to build.

### FAQ: Why are there `if` statements in `.travis.yml`?

E.g.:

```
    script:
      - if has_testable_files; then npm run lint; else log_untestable_files; fi
```

I originally called `exit` in the shell script, but that caused Travis job logs to get truncated:

- https://travis-ci.org/material-components/material-components-web/jobs/408765826
- https://travis-ci.org/material-components/material-components-web/jobs/408760221

That's because:

1. I'm `source`-ing our shell scripts so that their env vars are available to subsequent commands; and
2. [Travis wraps every command from `.travis.yml` in its own shell script](https://docs.travis-ci.com/user/customizing-the-build/#note-on-), so calling `exit` in _our_ script actually exits _Travis's_ special wrapper script. This, in turn, prevents Travis from flushing the log stream.

The only way I could find to skip the tests _without_ failing the build and _without_ truncating logs was to add `if` statements in `.travis.yml`. 🤷‍♂️ 

## Example Travis logs

Example PR with only Markdown changes: #3240

#### Unit tests on a PR with only Markdown changes:

https://travis-ci.org/material-components/material-components-web/jobs/408829485

image

#### Screenshot tests on a PR with only Markdown changes:

https://travis-ci.org/material-components/material-components-web/jobs/408829488

image

#### Normal build times for a PR with source code changes:

https://travis-ci.org/material-components/material-components-web/builds/408790287

image

#### Short build times for a PR with only Markdown changes:

https://travis-ci.org/material-components/material-components-web/builds/408790304

image


.travis.yml
```diff
@@ -36,10 +36,12 @@ matrix:
...
-      - source test/screenshot/infra/commands/travis.sh
+    # Source the scripts to export their env vars. See https://superuser.com/a/176788/62792
+    - source scripts/skip-unnecessary-travis-tests.sh
+    - source test/screenshot/infra/commands/travis.sh
     install:
-      - npm install
-      #- npm ls # Noisy output, but useful for debugging npm package dependency version issues
+    - npm install
+    #- npm ls # Noisy output, but useful for debugging npm package dependency version issues
+before_install:
```
kfranqueiro: Is this really supposed to be down here like this? Is this supposed to take effect globally? If so, does it still need to be above for screenshot tests?

scripts/skip-unnecessary-travis-tests.sh
```diff
@@ -0,0 +1,96 @@
...
+    exit 1
+  fi
+}
+
+print_travis_env_vars
+print_all_changed_files
+
+if [[ "$TEST_SUITE" == 'unit' ]]; then
+  # Only run unit tests if JS files changed
+  exit_if_files_not_changed '^packages/.+\.js$'
```
kfranqueiro: Shouldn't this also test for js files within the test/unit folder?

scripts/skip-unnecessary-travis-tests.sh
```diff
@@ -0,0 +1,96 @@
...
+  # Only run site-generator test if docs, Markdown, or image files changed
+  exit_if_files_not_changed '^docs/' '\.md$' '\.(png|jpg|jpeg|gif|svg)$'
+fi
+
+if [[ "$TEST_SUITE" == 'screenshot' ]]; then
+  # Only run screenshot tests if package JS/Sass files, non-Markdown screenshot test files, or image files changed.
+  check_for_testable_files '^packages/.+\.(js|css|sass)$' '^test/screenshot/.+[^m][^d]$' '\.(png|jpg|jpeg|gif|svg)$'
+
+  if [[ $? != 0 ]]; then
+    export TESTABLE_FILE_COUNT=0
```
kfranqueiro: What's the motive behind doing this rather than exiting early like with the others? Will this still test the build this way?

scripts/skip-unnecessary-travis-tests.sh
acdvorak: Yep, good catch! I meant to add that but forgot. Thanks!

scripts/skip-unnecessary-travis-tests.sh
acdvorak: If we exit early, it won't build any of the code. Is that OK?

.travis.yml
acdvorak: Yes, this is global, and I already added it for screenshot tests above (because the screenshot config overrides `before_install`) 😀

scripts/skip-unnecessary-travis-tests.sh
kfranqueiro: If that's why this is this way, then that seems reasonable, but maybe leave a comment on it

PR #3245, opened 2018-07-26
feat: update default npm export to ES5 js files

moog16:
Currently npm is setup to export /index.js by default. The index.js file is ES6 and less universally accepted by other developers. This PR is to support those developers that are on other versions of EcmaScript, by exporting by default the dist js file.

Refs #929

See also https://github.com/material-components/material-components-web-react/issues/107

BREAKING CHANGE: Anyone intending to build MDC Web's ES2015+ sources must directly import `@material/foo/index`. `@material/foo` will now resolve to UMD modules.


packages/mdc-tab/package.json
```diff
@@ -9,7 +9,6 @@
     "material design",
     "tab"
   ],
-  "main": "index.js",
```
moog16: removed this because this isn't a package. Developer should be installing mdc-tabs

kfranqueiro: This is incorrect. mdc-tab is part of the new tabs work being done in the feat/tabs/tabs branch. Users will be installing this as well as mdc-tab-bar, mdc-tab-scroller, and mdc-tab-indicator.

moog16: Ah gotcha...I saw the `export {MDCTabFoundation, MDCTab} from './tab';` line in mdc-tabs, and thought it to be this mdc-tab package. Just missing a `.` to become `../tab`

moog16:
I think we'll also need to do a similar change as this PR: https://github.com/material-components/material-components-web-react/pull/198

kfranqueiro:
That's precisely the kind of thing I'm concerned about. I think that corroborates my concern about needing to update any ES2015 examples in our READMEs, because they would need to explicitly point to `/index` now.

Incidentally, our modules themselves should already be referencing `/index` where necessary, since #2111 was a thing. Otherwise we'd potentially have problems even in the module code itself.

moog16:
Would we want to point to the `/index` or the `/dist/mdc.foo.js` files? I believe we would want to point to `/dist/mdc.foo`, since pointing to `/index` wouldn't fix the issue.

Currently our readmes don't point to `/index` now, so why would they need to now? I think what #1991 is referring to is that Webpack/node will default to whatever the package.json main's prop is set to.

kfranqueiro:
> Would we want to point to the `/index` or the `/dist/mdc.foo.js` files? I believe we would want to point to `/dist/mdc.foo`, since pointing to `/index` wouldn't fix the issue.
> 
> Currently our readmes don't point to `/index` now, so why would they need to now? I think what #1991 is referring to is that Webpack/node will default to whatever the package.json main's prop is set to.

Do transpilers honor `package.json`'s `main` property when resolving `import`s the way that node itself does? If they do, then all of our documented ES2015 usages would break, because they're all simply referencing the directory, not a specific file, and what they reference would change when we change `main`. This would also presumably break a large percentage of end-user code that consumes our ES2015 modules.

moog16:
I wrote up a blog post of my learnings of the benefits of having npm packages export ES5 instead of ES6.

https://medium.com/@moog16/rules-of-npm-packaging-16a40c5d7e30

packages/mdc-tab/package.json
kfranqueiro: This was never updated after my comment. Also, mdc-tab-bar, mdc-tab-scroller, and mdc-tab-indicator were added since this.

packages/mdc-top-app-bar/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.37.1",
   "description": "The Material Components for the web top app bar component",
   "license": "Apache-2.0",
+  "main": "mdc.topAppBar.js",
```
kfranqueiro: Missing `dist`?

packages/mdc-form-field/package.json
```diff
@@ -8,7 +8,7 @@
     "material design",
     "form"
   ],
-  "main": "index.js",
+  "main": "mdc.formField.js",
```
kfranqueiro: Missing `dist`

packages/mdc-linear-progress/package.json
```diff
@@ -3,6 +3,7 @@
   "description": "The Material Components for the web linear progress indicator component",
   "version": "0.38.0",
   "license": "MIT",
+  "main": "dist/linearProgress.js",
```
kfranqueiro: Missing `mdc.`

moog16:
Pushed changes and performed following:

Linux:
npm run test:unit (passes)
npm run start (chrome)
npm run dev (chrome)

Windows:
npm run test:unit (passes)
*npm run start (ie11 & edge)*
npm run dev (ie11 & edge)

`npm run start` is not working on this branch or master branch on Windows 10 - I don't think it is an issue, since there are some `PWD` and `$` in the start script. So all is good (regarding this PR)

kfranqueiro:
I apparently neglected to respond to @abhiomkar's comment earlier, though I had discussed it with Matt a few weeks ago.

I think specifying `"module"` could end up causing trouble. Partially because it's a proposal and its behavior isn't fully standardized, but in particular I've found at least 2 notes from different sources that give me pause about using it.

From [rollup's pkg.module docs](https://github.com/rollup/rollup/wiki/pkg.module):

> `pkg.module` will point to a module that has ES2015 module syntax but otherwise _only syntax features that the target environments support._

This is not true in our case, e.g. our reliance on Object.assign (which we use an additional babel plugin to transform).

From [a proposal for Node.js modules](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md#typical-usage):

> To opt into standard modules, packages can specify a `"module"` key in their `package.json`. New versions of Node.js will prefer the `"module"` to an existing `"main"` or `index.js`.

This could have an effect on how build toolchains process our code, if it were to cause them to pick up one _or_ the other depending on Node version.

docs/importing-js.md
```diff
@@ -12,21 +12,36 @@ Most components ship with Component / Foundation classes which are used to provi
...
+
+However, if you want to take advantage of tree-shaking and dependency sharing within MDC Web's code to reduce the size
+of your built assets, you will want to explicitly reference the package's `index.js`, which contains the ES2015+ source:
+
+```js
+import {MDCFoo, MDCFooFoundation} from '@material/foo/index';
+```
+
+Note that in this case, you must ensure your build toolchain is configured to process and transpile MDC Web's modules
+as well as your own. You will also need to include babel's `transform-object-assign` plugin for IE 11 support.
```
moog16: This piece of info should also be in the getting-started.md

https://github.com/material-components/material-components-web/blame/master/docs/getting-started.md#L222

scripts/verify-pkg-main.js
```diff
@@ -0,0 +1,66 @@
+/**
```
moog16: awesome script! Tested it and works nicely

docs/importing-js.md
kfranqueiro: Yeah, I had intended to look into the right place to add that to getting-started, thanks for reminding me.

kfranqueiro: I've fixed that in #3665, since that PR was already touching the lines around that point in the file. That way hopefully it'll merge cleanly the next time we sync this PR.

kfranqueiro:
Matt performed some tests with various build tools (e.g. browserify and webpack), and I tested this against our catalog and getting started guide. Between that and getting positive feedback from James on RMWC and Dominic on angular-mdc-web, I think we can merge this.

PR #3261, opened 2018-07-30
feat(menu-surface): Add ability to position a menu-surface by x/y

### LINKED ISSUE #2770: Support dynamically setting the anchoring element of an mdc-menu-surface
Also support anchoring to x, y coordinates for context menus

williamernest:
Completely fixes #2770. 
Adds the ability to specify an absolute x/y that the menu should open to instead of being based on an element's position in the DOM.

Updates/adds a couple unrelated tests for additional coverage.

### DIFF
--- a/packages/mdc-menu-surface/foundation.js
+++ b/packages/mdc-menu-surface/foundation.js
@@ -136,6 +136,8 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
     this.hoistedElement_ = false;
     /** @private {boolean} */
     this.isFixedPosition_ = false;
+    /** @private {{x: {number}, y: {number}}} */
+    this.position_ = {x: 0, y: 0};
   }
 
   init() {
@@ -178,14 +180,30 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
     this.anchorMargin_.left = typeof margin.left === 'number' ? margin.left : 0;
   }
 
+  /**
+   * Used to indicate if the menu-surface is hoisted to the body.
+   * @param {boolean} isHoisted
+   */
   setIsHoisted(isHoisted) {
-    this.close();
     this.hoistedElement_ = isHoisted;
   }
 
+  /**
+   * Used to set the menu-surface calculations based on a fixed position menu.
+   * @param {boolean} isFixedPosition
+   */
   setFixedPosition(isFixedPosition) {
-    this.close();
     this.isFixedPosition_ = isFixedPosition;
+  }
+
+  /**
+   * Sets the menu-surface position on the page.
+   * @param x
+   * @param y
+   */
+  setAbsolutePosition(x, y) {
+    this.position_.x = x;
+    this.position_.y = y;
   }
 
   /** @param {boolean} quickOpen */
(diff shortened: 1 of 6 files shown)


packages/mdc-menu-surface/foundation.js
```diff
@@ -178,16 +180,32 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
...
     this.isFixedPosition_ = isFixedPosition;
   }
 
+  /**
+   * Sets the menu-surface position on the page.
+   * @param {number} x
+   * @param {number} y
+   */
+  setAbsolutePosition(x, y) {
+    this.position_.x = isNaN(x) ? 0 : x;
```
acdvorak: Change `isNaN(x)` to `!Number.isFinite(x)`.

From [`isNaN()` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN):

> The `isNaN()` function determines whether a value is `NaN` or not. Note: coercion inside the isNaN function has **interesting** rules; you may alternatively want to use `Number.isNaN()`, as defined in ECMAScript 2015.

From [`Number.isFinite()` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite):

> In comparison to the global `isFinite()` function, `Number.isFinite()` doesn't forcibly convert the parameter to a number. This means only values of the type `number`, that are also finite, return `true`.

So it seems like `Number.isFinite()` is the closest to what you want: checking if the value is a usable x/y coordinate of type `number`.

packages/mdc-menu-surface/index.js
```diff
@@ -118,6 +118,16 @@ class MDCMenuSurface extends MDCComponent {
...
   }
 
+  /**
+   * Sets the absolute x/y position to position based on. Requires the menu to be hoisted.
+   * @param {number} x
+   * @param {number} y
+   */
+  setAbsolutePosition(x, y) {
+    this.foundation_.setAbsolutePosition(x, y);
+    this.setIsHoisted(true);
```
acdvorak: 💥 💥 💥

kfranqueiro: I'm a bit confused by `setHoisted` vs. `hoistMenuToBody`. We're setting `setHoisted` here but not actually enforcing that the menu is hoisted. Also, in the foundation, we seem to blindly assume that if it's hoisted, it's hoisted to the body.

Would it make more sense to _only_ expose a `hoistMenuToBody` method on the component? Or should we be supporting other hoisting scenarios?

williamernest: `hoistMenuToBody` is only on the component as a helper method. The `setHoisted` method is a way to for the foundation to know to calculate the position of the menu based on the body and not relative to an anchoring element. 

I think the current use cases we cover are enough. The menu is either positioned relative to an anchor element, or it has been hoisted to a position that is relative to the body. I think trying to cover the case where the menu is only hoisted to a certain level would be problematic.

packages/mdc-menu-surface/foundation.js
```diff
@@ -529,6 +556,17 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
   isOpen() {
     return this.isOpen_;
   }
+
+  /**
+   * Number.isNaN that doesn't force conversion to number type.
+   * Equivalent to Number.isNaN in ES2015, but is not included in IE11.
```
kfranqueiro: Did you mean `Number.isFinite`?

williamernest: Yes, just updated.

PR #3265, opened 2018-07-30
refactor(chips): Pass chip ids instead of foundations in events 

bonniezhou:
Passing an id rather than an `MDCChipFoundation` for Chip-to-ChipSet events allows users to do more with the component without needing to access foundations. For example, users can tell the `MDCChipSet` to select a chip by passing in the chip's ID. Previously, this would only be possible if users passed in an `MDCChipFoundation` or added a class to the DOM element.

This is necessary in MDC React, where the `ChipSet` wrapper needs to be able to set selection state without a DOM.

### DIFF
--- a/packages/mdc-chips/README.md
+++ b/packages/mdc-chips/README.md
@@ -248,22 +248,21 @@ Method Signature | Description
 `addClassToLeadingIcon(className: string) => void` | Adds a class to the leading icon element
 `removeClassFromLeadingIcon(className: string) => void` | Removes a class from the leading icon element
 `eventTargetHasClass(target: EventTarget, className: string) => boolean` | Returns true if target has className, false otherwise
-`notifyInteraction() => void` | Emits a custom event `MDCChip:interaction` denoting the chip has been interacted with
-`notifyTrailingIconInteraction() => void` | Emits a custom event `MDCChip:trailingIconInteraction` denoting the chip's trailing icon has been interacted with
-`notifyRemoval() => void` | Emits a custom event `MDCChip:removal` denoting the chip will be removed
+`notifyInteraction() => void` | Emits a custom event `MDCChip:interaction` denoting the chip has been interacted with*
+`notifyTrailingIconInteraction() => void` | Emits a custom event `MDCChip:trailingIconInteraction` denoting the chip's trailing icon has been interacted with*
+`notifyRemoval() => void` | Emits a custom event `MDCChip:removal` denoting the chip will be removed*
 `getComputedStyleValue(propertyName: string) => string` | Returns the computed property value of the given style property on the root element
 `setStyleProperty(propertyName: string, value: string) => void` | Sets the property value of the given style property on the root element
 
-> _NOTE_: The custom events emitted by `notifyInteraction` and `notifyTrailingIconInteraction` must pass along the target chip in its event `detail`, as well as bubble to the parent `mdc-chip-set` element.
-
-> _NOTE_: The custom event emitted by `notifyRemoval` must pass along the target chip and its root element in the event `detail`, as well as bubble to the parent `mdc-chip-set` element.
+> *_NOTE_: The custom events emitted by `notifyInteraction`, `notifyTrailingIconInteraction` and `notifyRemoval` must pass along the target chip's ID in its event `detail`, as well as bubble to the parent `mdc-chip-set` element.
 
 #### `MDCChipSetAdapter`
 
 Method Signature | Description
 --- | ---
 `hasClass(className: string) => boolean` | Returns whether the chip set element has the given class
-`removeChip(chip: MDCChip) => void` | Removes the chip object from the chip set
+`removeChip(chipId: string) => void` | Removes the chip with the given id from the chip set
+`setSelected(chipId: string, selected: boolean) => void` | Sets the selected state of the chip with the given id
 
 ### Foundations: `MDCChipFoundation` and `MDCChipSetFoundation`
 
@@ -284,7 +283,8 @@ Method Signature | Description
 
 Method Signature | Description
 --- | ---
-`select(chipFoundation: MDCChipFoundation) => void` | Selects the given chip
-`deselect(chipFoundation: MDCChipFoundation) => void` | Deselects the given chip
+`select(chipId: string) => void` | Selects the chip with the given id
+`deselect(chipId: string) => void` | Deselects the chip with the given id
+`toggleSelect(chipId: string) => void` | Toggles selection of the chip with the given id
 `handleChipInteraction(evt: Event) => void` | Handles a custom `MDCChip:interaction` event on the root element
 `handleChipRemoval(evt: Event) => void` | Handles a custom `MDCChip:removal` event on the root element
--- a/packages/mdc-chips/README.md
+++ b/packages/mdc-chips/README.md
@@ -209,24 +209,27 @@ To use the `MDCChip` and `MDCChipSet` classes, [import](../../docs/importing-js.
 
 #### `MDCChip`
 
-Method Signature | Description
---- | ---
-`isSelected() => boolean` | Proxies to the foundation's `isSelected` method
-`beginExit() => void` | Proxies to the foundation's `beginExit` method
-
 Property | Value Type | Description
 --- | --- | ---
-`foundation` | MDCChipFoundation | The foundation
-`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods
+`id` | string | Unique identifier on the chip*
+`selected` | Boolean | Proxies to the foundation's `isSelected`/`setSelected` methods
+`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods**
 `ripple` | `MDCRipple` | The `MDCRipple` instance for the root element that `MDCChip` initializes
 
->_NOTE_: If `shouldRemoveOnTrailingIconClick` is set to false, you must manually call `beginExit()` on the chip to remove it.
+> *_NOTE_: By default, this will be the same as the `id` attribute on the root element. If an `id` is not provided, a random one will be generated.
+
+> **_NOTE_: If `shouldRemoveOnTrailingIconClick` is set to false, you must manually call `beginExit()` on the chip to remove it.
+
+Method Signature | Description
+--- | ---
+`beginExit() => void` | Proxies to the foundation's `beginExit` method
 
 #### `MDCChipSet`
 
 Method Signature | Description
 --- | ---
 `addChip(chipEl: Element) => void` | Adds a new `MDCChip` instance to the chip set based on the given `mdc-chip` element
+`getSelectedChipIds() => boolean` | Returns an array of the IDs of all selected chips
 
 Property | Value Type | Description
 --- | --- | ---
@@ -286,5 +289,6 @@ Method Signature | Description
 `select(chipId: string) => void` | Selects the chip with the given id
 `deselect(chipId: string) => void` | Deselects the chip with the given id
 `toggleSelect(chipId: string) => void` | Toggles selection of the chip with the given id
+`getSelectedChipIds() => boolean` | Returns an array of the IDs of all selected chips
 `handleChipInteraction(evt: Event) => void` | Handles a custom `MDCChip:interaction` event on the root element
 `handleChipRemoval(evt: Event) => void` | Handles a custom `MDCChip:removal` event on the root element
--- a/packages/mdc-chips/chip-set/foundation.js
+++ b/packages/mdc-chips/chip-set/foundation.js
@@ -18,7 +18,7 @@
 import MDCFoundation from '@material/base/foundation';
 import MDCChipSetAdapter from './adapter';
 // eslint-disable-next-line no-unused-vars
-import {MDCChipFoundation, MDCChipInteractionEventType} from '../chip/foundation';
+import {MDCChipInteractionEventType} from '../chip/foundation';
 import {strings, cssClasses} from './constants';
 
 /**
@@ -60,6 +60,14 @@ class MDCChipSetFoundation extends MDCFoundation {
      * @private {!Array<string>}
      */
     this.selectedChipIds_ = [];
+  }
+
+  /**
+   * Returns an array of the IDs of all selected chips.
+   * @return {!Array<string>}
+   */
+  getSelectedChipIds() {
+    return this.selectedChipIds_;
   }
 
   /**
@@ -111,7 +119,7 @@ class MDCChipSetFoundation extends MDCFoundation {
    */
   handleChipInteraction(evt) {
     const {chipId} = evt.detail;
-    this.handleSelect(chipId);
+    this.toggleSelect(chipId);
   }
 
   /**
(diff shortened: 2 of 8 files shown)


packages/mdc-chips/chip-set/foundation.js
```diff
@@ -99,14 +118,8 @@ class MDCChipSetFoundation extends MDCFoundation {
    * @private
    */
   handleChipInteraction(evt) {
```
rlfriedman: Here or in another change, can we make this and handleChipRemoval public?

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -55,42 +56,60 @@ class MDCChipSetFoundation extends MDCFoundation {
...
+   * Selects the chip with the given id. Deselects all other chips if the chip set is of the choice variant.
+   * @param {string} chipId
    */
-  deselect(chipFoundation) {
-    const index = this.selectedChips_.indexOf(chipFoundation);
-    if (index >= 0) {
-      this.selectedChips_.splice(index, 1);
+  select(chipId) {
+    if (this.adapter_.hasClass(cssClasses.CHOICE)) {
+      const selectedChipId = this.selectedChipIds_.length > 0 && this.selectedChipIds_[0];
```
rlfriedman: Closure is not a fan of the typing here as it thinks selectedChipId could evaluate to boolean or string (although I don't think it would ever actually make it into the setSelected method as a boolean). Maybe just only go inside this if when this.adapter_.hasClass(cssClasses.CHOICE) && this.selectedChipIds_.length > 0 ?

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -99,14 +118,8 @@ class MDCChipSetFoundation extends MDCFoundation {
...
   handleChipInteraction(evt) {
-    const chipFoundation = evt.detail.chip.foundation;
-    if (this.adapter_.hasClass(cssClasses.CHOICE) || this.adapter_.hasClass(cssClasses.FILTER)) {
-      if (chipFoundation.isSelected()) {
-        this.deselect(chipFoundation);
-      } else {
-        this.select(chipFoundation);
-      }
-    }
+    const {chipId} = evt.detail;
```
rlfriedman: Since the MDCChipInteractionEventType has changed, I think we need to update the typedef in MDCChipFoundation to reflect that it now consists of a chipId instead of a foundation.

packages/mdc-chips/README.md
```diff
@@ -209,24 +209,27 @@ To use the `MDCChip` and `MDCChipSet` classes, [import](../../docs/importing-js.
...
 --- | --- | ---
-`foundation` | MDCChipFoundation | The foundation
-`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods
+`id` | string | Unique identifier on the chip*
+`selected` | Boolean | Proxies to the foundation's `isSelected`/`setSelected` methods
+`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods**
 `ripple` | `MDCRipple` | The `MDCRipple` instance for the root element that `MDCChip` initializes
 
->_NOTE_: If `shouldRemoveOnTrailingIconClick` is set to false, you must manually call `beginExit()` on the chip to remove it.
+> *_NOTE_: By default, this will be the same as the `id` attribute on the root element. If an `id` is not provided, a random one will be generated.
```
kfranqueiro: I would suggest escaping the `*`s (`\*`) in front of these notes and in the tables to avoid formatting bugs. It doesn't appear to cause any right now, but `*` is normally interchangeable with `_` in markdown (and they're currently fooling github's diff view as well as Sublime Text's markdown syntax highlighting).

packages/mdc-chips/README.md
```diff
@@ -209,24 +209,27 @@ To use the `MDCChip` and `MDCChipSet` classes, [import](../../docs/importing-js.
...
+
+> **_NOTE_: If `shouldRemoveOnTrailingIconClick` is set to false, you must manually call `beginExit()` on the chip to remove it.
+
+Method Signature | Description
```
kfranqueiro: Why was this moved below properties? chipset's methods are still listed above its properties immediately below this section.

packages/mdc-chips/README.md
```diff
@@ -248,22 +251,21 @@ Method Signature | Description
...
+`notifyTrailingIconInteraction() => void` | Emits a custom event `MDCChip:trailingIconInteraction` denoting the chip's trailing icon has been interacted with*
+`notifyRemoval() => void` | Emits a custom event `MDCChip:removal` denoting the chip will be removed*
 `getComputedStyleValue(propertyName: string) => string` | Returns the computed property value of the given style property on the root element
 `setStyleProperty(propertyName: string, value: string) => void` | Sets the property value of the given style property on the root element
 
-> _NOTE_: The custom events emitted by `notifyInteraction` and `notifyTrailingIconInteraction` must pass along the target chip in its event `detail`, as well as bubble to the parent `mdc-chip-set` element.
-
-> _NOTE_: The custom event emitted by `notifyRemoval` must pass along the target chip and its root element in the event `detail`, as well as bubble to the parent `mdc-chip-set` element.
+> *_NOTE_: The custom events emitted by `notifyInteraction`, `notifyTrailingIconInteraction` and `notifyRemoval` must pass along the target chip's ID in its event `detail`, as well as bubble to the parent `mdc-chip-set` element.
```
kfranqueiro: "in its event `detail`" -> "via `event.detail.chipId`"

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -55,69 +56,79 @@ class MDCChipSetFoundation extends MDCFoundation {
...
-   * Deselects the given chip.
-   * @param {!MDCChipFoundation} chipFoundation
+   * Selects the chip with the given id. Deselects all other chips if the chip set is of the choice variant.
+   * @param {string} chipId
    */
-  deselect(chipFoundation) {
-    const index = this.selectedChips_.indexOf(chipFoundation);
-    if (index >= 0) {
-      this.selectedChips_.splice(index, 1);
+  select(chipId) {
```
kfranqueiro: I'm pretty sure this API creates a bug if called repeatedly for the same choice or filter chip, because it's going to repeatedly push the same chip ID into the array multiple times (without ever clearing the array, even for choice chips).

This won't happen from user interaction since `toggleSelect` will be called which will never call `select` for an already-selected chip, but `select` is also a public API that could be called programmatically.

We should add a test for this bug and fix it. TBH, the earlier behavior seemed simpler - unconditionally deselect everything for choice chips.

packages/mdc-chips/chip-set/index.js
```diff
@@ -96,16 +96,32 @@ class MDCChipSet extends MDCComponent {
     this.chips.push(this.chipFactory_(chipEl));
   }
 
+  /**
+   * Returns an array of the IDs of all selected chips.
+   * @return {!Array<string>}
+   */
+  getSelectedChipIds() {
```
kfranqueiro: Any thoughts of making this `get selectedChipIds` instead?

packages/mdc-chips/chip/index.js
```diff
@@ -172,10 +176,11 @@ class MDCChip extends MDCComponent {
...
       },
       eventTargetHasClass: (target, className) => target.classList.contains(className),
-      notifyInteraction: () => this.emit(strings.INTERACTION_EVENT, {chip: this}, true /* shouldBubble */),
+      notifyInteraction: () => this.emit(strings.INTERACTION_EVENT, {chipId: this.id}, true /* shouldBubble */),
       notifyTrailingIconInteraction: () => this.emit(
-        strings.TRAILING_ICON_INTERACTION_EVENT, {chip: this}, true /* shouldBubble */),
-      notifyRemoval: () => this.emit(strings.REMOVAL_EVENT, {chip: this, root: this.root_}, true /* shouldBubble */),
+        strings.TRAILING_ICON_INTERACTION_EVENT, {chipId: this.id}, true /* shouldBubble */),
+      notifyRemoval: () =>
+        this.emit(strings.REMOVAL_EVENT, {chipId: this.id, root: this.root_}, true /* shouldBubble */),
```
kfranqueiro: Is `root` needed? I don't see it used or documented

test/unit/mdc-chips/mdc-chip-set.foundation.test.js
```diff
@@ -34,118 +34,75 @@ test('exports cssClasses', () => {
...
-  td.verify(chipA.foundation.setSelected(true));
-  assert.equal(foundation.selectedChips_.length, 1);
+  foundation.select('chipA');
+  td.verify(mockAdapter.setSelected('chipA', true));
+  assert.equal(foundation.getSelectedChipIds().length, 1);
 });
 
-test('in choice chips, #handleChipInteraction deselects chip if another chip is selected', () => {
-  const {foundation, mockAdapter, chipA, chipB} = setupTest();
+test('in choice chips, #select deselects chip if another chip is selected', () => {
```
kfranqueiro: Add another test for `in choice chips, #select does not affect selected state if the same chip is selected` and test that `getSelectedChipIds().length` is 1

test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -115,7 +97,15 @@ test('#adapter.removeChip removes the chip object from the chip set', () => {
   const root = getFixture();
   const component = new MDCChipSet(root, undefined, (el) => new FakeChip(el));
   const chip = component.chips[0];
-  component.getDefaultFoundation().adapter_.removeChip(chip);
+  component.getDefaultFoundation().adapter_.removeChip(chip.id);
   assert.equal(component.chips.length, 2);
   td.verify(chip.destroy());
 });
+
+test('#adapter.setSelected sets select on chip object', () => {
```
kfranqueiro: Nit: select -> selected

test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -43,7 +43,7 @@ test('attachTo returns an MDCChipSet instance', () => {
 class FakeChip {
   constructor() {
     this.destroy = td.func('.destroy');
-    this.isSelected = td.func('.isSelected');
+    this.selected = td.func('.selected');
```
kfranqueiro: Is this going to work? `chip.selected` is not invoked as a function...

test/unit/mdc-chips/mdc-chip-set.foundation.test.js
```diff
@@ -34,118 +34,75 @@ test('exports cssClasses', () => {
...
-  assert.equal(foundation.selectedChips_.length, 1);
+  foundation.select('chipA');
+  td.verify(mockAdapter.setSelected('chipB', false));
+  td.verify(mockAdapter.setSelected('chipA', true));
+  assert.equal(foundation.getSelectedChipIds().length, 1);
 });
 
-test('in filter chips, #handleChipInteraction selects multiple chips', () => {
-  const {foundation, mockAdapter, chipA, chipB} = setupTest();
+test('in filter chips, #select selects multiple chips', () => {
```
kfranqueiro: Add another test for calling `select` on a filter chip set with the same ID multiple times and checking `getSelectedChipIds().length`

packages/mdc-chips/chip/index.js
```diff
@@ -59,6 +61,7 @@ class MDCChip extends MDCComponent {
   }
 
   initialize() {
+    this.id = this.root_.id || this.root_.innerText.slice(0, -1) + Math.random();
```
kfranqueiro: `innerText` seems unwieldy for this, and `Math.random()` is going to produce incomprehensible decimal numbers. A better idea might be to use a module-local counter:

```js
let idCounter = 0;

...

initialize() {
  this.id = this.root_.id || `mdc-chip-${++idCounter}`;
```

Also, should we set the `id` attribute on the element too?

Tangentially, I'm not sure what the purpose of `slice(0, -1)` was (FYI `trim` should work in all supported browsers).

packages/mdc-chips/README.md
bonniezhou: Oh for some reason I thought it was the other way around...I'll move it back.

packages/mdc-chips/chip-set/foundation.js
```diff
@@ -55,69 +56,83 @@ class MDCChipSetFoundation extends MDCFoundation {
...
+    if (this.selectedChipIds_.indexOf(chipId) >= 0) {
+      return;
+    }
+
+    if (this.adapter_.hasClass(cssClasses.CHOICE) && this.selectedChipIds_.length > 0) {
+      const selectedChipId = this.selectedChipIds_[0];
+      if (selectedChipId != chipId) {
```
kfranqueiro: This should always be true now, right?

packages/mdc-chips/chip/index.js
bonniezhou: To use a counter, the id will be set in the chip set. I used the id of the chip set to minimize id collisions in case there's more than one chip set. Do you think that would be secure enough?

(And thanks for the tip, `trim` was what I was looking for 🙂)

packages/mdc-chips/chip/index.js
bonniezhou: Yep, it's needed for the user to remove the DOM node. Added it back to the README

packages/mdc-chips/chip/index.js
bonniezhou: I updated the README to recommend setting an id on the chip set

kfranqueiro: If you initialize a counter on every chip set, yes, you'd have to do this, and as implemented, this will break if the user ever doesn't set a chip set ID, so you've effectively just punted the issue up one component.

My original thought, if we don't care that much about the counter, was to scope the counter to the module (not an instance) and share/increment it across instances, which doesn't require distinguishing with an ID on the chip set. The autogenerated IDs might make a little less sense since they'll strictly increase across instances, but if the user isn't specifying IDs themselves anyway, does it really matter?

bonniezhou: Updated!

packages/mdc-chips/README.md
```diff
@@ -211,22 +231,25 @@ To use the `MDCChip` and `MDCChipSet` classes, [import](../../docs/importing-js.
...
 --- | --- | ---
-`foundation` | MDCChipFoundation | The foundation
-`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods
+`id` | string | Unique identifier on the chip\*
+`selected` | Boolean | Proxies to the foundation's `isSelected`/`setSelected` methods
+`shouldRemoveOnTrailingIconClick` | Boolean | Proxies to the foundation's `getShouldRemoveOnTrailingIconClick`/`setShouldRemoveOnTrailingIconClick` methods\*\*
 `ripple` | `MDCRipple` | The `MDCRipple` instance for the root element that `MDCChip` initializes
 
->_NOTE_: If `shouldRemoveOnTrailingIconClick` is set to false, you must manually call `beginExit()` on the chip to remove it.
+> \*_NOTE_: This will be the same as the `id` attribute on the root element. If an `id` is not provided, a pseudo-random one will be generated.
```
kfranqueiro: pseudo-random -> unique

Issue #3256, opened 2018-07-30
Dialog backdrop not hiding when closed immediately after opening


Alan-Liang:
** PLEASE READ THIS BEFORE FILING AN ISSUE **

## Bugs

Follow the template below **and include a codepen** to ensure the quickest and most accurate response to your issue.

### What MDC Web Version are you using?

0.37.1
### What browser(s) is this bug affecting?

`Mozilla/5.0 (Windows NT 5.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.112 Safari/537.36`;

`Mozilla/5.0 (X11; Linux armv7l) AppleWebKit/537.36 (KHTML, like Gecko) Raspbian Chromium/65.0.3325.181 Chrome/65.0.3325.181 Safari/537.36`
### What OS are you using?

`Windows XP SP3`, `Linux raspberrypi 4.14.52-v7+ #1123`
### What are the steps to reproduce the bug?

1. Open [codepen](https://codepen.io/Alan-Liang/pen/gjorya) / Open a dialog and close it (using script) fast after open.
2. Check if the link is clickable.

### What is the expected behavior?

The link will be clickable.

### What is the actual behavior?

The link isn't clickable, which indicates that there's a layer on it.
### Any other information you believe would be useful?

The class `mdc-dialog--animating` remains on the `aside.mdc-dialog` element.

acdvorak:
Thanks so much for reporting this issue!

I'm guessing it's due to how we're using the `transitionend` event. Definitely a bug!

PR #3267, opened 2018-07-31
fix(dialog): Fixes transitionend event not always being called

williamernest:
fixes: #3256 
When open/close is called to quickly, the `transitionend` event is not always fired. This causes the `mdc-dialog--animating` class to remain and the `backdrop` to still mask the body. Converting the `transitionend` event to a simple timeout resolves this issue.  

Also, it removes the unused adapter methods, but is not a breaking change.

The first commit incorrectly states this is for drawer.


packages/mdc-dialog/foundation.js
```diff
@@ -63,21 +61,19 @@ export default class MDCDialogFoundation extends MDCFoundation {
...
-      this.adapter_.deregisterInteractionHandler('click', this.componentClickHandler_);
-      this.adapter_.untrapFocusOnSurface();
-      this.adapter_.deregisterTransitionEndHandler(this.transitionEndHandler_);
-      this.adapter_.removeClass(MDCDialogFoundation.cssClasses.ANIMATING);
-      this.adapter_.removeClass(MDCDialogFoundation.cssClasses.OPEN);
-      this.enableScroll_();
+      this.close();
     }
+    // Final cleanup of animating class in case the timer has not completed.
+    this.adapter_.removeClass(MDCDialogFoundation.cssClasses.ANIMATING);
```
acdvorak: 💯

PR #3288, opened 2018-08-02
test(elevation): Add screenshot tests

acdvorak:


test/screenshot/spec/mdc-elevation/classes/baseline.html
```diff
@@ -0,0 +1,65 @@
...
+    <link rel="stylesheet" href="../../../out/mdc.elevation.css">
+    <link rel="stylesheet" href="../../../out/mdc.typography.css">
+    <link rel="stylesheet" href="../../../out/spec/fixture.css">
+    <link rel="stylesheet" href="../../../out/spec/mdc-elevation/fixture.css">
+  </head>
+
+  <body class="test-container mdc-typography">
+    <main class="test-viewport test-viewport--mobile">
+      <div class="test-layout">
+        <div class="test-cell test-cell--elevation"><span class="mdc-elevation--z0">0</span></div>
```
kfranqueiro: Do you think it might be a good idea to apply a class to each span that gives it a fixed size? e.g. in case of text rendering flakes between browsers/runs...

acdvorak: `test-cell--elevation` already sets a fixed width/height for that exact reason 😁

kfranqueiro: But that's the cell, not the thing being elevated.

My thought was that if the text causes flakes, that will cause flakes on the box shadows as well.

acdvorak: Ahh, I see what you're saying. Good idea - I'll do that!

acdvorak: Done

PR #3291, opened 2018-08-03
test(typography): Add screenshot tests

acdvorak:


kfranqueiro:
IE and Edge are apparently doing weird things on the mixin test page... cc @abhiomkar in case this is actually an issue w/ the baseline mixins

acdvorak:
Edge seems to be really flaky with font rendering on these pages:

- https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/08/03/17_29_13_502/report/report.html
- https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/08/03/17_42_52_220/report/report.html

The existing retry logic (which simply reloads the page and waits an extra second) doesn't help.

Either Edge's font rendering is inherently flaky, or we're getting slightly different versions of Windows that render fonts differently 🤷‍♀️

acdvorak:
The strange IE/Edge behavior was fixed in #3297.

Edge is still flaky with its text rendering, however.

PR #3297, opened 2018-08-03
fix(typography): Reverted baseline mixin to use display inline-block

abhiomkar:
Flexbox works differently with pseudo elements on Chrome/FF vs IE/Edge. Reverted mixin to use display inline-block.

BREAKING CHANGE: Updates to typography baseline mixin which requires the text element to strip the white-space. Helper text and MDC List two line text should strip the white-space.

### DIFF
--- a/demos/typography.html
+++ b/demos/typography.html
@@ -84,16 +84,9 @@
 
       <h6 class="mdc-typography--headline6 demo-typography--heading-baseline">Baseline</h6>
       <section class="demo-typography--section-baseline">
-        <h2 class="demo-typography__title mdc-typography--headline6">
-          Our Changing Planet
-        </h2>
-        <h3 class="demo-typography__subtitle mdc-typography--subtitle2">
-          by Kurt Wagner
-        </h3>
-        <div class="demo-typography__body mdc-typography--body2">
-          Visit ten places on our planet that are undergoing the biggest changes today.
-        </div>
-
+        <h2 class="demo-typography__title mdc-typography--headline6">Our Changing Planet</h2>
+        <h3 class="demo-typography__subtitle mdc-typography--subtitle2">by Kurt Wagner</h3>
+        <div class="demo-typography__body mdc-typography--body2">Visit ten places on our planet that are undergoing the biggest changes today.</div>
         <line class="demo-typography-line-1">+34px</line>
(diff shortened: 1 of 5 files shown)


demos/typography.html
```diff
@@ -84,16 +84,9 @@ <h6 class="mdc-typography--subtitle2">Subtitle 2</h6>
...
-          by Kurt Wagner
-        </h3>
-        <div class="demo-typography__body mdc-typography--body2">
-          Visit ten places on our planet that are undergoing the biggest changes today.
-        </div>
-
+        <h2 class="demo-typography__title mdc-typography--headline6">Our Changing Planet</h2>
+        <h3 class="demo-typography__subtitle mdc-typography--subtitle2">by Kurt Wagner</h3>
+        <div class="demo-typography__body mdc-typography--body2">Visit ten places on our planet that are undergoing the biggest changes today.</div>
         <line class="demo-typography-line-1">+34px</line>
```
acdvorak: Why are we using a non-standard `<line>` element instead of a `<div>` or something?

packages/mdc-list/README.md
```diff
@@ -77,6 +77,8 @@ in the double line list style as defined by
 </ul>
 ```
 
+> NOTE: Make sure there are no white-space characters before primary and secondary text content.
```
acdvorak: Is there another way to implement these mixins that doesn't choke on whitespace?

I suspect a lot of people are going to run into this, and it's going to take them a while to figure out why their UI suddenly broke.

kfranqueiro: I would like there to be a way, but flex-box was the one way we figured...but that causes the current problem where IE and Edge don't work correctly.

The odd bit is if anything, I'd say IE and Edge are working as I'd expect all browsers to. It's more of a mystery to me why switching to flex still worked in the others.

abhiomkar: Since the baseline struts (before & after pseudo elements) use `display: inline-block` it leaves [extra space](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) around before the text when white-space characters are inserted. We avoided this problem by using flex box before but IE / Edge renders them differently on pseudo elements.

It shouldn't create any issues for users who are using frameworks while vanilla component users should use this with caution.

PR #3309, opened 2018-08-07
feat(slider): Add tick marks to change color when active vs inactive

EstebanG23:


packages/mdc-slider/index.js
```diff
@@ -163,6 +163,20 @@ class MDCSlider extends MDCComponent {
           this.ripple_.deactivate();
         },
         isRTL: () => getComputedStyle(this.root_).direction === 'rtl',
+        updateTickMarkClasses: (currentTickMark) => {
```
bonniezhou: Generally we want the adapter methods to have as little logic as possible, so I think we should move the `if`s and`for`s to the foundation and have the adapter provide 4 methods:
```
tickMarkHasClass: (tickMark, className) => tickMark.classList.contains(className),
tickMarkAddClass: (tickMark, className) => tickMark.classList.add(className),
tickMarkRemoveClass: (tickMark, className) => tickMark.classList.remove(className),
getTickMarks: () => this.tickMarkSet_.children
```

kfranqueiro: I agree with this, but can we name them `addTickMarkClass`, etc. instead to be consistent with what we've done for other adapters in cases like this?

EstebanG23: Sounds good! Updated

test/unit/mdc-slider/mdc-slider.test.js
```diff
@@ -400,6 +400,52 @@ test('adapter#setLastTickMarkStyleProperty sets a style property on the last tic
...
+
+  assert.isFalse(tickMark.classList.contains('mdc-slider--test'));
+});
+
+test('adapter#getTickMarks returns the tickMarkSet children', () => {
+  const {root, component} = setupTest();
+  component.setUpTickMarks_();
+  const tickMarkset = root.querySelector('.mdc-slider__tick-mark-set').children;
+
+  assert.equal(tickMarkset, component.getDefaultFoundation().adapter_.getTickMarks());
```
bonniezhou: nits:
1. rename `tickMarkset` to `tickMarkSet`
2. Swap `tickMarkSet` and `component.getDefaultFoundation().adapter_.getTickMarks()` because the convention is to put the actual value (the one we're testing) first and the expected value second.

EstebanG23: Thank you! Updated

packages/mdc-slider/README.md
```diff
@@ -224,6 +228,8 @@ instance that was affected.
 | Method Signature | Description |
 | --- | --- |
 | `layout() => void` | Same as layout() detailed within the component methods table. Does the majority of the work; the component's layout method simply proxies to this. |
+| `calculateNumberOfTickMarks() => number` | Calculates the number of tick marks for discrete slider. |
+| `adjustLastTickMark() => void` | Adjusts the last tick mark style for discrete sliders. |
```
bonniezhou: This takes in 1 parameter

Issue #3277, opened 2018-08-01
Remove unused adapter methods from icon-button


moog16:
Icon button still has methods in the adapter that were removed in PR https://github.com/material-components/material-components-web/pull/3165

See lines:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-icon-button/adapter.js#L50

PR #3310, opened 2018-08-07
feat(icon-button): Add SVG support

### LINKED ISSUE #3107: mdc-icon-button: Restructure DOM to simplify and enable SVG for toggle
This issue was split out from #2918 (which also identified documentation/cleanup issues).

From @gjdev:

I don't really understand why the toggled variant works the way it does, with iconOn/iconOff properties that set an icon-font class or ligature. This needs the earlier workarounds for pseudo elements that are difficult to document/grasp. It also excludes the use of SVG icons. Imho a much better approach would be to have an icon-button-toggle DOM with two embedded icons, one marked as the 'on' icon, one as the 'off' icon. This would work with all types of icon-fonts without workarounds for pseudo elements, and would also work with SVG icons. It's also much easier to document for all those different use cases, as the usage would be exactly the same for any kind of icon.

### LINKED ISSUE #3154: Icon button does not have documentation showing SVG support
It seems like the regular (non-toggle) icon button can support SVG icons...but there is no documentation. Other packages in MDC Web have this kind of documentation, like button:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-button#icons

williamernest:
fixes: #3277 , resolves #3154, resolves #3107
BREAKING CHANGE: Removes the previous `data` attributes and no longer dynamically changes the label. Allows developers to add both elements to the button, with one indicated as the `on` state by using a `data-toggle-on` attribute. State is now changed by adding/removing the `mdc-icon-button--on` class to the `mdc-icon-button` element. All icon elements should have the `mdc-icon-button__icon` class.

### DIFF
--- a/packages/mdc-icon-button/README.md
+++ b/packages/mdc-icon-button/README.md
@@ -70,19 +70,20 @@ iconButtonRipple.unbounded = true;
 
 ### Icon Button Toggle
 
-The icon button can be used to toggle between an on and off icon. To style an icon button as an icon button toggle, add the
-`data-toggle-on` and `data-toggle-off` attributes to the `mdc-icon-button` element. Then instantiate an `MDCIconButtonToggle` on the root element.
+The icon button can be used to toggle between an on and off icon. To style an icon button as an icon button toggle, add
+both icons as child elements and place the `data-toggle-on` attribute on the icon that should be on. If the button
+should be initialized in the "on" state, then add the `mdc-icon-button--on` class to the parent `button`. Then
+instantiate an `MDCIconButtonToggle` on the root element.
 
 ```html
 <button id="add-to-favorites"
-   class="mdc-icon-button material-icons"
+   class="mdc-icon-button"
    aria-label="Add to favorites"
    aria-hidden="true"
-   aria-pressed="false"
-   data-toggle-on-content="favorite"
-   data-toggle-on-label="Remove from favorites"
-   data-toggle-off-content="favorite_border"
-   data-toggle-off-label="Add to favorites">favorite_border</button>
+   aria-pressed="false">
+   <i class="material-icons mdc-icon-button__icon">favorite</i>
+   <i class="material-icons mdc-icon-button__icon" data-toggle-on>favorite_border</i>
+</button>
 ```
 
 ```js
@@ -91,38 +92,38 @@ var toggleButton = new mdc.iconButton.MDCIconButtonToggle(document.getElementByI
 
 #### Icon Button Toggle States
 
-Note the use of `data-toggle-*` properties in the above examples. When an MDCIconButtonToggle
-instance is toggled, it looks at these data attributes to determine how to update the element. This is what
-allows MDCIconButtonToggle to be so flexible. The `data-toggle-on-*` properties will be used when the is
-MDCIconButtonToggle is toggled on, and vice versa for `data-toggle-off-*`.
+Note the use of `data-toggle-on` attribute in the above examples. When an MDCIconButtonToggle
+instance is toggled, it looks at this data attribute to determine which element is the on element. This is what
+allows MDCIconButtonToggle to be so flexible. The `data-toggle-on` properties will be as the `on` value, and the
+`mdc-icon-button--on` class will be used to indicate the `on` state.
 
 Attribute | Description
 --- | ---
-`data-toggle-<TOGGLE STATE>-label` | The value to apply to the element's "aria-label" attribute.
-`data-toggle-<TOGGLE STATE>-content` | The text content to set on the element. Note that if an inner icon is used, the text content will be set on that element instead.
-`data-toggle-<TOGGLE STATE>-class` | A CSS class to apply to the icon element. The same rules regarding inner icon elements described for `content` apply here as well.
+`data-toggle-on` | Used to indicate which element in an icon button toggle is the `on` value.
 
-#### Icon Button Toggle with Font Awesome
+#### Icon Button Toggle with SVG
 
-The icon button toggle can be used with other font libraries such as Font Awesome that use an inner icon element.
+The icon button toggle can be used with SVGs.
 
 ```html
 <button id="star-this-item"
-   class="mdc-icon-button"
+   class="mdc-icon-button mdc-icon-button--on"
    aria-label="Unstar this item"
    aria-hidden="true"
-   aria-pressed="true"
-   data-toggle-on-class="fa-star"
-   data-toggle-on-label="Unstar this item"
-   data-toggle-off-class="fa-star-o"
-   data-toggle-off-label="Star this item"><i class="fa fa-2x fa-star"></i></button>
+   aria-pressed="true">
+   <svg class="mdc-icon-button__icon">
+     ...
+   </svg>
+   <svg class="mdc-icon-button__icon" data-toggle-on>
+     ...
+  </svg>
+</button>
 ```
 
 ### Icons
 
-The icon button can be used with a standard icon library such as Material Icons or Font Awesome, or with an `svg`.
-The icon button toggle should only be used with an standard icon library. We recommend you use 
-[Material Icons](https://material.io/tools/icons) from Google Fonts.
+The icon button can be used with a standard icon library such as Material Icons, or with an `svg`.
+We recommend you use [Material Icons](https://material.io/tools/icons) from Google Fonts.
 
 ### Disabled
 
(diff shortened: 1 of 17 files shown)


packages/mdc-icon-button/README.md
```diff
@@ -91,38 +92,54 @@ var toggleButton = new mdc.iconButton.MDCIconButtonToggle(document.getElementByI
...
 #### Icon Button Toggle States
 
-Note the use of `data-toggle-*` properties in the above examples. When an MDCIconButtonToggle
-instance is toggled, it looks at these data attributes to determine how to update the element. This is what
-allows MDCIconButtonToggle to be so flexible. The `data-toggle-on-*` properties will be used when the is
-MDCIconButtonToggle is toggled on, and vice versa for `data-toggle-off-*`.
+Note the use of `data-toggle-on` attribute in the above examples. This attribute indicates which element is the
+on element. When the `mdc-icon-button--on` class is present, the element with the `data-toggle-on` attribute
+will be shown and the other element will be hidden. When the `mdc-icon-button--on` class is not present, the element
+with the `data-toggle-on` attribute will be hidden and the other element will beshown. This is what allows
```
kfranqueiro: beshown -> be shown

packages/mdc-icon-button/README.md
```diff
@@ -91,38 +92,54 @@ var toggleButton = new mdc.iconButton.MDCIconButtonToggle(document.getElementByI
...
+MDCIconButtonToggle to be so flexible.
```
kfranqueiro: "This is what allows MDCIconButtonToggle to be so flexible." -> "This structure allows MDCIconButton to be flexible enough to support both icon fonts and images/SVGs."

packages/mdc-icon-button/README.md
```diff
@@ -140,6 +157,7 @@ cannot be disabled. Disabled icon buttons cannot be interacted with and have no
 CSS Class | Description
 --- | ---
 `mdc-icon-button` | Mandatory.
+`mdc-icon-button--on` | Used to indicate the toggle button icon that is the not-selected option.
```
kfranqueiro: This doesn't sound right... This class is applied to the root element when the toggle should be in the "on" state, right?

packages/mdc-icon-button/README.md
```diff
@@ -171,9 +189,8 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
 
 Method Signature | Description
 --- | ---
-`addClass(className: string) => void` | Adds a class to the root element, or the inner icon element.
-`removeClass(className: string) => void` | Removes a class from the root element, or the inner icon element.
-`setText(text: string) => void` | Sets the text content of the root element, or the inner icon element.
+`addClass(className: string) => void` | Adds a class to a icon element.
```
kfranqueiro: Why were these docs changed? Doesn't this still apply to the root element?

packages/mdc-icon-button/adapter.js
```diff
@@ -44,31 +44,16 @@ class MDCIconButtonToggleAdapter {
...
-  deregisterInteractionHandler(type, handler) {}
-
-  /** @param {string} text */
-  setText(text) {}
-
-  /**
-   * @param {string} name
-   * @return {string}
-   */
-  getAttr(name) {}
```
kfranqueiro: You removed getAttr but it's still in the docs

packages/mdc-icon-button/foundation.js
```diff
@@ -34,48 +34,27 @@ class MDCIconButtonToggleFoundation extends MDCFoundation {
...
   constructor(adapter) {
     super(Object.assign(MDCIconButtonToggleFoundation.defaultAdapter, adapter));
 
-    const {ARIA_PRESSED} = MDCIconButtonToggleFoundation.strings;
+    const {ICON_BUTTON_ON_CLASS} = MDCIconButtonToggleFoundation.cssClasses;
 
     /** @private {boolean} */
-    this.on_ = this.adapter_.getAttr(ARIA_PRESSED) === 'true';
-
+    this.on_ = this.adapter_.hasClass(ICON_BUTTON_ON_CLASS) || false;
```
kfranqueiro: Why `|| false`? Shouldn't `hasClass` return a boolean anyway?

Also, is there a reason we even need the `on_` variable? Could we just implement `isOn` to check whether this class is present?

packages/mdc-icon-button/foundation.js
```diff
@@ -91,52 +70,20 @@ class MDCIconButtonToggleFoundation extends MDCFoundation {
...
-    if (cssClass) {
-      this.adapter_.addClass(cssClass);
-    }
-    if (content) {
-      this.adapter_.setText(content);
-    }
-    if (label) {
-      this.adapter_.setAttr(ARIA_LABEL, label);
-    }
+    this.adapter_.setAttr(strings.ARIA_PRESSED, `${this.isOn()}`);
```
kfranqueiro: `this.isOn()` -> `isOn`?

packages/mdc-icon-button/foundation.js
```diff
@@ -91,52 +70,20 @@ class MDCIconButtonToggleFoundation extends MDCFoundation {
...
-    this.adapter_.setAttr(ARIA_PRESSED, this.on_.toString());
+    const {ICON_BUTTON_ON_CLASS} = MDCIconButtonToggleFoundation.cssClasses;
 
-    const {cssClass: classToRemove} =
-        this.on_ ? this.toggleOffData_ : this.toggleOnData_;
-
-    if (classToRemove) {
-      this.adapter_.removeClass(classToRemove);
+    this.on_ = isOn;
+    if (this.isOn()) {
```
kfranqueiro: `this.isOn()` -> `isOn`? We already have the variable right here.

test/screenshot/golden.json
```diff
@@ -157,7 +157,7 @@
     "screenshots": {
       "desktop_windows_chrome@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/07/31/04_55_54_478/spec/mdc-card/classes/baseline.html.windows_chrome_67.png",
       "desktop_windows_edge@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/07/31/04_55_54_478/spec/mdc-card/classes/baseline.html.windows_edge_17.png",
-      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/07/31/04_55_54_478/spec/mdc-card/classes/baseline.html.windows_firefox_61.png",
+      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/08/07/22_51_57_136/spec/mdc-card/classes/baseline.html.windows_firefox_61.png",
```
kfranqueiro: This seems weird...is this diff consistent, and if so do we have any possible explanation?

test/unit/mdc-icon-button/foundation.test.js
```diff
@@ -22,6 +22,7 @@ import {verifyDefaultAdapter} from '../helpers/foundation';
 import MDCIconButtonToggleFoundation from '../../../packages/mdc-icon-button/foundation';
 
 const {strings} = MDCIconButtonToggleFoundation;
+const {cssClasses} = MDCIconButtonToggleFoundation;
```
kfranqueiro: Change above line to `const {cssClasses, strings} = ...` and you don't need this line

test/unit/mdc-icon-button/mdc-icon-button-toggle.test.js
```diff
@@ -77,85 +77,38 @@ if (supportsCssVariables(window)) {
...
 });
 
 test('#adapter.addClass adds a class to the root element', () => {
   const {root, component} = setupTest();
   component.getDefaultFoundation().adapter_.addClass('foo');
-  assert.isOk(root.classList.contains('foo'));
-});
-
-test('#adapter.addClass adds a class to the inner icon element when used', () => {
-  const {root, component} = setupTest({useInnerIconElement: true});
```
kfranqueiro: We seem to never use `useInnerIconElement` or `tabIndex` in `setupTest` anymore.

Are we losing anything by no longer supporting the inner icon element? I'm inclined to think that's mitigated by having separate DOM for each state now.

test/screenshot/golden.json
williamernest: Yes, it's consistent. This was a result of me accepting screenshot diffs from master, not from within this branch.

Issue #3201, opened 2018-07-24
Text in single line list item wraps to next line on text overflow


abhiomkar:
Text content in single line list item wraps to next line.

<img width="621" alt="screen shot 2018-07-24 at 5 41 31 pm" src="https://user-images.githubusercontent.com/47312/43167763-d67b7f9c-8f68-11e8-8b01-a28fbac7eb33.png">

Expected it to show ellipses on text overflow.

abhiomkar:
This would require the text content to be wrapped in separate div to show ellipsis on text overflow.

PR #3311, opened 2018-08-07
feat(menu): Adds new menu, menu-surface.

### LINKED ISSUE #1495: mdc-simple-menu: disabled menu items not taken into account when focusing

0.23.0

All browsers

### What OS are you using?

NA

1. Go to the following codepen: https://codepen.io/gjdev/pen/QOLYEZ
2. Click the "reveal menu" button.
3. The first menu element now gets focus (this is not visible because the element is disabled, but can be checked via developer tools)
4. Use the arrow-keys to focus other menu elements. All disabled menu elements will get focus if you key to them (again, not visible because of the styling for disbled menu items).

Disabled menu items are not focusable.

The component ignores the disabled state when focusing elements, but correctly uses it for everything else. How to disable a menu item is also documented in the readme. Therefore, it's only logical to take the disabled state into account for focus as well.

Disabled menu items can be focused (even though you can't see it due to the styling). This leads to a confusing UI experience.

### LINKED ISSUE #1845: MDCSimpleMenu has no support for menu item icons

### LINKED ISSUE #2288: mdc-menu inside mdc-list - cutting menu items
What MDC Web Version are you using?
0.30.0

What browser(s) is this bug affecting?
All

What OS are you using?
Windows

What are the steps to reproduce the bug?

I would like to put an icon to open a menu in each item in my list, but when opening the menu it stays inside the container of the item not being visible the menu options. Is there a way to make the menu in the list items?

img

Tks

### LINKED ISSUE #2691: Menu Close event should be fired AFTER menu is actually closed
https://github.com/material-components/material-components-web/blob/a3c2de599ec799566fe150fd30dc6d4be1f1250c/packages/mdc-menu/foundation.js#L255

Specifically I'm dealing with this issue in React. Because of the render cycle and the way react works, you'd expect the state of the component to be updated before the event fires, not after.

Currently, when React runs through a new render, the state of the Menu still appears to be open, even though it should be closed which is creating synchronicity issues. My current workaround is to wrap the adapter method in a timeout so it will actually fire after the the menu is really closed.

```javascript
adapter: {
    notifySelected: function(evtData) {
      const evt = this.emit(MDCMenuFoundation.strings.SELECTED_EVENT, {
        index: evtData.index,
        item: this.items[evtData.index]
      });
      this.props.onClose && this.props.onClose(evt);
    },
    notifyCancel: function() {
      setTimeout(() => {
        const evt = this.emit(MDCMenuFoundation.strings.CANCEL_EVENT, {});
        this.props.onClose && this.props.onClose(evt);
      });
    }
  }
```

### LINKED ISSUE #2756: Create a new mdc-menu-surface
- [x] README
- [x] Pull the code from existing mdc-menu that is responsible for showing just the surface of a menu

### LINKED ISSUE #2758: Re-make mdc-menu with mdc-menu-surface and mdc-list
- [ ] Use mdc-menu-surface for its...surface
- [ ] Use mdc-list, now that it is accessible (see #2757)

### LINKED ISSUE #2763: MDC Menu should not preventDefault for space keys within inputs
** PLEASE READ THIS BEFORE FILING AN ISSUE **

0.35.1

Firefox 52.7.3

### What OS are you using?

Windows 7

1.  Displaying an MDCMenu on a button click.  The MDCMenu has an integrated text input.

2.  Type 'MDC Test' into text input.

1.  I expect 'MDC Test' to be visible in the text input

1.  'MDCTest' is displayed.  All spaces are being eaten by the menu

I tracked down in the MDCMenu code a check for a space to stop any unwanted scrolling.  I'm not
sure why no one else is seeing this.  I am using React so the event model is different and might be
causing some strange event bubbling but it doesn't seem like it should be.
**Spaces are actually eaten everywhere in Firefox, even the debug console.

### LINKED ISSUE #2770: Support dynamically setting the anchoring element of an mdc-menu-surface

### LINKED ISSUE #2972: MDC Menu readme instructions are missing import statement

> 0.36.1

[View the MDC Menu readme here](https://github.com/material-components/material-components-web/tree/master/packages/mdc-menu). The Javascript instructions include the following:

```
 // Instantiation
  ...
  var menu = new mdc.menu.MDCMenu(menuEl);
  ...
```

`new mdc.menu.MDCMenu(menuEl)` cannot be instantiated without importing an external module, which is not provided. This is done on other components, such as the ripple effect on the button component:

```
import {MDCRipple} from '@material/ripple';
const buttonRipple = new MDCRipple(document.querySelector('.mdc-button'));
```

### What is needed

Please provide the missing import statement or clarify how this is supposed to work. Thanks!

williamernest:
Resolves #2972, resolves #2770, resolves #2763, resolves #2758, resolves #2756, resolves #2691, resolves #2288, resolves #1845, resolves #1495

Adds the ability for a menu-surface to be positioned relative to an element, relative to the page, or fixed to the viewport. Adds the ability to dynamically change the anchoring element. Adds the ability to hoist the menu to the body.

BREAKING CHANGE: Menu positioning logic has been split into its own package (`mdc-menu-surface`). `mdc-menu` is rebuilt to use `mdc-menu-surface` and `mdc-list` styles and JavaScript.


packages/mdc-menu-surface/README.md
```diff
@@ -0,0 +1,164 @@
...
+### `MDCMenuSurfaceFoundation`
+
+Method Signature | Description
+--- | ---
+`setAnchorCorner(corder: Corner) => void` | Sets the corner that the menu surface will be anchored to. See [constants.js](./constants.js)
+`setAnchorMargin(margin: AnchorMargin) => void` | Sets the distance from the anchor point that the menu surface should be shown.
+`setIsHoisted(isHoisted: boolean) => void` | Sets whether the menu surface has been hoisted to the body so that the offsets are calculated relative to the page and not the anchor.
+`setFixedPosition(isFixed: boolean) => void` | Sets whether the menu surface is using fixed positioning.
+`setAbsolutePosition(x: number, y: numnber) => void` | Sets the absolute x/y position of the menu. Should only be used when the menu is hoisted or using fixed positioning.
+`handleDocumentClick(event: Event) => void` | Method used as the callback function for the `click` event.
```
kfranqueiro: Should this be `handleBodyClick`, or should the event [de]registration APIs actually use `document.documentElement`, not `document.body` (in case the body doesn't span the full viewport)?

packages/mdc-menu-surface/constants.js
```diff
@@ -0,0 +1,86 @@
...
+  FOCUSABLE_ELEMENTS: 'button:not(:disabled), [href]:not([aria-disabled="true"]), input:not(:disabled), ' +
+  'select:not(:disabled), textarea:not(:disabled), [tabindex]:not([tabindex="-1"]):not([aria-disabled="true"])',
+};
+
+/** @enum {number} */
+const numbers = {
+  // Amount of time to wait before triggering a selected event on the menu-surface. Note that this time
+  // will most likely be bumped up once interactive lists are supported to allow for the ripple to
+  // animate before closing the menu-surface
+  SELECTED_TRIGGER_DELAY: 50,
```
kfranqueiro: Does this really belong on surface, not menu?

Actually, is this even used (other than in tests apparently)?

packages/mdc-menu-surface/constants.js
```diff
@@ -0,0 +1,86 @@
...
+  ANIMATING_CLOSED: 'mdc-menu-surface--animating-closed',
+  ANIMATING_OPEN: 'mdc-menu-surface--animating-open',
+  FIXED: 'mdc-menu-surface--fixed',
+  OPEN: 'mdc-menu-surface--open',
+  ROOT: 'mdc-menu-surface',
+};
+
+/** @enum {string} */
+const strings = {
+  ARIA_DISABLED_ATTR: 'aria-disabled',
```
kfranqueiro: Unused?

packages/mdc-menu-surface/foundation.js
```diff
@@ -0,0 +1,572 @@
...
+  getHorizontalOriginOffset_(corner) {
+    const {anchorWidth} = this.measures_;
+    // isRightAligned corresponds to using the 'right' property on the surface.
+    const isRightAligned = Boolean(corner & CornerBit.RIGHT);
+    const avoidHorizontalOverlap = Boolean(this.anchorCorner_ & CornerBit.RIGHT);
+
+    if (isRightAligned) {
+      const rightOffset = avoidHorizontalOverlap ? anchorWidth - this.anchorMargin_.left : this.anchorMargin_.right;
+
+      // Hoisted elements positioning doesn't account for the scrollbar, so the right property needs to be reduced by
```
kfranqueiro: This potentially sounds non-RTL-aware right now... Have we tested this on windows and linux?

packages/mdc-menu-surface/foundation.js
```diff
@@ -0,0 +1,572 @@
...
+
+    return avoidHorizontalOverlap ? anchorWidth - this.anchorMargin_.right : this.anchorMargin_.left;
+  }
+
+  /**
+   * @param {!Corner} corner Origin corner of the menu surface.
+   * @return {number} Vertical offset of menu surface origin corner from corresponding anchor corner.
+   * @private
+   */
+  getVerticalOriginOffset_(corner) {
```
kfranqueiro: It looks like the code here was reduced from the original code in MDC Menu, can you explain? Specifically, there used to be considerations for viewport size and distance from edges here. If I were to wager a guess, is this because we're resizing the menu instead of repositioning?

packages/mdc-menu-surface/foundation.js
```diff
@@ -0,0 +1,572 @@
...
+      [horizontalAlignment]: horizontalOffset ? horizontalOffset : '0',
+      [verticalAlignment]: verticalOffset ? verticalOffset : '0',
+    };
+    const {anchorWidth, surfaceWidth} = this.measures_;
+    // Center align when anchor width is comparable or greater than menu surface, otherwise keep corner.
+    if (anchorWidth / surfaceWidth > numbers.ANCHOR_TO_MENU_SURFACE_WIDTH_RATIO) {
+      horizontalAlignment = 'center';
+    }
+
+    // If the menu-surface has been hoisted to the body, it's no longer relative to the anchor element
```
kfranqueiro: I noticed that this method used to bail out if there was no anchor element but it no longer does that. What are the implications of that on this method?

packages/mdc-menu-surface/foundation.js
```diff
@@ -0,0 +1,572 @@
...
+    this.isOpen_ = false;
+    this.shouldRestoreFocus_();
+  }
+
+  /**
+   * The last focused element when the menu surface was opened should regain focus, if the user is
+   * focused on or within the menu surface when it is closed.
+   * @private
+   */
+  shouldRestoreFocus_() {
```
kfranqueiro: This is a weird name for how this method works. I'd expect this to return a boolean and for `close` to perform the actual adapter call. Otherwise rename this to something like `maybeRestoreFocus_` at least.

packages/mdc-menu-surface/mdc-menu-surface.scss
```diff
@@ -0,0 +1,86 @@
...
+// limitations under the License.
+//
+
+@import "@material/animation/variables";
+@import "@material/elevation/mixins";
+@import "@material/theme/mixins";
+@import "@material/rtl/mixins";
+@import "./mixins";
+
+$mdc-menu-surface-fade-in-duration: .03s;
```
kfranqueiro: -> _variables.scss?

packages/mdc-menu-surface/mdc-menu-surface.scss
```diff
@@ -0,0 +1,86 @@
...
+  box-sizing: border-box;
+  max-width: calc(100vw - 32px);
+  max-height: calc(100vh - 32px);
+  margin: 0;
+  padding: 0;
+  transform: scale(1);
+  transform-origin: top left;
+  border-radius: 4px;
+  opacity: 0;
+  white-space: nowrap;
```
kfranqueiro: Does this make sense on `menu-surface`? Seems ultimately pertinent to #3201, maybe, but maybe not right here?

packages/mdc-menu-surface/mdc-menu-surface.scss
```diff
@@ -0,0 +1,86 @@
...
+.mdc-menu-surface {
+  @include mdc-elevation(8);
+  @include mdc-menu-surface-fill-color(surface);
+  @include mdc-menu-surface-ink-color(on-surface);
+  @include mdc-rtl-reflexive-property(transform-origin, top left, top right);
+
+  display: none;
+  position: absolute;
+  box-sizing: border-box;
+  max-width: calc(100vw - 32px);
```
kfranqueiro: Should 32px be a variable?

packages/mdc-menu-surface/mdc-menu-surface.scss
```diff
@@ -0,0 +1,86 @@
...
+  max-height: calc(100vh - 32px);
+  margin: 0;
+  padding: 0;
+  transform: scale(1);
+  transform-origin: top left;
+  border-radius: 4px;
```
kfranqueiro: Should this have a corner-radius mixin?

packages/mdc-menu-surface/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "material design",
+    "menu surface"
+  ],
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/material-components/material-components-web.git"
+  },
+  "dependencies": {
+    "@material/animation": "^0.34.0",
```
kfranqueiro: Are these versions up to date?

packages/mdc-menu-surface/README.md
```diff
@@ -0,0 +1,164 @@
...
+`open` | Boolean | Proxies to the foundation's `isOpen`/(`open`, `close`) methods.
+`quickOpen` | Boolean | Proxies to the foundation's `setQuickOpen()` method.
+
+Method Signature | Description
+--- | ---
+`show() => void` | Proxies to the foundation's `open()` method.
+`hide() => void` | Proxies to the foundation's `close()` method.
+`setAnchorCorner(Corner) => void` | Proxies to the foundation's `setAnchorCorner(Corner)` method.
+`setAnchorMargin(AnchorMargin) => void` | Proxies to the foundation's `setAnchorMargin(AnchorMargin)` method.
+`setFixedPosition(isFixed: boolean) => void` | Adds the `mdc-menu-surface--fixed` class to the `mdc-menu-surface` element. Proxies to the foundation's `setIsHoisted()` and `setFixedPosition()` methods.
```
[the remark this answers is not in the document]
williamernest: `setFixedPosition` doesn't require hoisting. `setAbsolutePosition` requires you to use the provided `hoistMenuToBody` method or instantiate the menu in a position that already is positioned relative to the body.

packages/mdc-menu-surface/README.md
```diff
@@ -0,0 +1,164 @@
...
+## `MDCMenuSurface` Properties and Methods
+
+Property | Value Type | Description
+--- | --- | ---
+`open` | Boolean | Proxies to the foundation's `isOpen`/(`open`, `close`) methods.
+`quickOpen` | Boolean | Proxies to the foundation's `setQuickOpen()` method.
+
+Method Signature | Description
+--- | ---
+`show() => void` | Proxies to the foundation's `open()` method.
```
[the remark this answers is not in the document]
williamernest: Removed `show`/`hide`

packages/mdc-menu-surface/foundation.js
williamernest: The menu-surface can work without an anchor element (e.g. context menus). If the menu is opened without an anchor element and without having a position, the menu will open at `x: 0`/`y: 0`.

packages/mdc-menu-surface/README.md
```diff
@@ -0,0 +1,164 @@
...
+`hide() => void` | Proxies to the foundation's `close()` method.
+`setAnchorCorner(Corner) => void` | Proxies to the foundation's `setAnchorCorner(Corner)` method.
+`setAnchorMargin(AnchorMargin) => void` | Proxies to the foundation's `setAnchorMargin(AnchorMargin)` method.
+`setFixedPosition(isFixed: boolean) => void` | Adds the `mdc-menu-surface--fixed` class to the `mdc-menu-surface` element. Proxies to the foundation's `setIsHoisted()` and `setFixedPosition()` methods.
+`setAbsolutePosition(x: number, y: number) => void` | Proxies to the foundation's `setAbsolutePosition(x, y)` method. Used to set the absolute x/y position of the menu on the page. Should only be used when the menu is hoisted to the body.
+`setMenuSurfaceAnchorElement(element: Element) => void` | Changes the element used as an anchor for `menu-surface` positioning logic. Should be used with conjunction with `hoistMenuToBody()`.
+`hoistMenuToBody() => void` | Removes the `menu-surface` element from the DOM and appends it to the `body` element. Should be used to overcome `overflow: hidden` issues.
+`setIsHoisted() => void` | Proxies tot he foundation's `setIsHoisted` method.
```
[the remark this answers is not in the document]
williamernest: This function sets the state of the foundation so calculate everything based on the body. Since the menu could technically be a few level deep into the body, but still be positioned relative to the body, the user may not always have the menu completely hoisted to the body.

packages/mdc-menu-surface/foundation.js
williamernest: Yes, there was extra logic here to allow the menu to go all the way to the edge of the screen and move the origin up/down based on the height of the menu. 

This behavior is not described or documented in the spec, and ignored the spec guidelines of minimum distance to the edge.

demos/menu.scss
```diff
@@ -16,5 +16,67 @@
...
+@import "../packages/mdc-icon-button/mdc-icon-button";
+@import "../packages/mdc-checkbox/mdc-checkbox";
 @import "../packages/mdc-list/mdc-list";
 @import "../packages/mdc-menu/mdc-menu";
+@import "../packages/mdc-menu-surface/mdc-menu-surface";
+
+html,
+body,
+main {
+  position: relative;
```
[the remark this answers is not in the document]
williamernest: `position: relative` is required for body, but I removed the rest.

packages/mdc-menu-surface/foundation.js
williamernest: Outdated comment. Yes, this works on windows/linux.

packages/mdc-menu-surface/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+The menu surface can use fixed positioning when being displayed.
+
+```html
+<div class="mdc-menu-surface mdc-menu-surface--fixed">
+...
+</div>
+```
+
+```js
```
kfranqueiro: There should probably be a "Or in JS:" before this, right? You only need one or the other IIUC?

packages/mdc-menu/index.js
```diff
@@ -111,85 +130,100 @@ class MDCMenu extends MDCComponent {
...
+
+  /**
+   * Sets the element that the menu-surface is anchored to.
+   * @param {!HTMLElement} element
+   */
+  setAnchorElement(element) {
+    this.menuSurface_.anchorElement = element;
+  }
+
+  registerListeners_() {
```
kfranqueiro: I'm curious why we are registering and deregistering listeners based on when it's open/closed. Is there any harm to just having them hooked up all the time? (It'd be one thing if they were registered on the body, but these are on the menu surface element.)

packages/mdc-menu/index.js
```diff
@@ -27,79 +29,96 @@ class MDCMenu extends MDCComponent {
...
    */
   get items() {
-    const {itemsContainer_: itemsContainer} = this;
-    return [].slice.call(itemsContainer.querySelectorAll('.mdc-list-item[role]'));
+    return this.list_.listElements_;
   }
 
   /**
-   * Return the item within the menu that is selected.
+   * Return the item within the menu at the index specified.
```
kfranqueiro: Do we foresee a purpose for this API? I know it existed before but I'm not sure why.

packages/mdc-menu-surface/foundation.js
williamernest: I guess it isn't completely outdated. For hoisted or fixed elements, adjust the offset by the difference between viewport width and body width so when we calculate the right value (`adjustPositionForHoistedElement_`) based on the element position, the right property is correct.

packages/mdc-menu/index.js
williamernest: I believe this was here previously so we could attempt to replicate `.item(index)` functionality from the native select. I planned on implementing this in select so I included it here, but it may not be required.

packages/mdc-menu-surface/util.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 Google Inc. All Rights Reserved.
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
kfranqueiro: While we're here, I think there are a bunch of these held over from old JS animation logic in "Simple" Menu that we don't need anymore (they're only referenced in tests):

* `clamp`
* `bezierProgress` (and `getBezierCoordinate_` and `solvePositionFromXValue_`)

packages/mdc-menu/README.md
```diff
@@ -73,131 +123,120 @@ The menu can be positioned to automatically anchor to a parent element when open
...
 
 See [Importing the JS component](../../docs/importing-js.md) for more information on how to import JavaScript.
 
 Property | Value Type | Description
 --- | --- | ---
-`open` | Boolean | Proxies to the foundation's `isOpen`/(`open`, `close`) methods.
-`items` | Array<Element> | Proxies to the foundation's container to query for all `.mdc-list-item[role]` elements.
-`itemsContainer` | Element | Queries the foundation's root element for the `mdc-menu__items` container element.
-`quickOpen` | Boolean | Proxies to the foundation's `setQuickOpen()` method.
+`open` | Boolean | `get` proxies to the menu surface's `open` property. `set` proxies to the `show` and `hide`methods.
```
kfranqueiro: Missing space before "methods"

packages/mdc-menu/README.md
```diff
@@ -73,131 +123,120 @@ The menu can be positioned to automatically anchor to a parent element when open
...
+`items` | Array<Element> | Proxies to the list to query for all `.mdc-list-item` elements.
+`quickOpen` | Boolean | Proxies to the menu surface `quickOpen` property.
 
 Method Signature | Description
 --- | ---
-`show({focusIndex: ?number}) => void` | Proxies to the foundation's `open()` method. An optional config parameter allows the caller to specify which item should receive focus after the menu opens.
+`show() => void` | Proxies to the menu surface's `open()` method.
```
kfranqueiro: show and hide were removed, right?

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -0,0 +1,484 @@
...
+});
+
+test('setAbsolutePosition calls the foundation setAbsolutePosition function', () => {
+  const {component, mockFoundation} = setupTest();
+  component.setAbsolutePosition(10, 10);
+  td.verify(mockFoundation.setAbsolutePosition(10, 10));
+  td.verify(mockFoundation.setIsHoisted(true));
+});
+
+test('setFixedPosition is false', () => {
```
kfranqueiro: Move this up below the setFixedPosition is true test?

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -0,0 +1,484 @@
...
+  td.verify(mockFoundation.setAnchorMargin({top: 0, right: 0, bottom: 0, left: 0}));
+});
+
+test('setQuickOpen', () => {
+  const {component, mockFoundation} = setupTest();
+  component.quickOpen = false;
+  td.verify(mockFoundation.setQuickOpen(false));
+});
+
+test('foundationAdapter#addClass adds a class to the root element', () => {
```
kfranqueiro: foundationAdapter -> adapter

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -0,0 +1,484 @@
...
+  button.focus();
+  component.getDefaultFoundation().adapter_.saveFocus();
+  root.focus();
+  component.getDefaultFoundation().adapter_.restoreFocus();
+  assert.equal(document.activeElement, button);
+  document.body.removeChild(button);
+  document.body.removeChild(root);
+});
+
+test('adapter#restoreFocus does not restores the focus if never called adapter#saveFocus', () => {
```
kfranqueiro: restores -> restore

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -0,0 +1,484 @@
...
+  assert.equal(component.getDefaultFoundation().adapter_.getAnchorDimensions().height, 21);
+  assert.equal(component.getDefaultFoundation().adapter_.getAnchorDimensions().width, 42);
+  document.body.removeChild(anchor);
+});
+
+test('adapter#getAnchorDimensions returns undefined if there is no anchor element', () => {
+  const {root, component} = setupTest(true);
+  document.body.appendChild(root);
+  component.initialSyncWithDOM();
+  assert.equal(component.getDefaultFoundation().adapter_.getAnchorDimensions(), undefined);
```
kfranqueiro: `assert.isUndefined`

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -0,0 +1,484 @@
...
+
+test(`adapter#notifyOpen fires an ${strings.OPENED_EVENT} custom event`, () => {
+  const {root, component} = setupTest();
+  const handler = td.func('notifyOpen handler');
+  root.addEventListener(strings.OPENED_EVENT, handler);
+  component.getDefaultFoundation().adapter_.notifyOpen();
+  td.verify(handler(td.matchers.anything()));
+});
+
+test('adapter#restoreFocus restore focus saved by adapter#saveFocus', () => {
```
kfranqueiro: This one was correct with "restores" :|

packages/mdc-menu-surface/util.js
```diff
@@ -0,0 +1,35 @@
...
+function getTransformPropertyName(globalObj, forceRefresh = false) {
+  if (storedTransformPropertyName_ === undefined || forceRefresh) {
+    const el = globalObj.document.createElement('div');
+    const transformPropertyName = ('transform' in el.style ? 'transform' : 'webkitTransform');
+    storedTransformPropertyName_ = transformPropertyName;
+  }
+
+  return storedTransformPropertyName_;
+}
+export {getTransformPropertyName};
```
kfranqueiro: Add blank line before export

packages/mdc-menu-surface/adapter.js
```diff
@@ -0,0 +1,142 @@
...
+  hasClass(className) {}
+
+  /** @return {boolean} */
+  hasAnchor() {}
+
+  /**
+   * @param {string} type
+   * @param {function(!Event)} handler
+   */
+  registerInteractionHandler(type, handler) {}
```
kfranqueiro: Do we want to move keydown registration to the component now or in a separate PR?

packages/mdc-menu-surface/foundation.js
```diff
@@ -0,0 +1,573 @@
...
+  static get Corner() {
+    return Corner;
+  }
+
+  /**
+   * {@see MDCMenuSurfaceAdapter} for typing information on parameters and return
+   * types.
+   * @return {!MDCMenuSurfaceAdapter}
+   */
+  static get defaultAdapter() {
```
kfranqueiro: default adapter is missing `focus`

Issue #2269, opened 2018-02-17
Slider error if min is greater than default max value of 100


entone:
If the minimum value for your slider is greater than 100, the MDCComponent throws `Error: Cannot set min to be greater than the slider's maximum value`.

I tried updating the logic to always set the max value first here https://github.com/prateekbh/preact-material-components/blob/master/Slider/Slider.jsx#L49

But that didn't seem to fix it. Not sure what's going on.

none23:
I also experience this issue with https://github.com/jamesmfriedman/rmwc

entone:
I was able to work around the bug by using smaller values and providing a
multiplier for displaying and storing the value.

moog16:
@entone, interesting that it appears in both React and Preact implementations of MDC. Would you be able to provide a [Codepen](https://codepen.io/traviskaufman/pen/pNQmRp) for us to replicate your work just using https://github.com/material-components/material-components-web/tree/master/packages/mdc-slider.

vervallsweg:
@moog16 Sorry I don't have the time to setup a Codepen now. However also with pure mdc you can easily reproduce this bug.

Add a standard slider to your DOM:
`<div class="mdc-slider mdc-slider--discrete" data-mdc-auto-init="MDCSlider" tabindex="0" role="slider" aria-valuemin="200" aria-valuemax="600" aria-valuenow="200" data-step="10" aria-label="Select Value"><div class="mdc-slider__track-container"><div class="mdc-slider__track"></div></div><div class="mdc-slider__thumb-container"><div class="mdc-slider__pin"><span class="mdc-slider__pin-value-marker"></span></div><svg class="mdc-slider__thumb" width="21" height="21"><circle cx="10.5" cy="10.5" r="7.875"></circle></svg><div class="mdc-slider__focus-ring"></div></div></div>`

Run auto init:
`window.mdc.autoInit();`

Check your console:
`Uncaught Error: Cannot set min to be greater than the slider's maximum value
    at MDCSliderFoundation.setMin (material-components-web.js:13482)
    at MDCSlider.set (material-components-web.js:13110)
    at MDCSlider.initialSyncWithDOM (material-components-web.js:13055)
    at MDCSlider.MDCComponent (material-components-web.js:259)
    at new MDCSlider (material-components-web.js:12929)`

If you replace your values with ` aria-valuemin="20" aria-valuemax="60" aria-valuenow="20" ` and run auto init again, results in no error.

[...]

PR #3314, opened 2018-08-08
Provide fix for 'Slider error if min is greater than default max value of 100' #2269

pvogel2:
Fixes [#2269](https://github.com/material-components/material-components-web/issues/2269). Added test for custom range triggering the issue.


pvogel2:
I signed it!

Issue #2826, opened 2018-05-25
mdc-textfield Outlined incorrect specs


EstebanG23:
Specs show the whole component should come to be 56px in height. Also 16px in spacing between the icon, and 28px from the label baseline to the input baseline.
https://material.io/design/components/text-fields.html#spec

![screen shot 2018-05-25 at 2 23 45 pm](https://user-images.githubusercontent.com/18471209/40566797-4224de80-6027-11e8-9389-a16ad3429331.png)

But our components show:
- 54px in height
- 16px on top of icon, 24px for icon and **14px below icon**
- 30px from label baseline to input baseline

![screen shot 2018-05-25 at 2 09 09 pm](https://user-images.githubusercontent.com/18471209/40566888-af972c98-6027-11e8-90e2-2a5a0b9cc25c.png)
![screen shot 2018-05-25 at 2 09 37 pm](https://user-images.githubusercontent.com/18471209/40566886-af6719fe-6027-11e8-9a62-4aca3a26cdf4.png)
![screen shot 2018-05-25 at 2 12 20 pm](https://user-images.githubusercontent.com/18471209/40566887-af7f517c-6027-11e8-9f44-036ebd3aebc4.png)

rubenmendoza:
![image](https://user-images.githubusercontent.com/3853885/40601275-d48d0f14-6297-11e8-9300-98f8ed6aabed.png)

My fixes so far, :

.mdc-text-field--outlined .mdc-text-field__input:-webkit-autofill+.mdc-floating-label{
	-webkit-transform: translateY(calc(-100% - 9px)) scale(.75);
    transform: translateY(calc(-100% - 9px)) scale(.75);
}
.mdc-text-field--outlined .mdc-floating-label--float-above{
	    -webkit-transform: translateY(calc(-130% - 4px)) scale(.75);
    transform: translateY(calc(-130% - 4px)) scale(.75);
}
.mdc-text-field--outlined{
	margin-bottom: 0;
}
.mdc-text-field--outlined+.mdc-text-field-helper-text{
	margin-left: 12px;
}

/* Change autocomplete styles in WebKit */
input:-webkit-autofill,
input:-webkit-autofill:hover, 
input:-webkit-autofill:focus
input:-webkit-autofill, 
textarea:-webkit-autofill,
textarea:-webkit-autofill:hover
textarea:-webkit-autofill:focus,
select:-webkit-autofill,
select:-webkit-autofill:hover,
[...]

PR #3306, opened 2018-08-06
test(text-field): Add screenshot tests

williamernest:
Adds screenshot tests for text field

PR #2859, opened 2018-06-01
refactor(text-field): Change text-field--box to be the new default

EstebanG23:
Add changes to CSS classes to override necessary styles. Edit README to reflect the new default. Edit JS to match the changes. Change tests and constants to reflect the new changes.

Fixes #2780

PR #3321, opened 2018-08-09
fix(text-field): Input position and textarea size

acdvorak:
BREAKING CHANGE: The `mdc-text-field--upgraded` class has been removed. `mdc-text-field__input` position has changed by 2px to match spec. `mdc-text-field--textarea` width in IE and Edge now matches other browsers.

### What it does

- Fixes vertical text alignment of `mdc-text-field__input` in the `mdc-text-field--box` variant to match spec. Baseline is now 28px instead of 30px. See #2826.
- Fixes `textarea` width in IE and Edge:
    - Replaces unsupported values `fit-content` and `initial` with `auto`
- Removes the `mdc-text-field--upgraded` class:
    - Prevents text fields from resizing and causing browser reflow jank after the JS initializes
    - Most properties from `mdc-text-field--upgraded` have been moved to `mdc-text-field`
    - Sets `display: inline-flex`, `box-sizing: border-box`, and `margin-top: 16px` on all variants

#### Before:

before

#### After:

after

#### Spec:

[Text field guidelines](https://material-io-project.appspot.com/design/components/text-fields.html#spec)

image


acdvorak:
Wait for #3306 to be merged before merging this PR.

kfranqueiro:
Might also want to wait for #2859 which will almost undoubtedly conflict with this?

PR #3335, opened 2018-08-12
Allow Snackbar SCSS variables to be customized

ZhangYiJiang:
Using SCSS `!default` to allow Snackbar variables to be overriden by user.


ZhangYiJiang:
I signed it!

PR #3338, opened 2018-08-13
fix(text-field): Fix label shake bug. Update invalid screenshots to show required star.

williamernest:
Fixes an issue with the label shaking on blur when the text field is invalid and empty. This causes the label to shake left/right while also animating down resulting in a circular animation. 

Also updates the screenshot tests to show the required state.


test/screenshot/spec/mdc-textfield/classes/invalid-focused-helper-text-persistent-validation-msg.html
```diff
@@ -41,7 +41,7 @@
 
     <div class="test-cell test-cell--textfield">
       <div class="mdc-text-field mdc-text-field--outlined mdc-text-field--invalid mdc-text-field--focused">
-        <input type="text" id="outlined-text-field" class="mdc-text-field__input test-text-field__input" value=" ">
+        <input type="text" id="outlined-text-field" class="mdc-text-field__input test-text-field__input" value=" " required pattern=".{4,8}">
```
acdvorak: Is the space in `value=" "` intentional?

If so, maybe add an HTML comment explaining what it's for.

test/screenshot/spec/mdc-textfield/classes/invalid-focused-helper-text-persistent-validation-msg.html
```diff
@@ -32,7 +32,7 @@
 
     <div class="test-cell test-cell--textfield">
       <div class="mdc-text-field mdc-text-field--box mdc-text-field--invalid mdc-text-field--focused">
-        <input type="text" id="filled-text-field" class="mdc-text-field__input test-text-field__input">
+        <input type="text" id="filled-text-field" class="mdc-text-field__input test-text-field__input" required pattern=".{4,8}">
```
acdvorak: Why do we need both `required` _and_ `pattern`?

williamernest: Required is what forces the asterisk to show. The `pattern` is useful when trying to debug the invalid state using the screenshot tests since you're unable to test the label shake without it.

test/screenshot/spec/mdc-textfield/classes/invalid-focused-helper-text-persistent-validation-msg.html
williamernest: I added a comment to explain. The text field does not draw the SVG path when the label is floating unless the input has a value or has focus. Since I'm forcing the focused state for the purpose of screenshot tests, I added a single space value to force the outline to notch.

PR #3340, opened 2018-08-13
feat(drawer): Update to dismissible drawer foundation

### LINKED ISSUE #3339: Update Dismissible drawer foundation for new motion and to support modal
Dismissible drawer foundation needs to be updated to accommodate changes required for modal and the updated motion makes the app content snap on open/close.

Remove logic used to calculate app content dimensions from Dismissible and use CSS sibling selector instead.

abhiomkar:
Fixes #3339

Changes include:

- Remove logic used to calculate app content dimensions, uses CSS sibling selector to style app content.
- Updates to Dismissible drawer to accommodate Modal drawer.
- Updated README.
- Updated demos / docs to support list keyboard navigation.


moog16:
The new animation seems to be overlapping with the main content of the page. Is this intended? 


screen shot 2018-08-13 at 13 59 15

moog16:
Also there is a gap between the opening animation and the main content.

screen shot 2018-08-13 at 14 00 19

packages/mdc-drawer/README.md
```diff
@@ -100,22 +153,22 @@ Dismissible drawers are by default hidden off screen, and can slide into view. D
     </div>
   </header>
 
-  <div class="mdc-drawer-app-content">
+  <div class="mdc-drawer__app-content">
```
moog16: I had this as `-app-content`, since I thought `__app-content` meant that it is contained within `mdc-drawer`. Does BEM say otherwise?

packages/mdc-drawer/index.js
```diff
@@ -74,14 +65,15 @@ export class MDCDrawer extends MDCComponent {
   }
 
   destroy() {
-    document.removeEventListener('keydown', this.handleKeydown_);
+    this.root_.removeEventListener('keydown', this.handleKeydown_);
```
moog16: Why was this changed to the root element? It doesn't seem to close when I his escape.

moog16:
Also there is a gap between the opening animation and the main content.

screen shot 2018-08-13 at 14 00 19

packages/mdc-drawer/README.md
abhiomkar: Makes sense. Changed it back to `-app-content`.

packages/mdc-drawer/index.js
abhiomkar: Wanted to make ESC work only when the focus is on Drawer. So this works only when the focus is on Drawer. Hope that makes sense.

moog16: I did click into the drawer, and hit escape, but it didn't close. Are you seeing this behavior as well?

abhiomkar: keydown wouldn't trigger when drawer is not in focus. drawer itself is not focusable element but nav items are.

abhiomkar:
As discussed on chat, Animating on width causes the app content to reflow & animating app content on transform causes it to snap when animation ends which looks distracting. So removed animation from app content and snaps immediately on open / close while the drawer still animates on transform.

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,74 @@ npm install @material/drawer
...
+        <span class="mdc-list-item__label mdc-drawer__list-item-label">Work</span>
+      </a>
+    </nav>
+  </div>
+</div>
+```
+
+> NOTE: First navigation item in list group should be explicitely set tabindex to '0'.
+
+And then list group needs to be instantiated for keyboard navigation.
```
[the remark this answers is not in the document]
kfranqueiro: It doesn't seem like part of the note, if that's what you mean.

I would probably leave out the "And then".

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,74 @@ npm install @material/drawer
...
+      <a class="mdc-list-item" href="#">
+        <i class="material-icons mdc-list-item__graphic" aria-hidden="true">bookmark</i>
+        <span class="mdc-list-item__label mdc-drawer__list-item-label">Work</span>
+      </a>
+    </nav>
+  </div>
+</div>
+```
+
+> NOTE: First navigation item in list group should be explicitely set tabindex to '0'.
```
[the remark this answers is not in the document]
kfranqueiro: also remove "be"

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,74 @@ npm install @material/drawer
...
+var listList = document.querySelectorAll('.mdc-list');
+
+listList.forEach(function(list) {
+  var list = mdc.list.MDCList.attachTo(list);
+  list.singleSelection = true;
+  list.vertical = true;
+});
+```
+
+Please see the [list readme](https://github.com/material-components/material-components-web/tree/master/packages/mdc-list#list-groups)
```
kfranqueiro: Change link to relative: `(../mdc-list#list-groups)`

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,74 @@ npm install @material/drawer
...
+  </div>
+</div>
+```
+
+> NOTE: First navigation item in list group should be explicitely set tabindex to '0'.
+
+And then list group needs to be instantiated for keyboard navigation.
+
+```javascript
+var listList = document.querySelectorAll('.mdc-list');
```
kfranqueiro: `listList` -> `lists` as with the demo above

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,74 @@ npm install @material/drawer
...
+
+listList.forEach(function(list) {
```
kfranqueiro: `(list)` -> `(el)` or `(listEl)` as with the demo above, to avoid redeclaring the same variable in the function

packages/mdc-drawer/README.md
```diff
@@ -172,15 +225,14 @@ body {
...
 --- | ---
 `mdc-drawer` |  Mandatory.
-`mdc-drawer--animating-close` | Applies the transition to the dismissible drawer while it is animating from the open to the closed position.
-`mdc-drawer--animating-open` | Applies the transition to the dismissible drawer while it is animating from the closed to the open position.
+`mdc-drawer--closing` | Applies the transition to the dismissible drawer while it is animating from the open to the closed position.
+`mdc-drawer--opening` | Applies the transition to the dismissible drawer while it is animating from the closed to the open position.
 `mdc-drawer-app-content` | Used for dismissible drawer variant sibling element that should animate open/closed with it.
-`mdc-drawer-app-content--animating-open` | Applies the transition to the app content element while it is animating to the open position.
-`mdc-drawer-app-content--animating-close` | Applies the transition to the app content element while it is animating to the closed position.
+while it is animating to the closed position.
```
kfranqueiro: This line looks like it doesn't belong here?

packages/mdc-drawer/dismissible/foundation.js
```diff
@@ -131,20 +120,29 @@ class MDCDismissibleDrawerFoundation extends MDCFoundation {
...
-    this.adapter_.removeClass(ANIMATING_CLOSE);
+
+    this.adapter_.removeClass(ANIMATE);
+    this.adapter_.removeClass(OPENING);
+    this.adapter_.removeClass(CLOSING);
   }
 }
 
+const delay = function(fn) {
+  setTimeout(fn, 10);
```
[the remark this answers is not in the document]
kfranqueiro: I agree the usage of this function looks like a place for `requestAnimationFrame`.

packages/mdc-drawer/dismissible/mdc-drawer-dismissible.scss
```diff
@@ -17,62 +17,30 @@
...
 
-  @include mdc-rtl {
-    transform: translateX(100%);
-    border-left-width: 0;
-  }
-}
+  display: none;
+  position: absolute;
+  top: 0;
+  transition-property: transform;
```
kfranqueiro: transition-property and transition-timing-function are already applied to `.mdc-drawer` in the standard styles so they're redundant here.

On a separate topic (for a separate PR), maybe it'd be a better idea to put any common drawer styles directly within `mdc-drawer.scss` (instead of under `standard`) to make it evident that it has an effect on all variants.

test/unit/mdc-drawer/dismissible.foundation.test.js
```diff
@@ -181,93 +117,102 @@ test(`#isOpen returns false when it lacks ${cssClasses.OPEN} class`, () => {
...
 });
 
-test('#handleTransitionEnd doesn\'t remove open class after closing and calls notifyOpen', () => {
+test('#handleTransitionEnd doesn\'t remove open class after opening and calls notifyOpen', () => {
   const {foundation, mockAdapter} = setupTest();
-  td.when(mockAdapter.hasClass(cssClasses.ANIMATING_CLOSE)).thenReturn(false);
-  foundation.handleTransitionEnd();
+  const mockEventTarget = bel`<div>root</div>`;
+  td.when(mockAdapter.eventTargetHasClass(mockEventTarget, cssClasses.ROOT)).thenReturn(true);
+  td.when(foundation.isClosing(cssClasses.OPEN)).thenReturn(false);
```
kfranqueiro: `isClosing` would never be passed an argument, right?

test/unit/mdc-drawer/dismissible.foundation.test.js
```diff
@@ -181,93 +117,102 @@ test(`#isOpen returns false when it lacks ${cssClasses.OPEN} class`, () => {
...
+  td.verify(mockAdapter.removeClass(cssClasses.CLOSING), {times: 1});
 });
 
 test('#handleTransitionEnd removes open class after closing and calls notifyClose', () => {
   const {foundation, mockAdapter} = setupTest();
-  td.when(mockAdapter.hasClass(cssClasses.ANIMATING_CLOSE)).thenReturn(true);
-  foundation.handleTransitionEnd();
+  const mockEventTarget = bel`<div>root</div>`;
+  td.when(mockAdapter.eventTargetHasClass(mockEventTarget, cssClasses.ROOT)).thenReturn(true);
+  td.when(foundation.isClosing(cssClasses.OPEN)).thenReturn(true);
```
kfranqueiro: `isClosing` would never be passed an argument, right?

test/unit/mdc-drawer/mdc-drawer.test.js
```diff
@@ -104,35 +73,36 @@ test('#set open false calls foundation.close', () => {
 });
 
 test('keydown event calls foundation.handleKeydown method', () => {
-  const {component} = setupTest();
+  const {component, drawer} = setupTest();
   component.foundation_.handleKeydown = td.func();
-  component.initialSyncWithDOM();
-  domEvents.emit(document, 'keydown');
+  drawer.querySelector('.demo-drawer-list-item').focus();
```
kfranqueiro: Is there a reason we actually need the `demo-drawer-list-item` class in the fixture at all? Can we just use `mdc-list-item`? (I realize it was there prior to this PR)

test/unit/mdc-drawer/mdc-drawer.test.js
```diff
@@ -104,35 +73,36 @@ test('#set open false calls foundation.close', () => {
...
   domEvents.emit(drawer, 'transitionend');
   td.verify(component.foundation_.handleTransitionEnd(td.matchers.isA(Object)), {times: 1});
 });
 
 test('#destroy removes keydown event listener', () => {
-  const {component} = setupTest();
-  const originalRemove = document.removeEventListener;
-  document.removeEventListener = td.func();
+  const {component, drawer} = setupTest();
+  const originalRemove = drawer.removeEventListener;
```
kfranqueiro: There's probably a more natural way of performing these event handler tests... see example in menu: https://github.com/material-components/material-components-web/blob/1245573ea4350682b6548af89a43171302624e63/test/unit/mdc-menu/mdc-menu.test.js#L126-L140

demos/drawer/standard-drawer.html
```diff
@@ -33,7 +33,7 @@
         <h3 class="mdc-drawer__title">Mail</h3>
         <h6 class="mdc-drawer__subtitle">email@material.io</h6>
       </div>
-      <div class="mdc-drawer__scrollable">
+      <div class="mdc-drawer__content">
```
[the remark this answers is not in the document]
moog16: gotcha - I remember lynn saying that __content is too generic. But yes technically you're right @abhiomkar

demos/drawer/modal-drawer.html
```diff
@@ -0,0 +1,200 @@
+<!DOCTYPE html>
```
moog16: i think this should be a different pr

packages/mdc-drawer/dismissible/foundation.js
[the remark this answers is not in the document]
moog16: This still needs to be updated*

abhiomkar:
@moog16 Fixed the scrolling issue, this was happening on other browsers too not just IE11.

packages/mdc-drawer/dismissible/foundation.js
```diff
@@ -55,9 +54,13 @@ class MDCDismissibleDrawerFoundation extends MDCFoundation {
...
       return;
     }
+
     this.adapter_.addClass(cssClasses.OPEN);
-    this.adapter_.addClass(cssClasses.ANIMATING_OPEN);
-    this.animateAppContent_(true);
+    this.adapter_.addClass(cssClasses.ANIMATE);
+
+    // Runs this after Style and Layout information have been calculated and the paint has occurred.
+    requestAnimationFrame(() => setTimeout(() => this.adapter_.addClass(cssClasses.OPENING), 0));
```
kfranqueiro: The idea with rAF was that setTimeout wouldn't be needed... is it still needed here? Or can we force recalc and not require waiting at all? (e.g. in https://github.com/material-components/material-components-web/blob/master/packages/mdc-tab-indicator/sliding-foundation.js)

test/unit/mdc-drawer/dismissible.foundation.test.js
```diff
@@ -33,23 +32,6 @@ const setupTest = () => {
   return {foundation, mockAdapter};
 };
 
-test('exports strings', () => {
```
kfranqueiro: Why were these 3 tests removed? They should exist for any foundation.

test/unit/mdc-drawer/mdc-drawer.test.js
```diff
@@ -104,35 +76,34 @@ test('#set open false calls foundation.close', () => {
...
 test('#destroy removes keydown event listener', () => {
-  const {component} = setupTest();
-  const originalRemove = document.removeEventListener;
-  document.removeEventListener = td.func();
+  const {component, drawer, mockFoundation} = setupTest();
   component.destroy();
-  td.verify(document.removeEventListener('keydown', td.matchers.isA(Function)), {times: 1});
-  document.removeEventListener = originalRemove;
+
+  domEvents.emit(drawer, 'keydown');
```
kfranqueiro: Does this need a focus call first to be consistent with the other keydown test? (i.e. would this not even call handleKeydown even if it weren't destroyed?)

packages/mdc-drawer/dismissible/foundation.js
moog16: Why does this need to be in a setTimeout?

abhiomkar: As per MDN documentation:

> It's also possible to queue JavaScript to run very soon after the frame has been painted, where the likelihood is highest that the DOM has not been written to, and layout and style information queries are still cheap. This can be done by using a setTimeout or dispatching a runnable inside a requestAnimationFrame callback.

https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Performance_best_practices_for_Firefox_fe_engineers#Detecting_and_avoiding_synchronous_reflow#How_do_I_avoid_triggering_uninterruptible_reflow#How_do_I_avoid_triggering_uninterruptible_reflow

Just using `requestAnimationFrame` didn't work with Firefox browser, I'm suspecting the callback is executed even before the layout is complete. So using this alternative to wait for the layout / paint job complete from previous step.

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,72 @@ npm install @material/drawer
 
 ### Drawers with separate list groups
 
-If you need multiple list groups within the drawer please see the [list readme](https://github.com/material-components/material-components-web/tree/master/packages/mdc-list#list-groups)
-for detailed use and instruction.
+```html
+<nav class="mdc-drawer">
+  <div class="mdc-drawer__content">
+    <nav class="mdc-list">
+      <a class="mdc-list-item mdc-list-item--activated" href="#" tabindex="0">
```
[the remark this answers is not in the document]
abhiomkar: MDC List automatically sets tabindex to selected list item. Removed the need of explicitly adding `tabindex` attribute. `tabindex` is required if list has no selection state so user can reach to first list item using TAB key.

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,72 @@ npm install @material/drawer
...
+> NOTE: First navigation item in list group should explicitly set tabindex to '0'.
+
+List group needs to be instantiated for keyboard navigation.
+
+```javascript
+var listEl = drawerEl.querySelector('.mdc-list');
+
+var list = mdc.list.MDCList.attachTo(listEl);
+list.singleSelection = true;
+list.vertical = true;
```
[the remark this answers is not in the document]
abhiomkar: Yes, it is. Removed.

packages/mdc-drawer/README.md
```diff
@@ -60,21 +60,72 @@ npm install @material/drawer
...
+```
+
+Please see the [list readme](../mdc-list#list-group)
```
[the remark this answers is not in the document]
abhiomkar: Removed.

packages/mdc-drawer/dismissible/foundation.js
abhiomkar: Please see my [comment](https://github.com/material-components/material-components-web/pull/3340/files#r210674599) below.

Firefox browser executes the rAF callback before layout / paint where but works on Chrome without `setTimeout`. Causing force reflow works but wanted to additional reflow.

test/unit/mdc-drawer/dismissible.foundation.test.js
abhiomkar: I don't see any value in these tests. It is like testing the existence of class properties, there is no business logic involved. is it really needed?

test/unit/mdc-drawer/mdc-drawer.test.js
abhiomkar: Done.

demos/drawer/standard-drawer.html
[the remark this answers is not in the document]
abhiomkar: Any better name for this?

packages/mdc-drawer/dismissible/foundation.js
moog16: @abhiomkar Thanks for sharing. That makes sense and is interesting. This could lead to issues in the react component if the drawer is removed from the DOM while opening (for whatever reason), but I think I have a workaround.

I think the only change I suggest here is: `setTimeout(func, 0)` --> `setTimeout(func)`. You can omit the `0`

kfranqueiro: Is my idea really adding additional reflow, or is it just making the same reflow happen sooner? If it works it would simplify test cases and possibly also reduce potential for issues in framework wrappers...

demos/drawer/standard-drawer.html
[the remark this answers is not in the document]
moog16: Most of the time a list should go into this div right? I don't think there is anything else. What about `mdc-drawer__list-area`?

test/unit/mdc-drawer/dismissible.foundation.test.js
kfranqueiro: Given that I just cut a patch release yesterday with a fix related to us missing one of these, I would argue it is warranted :) see https://github.com/material-components/material-components-web/pull/3346

packages/mdc-drawer/dismissible/foundation.js
abhiomkar: If I omit delay parameter browser adds default delay which is more than `0`. So I think it is safer to have `0` delay than a default delay.

That is an interesting use case. If the drawer is removed from DOM immediately after open, I'm guessing it happens after above callback is complete since this is queued first. I just tested this didn't have any issues.

    drawer.open = true;
    drawerEl.remove()

test/unit/mdc-drawer/dismissible.foundation.test.js
abhiomkar: Ah ok! Added these tests back.

Thanks!

demos/drawer/standard-drawer.html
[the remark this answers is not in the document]
abhiomkar: It may also have footer content which is not a list. I'm leaving it as `__content` for now. MDC Tab uses `__content` too.

packages/mdc-drawer/dismissible/foundation.js
kfranqueiro: I suspect Matt is coming from the perspective of MDC React where destroying something while a rAF/setTimeout is still pending might cause issues unless they tiptoe around `addClass` to guard against it being called after/during destroy.

It might help if we stored references to the scheduled rAF and setTimeout and cleared them in destroy.

abhiomkar: As per MDN documentation it says rAF with setTimeout set to `0` causes it to execute immediately after previous layout + paint is complete.

It would be helpful to look at an example where it might cause an issue.

abhiomkar: Found an edge case where using rAF + setTimeout might be an issue.

    drawer.open // true
    drawer.isOpening // false - because rAF callback is still pending

This wouldn't happen if we force reflow. Added new foundation method `computeBoundingRect` for reflow. PTAL.

Thanks!

packages/mdc-drawer/README.md
```diff
@@ -272,6 +272,7 @@ Method Signature | Description
 `hasClass(className: string) => boolean` | Returns true if the root element contains the given `className`.
 `removeClass(className: string) => void` | Removes a class from the root element.
 `eventTargetHasClass(targetElement: !Element, className: string) => boolean` | Returns true if the target element contains the given class.
+`computeBoundingRect() => !ClientRect` | Returns the ClientRect for the surface.
```
kfranqueiro: For clarity: surface -> root element

test/unit/mdc-drawer/dismissible.foundation.test.js
```diff
@@ -45,8 +45,8 @@ test('exports cssClasses', () => {
 
 test('defaultAdapter returns a complete adapter implementation', () => {
   verifyDefaultAdapter(MDCDismissibleDrawerFoundation, [
-    'hasClass', 'addClass', 'removeClass', 'computeBoundingRect', 'setStyleAppContent',
-    'addClassAppContent', 'removeClassAppContent', 'isRtl', 'notifyClose', 'notifyOpen',
+    'hasClass', 'addClass', 'removeClass', 'eventTargetHasClass', 'computeBoundingRect', 'saveFocus', 'restoreFocus',
```
kfranqueiro: Also add a test for `computeBoundingRect` to the component unit test module for the vanilla implementation

packages/mdc-drawer/_mixins.scss
```diff
@@ -123,3 +122,11 @@
...
   }
 }
+
+@mixin mdc-drawer-subtext-ink-color($color, $opacity) {
+  $value: rgba(mdc-theme-prop-value($color), $opacity);
+
+  .mdc-drawer__subtext {
+    @include mdc-theme-prop(color, $value);
+  }
+}
```
kfranqueiro: Discussed offline: merge this back together with subtitle, and rename subtitle-text-color to subtitle-ink-color

Also make sure there's a newline at the end of the file after reverting (lint caught it)

packages/mdc-drawer/_variables.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2017 Google Inc. All Rights Reserved.
+// Copyright 2018 Google Inc. All Rights Reserved.
```
kfranqueiro: Revert this, since this file already existed

packages/mdc-drawer/standard/mdc-drawer-standard.scss
```diff
@@ -45,52 +56,78 @@
...
   }
 }
 
 .mdc-drawer__header {
+  flex-shrink: 0;
   box-sizing: border-box;
   min-height: 64px; // same as $mdc-top-app-bar-row-height
-  padding: $mdc-drawer-surface-padding;
-  padding-bottom: 4px;
+  padding: 0 $mdc-drawer-surface-padding 4px $mdc-drawer-surface-padding;
```
kfranqueiro: Nit: last value can be dropped

packages/mdc-drawer/_mixins.scss
```diff
@@ -73,32 +75,26 @@
 
 @mixin mdc-drawer-icon-activated-ink-color($color, $opacity) {
   $value: rgba(mdc-theme-prop-value($color), $opacity);
-  // stylelint-disable selector-no-qualifying-type
-  .mdc-list-item--activated,
-  a.mdc-list-item--activated, {
+
+  .mdc-list-item--activated {
```
kfranqueiro: I noticed you took out `a` selectors here, but there's still one in `mdc-drawer-standard.scss`; is it still needed?

test/screenshot/spec/mdc-drawer/fixture.scss
```diff
@@ -35,14 +42,38 @@ $custom-drawer-color: $material-color-orange-900;
...
-.custom-drawer--ink-color {
-  @include mdc-drawer-ink-color($custom-drawer-color);
+.test-main-content {
+  overflow: auto;
+  height: 100%;
+  padding-left: 16px;
+  padding-right: 16px;
 }
+
+.custom-drawer--fill-color {
```
kfranqueiro: Typically we aim for each screenshot test class to apply exactly one mixin. Thus I'd aim for these to only apply the relevant mixin, as they were doing before. We can add more screenshot test pages for the other mixins later.

packages/mdc-drawer/adapter.js
```diff
@@ -48,40 +48,33 @@ class MDCDrawerAdapter {
   hasClass(className) {}
 
   /**
-   * Returns ClientRect of the root Element.
-   * @return {!ClientRect}
+   * @param {!Element} targetElement target element
+   * @param {string} className class name
    */
-  computeBoundingRect() {}
+  eventTargetHasClass(targetElement, className) {}
```
patrickrodee: Nit: rename to "elementHasClass"

packages/mdc-drawer/index.js
```diff
@@ -91,28 +93,30 @@ export class MDCDrawer extends MDCComponent {
...
       computeBoundingRect: () => this.root_.getBoundingClientRect(),
-      addClassAppContent: (className) => {
-        if (this.appContent_) {
-          this.appContent_.classList.add(className);
+      saveFocus: () => {
+        this.previousFocus_ = document.activeElement;
+      },
+      restoreFocus: () => {
+        if (this.root_.contains(document.activeElement)) {
+          if (this.previousFocus_ && this.previousFocus_.focus) {
```
patrickrodee: Nit: collapse these into one `if` branch since there's no else case

PR #3342, opened 2018-08-14
feat(switch): Move component specific logic out of foundation

lynnjepsen:


lynnjepsen:
Re: #3147, "Firefox would remember the state of the hidden input element on refresh "

How do I reproduce this bug?

lynnjepsen:
Okay i fixed this bug where the checked state can be out of sync with the styling. I moved the logic to the vanilla component in the `initialSyncWithDom` method

kfranqueiro:
FYI I think this needs a `BREAKING CHANGE:` when merged because it removes 2 foundation APIs and makes `handleChange` require a parameter.

PR #3351, opened 2018-08-14
fix(auto-init): added tab, tabScroller, tabIndicator to auto init

moog16:


kfranqueiro:
When we were looking at updating auto-init registrations, we decided to intentionally omit these because it doesn't make sense to auto-init just a tab, tab scroller, or tab indicator. When you auto-init a tab bar, it will instantiate all of these components itself.

PR #3360, opened 2018-08-15
Add installation instructions to floating-label and line-ripple

ALMaclaine:
I added installation instructions to the other two places it was missing, floating-label and line-ripple.

For some reason the notched-outline commit is in here as well. Let me know if I need to change anything in regards to that.

### DIFF
--- a/packages/mdc-floating-label/README.md
+++ b/packages/mdc-floating-label/README.md
@@ -21,6 +21,12 @@ Floating labels display the type of input a field requires. Every Text Field and
   </li>
 </ul>
 
+## Installation
+
+```
+npm install @material/floating-label
+```
+
 ## Basic Usage
 
 ### HTML Structure
--- a/packages/mdc-line-ripple/README.md
+++ b/packages/mdc-line-ripple/README.md
@@ -21,6 +21,12 @@ The line ripple is used to highlight user-specified input above it. When a line 
   </li>
 </ul>
 
+## Installation
+
+```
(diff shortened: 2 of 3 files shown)


kfranqueiro:
First of all, thanks for spotting these!

The notched-outline commit is still there because your master branch is out of sync (and diverged) with our repository's master branch.

We highly recommend sending PRs from separate branches, especially since when we merge PRs, we squash-merge them, which means the commit hash will not match the original commit from the PR, so it isn't seen as the same commit.

In order to get back in sync, you'll likely need to do this:

```
git fetch upstream
git reset --hard upstream/master
```

Then you should be able to create a branch, cherry-pick the two commits (which will have been orphaned when you reset above), and push it:

```
git checkout -b docs/install-instructions
git cherry-pick 41f87ad 984345b
git push -u origin docs/install-instructions
```

(`checkout` with `-b` creates a new branch at the current point in history; just `checkout` checks out an existing branch/commit/tag. The `-u` option to `push` sets up the branch so that you can just `git push` subsequently if you need to push any more commits on the same branch.)

Then you can send a new PR using that branch, and close this one.

ALMaclaine:
Thank you so much for this detailed response on how to fix my divergence issue. This is the first time I've contributed to open source or really done anything this complicated on github and I really appreciate the guidance. I had lots of trepidation about messing something up, so thank you.

Closing this pr and doing a pr from the branch docs/install-instructions

PR #3362, opened 2018-08-15
feat(drawer): Add modal variant

### LINKED ISSUE #2875: New `mdc-drawer--modal`
- [x] New `mdc-drawer-scrim-fill-color` Sass mixin
- [x] New HTML structure for this variant, including `mdc-drawer__scrim`
- [x] New `MDCModalDrawerFoundation` which extends `MDCDismissibleDrawerFoundation`
- [x] Documentation notes that `mdc-drawer--modal` cannot be used with `mdc-drawer--rail`

abhiomkar:
Fixes #2875


packages/mdc-drawer/README.md
```diff
@@ -283,15 +359,29 @@ Method Signature | Description
...
-### Foundations: `MDCDismissibleDrawerFoundation`
+### Foundations
+
+#### `MDCDismissibleDrawerFoundation`
 
 Method Signature | Description
 --- | ---
 `open() => void` | Opens the drawer from the closed state.
 `close() => void` | Closes the drawer from the open state.
+`opened() => void` | Abstract method which gets called when drawer finished opening.
```
kfranqueiro: I'd use the word "Extensible" for these rather than "Abstract".

Shouldn't these _not_ technically be `@abstract` on dismissible drawer? They're just no-ops for it. Presence of abstract methods normally suggests the class can't be instantiated, doesn't it?

packages/mdc-drawer/_variables.scss
```diff
@@ -41,5 +41,9 @@ $mdc-drawer-surface-padding: 16px;
 $mdc-drawer-animation-enter: 250ms;
 $mdc-drawer-animation-exit: 200ms;
 
-$mdc-drawer-z-index: 5;
+// Scrim
+$mdc-drawer-modal-scrim-color: #000 !default;
```
kfranqueiro: I think we'd want to encourage use of the mixin rather than make these overridable. We're pretty selective about adding `!default` and it tends to more often apply to systems.

packages/mdc-drawer/modal/foundation.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import MDCDrawerAdapter from '../adapter';
+import MDCDismissibleDrawerFoundation from '../dismissible/foundation';
+
+/**
+ * @extends {MDCDismissibleDrawerFoundation<!MDCDrawerAdapter>}
+ */
+class MDCModalDrawerFoundation extends MDCDismissibleDrawerFoundation {
+  static get defaultAdapter() {
```
kfranqueiro: Shouldn't we update `MDCDismissibleDrawerFoundation`'s `defaultAdapter` to include everything and then we don't need this? (That's what e.g. top app bar does.)

It doesn't really make sense to have partial default adapters when there's only one actual adapter interface which includes everything. If there are variants that don't make use of everything, we tend to include guards within the adapter implementations to ensure they no-op rather than cause an error (e.g. in text field we'll check for existence of the line ripple or notched outline components).

packages/mdc-drawer/modal/mdc-drawer-modal.scss
```diff
@@ -0,0 +1,66 @@
...
+  &.mdc-drawer--open {
+    display: block;
+  }
+}
+
+.mdc-drawer-scrim {
+  display: none;
+  position: fixed;
+  top: 0;
+  right: 0;
```
kfranqueiro: I think you can omit right/bottom since you have width/height? (At least that's what I've seen in our existing code for dialog and drawer)

packages/mdc-drawer/package-lock.json
```diff
@@ -0,0 +1,42 @@
+{
```
kfranqueiro: I'm curious, did this file become a thing just from a top-level npm install, or did an npm install get performed within packages/mdc-drawer? I've seen this pop up a couple of times in specific packages but most of our packages don't have them, and we've gotten in the habit of removing them from inner folders (I've seen git tend to trip over them when switching between feature branches).

test/unit/mdc-drawer/modal.foundation.test.js
```diff
@@ -0,0 +1,42 @@
...
+  const {foundation, mockAdapter} = setupTest();
+  foundation.opened();
+  td.verify(mockAdapter.trapFocusOnSurface(), {times: 1});
+});
+
+test('#closed untraps the focus when drawer finishes animating close', () => {
+  const {foundation, mockAdapter} = setupTest();
+  foundation.closed();
+  td.verify(mockAdapter.untrapFocusOnSurface(), {times: 1});
+});
```
kfranqueiro: Needs test for `handleScrimClick`?

Also do we need component tests to test modal vs. dismissible code paths?

packages/mdc-drawer/README.md
```diff
@@ -121,9 +127,15 @@ Drawers can contain a header element which will not scroll with the rest of the
   </div>
   <div class="mdc-drawer__content">
     <nav class="mdc-list">
-      <a class="mdc-list-item mdc-list-item--activated" href='#'>
+      <a class="mdc-list-item mdc-list-item--selected" href='#' tabindex="0">
```
kfranqueiro: I noticed you reverted `selected` to `activated` where I commented, but there are still a bunch of changes from `activated` to `selected` in this PR.

packages/mdc-drawer/README.md
```diff
@@ -234,6 +251,35 @@ topAppBar.listen('MDCTopAppBar:nav', () => {
 });
 ```
 
+### Modal Drawer
+
+Modal drawers are elevated above most of the app’s UI and don’t affect the screen’s layout grid.
+
+```html
+<body>
+  <header class="mdc-drawer mdc-drawer--modal">
```
kfranqueiro: The root element is an `aside` in the demo page, but it's a `header` here?

There's also a case of this under Dismissible, and also a case of an outer `nav` (when the list inside content is also a `nav`).

packages/mdc-drawer/README.md
```diff
@@ -234,6 +251,35 @@ topAppBar.listen('MDCTopAppBar:nav', () => {
...
+      </nav>
+    </div>
+  </header>
+
+  <div class="mdc-drawer-scrim"></div>
+  <div>Main Content</div>
+</body>
+```
+
+> Use the `mdc-drawer-scrim` class on next sibling element to add backdrop to block app's UI.
```
kfranqueiro: I think backdrop might potentially mean something else in Material terms...

Maybe:

```
> The `mdc-drawer-scrim` next sibling element protects the app's UI from interactions while the drawer is open.
```

packages/mdc-drawer/README.md
```diff
@@ -246,10 +292,13 @@ Class | Description
 `mdc-drawer__title` | Title text element of the drawer.
 `mdc-drawer__subtitle` | Subtitle text element of the drawer.
 `mdc-drawer--dismissible` | Dismissible drawer variant class.
+`mdc-drawer--modal` | Modal drawer variant class.
 `mdc-drawer--open` | If present, indicates that the dismissible drawer is in the open position.
 `mdc-drawer--opening` | Applied while the drawer is animating from the closed to the open position.
 `mdc-drawer--closing` | Applied while the drawer is animating from the open to the closed position.
 `mdc-drawer-app-content` | Dismissible variant only. Sibling element that is resized when the drawer opens/closes.
+`mdc-drawer-scrim` | Modal variant only. Used for backdrop to overlay on the app content.
```
kfranqueiro: "Used for the overlay on the app content." (avoiding the word backdrop again)

packages/mdc-drawer/README.md
```diff
@@ -301,15 +351,29 @@ Method Signature | Description
 `focusActiveNavigationItem() => void` | Focuses the active / selected navigation item.
 `notifyClose() => void` | Emits the `MDCDrawer:close` event.
 `notifyOpen() => void` | Emits the `MDCDrawer:open` event.
+`trapFocus() => void` | Traps focus on root element and focuses the active navigation element.
+`releaseFocus() => void` | Releases focus on root element which trapFocus was previously called on and resumes focus to nav button.
```
kfranqueiro: This technically restores focus to whatever had it before, right? Maybe we should loosen the wording?

"Releases focus from the root element and restores it to where it was prior to calling `trapFocus`."

(Also update adapter doc comment)

packages/mdc-drawer/package-lock.json
```diff
@@ -0,0 +1,110 @@
...
```
kfranqueiro: I think you previously removed this on the other branch? Not sure if that was with the intent of adding it here or removing it altogether.

test/unit/mdc-drawer/mdc-drawer.test.js
```diff
@@ -27,14 +27,20 @@ import domEvents from 'dom-events';
...
 import {MDCDrawer} from '../../../packages/mdc-drawer';
-import {strings} from '../../../packages/mdc-drawer/constants';
+import * as util from '../../../packages/mdc-drawer/util';
+import {strings, cssClasses} from '../../../packages/mdc-drawer/constants';
 import {MDCListFoundation} from '../../../packages/mdc-list';
 import MDCDismissibleDrawerFoundation from '../../../packages/mdc-drawer/dismissible/foundation';
 
-function getFixture() {
+function getFixture(variantClass) {
+  let scrimEl;
```
kfranqueiro: Should this be initialized to `''`? Will it produce `'undefined'` below for dismissible? (Though I suppose it won't affect anything)

test/unit/mdc-drawer/mdc-drawer.test.js
```diff
@@ -44,22 +50,28 @@ function getFixture() {
...
+
+function hasClassMatcher(className) {
+  return td.matchers.argThat((el) => el.classList && el.classList.contains(className));
+}
+
 suite('MDCDrawer');
 
 test('attachTo initializes and returns a MDCDrawer instance', () => {
-  const drawer = getFixture().querySelector('.mdc-drawer');
+  const drawer = getFixture(cssClasses.DISMISSIBLE).querySelector('.mdc-drawer');
```
kfranqueiro: Out of curiosity, what happens right now if someone mistakenly instantiates Drawer JS for a standard drawer? I'm wondering if we can do anything to clearly indicate user error in that case.

abhiomkar: Yah, it can throw an Error. Added an extra else statement to component.

Also added a test case for this.

PR #3375, opened 2018-08-16
Docs/snackbar update format

ALMaclaine:
I updated the snackbar readme to the correct format. I would say it's about 90% converted. Some of the information in the readme didn't fully fit with the format specified, but I thought better to leave it at the bottom than delete entirely.

From my understanding, the format is correct up to "Responding to a Snackbar Action", at that stage there are slight deviations from the specified format. From my understanding of the template, the example sections "Responding to a Snackbar Action" and "Keep snackbar when the action button is pressed" should be deleted, but I didn't want to do so without making an issue about it.

Also the "Usage Within Frameworks" section is not complete because snackbar does not have an adapter, so I was unsure how to grok what that section was saying because it seemed to match the foundation API. Again, didn't want to make a change without making an issue about it.

Finally, I'm not sure if the "Avoiding Flash-Of-Unstyled-Content (FOUC)" should be included either, it doesn't fit the format.

I believe this version of the readme more closely matches the prescribed format. I think it's a decent intermediate step towards fully converting the snackbar readme to the correct format. I intend to see it through regardless if this PR is accepted, but I think it would be easier to file issues on a 90% converted readme over a radically different format.


kfranqueiro:
Thanks for sending this PR, and sorry for the delay. This looks pretty good, but I did some more cleanup on top of it. I moved the avoiding FOUC section to an "additional information" subsection, and removed a couple of the sections that are redundant of documentation within the API tables. I also updated the columns in the tables to match other READMEs.

RE the adapter, snackbar doesn't (yet) have a dedicated adapter.js, but it still has an adapter API which you can see in defaultAdapter in foundation.js and implemented in index.js.

We will be updating Snackbar in the coming month, but I wanted to be able to merge this PR before we start doing that.

PR #3377, opened 2018-08-16
fix(slider): Center value label

### LINKED ISSUE #3372: New Slider: Discrete value indicator should be centered over thumb by default
Applicable only to feat/slider branch.

Currently the indicator is purposely off-center to a degree, possibly due to a misinterpretation of design intent; this should involve removing some of the math that's being performed.

bonniezhou:
Fixes #3372

### DIFF
--- a/packages/mdc-slider/foundation.js
+++ b/packages/mdc-slider/foundation.js
@@ -144,7 +144,7 @@ class MDCSliderFoundation extends MDCFoundation {
     /** @private {function(): undefined} */
     this.windowResizeHandler_ = () => this.layout();
 
-    //For testing path calculations
+    // For testing path calculations
     /** @private {number} */
     this.pathTotalHorizontalDistance_ = 0;
     /** @private {number} */
@@ -699,7 +699,7 @@ class MDCSliderFoundation extends MDCFoundation {
         this.pathLeftTopNeckCenterYPos_ = Math.sqrt(Math.pow(TOP_LOBE_RADIUS + TOP_NECK_RADIUS, 2) -
           Math.pow(MAX_TOP_NECK_WIDTH - topLobeHorizontalLeft/2, 2));
 
-        pointI.x = MAX_TOP_NECK_WIDTH - TOP_NECK_RADIUS + 
+        pointI.x = MAX_TOP_NECK_WIDTH - TOP_NECK_RADIUS +
           (Math.cos(this.pathLeftTopNeckTheta_) * TOP_NECK_RADIUS) + OFFSET_X;
         pointI.y = TOP_LOBE_RADIUS +
           (this.pathLeftTopNeckCenterYPos_ - (Math.sin(this.pathLeftTopNeckTheta_) * TOP_NECK_RADIUS)) + OFFSET_Y;
@@ -771,7 +771,7 @@ class MDCSliderFoundation extends MDCFoundation {
     let extraTranslateValue = 0;
     const topLobeHorizontal = this.value_.toString().length > 2 ? localeStringWidth - (2 * digitWidth) : 0;
     const extraHorizontalWidth = topLobeHorizontal - 30;
-    if(extraHorizontalWidth > 0) {
+    if (extraHorizontalWidth > 0) {
       const extraHorizontalWidthSide = extraHorizontalWidth / 2;
       if (translatePx - MAX_TOP_NECK_WIDTH < extraHorizontalWidthSide) {
         extraTranslateValue = extraHorizontalWidthSide - translatePx + MAX_TOP_NECK_WIDTH;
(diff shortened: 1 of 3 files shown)


packages/mdc-slider/foundation.js
```diff
@@ -753,8 +748,8 @@ class MDCSliderFoundation extends MDCFoundation {
       topLobeHorizontal = localeStringWidth - (2 * digitWidth);
     }
     const extraHorizontalWidth = topLobeHorizontal - 30;
-    let extraHorizontalWidthLeft = extraHorizontalWidth * 3 / 4;
-    let extraHorizontalWidthRight = extraHorizontalWidth / 4;
+    let extraHorizontalWidthLeft = extraHorizontalWidth / 2;
```
kfranqueiro: Can this be further simplified to just use a single variable in this function (e.g. `const extraHorizontalWidthSide = (topLobeHorizontal - 30) / 2`)? It doesn't look like the left/right values are further modified dependent on left/right, and you'd also be able to remove the isRTL block.

bonniezhou:
Sorry to have left this hanging for so long! I edited a bunch of the constant names and comments in `calcPath_` and `calcValueLabelTextXValue_` so they're easier to understand.
Also added some very basic units tests for the math calculations...it's hard to be comprehensive but these will hopefully make sure that if someone edits the math they have to know what they're doing.

packages/mdc-slider/foundation.js
bonniezhou: Done!

packages/mdc-slider/foundation.js
```diff
@@ -577,146 +594,156 @@ class MDCSliderFoundation extends MDCFoundation {
...
-      const leftTopNeckCornerTheta =
-        Math.acos((topNeckArcWidth - (topLobeHorizontalRight)/2)/(topLobeRadius+topNeckRadius));
-      const leftTopNeckCornerCenterY = Math.sqrt(Math.pow(topLobeRadius+topNeckRadius, 2) -
-        Math.pow(topNeckArcWidth - (topLobeHorizontalRight)/2, 2));
-
-      pointA.x = rightStemX + (topNeckRadius - (Math.cos(leftTopNeckCornerTheta) * topNeckRadius)) + offsetX;
-      pointA.y =
-        topLobeRadius + (leftTopNeckCornerCenterY - (Math.sin(leftTopNeckCornerTheta) * topNeckRadius)) + offsetY,
-      pointB.y = topLobeRadius + leftTopNeckCornerCenterY + offsetY;
+    // When the slider reaches the edge of the slider, the extra horizontal distance on the side closest to the edge
```
kfranqueiro: I'm wondering whether the end-of-slider case is triggering correctly in all instances it should? On http://localhost:8080/slider.html if I look at the first "large values" slider, it always remains centered, whereas the last one shifts the pointer in an effort to keep the balloon mostly in line with the slider. I would think the latter should happen in both cases.

test/unit/mdc-slider/foundation-value-label-path.test.js
```diff
@@ -0,0 +1,115 @@
...
+test('path calculated correctly with label at beginning of slider', () => {
+  const {foundation, mockAdapter} = setupTest();
+  td.when(mockAdapter.getDigitWidth()).thenReturn(DIGIT_WIDTH);
+  td.when(mockAdapter.getCommaWidth()).thenReturn(COMMA_WIDTH);
+  foundation.rect_ = {width: 1000};
+
+  foundation.value_ = 10000000000; // 10,000,000,000
+  const numCommas = 3;
+  const labelHorizontalWidth = ('10000000000'.length - 2) * DIGIT_WIDTH + numCommas * COMMA_WIDTH;
+  const topLobeHorizontal = Math.min(labelHorizontalWidth, MAX_TOP_LOBE_HORIZONTAL);
```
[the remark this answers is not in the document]
bonniezhou: The asserts are different: in the 1st test `pathLeftTopNeckTheta_` and `pathLeftTopNeckCenterYPos_` are adjusted and in the 3rd test the `pathRightTopNeckTheta_` and `pathRightTopNeckCenterYPos_` are adjusted

packages/mdc-slider/foundation.js
bonniezhou: This is actually pretty difficult because the way it's set up now is that first the bubble expands to be as wide as possible without necessitating any extra width at the "neck", second to add that extra width if necessary, and lastly shift the bubble back to be in line with the slider. So this should probably be in a separate PR/issue.

kfranqueiro: Entered #4314.

moog16:
Going to close the slider PR, as it's gone stale. We will leave the branch open for future use.

PR #3386, opened 2018-08-20
fix(tab-bar): Early exit

bonniezhou:
Move the early exit for activating a tab from `Tab` to `TabBar`. 

Adapter methods may be asynchronous in other frameworks, resulting in a race condition when `deactivateTabAtIndex` and `activateTabAtIndex` are called synchronously in `MDCTabBarFoundation.activateTab`.


packages/mdc-tab/foundation.js
```diff
@@ -92,11 +92,6 @@ class MDCTabFoundation extends MDCFoundation {
    * @param {!ClientRect=} previousIndicatorClientRect
    */
   activate(previousIndicatorClientRect) {
-    // Early exit
-    if (this.isActive()) {
```
patrickrodee: Does this have any problems if an already active tab is activated again when the tab is outside of the tab bar?

bonniezhou: Nope it looks fine, based on the demos page which calls `activate` on the tab directly (without a tab bar)

Issue #3383, opened 2018-08-17
MDC List: Make activated list item behave similar to selected list item for singleSelection


abhiomkar:
MDC List item currently has `--activated` state which stays activated till lifetime of a page. Activated state should be behave similar to selected state for single selection item. List should update the activated state for respective list item when user clicks.

Issue #3361, opened 2018-08-15
MDC List: Allow disabling toggle for single select variant


abhiomkar:
Right now the single select variant of MDC list allows user to deselect activated / selected, it needs a support to disabling toggle of selected item so that MDC list can always have one selected item.

PR #3388, opened 2018-08-20
fix(list): Change default class to activated

williamernest:
fixes: #3383, fixes: #3361 

This changes the default class to add/remove from `--selected` to `--activated`. It preserves the ability to use `--selected` by adding a foundation variable `useSelectedClass_` (and setter methods). 

Removes the ability to toggle the activated class off by clicking the same element twice.


packages/mdc-list/README.md
```diff
@@ -134,7 +134,7 @@ OR
 ### Single Selection List
 
 MDC List can handle selecting/deselecting list elements based on click or keyboard action. When enabled, the `space` and `enter` keys (or `click` event) will trigger an
```
abhiomkar: > selecting/deselecting

is this still relevant?

packages/mdc-list/README.md
```diff
@@ -267,10 +268,10 @@ these should also receive `tabIndex="-1"`.
 #### Setup in `singleSelection()`
 
 When implementing a component that will use the single selection variant, the HTML should be modified to include
-the `aria-selected` attribute, the `mdc-list-item--selected` class should be added, and the `tabindex` of the selected
-element should be `0`. The first list item should have the `tabindex` updated to `-1`. The foundation method 
-`setSelectedIndex()` should be called with the initially selected element immediately after the foundation is 
-instantiated.
+the `aria-selected` attribute, the `mdc-list-item--selected` or `mdc-list-item--activated` class should be added, 
+and the `tabindex` of the selected element should be `0`. The first list item should have the `tabindex` updated
```
abhiomkar: is it mandator for user to set `tabindex` -1 explicitly on first element? Doesn't list automatically sets tabindex to -1 to all other list items?

packages/mdc-list/index.js
```diff
@@ -78,6 +79,21 @@ class MDCList extends MDCComponent {
...
+    // Automatically set single selection if selected/activated classes are present.
+    const preselectedElement =
+      this.root_.querySelector(`.${cssClasses.LIST_ITEM_ACTIVATED_CLASS}, .${cssClasses.LIST_ITEM_SELECTED_CLASS}`);
+
+    if (preselectedElement) {
+      if (preselectedElement.classList.contains(cssClasses.LIST_ITEM_ACTIVATED_CLASS)) {
+        this.foundation_.setUseActivatedClass(true);
+      }
+
+      this.singleSelection = true;
```
abhiomkar: is it safe to assume user wants to have single selection mode when list has selected item? 

Multi-selectable list can initially have one item selected.

demos/list.html
```diff
@@ -53,7 +53,7 @@
               info
             </a>
           </li>
-          <li class="mdc-list-item">
+          <li class="mdc-list-item mdc-list-item--activated">
```
abhiomkar: It would be helpful to have separate demos for following cases:

- singleSelect: with `--selected`
- singleSelect: with `--activated`
- multiSelect

packages/mdc-list/README.md
williamernest: Yes, when clicked it will deselect the previously selected list item and select the currently selected list item.

packages/mdc-list/index.js
williamernest: I think it is safe to assume that the user has a single selection list when using the selected or activated classes. There isn't a multi-selectable list, but this can be changed if we implement one.

demos/list.html
williamernest: I'll add a demo for `--activated`.

packages/mdc-list/README.md
williamernest: If the user has marked an element as `selected`, they should render the first list element to `tabindex=-1`. The foundation may not be instantiated immediately when the page is loaded (such as in Wiz).

demos/list.html
abhiomkar: `aria-selected` and `tabindex` attr are missing for activated item.

PR #3390, opened 2018-08-22
Add missing material-icons class to tab-bar example

chalin:
AFAICT the icon won't render w/o the `material-icons` class. /cc @filiph @Sfshaza


kfranqueiro:
Huh. I thought I'd come across these and fixed them, but I guess I missed some. Thanks!

PR #3394, opened 2018-08-22
fix(tab-bar): Move activateTab to adapter

bonniezhou:
Move the job of calling `activateTab` into an adapter method instead of directly in the foundation's handler methods. In the React wrapper, handler methods should only update props. `activateTab` will then be called in the `componentDidUpdate` lifecycle method. 

Also rename `getActiveTabIndex` to `getPreviousActiveTabIndex` to disambiguate. React wrapper stores the previous active index in state and returns it in this adapter method.

### DIFF
--- a/packages/mdc-tab-bar/foundation.js
+++ b/packages/mdc-tab-bar/foundation.js
@@ -121,7 +121,7 @@ class MDCTabBarFoundation extends MDCFoundation {
   }
 
   /**
-   * Sets the tab at the given index to be activated
+   * Activates the tab at the given index
    * @param {number} index
    */
   activateTab(index) {
--- a/packages/mdc-tab-bar/foundation.js
+++ b/packages/mdc-tab-bar/foundation.js
@@ -81,13 +81,13 @@ class MDCTabBarFoundation extends MDCFoundation {
       getScrollContentWidth: () => {},
       getOffsetWidth: () => {},
       isRTL: () => {},
-      activateTab: () => {},
+      setActiveTab: () => {},
       activateTabAtIndex: () => {},
       deactivateTabAtIndex: () => {},
       focusTabAtIndex: () => {},
       getTabIndicatorClientRectAtIndex: () => {},
       getTabDimensionsAtIndex: () => {},
-      getActiveTabIndex: () => {},
+      getPreviousActiveTabIndex: () => {},
       getFocusedTabIndex: () => {},
       getIndexOfTab: () => {},
       getTabListLength: () => {},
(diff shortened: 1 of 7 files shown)


kfranqueiro:
This seems super-weird from a MDC Web core perspective. We're taking what used to be one foundation method calling another, and changing it to make the foundation call the adapter to immediately call into the foundation again. Typically if anything we've been trying lately to _avoid_ "hollow" loops between component->foundation->adapter, and this seems like an even smaller loop than those.

You said that "handler methods should only update props"... can you explain or point to what exactly is happening right now and why it's a problem?

bonniezhou:
This is different from the component->foundation->adapter loops, it's more like foundation->adapter->foundation. The main reason for doing this is to separate the foundation's `activateTab` method from the foundation's handler methods.

The problem right now is that in React, the user expects to have full control over the tab bar's active index. Say the user keeps this `activeIndex` in some state variable, and passes it as a prop to the `TabBar`. Every time the `TabBar` detects that props have changed, it will call the foundation's `activateTab` (through the `componentDidUpdate` lifecycle method).

This means that tab bar's click/keydown handlers should be calling a callback that updates the user's `activeIndex` state, not calling `activateTab` directly.

kfranqueiro:
I'm trying to think of a way to formalize this so we can understand why this is done and know when we need to do it (vs. when it's not necessary). It's hard for me to pin this down generically - is it something like "if a foundation's handler needs to change state, it needs to go through an adapter"? I guess that would conform to e.g. Switch's `handleChange` applying CSS classes via the adapter...

bonniezhou:
I think "if a foundation's handler needs to change state, it needs to go through an adapter" makes sense.

Yeah foreseeing "state" is tricky. If it helps, I found an example of something that does _not_ hold state: the foundation's `scrollIntoView` method. It's similar to `activateTab` in that it's something that the user should be able to directly manipulate (via MDCW's `MDCTabBar.scrollIntoView` method or MDCR's `indexInView` prop), but it doesn't manage something that changes internally and needs to be propagated back up to the user.

packages/mdc-tab-bar/foundation.js
```diff
@@ -106,7 +106,8 @@ class MDCTabBarFoundation extends MDCFoundation {
   }
 
   init() {
-    const activeIndex = this.adapter_.getActiveTabIndex();
+    // Upon init, the active tab index is the same as the previous active tab index
+    const activeIndex = this.adapter_.getPreviousActiveTabIndex();
```
patrickrodee: This one line feels weird. All other cases seem like improvements.

kfranqueiro:
Remember `BREAKING CHANGE:` for renamed and added adapter APIs

PR #3397, opened 2018-08-22
fix(text-field): Update to match spec

williamernest:


packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -29,11 +29,8 @@
 // Notched Outline is intended for use by multiple components, but its styles should only be emitted once when bundled
 @include mdc-base-emit-once("mdc-notched-outline") {
   .mdc-notched-outline {
-    position: absolute;
-    top: 0;
-    left: 0;
-    width: calc(100% - 1px);
-    height: calc(100% - 2px);
+    @include mdc-notched-outline-base_;
```
kfranqueiro: Rather than put this in a mixin, should we separate these styles and put them under `.mdc-notched-outline, .mdc-notched-outline__idle`?

packages/mdc-textfield/_mixins.scss
```diff
@@ -371,15 +370,19 @@
 }
 
 @mixin mdc-text-field-with-trailing-icon_ {
-  @include mdc-text-field-icon-horizontal-position_(right, $mdc-text-field-icon-position, $mdc-text-field-icon-padding);
+  @include mdc-text-field-icon-horizontal-position_(right, $mdc-text-field-trailing-icon-position, $mdc-text-field-icon-padding);
+
+  // Outlined uses 16px for text alignment when using a trailing icon.
```
kfranqueiro: Where did you get this from? The spec seems to show 8px between the trailing icon and the side for outlined text field. (The spec also doesn't seem to show leading icons at all though...)

williamernest:
I confirmed with designers the actual numbers. The spec tool is broken so
they can't update the values on the site.

kfranqueiro:
I pushed a fix to the demo page since we never removed leading/trailing whitespace from the helper text elements after we had to revert to inline-block for the baseline typography mixins.

I added the bugs from the backlog that this seems to resolve, in the description.

I'm wondering if we want a `BREAKING CHANGE:` note here because of the removed margins?

williamernest:
I'll add `BREAKING CHANGE:` to the commit message since we are changing properties outside of the text field and that can break some UI's.

PR #3400, opened 2018-08-23
Allow fullscreen dialog

arpontes:
Althought noted in documentation of MDC's dialog that a full screen dialog is out of scope, there are specification in Material Design to full screen dialogs (only for mobile).
I'm proposing a change to allow bigger dialogs that become full screen on mobile.


williamernest:
Thanks for submitting this PR. 

Fullscreen dialogs are not on the roadmap for any platform, because supposedly native mobile OSes already provide similar functionality, and none of our GM clients are asking for it on the Web. In addition, fullscreen dialogs look and behave very differently than modal dialogs, so it would take more than just a few lines of CSS to fully implement it. E.g., fullscreen dialogs have a top app bar.

You're welcome to implement this for your own needs, but we'll be closing this PR until we can scope all of the work required to implement this.

Issue #3379, opened 2018-08-17
New Drawer: Spec discrepancies


kfranqueiro:
I noticed a couple of discrepancies with the spec when reviewing the current pending drawer branches:

* [x] Group labels don't have sufficient vertical padding
* [x] List dividers do not span the full drawer, because the drawer styles are adding padding to the entire mdc-list rather than just to items if necessary

PR #3404, opened 2018-08-23
feat(drawer): sass improvements & changes to match spec.

### LINKED ISSUE #3380: New Drawer: Sass Improvements
* [ ] Move `standard/mdc-drawer-standard.scss` to `mdc-drawer.scss` since its styles are actually inherited by all variants (I'd suggest doing this in its own PR before continuing to the other items here, it might make the diffs clearer)
* [x] Consider renaming `mdc-drawer-group-header-text-color` to `mdc-drawer-subtitle-ink-color` since that's what it pertains to and "ink" is consistent with other mixin names
* [ ] Several mixins take an opacity parameter but in many cases there is a sane default for this in the drawer variables; I'd be inclined to hook up those defaults within the parameters instead of passing them in explicitly from the emitted styles, so that they can automatically take effect for custom colors as well
* [x] `mdc-drawer-fill-color-accessible` should also invoke `mdc-drawer-title-ink-color` and `mdc-drawer-subtitle-ink-color` at the proper opacities

abhiomkar:
Fixes #3380 & #3379


packages/mdc-theme/_variables.scss
```diff
@@ -144,3 +150,7 @@ $mdc-theme-property-values: (
 @function mdc-theme-is-valid-theme-prop-value_($style) {
   @return type-of($style) == "color" or $style == "currentColor" or str_slice($style, 1, 4) == "var(";
 }
+
+@function mdc-theme-text-emphasis($emphasis) {
```
kfranqueiro: This seems like something we should PR as feat(theme) on master then water into this branch, so it can have its own changelog item.

Also, let's make the number formatting consistent (remove the leadaing 0 on high/medium values).

abhiomkar: Created PR #3411 for this. This diff will be removed when other PR gets into master and once we water it .

PR #3408, opened 2018-08-23
fix(checkbox): remove native control from getters/setters of foundation

### LINKED ISSUE #3396: Checkbox should not have foundation methods that proxy to adapter
Checkbox has a lot of foundation methods that proxy to adapter.getNativeControl(). These should just be methods on the component.

Example:

```js
// foundation
  // remove this method
  getValue() {
    return this.adapter_.getNativeControl_().value;
  }


// component
  // update this method:
  get value() {
     return this.nativeCb_.value;
  }
```

moog16:
refs #3396 

Will move the ObjectPropertyDescriptor override into the component in a follow up PR


moog16:
@kfranqueiro I don't want to do the changes until this PR is merged in, since it is highly dependent on this PR.

packages/mdc-checkbox/index.js
```diff
@@ -93,39 +93,59 @@ class MDCCheckbox extends MDCComponent {
...
       isAttachedToDOM: () => Boolean(this.root_.parentNode),
     });
   }
 
+  /**
+   * @return {boolean}
+   * @private
+   */
+  isIndeterminate_() {
+    return this.nativeCb_.indeterminate;
```
williamernest: Can we move these to the `get` methods and then update the adapter to use the public `get` ? Seems like we don't really need the private function.

PR #3409, opened 2018-08-23
chore(list): Move focusIn/Out logic to a new MDCListItem

lynnjepsen:
The focus logic is specific to each list item. The list item doesn't need to interact with the parent list during the focusIn/Out events. So to make the MDCList and MDCListFoundation simpler, we've moved the logic to a new MDCListItem and MDCListItemFoundation.


kfranqueiro:
You say this is making the List component/foundation simpler... but meanwhile, this is also creating a new component which requires N instances (N = number of list items) and N*2 event listeners, which didn't exist at all before. Lists are a textbook case for event delegation, which this is (partially) undoing.

I'd also argue that subcomponents inherently carry a bit of complexity, and given the small amount of code this moves, the added boilerplate likely outweighs the actual code that was separated.

Is there something driving this from the perspective of wrapper libraries? Because otherwise I'm kinda -1 to this idea from the perspective of browser resources.

kfranqueiro:
Closing this in the absence of further discussion/justification, and since it's fallen out of date by this point anyway.

PR #3414, opened 2018-08-24
fix(drawer): list-item activation demo

Megidd:
Fix the drawer list item selection/activation demo which causes strange behaviour in certain cases, like when `mdc-list-item` inside drawer has multiple children like below:

        <a class="mdc-list-item" href="#">
            <i class="material-icons mdc-list-item__graphic" aria-hidden="true">add_box</i>
            <span class='global-lang'>Hello</span>
            <span class='local-lang'>שלום</span>
        </a>

### DIFF
--- a/demos/drawer/temporary-drawer.html
+++ b/demos/drawer/temporary-drawer.html
@@ -165,7 +165,7 @@
             if (activatedItem) {
               activatedItem.classList.remove(activatedClass);
             }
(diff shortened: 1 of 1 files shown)


Megidd:
@googlebot I signed it!

williamernest:
Thanks for submitting this PR. That file no longer exists and the functionality modified has been moved into the list component.

PR #3422, opened 2018-08-26
fix(drawer): Add exports for Dismissible, Modal and util

trimox:
- Add export for `./dismissible/foundation.js`
- Add export for `./modal/foundation.js`
- Add export for `./util.js`

PR #3424, opened 2018-08-27
fix(drawer): Fix exports and closure tests

kfranqueiro:
First of all, thanks so much @trimox for spotting this in #3422 before we even released!

This builds off of #3422 to fix closure tests, and also fix unit tests to reference mocks where needed.


kfranqueiro:
@trimox can you give an OK in this PR since I based the branch off of your commits from #3422? Thanks!

PR #3427, opened 2018-08-27
docs: Update release-process docs to include fetching tags

williamernest:
I didn't have the latest tags and lerna failed as a result since I was using the `--since` option.


kfranqueiro:
I'm somewhat confused, because for me just `git pull` is enough to pull down any tags I don't already have. (I just tested by deleting the `v0.39.0` tag locally and running again.)

Out of curiosity, does `git pull --tags` make any difference? `git pull` should forward its options to `git fetch`, and `fetch --all` seems like overkill since our local master should already be pointing at the material-components remote and we don't need to download from any other forks.

PR #3450, opened 2018-08-28
chore(elevation): Delete `package-lock.json`

acdvorak:
This file is not needed because the package does not have any external npm dependencies.


kfranqueiro:
I wonder if this explains how the theme version formerly got stuck too... (#2870)

Issue #3302, opened 2018-08-06
MDCTextField: Leading and trailing icons do not render properly when used at the same time


williamernest:
When adding both a leading and trailing icon at the same time, the icons end up being stacked on top of each other on the 'end' side of the text field. 

```html
<div class="mdc-text-field mdc-text-field--outlined mdc-text-field--with-leading-icon	 mdc-text-field--with-trailing-icon">
  <i class="material-icons mdc-text-field__icon" tabindex="0" role="button">directions_transit</i>
  <input type="text" id="tf-outlined" class="mdc-text-field__input">
  <label for="tf-outlined" class="mdc-floating-label">Your Name</label>
  <i class="material-icons mdc-text-field__icon" tabindex="0" role="button">3d_rotation</i>
  <div class="mdc-notched-outline">
    <svg>
      <path class="mdc-notched-outline__path"/>
    </svg>
  </div>
  <div class="mdc-notched-outline__idle"></div>
</div>
```

https://codepen.io/williamernest/pen/vazLPB

acdvorak:
This was originally not in spec, but now it is (yay!). Might take a bit of refactoring.

PR #3451, opened 2018-08-28
feat(text-field): Add support for leading/trailing icons at the same time

williamernest:
fixes: #3302


GusRuss89:
Once this is merged will it also pave the way for leading icons on select fields? The spec says nothing about not supporting icons on select fields as far as I can see...

williamernest:
@gusruss89 Yes. The goal is to add support for leading icons and invalid styles to the current native select. We're also adding an enhanced select with the menu component that should also have feature parity with the text field.

porfirioribeiro:
@williamernest That would be great to have also!

packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -55,8 +55,37 @@
   }
 }
 
-@mixin mdc-text-field-icon-color_($color) {
+@mixin mdc-text-field-icon-horizontal-position-two-icons_($position-left, $padding-left, $position-right, $padding-right) {
   .mdc-text-field__icon {
-    @include mdc-theme-prop(color, $color);
+    &:first-of-type {
```
kfranqueiro: Just pointing out this will only work as long as it's safe to assume that the icons always use a distinct type (e.g. `i`, `svg`, etc., not `div`), since `first-of-type` and `last-of-type` apply against the element type, not the selector. I think this is a safe assumption...

packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -22,9 +22,9 @@
 
 // Public mixins
 
-@mixin mdc-text-field-icon-color($color) {
+@mixin mdc-text-field-icon-color($color, $onlyTrailing: false) {
```
kfranqueiro: Is there a reason why we're singling out trailing like this?

Would it make more sense to have there be 2 color parameters? i.e. `$color, $trailingColor: $color` where if you specify one it applies to both, but if you specify two they apply to leading/trailing?

(Could also do `$trailingColor: null` and still conditionalize the output like you're currently doing.)

packages/mdc-textfield/icon/_mixins.scss
porfirioribeiro: Maybe add a class like `mdc-text-field__icon--leading` and `mdc-text-field__icon--trailing` can prevent that?

kfranqueiro:
This looks OK from a styling perspective, but I think there are things on the JS side that this isn't taking into account...

The leading/trailing icon also has a controller/foundation, and text field is currently still assuming that there's only ever one of them to manage. This affects a few things:

* When the text field is set enabled/disabled, that is passed down to the icon foundation to remove its tabindex if it has one. (It properly remembers its previous tabindex and restores it when it is enabled again.)
* There are APIs on the top-level text-field component and foundation which forward to setting the icon's aria-label and content; these are sometimes used to swap between 2 icons.

packages/mdc-textfield/icon/_mixins.scss
williamernest: This was only used to style the trailing icon when in an invalid state, since that's the only time the 2 icons have different colors. I don't think it makes sense to support passing in 2 colors, since majority of the time the icons will always be the same color. There's only 1 time we need an exception (and to be honest, I'm not sure why leading icons don't turn red when the text field is invalid).

packages/mdc-textfield/icon/_mixins.scss
williamernest: I can work around the `last-of-type` issue using a sibling selector
```
.mdc-text-field__icon ~ .mdc-text-field__icon {
    color: red;
}
```

PR #3452, opened 2018-08-28
chore(infrastructure): More helpful error message in Closure test

acdvorak:
### What it does

* Adds a more accurate and helpful error message to `npm run test:closure` when it encounters a JS file that does not contain a multiline comment.
* Previously, it threw an error on a seemingly random line of code, with no indication of which JS file it was compiling or what the compiled output was.

### How to test

```bash
echo > packages/mdc-base/bork.js && npm run test:closure
```

#### Before:

image

#### After:

image

```
Error: Missing multiline comment terminator!

{
  "srcFile": "/Users/advorak/dev/mdc-web/.closure-tmp/packages/mdc-base/foo.js",
  "pathbasedPackageName": "mdc.base.foo",
  "packageStr": "mdc.base.foo",
  "outputCode": "exports = {};\n"
}
```


kfranqueiro:
I'm a little confused... is the missing multiline comment terminator at all correlated with the incorrect export format?

acdvorak:
Yes.

If you create a file that contains unsupported syntax `export function foo() {}`, the output generated by Closure Compiler is literally just `"{}\n"`.

So our script dies when it can't find a multiline comment terminator in `"{}\n"`.

I suppose we could check if the output is exactly equal to `"{}\n"` instead - WDYT?

acdvorak:
Need to investigate Closure Compiler supported syntax.

acdvorak:
I was wrong about the root cause: Closure Compiler _does_ support `export function foo() {}` syntax.

I removed any mention of unsupported syntax from the code comment and PR description.

PR #3453, opened 2018-08-28
feat(drawer): New sass mixin to set z-index

abhiomkar:


abhiomkar:
Right now we're not opinionated about clipped being variant because it requires applying additional CSS to external elements. Clipped variant also has open issue of extra space being left by auto dismissible top app bar.

We can definitely document in future. Created #3456

PR #3459, opened 2018-08-28
feat(drawer): Allow customizing drawer width

### LINKED ISSUE #2603: MDC Drawer width
Why is an MDC drawer a fixed 240px width regardless of device?

It seems crazy that it's so small on an XL desktop screen and it puts me off using the component in favour of my own push drawer.

If I override the fixed px width with a percentage in the SCSS does that work for all of the animation calculations?

abhiomkar:
Fixes #2603


packages/mdc-drawer/README.md
```diff
@@ -320,6 +320,7 @@ Mixin | Description
 `mdc-drawer-activated-overlay-color($color)` | Sets the overlay color of the activated drawer list item.
 `mdc-drawer-scrim-fill-color($color)` | Sets the fill color of `mdc-drawer-scrim`.
 `mdc-drawer-z-index($value)` | Sets the z index of drawer. Drawer stays on top of top app bar except for clipped variant of drawer.
+`mdc-drawer-width($width)` | Sets the width of drawer and margin space required for `mdc-drawer-app-content` in case of dismissible variant.
```
kfranqueiro: IMO this can be a little hard to read/parse (esp. the part that is dismissible-specific).

```
Sets the width of the drawer. In the case of the dismissible variant, also sets margin required for `mdc-drawer-app-content`.
```

packages/mdc-drawer/_mixins.scss
```diff
@@ -122,3 +122,12 @@
 @mixin mdc-drawer-z-index($value) {
   z-index: $value;
 }
+
+@mixin mdc-drawer-width($width) {
+  width: $width;
+  max-width: $width;
```
kfranqueiro: I know this was already here, but I forget, why'd we need both width and max-width?

abhiomkar: Good point. Removed.

kfranqueiro:
The latest screenshot test results seem to be showing a bunch of diffs I wouldn't expect. It looks like there's more content in the snapshot images (look at the scrollbars)...

packages/mdc-drawer/dismissible/mdc-drawer-dismissible.scss
```diff
@@ -42,11 +42,8 @@
 
   position: relative;
 
-  .mdc-drawer--open + & {
-    @include mdc-rtl-reflexive-box(margin, left, $mdc-drawer-width);
-  }
-
-  .mdc-drawer--closing + & {
+  // To level the specificity with app-content selector in `mdc-drawer-width` mixin.
+  .mdc-drawer.mdc-drawer--closing + & {
```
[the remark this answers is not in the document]
abhiomkar: Extra specificity is no longer needed. Removed `.mdc-drawer` from this selector.

PR #3464, opened 2018-08-29
chore(infrastructure): Add Google Analytics to all screenshot test pages

acdvorak:


kfranqueiro:
Why are we doing this?

PR #3470, opened 2018-08-29
fix(select): Only add line ripple listeners when line ripple is present

williamernest:
Modifies the line ripple logic in the `select` to only add the event listeners or updates if the line ripple element exists.


packages/mdc-select/index.js
```diff
@@ -287,7 +290,9 @@ class MDCSelect extends MDCComponent {
     const targetClientRect = evt.target.getBoundingClientRect();
     const xCoordinate = evt.clientX;
     const normalizedX = xCoordinate - targetClientRect.left;
-    this.lineRipple_.setRippleCenter(normalizedX);
+    if (this.lineRipple_) {
```
kfranqueiro: This would never not be true since we're not hooking up the listener to this method if line ripple doesn't exist, right?

Issue #2899, opened 2018-06-07
Deprecate and delete the non-box version of Select


kfranqueiro:
The Material guidelines only specify the filled and outlined variants of text field.
https://material.io/design/components/text-fields.html#spec

MDC Select was designed such that the select "box" is styled consistently to MDC Text Field. The current "non-box" version of select should thus be deprecated.

The Material research team showed that the filled and outlined variants are much more discoverable by users, which is why the Material guidelines stopped recommending the non-box version.

PR #3476, opened 2018-08-30
chore(select): Remove default variant

williamernest:
fixes: #2899


packages/mdc-select/_mixins.scss
```diff
@@ -176,3 +176,12 @@
     @include mdc-notched-outline-color($color);
   }
 }
+
+@mixin mdc-select-floating-label_ {
+  .mdc-floating-label {
+    @include mdc-rtl-reflexive-position(left, 16px);
+
+    bottom: 12px;
+    line-height: 1.75rem;
```
acdvorak: Is this related to any of our typography variables or mixins? Or just a component-specific value?

packages/mdc-select/index.js
```diff
@@ -145,7 +145,7 @@ class MDCSelect extends MDCComponent {
       this.outline_ = outlineFactory(outlineElement);
     }
 
-    if (this.root_.classList.contains(cssClasses.BOX)) {
+    if (!this.root_.classList.contains(cssClasses.OUTLINED)) {
```
acdvorak: This seems like a change of behavior. Did outlined previously not have ripples?

acdvorak: Never mind... I missed the `!`.

packages/mdc-select/_mixins.scss
williamernest: This is required to get the notched outline to float to the correct position. This will be refactored in a follow up PR so that the notched outline requires no changes in the default version of text field and the select.

packages/mdc-select/index.js
williamernest: The outlined variant does not have ripples. This code used to single out the box variant to add ripples, but since there is no longer a default variant this needs to exclude the outlined variant.

packages/mdc-select/README.md
```diff
@@ -83,21 +83,6 @@ See [Importing the JS component](../../docs/importing-js.md) for more informatio
 
 ## Variants
 
-### Select Box
```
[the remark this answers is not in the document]
williamernest: I think we should leave the deprecation notice in so people see it if they come looking for why their default variant changes.

PR #3479, opened 2018-08-30
fix(list): Set all other list item child elements pointer event to none

abhiomkar:


kfranqueiro:
Aren't there cases where these regions may have elements that can be interacted with, and wouldn't this prevent that? Why are we making this change?

abhiomkar:
That's correct. Closing this one.

Please see #3480 for explanation.

PR #3460, opened 2018-08-29
fix(list): Update single line list to ellipsis

williamernest:
fixes: #3201

PR #3480, opened 2018-08-30
fix(drawer): Remove list item children to be included in click target.

abhiomkar:
Since #3460 adds additional wrapper element to text label inside list item it might be included as click target. In this change, drawer removes the pointer events from list item children including `__text` & `__graphic` so click event always originates from `list-item` instead of its children.

The long-term solution would be to have a custom event for MDC List that drawer & menu can listen to.


packages/mdc-drawer/common.scss
```diff
@@ -89,6 +89,12 @@
   .mdc-list-divider {
     margin: 3px 0 4px 0;
   }
+
+  // Prevents list item children to be included in click target.
```
williamernest: Prevents list item children from being included in the click target.

PR #3493, opened 2018-09-01
docs(drawer): Fix typo in README.md

robbyrice:
The SASS import for mdc-list was pointing to the drawer folder. This changes the import to point to the list folder. It also removes an unnecessary closing div tag in one of the code examples.


kfranqueiro:
Thanks for spotting these!

PR #3513, opened 2018-09-05
fix(select): Disabled color and opacity

acdvorak:
### What it does

* Sets disabled `color` directly instead of giving the entire `mdc-select` element `opacity: .38` and relying on the default browser styles for text color.

#### Before:

image

#### After:

image

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -71,6 +71,11 @@
       text-indent: -2px;
     }
 
+    // stylelint-disable-next-line selector-max-type, plugin/selector-bem-pattern
+    > option {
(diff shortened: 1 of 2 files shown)


kfranqueiro:
Why are we changing this? The default guidance is 38% opacity which is what the code is currently doing.

acdvorak:
Look at the screenshots in the PR description. The existing version's text color is incorrect.

Browsers automatically style disabled `<select>` elements with `color: graytext` (WebKit). On top of that, we're applying 38% opacity. So the end result is opacity much less than 38%, which makes the text really hard to read.

In addition, the old version has transparent backgrounds, whereas the new one has opaque backgrounds.

PR #3514, opened 2018-09-05
chore(infrastructure): Run headless Chrome for local unit tests

acdvorak:


kfranqueiro:
I don't understand why we would want this. This makes `test:watch` pretty much useless.

PR #3516, opened 2018-09-05
chore(infrastructure): Better error messages in verifyDefaultAdapter

acdvorak:
This PR adds more helpful error messages when adapters don't match the expected method names.

#### Before:

```
expected [ Array(21) ] to deeply equal [ Array(21) ]
```

#### After:

```
Found 1 unexpected method: addClass; 1 missing method: addClassX
```

image


test/unit/helpers/foundation.js
```diff
@@ -24,44 +24,132 @@
...
+    const set = new Set();
+    array.forEach((value) => set.add(value));
+    return set;
+  };
+
+  const actualSet = toSet(actualArray);
+  const expectedSet = toSet(expectedArray);
+  const addedStr = getAddedStr(actualSet, expectedSet);
+  const removedStr = getRemovedStr(actualSet, expectedSet);
+  const messages = [addedStr, removedStr].filter((val) => Boolean(val));
```
kfranqueiro: `(val) => val.length > 0` ? Makes it a little more clear this is just filtering out empty strings...

test/unit/helpers/setup.js
```diff
@@ -23,9 +23,23 @@
...
+ * Returns a foundation configured to use a mock object with the same API as a default adapter,
+ * as well as that adapter itself.
+ * The trailing `.` in the `@param` type below is intentional: It indicates a reference to the class itself instead of
+ * an instance of the class.
+ * See https://youtrack.jetbrains.com/issue/WEB-10214#focus=streamItem-27-1305930-0-0
+ * @param {!MDCFoundation.} FoundationClass
+ * @return {{mockAdapter: !Object, foundation: !MDCFoundation}}
+ */
 export function setupFoundationTest(FoundationClass) {
+  // Make code coverage happy by instantiating the class with no arguments.
```
kfranqueiro: I'm confused... where were we missing coverage due to this? Judging by this comment it should currently be impossible for us to have 100% branch coverage on foundations, but that's clearly not the case...

acdvorak: All I know is that Dialog code coverage was stuck at 19/20 until I added this line. 🤷‍♂️

acdvorak: ### Without this line:

image

File |   | Statements |   | Branches |   | Functions |   | Lines |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
constants.js |   | 100% | 3/3 | 100% | 0/0 | 100% | 0/0 | 100% | 3/3
foundation.js |   | 100% | 64/64 | 92.86% | 13/14 | 100% | 22/22 | 100% | 61/61
index.js |   | 100% | 46/46 | 100% | 6/6 | 100% | 31/31 | 100% | 45/45
util.js |   | 100% | 6/6 | 100% | 2/2 | 100% | 4/4 | 100% | 5/5

---

### With this line:

image

File |   | Statements |   | Branches |   | Functions |   | Lines |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
constants.js |   | 100% | 3/3 | 100% | 0/0 | 100% | 0/0 | 100% | 3/3
foundation.js |   | 100% | 64/64 | 100% | 14/14 | 100% | 22/22 | 100% | 61/61
index.js |   | 100% | 46/46 | 100% | 6/6 | 100% | 31/31 | 100% | 45/45
util.js |   | 100% | 6/6 | 100% | 2/2 | 100% | 4/4 | 100% | 5/5

test/unit/helpers/foundation.js
acdvorak: Done

test/unit/helpers/setup.js
kfranqueiro: I just ran coverage with/without this line and it only affects branch coverage in dialog, line-ripple, and list. I'll see if I can figure out why, but I am inclined to leave it out of this PR.

kfranqueiro: I figured out why 3 specific components have diffs with this line. It's because for some reason those components are using a default parameter in their foundation constructors, whereas all other foundations use an `Object.assign` within the constructor instead. I'm not sure what prompted this deviation, as the `Object.assign` pattern existed much earlier.

acdvorak: #3522 removes the default parameters from `line-ripple` and `list`.

PR #3520, opened 2018-09-05
chore(infrastructure): Fix flaky Edge font rendering in Drawer pages

acdvorak:
For some reason, the last letter of every Drawer list item - and the last letter of the Top App Bar's title - are super flaky in Edge. They randomly shift left/right by about 1px ([example](https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/05/16_31_23_721/report/report.html?utm_source=github&utm_content=pr_status)).

The flakiness is consistent enough that redacting specific characters seems to be effective.


test/screenshot/spec/fixture.scss
```diff
@@ -67,10 +67,19 @@ $test-layout-cell-grid-color: #ddd;
     // stylelint-enable selector-class-pattern
   }
 
-  .test-container--edge-fonts .test-font--flaky {
+  .test-container--edge-fonts .test-font--redact-all {
```
kfranqueiro: I would assume there are places that were referencing this by its old name?

acdvorak: Only #3413 references it for now

PR #3522, opened 2018-09-05
chore: Remove unnecessary default arguments that affect coverage

kfranqueiro:
This solves the mystery in https://github.com/material-components/material-components-web/pull/3516#discussion_r215364312.

All of our foundation constructors follow the pattern of calling `super(Object.assign(MDCFooFoundation.defaultAdapter, adapter))`, but two components in particular (three if you include the Dialog prototype) were also assigning an empty object as the default value, which caused coverage to decrease due to the default "branch" never being taken. The default assignment is not necessary, since `Object.assign` effectively no-ops on `undefined` sources anyway.


kfranqueiro:
I saw the %s go up on my end, I'll look at the absolute numbers...

PR #3530, opened 2018-09-06
feat(chips): Add mixins to customize horizontal padding and icon margins

rlfriedman:


kfranqueiro:
I would've asked for a screenshot test if it weren't for the fact that we don't have any screenshot tests for chips yet... Do we want to work on that in a separate PR?

rlfriedman:
I'll be adding screenshot tests in a separate PR first, and then will update this to include them for these changes.

rlfriedman:
Updated with screenshot tests for each mixin: https://storage.googleapis.com/mdc-web-screenshot-tests/rlfriedman/2018/09/13/22_36_05_454/report/report.html?utm_source=cli&utm_content=test_results

packages/mdc-chips/_mixins.scss
```diff
@@ -159,3 +163,23 @@
   height: $size;
   font-size: $size;
 }
+
+@mixin mdc-chip-leading-icon-margin(
+  $top: $mdc-chip-leading-icon-margin-top,
+  $right: $mdc-chip-leading-icon-margin-right,
+  $bottom: $mdc-chip-leading-icon-margin-bottom,
+  $left: $mdc-chip-leading-icon-margin-left) {
+  .mdc-chip__icon.mdc-chip__icon--leading {
```
lynnjepsen: the `.mdc-chip__icon` is overly specific in this instance. `mdc-chip__icon--leading` can only apply to `mdc-chip__icon`, so you don't need to specify it again.

We did some research offline and realize that sometimes we need to be overly specific about our CSS selectors for icon elements, because we want to override styles applied by `material-icons-extended`

However,`margin` is applied by `material-icons-extended`, so it does not apply in this case.

So remove `mdc-chip__icon`, pretty please, and re-run the screenshot tests.

packages/mdc-chips/_mixins.scss
```diff
@@ -159,3 +163,23 @@
...
+    margin: $top $right $bottom $left;
+  }
+}
+
+@mixin mdc-chip-trailing-icon-margin(
+  $top: $mdc-chip-trailing-icon-margin-top,
+  $right: $mdc-chip-trailing-icon-margin-right,
+  $bottom: $mdc-chip-trailing-icon-margin-bottom,
+  $left: $mdc-chip-trailing-icon-margin-left) {
+  .mdc-chip__icon.mdc-chip__icon--trailing {
```
lynnjepsen: also here.

packages/mdc-chips/chip/mdc-chip.scss
```diff
@@ -77,13 +77,13 @@
 }
 
 .mdc-chip__icon--trailing {
-  margin: 0 -4px 0 4px;
+  margin: $mdc-chip-trailing-icon-margin-top $mdc-chip-trailing-icon-margin-right $mdc-chip-trailing-icon-margin-bottom $mdc-chip-trailing-icon-margin-left;
```
lynnjepsen: this is basically duplicating the logic in `mdc-chip-trailing-icon-margin`...so instead call `mdc-chip-trailing-icon-margin` from `@at-root`

and then re-run the screenshots :)

PR #3531, opened 2018-09-07
fix(theme): Declare error variables as !default

mathiasschopmans:
To allow customization SASS variables should be declared as `!default`


mathiasschopmans:
I signed it!

Issue #3510, opened 2018-09-05
MDCList: Move toggle of checkbox from Menu to List. 


williamernest:
Currently, the MDC Menu handles toggling a checkbox if present in a list item. This functionality can be expanded to include radio buttons and can be moved to the list since the Dialog can also use this behavior.

PR #3546, opened 2018-09-11
feat(list): Toggle radio checkbox

williamernest:
fixes: #3510 

BREAKING CHANGE: Moves the toggling of radio buttons and checkboxes to the list from the menu. Adds toggleCheckbox List adapter API.


packages/mdc-list/index.js
```diff
@@ -234,6 +229,26 @@ class MDCList extends MDCComponent {
...
       },
+      toggleCheckbox: (index) => {
+        const listItem = this.listElements[index];
+        const elementsToToggle =
+          [].slice.call(listItem.querySelectorAll('input[type="radio"], input[type="checkbox"]'));
+        elementsToToggle.forEach((element) => {
+          const event = document.createEvent('Event');
+          event.initEvent('change', false, true);
+
+          if (!element.checked) {
```
kfranqueiro: I think this can be simplified to `if (!element.checked || element.type !== 'radio')` and avoid needing the else and repeating the 2 lines inside? (Move the comment up)

packages/mdc-list/index.js
```diff
@@ -234,6 +229,26 @@ class MDCList extends MDCComponent {
           listItem.click();
         }
       },
+      toggleCheckbox: (index) => {
+        const listItem = this.listElements[index];
+        const elementsToToggle =
+          [].slice.call(listItem.querySelectorAll('input[type="radio"], input[type="checkbox"]'));
+        elementsToToggle.forEach((element) => {
```
kfranqueiro: If we `let` a boolean outside of this loop and set it to true when this function is run, we can return it as an indicator of whether an input was found to toggle, and `preventDefault` on the event within the foundation if this returns true.

packages/mdc-list/index.js
```diff
@@ -234,6 +229,23 @@ class MDCList extends MDCComponent {
...
         }
       },
+      toggleCheckbox: (index) => {
+        let checkboxOrRadioExists = false;
+        const listItem = this.listElements[index];
+        const elementsToToggle =
+          [].slice.call(listItem.querySelectorAll('input[type="radio"], input[type="checkbox"]'));
+        elementsToToggle.forEach((element) => {
+          const event = document.createEvent('Event');
+          event.initEvent('change', false, true);
```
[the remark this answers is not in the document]
kfranqueiro: Also, `initEvent` is apparently deprecated in favor of directly using Event constructors; see https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -273,22 +273,6 @@ test('wrapFocus calls setWrapFocus on foundation', () => {
...
 
-
-test('singleSelection true sets the click handler from the root element', () => {
-  const {root, component, mockFoundation} = setupTest();
-  component.singleSelection = true;
-  domEvents.emit(root, 'click');
-  td.verify(mockFoundation.handleClick(td.matchers.anything()), {times: 1});
-});
-
-test('singleSelection false removes the click handler from the root element', () => {
```
kfranqueiro: FYI there is no equivalent test to this in the new code (i.e. no test that tests MDCListFoundation#handleClick when `isSingleSelectionList_` is false)

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -374,3 +358,23 @@ test('keydown handler is removed from the root element on destroy', () => {
   listElementItem.dispatchEvent(event);
   td.verify(mockFoundation.handleKeydown(event, true, 0), {times: 0});
 });
+
+test('adapter#toggleCheckbox toggles a checkbox', () => {
```
kfranqueiro: Add a third test with no checkbox/radio to verify it returns false

packages/mdc-list/index.js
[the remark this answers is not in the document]
williamernest: Event constructors are not supported in IE11. 
https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

test/unit/mdc-list/foundation.test.js
```diff
@@ -483,75 +483,64 @@ test('#handleKeydown space key is triggered and focused is moved to a different
...
+  foundation.setSingleSelection(true);
+  foundation.handleClick(0, false);
+  foundation.handleClick(0, false);
 
-    td.verify(mockAdapter.setAttributeForElementIndex(0, 'tabindex', 0), {times: 0});
-  });
+  td.verify(mockAdapter.setAttributeForElementIndex(0, 'tabindex', 0), {times: 0});
+});
+
+test('#handleClick proxies to the adapter#toggleCheckbox is toggleCheckbox is true', () => {
```
kfranqueiro: is toggleCheckbox -> if toggleCheckbox

test/unit/mdc-list/foundation.test.js
```diff
@@ -481,62 +481,66 @@ test('#handleKeydown space key is triggered and focused is moved to a different
   td.verify(mockAdapter.setAttributeForElementIndex(2, 'tabindex', 0), {times: 1});
 });
 
+test('#handleClick when singleSelection=true on a list item should not cause the list item to be selected', () => {
```
kfranqueiro: true -> false

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -226,11 +225,11 @@ test('adapter#setTabIndexForListItemChildren sets the child button/a/radio/check
   document.body.appendChild(root);
   const listItems = root.querySelectorAll('.mdc-list-item');
 
-  for (let index = 0; index < listItems.length; index++) {
-    assert.equal(0, listItems[index].querySelectorAll('[tabindex="0"]').length);
-    component.getDefaultFoundation().adapter_.setTabIndexForListItemChildren(index, 0);
-    assert.equal(1, listItems[index].querySelectorAll('[tabindex="0"]').length);
-  }
+  component.getDefaultFoundation().adapter_.setTabIndexForListItemChildren(0, 0);
```
kfranqueiro: The descriptions of this test and in the README need updating for the new behavior.

kfranqueiro:
FYI there's currently no test exercising the preventDefault branch in handleKeydown.

packages/mdc-list/README.md
```diff
@@ -268,12 +268,13 @@ The MDCList JavaScript component implements the WAI-ARIA best practices for
 [Listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox). This includes overriding the default tab behavior
 within the list component. You should not add `tabindex` to any of the `li` elements in a list.
 
-As the user navigates through the list, any `button`, `a`, `input[type="radio"]` and `input[type="checkbox"]` elements 
-within the list will receive `tabindex="-1"` when the list item is not focused. When the list item receives focus, the 
-aforementioned elements will receive `tabIndex="0"`. This allows for the user to tab through list item elements
-and then tab to the first element after the list. The `Arrow`, `Home`, and `End` keys should be used for navigating
-internal list elements. If `singleSelection=true`, the list will allow the user to use the `Space` or `Enter` keys to
-select or deselect a list item. The MDCList will perform the following actions for each key press
+As the user navigates through the list, any `button`, and `a` elements within the list will receive `tabindex="-1"` when
```
kfranqueiro: Nit: remove comma between "`button` and `a`"

packages/mdc-list/foundation.js
```diff
@@ -203,17 +204,31 @@ class MDCListFoundation extends MDCFoundation {
         this.adapter_.followHref(currentIndex);
       }
     }
+
+    if (isEnter || isSpace) {
```
kfranqueiro: Would it be more straightforward to put this within the `else if` above, make that `else if` check only for `isEnter || isSpace`, then nest `if (this.isSingleSelectionList_)` within that? IMO it's a little hard to follow right now since there's two separate blocks with logic for enter/space, and potentially both of them can be executed.

Tangentially, I also can't remember why the preventDefault for single-selection is outside the isRootListItem check?

packages/mdc-menu/adapter.js
```diff
@@ -110,19 +110,6 @@ class MDCMenuAdapter {
  *   }} evtData
    */
   notifySelected(evtData) {}
-
```
kfranqueiro: These also need to be removed from the menu README

packages/mdc-list/foundation.js
```diff
@@ -193,27 +194,41 @@ class MDCListFoundation extends MDCFoundation {
     } else if (isEnd) {
       this.preventDefaultEvent_(evt);
       this.focusLastElement();
-    } else if (this.isSingleSelectionList_ && (isEnter || isSpace)) {
-      this.preventDefaultEvent_(evt);
-      // Check if the space key was pressed on the list item or a child element.
-      if (isRootListItem) {
-        this.setSelectedIndex(currentIndex);
+    } else if (isEnter || isSpace) {
+      if (this.isSingleSelectionList_) {
```
kfranqueiro: I think we can collapse these two nested if's into one if && now?

packages/mdc-list/foundation.js
```diff
@@ -193,27 +194,41 @@ class MDCListFoundation extends MDCFoundation {
...
+        // Check if the space key was pressed on the list item or a child element.
+        if (isRootListItem) {
```
kfranqueiro: We don't seem to have a test for single-selection + non-root-list-item

Issue #3176, opened 2018-07-23
Integrate shape system API with components


abhiomkar:
Integrate shape system APIs with below components first:

- [x] Buttons (#3547)
- [x] Cards (#3547)
- [ ] Extended FAB (#3552)
- [x] Chips (#3547)

We should ultimately aim to replace any corner-radius mixins with shape mixins, which additionally includes the following components;

- [x] Dialog (#3552)
- [x] Drawer (item) (#3547)
- [x] Image List (#3547)
- [x] List (item) (#3547)
- [x] Menu Surface (#3552)
- [x] Notched Outline (#3552)
- [x] Select (#3552)
- [x] Text Field (#3552)
- [x] Top App Bar (Short) (#3552)

kfranqueiro:
Resolved via #3547 and #3552.

PR #3547, opened 2018-09-12
feat(shape): Integrate Shape API with components

abhiomkar:
This branch has integration to only following components:

- button
- card
- chips
- drawer
- image-list
- list

Rest of the components in following PR.

Refs #3176

BREAKING CHANGE: Replace `*-corner-radius` component mixins with `*-shape-radius` mixins to integrate with Shape system.


packages/mdc-drawer/README.md
```diff
@@ -331,7 +331,8 @@ Mixin | Description
 `mdc-drawer-item-text-ink-color($color)` | Sets drawer list item text ink color.
 `mdc-drawer-item-activated-icon-ink-color($color)` | Sets activated drawer list item icon ink color.
 `mdc-drawer-item-activated-text-ink-color($color)` | Sets activated drawer list item text ink color.
-`mdc-drawer-item-corner-radius($radius)` | Sets the corner border radius of the drawer list item.
+`mdc-drawer-shape-radius($radius, $rtl-reflexive)` | Sets the rounded shape to drawer with given radius size. `$radius` can be single radius or list of 2 radius values for trailing-top and trailing-bottom. Set `$rtl-reflexive` to true to flip radius values in RTL context, defaults to true.
```
kfranqueiro: Remove `$rtl-reflexive` here since IIUC we're intentionally not exposing it for drawer (maybe explain instead that only the side farthest from the edge of the screen has radius applied)

packages/mdc-shape/README.md
```diff
@@ -65,6 +65,42 @@ Styles for applying shapes for specific corners such as drawer looks like this:
 
 Where, only top-right & bottom-right corners are customizable and it automatically flips radius values based on RTL context when `$rtl-reflexive` is set to true.
 
+### Global theming
```
kfranqueiro: I'm very wary of enumerating components and overrides here because they will _undoubtedly_ end up out of sync the moment design makes us change something in the respective package and we forget to change it here. This seems like strictly design documentation, so I would push on designers that this should be something that gets documented on material.io so we can cross-link to it and they can update it directly.

demos/shape.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
```
kfranqueiro: Rather than re-add this, can we add a pill screenshot test for button and then remove this (and remove mention of it from demos/index.html)?

packages/mdc-image-list/_mixins.scss
```diff
@@ -26,13 +27,9 @@
...
 
-@mixin mdc-image-list-corner-radius($radius) {
+@mixin mdc-image-list-shape-radius($radius, $rtl-reflexive: false) {
   .mdc-image-list__image {
-    border-radius: $radius;
-  }
-
-  .mdc-image-list__supporting {
-    border-radius: 0 0 $radius $radius;
+    @include mdc-shape-radius($radius, $rtl-reflexive);
```
kfranqueiro: This lost the details of how `.mdc-image-list__supporting` was styled, which constrains to only the bottom corners (the top corners are always flush with the image; only the bottom is considered to be along the boundary of the item).

test/screenshot/spec/mdc-button/fixture.scss
```diff
@@ -54,7 +54,7 @@ $custom-button-custom-horizontal-padding: 24px;
 }
 
 .custom-button--corner-radius {
-  @include mdc-button-corner-radius($custom-button-custom-corner-radius);
+  @include mdc-button-shape-radius($custom-button-custom-corner-radius);
```
kfranqueiro: Should we also update the class name and the name of the button and card mixin test pages since we're updating the name of the mixin?

This also needs to be done for chips.

packages/mdc-image-list/_mixins.scss
abhiomkar: With the previous implementation __supporting border-radius style was doing nothing since it doesn't have any background. Also, the corner-radius mixin applies border radius to 4 corners of image with and without existence of __supporting element.

kfranqueiro: It does have a background when the image list uses overlaid labels with text protection. See the first example under Demos at https://material-components.github.io/material-components-web-catalog/#/component/image-list

packages/mdc-button/_mixins.scss
```diff
@@ -68,8 +70,12 @@
   }
 }
 
-@mixin mdc-button-corner-radius($corner-radius) {
-  border-radius: $corner-radius;
+@mixin mdc-button-shape-radius($radius, $rtl-reflexive: false) {
+  @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-button-height, $radius), $rtl-reflexive);
+
+  &.mdc-button--dense {
```
kfranqueiro: This makes it impossible to customize the corner radius of dense and non-dense buttons independently, which was previously possible with the old mixin (you'd just apply it to whichever one you wanted). I could see where people might want to do that if they're using an absolute value... though I wonder if it's ever expected to mix and match dense and non-dense variants?

The same is true of e.g. FAB in #3552... and I'd expect it to be true of text field as well, though I don't see it there, which makes me suspect we're not treating this consistently.

I think we need to think through how we want to support dense vs. non-dense variants WRT shape. I realize that for components with fixed heights it needs to be treated separately for purposes of supporting percentage, but I'm not sure we should be forcing different-sized variants to all use the same radius values.

abhiomkar: As discussed, we apply same shapes for all component variants. User can override a specific variant as a one-off.

packages/mdc-image-list/_mixins.scss
abhiomkar: Good catch. This wasn't applying with previous implementation - corrected that.

packages/mdc-shape/_functions.scss
```diff
@@ -143,6 +124,61 @@
...
+// 3. mdc-shape-mask-top-radius(4px 4px 4px 4px, 0 1 1 0) => 0 4px 4px 0
+//
+@function mdc-shape-mask-radius($radius, $masked-corners) {
+  @if type-of($radius) == "list" {
+    @if length($radius) > 4 {
+      @error "Invalid radius: '#{$radius}' is more than 4 values";
+    }
+  }
+
+  @if type-of($masked-corners) != "list" and length($masked-corners) != "length" {
```
kfranqueiro: I don't think this condition is quite right... `length($masked-corners)` will never equal `"length"` so this always evaluates to false.

I think this can be simplified to `@if length($masked-corners) != 4` since `length` of something that isn't a list is always `1`.

packages/mdc-shape/_functions.scss
```diff
@@ -143,6 +124,61 @@
...
+// Accepts radius number or list of 2 radius values and returns 4 value list with
+// top-left value, top-right value, bottom-right 0, bottom-left 0.
+// Throws error when length of radius is more than 2.
+//
+// Example:
+//
+// 1. mdc-shape-mask-top-radius((2px, 3px)) => 2px 3px 0 0
+// 2. mdc-shape-mask-top-radius(8px) => 8px 8px 0 0
+//
+@function mdc-shape-mask-top-radius($radius) {
```
kfranqueiro: Can this be simplified to just reuse `mdc-shape-mask-radius($radius, 1 1 0 0)` now?

Should we even just remove this and always use mask-radius?

packages/mdc-shape/_functions.scss
```diff
@@ -143,6 +124,61 @@
...
+// Accepts radius number or list of 2-4 radius values and returns 4 value list with
+// masked corners as mentioned in `$masked-corners`
+//
+// Example:
+//
+// 1. mdc-shape-mask-radius(2px 3px, 1 1 0 0) => 2px 3px 0 0
+// 2. mdc-shape-mask-top-radius(8px, 0 0 1 1) => 0 0 8px 8px
+// 3. mdc-shape-mask-top-radius(4px 4px 4px 4px, 0 1 1 0) => 0 4px 4px 0
+//
+@function mdc-shape-mask-radius($radius, $masked-corners) {
```
kfranqueiro: Huge +1 to this function. I was debating asking for something like this in lieu of mask-top-radius earlier but didn't want to make more work if it wasn't warranted.

packages/mdc-shape/_functions.scss
abhiomkar: Removed `mask-top-radius` mixin in next PR.

Issue #3550, opened 2018-09-12
MDC Select - Options cannot be seen and down arrow color stays black on dark theme


JPlow85:
### What MDC Web Version are you using?

v0.39.0-0

### What browser(s) is this bug affecting?

Vivaldi, Chrome, Edge

### What OS are you using?

Windows 10

### What are the steps to reproduce the bug?

1. Run the demo server
2. Go to http://localhost:8081/theme/index.html?theme=dark (note 8080 on most machines)
3. Observe the mdc-select component's behavior when opened

### What is the expected behavior?

1. The down arrow should be white (with opacity applied) due to dark background
2. The user should be able to read all options associated with select. The text color needs to be set back to black or the select background needs to be set to theme color.

### What is the actual behavior?

1. Down arrow continues to use black.
2. User cannot see the select options as they are the same color as the select background.

### Images

![image](https://user-images.githubusercontent.com/18597960/45451207-3d4f8400-b68f-11e8-9c88-c943ad4f5070.png)

williamernest:
@JPlow85 Thanks for filing this issue. We have a PR to fix this behavior ready so it should be fixed by the next release.

PR #3553, opened 2018-09-13
fix(select): Fix dropdown color/opacity and options background

williamernest:
fixes: #3550


packages/mdc-select/_mixins.scss
```diff
@@ -147,7 +147,10 @@
   }
 }
 
-@mixin mdc-select-dd-arrow-svg-bg_($fill-hex-number: 000000, $opacity: .54) {
+@mixin mdc-select-dd-arrow-svg-bg_($fill-hex-number: #000000, $opacity: .54) {
+  // Remove leading # from
```
kfranqueiro: Is this comment missing some words? Or drop "from"?

packages/mdc-select/_mixins.scss
```diff
@@ -147,7 +147,10 @@
   }
 }
 
-@mixin mdc-select-dd-arrow-svg-bg_($fill-hex-number: 000000, $opacity: .54) {
+@mixin mdc-select-dd-arrow-svg-bg_($fill-hex-number: #000000, $opacity: .54) {
```
kfranqueiro: * Remove the default value for the first parameter since we always specify it
* Should the default opacity be represented in a variable? Should we also always specify that in the mixin invocation rather than providing a default? (Currently we pass 1 parameter in one invocation, and 2 parameters in the other)

packages/mdc-select/mdc-select.scss
```diff
@@ -89,6 +89,8 @@
 
     // stylelint-disable-next-line selector-max-type, plugin/selector-bem-pattern
     > option {
+      @include mdc-theme-prop(background-color, surface);
```
kfranqueiro: What effect does this have on native selects?

williamernest: This causes the `background-color` to be the `surface` color on browsers that allow it. Since we're changing the color on Line 94, we also need to adjust the background-color. On browsers that don't allow modifications to the dropdown options, neither of these apply and the dropdown is rendered with a white background and black text.

Issue #3551, opened 2018-09-12
MDC Ripple - Dark theme does not change ripple to use contrasting color


JPlow85:
### What MDC Web Version are you using?

v0.39.0-0

### What browser(s) is this bug affecting?

Vivaldi, Chrome, Edge

### What OS are you using?

Windows 10

### What are the steps to reproduce the bug?

1. Run the demo server
2. Go to http://localhost:8081/theme/index.html?theme=dark (note 8080 on most machines)
3. Observe the component's behavior on hover or select

### What is the expected behavior?

1. The hover and select color should be contrasting, in this case white, rather than black.

### What is the actual behavior?

1. The hover color remains black regardless of theme colors

### Images

Outline of icon is visible due to black contrasting color (note: cursor hovering in screenshot)
![image](https://user-images.githubusercontent.com/18597960/45451424-e8603d80-b68f-11e8-9684-68cd88944247.png)

Outline of icon is not visible due to black non-contrasting color (note: cursor hovering in screenshot)
![image](https://user-images.githubusercontent.com/18597960/45451413-dd0d1200-b68f-11e8-8ce7-f83e158ee04c.png)

williamernest:
@JPlow85 Thanks for filing this issue. We have a PR to fix this behavior ready so it should be fixed by the next release.

PR #3554, opened 2018-09-13
fix(ripple): Change default color from black to on-surface

williamernest:
fixes: #3551 

Per our discussion, we may not accept this change but I wanted to see the diffs and have a solution in place in case we did.


williamernest:
From the states article "The overlay color matches the color of the text or iconography on the element it’s applied to." It seems like this should match the `on-surface` color.  

https://material.io/design/interaction/states.html#anatomy

Issue #3511, opened 2018-09-05
MDCList: Make toggling tabindex of interactive elements optional


williamernest:
MDC List currently toggles the tabindex of `button` and `a` elements contained in a list item when focus/blue occur. Developers should be able to optionally disable this behavior.

PR #3564, opened 2018-09-14
feat(list): Added setFocusInListItemChildren new API to list foundation.

abhiomkar:
This change sets the list to enable or disable focusing in the list item children.

Fixes Issue #3511

### DIFF
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -268,8 +268,8 @@ The MDCList JavaScript component implements the WAI-ARIA best practices for
 [Listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox). This includes overriding the default tab behavior
 within the list component. You should not add `tabindex` to any of the `li` elements in a list.
 
-As the user navigates through the list, any `button`, `a`, `input[type="radio"]` and `input[type="checkbox"]` elements 
-within the list will receive `tabindex="-1"` when the list item is not focused. When the list item receives focus, the 
+As the user navigates through the list, any `button`, `a`, `input[type="radio"]` and `input[type="checkbox"]` elements
+within the list will receive `tabindex="-1"` when the list item is not focused. When the list item receives focus, the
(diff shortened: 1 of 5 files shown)


packages/mdc-list/README.md
```diff
@@ -364,6 +364,7 @@ Method Signature | Description
 `setSingleSelection(value: Boolean) => void` | Sets the list to be a selection list. Enables the `enter` and `space` keys for selecting/deselecting a list item.
 `setSelectedIndex(index: Number) => void` | Toggles the `selected` state of the list item at index `index`.
 `setUseActivated(useActivated: boolean) => void` | Sets the selection logic to apply/remove the `mdc-list-item--activated` class.
+`setFocusInListItemChildren(enable: Boolean) => void` | Sets the list to enable or disable focusing in list item children.
```
williamernest: I think this name doesn't indicate what it's enabling or disabling. Maybe `enableToggleTabindexForChildren` ? That's probably worse. We should at least add more to the description to indicate what it's doing.  Maybe try to link this description back to the description on 271-273.

abhiomkar: Done.

kfranqueiro: How about...

```
`setTabbableChildren(tabbable: boolean) => void` | Sets whether focusable children of the currently-focused list item can also be individually focused via the Tab key.
```

williamernest:
The List sets every checkbox/radio button to tabindex=-1 when it's instantiated. Once this behavior is disabled, it will prevent the List from making the checkbox/radio's tabindex=0 during `focusIn` event so that when you hit the `tab` key, you skip over the rest of the list.

abhiomkar:
Discussed this with Will & we decided that PR #3546 would detect for existence of checkbox & radio and also would keep the tabindex to -1 so those wouldn't be tabbable when list item receives focus. This would be the default MDC List behaviour, then we don't even need this change.

PR #3566, opened 2018-09-14
feat(select): Rotate dropdown arrow

williamernest:
Separates the dropdown arrow into it's own element and adds the ability for the dropdown arrow to rotate when the menu is opened.

### DIFF
--- a/packages/mdc-select/mdc-select.scss
+++ b/packages/mdc-select/mdc-select.scss
@@ -69,6 +69,23 @@
     background-position: left 10px center;
   }
 
+  &__dropdown-icon {
+    @include mdc-select-dd-arrow-svg-bg_;
+    @include mdc-rtl-reflexive(left, auto, right, 8px);
+
+    position: absolute;
+    bottom: 16px;
+    width: 24px;
+    height: 24px;
+    transition: transform 250ms;
+    pointer-events: none;
+
+    .mdc-select--focused & {
+      transform: rotate(180deg);
+      transition: transform 250ms;
+    }
+  }
+
   &__native-control {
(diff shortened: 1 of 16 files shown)


packages/mdc-select/mdc-select.scss
```diff
@@ -151,13 +167,16 @@
...
 .mdc-select--disabled {
-  @include mdc-select-dd-arrow-svg-bg_($mdc-select-dropdown-color, $mdc-select-disabled-dropdown-opacity);
   @include mdc-select-container-fill-color_($mdc-select-box-disabled-fill-color);
 
   .mdc-floating-label {
     @include mdc-floating-label-ink-color($mdc-select-disabled-label-color);
   }
 
+  .mdc-select__dropdown-icon {
+    @include mdc-select-dd-arrow-svg-bg_($mdc-select-dropdown-color, .37);
```
kfranqueiro: `.37` -> `$mdc-select-disabled-dropdown-opacity`

packages/mdc-select/mdc-select.scss
```diff
@@ -69,6 +68,23 @@
     background-position: left 10px center;
   }
 
+  &__dropdown-icon {
+    @include mdc-select-dd-arrow-svg-bg_($mdc-select-dropdown-color, $mdc-select-dropdown-opacity);
+    @include mdc-rtl-reflexive(left, auto, right, 8px);
+
+    position: absolute;
+    bottom: 16px;
```
kfranqueiro: This seems like it sits higher vs. text in the default variant than in the outlined variant. Is there supposed to be a different `bottom` style for outlined?

williamernest: No, this is intended based on the design.

PR #3571, opened 2018-09-17
fix(dialog): Update aria-hidden attribute on open/close

### LINKED ISSUE #3535: Dialog never updates aria-hidden attribute

Any

All

Go to any dialog demo page e.g. https://material-components.github.io/material-components-web-catalog/#/component/dialog

The dialog root element should have `aria-hidden="true"` when closed and remove the attribute when opened

The JS component doesn't update/sync `aria-hidden` at all

I noticed this while working on the dialog prototype branch for #3413, but this is already an existing issue. I'm aiming to fix it for new dialog.

kfranqueiro:
Fixes #3535.

BREAKING CHANGE: Adds setAttribute and removeAttribute adapter APIs.

### DIFF
--- a/packages/mdc-dialog/foundation.js
+++ b/packages/mdc-dialog/foundation.js
@@ -44,6 +44,8 @@ class MDCDialogFoundation extends MDCFoundation {
       removeClass: (/* className: string */) => {},
       addBodyClass: (/* className: string */) => {},
       removeBodyClass: (/* className: string */) => {},
+      setAttribute: (/* name: string, value: string */) => {},
+      removeAttribute: (/* name: string */) => {},
       eventTargetHasClass: (/* target: !EventTarget, className: string */) => {},
       computeBoundingRect: () => {},
       trapFocus: () => {},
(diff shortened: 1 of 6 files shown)


packages/mdc-dialog/foundation.js
```diff
@@ -127,6 +131,8 @@ class MDCDialogFoundation extends MDCFoundation {
     this.adapter_.removeClass(cssClasses.OPEN);
     this.adapter_.removeBodyClass(cssClasses.SCROLL_LOCK);
 
+    this.adapter_.setAttribute('aria-hidden', 'true');
```
abhiomkar: Screen readers usually don't pronounce elements that are set to "display: none". Do we need this?

kfranqueiro: OK, so now I'm confused...

I've seen the following quote several places:

> Authors MUST set aria-hidden="true" on content that is not displayed, regardless of the mechanism used to hide it.

However, the [current WAI-ARIA spec](https://www.w3.org/TR/using-aria/#fourth) apparently contradicts this:

> If an interactive element is hidden using display:none or visibility:hidden, it will also be removed from the accessibility tree, which makes the addition of aria-hidden="true" unnecessary.

So yeah... in that case, maybe any other components we have that are currently doing this don't need to do it? I'll have to double-check tomorrow.

kfranqueiro:
Based on further discussion with the team, it does seem like dialogs just shouldn't be using aria-hidden at all now. The previous implementation of dialog used visibility: hidden rather than display: none, and included aria-hidden in its documentation examples, but then never removed the attribute.

I'm going to update this PR to just remove mention of aria-hidden in our examples instead.

Issue #3481, opened 2018-08-30
MDC List: Add a custom event for list item clicked.


abhiomkar:
Currently MDC Menu has custom way to find list item clicked [logic](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/foundation.js#L200) which can be abstracted with custom MDC List event (Example: `MDCList:action` or `MDCList:clicked`).

This custom event can be reused within drawer & menu.

abhiomkar:
Fixed in #4144 & #4151

PR #3572, opened 2018-09-17
feat(dialog): Support reporting action in ancestor element

### LINKED ISSUE #3560: Dialog should detect action from ancestor of event target
The logic landing from #3413 currently only detects actions based on `event.target` directly.

This means that if you have any child elements within an element that represents an action (e.g. an icon inside of a list item), interacting with that child element will not properly reflect that the action was performed.

The dialog code should use `mdc-dom/ponyfill.closest` (#3559) to identify an action associated with interaction events.

kfranqueiro:
Note: This relies on #3559.

Fixes #3560.

### DIFF
--- a/packages/mdc-dialog/foundation.js
+++ b/packages/mdc-dialog/foundation.js
@@ -197,7 +197,7 @@ class MDCDialogFoundation extends MDCFoundation {
    */
   handleClick(evt) {
     // Check for scrim click first since it doesn't require querying ancestors
-    if (this.adapter_.eventTargetHasClass(evt.target, cssClasses.SCRIM) && this.scrimClickAction_) {
+    if (this.adapter_.eventTargetHasClass(evt.target, cssClasses.SCRIM) && this.scrimClickAction_ !== '') {
       this.close(this.scrimClickAction_);
     } else {
       const action = this.adapter_.getActionFromEvent(evt);
@@ -212,7 +212,7 @@ class MDCDialogFoundation extends MDCFoundation {
    * @private
    */
   handleDocumentKeydown(evt) {
-    if ((evt.key === 'Escape' || evt.keyCode === 27) && this.escapeKeyAction_) {
+    if ((evt.key === 'Escape' || evt.keyCode === 27) && this.escapeKeyAction_ !== '') {
       this.close(this.escapeKeyAction_);
     }
   }
--- a/packages/mdc-dialog/foundation.js
+++ b/packages/mdc-dialog/foundation.js
@@ -196,11 +196,14 @@ class MDCDialogFoundation extends MDCFoundation {
    * @private
    */
   handleClick(evt) {
-    const action = this.adapter_.getActionFromEvent(evt);
-    if (action) {
-      this.close(action);
-    } else if (this.adapter_.eventTargetHasClass(evt.target, cssClasses.SCRIM) && this.scrimClickAction_) {
+    // Check for scrim click first since it doesn't require querying ancestors
+    if (this.adapter_.eventTargetHasClass(evt.target, cssClasses.SCRIM) && this.scrimClickAction_) {
       this.close(this.scrimClickAction_);
+    } else {
+      const action = this.adapter_.getActionFromEvent(evt);
+      if (action) {
+        this.close(action);
+      }
     }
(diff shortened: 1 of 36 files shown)


packages/mdc-dialog/index.js
```diff
@@ -171,7 +172,10 @@ class MDCDialog extends MDCComponent {
       releaseFocus: () => this.focusTrap_.deactivate(),
       isContentScrollable: () => !!this.content_ && util.isScrollable(/** @type {!Element} */ (this.content_)),
       areButtonsStacked: () => util.areTopsMisaligned(this.buttons_),
-      getActionFromEvent: (event) => event.target.getAttribute(strings.ACTION_ATTRIBUTE),
+      getActionFromEvent: (event) => {
+        const element = closest(event.target, `[${strings.ACTION_ATTRIBUTE}]`);
```
abhiomkar: Not in the scope of this PR, But, just like MDC List custom event proposal (#3481), button / icon button can also have a custom event that encapsulates this logic.

packages/mdc-dialog/foundation.js
```diff
@@ -196,11 +196,14 @@ class MDCDialogFoundation extends MDCFoundation {
    * @private
    */
   handleClick(evt) {
-    const action = this.adapter_.getActionFromEvent(evt);
-    if (action) {
-      this.close(action);
-    } else if (this.adapter_.eventTargetHasClass(evt.target, cssClasses.SCRIM) && this.scrimClickAction_) {
+    // Check for scrim click first since it doesn't require querying ancestors
+    if (this.adapter_.eventTargetHasClass(evt.target, cssClasses.SCRIM) && this.scrimClickAction_) {
```
abhiomkar: This is bit confusing. Instead of setting scrim action with `setScrimClickAction` `closeOnBackgroundClick(boolean)` would have more intuitive. WDYT?

packages/mdc-dialog/index.js
```diff
@@ -171,7 +172,10 @@ class MDCDialog extends MDCComponent {
       releaseFocus: () => this.focusTrap_.deactivate(),
       isContentScrollable: () => !!this.content_ && util.isScrollable(/** @type {!Element} */ (this.content_)),
       areButtonsStacked: () => util.areTopsMisaligned(this.buttons_),
-      getActionFromEvent: (event) => event.target.getAttribute(strings.ACTION_ATTRIBUTE),
+      getActionFromEvent: (event) => {
```
abhiomkar: Need documentation for `getActionFromEvent` in dialog's README.

packages/mdc-dialog/index.js
kfranqueiro: That's making a strong assumption that those are the only things to ever report an action. List items will also report actions, and I'd rather remain unopinionated on what types of things people assign actions to (e.g. links?)

Previously MDC Dialog has received complaints because it was way too strongly opinionated about actions. The way it's reimplemented makes it pretty flexible.

packages/mdc-dialog/foundation.js
kfranqueiro: 1. Scrim is the word that the [spec](https://material.io/design/components/dialogs.html#anatomy) uses to refer to this element.
2. This API doesn't control a boolean, it controls a string. You can assign any string to the action in case you want to distinguish it for some reason, but both the escape key and scrim click default to reporting `close` action.

This is implemented for flexibility with widely varying use cases of dialogs.

packages/mdc-dialog/index.js
kfranqueiro: Thanks, I forgot to update the wording there.

packages/mdc-dialog/foundation.js
kfranqueiro: Inspired by @patrickrodee in #3576, I added explicit comparisons here to clarify how these variables work.

packages/mdc-dialog/index.js
abhiomkar: It is up to consumer whether to use the custom event or go with traditional approach. The benefit of proxying native click event I think is that not every wrapper has to walk through the DOM to find the closest. But this looks ok for now.

PR #3575, opened 2018-09-17
fix(dialog): Conform more closely with spec

kfranqueiro:
Note: This is based on the dialog-stacked-reverse branch since that already has other golden updates.

This fixes the following:

* Increases spacing between top of dialog and body text when there is no title
* Reduces spacing between title and body text to get as close to spec as possible for alert dialogs
  * In particular, however, the specs WRT simple vs. alert dialogs seem to contradict each other in terms of how much space is between title and content
* Reduces spacing when dialog is scrollable to match spec
  * Also eliminates padding from MDC List for confirmation dialogs, since Dialog styles already add content padding

I've updated goldens, but I think it's out of sync, because the screenshot test report showed "added" chips screenshot tests which are already in master. I'll have to straighten that out after the other dialog PRs get merged.

### DIFF
--- a/packages/mdc-dialog/mdc-dialog.scss
+++ b/packages/mdc-dialog/mdc-dialog.scss
@@ -114,7 +114,7 @@
   flex-grow: 1;
   box-sizing: border-box;
   margin: 0;
-  padding: 10px 24px 20px;
+  padding: 20px 24px; // Note: the top padding is only 20px for dialogs without titles; see below for override
   overflow: auto;
   -webkit-overflow-scrolling: touch;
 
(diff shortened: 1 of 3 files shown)


packages/mdc-dialog/mdc-dialog.scss
```diff
@@ -208,6 +208,25 @@
 
 // postcss-bem-linter: end
 
+.mdc-dialog__title + .mdc-dialog__content {
```
abhiomkar: Move these two style rules next to .mdc-dialog__content above?

kfranqueiro: I'll try that, but I suspect I'll have to exclude them from bem-linter anyway and might also get increasing-specificity warnings. That's why I put them at the end.

kfranqueiro: Yep, had to exclude from bem-linter. No other problems though. Also apparently had a HTML lint error that I fixed.

PR #3576, opened 2018-09-17
chore(ripple): Clean up types

patrickrodee:
After attempting to port MDCRipple to TypeScript, I found the typing of ripple to be unclear or wrong. This PR cleans up the typing of MDCRipple. Some of the main offenders were the `activate` and `deactivate` methods, which took *three* different types: `Event|null|undefined`. I've cleaned them up to only take two types, `Event|undefined`, by removing `null`.

### DIFF
--- a/packages/mdc-ripple/foundation.js
+++ b/packages/mdc-ripple/foundation.js
@@ -32,7 +32,7 @@ import {getNormalizedEventCoords} from './util';
  *   hasDeactivationUXRun: (boolean|undefined),
  *   wasActivatedByPointer: (boolean|undefined),
  *   wasElementMadeActive: (boolean|undefined),
- *   activationEvent: (Event|undefined),
+ *   activationEvent: (!Event|undefined),
  *   isProgrammatic: (boolean|undefined)
  * }}
  */
@@ -51,7 +51,7 @@ let ListenerInfoType;
 /**
  * @typedef {{
  *   activate: function(!Event),
- *   deactivate: function(Event=),
+ *   deactivate: function(!Event=),
  *   focus: function(),
  *   blur: function()
  * }}
@@ -134,13 +134,13 @@ class MDCRippleFoundation extends MDCFoundation {
     /** @private {function(!Event)} */
     this.activateHandler_ = (e) => this.activate_(e);
 
-    /** @private {function(Event=)} */
+    /** @private {function(!Event=)} */
     this.deactivateHandler_ = () => this.deactivate_();
 
-    /** @private {function(Event=)} */
+    /** @private {function(!Event=)} */
     this.focusHandler_ = () => this.handleFocus();
 
-    /** @private {function(Event=)} */
+    /** @private {function(!Event=)} */
     this.blurHandler_ = () => this.handleBlur();
 
     /** @private {!Function} */
@@ -170,7 +170,7 @@ class MDCRippleFoundation extends MDCFoundation {
       this.runDeactivationUXLogicIfReady_();
     };
 
-    /** @private {Event|undefined} */
+    /** @private {!Event|undefined} */
     this.previousActivationEvent_;
   }
 
@@ -195,7 +195,7 @@ class MDCRippleFoundation extends MDCFoundation {
       hasDeactivationUXRun: false,
       wasActivatedByPointer: false,
       wasElementMadeActive: false,
-      activationEvent: null,
+      activationEvent: undefined,
       isProgrammatic: false,
     };
   }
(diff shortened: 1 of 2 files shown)


packages/mdc-ripple/foundation.js
```diff
@@ -325,27 +325,27 @@ class MDCRippleFoundation extends MDCFoundation {
 
     // Avoid reacting to follow-on events fired by touch device after an already-processed user interaction
     const previousActivationEvent = this.previousActivationEvent_;
-    const isSameInteraction = previousActivationEvent && e && previousActivationEvent.type !== e.type;
+    const isSameInteraction = previousActivationEvent && e !== undefined && previousActivationEvent.type !== e.type;
```
kfranqueiro: I see at least 5 places where `!== undefined` was added... but isn't it unnecessary when undefined is the only acceptable falsy value?

packages/mdc-ripple/foundation.js
```diff
@@ -325,27 +325,27 @@ class MDCRippleFoundation extends MDCFoundation {
...
     if (isSameInteraction) {
       return;
     }
 
     activationState.isActivated = true;
-    activationState.isProgrammatic = e === null;
+    activationState.isProgrammatic = e === undefined;
```
kfranqueiro: Come to think of it, would `!e` work here too?

packages/mdc-ripple/foundation.js
```diff
@@ -505,32 +504,27 @@ class MDCRippleFoundation extends MDCFoundation {
...
+    this.deactivate_();
   }
 
   /**
-   * @param {Event} e
    * @param {!ActivationStateType} options
    * @private
    */
-  animateDeactivation_(e, {wasActivatedByPointer, wasElementMadeActive}) {
+  animateDeactivation_({wasActivatedByPointer, wasElementMadeActive}) {
```
kfranqueiro: Nice catch on the event not actually being used in the whole deactivate flow... I must've missed that as I was gradually refactoring stuff out.

packages/mdc-ripple/util.js
```diff
@@ -97,31 +97,46 @@ function supportsCssVariables(windowObj, forceRefresh = false) {
...
- * @return {boolean|{passive: boolean}}
+ * @return {boolean|!EventListenerOptions}
  */
 function applyPassive(globalObj = window, forceRefresh = false) {
   if (supportsPassive_ === undefined || forceRefresh) {
     let isSupported = false;
     try {
       globalObj.document.addEventListener('test', null, {get passive() {
         isSupported = true;
+        return isSupported;
```
kfranqueiro: Adding this here is going to skip caching the result, which is done below.

packages/mdc-ripple/util.js
```diff
@@ -97,31 +97,46 @@ function supportsCssVariables(windowObj, forceRefresh = false) {
...
 /**
  * @param {!Object} HTMLElementPrototype
- * @return {!Array<string>}
+ * @return {string}
  */
 function getMatchesProperty(HTMLElementPrototype) {
-  return [
-    'webkitMatchesSelector', 'msMatchesSelector', 'matches',
-  ].filter((p) => p in HTMLElementPrototype).pop();
+  /**
```
kfranqueiro: I'm a little unclear on how changing this function body is related to the purpose of this PR... AFAICT it doesn't change the behavior of what's returned.

packages/mdc-ripple/foundation.js
patrickrodee: Yeah, `!e` would work, but my aim here is to make our typing more strict and obvious.

packages/mdc-ripple/util.js
patrickrodee: The original function had to guarantee of returning a string since `filter` can return an empty array which causes `pop` to return undefined. We know it will always return a string since every HTMLElement has one of those properties but a type system doesn’t know that. I’m happy to refactor this if you have suggestions!

packages/mdc-ripple/util.js
patrickrodee: The return here is in the context of the getter. I added this to solve for cases where the type system requires all getters to return a value.

packages/mdc-ripple/foundation.js
patrickrodee: This case is to ensure the event object is present. I’d lean towards strict type checking.

packages/mdc-ripple/util.js
kfranqueiro: Oh whoops, sorry, I misread. I see what you mean.

PR #3597, opened 2018-09-20
fix(dialog): Increase z-index above Drawer

### LINKED ISSUE #3596: Drawer z-index causes it to overlap dialogs
When MDC Drawer was updated, its z-index changed:

* It was raised from 5 to 6
* It is now applied to _all_ drawer variants, whereas it used to only be applied to temporary (modal)

Meanwhile, Dialog currently has a z-index of 5.

So I have the following questions:

* Why was the z-index of drawer increased?
* Does it really need to be applied to all variants? (I suspect it does need to be applied to dismissable, which means there's still the potential for conflict with Dialog)

I suspect we might not need to raise Dialog's z-index, if it's feasible for Drawer to be restored to 5, since dialogs are most likely to come after drawers in the DOM, but if we want to be safe we should raise Dialog's as well.

kfranqueiro:
Fixes #3596.

Also adds a screenshot test page.

Before:
before

After:
after

### DIFF
--- a/packages/mdc-dialog/_variables.scss
+++ b/packages/mdc-dialog/_variables.scss
@@ -35,3 +35,5 @@ $mdc-dialog-max-width: 560px;
 $mdc-dialog-margin: 16px;
 
 $mdc-dialog-corner-radius: 4px;
+
+$mdc-dialog-z-index: 7;
--- a/packages/mdc-dialog/mdc-dialog.scss
+++ b/packages/mdc-dialog/mdc-dialog.scss
@@ -56,7 +56,7 @@
 
   // Use `display: none` instead of `visibility: hidden` to avoid recalculating layout when the dialog is closed.
   display: none;
-  z-index: 5;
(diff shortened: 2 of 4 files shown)


amsheehan:
Does this also put it above top app bars?

kfranqueiro:
Yes. See https://github.com/material-components/material-components-web/issues/3596#issuecomment-423288588

kfranqueiro:
Now I'm actually wondering something...

Why is drawer unconditionally above top app bar? Shouldn't that only be necessary for modal drawer? Should other drawer variants actually be _lower_ to support the drawer being clipped by the top app bar? (which would avoid the issue in dialog entirely since modal drawer and modal dialog should never be open at the same time)?

cc @abhiomkar

abhiomkar:
We need the drawer on top of TopAppBar for Dismissible variant where it slides over it on close. So it is intended to be always on top of TopAppBar.

PR #3599, opened 2018-09-20
test(image-list): Add screenshot tests for standard variant

kfranqueiro:
This should help test Shape API updates.


kfranqueiro:
This is now being integrated directly into #3547 instead.

PR #3606, opened 2018-09-21
chore(infrastructure): Reduce dialog flakiness by avoiding reflow

acdvorak:
- Reduces text rendering flakiness on dialog pages in IE 11 💩, Edge, and Firefox
- Avoids reflow by adding CSS classes directly to the HTML:
    - `mdc-dialog--open`
    - `mdc-dialog-scroll-lock` (without this, scrollbars flicker in IE 11 💩 and Edge)
- Fixes a broken regex that caused screenshot tests to be skipped on PRs that only have HTML or YAML file changes 😞

### DIFF
--- a/test/screenshot/golden.json
+++ b/test/screenshot/golden.json
@@ -422,11 +422,11 @@
     }
   },
   "spec/mdc-dialog/classes/overflow-bottom.html": {
-    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/18/18_18_22_548/spec/mdc-dialog/classes/overflow-bottom.html?utm_source=golden_json",
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/21/11_20_28_699/spec/mdc-dialog/classes/overflow-bottom.html?utm_source=golden_json",
     "screenshots": {
       "desktop_windows_chrome@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/18/18_18_22_548/spec/mdc-dialog/classes/overflow-bottom.html.windows_chrome_69.png",
       "desktop_windows_edge@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/17/15_33_59_538/spec/mdc-dialog/classes/overflow-bottom.html.windows_edge_17.png",
-      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/18/18_18_22_548/spec/mdc-dialog/classes/overflow-bottom.html.windows_firefox_62.png"
+      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/21/11_20_28_699/spec/mdc-dialog/classes/overflow-bottom.html.windows_firefox_62.png"
     }
   },
   "spec/mdc-dialog/classes/overflow-top.html": {
--- a/test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
+++ b/test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
@@ -39,7 +39,7 @@
 
       <button class="test-dialog-open-button" data-test-dialog-id="test-dialog" autofocus>Open Dialog</button>
 
-      <div class="mdc-dialog mdc-dialog--open test-dialog test-dialog--simple"
+      <div class="mdc-dialog mdc-dialog--open test-dialog"
            role="alertdialog"
            aria-modal="true"
            aria-labelledby="test-dialog__title--simple"
--- a/test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
+++ b/test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
@@ -34,12 +34,12 @@
     <link rel="stylesheet" href="../../../out/spec/mdc-dialog/fixture.css">
   </head>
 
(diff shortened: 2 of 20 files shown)


test/screenshot/spec/mdc-dialog/fixture.js
```diff
@@ -41,16 +41,21 @@ window.mdc.testFixture.fontsLoaded.then(() => {
...
+        const lastChild = contentEl.children[contentEl.children.length - 1];
+        if (lastChild && lastChild.scrollIntoView) {
+          lastChild.scrollIntoView(/* topAligned */ false);
+        }
       };
-      dialog.listen(strings.OPENING_EVENT, scrollToBottom);
-      dialog.listen(strings.OPENED_EVENT, scrollToBottom);
+
+      const handleOpenEvent = () => {
+        for (let ms = 0; ms < 500; ms += 50) {
```
kfranqueiro: What's the point of doing this incrementally instead of just waiting until the final interval? (Also, is this final interval longer than the system actually waits to take screenshots?)

test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
```diff
@@ -34,12 +34,12 @@
...
   </head>
 
-  <body class="test-container test-container--edge-fonts">
+  <body class="test-container test-container--edge-fonts mdc-dialog-scroll-lock">
     <main class="test-viewport test-viewport--center">
 
       <button class="test-dialog-open-button" data-test-dialog-id="test-dialog" autofocus>Open Dialog</button>
 
-      <div class="mdc-dialog test-dialog"
+      <div class="mdc-dialog mdc-dialog--open test-dialog test-dialog--simple"
```
kfranqueiro: What is the --simple class doing? This is the only place I see it. Ditto the other --simple classes later in this page.

I noticed you removed --alert classes in another file, should these also be removed?

test/screenshot/spec/mdc-dialog/fixture.js
acdvorak: Because the screenshot could potentially get captured _before_ the final interval. This way it covers all the bases.

test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
acdvorak: I split this PR out from #3602. Looks like I missed a class :-)

acdvorak:
1. This PR does not affect the dialog component - only screenshot tests 😀 
2. This PR does not change `golden.json`

test/screenshot/spec/mdc-dialog/fixture.js
kfranqueiro: I'm still confused, because if the screenshots are already captured, what's the point of still looping here?

Like, if the point of this loop is to set up the test page correctly for the screenshot, it seems like the uncertainty of when the screenshot is taken relative to where this loop currently is defeats the point. Either it's scrolled to the bottom at the right time, or it's not, no sense in trying to do it 10 times.

test/screenshot/spec/mdc-dialog/classes/baseline-simple.html
acdvorak: Removed

test/screenshot/spec/mdc-dialog/fixture.js
acdvorak: It's not only the uncertainty of when the screenshot gets taken - it's also the uncertainty of exactly when the browser finishes reflowing after the `--scrollable` class gets added by JS.

When `--scrollable` gets added, it affects the `__content`'s `scrollHeight`, which means we might need to scroll again (depending on the browser). This is what was previously happening in Firefox.

I'll just remove the loop for now. We can always add it back later if necessary.

test/screenshot/golden.json
```diff
@@ -431,11 +431,11 @@
...
-    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/18/18_18_22_548/spec/mdc-dialog/classes/overflow-bottom.html?utm_source=golden_json",
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/21/11_20_28_699/spec/mdc-dialog/classes/overflow-bottom.html?utm_source=golden_json",
     "screenshots": {
       "desktop_windows_chrome@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/18/18_18_22_548/spec/mdc-dialog/classes/overflow-bottom.html.windows_chrome_69.png",
       "desktop_windows_edge@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/17/15_33_59_538/spec/mdc-dialog/classes/overflow-bottom.html.windows_edge_17.png",
-      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/18/18_18_22_548/spec/mdc-dialog/classes/overflow-bottom.html.windows_firefox_62.png"
+      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/21/11_20_28_699/spec/mdc-dialog/classes/overflow-bottom.html.windows_firefox_62.png"
```
kfranqueiro: You said this PR doesn't change any goldens.

I think I might've seen a change like this come up on me exactly once but it seems like a flake to me. Is this consistently reproducible?

kfranqueiro: Whoops, just noticed you edited your earlier comment. I guess that means this change should be consistent now.

PR #3608, opened 2018-09-21
feat(select): Add enhanced select

williamernest:


packages/mdc-select/index.js
```diff
@@ -214,29 +252,109 @@ class MDCSelect extends MDCComponent {
...
+          } else {
+            return listItem.textContent.trim();
+          }
+        }
+        return '';
+      },
+      setValue: (value) => {
+        const element = this.menuElement_.querySelector(`[value="${value}"]`);
+        if (element) {
+          this.setEnhancedSelectedIndex(this.menu_.items.indexOf(element));
```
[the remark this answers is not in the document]
williamernest: The component doesn't maintain a reference to the adapter. Are you saying we should maintain a reference to the adapter in the component ?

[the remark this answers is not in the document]
kfranqueiro: Whoops, I spaced out on that, which explains why it is the way it is.

Is it possible to make the method private if it's only intended to be called from within the component?

packages/mdc-select/index.js
```diff
@@ -40,8 +44,14 @@ class MDCSelect extends MDCComponent {
    */
   constructor(...args) {
     super(...args);
-    /** @private {?Element} */
+    /** @private {?HTMLElement} */
```
kfranqueiro: Doesn't `HTMLElement` have a tendency to fail closure validation whereas `Element` works?

packages/mdc-select/index.js
```diff
@@ -72,66 +86,77 @@ class MDCSelect extends MDCComponent {
...
   }
 
   /**
    * @return {number} The selected index of the select.
    */
   get selectedIndex() {
-    return this.nativeControl_.selectedIndex;
+    let selectedIndex;
+    if (this.menuElement_) {
+      const selectedEl = /** @type{!HTMLElement} */ (this.menuElement_.querySelector(strings.SELECTED_ITEM_SELECTOR));
```
kfranqueiro: Missing space after `@type`

packages/mdc-select/index.js
```diff
@@ -214,29 +286,148 @@ class MDCSelect extends MDCComponent {
...
+      addClass: (className) => this.root_.classList.add(className),
+      removeClass: (className) => this.root_.classList.remove(className),
+      hasClass: (className) => this.root_.classList.contains(className),
+      isRtl: () => window.getComputedStyle(this.root_).getPropertyValue('direction') === 'rtl',
+      setRippleCenter: (normalizedX) => this.lineRipple_ && this.lineRipple_.setRippleCenter(normalizedX),
+      activateBottomLine: () => this.lineRipple_ && this.lineRipple_.activate(),
+      deactivateBottomLine: () => this.lineRipple_ && this.lineRipple_.deactivate(),
+      notifyChange: (evtData) => {
+        evtData.index = this.selectedIndex;
+        this.emit(strings.CHANGE_EVENT, evtData);
```
kfranqueiro: I think we should make this bubble to match native `change` which we're effectively replacing.

packages/mdc-select/index.js
```diff
@@ -150,14 +175,49 @@ class MDCSelect extends MDCComponent {
...
+      // will return false if the menu is still closing and this method listens to the closed event which
+      // occurs after the menu is already closed.
+      this.menuOpened_ = false;
+      if (document.activeElement !== this.selectedText_) {
+        this.foundation_.handleBlur();
+      }
+    });
+
+    // Menu should open to the last selected element.
+    this.menu_.listen(menuSurfaceConstants.strings.OPENED_EVENT, () => {
```
kfranqueiro: Are these listeners ever being removed?

packages/mdc-select/index.js
```diff
@@ -150,14 +175,49 @@ class MDCSelect extends MDCComponent {
...
+      if (this.selectedIndex >= 0) {
+        this.menu_.items[this.selectedIndex].focus();
+      }
+    });
+
+    this.menu_.listen(menuConstants.strings.SELECTED_EVENT, (evtData) => {
+      this.selectedIndex = evtData.detail.index;
```
kfranqueiro: Can we add a comment here that explains that this ultimately runs `handleChange`? Otherwise it requires tracing 2 method calls deep to realize how the change event ends up emitted.

packages/mdc-select/index.js
williamernest: If we change this to emit the `change` event, the native select will emit 2 change events (because this is also emitted during the handleChange event and the native one) . By using our custom change event, it allows you to use the same event for both implementations while only getting 1 event.

kfranqueiro: I think I clarified this offline, but I was referring to the bubbling behavior, not the event name. (Currently this event doesn't bubble, but native `change` does.)

test/screenshot/golden.json
```diff
@@ -855,12 +927,12 @@
...
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/28/20_41_45_808/spec/mdc-select/mixins/shape-radius.html?utm_source=golden_json",
     "screenshots": {
-      "desktop_windows_chrome@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/24/15_50_35_891/spec/mdc-select/mixins/shape-radius.html.windows_chrome_69.png",
-      "desktop_windows_edge@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/24/15_50_35_891/spec/mdc-select/mixins/shape-radius.html.windows_edge_17.png",
-      "desktop_windows_firefox@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/24/15_50_35_891/spec/mdc-select/mixins/shape-radius.html.windows_firefox_62.png",
-      "desktop_windows_ie@11": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/24/15_50_35_891/spec/mdc-select/mixins/shape-radius.html.windows_ie_11.png"
+      "desktop_windows_chrome@latest": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/09/28/20_41_45_808/spec/mdc-select/mixins/shape-radius.html.windows_chrome_69.png",
```
kfranqueiro: In this image, the arrows moved further left for the filled selects, and disappeared for the outlined selects.

test/screenshot/spec/mdc-select/classes/baseline-enhanced.html
```diff
@@ -0,0 +1,160 @@
+<!DOCTYPE html>
```
kfranqueiro: Under `classes/` you went with `...-enhanced` but under `mixins` you went with `enhanced-...`... pick one? I think just `enhanced.html` for this and `enhanced-disabled.html` for the disabled one could work.

test/screenshot/spec/mdc-select/classes/baseline-enhanced.html
```diff
@@ -0,0 +1,160 @@
...
+
+  <body class="test-container">
+    <main class="test-viewport test-viewport--mobile">
+      <div class="test-layout">
+
+        <div class="test-cell test-cell--select">
+          <div class="mdc-select custom-enhanced-select-width">
+            <i class="mdc-select__dropdown-icon"></i>
+            <div class="mdc-select__selected-text"></div>
+            <div class="mdc-select__menu mdc-menu mdc-menu-surface custom-enhanced-select-width">
```
kfranqueiro: This leads me to a question: are we not using JS to auto-size the menu based on the select? If this is necessary to be styled manually to match, this should also be documented.

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+  const {component, menuSurface} = setupTest(hasOutline, hasLabel, hasMockFoundation, hasMockMenu);
+  assert.equal(component.value, '');
+  component.selectedIndex = 2;
+  assert.equal(component.value, 'apple');
+  component.selectedIndex = 1;
+  assert.equal(component.value, 'orange');
+  menuSurface.parentElement.removeChild(menuSurface);
+});
+
+test('#set value calls foundation.handleChange', () => {
```
kfranqueiro: handleChange -> setValue

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+  const hasMockFoundation = true;
+  const hasMockMenu = true;
+  const hasOutline = false;
+  const hasLabel = true;
+  const {component, mockFoundation} = setupTest(hasOutline, hasLabel, hasMockFoundation, hasMockMenu);
+  component.value = 'orange';
+  td.verify(mockFoundation.setValue('orange'), {times: 1});
+});
+
+test('#set selectedIndex calls foundation.handleChange', () => {
```
kfranqueiro: handleChange -> setSelectedIndex

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+  const hasMockFoundation = true;
+  const hasMockMenu = true;
+  const hasOutline = false;
+  const hasLabel = true;
+  const {component, mockFoundation} = setupTest(hasOutline, hasLabel, hasMockFoundation, hasMockMenu);
+  component.selectedIndex = 1;
+  td.verify(mockFoundation.setSelectedIndex(1), {times: 1});
+});
+
+test('#set disabled calls foundation.updateDisabledStyle_', () => {
```
kfranqueiro: `updateDisabledStyle_` -> `setDisabled`

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+test('keydown event is removed from selected-text when destroyed', () => {
+  const {fixture, mockFoundation, component} = setupWithMockFoundation();
+  document.body.appendChild(fixture);
+  component.destroy();
+  domEvents.emit(fixture.querySelector('.mdc-select__selected-text'), 'keydown');
+  td.verify(mockFoundation.handleKeydown(td.matchers.anything()), {times: 0});
+  document.body.removeChild(fixture);
+});
+
+test('open menu when menu is opened does not set ', () => {
```
kfranqueiro: I think this test is unfinished... it looks like the description was cut off and the body looks similar to the keydown event deregistration test.

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+
+  component.destroy();
+  // IE11 mousedown event.
+  event.initMouseEvent('mousedown', true, true, window, 0, 0, 0, clientX, clientY, false, false, false, false, 0, null);
+  fixture.querySelector('.mdc-select__selected-text').dispatchEvent(event);
+
+  td.verify(bottomLine.setRippleCenter(200), {times: 0});
+});
+
+test('menu surface opened event causes the first element (if not element is selected) to be focused', () => {
```
kfranqueiro: You have tests for the opened event but we need one for the menu selected event (its handler is currently not covered).

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+  assert.isFalse(fixture.classList.contains('foo'));
+});
+
+test('adapter#hasClass returns true if a class exists on the root element', () => {
+  const {component, fixture} = setupTest();
+  fixture.classList.add('foo');
+  assert.isTrue(component.getDefaultFoundation().adapter_.hasClass('foo'));
+});
+
+test('adapter_.floatLabel does not throw error if label does not exist', () => {
```
kfranqueiro: Please use `adapter#methodName` notation for all of these descriptions (it's a mixed bag from here down)

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+  assert.isFalse(component.getDefaultFoundation().adapter_.isRtl());
+
+  wrapper.appendChild(fixture);
+  document.body.appendChild(wrapper);
+  assert.isTrue(component.getDefaultFoundation().adapter_.isRtl());
+
+  document.body.removeChild(wrapper);
+});
+
+test('instantiates ripple', function() {
```
kfranqueiro: Move these above adapter tests so that the adapter tests above and below this are together?

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -0,0 +1,654 @@
...
+  component.getDefaultFoundation().adapter_.notchOutline(LABEL_WIDTH, isRtl);
+  td.verify(outline.notch(LABEL_WIDTH, isRtl), {times: 0});
+});
+
+test('adapter#getLabelWidth returns the width of the label', () => {
+  const {component} = setupTest();
+  assert.equal(component.getDefaultFoundation().adapter_.getLabelWidth(), LABEL_WIDTH);
+});
+
+test('adapter#getLabelWidth returns 0 if the label does not exist', () => {
```
kfranqueiro: It looks like we're not actually testing the enhanced-specific adapter methods here, which are probably the most important ones to test in this file in particular... while most of it ends up being covered anyway, the else condition in openMenu and the if condition in closeMenu aren't covered.

demos/select.scss
```diff
@@ -51,4 +54,29 @@
...
+  flex-direction: row;
+  justify-content: space-evenly;
+}
+
+.example {
+  min-width: 300px;
+}
+
+.demo-enhanced-select {
+  width: 400px;
```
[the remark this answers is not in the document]
williamernest: Yes, it has to be set explicitly or the width will change based on the option selected. It also needs to be added to the menu itself, since the menu is hoisted when the select is initialized. 

I'll add this to the docs.

packages/mdc-select/index.js
```diff
@@ -168,37 +214,78 @@ class MDCSelect extends MDCComponent {
...
+    };
+
+    const element = this.nativeControl_ ? this.nativeControl_ : this.selectedText_;
+
+    element.addEventListener('change', this.handleChange_);
+    element.addEventListener('focus', this.handleFocus_);
+    element.addEventListener('blur', this.handleBlur_);
+    element.addEventListener('keydown', this.handleKeydown_);
+
+    ['mouseup', 'touchend'].forEach((evtType) => {
```
kfranqueiro: Listening to up events here causes 2 problems.

* This causes line ripple to behave incorrectly, since it happens before the ripple center is set. (In actual testing, it seems to come from the center on the first time, and on subsequent times I'm not seeing it ripple at all, it seems like the border is just growing vertically.)
* On the enhanced select, this triggers the menu to open as long as you release it over the select - regardless of whether the initial press happened on the select or not.

I suspect we're going to need to handle both down and up events if we go this route...

* Down would set the ripple center
* Down would set a flag that up would check... but this flag would also need to be cleared on up events anywhere else on the document to clear the flag in case the user dragged off of the select

This is becoming rather complex... IIRC the original problem is that reacting to click doesn't give the selected text time to become focused first. Would a setTimeout within the handler logic end up being simpler and allow us to continue listening to `click`?

packages/mdc-select/foundation.js
```diff
@@ -77,47 +84,89 @@ class MDCSelectFoundation extends MDCFoundation {
...
+    this.adapter_.setValue(value);
+    const didChange = true;
+    this.handleChange(didChange);
+  }
+
+  getValue() {
+    return this.adapter_.getValue();
+  }
+
+  setDisabled(isDisabled) {
```
kfranqueiro: I noticed you removed the code that adds/removes a CSS class in `setDisabled`... isn't that still needed?

It looks like you baked in the root class addition/removal into the adapter for enhanced only, but AFAICT it is still needed for both variants, so I'm not sure why that was removed from the foundation.

packages/mdc-select/index.js
```diff
@@ -214,29 +301,148 @@ class MDCSelect extends MDCComponent {
...
+      addClass: (className) => this.root_.classList.add(className),
+      removeClass: (className) => this.root_.classList.remove(className),
+      hasClass: (className) => this.root_.classList.contains(className),
+      isRtl: () => window.getComputedStyle(this.root_).getPropertyValue('direction') === 'rtl',
+      setRippleCenter: (normalizedX) => this.lineRipple_ && this.lineRipple_.setRippleCenter(normalizedX),
+      activateBottomLine: () => this.lineRipple_ && this.lineRipple_.activate(),
+      deactivateBottomLine: () => this.lineRipple_ && this.lineRipple_.deactivate(),
+      notifyChange: (evtData) => {
+        evtData.index = this.selectedIndex;
+        this.emit(strings.CHANGE_EVENT, evtData, true);
```
kfranqueiro: Add `/* shouldBubble */` comment after `true`

test/unit/mdc-select/foundation.test.js
```diff
@@ -164,15 +167,129 @@ test('#handleFocus calls adapter.activateBottomLine()', () => {
   td.verify(mockAdapter.activateBottomLine(), {times: 1});
 });
 
+test('#handleFocus does not call adapter.activateBottomLine() if isMenuOpen=true', () => {
+  const {foundation, mockAdapter} = setupTest();
+  td.when(mockAdapter.isMenuOpen()).thenReturn(true);
+  foundation.handleFocus();
+  td.verify(mockAdapter.activateBottomLine(), {times: 1});
```
kfranqueiro: This test's verification was changed to contradict its description. What's the right answer?

test/unit/mdc-select/foundation.test.js
```diff
@@ -164,15 +167,129 @@ test('#handleFocus calls adapter.activateBottomLine()', () => {
...
+
+test('#handleClick opens the menu if the select is focused and isMenuOpen=false', () => {
+  const {foundation, mockAdapter} = setupTest();
+  td.when(mockAdapter.isMenuOpen()).thenReturn(false);
+  td.when(mockAdapter.hasClass(td.matchers.anything())).thenReturn(true);
+  foundation.handleClick(0);
+  td.verify(mockAdapter.openMenu(), {times: 1});
+});
+
+test('#handleKeydown calls adapter.openMenu if valid keys is pressed, menu is not open and select is focused',
```
kfranqueiro: is pressed -> are pressed

test/unit/mdc-select/foundation.test.js
```diff
@@ -164,15 +167,129 @@ test('#handleFocus calls adapter.activateBottomLine()', () => {
...
+  () => {
+    const {foundation, mockAdapter} = setupTest();
+    const preventDefault = td.func();
+    const event = {key: 'Enter', preventDefault};
+    td.when(mockAdapter.isMenuOpen()).thenReturn(false);
+    td.when(mockAdapter.hasClass('mdc-select--focused')).thenReturn(true);
+    foundation.handleKeydown(event);
+    event.key = 'Space';
+    foundation.handleKeydown(event);
+    event.key = '';
```
kfranqueiro: No `event.key` tests for arrows?

packages/mdc-select/index.js
```diff
@@ -214,29 +301,148 @@ class MDCSelect extends MDCComponent {
...
+          }
+        }
+        return this.selectedText_.textContent.trim();
+      },
+      setValue: (value) => {
+        const element = /** @type {HTMLElement} */ (this.menuElement_.querySelector(`[value="${value}"]`));
+        if (element) {
+          this.setEnhancedSelectedIndex_(this.menu_.items.indexOf(element));
+        } else {
+          this.selectedText_.textContent = value;
```
kfranqueiro: Falling back to just directly setting `selectedText_.textContent` (and falling back to reading it in `getValue`) also opens the floodgates to this not really behaving like a select anymore (i.e. you can pass in any value, not limited to one in the list).

Moreover, `getValue` falls back to reading the list item's `textContent`, but there's no representation of that case in `setValue`, which IIUC means that list items will never get properly selected if you don't include the `value` attribute, so there's very few cases where that `getValue` branch will be useful.

This all makes me wonder if we should just enforce usage of the attribute.

Meanwhile... is `value` invalid on these elements? Should we be using something like `data-value` instead for enhanced select?

packages/mdc-select/index.js
williamernest: I'll convert these back to `mousedown`/`touchstart` and handle updating this logic in a later PR.

packages/mdc-select/index.js
williamernest: `value` is valid on `li` elements. 

I'll update the PR to indicate that using `value` is required.

kfranqueiro: Hm. It's valid, but not the way we're using it.

> This integer attribute indicates the current ordinal value of the list item as defined by the `<ol>` element. The only allowed value for this attribute is a number, even if the list is displayed with Roman numerals or letters. List items that follow this one continue numbering from the value set. The value attribute has no meaning for unordered lists (`<ul>`) or for menus (`<menu>`).

williamernest: So it's valid, but it has no meaning. Why shouldn't we use it?

kfranqueiro: image

(If you switch to `ol` for argument's sake, it turns into an error about encountering something other than a `-` or digit.)

I expect someone will log this as an issue, requiring us to change the required markup for the component afterwards, if we don't do it initially.

packages/mdc-select/README.md
```diff
@@ -64,17 +59,56 @@ npm install @material/select
...
 
+When using the enhanced select, you will also require the menu components. 
+
+```scss
+@import "@material/list/mdc-list";
+@import "@material/menu-surface/mdc-menu-surface";
+@import "@material/menu/mdc-menu";
+@import "@material/select/mdc-select";
+```
+
```
kfranqueiro: If we're discussing enhanced select here, should we also mention setting width of the select and menu?

packages/mdc-select/README.md
```diff
@@ -235,21 +348,40 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
...
-| `hasLabel() => boolean` | Returns true if the `select` has a label associated with it. |
 | `floatLabel(value: boolean) => void` | Floats or defloats label. |
 | `getLabelWidth() => number` | Returns the offsetWidth of the label element. |
 | `hasOutline() => boolean` | Returns true if the `select` has the notched outline element. |
 | `notchOutline(labelWidth: number, isRtl, boolean) => void` | Switches the notched outline element to its "notched state." |
 | `closeOutline() => void` | Switches the notched outline element to its closed state. |
+| `openMenu() => void` | Causes the menu element in the enhanced select to open. |
+| `closeMenu() => void` | Causes the menu element in the enhanced select to close. |
+| `setValue(value: string) => void` | Sets the value of the select or text content of the selected-text element. |
+| `isMenuOpen() => boolean` | Returns true if the menu is currently opened. |
```
kfranqueiro: add "in the enhanced select" after menu?

packages/mdc-select/README.md
```diff
@@ -235,21 +348,40 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
...
 | `handleFocus() => void` | Handles a focus event on the `select` element. |
 | `handleBlur() => void` | Handles a blur event on the `select` element. |
+| `handleClick(normalizedX: number) => void` | Sets the line ripple center to the normalizedX for the line ripple. |
 | `handleChange() => void` | Handles a change to the `select` element's value. This must be called both for `change` events and programmatic changes requested via the component API. |
+| `handleKeydown(event: Event) => void` | Handles opening the menu (enhanced select) when the `mdc-select__selected-text` element is focused and the user presses the `Enter` or `Space` key. |
+| `setSelectedIndex(index: number) => void` | Handles setting the `mdc-select__selected-text` element and closing the menu (enhanced select only). Also causes the label to float and outline to notch if needed. |
+| `setValue(value: string) => void` | Handles setting the value through the adapter and causes the label to float and outline to notch if needed. |
+| `getValue() => string` | Handles getting the value through the adapter. |
```
kfranqueiro: Should this not be public on the foundation? It seems like the foundation only needs it internally (or could just call the adapter API each time).

packages/mdc-select/mdc-select.scss
```diff
@@ -205,3 +225,8 @@
   cursor: default;
   pointer-events: none;
 }
+
+.mdc-select__menu .mdc-list .mdc-list-item--selected {
+  @include mdc-theme-prop(color, on-surface);
+  @include mdc-states;
```
kfranqueiro: I'd suggest specifying `on-surface` here so it's clear this is intentionally aligning with the color above. (Took me a minute to realize on-surface is currently the default for states.)

test/unit/mdc-select/foundation.test.js
```diff
@@ -59,22 +58,21 @@ function setupTest() {
   return {mockAdapter, foundation};
 }
 
-test('#updateDisabledStyle(true) calls adapter.addClass', () => {
+test('#setDisabled(true) calls adapter.addClass', () => {
```
kfranqueiro: This should presumably be testing both addClass and setDisabled now?

test/unit/mdc-select/foundation.test.js
```diff
@@ -59,22 +58,21 @@ function setupTest() {
...
   const {mockAdapter, foundation} = setupTest();
-  foundation.updateDisabledStyle(true);
-  td.verify(mockAdapter.addClass(MDCSelectFoundation.cssClasses.DISABLED));
+  foundation.setDisabled(true);
+  td.verify(mockAdapter.setDisabled(true));
 });
 
-test('#updateDisabledStyle(false) calls adapter.removeClass', () => {
+test('#setDisabled(false) calls adapter.removeClass', () => {
```
kfranqueiro: This should presumably be testing both removeClass and setDisabled now?

test/unit/mdc-select/foundation.test.js
```diff
@@ -164,15 +167,145 @@ test('#handleFocus calls adapter.activateBottomLine()', () => {
...
+  const {foundation, mockAdapter} = setupTest();
+  td.when(mockAdapter.isMenuOpen()).thenReturn(false);
+  foundation.handleClick(0);
+  td.verify(mockAdapter.setRippleCenter(0), {times: 1});
+});
+
+test('#handleClick opens the menu if the select is focused and isMenuOpen=false', () => {
+  const {foundation, mockAdapter} = setupTest();
+  td.when(mockAdapter.isMenuOpen()).thenReturn(false);
+  td.when(mockAdapter.hasClass(td.matchers.anything())).thenReturn(true);
```
kfranqueiro: change `td.matchers.anything()` to `cssClasses.FOCUSED`? It's unclear whether there are intentional or unintentional side effects of this matcher.

packages/mdc-select/README.md
williamernest: We route the component `value` getter through the foundation.

kfranqueiro: Whoops, I see. That's probably fair then, since both the foundation and the component need it.

williamernest:
Updated

kfranqueiro:
I just tested the demo page on Chrome/Firefox OS X and Edge/IE Windows. On all but Firefox, space/arrows no longer open native selects for some reason. Further, Edge/IE don't seem to respond to arrow keys while the native select is open. Firefox OS X seems to be fine.

These cases work on our [catalog page](https://material-components.github.io/material-components-web-catalog/#/component/select).

PR #3617, opened 2018-09-21
feat(chips): make deselect and toggleSelect private. Update handleChipInteraction/Removal API

### LINKED ISSUE #3612: Update ChipSet to make deselect private
Deselect does not need to be public. `deselect` only needs to be called from `handleChipRemoval`. Users can programmatically remove chips by setting `select` on the chip component.

### LINKED ISSUE #3613: update `handleChipRemoval` to accept a chipId (not event)
handleChipRemoval only uses the chipId. External libraries including MDC React need to create a "fake" event to call this function. To simplify this for external libs, `handleChipRemoval` should be updated to only accept a chipId.

### LINKED ISSUE #3615: make toggleSelect private
ChipSet does not need a deselect method, nor should there need to be 2 methods (select and toggleSelect) to select a chip. Instead what external libraries should be calling `handleChipInteraction`.

moog16:
Fixes #3612 
Fixes #3613 
Fixes #3615 

BREAKING CHANGE: `deselect` and `toggleSelect` are private methods. `handleChipInteraction` and `handleChipRemoval` now accept `chipId` instead of an event.


test/unit/mdc-chips/mdc-chip-set.test.js
```diff
@@ -89,12 +89,18 @@ test('#destroy cleans up child chip components', () => {
...
-  domEvents.emit(root, MDCChipFoundation.strings.INTERACTION_EVENT);
-  td.verify(mockFoundation.handleChipInteraction(td.matchers.anything()), {times: 1});
-
-  domEvents.emit(root, MDCChipFoundation.strings.REMOVAL_EVENT);
-  td.verify(mockFoundation.handleChipRemoval(td.matchers.anything()), {times: 1});
+  const evtData = {
+    detail: {chipId: 'chipA'},
+    bubbles: true,
+  };
+  const evt1 = new CustomEvent(MDCChipFoundation.strings.INTERACTION_EVENT, evtData);
```
kfranqueiro: I don't think this is going to fly in IE. (@williamernest can probably give you more details quickly.)

moog16: Yup, talked to Will - thanks for catching it. Will fix it.

kfranqueiro:
2 questions:

* Given that deselect and toggleSelect are now private, what still uses `select` publicly?
* We still have a unit test testing `deselect_` directly even though it's a private API now... should we rework that test to test the API that calls `deselect_` instead?

moog16:
If you want to initialize a chip as selected, `select` will need to be called.

Should we not test private methods? I can add a test that triggers the `_deselect` method to be thorough.

kfranqueiro:
RE `select`, I see it now in the component's `initialSyncWithDOM`.

We typically directly test public APIs with enough cases to cover all private code. Testing private methods directly veers towards testing implementation details.

moog16:
@kfranqueiro ... pushed changes for tests.

RE `select`, I think you're on to something. I think it should be private too. Instead of using it in `initialSyncWithDOM`, the ChipSet should instead listen to the `SELECTION_EVENT`. If you agree, I can do the changes/update the tests and Readme.

kfranqueiro:
I don't necessarily think the use of `select` is invalid... How would listening to the event work for initially-selected chips at page load based on applied classes?

moog16:
@rlfriedman what do you think about my above suggestion? Does that work for you?

moog16:
ChipSet will be initialized with the SELECTION_EVENT listening. The Chips will each be initialized and put into selection state: `chip.selected = true`, which emits the SELECTION_EVENT. Currently the ChipSet already has a listener, which looks like it already calls the `foundation.handleChipSelection`, which calls `foundation.select` (which is what the `initialSyncWithDOM` calls). So seems like what we have now is redundant.

kfranqueiro:
I'm not confident that's the case, because AFAICT nothing else will ever actually invoke the Chip's selected setter when it initializes. Chip Set's `initialSyncWithDOM` invokes Chip's selected _getter_ which determines its result based on the already-applied CSS class.

I'm also not sure it'd be correct to change that, since emitting events for something that arguably has already happened might seem unnatural.

moog16:
I see...I did some more investigation and it looks like the demo only initializes the ChipSet, which also initializes all it's Chips, which is why it is needed. So nevermind. :)

moog16:
@kfranqueiro - is this PR good to go?

kfranqueiro:
Yeah I think so.

PR #3631, opened 2018-09-24
chore(infrastructure): Run `npm run build` on Travis CI

### LINKED ISSUE #3625: Travis CI should test the all-in-one CSS build
#3624 slipped through because the screenshot tests only compile the a-la-carte CSS targets, not the all-in-one target (so `material-components-web.scss` effectively isn't tested).

We should aso run the all-in-one CSS target at least on CI for the purpose of validating that it compiles. (The screenshot testing CI task currently doubles as our compilation check.)

acdvorak:
Fixes #3625


kfranqueiro:
If we're just going to additionally run `npm run dist`...

* Do it as a separate CI task (like we [used to do](https://github.com/material-components/material-components-web/blob/v0.37.0/.travis.yml#L7-L10))
* Run `build`, not `dist` - `build` is enough to flag issues in the all-in-one targets
* Add conditions for running the new CI task, and update conditions for running the screenshot tests since we no longer need to do so for the purpose of build sanity

acdvorak:
Done!

I left the conditions for running screenshot tests as-is because I think they still make sense:

https://github.com/material-components/material-components-web/blob/9842a07e82db44963d40af2f5713f9de44a804bd/scripts/travis-env-vars.sh#L100

We want to run screenshot tests whenever package files or screenshot test files change.

PR #3635, opened 2018-09-25
feat(checkbox): Declare all Sass variables as `!default`

mathiasschopmans:
All Sass variables are now declared as `!default` so they can be overridden by clients.

Refs #3708


kfranqueiro:
Some of these variables are really constants based on calculations, which it doesn't make sense to override. Meanwhile, others are already overridable via mdc-checkbox's mixins.

Is there a particular use case you can't accomplish using mdc-checkbox's mixins?

mathiasschopmans:
I get your point for the calculated ones like `$mdc-checkbox-ui-pct`.

Yep, I'm able to style them via the mixins, **but** this would result in superfluous CSS declarations and since every byte in the output counts, I thought it would make more sense to be able to change the variables beforehand. 

**Currently:**
```sass
@import "@material/checkbox/mdc-checkbox"
@include mdc-checkbox-container-colors($unmarked-stroke-color: $mdc-theme-primary, $marked-fill-color: $mdc-theme-primary, $generate-keyframes: false)
@include mdc-checkbox-focus-indicator-color($mdc-theme-primary)
```

**After:**
```sass
$mdc-checkbox-baseline-theme-color: primary
@import "@material/checkbox/mdc-checkbox"
```

What's your opinion, @williamernest? :)

williamernest:
Just to circle back on this, we plan on implementing this by adding !default to all of our global scss variables, but want to do it on a much wider scale and also add a strong statement in our Readme that indicates you're on your own if you modify variables and something breaks.

PR #3645, opened 2018-09-25
docs(chips): Document events and addChip usage

### LINKED ISSUE #3282: mdc-tab, mdc-tab-bar: Update icon usage in READMEs
* [x] Use `material-icons` and reference a valid icon (e.g. `favorite` instead of `heart`) (#3390, #3391)
* [ ] Document inclusion of Material Icons consistently with #3266 once that's settled and merged
* [ ] Add `aria-hidden="true"` to icons in examples where Material Icons are accompanied by text labels

kfranqueiro:
Fixes #3282.

Also removes a property from removal events that AFAICT is unused.


kfranqueiro:
I was a little wary of doing that since it'd also affect the notes below the table, but I think I've made it work, given that there's also an events table for the vanilla DOM implementation now.

PR #3652, opened 2018-09-26
chore(infrastructure): Increase max CBT parallels off-hours

acdvorak:
### What it does

Increases the default number of parallel CBT browsers from **`2`** to **`4`** when running outside of \
normal business hours (M–F 9–6).

This saves about 5–10 minutes on each Travis job.

### New defaults

When nobody else is running screenshot tests:

* **Outside of normal business hours** (evenings and weekends):\
    Defaults to **`4`** parallel browsers 🆕
* **During normal business hours** (Monday through Friday, 9am EST to 6pm PST): \
    Defaults to **`2`** parallel browsers (same behavior as before)

When other users are already running screenshot tests:

* **Any day/time**: \
    Defaults to **`1`** parallel browser (same behavior as before)

### CLI overrides

Just like before, you can override the default number of parallels with the `--parallels=N` CLI option:

```bash
npm run screenshot-test -- --parallels=4
```

### DIFF
--- a/test/screenshot/infra/lib/selenium-api.js
+++ b/test/screenshot/infra/lib/selenium-api.js
@@ -392,10 +392,15 @@ class SeleniumApi {
         return Math.min(cliParallels, available);
       }
 
-      // If nobody else is running any tests, run half the number of concurrent tests allowed by our CBT account.
-      // This gives us _some_ parallelism while still allowing other users to run their tests.
+      // Nobody else is running tests
       if (active === 0) {
-        return Math.floor(max / 2);
+        if (this.isBusinessHours_()) {
+          // Run half the number of concurrent tests allowed by our CBT account.
+          // This gives us _some_ parallelism while allowing other users to run their tests.
+          return Math.floor(max / 2);
+        } else {
+          return available;
+        }
       }
 
       // If someone else is already running tests, only run one test at a time.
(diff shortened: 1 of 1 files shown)


test/screenshot/infra/lib/selenium-api.js
```diff
@@ -972,6 +977,25 @@ class SeleniumApi {
...
   }
 
+  /**
+   * @return {boolean}
+   * @private
+   */
+  isBusinessHours_() {
+    const MORNING_UTC_HOURS = 13; // 1pm UTC === 9am EST
+    const EVENING_UTC_HOURS = 1; // 1am UTC === 6pm PST
+    const SATURDAY = 7;
```
kfranqueiro: Saturday is 6, not 7. (There are not 8 days in a week, as much as we might wish there were.)

test/screenshot/infra/lib/selenium-api.js
```diff
@@ -972,6 +977,25 @@ class SeleniumApi {
...
+    const SUNDAY = 0;
+
+    const nowDate = new Date();
+    const nowUtcHours = nowDate.getUTCHours();
+
+    const isWeekDay = nowDate.getDay() > SUNDAY || nowDate.getDay() < SATURDAY;
```
kfranqueiro: This is determining `getDay` by local time. What time zone does Travis run in?

Asking because 12-1am UTC Saturday would currently return false (assuming we fix `SATURDAY` above).

test/screenshot/infra/lib/selenium-api.js
acdvorak: Good catch. I'll create a follow up PR. Thanks!

test/screenshot/infra/lib/selenium-api.js
acdvorak: Travis runs on EC2 in the US. I'm not sure exactly which region. I'll create a follow up PR to explicitly set the timezone. Thanks!

PR #3658, opened 2018-09-26
chore: Fix copyright notice in HTML/Sass files

acdvorak:
The MIT copyright notice in all Sass files has an extraneous trailing `//` on two lines:

```
furnished to do so, subject to the following conditions://

The above copyright notice and this permission notice shall be included in
all copies or substantial portions of the Software.//
```

Several HTML files also had an extraneous trailing `.`:

```
THE SOFTWARE..
```


kfranqueiro:
I had noticed this a while ago and started a local branch for it, but held off on PR-ing it until we were also going to merge #3245 since both of these are likely to cause version revs on every package.

In addition to the scss files, there are also extraneous duplicate periods in some of the spec html files. Look for `THE SOFTWARE..` and change to `THE SOFTWARE.`

Issue #3486, opened 2018-08-31
Menu with Surface prevents links with href from firing


nerg4l:
### What MDC Web Version are you using?

0.39.0

### What browser(s) is this bug affecting?

All

### What OS are you using?

> Windows 10 Pro N 1709

### What are the steps to reproduce the bug?

1. Create a mdc Menu with anchor
2. Add anchor link with href to mdc Menu as a list item
3. Open menu
4. Click on item

[CodePen](https://codepen.io/nerg4l/pen/VGPzbX)

### What is the expected behavior?

Execute event of `a` tag (anchor link)

### What is the actual behavior?

Menu executes `preventDefault`

### Any other information you believe would be useful?

When a list item is clicked [`handleAction_`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/foundation.js#L115) of menu is called. As a list item is clicked `handleSelection_` and `preventDefaultEvent_` methods are called.

`handleSelection_` in this cases only executes the close animation.

[`preventDefaultEvent_`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/foundation.js#L221) executes `preventDefault` at line [225](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu/foundation.js#L225) as `a` tag is not added to `ELEMENTS_KEY_ALLOWED_IN`.

ahlechandre:
Menus are my nightmare with MDC. Before Menu Surface, we couldn't use it inside a list item, now that we can, it prevents links default behavior...

williamernest:
@Nerg4l Thanks for filing this issue. This was an oversight when we built the new menu and it should be resolved in the next major release.

stof:
note that adding `a` in the list of `ELEMENTS_KEY_ALLOWED_IN` might not resolve things entirely. If you have a `<span>` inside your `<a>`, the target will not always be the `<a>` (it can be the `<span>`, but the default action would still execute the action of the link). We need to make sure that case work too.

robbyrice:
@williamernest is there any update as to when this will be fixed? I just updated to 0.40.0 and it appears that links inside a menu still do not work correctly.

AlexZarivayskiy:
Perhaps, this is a temporary solution, until that issue will be solved.
```
...
<li class="mdc-list-item" role="menuitem" onclick="location.href='myURL'">
...
```

PR #3680, opened 2018-10-01
fix(menu): Allow anchor links as menu list items

jamesmfriedman:
The current logic prevents anchor links from firing when clicked on in the menu, a fairly common case.
fixes: #3486


williamernest:
This doesn't work when the user clicks on the text or graphic of the link. We can set `pointer-events: none;` on the `mdc-list-item__text` and `mdc-list-item__grapic` classes when the `mdc-list-item` is a link. 

Adding the following to the end of the `mdc-menu` 
```
.mdc-menu {
...
  //stylelint-disable selector-max-type, selector-no-qualifying-type
  a.mdc-list-item .mdc-list-item__text,
  a.mdc-list-item .mdc-list-item__graphic {
    pointer-events: none;
  }
  // stylelint-enable selector-max-type, selector-no-qualifying-type
}
```

jamesmfriedman:
Ah, yeah that makes sense.

williamernest:
@jamesmfriedman Can you give me access to your repo to add this snippet?

jamesmfriedman:
Sorry for the delay. Added you as a collaborator.

jamesmfriedman:
Thanks @williamernest!

yguedidi:
@jamesmfriedman @williamernest thanks for the fix! But it looks like it doesn't solve the following comment: https://github.com/material-components/material-components-web/issues/3486#issuecomment-423485685

williamernest:
@yguedidi we set pointer events none on the elements within the list to ensure the anchor tag gets clicked and we don't prevent default on the event. Any custom markup you add inside the list item should also set pointer events to none.

PR #3723, opened 2018-10-09
fix(select): Fix color for dropdown arrow when --focused is applied

williamernest:


williamernest:
The shape-radius screenshot test is a merge issue in the feature branch. It was missing the dropdown icon and still had some unused styles. The goldens from that PR have the icon in it's original location. I'll fix them with this PR (although I've fixed them in 3 different PR's now, this one ~should~ make it in first)

PR #3724, opened 2018-10-10
feat(select): Add invalid

williamernest:


packages/mdc-select/mdc-select.scss
```diff
@@ -108,6 +108,38 @@
   @include mdc-select-outlined_;
 }
 
+.mdc-select--invalid {
+  @include mdc-select-label-color($mdc-select-label-error-color);
+  @include mdc-select-bottom-line-color($mdc-select-bottom-line-idle-error-color);
+  @include mdc-select-focused-bottom-line-color(error);
```
kfranqueiro: I'm unclear on why a few of these don't have individual variables while the rest do. (I'm also not clear on whether less or more is the right answer... if they're all error, do they all deserve separate variables?)

Issue #3629, opened 2018-09-24
Select: Add support for leading icons (static) 


williamernest:
Add support for leading icons to match text field. This will also cause the menu options to have extra left padding to ensure the selected text and the menu options are left aligned.

PR #3745, opened 2018-10-12
feat(select): Add leading icon support

williamernest:
Adds leading icon support to the `mdc-select` component. The icon JS is largely a copy of the `MDCTextFieldIcon` element to keep the two components nearly identical. The scss has been modified to only accommodate a single icon as well as to support SVG elements as leading icons for the select. 
refs: #3629

### DIFF
--- a/packages/mdc-select/README.md
+++ b/packages/mdc-select/README.md
@@ -291,6 +291,11 @@ programmatically select a disabled list item in the enhanced select.
 </div>
 ```
 
+### Select with Leading  Icons
+
+Leading icons can be added within the default or outlined variant of MDC Select as visual indicators as
+well as interaction targets. See [here](icon/) for more information on using icons.
+
 ## Style Customization
 
 #### CSS Classes
@@ -300,10 +305,12 @@ programmatically select a disabled list item in the enhanced select.
 | `mdc-select` | Mandatory. |
 | `mdc-select__menu` | Mandatory when using the enhanced select. This class should be placed on the `mdc-menu` element within the `mdc-select` element. |
 | `mdc-select__dropdown-icon` | Mandatory. Should be placed on an `i` element within the `mdc-select` element. Used for the dropdown arrow svg and animation.
+| `mdc-select__icon` | Optional. Should be placed on an `i` or `svg` element within the `mdc-select` element. Used for the leading icon. 
 | `mdc-select--disabled` | Optional. Styles the select as disabled. This class should be applied to the root element when the `disabled` attribute is applied to the `<select>` element. |
 | `mdc-select--outlined` | Optional. Styles the select as outlined select. |
 | `mdc-select__native-control` | Mandatory for the native select. The native `<select>` element. |
 | `mdc-select__selected-text` | Mandatory for the enhanced select. This class should be placed on a `div` within the `mdc-select` element. |
+| `mdc-select--with-leading-icon` | Styles the select as a select with a leading icon. |
 
 > Note: To further customize the [MDCMenu](./../mdc-menu) or the [MDCList](./../mdc-list) component contained within the select, please refer to their respective documentation.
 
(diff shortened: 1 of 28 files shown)


packages/mdc-select/README.md
```diff
@@ -386,9 +395,13 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
...
+| `setLeadingIconAriaLabel(label: string) => void` | Sets the aria label of the leading icon. |
+| `setLeadingIconContent(content: string) => void` | Sets the text content of the leading icon. |
 
 ### Events
 
 Event Name | Data | Description
 --- | --- | ---
 `MDCSelect:change` | `{value: string, index: number}` | Used to indicate when an element has been selected. This event also includes the value of the item and the index.
+
+`MDCTextFieldFoundation` supports an optional icon sub-element. The foundation of the icon must be passed in as constructor arguments to `MDCSelectFoundation`.
```
kfranqueiro: `MDCTextFieldFoundation` -> `MDCSelectFoundation`

packages/mdc-select/_mixins.scss
```diff
@@ -218,3 +218,153 @@
     pointer-events: none;
   }
 }
+
+@mixin mdc-select-with-leading-icon_ {
+  @include mdc-select-icon-horizontal-position_(16px, 48px);
+
+  &.mdc-select--outlined {
+    @include mdc-floating-label-float-position($mdc-select-outlined-label-position-y, $mdc-select-outlined-with-leading-icon-label-position-x);
+    @include mdc-floating-label-shake-animation(text-field-outlined-leading-icon);
```
kfranqueiro: This is currently assuming that text field is also installed and included. Should we replicate the shake keyframes for select, since I assume we're not currently making the jump to depending on text field?

packages/mdc-select/_variables.scss
```diff
@@ -28,21 +28,23 @@ $mdc-select-height: 56px;
...
 $mdc-select-bottom-line-hover-color: rgba(mdc-theme-prop-value(on-surface), .87);
 
-$mdc-select-box-fill-color: mix(mdc-theme-prop-value(on-surface), mdc-theme-prop-value(surface), 4%);
-$mdc-select-box-disabled-fill-color: mix(mdc-theme-prop-value(on-surface), mdc-theme-prop-value(surface), 2%);
+$mdc-select-fill-color: mix(mdc-theme-prop-value(on-surface), mdc-theme-prop-value(surface), 4%);
 
 $mdc-select-outlined-idle-border: rgba(mdc-theme-prop-value(on-surface), .24);
+$mdc-select-dropdown-opacity: .54;
+
+/** Disabled Styles **/
```
kfranqueiro: Change to `//` comment

packages/mdc-select/icon/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Select Icon
+
+Icons describe the type of input a select requires. They can also be interaction targets.
+
+## Design & API Documentation
+
+<ul class="icon-list">
+  <li class="icon-list-item icon-list-item--spec">
+    <a href="https://material.io/go/design-selects#selects-layout">Material Design guidelines: Text Fields Layout</a>
```
kfranqueiro: I think maybe something got replaced here that shouldn't have - this URL doesn't exist

packages/mdc-select/icon/_mixins.scss
```diff
@@ -0,0 +1,75 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+@import "@material/theme/mixins";
+@import "./variables";
+
+// Public mixins
+
+@mixin mdc-select-icon-color_($color) {
```
kfranqueiro: Move this below the public mixin, and move the "private mixins" comment up above the two private mixins that are currently above it

test/unit/mdc-select/foundation.test.js
```diff
@@ -305,9 +342,34 @@ test('#layout calls notchOutline(true) if value is not an empty string', () => {
...
+  td.verify(leadingIcon.setAriaLabel('foo'), {times: 1});
+});
+
+test('#setLeadingIconContent sets the content of the leading icon element', () => {
+  const {foundation, leadingIcon} = setupTest();
+  foundation.setLeadingIconContent('foo');
+  td.verify(leadingIcon.setContent('foo'), {times: 1});
+});
+
+test('#setLeadingIconAriaLabel does nothing if the leading icon element is undefined', () => {
```
kfranqueiro: Should these last two tests' foundation calls be wrapped in assert.doesNotThrow?

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -527,3 +553,9 @@ test('keydown is not added to the native select when initialized', () => {
   td.verify(component.foundation_.handleKeydown(td.matchers.anything()), {times: 0});
   document.body.removeChild(fixture);
 });
+
+test('#constructor instantiates a leading icon if an icon element is present', () => {
```
kfranqueiro: One question out of curiosity: coverage is telling me we're covering both the if and else case in `initialize` where it conditionally instantiates leading icon... but are we specifically testing the `else`, given that all the fixtures have an icon?

PR #3748, opened 2018-10-12
feat(tab-bar): Add focusOnActivate flag

### LINKED ISSUE #3585: fix(tabs): Focus should be set optionally
There was a [refactoring](https://github.com/material-components/material-components-web/pull/3164) in July where the active tab would always gain focus once activated.

In my recent project I'm (mis)using tabs as a navigation element in a fixed header to jump between sections on a long onepage website. In addition to this if a user scrolls the page, I highlight the current section once a new section becomes visible in the viewport by using `navigation.activateTab(sectionIndex)`

This is causing an issue, because when you try to navigate the page via the tab-key and you reach an element in the next section, the viewport scrolls, the event-listener will activate the next tab and focus is back in the header. 😢 

Back then @kfranqueiro also raised a question related to this:
> Are we sure we want to do this in all cases? Couldn't this be really confusing if you're focused e.g. on a part of the page unaffected by tabs, then a tab is activated programmatically and grabs the user's focus?

In my case it would be awesome to be able to configure/disable this behaviour when creating the instance or add a second parameter to `.activateTab(sectionIndex, false)` to not set the focus when programmatically activating a tab.

_Related PR: #3164 "feat(tab): Give focus to tab when activated" by @patrickrodee_

kfranqueiro:
Fixes #3585.

This also adds docs and tests I apparently missed for `useAutomaticActivation`, and removes an unused property back from before event registration was moved to the components.

### DIFF
--- a/packages/mdc-tab/foundation.js
+++ b/packages/mdc-tab/foundation.js
@@ -72,8 +72,8 @@ class MDCTabFoundation extends MDCFoundation {
   constructor(adapter) {
     super(Object.assign(MDCTabFoundation.defaultAdapter, adapter));
 
-    /** @private {function(?Event): undefined} */
-    this.handleClick_ = () => this.handleClick();
+    /** @private {boolean} */
+    this.focusOnActivate_ = true;
   }
 
   /**
@@ -94,6 +94,14 @@ class MDCTabFoundation extends MDCFoundation {
   }
 
   /**
+   * Sets whether the tab should focus itself when activated
+   * @param {boolean}
+   */
+  setFocusOnActivate(focusOnActivate) {
+    this.focusOnActivate_ = focusOnActivate;
+  }
+
+  /**
    * Activates the Tab
    * @param {!ClientRect=} previousIndicatorClientRect
    */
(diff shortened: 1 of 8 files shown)


packages/mdc-tab/foundation.js
```diff
@@ -102,7 +110,9 @@ class MDCTabFoundation extends MDCFoundation {
     this.adapter_.setAttr(strings.ARIA_SELECTED, 'true');
     this.adapter_.setAttr(strings.TABINDEX, '0');
     this.adapter_.activateIndicator(previousIndicatorClientRect);
-    this.adapter_.focus();
+    if (this.focusOnActivate_) {
```
patrickrodee: I'd be tempted to make this a parameter for the `activate` function that can be passed by the `TabBar`. Then, `focusOnActivate` won't need to be stored in each `Tab` instance.

patrickrodee: So `TabBar` would pass it's value for `focusOnActivate` to a `Tab` via the `activate` method. I realize this makes more changes to the `Tab` foundation interface. If you're trying to prevent breaking changes, we can roll with the existing solution. Let me know what you think.

kfranqueiro: Your idea is the first idea I had, but I thought against it for a couple of reasons:

* It would make the `activate` API more awkward, given that it already has an optional `ClientRect` parameter
* Semantically, it seems unlikely that you'd want to call this with different values across the same instance's lifetime; you'd probably just want to set the behavior once and roll with it

It wouldn't necessarily be a breaking change that way, since it could also be optional with a default.

patrickrodee: True true, those are all good points.

PR #3749, opened 2018-10-12
fix(dialog): Apply max-width to same element as min-width

### LINKED ISSUE #3717: Unable to properly set dialog min width using mixin
Note: I couldn't include a codepen because this issue was with a mixin
## Bug

**MDC Web Version:** 0.40.0

**Browser:** Firefox Quantum 62.0.3 (64 Bit)

**OS:** Ubuntu 18.04 LTS

### Steps:

1. Add the HTML for a dialog, as shown [here](https://github.com/material-components/material-components-web/blob/master/packages/mdc-dialog/README.md).
2. Use the mixin `@include mdc-dialog-min-width($value)` with `$value` as `75vw` or something
3. Now visit the page in the browser, you should see that the dialog is **not** center-aligned, the left margin is the same. [This](https://imgur.com/gallery/L9uCsEe) is how it looks for me.

The dialog should be centered

The dialog is not centered.

kfranqueiro:
Fixes #3717.

I'm sending this to run screenshot tests and also to let Andy poke at it, since what I'm changing should _theoretically_ resurface an issue in IE 11, but I'm not able to reproduce it.


acdvorak:
This PR does not seem to trigger the bug I was worried about in IE 11: https://github.com/philipwalton/flexbugs/issues/216

However, if you comment out the `display: flex` on `.mdc-dialog__container`, the bug is triggered:

https://github.com/material-components/material-components-web/blob/313618a52e930ecaeb5d3081735d719f30ca4e2b/packages/mdc-dialog/mdc-dialog.scss#L69-L75

image

Which means we probably need both `__container` and `__surface` elements.

kfranqueiro:
Yeah, I concluded that while trying to get rid of the bug in Safari (and one of the things I did was take out the container element, which made IE break). Fortunately I found one ~~cool trick~~ style rule that made Safari behave itself.

packages/mdc-dialog/mdc-dialog.scss
```diff
@@ -69,6 +69,7 @@
 .mdc-dialog__container {
   display: flex;
   flex-direction: row; // IE 11
+  justify-content: space-around;
```
acdvorak: Is this the magic property that fixes something in Safari?

If so, can you add an inline comment explaining why it's there? I'm not sure what issue you're seeing.

test/screenshot/spec/mdc-dialog/issues/3717.html
```diff
@@ -0,0 +1,94 @@
...
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Overflow Dialog Scrolled to Top - MDC Web Screenshot Test</title>
```
acdvorak:
```html
<title>Dialog Issue #3717 - MDC Web Screenshot Test</title>
```

Issue #2239, opened 2018-02-12
Deprecation warnings when compiling with the sass Ruby gem (3.5.5)


kfranqueiro:
We might want to look into resolving these, hopefully before it ends up affecting people.

```
DEPRECATION WARNING on line 24 of .../@material/layout-grid/_mixins.scss:
The result of `0px != 0` will be `true` in future releases of Sass.
Unitless numbers will no longer be equal to the same numbers with units.

DEPRECATION WARNING on line 27 of .../@material/linear-progress/_mixins.scss:
#{} interpolation near operators will be simplified in a future version of Sass.
To preserve the current behavior, use quotes:

  unquote("#{$color-value-for-css}")

DEPRECATION WARNING on line 61 of .../@material/textfield/label/_mixins.scss:
The result of `0% == 0` will be `false` in future releases of Sass.
Unitless numbers will no longer be equal to the same numbers with units.
```

acdvorak:
### Comparison operators

Interestingly, although `0px != 0` and `0% == 0` return `false`, less-than and greater-than operators still work like they did before (as of `dart-sass 1.14.2`):

```
$ echo '@debug (0 == 0px)' | sass --stdin
false

$ echo '@debug (0 < 1px)' | sass --stdin
true

$ echo '@debug (1 > 0px)' | sass --stdin
true
```

### Specifying units

Some comparisons can probably be rewritten to specify units.

E.g., this could be `$min != 0px` (or `$min > 0`):

https://github.com/material-components/material-components-web/blob/92eed058f3e2eb77b6b81d99b506890556efc7a9/packages/mdc-layout-grid/_mixins.scss#L23-L31

https://github.com/material-components/material-components-web/blob/92eed058f3e2eb77b6b81d99b506890556efc7a9/packages/mdc-layout-grid/_variables.scss#L21-L25

### Ignoring units

Other comparisons need to support multiple units:

https://github.com/material-components/material-components-web/blob/92eed058f3e2eb77b6b81d99b506890556efc7a9/packages/mdc-floating-label/_mixins.scss#L57-L69

In these cases, we could strip out the units and compare them to unitless zero:

```scss
/// Remove the unit of a length
/// @param {Number} $number - Number to remove unit from
/// @return {Number} - Unitless number
@function strip-unit($number) {
  @if type-of($number) == 'number' and not unitless($number) {
    @return $number / ($number * 0 + 1);
  }

  @return $number;
}

@debug strip-unit(0px) == 0;
```

[...]

PR #3752, opened 2018-10-13
chore: Switch from node-sass to dart-sass

kfranqueiro:
Note: This is based off of the work for #3751, but I'm submitting it against master to run tests.

This adds dart-sass and fibers (for faster compilation as mentioned in [sass-loader's readme](https://github.com/webpack-contrib/sass-loader#examples)), updates the webpack configuration, and removes node-sass.

With dart-sass, `npm run build` seems to run in around 3/4 the time it did with node-sass (the difference seems even more pronounced when running `npm run build:min`). The CSS output has some whitespace and decimal precision differences, but otherwise looks the same as far as I could tell.

I ran screenshot tests twice to check that the same set of diffs was reported (i.e. no flakes).

TODO: Test on Windows (I'm hoping this works out better than node-sass which doesn't have pre-built versions for all major versions of node). (Update: Works with both node 8 and 10.)


kfranqueiro:
I expected the floating point precision change, because of this in the [docs](https://github.com/sass/dart-sass#javascript-api):

> No support is intended for the following options:
>
> `precision`. Dart Sass defaults to a sufficiently high precision for all existing browsers, and making this customizable would make the code substantially less efficient.

(node-sass's default precision is 5.)

RE floating-label's change, this is a really good catch and I think it has to do with a deprecation notice I remember seeing when testing our code with ruby-sass:

```
The result of `0% == 0` will be `false` in future releases of Sass.
```

I'm pretty sure the culprit is https://github.com/material-components/material-components-web/blob/v0.40.1/packages/mdc-floating-label/_mixins.scss#L59

Looks like I should probably tackle #2239 before we merge this.

PR #3759, opened 2018-10-15
fix(ripple): Deactivate on contextmenu event

### LINKED ISSUE #3524: MDCRipple: Context menu causes ripple to remain activated
Steps to reproduce:
1. Long press on a button when in chrome touch mode. 
2. Click Esc when the menu appears. 
3. Click Tab to focus the next element. 
Notice that the button stays dark while focus has moved to other elements. The button element retains the `:hover` psuedo-class (See [Note](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover)) but it also has the `FG_ACTIVATION` class that causes it be a darker color and more noticeable.

I was able to get the button to return to almost normal (still has the `:hover` state), by removing the activation class and resetting the activate state in the `handleBlur` function.

```
  handleBlur() {
    requestAnimationFrame(() => {
      this.adapter_.removeClass(MDCRippleFoundation.cssClasses.BG_FOCUSED);
      this.adapter_.removeClass(MDCRippleFoundation.cssClasses.FG_ACTIVATION);
      this.resetActivationState_();
    });
  }
```

kfranqueiro:
Fixes #3524.

While testing I wasn't able to get tap-hold to bring up the context menu at all on mobile for some reason... but I confirmed that this fixes the issue for right-click in Chrome and Safari on OS X, which are the only browsers I could reproduce it on.

OS | Browser | Before | After
--- | --- | --- | ---
OS X | Chrome | Buggy | OK
OS X | Safari | Buggy | OK
OS X | Firefox | OK | OK
Win10 | Chrome | OK | OK
Win10 | Firefox | OK | OK
Win10 | Edge | OK | OK

Tested on Win10 with both mouse and touch. IE is N/A because it doesn't use JS for ripple effect.

### DIFF
--- a/packages/mdc-ripple/foundation.js
+++ b/packages/mdc-ripple/foundation.js
@@ -70,7 +70,7 @@ let PointType;
 const ACTIVATION_EVENT_TYPES = ['touchstart', 'pointerdown', 'mousedown', 'keydown'];
 
 // Deactivation events registered on documentElement when a pointer-related down event occurs
-const POINTER_DEACTIVATION_EVENT_TYPES = ['touchend', 'pointerup', 'mouseup'];
+const POINTER_DEACTIVATION_EVENT_TYPES = ['touchend', 'pointerup', 'mouseup', 'contextmenu'];
 
 // Tracks activations that have occurred on the current frame, to avoid simultaneous nested activations
 /** @type {!Array<!EventTarget>} */
--- a/test/unit/mdc-ripple/foundation-deactivation.test.js
+++ b/test/unit/mdc-ripple/foundation-deactivation.test.js
@@ -81,6 +81,29 @@ testFoundation('runs deactivation UX on pointerup after pointerdown', ({foundati
   clock.uninstall();
 });
 
+testFoundation('runs deactivation UX on contextmenu after pointerdown', ({foundation, adapter, mockRaf}) => {
+  const handlers = captureHandlers(adapter, 'registerInteractionHandler');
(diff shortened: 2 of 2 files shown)


williamernest:
I'm not sure this is resolving all of the use cases where the ripple ends up stuck being activted. If you click on a button and then open a context menu on the button it ends up stuck again.

kfranqueiro:
I haven't been able to reproduce what you're describing, even if I purposely try opening the context menu before the first click's ripple activation finishes. The only state that stays for me is the focus state, which is expected.

Can you give full repro steps when you get a chance?

williamernest:
1. Enable device toolbar and select a device (Pixel 2) 
2. Long press on button until context menu shows up. (I'm using `baseline-button-with-icons.html`) 
3. Click `ESC` to close context menu. 
4. Click `Tab` to go to next element. 
The ripple on the button remains in the activated state. 

I was able to reproduce using these steps on OSX/Chrome. On Linux/Chrome substitute the following for step 2. 
1. Click on the button to activate the ripple.
2. Right click on the button to show context menu.

kfranqueiro:
I've managed to reproduce this in Chrome + device emulation on desktop, but I think it's a device emulation quirk. I tested with demos/button.html with an added style to color `:active` elements red, and Chrome on desktop w/ device emulation seems to get the active state stuck when a long-press opens the context menu. Android Chrome and iOS Safari (simulated) don't seem to have any issue.

(GitHub doesn't allow attaching videos, but I've sent a few to Will outside of this thread.)

PR #3761, opened 2018-10-15
fix(shape): Rename surface term with component.

abhiomkar:
BREAKING CHANGE: Renamed shape global variables from `$mdc-shape-*-surface-radius` to `$mdc-shape-*-component-radius`

### DIFF
--- a/packages/mdc-shape/README.md
+++ b/packages/mdc-shape/README.md
@@ -44,7 +44,7 @@ Components are categorized as small, medium and large in shape system. Overridin
 Variable | Description
 --- | ---
 `$mdc-shape-small-component-radius` | Rounded shape radius size for small components. Default value `4px`.
-`$mdc-shape-medium-component-radius` | Rounded shape radius size for medium components. Default value `8px`.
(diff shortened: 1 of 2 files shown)


packages/mdc-shape/README.md
```diff
@@ -43,9 +43,9 @@ Components are categorized as small, medium and large in shape system. Overridin
 
 Variable | Description
 --- | ---
-`$mdc-shape-small-surface-radius` | Rounded shape radius size for small surface components. Default value `4px`.
-`$mdc-shape-medium-surface-radius` | Rounded shape radius size for medium surface components. Default value `8px`.
-`$mdc-shape-large-surface-radius` | Rounded shape radius size for large surface components. Default value `0`.
+`$mdc-shape-small-component-radius` | Rounded shape radius size for small components. Default value `4px`.
+`$mdc-shape-medium-component-radius` | Rounded shape radius size for medium components. Default value `8px`.
```
kfranqueiro: While we're here: this says 8px is the default, but it's 4px in the code.

PR #3921, opened 2018-10-16
fix: Future-proof Sass usage

### LINKED ISSUE #2239: Deprecation warnings when compiling with the sass Ruby gem (3.5.5)

kfranqueiro:
Fixes #2239 and avoids unwanted diffs in #3752.

### DIFF
--- a/packages/mdc-floating-label/_mixins.scss
+++ b/packages/mdc-floating-label/_mixins.scss
@@ -56,14 +56,14 @@
 
 @mixin mdc-floating-label-float-position($positionY, $positionX: 0%, $scale: .75) {
   .mdc-floating-label--float-above {
-    @if $positionX == 0% or $positionX == 0 {
-      transform: translateY(-1 * $positionY) scale($scale);
-    } @else {
+    @if $positionX > 0 {
       transform: translateY(-1 * $positionY) translateX(-1 * $positionX) scale($scale);
 
       @include mdc-rtl {
         transform: translateY(-1 * $positionY) translateX($positionX) scale($scale);
       }
+    } @else {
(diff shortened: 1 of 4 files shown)


packages/mdc-floating-label/_mixins.scss
```diff
@@ -56,14 +56,14 @@
 
 @mixin mdc-floating-label-float-position($positionY, $positionX: 0%, $scale: .75) {
   .mdc-floating-label--float-above {
-    @if $positionX == 0 {
-      transform: translateY(-1 * $positionY) scale($scale);
-    } @else {
+    @if $positionX > 0 {
```
acdvorak: Is there a valid use case for clients to pass a negative value here?

kfranqueiro: There isn't presently, but I could see this becoming a thing depending on how we fix floating label in heavily-rounded outlined text fields.

PR #3923, opened 2018-10-16
feat(select): Add helper-text sub-component

williamernest:
refs: #3706


packages/mdc-select/mdc-select.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2017 Google Inc.
+// Copyright 2018 Google Inc.
```
kfranqueiro: Revert copyright year change here too

test/screenshot/spec/mdc-select/classes/enhanced-helper-text-persistent.html
```diff
@@ -0,0 +1,172 @@
...
+  </head>
+
+  <body class="test-container">
+    <main class="test-viewport test-viewport--mobile">
+      <div class="test-layout">
+
+        <div class="test-cell test-cell--select">
+          <div class="mdc-select custom-enhanced-select-width">
+            <i class="mdc-select__dropdown-icon"></i>
+            <div class="mdc-select__selected-text" role="combobox" aria-controls="my-helper-text"></div>
```
kfranqueiro: These enhanced pages will eventually need to be updated to reflect the updated ARIA stuff in the other PR

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -110,32 +115,42 @@ function getFixture() {
...
-    </div>
   `;
 }
 
+function getHelperTextFixture() {
+  const containerDiv = document.createElement('div');
+  const root = getFixture();
+  root.querySelector('.mdc-select__selected-text').setAttribute('aria-controls', 'test-helper-text');
+  containerDiv.appendChild(root);
+  containerDiv.innerHTML = containerDiv.innerHTML
```
kfranqueiro: This seems like asking for trouble... can we do this instead?

```js
containerDiv.appendChild(bel`<p class="mdc-select-helper-text" id="test-helper-text">Hello World</p>`);
```

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -967,3 +991,37 @@ test('#constructor instantiates a leading icon if an icon element is present', (
   assert.isTrue(root.classList.contains(cssClasses.WITH_LEADING_ICON));
   document.body.removeChild(menu);
 });
+
+test('#constructor instantiates the helper text if present', () => {
```
kfranqueiro: Is there a reason these don't use `setupTest` while `setupTest` was seemingly updated to support helper text?

In particular would that make it possible to just reference `helperText` directly from `setupTest`'s return value in the destroy test?

(Ditto for mdc-select.test.js)

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -992,36 +995,41 @@ test('#constructor instantiates a leading icon if an icon element is present', (
...
+  const hasMockFoundation = true;
+  const hasMockMenu = true;
+  const hasLabel = true;
+  const hasOutline = false;
+  const hasHelperText = true;
+  const {container, helperText} = setupTest(hasLabel, hasOutline, hasMockFoundation, hasMockMenu,
+    hasHelperText);
 
-  assert.instanceOf(component.helperText_, MDCSelectHelperText);
+  assert.instanceOf(helperText, FakeHelperText);
```
kfranqueiro: This line (as well as in mdc-select.test.js) will still need to check the component property - otherwise this is effectively just testing `setupTest`.

Issue #3741, opened 2018-10-11
Side Sheet: Migrate JS from mdc-drawer to mdc-side-sheet and make reusable


kfranqueiro:
* [x] Copy JS files from `mdc-drawer` to `mdc-sidesheet` and rename classes
* [x] Update constants to use `mdc-side-sheet` as the base name of classes
* [x] Remove list logic from `MDCSideSheet` class
* [ ] Create common helper for returning adapter implementation, to be used by `MDCSideSheet` and `MDCDrawer`
  * This would at least need a parameter to distinguish the prefix for the opened and closed events for each component
  * This probably makes sense to be a separate util function; it seems unlikely to be able to easily extend one component class from the other without overriding a bunch of code in the parent anyway

abhiomkar:
Closing this issue as obsolete.

Issue #2665, opened 2018-04-26
Add new `mdc-side-sheet` component


sfdexter:
https://material.io/go/design-sheets-side

vgangisetti:
Hi Bonnie,
I would love to work on this component as I have a similar requirement in my current project.

zq-97:
Is there any progress for this issue? branch `feat/side-sheet` hadn't been update for months.

bdgit:
Are there any updates on this forgotten/stale feature which is listed in the documentation?
[Sheets: Side](https://material.io/components/sheets-side/web)

Is there a roadmap?

PR #3924, opened 2018-10-16
feat(side-sheet): Create side-sheet package

### LINKED ISSUE #3739: Side Sheet: Migrate Sass from mdc-drawer to mdc-side-sheet and make reusable
* [x] Copy content over from mdc-drawer to mdc-side-sheet package, updating for `mdc-side-sheet` base class
* [x] Exclude mixins that only make sense for drawer (i.e. side sheet will only have border-color, shape-radius, width, surface-fill-color, ink-color, fill-color-accessible, scrim-fill-color, z-index)
* [x] Update width and shape mixins to take an optional side parameter
  * Side parameter should be phrased as `leading` and `trailing` to indicate being RTL agnostic
* [x] Remove parts that are drawer-specific (e.g. list styles)
* [x] Expose all remaining styles via mixins (parameterized by side where necessary, defaulting to trailing side), then invoke them within `mdc-side-sheet.scss`

Drawer will be updated to reuse these new mixins in a separate task.

### LINKED ISSUE #3740: Side Sheet: Create screenshot test pages for dismissible and modal side sheets
These can initially be tested without JS.

It will probably make sense to include this in the same PR as #3739 in order to develop and demonstrate it.

acdvorak:
I copied the `mdc-drawer` package, renamed it, and deleted all drawer-specific code.

Applicable mixins have a `$side` param that defaults to `trailing`, and can be overridden to `leading` for drawer.

#### Related issues:

* Refs #2665
* Refs #3741
* Resolves #3739
* Resolves #3740

#### TODO:

* [x] Verify README metadata is correct
* [x] Determine if we need to set `padding: 16px` on content
    - NO. Drawer doesn't set padding on content, so neither should side sheet.
* [x] Determine if we need to call `mdc-typography-base()` mixin for content
    - YES. Call it within the top-level `mdc-side-sheet` class.

#### Follow up PRs:

* Create common helper for returning adapter implementation, to be used by MDCSideSheet and MDCDrawer (#3741)
* Dividers (mdc-list)


packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+section: components
+excerpt: "Permanent, dismissible, and modal side sheets."
+iconId: side_navigation
+path: /catalog/side-sheets/
+-->
 
-The [Side Sheet component](https://material.io/go/design-sheets-side) is yet to be completed, please follow the [tracking issue](https://github.com/material-components/material-components-web/issues/2665) for more information.
+# Side Sheets
 
+The MDC Side Sheet is used to organize access to destinations and other functionality on an app.
```
kfranqueiro: Based on material.io wording:

The MDC Side Sheet is a supplementary surface primarily used on tablet and desktop.

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+```
+npm install @material/side-sheet
+```
+
+### HTML Structure
+
+```html
+<aside class="mdc-side-sheet">
+  <div class="mdc-side-sheet__content">
+    <div><label><input type="checkbox"> Events</label></div>
```
kfranqueiro: Leave the inner contents out, just put ..., since we have no opinion on the internals (but what's here now isn't MD-ified at all).

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+    <div><label><input type="checkbox" checked> Personal</label></div>
+    <div><label><input type="checkbox" checked> Projects</label></div>
+    <div><label><input type="checkbox"> Reminders</label></div>
+    <div><label><input type="checkbox"> Work</label></div>
+  </div>
+</aside>
+```
+
+#### Menu Icon
```
kfranqueiro: Genericize this to just "Icon" and add a leading sentence such as the following:

Side sheets are typically dismissible or modal, and are opened via an affordance such as an icon in a Top App Bar.

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+```js
+import {MDCSideSheet} from "@material/side-sheet";
+const sideSheet = MDCSideSheet.attachTo(document.querySelector('.mdc-side-sheet'));
+```
+
+## Variants
+
+### Dismissible Side Sheet
+
+Dismissible side sheets are by default hidden off screen, and can slide into view. Dismissible side sheets should be used when navigation is not common, and the main app content is prioritized.
```
kfranqueiro: Remove the second sentence, which is drawer-specific

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+
+```html
+<body>
+  <aside class="mdc-side-sheet mdc-side-sheet--dismissible">
+    <div class="mdc-side-sheet__content">
+      <div><label><input type="checkbox"> Events</label></div>
```
kfranqueiro: Replace internals with ...

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+
+```
+npm install @material/side-sheet
+```
+
+### HTML Structure
+
+```html
+<aside class="mdc-side-sheet">
+  <div class="mdc-side-sheet__content">
```
kfranqueiro: Is `__content` here for independent scrolling? Do we need it? Are we keeping it for future-proofing if there is need for fixed sections above/below?

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+    </div>
+  </aside>
+
+  <div class="mdc-side-sheet-app-content">
+    App Content
+  </div>
+</body>
+```
+
+> Apply the `mdc-side-sheet-app-content` class to the sibling element after the side-sheet for the open/close animations to work.\
```
kfranqueiro: "the sibling element" -> "a subsequent sibling element" here, and change style to use `~` rather than `+` to make this more agnostic to drawer + side sheet

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+In cases where the side-sheet occupies the full viewport height, some styles must be applied to get the dismissible side-sheet and the content below the top app bar to independently scroll and work in all browsers.
+
+In the following example, the `mdc-side-sheet__content` and `main-content` elements should scroll independently of each other. The `mdc-side-sheet--dismissible` and `mdc-side-sheet-app-content` elements should then sit side-by-side. The markup looks something like this:
+
+```html
+<body>
+  <aside class="mdc-side-sheet mdc-side-sheet--dismissible">
+    <div class="mdc-side-sheet__content">
+      <div class="mdc-side-sheet__content">
+        <div><label><input type="checkbox"> Events</label></div>
```
kfranqueiro: "..." here too

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+        <div><label><input type="checkbox"> Reminders</label></div>
+        <div><label><input type="checkbox"> Work</label></div>
+      </div>
+    </div>
+  </aside>
+
+  <div class="mdc-side-sheet-app-content">
+    <header class="mdc-top-app-bar app-bar" id="app-bar">
+      <div class="mdc-top-app-bar__row">
+        <section class="mdc-top-app-bar__section mdc-top-app-bar__section--align-start">
```
kfranqueiro: This should be updated to use an icon at the end of the top app bar consistent with the side sheet design article. e.g.:

```html
    <section class="mdc-top-app-bar__section mdc-top-app-bar__section--align-end" role="toolbar">
      <a href="#" class="material-icons mdc-top-app-bar__action-item" aria-label="Filters" alt="Filters">filter_list</a>
    </section>
```

Could we also update this in the screenshot test pages? (That could be in a separate task if you really want.)

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+  overflow: auto;
+  height: 100%;
+}
+
+.app-bar {
+  position: absolute;
+}
+```
+
+The JavaScript to toggle the side-sheet when the navigation button is clicked looks like this:
```
kfranqueiro: navigation -> filter

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+
+```js
+import {MDCTopAppBar} from "@material/top-app-bar";
+const topAppBar = MDCTopAppBar.attachTo(document.getElementById('app-bar'));
+topAppBar.setScrollTarget(document.getElementById('main-content'));
+topAppBar.listen('MDCTopAppBar:nav', () => {
```
kfranqueiro: This would need to change to a click handler on the specific icon, since only the top-left icon has a dedicated Top App Bar event.

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+
+### Modal Side Sheet
+
+Modal side sheets are elevated above most of the app's UI and don't affect the screen's layout grid.
+
+```html
+<body>
+  <aside class="mdc-side-sheet mdc-side-sheet--modal">
+    <div class="mdc-side-sheet__content">
+      <div><label><input type="checkbox"> Events</label></div>
```
kfranqueiro: contents -> ...

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+const topAppBar = MDCTopAppBar.attachTo(document.getElementById('app-bar'));
+topAppBar.setScrollTarget(document.getElementById('main-content'));
+topAppBar.listen('MDCTopAppBar:nav', () => {
+  sideSheet.open = !sideSheet.open;
+});
+```
+
+### Modal Side Sheet
+
+Modal side sheets are elevated above most of the app's UI and don't affect the screen's layout grid.
```
kfranqueiro: most of -> the rest of?

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+`mdc-side-sheet-stroke-color($color)` | Sets border color of `mdc-side-sheet` surface.
+`mdc-side-sheet-shape-radius($radius)` | Sets the rounded shape to side-sheet with given radius size. `$radius` can be single radius or list of 2 radius values for trailing-top and trailing-bottom. Automatically flips the radius values in RTL context.
+`mdc-side-sheet-z-index($value)` | Sets the z index of side-sheet. Side Sheet stays on top of top app bar except for clipped variant of side-sheet.
+`mdc-side-sheet-width($width)` | Sets the width of the side-sheet. In the case of the dismissible variant, also sets margin required for `mdc-side-sheet-app-content`.
+
+## Accessibility
+
+### Focus Management
+
+It is recommended to shift focus to the first focusable element in the main content when side-sheet is closed or one of the destination items is activated. (By default, MDC Side Sheet restores focus to the menu button which opened it.)
```
kfranqueiro: "menu button" -> "button"?

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+
+#### Dismissible Side Sheet
+
+Restore focus to the first focusable element when a list item is activated or after the side-sheet closes. Do not close the side-sheet upon item activation, since it should be up to the user when to show/hide the dismissible side-sheet.
```
kfranqueiro: This is left over from drawer and needs to be updated or removed. Check if there's particular guidance RE whether anything aside from an optional close affordance should close the side sheet (and I'm not sure a close affordance is required since you can toggle it via the same icon you opened it with?)

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+});
+
+document.body.addEventListener('MDCSideSheet:closed', () => {
+  mainContentEl.querySelector('input, button').focus();
+});
+```
+
+#### Modal Side Sheet
+
+Close the side-sheet when an item is activated in order to dismiss the modal as soon as the user performs an action. Only restore focus to the first focusable element in the main content after the side-sheet is closed, since it's being closed automatically.
```
kfranqueiro: Ditto previous comment RE drawer-specific leftovers

packages/mdc-side-sheet/README.md
```diff
@@ -1,4 +1,318 @@
...
+### Events
+
+Event Name | Event Data Structure | Description
+--- | --- | ---
+`MDCSideSheet:opened` | None | Emits when the navigation side-sheet has opened.
+`MDCSideSheet:closed` | None | Emits when the navigation side-sheet has closed.
+
+## Usage within Web Frameworks
+
+If you are using a JavaScript framework, such as React or Angular, you can create a Navigation Side Sheet for your framework. Depending on your needs, you can use the _Simple Approach: Wrapping MDC Web Vanilla Components_, or the _Advanced Approach: Using Foundations and Adapters_. Please follow the instructions [here](../../docs/integrating-into-frameworks.md).
```
kfranqueiro: Remove "Navigation"

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -0,0 +1,216 @@
...
+// Public
+//
+
+@mixin mdc-side-sheet-stroke-color($color, $opacity: $mdc-side-sheet-stroke-opacity) {
+  $value: rgba(mdc-theme-prop-value($color), $opacity);
+
+  @include mdc-theme-prop(border-color, $value);
+}
+
+@mixin mdc-side-sheet-shape-radius($radius, $side: trailing) {
```
kfranqueiro: Nit: We should be consistent on whether `leading` and `trailing` are quoted or unquoted. Currently we have the default unquoted but everything else seems to be quoted. I think we do unquoted elsewhere in MDC but I'd have to dig for examples.

(Fortunately Sass treats the two as equal, though I wonder if that would ever change...)

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -0,0 +1,216 @@
...
+
+  ~ .mdc-side-sheet-scrim {
+    @include mdc-theme-prop(background-color, $value);
+  }
+}
+
+@mixin mdc-side-sheet-ink-color($color, $opacity: $mdc-side-sheet-ink-opacity) {
+  $value: rgba(mdc-theme-prop-value($color), $opacity);
+
+  .mdc-side-sheet__title {
```
kfranqueiro: This selector shouldn't be here, right? (This is probably why I thought the accessible fill color test page looked odd, and we don't seem to have a test page for ink color?)

packages/mdc-side-sheet/README.md
acdvorak: I thought I already replaced the `+` (adjacent sibling selector) with `~` (general sibling selector), but it looks like I missed a few in `mdc-side-sheet-modal.scss`.

Fixed!

acdvorak:
Now I remember why I originally kept `__content`: So that drawer and side-sheet can both use the same base styles.

* `.mdc-drawer` and `.mdc-side-sheet` are explicitly set to **`display: flex`**
* `.mdc-drawer__content` and `.mdc-side-sheet__content` are implicitly **`display: block`**

As a result, the same content will be laid out differently depending on whether it's inside a
`.mdc-drawer__content` or `.mdc-side-sheet` element. That seems weird.

We could potentially make `.mdc-side-sheet` `display: block` and then override it to
`display: flex` in `.mdc-drawer`, but that sounds messy and confusing.

It seems like the simplest option is to keep the `.mdc-side-sheet__content` element. WDYT?

packages/mdc-side-sheet/_mixins.scss
acdvorak: Removed double quotes everywhere

packages/mdc-side-sheet/README.md
acdvorak: Done

acdvorak:
### Changes:

* Added screenshot test pages for all mixins
* Updated all pages to show scrolling behavior when side sheet content overflows its surface
* Added a "Close" button to all side sheets (except "permanent") to demonstrate focus trap behavior
* Added support for focus trapping on initially-open side sheets (those with `--open` in the markup)

kfranqueiro:
I'm fine with leaving `__content` in if it makes it more consistent. There's a chance we (or users) will have a use for it being a dedicated element anyway.

kfranqueiro:
I would revert the focus trapping on initially-open side sheets commit.

* This is only useful for modal and I don't think we should be going out of our way to support initially-open modals...
* This ends up emitting an "opened" event even when the side sheet was already opened
* This is adding features beyond the scope of this task that would need to be reconciled with Drawer

acdvorak:
#### Changes:

* Added `__content` element back
* Reverted focus trapping on initially-opened side sheets

acdvorak:
#### Changes:

* Added `mdc-typography-base()` mixin to root `.mdc-side-sheet` CSS class
* Added `autofocus` attribute to all "close" buttons to reduce screenshot test flakiness
* Updated `golden.json`

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -0,0 +1,214 @@
...
+  &--closing {
+    transform: translateX(-100% * $xFactor);
+
+    @include mdc-rtl {
+      transform: translateX(100% * $xFactor);
+    }
+  }
+}
+
+@mixin mdc-side-sheet-scrim_($side: trailing) {
```
kfranqueiro: I don't think the branching in this mixin is necessary since the scrim is 100% width.

Meanwhile, why aren't the rest of the scrim styles in here? They should all be shareable/common with drawer, right? The point of these is to avoid needing to repeat the same styles between mdc-side-sheet and mdc-drawer codebases. (I think this question also applies to other mixins here)

packages/mdc-side-sheet/_variables.scss
```diff
@@ -0,0 +1,47 @@
...
+$mdc-side-sheet-ink-color: mdc-theme-prop-value(on-surface);
+$mdc-side-sheet-stroke-color: mdc-theme-prop-value(on-surface);
+
+// Opacity
+$mdc-side-sheet-ink-opacity: mdc-theme-text-emphasis(high);
+$mdc-side-sheet-stroke-opacity: .12;
+
+// Widths
+$mdc-side-sheet-width: 256px;
+$mdc-side-sheet-surface-padding: 16px;
```
kfranqueiro: Remove, unused

packages/mdc-side-sheet/common.scss
```diff
@@ -0,0 +1,73 @@
...
+  @include mdc-side-sheet-base_;
+  @include mdc-side-sheet-stroke-width_(1px);
+  @include mdc-side-sheet-stroke-color($mdc-side-sheet-stroke-color);
+  @include mdc-side-sheet-surface-fill-color($mdc-side-sheet-surface-fill-color);
+  @include mdc-side-sheet-shape-radius(large);
+  @include mdc-side-sheet-z-index($mdc-side-sheet-z-index);
+  @include mdc-side-sheet-width($mdc-side-sheet-width);
+  @include mdc-typography-base;
+
+  display: flex;
```
kfranqueiro: Related to previous comment, as much of this as possible should be within the `base_` mixin (including the open/opening/closing/content stuff) if it will be the same for Drawer, so that Drawer can just include these mixins and have a minimum of additional styles only pertinent to Drawer.

Same goes for the dismissible and modal files/mixins

packages/mdc-side-sheet/constants.js
```diff
@@ -0,0 +1,43 @@
...
+  MODAL: 'mdc-side-sheet--modal',
+  OPEN: 'mdc-side-sheet--open',
+  ANIMATE: 'mdc-side-sheet--animate',
+  OPENING: 'mdc-side-sheet--opening',
+  CLOSING: 'mdc-side-sheet--closing',
+};
+
+/** @enum {string} */
+const strings = {
+  APP_CONTENT_SELECTOR: '.mdc-side-sheet-app-content',
```
kfranqueiro: This appears to be currently unused (and was also unused in drawer).

packages/mdc-side-sheet/dismissible/foundation.js
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Keydown handler to close side-sheet when key is escape.
+   * @param evt
+   */
+  handleKeydown(evt) {
+    const {keyCode, key} = evt;
+
+    const isEscape = key === 'Escape' || keyCode === 27;
+    if (isEscape) {
+      this.close();
```
kfranqueiro: I think we'll need to parameterize this, because drawers should do this by default but side sheets should not (they may contain child controls that do something on escape).

We can enter a separate task for this if you would rather do it that way.

packages/mdc-side-sheet/dismissible/foundation.js
```diff
@@ -0,0 +1,200 @@
...
+      return;
+    }
+
+    if (this.isClosing()) {
+      this.adapter_.removeClass(OPEN);
+      this.adapter_.restoreFocus();
+      this.closed();
+      this.adapter_.notifyClose();
+    } else {
+      this.adapter_.focusActiveNavigationItem();
```
kfranqueiro: This concept doesn't exist in side sheet and should be removed from its adapter entirely. I guess our drawer adapter will be a superset of side sheet's...

packages/mdc-side-sheet/index.js
```diff
@@ -0,0 +1,168 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc.
```
kfranqueiro: 2018

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -0,0 +1,214 @@
+//
+// Copyright 2017 Google Inc.
```
kfranqueiro: 2018

packages/mdc-side-sheet/_variables.scss
```diff
@@ -0,0 +1,47 @@
+//
+// Copyright 2017 Google Inc.
```
kfranqueiro: 2018

packages/mdc-side-sheet/constants.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc.
```
kfranqueiro: 2018

packages/mdc-side-sheet/mdc-side-sheet.scss
```diff
@@ -0,0 +1,25 @@
+//
+// Copyright 2017 Google Inc.
```
kfranqueiro: 2018

packages/mdc-side-sheet/util.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc.
```
kfranqueiro: 2018

packages/mdc-side-sheet/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@material/side-sheet",
+  "version": "0.0.0",
+  "description": "The Material Components Web side-sheet component",
```
kfranqueiro: Remove hyphen

packages/mdc-side-sheet/util.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @param {!Element} surfaceEl
+ * @param {!Function} focusTrapFactory
+ * @return {!FocusTrapInstance}
+ */
+function createFocusTrapInstance(surfaceEl, focusTrapFactory = createFocusTrap) {
+  return focusTrapFactory(surfaceEl, {
+    clickOutsideDeactivates: true,
+    initialFocus: false, // Navigation side-sheet handles focusing on active nav item.
```
kfranqueiro: Remove "Navigation" from these comments

test/screenshot/spec/mdc-side-sheet/fixture.scss
```diff
@@ -0,0 +1,98 @@
...
+.test-side-sheet-main-content {
+  display: block; // Needed for IE 11
+  box-sizing: border-box;
+  height: 100%;
+  padding-right: 16px;
+  padding-left: 16px;
+  overflow: auto;
+}
+
+.test-side-sheet > :first-child,
```
kfranqueiro: Are these to avoid UA differences?

acdvorak: No, they're to avoid giving the false impression that `__content` has padding.
I added a comment here explaining that.

The default margins on `<p>` elements could easily be mistaken for container padding, and I want to make it obvious that `__content` does not have any padding.

Another option would be to remove the `<p>` elements and use `<br>` tags for spacing.

packages/mdc-side-sheet/dismissible/foundation.js
acdvorak: Created #3938 for this 🙂

packages/mdc-side-sheet/dismissible/foundation.js
acdvorak: Doesn't modal side sheet still need to trap focus?

How should we generically support that?

kfranqueiro: I'm not talking about the trap/restore focus APIs. I'm talking about the "focus active navigation item" API, which is drawer-specific, and typically won't actually do anything for side sheet (unless it just happens to have a MDC List with an activated item.

We probably do still want to focus something within the dismissible side sheet when it opens...but we'd probably have to just look for the first focusable thing within it. I forget if we've already implemented or used something like that somewhere else in our repo? Otherwise we could probably use `tabbable` (focus-trap's dependency) for it, but might want to delegate that to a `util` function as well.

We can split this to another task if it makes it clearer to leave this as it was from drawer in this first PR.

packages/mdc-side-sheet/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "main": "dist/mdc.side-sheet.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/material-components/material-components-web.git"
+  },
+  "dependencies": {
+    "@material/animation": "^0.40.1",
+    "@material/base": "^0.40.1",
+    "@material/elevation": "^0.40.1",
+    "@material/list": "^0.40.1",
```
kfranqueiro: I don't think side sheet needs list or ripple?

packages/mdc-side-sheet/dismissible/foundation.js
acdvorak: Created #3940 for this 🙂

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -110,90 +113,127 @@
...
+  flex-direction: column;
+  flex-shrink: 0;
+  order: $flexOrder;
+  box-sizing: border-box;
+  height: 100%;
+  transition-property: transform;
+  transition-timing-function: $mdc-animation-standard-curve-timing-function;
+  overflow: hidden;
+
+  #{&}--open {
```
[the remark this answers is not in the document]
acdvorak: I originally had `&`, but stylelint threw a fit:

```
Expected "&--open" to have no more than 0 type selectors
```

Strangely, it only complains about `&--open`, but not any of the other similar selectors
(`&--animate`, `&--opening`, or `&--closing`). It must be a bug in stylelint.

I reverted back to `&` and added a `stylelint-disable-next-line selector-max-type` comment to suppress the linter error.

Thanks!

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -110,90 +113,127 @@
...
+  @include mdc-typography-base;
+
+  $flexOrder: if($side == leading, 0, 1);
   $xFactor: if($side == leading, 1, -1);
 
-  order: if($side == leading, 0, 1);
+  display: flex;
+  flex-direction: column;
+  flex-shrink: 0;
+  order: $flexOrder;
```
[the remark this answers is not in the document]
acdvorak: Yes, that's correct. I added a comment in the Sass to clarify this.

I manually tested drawer + app content + side sheet, and `1` seemed to work fine. According to the Chrome Dev Tools, the default `order` for flex items is `0`, so `1` will always "win".

Should I create an issue to track finding a better solution that doesn't require a specific source order?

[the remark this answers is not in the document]
kfranqueiro: Yeah, I haven't gotten around to adding one but we need to, in terms of finding something that doesn't present problems for tab order as well as flex order.

packages/mdc-side-sheet/_mixins.scss
```diff
@@ -110,90 +113,127 @@
...
+    transition-duration: $mdc-side-sheet-animation-exit;
 
     @include mdc-rtl {
       transform: translateX(100% * $xFactor);
     }
   }
 }
 
-@mixin mdc-side-sheet-modal_($side: trailing) {
+@mixin mdc-side-sheet-dismissible_($rootClass, $side: trailing) {
```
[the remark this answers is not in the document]
acdvorak: It's possible, yes, but it feels a little weird to me: 4d73a7a3e9834caedc85579e36b9f26699185f52

Specifically, `mdc-side-sheet.scss` ends up looking like this:

```scss
.mdc-side-sheet {
  @include mdc-side-sheet-base_;
  @include mdc-side-sheet-dismissible_;
  @include mdc-side-sheet-modal_;
  @include mdc-side-sheet-scrim_;
  @include mdc-side-sheet-stroke-width_(1px);
  @include mdc-side-sheet-stroke-color($mdc-side-sheet-stroke-color);
  @include mdc-side-sheet-surface-fill-color($mdc-side-sheet-surface-fill-color);
  // ...
}

.mdc-side-sheet--modal {
  @include mdc-side-sheet-scrim-fill-color($mdc-side-sheet-modal-scrim-color);
  @include mdc-elevation($mdc-side-sheet-modal-elevation);
}
```

1. The root CSS class calls mixins for itself _and_ the two variants _and_ the scrim, which feels odd.

2. The root CSS class calls a `modal_` mixin, but right below it, there's _also_ a `--modal` CSS class. I could move those two mixin calls into `mdc-side-sheet-modal_`, but we normally try to keep all the theme configuration in the main `mdc-foo.scss` file.

3. It forces most of the CSS classes to go in mixins instead of the main `mdc-side-sheet.scss` file, which IMO makes the styles harder to read and understand.
    - On a related note, `&--open` selectors make search/replace harder, and it's easy to forget about them when you're renaming something.

The `$rootClass` param isn't great, but it seems cleaner than the alternative. And the mixins that have it are all "private", so we don't need to worry about clients using it.

WDYT?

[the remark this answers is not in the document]
kfranqueiro: My original concern was mostly just the required repetition of the base class which we could easily mess up in copypasta, but I guess that's a pretty small price to pay which should also be obvious in screenshot tests. So I'm willing to go with that since this didn't pan out.

packages/mdc-side-sheet/_mixins.scss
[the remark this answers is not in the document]
acdvorak: Created #3944 for this 🙂

PR #3925, opened 2018-10-16
feat(select): Update aria handling

williamernest:
The enhanced select was initially designed to use the `role="combobox"` based on feedback from another team. After going through and verifying that our component will meet the `combobox` specification, I discovered that it does not meet the definition of a `combobox` because it does not contain a text field inside the `role="combobox"` element. 

I found another suitable example on the aria site for the enhanced select component, the [collapsible dropdown listbox](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html). This example uses a `role="button"`, `role="listbox"` and `role="option"`. Our component will differ from the aria example above in 2 key ways.

First, the enhanced select will not set `tabindex="-1"` on the `ul` element, because the `ul` element will not receive focus. Design guidance has indicated that the menu should focus on the previously selected element or the first element of the list. 

Second, the enhanced select will not use the `aria-activedescendant` attribute on the `ul` or the `mdc-menu` element. According to the example, this attribute is used to indicate which elements appear focused with actual DOM focus remaining on the `ul` element. Since the `mdc-menu` component allows the DOM focus to move when using the arrow keys in the list, this attribute was redundant.


packages/mdc-select/README.md
```diff
@@ -291,6 +291,35 @@ programmatically select a disabled list item in the enhanced select.
...
+#### Enhanced Select Accessibility (a11y)
+
+In order to have an accessible component for users, it's recommended that you follow the WAI-ARIA example for
+[Collapsible Dropdown Listbox](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html). An
+example of the enhanced select component with all the necessary aria attributes is listed below.
+
+```html
+<div class="mdc-select">
+  <i class="mdc-select__dropdown-icon"></i>
+  <div id="demo-selected-text" class="mdc-select__selected-text" role="button" aria-haspopup="listbox" aria-labelledby="demo-label demo-selected-text">Vegetables</div>
```
kfranqueiro: Does this need `aria-controls`?

> If the element with the aria-expanded attribute controls the expansion of another grouping container that is not 'owned by' the element, the author SHOULD reference the container by using the aria-controls attribute.

packages/mdc-select/index.js
```diff
@@ -264,6 +264,7 @@ class MDCSelect extends MDCComponent {
       // will return false if the menu is still closing and this method listens to the closed event which
       // occurs after the menu is already closed.
       this.menuOpened_ = false;
+      this.selectedText_.removeAttribute('aria-expanded');
```
kfranqueiro: I think we want to be setting this to false, not removing it.

> **true/false/undefined**
> Value representing true, false, or not applicable. For example, an element with aria-expanded set to false is not currently expanded; an element with aria-expanded set to undefined is not expandable. The default value for this value type is undefined unless otherwise specified.

packages/mdc-select/index.js
```diff
@@ -367,7 +368,10 @@ class MDCSelect extends MDCComponent {
       setSelectedIndex: (index) => {
         this.nativeControl_.selectedIndex = index;
       },
-      setDisabled: (isDisabled) => this.nativeControl_.disabled = isDisabled,
+      setDisabled: (isDisabled) => {
+        this.nativeControl_.disabled = isDisabled;
+        this.nativeControl_.setAttribute('aria-disabled', isDisabled.toString());
```
kfranqueiro: Isn't this redundant for native select? Shouldn't `disabled` suffice?

packages/mdc-select/README.md
williamernest: I'm using `aria-controls` for the helper text and the way it was written for text field it won't support us using `aria-controls` for anything other than the helper text. Maybe switch the helper text logic to use `data-has-helper-text` to instantiate the MDCSelectHelperText and add multiple `aria-controls`.

The example I worked from on the WAI-ARIA site didn't use `aria-controls` either so I wasn't sure it was required.

kfranqueiro: Huh, OK. Well, it's a "SHOULD" so that's more like strongly encouraged than outright required.

packages/mdc-select/index.js
williamernest: The example I linked above indicates that `aria-expanded` is not present when the popup is closed. Also, from the `menu` spec, which is closely associated with the listbox example we're implementing: 

> When the menu is displayed, the element with role button has aria-expanded set to true. When the menu is hidden, it is recommended that aria-expanded is not present. If aria-expanded is specified when the menu is hidden, it is set to false.

From what I gather `aria-expanded` should be used to indicate a control is in an expanded (or displayed) state but when focus has not moved from the original element. When focus moves into the popup, the expanded attribute should not be used (because it's ignored by screen readers since focus has moved). It is only used when the menu is open to accommodate for touch devices where screen reader users can touch the menu button when the menu is displayed. Keyboard users cannot focus the menu when the menu is open. 

https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html
https://www.w3.org/TR/wai-aria-practices/#menubutton

test/screenshot/spec/mdc-select/classes/enhanced-baseline.html
```diff
@@ -49,45 +49,46 @@
...
+            <span id="screenshot-select-label" class="mdc-floating-label">Pick a Food Group</span>
             <div class="mdc-line-ripple"></div>
           </div>
         </div>
 
         <div class="test-cell test-cell--select">
           <div class="mdc-select mdc-select--outlined custom-enhanced-select-width">
-            <i class="mdc-select__dropdown-icon"></i>
-            <div class="mdc-select__selected-text" role="combobox"></div>
+
```
kfranqueiro: This lost its dropdown icon?

PR #3929, opened 2018-10-17
fix(notched-outline): Auto position the notch and floating label based on corner size

### LINKED ISSUE #3158: Notched outline overlaps with text field label

v0.37.1

All browsers

### What OS are you using?

Mac

1. Override `$mdc-text-field-border-radius` variable to 28px.

    `$mdc-text-field-border-radius: 28px;`

2. Focus on any text field outlined example in textfield demo page.

The notched outline and the text field label should not be overlapped.

Notched outline and text field labels are overlapped.

screen shot 2018-07-20 at 3 28 39 pm

abhiomkar:
This change fixes the issue where notched outline overlaps with floating label on text field & select components.

Notch position is calculated based on the corner size of notched outline and accordingly the shape mixins in text field and select components sets the floating label position based on radius size.

The leading stroke length is calculated with: `Math.max(0, numbers.MIN_LEADING_STROKE_EDGE_POSITION - radius - 1.2)` which gets shrunk when large radius is used.

The floating label position calculation with: `$cornerSize + $leadingStrokeLength + $mdc-notched-outline-notch-gutter-size + 1.2` where `$leadingStrokeLength` is the stroke length that applies the same formula as above.

Fixes #3158, #3936

### DIFF
--- a/packages/mdc-notched-outline/README.md
+++ b/packages/mdc-notched-outline/README.md
@@ -80,6 +80,13 @@ Mixin | Description
 `mdc-notched-outline-shape-radius($radius, $rtl-reflexive)` | Sets the rounded shape to notched outline element with given radius size. Set `$rtl-reflexive` to true to flip radius values in RTL context, defaults to false.
 `mdc-notched-outline-idle-shape-radius($radius, $rtl-reflexive)` | Sets the rounded shape to notched outline element in idle state with given radius size. Set `$rtl-reflexive` to true to flip radius values in RTL context, defaults to false.
 
+
+### Sass Functions
+
+Function | Description
+--- | ---
+`mdc-notched-outline-get-notch-padded-position($cornerSize)` | Returns the notch padded position based on given radius. This is 'x' position where the floating label starts.
+
 #### Calling Mixins with Parent Selectors
 
 Because notched-outline has sibling elements, you need to call the "idle" Sass mixins with parent selectors.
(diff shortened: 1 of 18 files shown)


abhiomkar:
I updated this PR to not move notch farther when setting large corner size. Now it stays close to corner with a minimum leading notch length. PTAL.

image

riichard:
I got the following error after upgrading to this version (0.41.0) and using the standard implementation from the readme. The error seems to be related to the changes in this PR. Downgrading to the previous version (0.40.1) resolves the issue.

```
🚨 Build Error

/my-project-directory/styles.scss:33:25: "12px-mdc-shape-prop-value(small)-1.2" is not a number for `max'

Error: "12px-mdc-shape-prop-value(small)-1.2" is not a number for `max'
        on line 33 of node_modules/@material/notched-outline/_functions.scss, in function `max`
        from line 33 of node_modules/@material/notched-outline/_functions.scss, in function `mdc-notched-outline-get-notch-padded-position`
        from line 249 of node_modules/@material/textfield/_mixins.scss, in mixin `mdc-text-field-outline-shape-radius`
        from line 316 of node_modules/@material/textfield/_mixins.scss, in mixin `mdc-text-field-outlined-`
        from line 116 of node_modules/@material/textfield/mdc-text-field.scss
        from line 8 of stdin
>>   $leadingStrokeLength: max(0, $mdc-notched-outline-min-leading-stroke-edge-
   ------------------------^```

kfranqueiro:
@riichard your textfield and shape package versions are out of sync. See https://github.com/material-components/material-components-web/issues/4038#issuecomment-435928192.

Issue #3742, opened 2018-10-11
Drawer: Update mdc-drawer styles to reuse mixins created under mdc-side-sheet


kfranqueiro:
* Mixins which are side-dependent should explicitly use the `leading` side
* Drawer should reuse `mdc-side-sheet-app-content`

abhiomkar:
Closing this issue as obsolete.

PR #3946, opened 2018-10-19
feat(side-sheet): Make mixins more consistent with Drawer

kfranqueiro:
Refs #3742.

This removes opacity parameters to match how drawer's mixin APIs are currently exposed (with the reasoning that it's easier to add later than remove), and reorganizes some styles to make them reusable between side sheet and drawer.

### DIFF
--- a/packages/mdc-side-sheet/README.md
+++ b/packages/mdc-side-sheet/README.md
@@ -86,7 +86,7 @@ Dismissible side sheets are by default hidden off screen, and can slide into vie
 ```
 
 > The main content of your page/app should reside in a `mdc-side-sheet-app-content` sibling element after the side sheet,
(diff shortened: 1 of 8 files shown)


packages/mdc-side-sheet/README.md
```diff
@@ -85,7 +85,8 @@ Dismissible side sheets are by default hidden off screen, and can slide into vie
 </body>
 ```
 
-> Apply the `mdc-side-sheet-app-content` class to a subsequent sibling element after the side-sheet for the open/close animations to work.\
+> The main content of your page/app should reside in a `mdc-side-sheet-app-content` sibling element after the side sheet,
+> in order to adjust appropriately when the drawer opens and closes.
```
abhiomkar: did you mean side sheet?

PR #3947, opened 2018-10-19
feat(drawer): Reuse side-sheet styles

### LINKED ISSUE #3742: Drawer: Update mdc-drawer styles to reuse mixins created under mdc-side-sheet

kfranqueiro:
Fixes #3742.

This branch relies on #3946, so I currently have it based on that branch for diff/review purposes.

BREAKING CHANGE: mdc-drawer-app-content class is replaced by mdc-side-sheet-app-content class, which is shared between the two components.


packages/mdc-drawer/_mixins.scss
```diff
@@ -119,13 +110,11 @@
...
 @mixin mdc-drawer-surface-fill-color($color) {
-  $value: mdc-theme-prop-value($color);
-
-  @include mdc-theme-prop(background-color, $value);
+  @include mdc-side-sheet-surface-fill-color($color);
 }
 
 @mixin mdc-drawer-scrim-fill-color($color) {
-  $value: rgba(mdc-theme-prop-value($color), $mdc-drawer-modal-scrim-opacity);
+  $value: rgba(mdc-theme-prop-value($color), $mdc-side-sheet-modal-scrim-opacity);
```
abhiomkar: So, are we not sharing `mdc-side-sheet-scrim` with drawer?

packages/mdc-drawer/mdc-drawer.scss
```diff
@@ -20,6 +20,108 @@
...
-@import "./common";
-@import "./dismissible/mdc-drawer-dismissible";
-@import "./modal/mdc-drawer-modal";
+@import "@material/side-sheet/mixins";
+@import "@material/side-sheet/variables";
+@import "@material/typography/mixins";
+@import "./mixins";
+@import "./variables";
+
+// postcss-bem-linter: define drawer
```
abhiomkar: Curious, why is this annotation required?

packages/mdc-drawer/mdc-drawer.scss
```diff
@@ -20,6 +20,108 @@
...
+  .mdc-list-item__text,
+  .mdc-list-item__graphic {
+    pointer-events: none;
+  }
+}
+
+.mdc-drawer__header {
+  flex-shrink: 0;
+  box-sizing: border-box;
+  min-height: 64px; // same as $mdc-top-app-bar-row-height
```
acdvorak: Would it make sense to use the value of `$mdc-top-app-bar-row-height` directly?

packages/mdc-drawer/README.md
```diff
@@ -68,6 +68,8 @@ However, you can also use SVG, [Font Awesome](https://fontawesome.com/), or any
 ```scss
 @import "@material/drawer/mdc-drawer";
 @import "@material/list/mdc-list";
+// For dismissible drawer, include mdc-side-sheet for the shared mdc-side-sheet-app-content class
```
acdvorak: Would this be a good candidate for the `mdc-base-emit-once()` mixin?

packages/mdc-drawer/mdc-drawer.scss
acdvorak: I think it enforces that each `.scss` file should only contain one top-level "block" class
(in this case, `mdc-drawer`).

Referencing any other block classes will throw an error, which is why we need a
`// postcss-bem-linter: ignore` comment on selectors like `.mdc-list-item:nth-child(1)`.

packages/mdc-drawer/README.md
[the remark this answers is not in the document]
kfranqueiro: Hm. Actually we could just genericize that heading to just "Icons" since the drawer items each have an icon, too, and the note is equally applicable to that.

packages/mdc-drawer/mdc-drawer.scss
kfranqueiro: I suspect we didn't do that to avoid adding a hard dependency on the top-app-bar package, when drawer can be used with or without one.

packages/mdc-drawer/README.md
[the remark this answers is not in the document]
kfranqueiro: Sent #3990 to fix this in master.

kfranqueiro: There are probably several places we could make this argument... so far we've only used emit-once for the packages that support text field / select and are basically never used standalone (line-ripple, notched-outline, floating-label).

It occurs to me that future discussions on how we organize our styles might impact how we use emit-once, so I'm going to hold off on opting into this here for now, but by the time we return to side sheet to finish it, we might have a better idea.

packages/mdc-drawer/_mixins.scss
kfranqueiro: I guess we can do that provided no one needs to customize the scrim colors for each independently (which should probably be discouraged anyway). I'll see what that ends up looking like.

kfranqueiro: On second thought, I think these need to be separate, otherwise you're going to run into issues if you have a dismissible drawer but a modal side sheet, or vice versa.

acdvorak: I was wondering about app-content too.

packages/mdc-drawer/mdc-drawer.scss
acdvorak: If we're only importing `_variables.scss`, why does it matter?

It won't output any additional CSS, and it has the advantage that we don't have to remember to keep two magic values in sync.

What's more, `@material/top-app-bar` only adds `736K` to `node_modules` ([I checked](https://gist.github.com/acdvorak/706227c1458cfbff229e313c2a8ebc31)).
NPM dependency trees are typically _massive_; `736K` is barely even noticeable.

Personally, I would prioritize maintainability over NPM dependency size.

kfranqueiro: Yeah, that seems fair, I'll add it.

PR #3949, opened 2018-10-22
feat(select): Add enhanced select variant

### LINKED ISSUE #291: Add invalid styles to mdc-select

0.2.0

All browsers, but most visible in Safari.

### What OS are you using?

OSX

1. Create an mdc-select component in a form.
1. Add a default option with no value.
1. Add some more options with values.
1. Add the `required` attribute to the select.
1. In the browser, change from the default option.
1. Change back to the default option.

The `mdc-select` component does not apply a class such as `mdc-select--invalid` nor does it appear to exist in the source code. This should also happen when the form is submitted in an invalid state.

This should work the same as it already does for `mdc-textfield`, expected to automatically apply and remove a class such as `mdc-select--invalid` when the select changes validation state.

No class is applied and in Safari there is no visual indication that the field is invalid.

### LINKED ISSUE #1835: MDC Select floating label overflows to top

0.0.27

Chrome (also others I guess)

### What OS are you using?

Mac

1. Create a select box with very small options e.g. `Opt1, Opt2, Opt3`

2. Put the surface label bigger than options length e.g. `Select an option`

3. When selection an option in browser, notice that floating label is broken

If the floating label is quite big to display, may be add ellipses to it.

The text overflows out of the select viewport
screen shot 2017-12-24 at 8 52 31 am

### LINKED ISSUE #2211: Disabled select: bottom line position jumps

> 0.30.0

Chrome

### What OS are you using?

Windows 10

* Open the select demo page: https://material-components-web.appspot.com/select.html
* Scroll down to **Fully-Featured JS Component**
* Toggle **Disabled** on and off.

The position of the disabled bottom line should not change,

The bottom line shifts up a few pixels.

The "box" variant is OK.

### LINKED ISSUE #2227: Alignment between textfields and selects are off
0.29.0

All

### What OS are you using?
Mac OS

https://codepen.io/anon/pen/aqWoRP

Take a look at the codepen. The underline should be visually aligned vertically, but is not.

### LINKED ISSUE #2275: MDCSelect's options do not display

### LINKED ISSUE #2552: Add a non-native select box: mdc-select-menu
Recently the mdc-select component was regressed to only allow native select. The technical reasons are reasonable but this left the community without a reasonable looking menu-style select. The native select works but has a **very** inconsistent look and feel. 

What I propose is two separate components for select components:
* **mdc-select-menu**: this would be the removed component that was based on the mdc-menu
* **mdc-select-native**: the native select that we have now

As others have pointed out, this change is discouraging to both current and would be users of this library. Please allow developers to decide which component is best for their use case.

Thanks for listening to my proposal. 

Lastly, I can never give enough thanks to the team and contributors for building and maintaining this great library in the first place. So, thank you, again.

### LINKED ISSUE #3627: Select: Add dropdown arrow as new icon (with rotating on focus)
Add the dropdown arrow as a separate icon and have the icon rotate to point up when the menu is opened (on focus).

### LINKED ISSUE #3628: Select: Add --invalid variant to select
Add the `--invalid` variant to the select. 

This includes adding styles for the `--invalid` class to match text field, as well as adding support for a `--required` class (and behavior) for both the native and non-native select.

### LINKED ISSUE #3629: Select: Add support for leading icons (static)

### LINKED ISSUE #3648: Select hasLabel Adapter can be remove
https://github.com/material-components/material-components-web/pull/3433/files

After this change, I think hasLabel can be removed from the Select component.

### LINKED ISSUE #3706: Add helper text to MDC Select

### LINKED ISSUE #3728: Enhanced Select: Collect HTML/SCSS documentation under its own Variant subsection
Logging this separately to not hold up #3608.

Consider moving the html/scss parts about Enhanced Select out of Basic Usage to within a subheading in the Variants section (it can be the first one, to make further mentions under other variants make sense).

That way we can put the html and scss right next to each other which will make them make more sense in context. As-is, you have to bounce back and forth between other content in between to realize the connection with `demo-width-class`.

Creating a variant subsection dedicated to enhanced select would also give us a place to discuss the usability differences/trade-offs between the two.

### LINKED ISSUE #3729: Enhanced Select should synchronize with a hidden input
This is basically #2221 for Enhanced Select, since we're re-adding a version that uses MDC Menu rather than a select element.

The Enhanced Select needs to be able to synchronize its value with a hidden input element in order for it to be visible to HTML forms. This is critical functionality for a component intended to behave like a form input element.

williamernest:
Adds the leading icon variant to the select.
Adds the helper text to the select. 
Adds the rotating dropdown icon to the select. 
Adds the enhanced menu version of the select.
fixes #2552, fixes #3706, fixes #3629, fixes #3627, fixes #3628, fixes #291, fixes #3728, fixes #3648, fixes #3729, fixes #2275, fixes #2211,  fixes #2227, fixes #1835
BREAKING CHANGE: Several adapter APIs were added to support the enhanced variant. The drop-down arrow is now its own element. The change event is now MDCSelect:change for all variants. See the README for full details.


.travis.yml
```diff
@@ -13,6 +13,7 @@ branches:
   # This prevents excessive resource usage and CI slowness.
   only:
     - master
+    - feat/select-new
```
kfranqueiro: Remove this before merging

packages/mdc-select/index.js
```diff
@@ -168,43 +269,91 @@ class MDCSelect extends MDCComponent {
...
+    element.addEventListener('blur', this.handleBlur_);
+
+    ['mousedown', 'touchstart'].forEach((evtType) => {
+      element.addEventListener(evtType, this.handleClick_);
+    });
+
+    if (this.menuElement_) {
+      this.selectedText_.addEventListener('keydown', this.handleKeydown_);
+      this.menu_.listen(menuSurfaceConstants.strings.CLOSED_EVENT, this.handleMenuClosed_);
+      // Menu should open to the last selected element.
```
kfranqueiro: Move this comment into the body of `handleMenuOpened_` (around L279)

packages/mdc-select/_mixins.scss
```diff
@@ -212,3 +225,173 @@
...
+  cursor: default;
+  pointer-events: none;
+}
+
+@mixin mdc-select-outlined_ {
+  @include mdc-select-container-fill-color(transparent);
+  @include mdc-select-outline-color($mdc-select-outlined-idle-border);
+  @include mdc-select-hover-outline-color($mdc-select-outlined-hover-border);
+  @include mdc-select-focused-outline-color(primary);
+  @include mdc-floating-label-float-position($mdc-select-outlined-label-position-y);
```
kfranqueiro: I think this line is now redundant since mdc-select-outline-shape-radius invokes this a few lines later. This line was removed in #3929 (and was one of the conflicts listed earlier, I think).

porfirioribeiro:
I don't know much about MDC internals, but couldn't helper-text and probably icon be moved to their own component avoiding code duplication? As you did with line-ripple

williamernest:
@porfirioribeiro That is the plan in the future. We didn't have time for this release to separate the internals into their own packages but we'll end up doing that in the long run.

porfirioribeiro:
Ok thats great, avoiding code duplication if you use both.

PR #3982, opened 2018-10-23
fix(text-field): Fix textarea-shape-radius mixin behavior for input

kfranqueiro:
This fixes a gap which can end up appearing between the inner and outer elements of the textarea DOM. The style calculations were not taking into account the new radius types (e.g. `small`) and were also being skipped for single-value radius arguments.

### DIFF
--- a/test/screenshot/spec/mdc-textfield/fixture.js
+++ b/test/screenshot/spec/mdc-textfield/fixture.js
@@ -26,5 +26,11 @@ window.mdc.testFixture.fontsLoaded.then(() => {
     mdc.textField.MDCTextField.attachTo(el);
   });
 
+  const firstFocusField = document.querySelector('.mdc-text-field--focused input, .mdc-text-field--focused textarea');
+  if (firstFocusField) {
+    // Programmatically focus to get :focus applied
+    firstFocusField.focus();
+  }
+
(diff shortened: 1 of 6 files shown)


test/screenshot/spec/mdc-textfield/fixture.js
```diff
@@ -26,5 +26,11 @@ window.mdc.testFixture.fontsLoaded.then(() => {
     mdc.textField.MDCTextField.attachTo(el);
   });
 
+  const firstFocusField = document.querySelector('.mdc-text-field--focused input, .mdc-text-field--focused textarea');
+  if (firstFocusField) {
+    // Programmatically focus to get :focus applied
+    firstFocusField.focus();
```
abhiomkar: This'll make only the first element in the demo to have focus?

kfranqueiro: Yes. Only one element in the document can have focus at a time, but this is enough to test the textarea issue and I want to see how the screenshot tests come out to see if it can be useful for other text field pages as well.

kfranqueiro:
Unfortunately, forcing focus in the screenshot tests is flake-prone, due at least in part to the blinking text cursor, so I'm leaving the screenshot tests unaltered.

PR #3997, opened 2018-10-26
style(list): ✌️ out whitespace

patrickrodee:


kfranqueiro:
PS I think this commit can go in as "style(list):" since it's code formatting.

PR #4021, opened 2018-10-30
fix(list): Fix font size and placement for avatar graphic

### LINKED ISSUE #4007: mdc-list-item__graphic font-size to big in mdc-list--avatar-list

0.40.1

- Chrome
- IE
- Edge
- Firefox

### What OS are you using?

Windows 10

The font-size of the graphic in an mdc-list--avatar-list is set to 40px which creates an huge icon. This can be seen in the demo page: https://material-components.github.io/material-components-web-catalog/#/component/list

When looking at the design specs https://material.io/design/components/lists.html#types the icon is smaller and is what we expected.

An icon with the normal font-size in a mdc-list--avatar-list

An icon with a font-size set to 40px

kfranqueiro:
Fixes #4007.

### DIFF
--- a/packages/mdc-list/mdc-list.scss
+++ b/packages/mdc-list/mdc-list.scss
@@ -87,7 +87,9 @@
   @include mdc-list-item-primary-text-ink-color(text-disabled-on-background);
 }
 
-.mdc-list-item__graphic {
+// Extra specificity is to override .material-icons display style if used in conjunction with mdc-list-item__graphic
+// stylelint-disable plugin/selector-bem-pattern
+.mdc-list .mdc-list-item__graphic {
   @include mdc-list-graphic-size_(24px);
 
   display: inline-flex;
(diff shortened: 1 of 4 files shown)


packages/mdc-list/mdc-list.scss
```diff
@@ -87,7 +87,9 @@
   @include mdc-list-item-primary-text-ink-color(text-disabled-on-background);
 }
 
-.mdc-list-item__graphic {
+// Extra specificity is to override .material-icons display style if used in conjunction with mdc-list-item__graphic
```
acdvorak: Would it make sense to only increase specificity for `display: inline-flex`?

E.g.:

```scss
.mdc-list-item__graphic {
  @include mdc-list-graphic-size_(24px);

  flex-shrink: 0;
  align-items: center;
  justify-content: center;
  fill: currentColor;
}

// Extra specificity is to override .material-icons display style if used in conjunction with mdc-list-item__graphic
.mdc-list .mdc-list-item__graphic {
  display: inline-flex;
}
```

PR #4026, opened 2018-10-31
docs: Added link to Ember Material Components

hilljh82:


kfranqueiro:
I took a quick look and it looks like Ember Material Components meets the criteria. Thank you!

hilljh82:
@kfranqueiro Awesome! It was a challenge breaking it into packages, but I really like the approach when compared to MDL. The Ember applications are much smaller in size now.

Issue #3231, opened 2018-07-26
MDCTextField: Rebuild textarea to match spec. 


jamesmfriedman:
## Bugs

Follow the template below **and include a codepen** to ensure the quickest and most accurate response to your issue.

### What MDC Web Version are you using?

> 0.37.1

### What browser(s) is this bug affecting?

> all

### What OS are you using?

> MacOS High Sierra

### What are the steps to reproduce the bug?

It's a visual bug. I looked through the Material spec and it doesn't specifically talk about Textareas which I'm assuming means they should match non textarea inputs.

- With no additional additional stying. My guess is the texarea should just be underlined
  ![image](https://user-images.githubusercontent.com/975256/43262918-c54933d4-90af-11e8-9b5c-b776e11103ef.png)
- Both with outlined. I'd think this would be a common example and these should match
  ![image](https://user-images.githubusercontent.com/975256/43262977-e88fd906-90af-11e8-8caa-583d80aa3aa2.png)
- Both with box. Also seems like a viable example where they should match. This also shows a rendering issue where the textarea label actually has a light background color.
  ![image](https://user-images.githubusercontent.com/975256/43263016-0a48f082-90b0-11e8-82be-e63c71d7bcdf.png)

### What is the expected behavior?

Pending someone telling me the spec says otherwise, they should all match.

### What is the actual behavior?

They don't 

I'm really looking for opportunities to fix things upstream, this is another one I'd be happy to take on if someone just points me in the right direction.

acdvorak:
Thanks so much for bringing this to our attention, James!

@kfranqueiro will check with a few folks to see if there's a technical reason it was done this way, or if it was just an oversight/resourcing issue.

My gut tells me we should fix this, though.

nwpappas:
Design guidelines do imply the visual style should match:

![image](https://user-images.githubusercontent.com/2529208/43470403-540c0ae2-949d-11e8-992c-1f22e2cfa396.png)

kfranqueiro:
I checked with @amsheehan about the history of our textarea styles. We wrote what's currently in MDC Web for textarea prior to the updated text field spec being fully fleshed out - meaning that at the time, we only had underline-only text fields, and that clearly wouldn't fly for a textarea.

Now that filled and outlined variants exist, we should update textarea to match.

Note that we might want to try to land #2859 before anyone undertakes this, since otherwise there are likely to be a bunch of conflicts between the two.

ahlechandre:
at this time: 

[...]

PR #4035, opened 2018-11-01
refactor(notched-outline): Refactor notched outline to use 3 divs

### LINKED ISSUE #3032: mdc text field textarea text is hidden behind label whitespace

> 0.37.0

> Chome Browser

> Type the long text on textarea.
> Textarea padding will not take and overlap with the label.

> Textarea text should not overlap with the label.

> Textarea text is overlapping with the label.

image

### LINKED ISSUE #3472: MDCTextArea: Label doesn't cover all text/highlighting

0.39.0

Chrome

### What OS are you using?

macOS

Go to the catalog page for text fields and enter enough text into the textarea. 

The label should cover all of the text/highlighting. 

When scrolling up and down you can see text/highlighting on the far right side of the label. 

screen shot 2018-08-29 at 1 35 07 pm
screen shot 2018-08-29 at 1 20 03 pm

### LINKED ISSUE #3582: A11Y: SVG of notched-outline should have focusable="false" attribute for IE11
- Version: latest / 0.39.x
- Components: textfield with outline

Internet Explorer 11

### What OS are you using?
Windows 7 (modern.ie virtual box virtual machine)

The bug can be reproduced, when viewing the docs with the mentioned browser.
- visit https://material-components.github.io/material-components-web-catalog/#/component/text-field
- click inside the first outlined textfield (Standard)
- press "tab"

- The SVG element shouldn't be focusable
- Focus should move to the next visible input-field (Dense)

- In IE11 the SVG will gain the focus therefore you need to
- press "tab" again to move the focus to the second element

This can be fixed via the `focusable="false"` attribute:

```html
<svg focusable="false">
    <path class="mdc-notched-outline__path" d=""/>
</svg>
```

If this issue will be accpeted, there README.mds of textfield and notched outline needs to be adapted and examples needs to be updated.

### LINKED ISSUE #3721: Textfield w/ float label gives the border a blur/shadow
See bottom border rendering.

image

### LINKED ISSUE #3987: Outlined Text Field: shape-radius doesn't factor in Leading Icon

0.40.x

All

Edit the `baseline-leading-icon.html` screenshot test page, adding the `test-shaped-text-field--outline` class to each `mdc-text-field--outlined` element.

Floating label should align with the notched outline properly like it does in the shape-radius tests.

The floating-label style for text field with leading icon is overriding the custom shape radius styles.

We may be able to factor this into the shape-radius mixin.

The same problem exists in the MDC Select package on the feat/select-new branch (which will be merged for 0.41.0).

### LINKED ISSUE #4006: TextField Textarea is always full width
## Bugs

Please see https://codepen.io/anon/pen/aRxpXL . Note  that no matter how the text area is specified it takes the full width of its container, unlike the Text field. 

The demo https://material-components.github.io/material-components-web-catalog/#/component/text-field seems to indicate that it should be possible to create a text area which is not full width as there is a section for 'full width' versions of the textarea compared to those show above them.

0.40.1

Safari, Chrome

See the codepen

If `mdc-text-field--fullwidth` is not specified, the textarea should size according to say the `rows` attribute, and not take the full width of its container. 

The textarea always fills its container.

I believe this is because the textarea is defined with `display: flex` and `width: auto` result in this behaviour https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/_mixins.scss#L481

### LINKED ISSUE #4066: MDCNotchedOutline: Refactor notched-outline to not use SVG
The notched outline uses SVG path and this causes a lot of problems with rendering. This is a tracking issue to track migrating the svg outlne to use div and borders to get the same result.

williamernest:
Placeholder PR so I can run screenshots. 

Notes for diffs: 
 - The previous notched outline was drawn in SVG and was not exactly 56px tall, so the SVG-less outline for all instances of a notched-outline that is notched will show diffs in both height and width. 
 - The text will shift slightly either up or down and will not render in the exact same location (subpixel shift). I will get them as close as possible but some text will not be possible to align perfectly. 

* Fixes #4066
* Fixes #3032
* Fixes #3472
* Fixes #3582 
* Fixes #3721 
* Fixes #3987 
* Fixes #4006
* Refs #3231 

BREAKING CHANGE: The notched outline has been changed from using an SVG for the outline to using 3 div elements. This approach resolves initial rendering issues as well as inconsistencies between the different types of outlines. Please refer to the [Readme](./packages/mdc-notched-outline/README.md) or the [screenshot test pages](./test/screenshot/spec/mdc-textfield/classes) for details and examples.


packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -28,57 +28,58 @@
 
 // Notched Outline is intended for use by multiple components, but its styles should only be emitted once when bundled
 @include mdc-base-emit-once("mdc-notched-outline") {
-  .mdc-notched-outline,
-  .mdc-notched-outline__idle {
+  .mdc-notched-outline {
```
kfranqueiro: It looks like this branch removes the idle outline element completely, and thus also removes the fade. This seems a bit jarring... we might want to run that by one of our motion designers.

I don't think keeping the idle element would present too much of a problem? (Since now the borders should behave exactly the same, since the notched outline is also simply an element of the same size using CSS borders, rather than an SVG.)

packages/mdc-textfield/foundation.js
```diff
@@ -213,8 +213,7 @@ class MDCTextFieldFoundation extends MDCFoundation {
       const isDense = this.adapter_.hasClass(cssClasses.DENSE);
       const labelScale = isDense ? numbers.DENSE_LABEL_SCALE : numbers.LABEL_SCALE;
       const labelWidth = this.adapter_.getLabelWidth() * labelScale;
-      const isRtl = this.adapter_.isRtl();
-      this.adapter_.notchOutline(labelWidth, isRtl);
+      this.adapter_.notchOutline(labelWidth);
```
kfranqueiro: Are we able to completely remove the isRtl adapter API from text field and select now?

packages/mdc-notched-outline/index.js
```diff
@@ -39,15 +40,30 @@ class MDCNotchedOutline extends MDCComponent {
...
+    this.notchElement_;
+  }
+
+  initialSyncWithDOM() {
+    const label = this.root_.querySelector('.' + MDCFloatingLabelFoundation.cssClasses.ROOT);
+    this.notchElement_ = this.root_.querySelector(strings.NOTCH_ELEMENT_SELECTOR);
+
+    if (label) {
+      label.style.transitionDuration = '0s';
+      label.classList.add(MDCFloatingLabelFoundation.cssClasses.UPGRADED);
```
kfranqueiro: This seems odd... modifying the class list of a floating label from within notched outline code. If this logic needs to be within notched-outline, I would suggest we apply the upgraded class to notched-outline and adjust CSS selectors accordingly.

packages/mdc-select/_mixins.scss
```diff
@@ -221,12 +217,14 @@
...
   .mdc-floating-label {
-    @include mdc-rtl-reflexive-position(left, $mdc-select-outline-label-offset);
-
     bottom: 12px;
     line-height: 1.75rem;
     pointer-events: none;
   }
+
+  &:not(.mdc-select--outlined) .mdc-floating-label {
+    @include mdc-rtl-reflexive-position(left, $mdc-select-outline-label-offset);
```
kfranqueiro: Should this also be using the notched-outline-leading-width variable? It's unclear to me why this is using a different value (16px), and also why it's using a variable named for outlined select, in a selector that's specifically not outlined select.

test/screenshot/spec/mdc-select/classes/baseline.html
```diff
@@ -77,14 +77,14 @@
...
             </select>
-            <label class="mdc-floating-label" for="my-select2">Pick a Food Group</label>
             <i class="mdc-select__dropdown-icon"></i>
             <div class="mdc-notched-outline">
-              <svg>
-                <path class="mdc-notched-outline__path"/>
-              </svg>
+              <div class="mdc-notched-outline__leading"></div>
+              <div class="mdc-notched-outline__notch">
+                <label class="mdc-floating-label" for="my-select2">Pick a Food Group</label>
```
kfranqueiro: I've seen some instances where the floating label is still outside the notched outline element, and some where it's inside. Is that accidental or intentional?

williamernest: Accidental. The label should always be inside the notch if one exists.

packages/mdc-notched-outline/mdc-notched-outline.scss
williamernest: I plan to apply the transition properties from the idle element to the notched outline to animate color. That should result in a very similar appearance.

kfranqueiro: We'll have to see how that performs. Opacity can be GPU-composited; color can't.

williamernest: The previous idle element transitioned border-color and opacity.

packages/mdc-select/_mixins.scss
```diff
@@ -126,11 +126,7 @@
     @include mdc-shape-radius($resolved-radius, $rtl-reflexive);
   }
 
-  $cornerSize: mdc-shape-prop-value($resolved-radius);
-  $positionX: mdc-notched-outline-get-notch-padded-position($cornerSize);
-  $labelPositionOffset: - $mdc-select-outline-label-offset + $positionX;
-
-  @include mdc-floating-label-float-position($mdc-select-outlined-label-position-y, -1 * ($labelPositionOffset));
+  @include mdc-floating-label-float-position($mdc-select-outlined-label-position-y, 0);
```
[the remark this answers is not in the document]
williamernest: Yes

packages/mdc-floating-label/_mixins.scss
```diff
@@ -68,6 +68,38 @@
   }
 }
 
+@mixin mdc-floating-label-outlined-float-position($positionY, $positionX: 0%, $scale: .75) {
+  .mdc-floating-label--float-above {
+    font-size: ($scale * 1rem);
+
+    @if $positionX > 0 or $positionX < 0 {
+      transform: translateY(-1 * $positionY - 14%) translateX(-1 * $positionX) scale(1);
```
kfranqueiro: What's the significance of 14%? Is there some existing variable this is related to?

packages/mdc-textfield/_mixins.scss
```diff
@@ -444,70 +447,57 @@
...
+    border: none;
   }
 
   .mdc-floating-label {
-    @include mdc-floating-label-fill-color($mdc-textarea-background);
-    @include mdc-rtl-reflexive(left, 1px, right, 0);
-    @include mdc-rtl-reflexive-property(margin, 8px, 0);
-
-    top: $label-offset-y;
+    top: 17px;
```
kfranqueiro: This was formerly using `$label-offset-y` but is now hard-coded, and the `$label-offset-x`/`y` variables are now unused.

packages/mdc-textfield/mdc-text-field.scss
```diff
@@ -63,10 +63,6 @@
 
   // Floating Label
   @include mdc-text-field-floating-label_;
-  // The following mixins will be removed after select is updated to default to the box.
-  @include mdc-floating-label-float-position($mdc-text-field-label-position-y);
-  @include mdc-floating-label-shake-animation(text-field-box);
-  @include mdc-floating-label-shake-keyframes(text-field-box, $mdc-text-field-label-position-y);
```
kfranqueiro: Is filled text field using the default shake keyframes now?

packages/mdc-floating-label/_mixins.scss
```diff
@@ -68,6 +68,38 @@
   }
 }
 
+@mixin mdc-floating-label-outlined-float-position($positionY, $positionX: 0%, $scale: .75) {
```
kfranqueiro: Is there some way we could fathom integrating this into the notched-outline package rather than adding a notched-outline-specific mixin to floating-label? (e.g. would it be feasible to invoke the `mdc-floating-label-float-position` mixin twice within a mixin intended to set up the floating label within notched-outline?)

Otherwise, this seems like it might be confusing to use externally.

packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -28,57 +28,66 @@
...
-    opacity: 1;
+    .mdc-notched-outline__notch {
+      padding: 0 8px 0 0;
+      border-top: none;
+    }
   }
 
-  .mdc-notched-outline--notched ~ .mdc-notched-outline__idle {
-    opacity: 0;
+  .mdc-notched-outline--no-label {
```
kfranqueiro: Is this something we plan to detect and apply in JS? Do we need to document it anyway for FOUC prevention?

Would it alternatively be feasible to apply margin to the label instead, or would that be problematic due to the scaling?

kfranqueiro:
Some observations on screenshot test diffs:

* Abhinay pointed out that even the resting-state outlined text fields show diffs in their corners on Chrome and (to a lesser extent) Firefox, which seems odd since the idle outline is a div right now, too
* Something's up with the shape radius page https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/20/20_52_48_719/report/report.html#spec/mdc-textfield/mixins/outline-shape-radius.html:desktop_windows_chrome@latest
* The textarea sizes changed on most of the textarea pages, not sure if intentional; e.g. https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/20/20_52_48_719/report/report.html#spec/mdc-textfield/classes/textarea.html:desktop_windows_chrome@latest
* This edge diff looks suspicious to me and I wonder if it's a flake; other trailing icon pages don't have this diff. https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/20/20_52_48_719/report/report.html#spec/mdc-textfield/classes/invalid-trailing-icon.html:desktop_windows_edge@latest
* I'm seeing a lot of filled-variant diffs where the label shifts right (both floating and resting); I assume this is intentional, but then I wonder if it's intentional that it _doesn't_ shift right in cases with a leading icon?

williamernest:
1. I think the corner diffs are un-avoidable. If they were consistent across multiple browsers in multiple use cases it might be something we could specifically track down, but the overall border-radius is consistent. 
2. I'll look at the shape radius mixin
3. I removed the width: 100% from the textarea (#4006) so the width changed. The height changed because the floating label now goes all the way up to the notched outline so the text area doesn't have as much padding-top as it previously had (which caused the components overall height to change. 
4. I believe that diff is a flake, but I'll check it again after the next run. 
5. Good catch! The filled variant incorrectly had 12px from the leading edge instead of 16px.

kfranqueiro:
The latest round of screenshot diffs cause a diff in the width of all non-enhanced selects (which also causes there to be 3 additional diffs). e.g. https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/20/23_43_04_788/report/report.html#spec/mdc-select/classes/baseline.html:desktop_windows_chrome@latest

Previous report: https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/20/20_52_48_719/report/report.html

New report: https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/20/23_43_04_788/report/report.html

williamernest:
Yes, this was intentional. The select specs indicate there should be 24dp padding from the end of the option to the 24x24dp box that contains the dropdown arrow icon.

packages/mdc-notched-outline/mdc-notched-outline.scss
williamernest: I'll update the notched outline to add this class if there is no label. 

I tried to use both padding and margin on the label but with the scale applied it's difficult to line up the label with the input and helper text.

packages/mdc-floating-label/_mixins.scss
williamernest: Since I'm changing the font-size it changes the amount of translateY needed to get the label centered vertically into the notch. This is simply a static amount to adjust for the 25% reduction in font-size. I'll move it into a variable when I move this into the notched-outline package.

packages/mdc-floating-label/_mixins.scss
williamernest: Yes, updated.

packages/mdc-notched-outline/_mixins.scss
```diff
@@ -30,23 +32,49 @@
...
+
+  .mdc-notched-outline__trailing {
+    @include mdc-shape-radius(mdc-shape-mask-radius($radius, 0 1 1 0), $rtl-reflexive: true);
+  }
 }
 
-@mixin mdc-notched-outline-idle-shape-radius($radius, $rtl-reflexive: false) {
-  .mdc-notched-outline__idle {
-    @include mdc-shape-radius($radius, $rtl-reflexive);
+@mixin mdc-notched-outline-floating-label-position($positionY, $positionX: 0%, $scale: .75) {
```
kfranqueiro: Should this be `mdc-notched-outline-floating-label-float-position` for consistency?

packages/mdc-textfield/_mixins.scss
```diff
@@ -377,23 +374,36 @@
 
 @mixin mdc-text-field-with-leading-icon_ {
   @include mdc-text-field-icon-horizontal-position_(left, $mdc-text-field-icon-position, $mdc-text-field-icon-padding);
+
+  .mdc-floating-label {
+    @include mdc-rtl-reflexive-position(left, $mdc-text-field-icon-padding);
+
+    &--float-above {
+      @include mdc-rtl-reflexive-position(left, $mdc-text-field-icon-padding);
```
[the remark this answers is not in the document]
williamernest: Yeah, this wasn't needed. Deleted.

packages/mdc-notched-outline/README.md
```diff
@@ -95,11 +102,10 @@ Consider the following example HTML:
 ```html
 <div class="foo__parent">
   <div class="mdc-notched-outline foo__child">
-    <svg>
-      <path class="mdc-notched-outline__path"/>
-    </svg>
+    <div class="mdc-notched-outline__leading"></div>
```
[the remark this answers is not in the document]
williamernest: This whole section can be removed since it pertains to special considerations due to the unconventional structure of the old outline.

thdk:
Will this issue fix this? 

image

There is no outline visible on the right side of the text field.
After you click the textfield the outline on the right side appears and will stay even when clicking outside the textfield.

https://codepen.io/kfranqueiro/pen/QJVObP

Running on windows 10, chrome 70.0.3538.110

Or is this another issue?

williamernest:
It should resolve that issue.

packages/mdc-select/_mixins.scss
```diff
@@ -108,29 +109,27 @@
...
 
-  @include mdc-notched-outline-idle-shape-radius($resolved-radius, $rtl-reflexive);
-
-  &__native-control {
+  .mdc-select__native-control {
     @include mdc-shape-radius($resolved-radius, $rtl-reflexive);
   }
 
-  $cornerSize: mdc-shape-prop-value($resolved-radius);
-  $positionX: mdc-notched-outline-get-notch-padded-position($cornerSize);
```
kfranqueiro: The `get-notch-padded-position` function seems to be unused now. Should we remove it (and the entire functions file), or should we still have a use for it?

packages/mdc-textfield/_mixins.scss
```diff
@@ -243,27 +223,48 @@
     @include mdc-notched-outline-shape-radius($resolved-radius, $rtl-reflexive);
   }
 
-  @include mdc-notched-outline-idle-shape-radius($resolved-radius, $rtl-reflexive);
+  $radius-pixels: mdc-shape-prop-value($resolved-radius);
 
-  $cornerSize: mdc-shape-prop-value($resolved-radius);
-  $positionX: mdc-notched-outline-get-notch-padded-position($cornerSize);
-  $labelPositionOffset: - $mdc-text-field-outline-label-offset + $positionX;
+  @if ($radius-pixels > $mdc-notched-outline-leading-width) {
```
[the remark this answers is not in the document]
kfranqueiro: Tracked #4140 for this.

packages/mdc-notched-outline/_mixins.scss
```diff
@@ -22,31 +22,53 @@
...
-  .mdc-notched-outline__idle {
-    @include mdc-shape-radius($radius, $rtl-reflexive);
+@mixin mdc-notched-outline-floating-label-float-position($positionY, $positionX: 0%, $scale: .75) {
+  @include mdc-floating-label-float-position($positionY + $mdc-notched-outline-label-adjust, $positionX, 1);
+
+  .mdc-floating-label--float-above {
+    font-size: ($scale * 1rem);
+  }
+
+  .mdc-notched-outline--upgraded {
```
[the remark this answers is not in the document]
williamernest: I'll make a quick commit adding both selectors and see if it changes anything.

packages/mdc-notched-outline/README.md
```diff
@@ -67,59 +73,24 @@ CSS Class | Description
 --- | ---
 `mdc-notched-outline` | Mandatory. Container for the SVG of the notched outline path.
```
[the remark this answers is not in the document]
kfranqueiro: Man, I _knew_ I was still going to miss something in the README. Thanks for pointing it out.

PR #4046, opened 2018-11-02
chore: Update karma to run Chrome+FF headless and only IE on SL

### LINKED ISSUE #4041: Update karma.conf to only use SauceLabs for IE, and run Chrome and Firefox headless locally
Currently, our karma.conf is configured to use SauceLabs to run unit tests in Chrome, Firefox, iOS Safari, and IE.

However, it should be feasible to run both Chrome and Firefox headless locally without a remote provider. (We already run Chrome headless for PRs from forks, which can't access Sauce Labs).

iOS Safari is likely redundant of Chrome as far as unit tests are concerned, but it is important that we test IE since it is the most likely browser where something could slip through and trip it up.

The goal of reducing the number of jobs requiring Sauce Labs is so that if we migrate to a single provider (as opposed to using SL for unit tests and CBT for screenshot tests), we will have as few total jobs per PR as possible.

kfranqueiro:
Fixes #4041.

Sending out initially to see what happens on Travis. Tested locally on OS X forcing the travis flag to true.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -27,122 +27,29 @@ const webpackConfig = require('./webpack.config')[0];
 const USING_TRAVISCI = Boolean(process.env.TRAVIS);
 const USING_SL = Boolean(process.env.SAUCE_USERNAME && process.env.SAUCE_ACCESS_KEY);
 
-const LOCAL_LAUNCHERS = {
+const HEADLESS_LAUNCHERS = {
   /** See https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-348248951 */
   'ChromeHeadlessNoSandbox': {
     base: 'ChromeHeadless',
     flags: ['--no-sandbox'],
   },
+  'FirefoxHeadless': {
+    base: 'Firefox',
+    flags: ['-headless'],
+  },
 };
 
 const SAUCE_LAUNCHERS = {
-  /*
-   * Chrome (desktop)
-   */
-
-  'sl-chrome-stable': {
-    base: 'SauceLabs',
-    browserName: 'chrome',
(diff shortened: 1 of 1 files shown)


kfranqueiro:
It looks like exactly one unit test failed on FF heaadless on Linux. I'll have to see if I can reproduce locally on a Linux workstation.

kfranqueiro:
Apparently the unit tests all passed now, without me changing anything? I'd still like to try running this on a linux workstation a few times; I don't want to be introducing new flakes.

Andy also pointed out that this might affect codecov numbers, and the codecov report seems to indicate that...though I can't fathom cutting out Safari affecting it this much, so I want to investigate what it's saying changed in coverage.

kfranqueiro:
Unfortunately I'm unable to reproduce the clearly-flaky top app bar test on headless firefox on a local linux workstation. It's rather odd though, since it ultimately should be accessing the same property (scrollTop) on the same element within the same execution turn for both the actual and expected values... I've made a couple of adjustments to the test in hopes of de-flaking it.

Meanwhile, regarding code coverage, I don't think this affects it. Chrome, Firefox, and Mobile Safari all each report the exact same percentages, while IE reports less due to skipped ripple tests. I think the `text-summary` coverage report format is seriously affected by the order in which the tests are run - if I run IE last, I get its percentages; if I run anything else last, I get the other (higher) numbers.

kfranqueiro:
I split out the script update to #4105 so that this PR shouldn't end up running screenshot tests in the future, and discovered the root cause of the test failure (it was legit, not a flake), which will be addressed in #4106. Once those both go in, I should be able to rebase this and pass CI.

Issue #3403, opened 2018-08-23
MDC List: Remove aria-selected="true" from selected item


abhiomkar:
As per [ARIA spec](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-selected) `aria-selected` property is used only to specific roles.

kfranqueiro:
Given that we previously augmented the List component to auto-detect `activated` behavior vs. `selected`, we may also be able to auto-detect the presence of roles in which `aria-selected` is valid, and in other cases opt for [`aria-current`](https://www.w3.org/TR/wai-aria-1.1/#aria-current) instead (which is the most semantically-correct attribute to use for drawer navigation).

I'm not sure if we'd want to simply build this in to the same check for activated vs. selected... it might work for now, but could still yield invalid combinations down the line.

PR #4055, opened 2018-11-05
fix(list): Update ARIA attributes for radio/checkbox based list

### LINKED ISSUE #3659: MDCList: List items with checkbox/radio buttons should use aria roles on li since the controls are not focusable
The screen reader does not read the state of the checkbox since the checkbox is not focusable. The list item should contain `role="checkbox|radio"` and should toggle the `aria-checked` property to match the checkbox state.

abhiomkar:
This change updates the aria attributes (`aria-selected` or `aria-checked`) for single select, checkbox or radio based lists.

BREAKING CHANGE: Replaced `toggleCheckbox` adapter method with `setCheckedCheckboxOrRadioAtIndex` and added 3 more new adapter methods for improved accessibility.

Fixes #3659, #3403, #4107


packages/mdc-list/README.md
```diff
@@ -39,7 +39,7 @@ npm install @material/list
 ### HTML Structure
 
 ```html
-<ul class="mdc-list" aria-orientation="vertical">
```
kfranqueiro: Why was aria-orientation removed from examples? The default is documented as horizontal.

packages/mdc-list/README.md
```diff
@@ -215,6 +215,131 @@ var list = new mdc.list.MDCList(listEle);
...
+      </div>
+    </span>
+    <label class="mdc-list-item__text" for="demo-list-radio-item-3">Option 3</label>
+  </li>
+</ul>
+```
+
+### List with checkbox items
+
+When rendering list with checkbox items all pre-selected list items should contain `aria-checked` set to `true` and the native checkbox input element should contain `checked` attribute, all other list items should have `aria-checked` set to `false`. Each list item in checkbox list contains `role="checkbox"` attribute.
```
kfranqueiro: `false` is the default value for `aria-checked`, so it shouldn't be necessary to specify that on every unselected item, right?

packages/mdc-list/constants.js
```diff
@@ -34,6 +34,9 @@ const strings = {
   ARIA_ORIENTATION: 'aria-orientation',
   ARIA_ORIENTATION_HORIZONTAL: 'horizontal',
   ARIA_SELECTED: 'aria-selected',
+  ARIA_CHECKED: 'aria-checked',
+  IS_ARIA_CHECKED: '[aria-checked="true"]',
```
kfranqueiro: Rename to ARIA_CHECKED_SELECTOR

packages/mdc-list/index.js
```diff
@@ -160,8 +160,10 @@ class MDCList extends MDCComponent {
 
   initializeListType() {
     // Automatically set single selection if selected/activated classes are present.
-    const preselectedElement =
-      this.root_.querySelector(`.${cssClasses.LIST_ITEM_ACTIVATED_CLASS}, .${cssClasses.LIST_ITEM_SELECTED_CLASS}`);
+    const preSelectedSelector = `.${cssClasses.LIST_ITEM_ACTIVATED_CLASS},
+                                    .${cssClasses.LIST_ITEM_SELECTED_CLASS},
+                                    ${strings.IS_ARIA_CHECKED}`;
```
kfranqueiro: I don't think we want this change... won't this automatically set singleSelection for lists containing radios or checkboxes? We definitely don't want this for checkbox lists, and I'm not sure if we even want it for radio lists right now.

cc @williamernest

packages/mdc-list/foundation.js
```diff
@@ -109,29 +112,36 @@ class MDCListFoundation extends MDCFoundation {
 
   /** @param {number} index */
   setSelectedIndex(index) {
-    if (index === this.selectedIndex_) {
+    if (index < 0 && index >= this.adapter_.getListItemCount()) {
```
kfranqueiro: Why does this no longer bail out if `setSelectedIndex` is called redundantly?

Also, I suspect we intentionally did not bail out immediately if selectedIndex was out of range, in order to support setting it to -1. @williamernest is that correct?

packages/mdc-list/foundation.js
```diff
@@ -109,29 +112,36 @@ class MDCListFoundation extends MDCFoundation {
...
+      this.adapter_.setAttributeForElementIndex(prevSelectedIndex, ariaAttribute, false);
+    }
+    const ariaAttributeValue = hasCheckbox ? this.adapter_.isCheckboxCheckedAtIndex(index) : true;
+    this.adapter_.setAttributeForElementIndex(index, ariaAttribute, ariaAttributeValue ? 'true' : 'false');
+
+    // Update class names
+    const selectedClassName = this.useActivatedClass_ ?
+      cssClasses.LIST_ITEM_ACTIVATED_CLASS :
+      cssClasses.LIST_ITEM_SELECTED_CLASS;
+    if (!hasCheckboxOrRadio) {
```
kfranqueiro: This seems like a confusing workaround for the fact that we're (incorrectly) setting single selection mode in the presence of checkbox or radio inputs now. (See my other comment in index.js)

packages/mdc-list/foundation.js
```diff
@@ -109,29 +112,36 @@ class MDCListFoundation extends MDCFoundation {
...
+    const ariaAttribute = hasCheckboxOrRadio ? strings.ARIA_CHECKED : strings.ARIA_SELECTED;
 
-    if (this.selectedIndex_ >= 0) {
-      this.adapter_.removeAttributeForElementIndex(this.selectedIndex_, strings.ARIA_SELECTED);
-      this.adapter_.removeClassForElementIndex(this.selectedIndex_, className);
-      this.adapter_.setAttributeForElementIndex(this.selectedIndex_, 'tabindex', -1);
+    // Update aria attributes
+    const hasCheckbox = this.adapter_.hasCheckboxAtIndex(index);
+    if (!hasCheckbox) {
+      this.adapter_.setAttributeForElementIndex(prevSelectedIndex, ariaAttribute, false);
```
kfranqueiro: Should this be checking hasCheckboxOrRadio? Otherwise wouldn't this set the attribute even if there were no checkbox/radio in the previous item at all? (Though I'm not sure if we even expect to have lists where some items have inputs and some don't, so maybe this idea is excessive)

packages/mdc-list/foundation.js
```diff
@@ -291,6 +304,20 @@ class MDCListFoundation extends MDCFoundation {
...
+   * @param {number} index
+   */
+  toggleCheckboxOrRadioAtIndex_(index) {
+    if (!this.adapter_.hasCheckboxOrRadioAtIndex(index)) return;
+
+    let isChecked = true;
+    if (this.adapter_.hasCheckboxAtIndex(index)) {
+      isChecked = !this.adapter_.isCheckboxCheckedAtIndex(index);
+    }
+    this.adapter_.setCheckedCheckboxOrRadioAtIndex(index, isChecked);
```
kfranqueiro: To clarify, was this adapter API changed primarily to move some of the logic into the foundation?

packages/mdc-list/README.md
abhiomkar: `listbox` is applicable only if one of the list item is selectable. But in our baseline version we don't add that support. 

I'll add a note about aria orientation in accessibility section.

abhiomkar: `role="list"` or `ul/li` elements are supposed to be keyboard interactive as per ARIA spec. I'm inclined towards using `aria-orientation` only when applicable. User may not initialize MDCList on non-interactive lists. WDYT?

williamernest: `aria-orienation` is used to indicate which arrow keys the foundation listens to for traversing the list. If we remove this from all `ul` examples that are not `role="listbox"` we need to change the attribute we use to indicate list direction.

abhiomkar:
EDIT: role="list" or ul/li (without ‘’listbox’) elements are not supposed
to be keyboard interactive as per ARIA spec. That includes not arrowing
through list.

Fixed typo. Sorry I missed ‘not’ in my previous comment.

packages/mdc-list/README.md
williamernest: Response to your comment through email: 

I understand that it's not a part of the spec, but this functionality is specifically what we wanted to make lists more accessible regardless if you can make a selection or not.

williamernest: I don't think the spec specifically disallows this behavior, it's just not mentioned IIRC.

packages/mdc-list/foundation.js
abhiomkar: Didn't see any documentation related to negative tab index deselects the list.

Regarding redundantly setting the `setSelectedIndex` on same index - it actually toggles the checkbox if called twice on same index.

packages/mdc-list/foundation.js
```diff
@@ -109,29 +112,74 @@ class MDCListFoundation extends MDCFoundation {
...
+    this.adapter_.setAttributeForElementIndex(index, strings.ARIA_CHECKED, ariaAttributeValue);
+  }
+
+  /**
+   * @param {number} index
+   * @private
+   */
+  setAriaAttributesForRadio_(index) {
+    if (this.selectedIndex_) {
+      this.adapter_.setAttributeForElementIndex(this.selectedIndex_, strings.ARIA_CHECKED, false);
```
kfranqueiro: `false` -> `'false'`

packages/mdc-list/foundation.js
kfranqueiro: Discussed w/ Abhinay; this logic should be reasonable for this PR, since `setSelectedIndex` formerly didn't do anything for checkboxes at all.

It also turns out that we already worded this as "Toggles" in the readme, which is more correct now than it was before. We might want to clarify that it toggles for checkboxes but always sets the indicated index as selected for radios and single selection.

packages/mdc-list/foundation.js
```diff
@@ -198,18 +249,21 @@ class MDCListFoundation extends MDCFoundation {
...
         if (this.isSingleSelectionList_) {
           // Check if the space key was pressed on the list item or a child element.
-          this.setSelectedIndex(currentIndex);
           this.preventDefaultEvent_(evt);
         }
 
-        // Explicitly activate links, since we're preventing default on Enter, and Space doesn't activate them.
-        this.adapter_.followHref(currentIndex);
-      }
+        const hasCheckboxOrRadio = this.hasCheckboxOrRadioAtIndex_(listItemIndex);
```
kfranqueiro: Can you clarify why these conditions were moved inside the `isRootListItem` check? I assume it has to do with avoiding double-triggers if e.g. a checkbox happens to be focused inside a list item, but I'm less clear on what this does for pure single-selection lists (if it even makes a difference).

packages/mdc-list/foundation.js
```diff
@@ -109,29 +112,77 @@ class MDCListFoundation extends MDCFoundation {
...
+    } else {
+      this.setAriaAttributesForSingleSelect_(index);
+      this.setClassNamesForSingleSelect_(index);
     }
 
-    const className = this.useActivatedClass_
-      ? cssClasses.LIST_ITEM_ACTIVATED_CLASS : cssClasses.LIST_ITEM_SELECTED_CLASS;
+    if (this.selectedIndex_ >= 0 && this.selectedIndex_ !== index) {
+      this.adapter_.setAttributeForElementIndex(this.selectedIndex_, 'tabindex', -1);
+    } else if (this.selectedIndex_ === -1 && index !== 0) {
```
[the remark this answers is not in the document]
abhiomkar: That's correct. We're saving additional `tabindex` change which will be set in below statement anyway in that case. Added a code comment about this line.

I didn't want to bring additional complexity by allowing `-1` tabindex since this method is now used for single select, checkbox and radio variants which was not the case before.

packages/mdc-list/foundation.js
abhiomkar: Checkbox / Radio input's are never reachable using keyboard. The keyboard action (space / enter) happens only on list item root level.

For single selection list case we're not changing anything, it was already inside this condition (`isRootListItem`).

packages/mdc-list/mdc-list.scss
```diff
@@ -105,6 +105,11 @@
   @include mdc-typography-overflow-ellipsis;
 }
 
+// Disable interaction on label elements that may automatically toggle correspoding checkbox / radio input.
```
kfranqueiro: Typo: corresponding

packages/mdc-list/mdc-list.scss
```diff
@@ -105,6 +105,11 @@
...
+label.mdc-list-item__text[for] {
```
kfranqueiro: Since we're including the class and the attribute selector, would this selector work just as well without `label`? (Since `for` pretty much scopes this down to `label` already)

test/unit/mdc-list/foundation.test.js
```diff
@@ -652,3 +691,27 @@ test('#setUseActivatedClass causes setSelectedIndex to use the --activated class
...
+
+test('#setSelectedIndex should bail out if not in the range', () => {
+  const {foundation, mockAdapter} = setupTest();
+
+  td.when(mockAdapter.getListItemCount()).thenReturn(4);
+  foundation.setSelectedIndex(-1);
+  td.verify(mockAdapter.setAttributeForElementIndex(-1, 'tabindex', 0), {times: 0});
+});
+
+test('#setSelectedIndex should set aria attributes on new index and should not change aria attributes on previous' +
```
kfranqueiro: This is specifically for checkbox, right? Can we include checkbox somewhere in the test description? And do we have a similar test for radio / single select?

packages/mdc-list/README.md
```diff
@@ -266,15 +391,21 @@ Mixin | Description
 
 The MDCList JavaScript component implements the WAI-ARIA best practices for
 [Listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox). This includes overriding the default tab behavior
-within the list component. You should not add `tabindex` to any of the `li` elements in a list.
+within the list component.
+
+The `tabindex` should be set to `0` for first list item element or selected list item element, remaining list item elements should not have `tabindex` set to `true`.
```
kfranqueiro: is `true` supposed to be `0` here? Or just leave it at "should not have `tabindex` set"?

PR #4069, opened 2018-11-08
fix(dialog): Release focus after style changes on close

### LINKED ISSUE #4033: Dialog: Some screen readers don't navigate correctly on open/close
The catalog page can be used to reproduce these issues: https://material-components.github.io/material-components-web-catalog/#/component/dialog

In iOS, upon opening a dialog with a MDC List in it (simple / confirmation), the screen reader is still focused on the button that opened the dialog, even though focus is on the first list item.

In iOS and OS X Safari, upon closing a dialog, the screen reader doesn't return to the button that opened the dialog. This is especially surprising in the case of Safari, since keyboard focus returns to the correct element, but for some reason the screen reader decides to navigate somewhere else. Chrome appears to be fine.

I'm not sure whether this is something we can resolve or work around, but it's worth investigating.

abhiomkar:
Fixes #4033

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+<a name="0.41.1"></a>
+## [0.41.1](https://github.com/material-components/material-components-web/compare/v0.41.0...v0.41.1) (2018-11-14)
+
+
+### Bug Fixes
+
+* **card:** Corrected baseline shape value of card small => medium ([#4060](https://github.com/material-components/material-components-web/issues/4060)) ([875b159](https://github.com/material-components/material-components-web/commit/875b159))
+* **drawer:** allow drawer below top app bar ([#4028](https://github.com/material-components/material-components-web/issues/4028)) ([1eff602](https://github.com/material-components/material-components-web/commit/1eff602))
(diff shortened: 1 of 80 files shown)


kfranqueiro:
Testing in Safari using the catalog page against this branch, it seems like the problem isn't fully resolved. It also seems like it might be somewhat dependent on how far VoiceOver is into its announcements before you close the dialog. Even without this change, sometimes VO will automatically navigate to the top bar of the app after announcing the cancel button and dialog.

image

(The black outline is the screen reader still navigated onto the cancel button in the now-closed dialog)

Also, I don't think this change will do anything for the other issue mentioned in #4033 RE opening the confirmation or simple dialog and the screen reader not properly navigating into it?

abhiomkar:
This change fixes Safari + VoiceOver on Mac. But doesn't fix iOS Safari + VoiceOver issue because of iOS platform limitation.

Apparently `document.activeElement` which is used to get the currently focused element doesn't work. I tested with this [codepen example](https://codepen.io/abhiomkar/pen/bQYRmQ?editors=1010) on iOS to demonstrate this bug. (But, works on Android browsers).

Workaround for this iOS issue would be to send button element selector that dialog needs to restore focus upon close instead of relying on `document.activeElement` as shown in [ARIA Modal Dialog example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html). But I'm inclined towards not to optimize the solution for single platform.

Issue #4062, opened 2018-11-07
Snackbar: Create JavaScript API


acdvorak:
- [x] Simplify foundation:
    - [x] Remove focus management
    - [x] Remove parameters from `show()`
- [x] Create getters/setters for timeout, label text, action text, and <kbd>esc</kbd> key behavior
- [x] Ensure that clients can easily implement queueing
- [x] Implement `[de]registerEventHandler` methods in `MDCSnackbar` instead of adapter (#2813)
- [x] Unit tests

PR #4070, opened 2018-11-09
feat(snackbar): Update styles to match guidelines

### LINKED ISSUE #4061: Snackbar: Create styles to match guidelines
- [x] CSS classes
- [x] Sass variables (with `!important`)
- [x] Sass mixins
- [x] Screenshot test pages
- [x] Delete old snackbar demo page

acdvorak:
Fixes #4061

This PR also deletes snackbar's JS files and unit tests to avoid spurious test failures.
They will be added back in the next PR (for issue #4062).


test/screenshot/golden.json
```diff
@@ -1077,6 +1077,123 @@
       "desktop_windows_ie@11": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/10/23/19_08_56_617/spec/mdc-select/mixins/shape-radius.html.windows_ie_11.png"
     }
   },
+  "spec/mdc-snackbar/classes/baseline-with-action.html": {
```
abhiomkar: Snackbar is being clipped in iPhone 5 viewport.

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,180 @@
...
-  color: $mdc-snackbar-foreground-color;
 
-  @media (min-width: $mdc-snackbar-tablet-breakpoint) {
-    @include mdc-rtl-reflexive-property(padding, 0, 24px);
+  // Prevent mdc-snackbar__overlay's background-color from bleeding past the surface's border-radius.
+  overflow: hidden;
+
+  .mdc-snackbar--stacked & {
+    flex-direction: column;
+    align-items: flex-start;
```
abhiomkar: Should this flip for RTL?

acdvorak: Nope! That's the beauty of flexbox and `flex-start`: It's already RTL-aware! 🎉 

BTW, you can add `?dir=rtl` to any screenshot test page URL to see what it looks like in RTL:

https://storage.googleapis.com/mdc-web-screenshot-tests/advorak/2018/11/09/02_51_51_193/spec/mdc-snackbar/index.html?dir=rtl

#### LTR:

image

#### RTL:

image

test/screenshot/golden.json
acdvorak: Wow, good catch! 👍

That's due to the default `min-width` value of `344px` from the spec:

https://material.io/design/components/snackbars.html#spec

image

I'll follow up with Design on this.

(OOC, why did you think to test it on an iPhone 5? Never would have occurred to me 😝)

packages/mdc-snackbar/_mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+@mixin mdc-snackbar-elevation($z-index) {
+  .mdc-snackbar__surface {
+    @include mdc-elevation($z-index);
+  }
+}
+
+@mixin mdc-snackbar-shape-radius($radius) {
+  // The mdc-snackbar__overlay selector is needed to prevent the overlay's
```
kfranqueiro: You say the `snackbar__overlay` selector is needed... but it's not there?

packages/mdc-snackbar/_mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+  // background-color from leaking past the border-radius of the
+  // mdc-snackbar__surface element in Safari 12 (and possibly other versions).
+  .mdc-snackbar__surface {
+    @include mdc-shape-radius($radius);
+  }
+}
+
+@mixin mdc-snackbar-surface-fill-color($color) {
```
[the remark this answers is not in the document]
kfranqueiro: I'd be agreeable to one mixin w/ arguments for both colors. Separate mixins seems prone to people accidentally overlooking one or the other and not realizing they influence each other.

packages/mdc-snackbar/_variables.scss
```diff
@@ -20,9 +20,37 @@
...
+
+$mdc-snackbar-elevation: 6 !default;
+$mdc-snackbar-shape-radius: small !default; // Key from $mdc-shape-category-values or CSS length value (e.g., 4px)
+$mdc-snackbar-z-index: 8 !default; // One above mdc-dialog
+
+//
+// Private
+//
+
+$mdc-snackbar-enter-delay_: 0ms;
```
kfranqueiro: Is there a reason these are distinguished as private, given the request we've gotten to expose everything as overrideable?

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2017 Google Inc.
+// Copyright 2018 Google Inc.
```
kfranqueiro: We typically don't change the copyright year in pre-existing files. I can buy the argument in other cases that the files are effectively being replaced, but this one especially seems somewhat intact.

packages/mdc-snackbar/package.json
```diff
@@ -16,7 +16,12 @@
   "dependencies": {
     "@material/animation": "^0.41.0",
     "@material/base": "^0.41.0",
+    "@material/button": "^0.41.0",
+    "@material/icon-button": "^0.41.0",
+    "@material/dom": "^0.41.0",
```
kfranqueiro: I take it you're expecting to use this in the JS portion?

acdvorak: Yep, I just forgot to remove it from this PR. Fixed!

packages/mdc-snackbar/mdc-snackbar.scss
acdvorak: I actually rewrote this file from scratch, so I figured that needed a year bump

packages/mdc-snackbar/_variables.scss
acdvorak: Good call—made them "public"

packages/mdc-snackbar/_mixins.scss
acdvorak: Good catch, removed

packages/mdc-snackbar/_mixins.scss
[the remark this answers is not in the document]
acdvorak: Done!

packages/mdc-snackbar/_mixins.scss
```diff
@@ -0,0 +1,80 @@
...
+}
+
+@mixin mdc-snackbar-shape-radius($radius) {
+  .mdc-snackbar__surface {
+    @include mdc-shape-radius($radius);
+  }
+}
+
+@mixin mdc-snackbar-fill-color($overlay-color, $overlay-opacity, $surface-color: surface) {
+  .mdc-snackbar__surface {
```
abhiomkar: `surface` is a global override and can be customized using sass variable - `$mdc-theme-surface`. User may not need to send surface color here again?

I'm assuming the overlay color should blend with the global surface color which might be rendered behind snackbar from other components.

test/screenshot/golden.json
abhiomkar: I was testing if snackbar snaps to screen edges for smaller browsers 🙂

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,183 @@
...
-.mdc-snackbar--action-on-bottom .mdc-snackbar__action-wrapper {
-  @include mdc-rtl-reflexive-box(margin, left, auto);
+.mdc-snackbar__label {
+  @include mdc-typography($mdc-snackbar-label-type-scale);
 
-  flex-direction: column;
-  justify-content: flex-start;
-  margin-top: -12px;
-  margin-bottom: 8px;
+  display: none;
```
abhiomkar: Do we still need to set this to `display: none` when the root has `aria-hidden="true"`?

test/screenshot/spec/mdc-snackbar/classes/wide.html
```diff
@@ -0,0 +1,77 @@
...
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Baseline Snackbar with Action - MDC Web Screenshot Test</title>
```
kfranqueiro: Update title?

test/screenshot/spec/mdc-snackbar/fixture.scss
```diff
@@ -20,9 +20,36 @@
...
 
-@import "./common";
-@import "../packages/mdc-button/mdc-button";
-@import "../packages/mdc-checkbox/mdc-checkbox";
-@import "../packages/mdc-form-field/mdc-form-field";
-@import "../packages/mdc-snackbar/mdc-snackbar";
-@import "../packages/mdc-textfield/mdc-text-field";
+@import "../../../../packages/mdc-snackbar/mixins";
+@import "../../../../packages/mdc-theme/color-palette";
+@import "../mixins";
```
kfranqueiro: We don't seem to be using this here. Should we remove it?

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,183 @@
 //
 
 @import "@material/animation/functions";
+@import "@material/button/mixins";
+@import "@material/icon-button/mixins";
+@import "@material/ripple/mixins";
 @import "@material/rtl/mixins";
+@import "@material/theme/functions";
```
kfranqueiro: Are all of these theme imports used? I don't see mdc-theme anywhere in this file. I think they're used in `_mixins` instead.

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,183 @@
...
-    transform: translate(0, 100%);
-  }
+  // Mobile Safari only supports `rgba` values for this property; named values like `transparent` are ignored.
+  // From Apple's docs:
+  // > This property obeys the alpha value, if specified.
+  // > If you don’t specify an alpha value, Safari on iOS applies a default alpha value to the color.
+  // > To disable tap highlighting, set the alpha value to 0 (invisible).
+  // > If you set the alpha value to 1.0 (opaque), the element is not visible when tapped.
+  // See https://github.com/ben-eb/postcss-colormin/issues/1
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```
kfranqueiro: Why is this needed on the entire snackbar element?

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,183 @@
...
-  @include mdc-rtl-reflexive-box(margin, right, auto, ".mdc-snackbar");
+.mdc-snackbar--open,
+.mdc-snackbar--closing {
+  transform: none;
+}
 
+.mdc-snackbar__surface {
   display: flex;
+
+  // Non-static positioning is required to make `overflow: hidden` work in Safari.
```
kfranqueiro: Are we really running into this bug? Wouldn't we want relative positioning anyway because the overlay element is absolutely-positioned inside the surface element?

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,183 @@
...
+  .mdc-snackbar--open & {
+    transform: scale(1);
+    transition:
+      mdc-animation-enter(opacity, $mdc-snackbar-enter-duration),
+      mdc-animation-enter(transform, $mdc-snackbar-enter-duration);
+    opacity: 1;
+    pointer-events: auto; // Allow mouse events on surface element while snackbar is open
+  }
+
+  .mdc-snackbar--closing & {
```
kfranqueiro: Can you explain why there's a --closing but no --opening? It's probably hard to surmise without having the JS APIs to actually cause animation.

packages/mdc-snackbar/mdc-snackbar.scss
acdvorak: Yes, `display: none` is necessary to prevent the elements from being focusable when the user presses the <kbd>Tab</kbd> key.

packages/mdc-snackbar/mdc-snackbar.scss
```diff
@@ -21,152 +21,183 @@
...
+.mdc-snackbar__action-icon {
+  @include mdc-icon-button-ink-color($mdc-snackbar-action-icon-ink-color);
 }
 
-// stylelint-disable plugin/selector-bem-pattern
-.mdc-snackbar--multiline.mdc-snackbar--action-on-bottom .mdc-snackbar__text {
-  margin: 0;
+// Two selectors are needed to increase specificity above `.material-icons`.
+// stylelint-disable-next-line selector-class-pattern
+.mdc-snackbar__action-icon.material-icons {
```
kfranqueiro: Should this really be using material-icons in the selector? What would happen if someone used an image or svg? Should we do the same-class-twice trick instead?

packages/mdc-snackbar/_mixins.scss
acdvorak: Individual components can technically have their own independent "surface" color, which may or may not be the same as the global surface color.

abhiomkar: I checked with designer today, Snackbar doesn't seem to allow customization for snackbar container (surface).

packages/mdc-snackbar/_mixins.scss
```diff
@@ -0,0 +1,87 @@
...
+@mixin mdc-snackbar-z-index($z-index) {
+  z-index: $z-index;
+}
+
+@mixin mdc-snackbar-viewport-margin($margin) {
+  margin: $margin;
+}
+
+@mixin mdc-snackbar-min-width($min-width) {
+  .mdc-snackbar__surface {
```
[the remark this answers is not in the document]
acdvorak: If possible, yes, but if we have to choose between supporting CSS variables and having a simple Sass API, we normally go with a simple Sass API.

I'll remove the `__overlay` element and simplify the `fill-color` mixin so that it only takes 1 color parameter. The `mix()` function will be called in `_variables.scss`, similar to what mdc-chips does.

test/screenshot/spec/mdc-snackbar/fixture.scss
acdvorak: Removed

packages/mdc-snackbar/mdc-snackbar.scss
acdvorak: Removed

packages/mdc-snackbar/mdc-snackbar.scss
acdvorak: For some reason, iOS Safari displays a tap highlight on the entire snackbar element 🤷‍♀️

packages/mdc-snackbar/mdc-snackbar.scss
acdvorak: Opening and closing use different animations, so I need a separate class for `--closing`.
The opening animation is defined in `--open`.

Issue #3666, opened 2018-09-27
Checkbox: Move property change hooks from foundation to component


bonniezhou:
Move `installPropertyChangeHooks_`/`uninstallPropertyChangeHooks_` from the foundation to the component, and add a `setChecked` method to the foundation (like Switch).

When a user sets a Checkbox's checked state, the [native control is updated](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/index.js#L117) and [a change hook is called](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/foundation.js#L144) to update the styling. Compare this to Switch, where [a single foundation method](https://github.com/material-components/material-components-web/blob/master/packages/mdc-switch/foundation.js#L58) updates the native control through the adapter and also updates the styling.

The change hook is useful as a fallback in case the `checked` attribute is directly set on the DOM element. This is very unlikely for users using a framework, so change hooks should be moved from the foundation to the vanilla JS component, and replaced with a `setChecked` foundation method that explicitly updates both the native control and styling.

kfranqueiro:
#4018 has additional information on how the presence of these hooks in the foundation seriously impacted the RMWC wrapper when it was updated to wrap the foundation directly.

PR #4073, opened 2018-11-10
fix(checkbox): remove adapter.getNativeCb and move property hooks to component

moog16:
fixes https://github.com/material-components/material-components-web/issues/3666


test/unit/mdc-checkbox/mdc-checkbox.test.js
```diff
@@ -263,3 +257,16 @@ test('#adapter.setNativeControlDisabled returns false when checkbox is not disab
   component.getDefaultFoundation().adapter_.setNativeControlDisabled(false);
   assert.isFalse(cb.disabled);
 });
+
+
+test('"checked" property change hook works correctly', () => {
```
kfranqueiro: These tests aren't actually testing the property change logic whatsoever. These would pass even if the hooks were never installed.

I'm working on fixing this.

PR #4082, opened 2018-11-13
fix(fab): Separate mixins for regular FAB and Extended FAB

### LINKED ISSUE #3989: Add support for customizing FAB dimensions
With the new shape system for FAB, the radius that used to be applied as a percentage is now converted to a pixel value.

When the height of a FAB grows to something larger than the default, the pixel border radius no longer represents a proper circle and looks wrong.

*Potential fixes*
- add the ability to specify height and width in the `mdc-fab-shape-radius` mixin
- don't convert 50% border radius to a pixel value and just leave it as 50%

abhiomkar:
BREAKING CHANGE: Fab now has 2 separate mixins - `mdc-fab-shape-radius` for regular & mini Fab variants & `mdc-fab-extended-shape-radius` for Extended FAB variant. 

Before:
```
mdc-fab-shape-radius(50%) => border-radius: 28px;
```

After:
```
mdc-fab-shape-radius(50%) => border-radius: 50%;
```

Where, `mdc-fab-shape-radius` doesn't resolve percentage to absolute value.

- `mdc-fab-shape-radius` sets rounded shape to regular & mini FAB.
- `mdc-fab-extended-shape-radius`sets rounded shape to only Extended FAB.

Fixes #3989

### DIFF
--- a/packages/mdc-fab/_mixins.scss
+++ b/packages/mdc-fab/_mixins.scss
@@ -89,17 +89,13 @@
 }
 
 @mixin mdc-fab-shape-radius($radius, $rtl-reflexive: false) {
-  @include mdc-shape-radius($radius, $rtl-reflexive);
-
-  &.mdc-fab--mini {
+  &:not(.mdc-fab--extended) {
     @include mdc-shape-radius($radius, $rtl-reflexive);
   }
 }
 
 @mixin mdc-fab-extended-shape-radius($radius, $rtl-reflexive: false) {
-  &.mdc-fab--extended {
-    @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-fab-extended-height, $radius), $rtl-reflexive);
-  }
+  @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-fab-extended-height, $radius), $rtl-reflexive);
 }
 
 $mdc-fab-icon-enter-delay_: 90ms;
--- a/packages/mdc-fab/_mixins.scss
+++ b/packages/mdc-fab/_mixins.scss
@@ -89,12 +89,14 @@
 }
 
 @mixin mdc-fab-shape-radius($radius, $rtl-reflexive: false) {
-  @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-fab-height, $radius), $rtl-reflexive);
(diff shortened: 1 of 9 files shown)


packages/mdc-fab/_mixins.scss
```diff
@@ -89,12 +89,14 @@
...
-  @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-fab-height, $radius), $rtl-reflexive);
+  @include mdc-shape-radius($radius, $rtl-reflexive);
 
   &.mdc-fab--mini {
-    @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-fab-mini-height, $radius), $rtl-reflexive);
+    @include mdc-shape-radius($radius, $rtl-reflexive);
   }
+}
 
+@mixin mdc-fab-extended-shape-radius($radius, $rtl-reflexive: false) {
```
kfranqueiro: Is the `&.mdc-fab--extended` selector now redundant since this is only expected to be used within that context?

test/screenshot/spec/mdc-fab/mixins/extended-shape-radius.html
```diff
@@ -0,0 +1,73 @@
...
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>extended-padding FAB Mixin - MDC Web Screenshot Test</title>
```
kfranqueiro: Update titles in new screenshot test pages

packages/mdc-fab/_mixins.scss
abhiomkar: Makes sense.

Also, added not() selector for regular/mini fab variant so extended variant don't get extra styles.

PR #4084, opened 2018-11-13
fix(text-field): Send client position to line ripple for touch events

### LINKED ISSUE #3643: Text Field touch event does not contain clientX or clientY
related https://github.com/material-components/material-components-web-react/issues/275

The touchstart event listener in text field's foundation.js is bound to this callback function:

```
this.setPointerXOffset_ = (evt) => this.setTransformOrigin(evt);

  setTransformOrigin(evt) {
    const targetClientRect = evt.target.getBoundingClientRect();
    const evtCoords = {x: evt.clientX, y: evt.clientY};
    const normalizedX = evtCoords.x - targetClientRect.left;
    this.adapter_.setLineRippleTransformOrigin(normalizedX);
  }

```

When the event is `mousedown`, the event has `clientX` and `clientY`. When the event is a `touchstart`, `clientX` and `clientY` are missing. `clientX` and `clientY` are instead found within `evt.touches` property.

**Reproduce**
1. put debugger at beginning of `setTransformOrigin` method
1. start text field demo
1. enter chrome into mobile mode with dev tools
1. click one of the text field's

Result:
You'll notice that `evt.clientX` or `evt.clientY` does not exist.

abhiomkar:
Fixes #3643

### DIFF
--- a/test/unit/mdc-textfield/foundation.test.js
+++ b/test/unit/mdc-textfield/foundation.test.js
@@ -772,14 +772,16 @@ test('mousedown on the input sets the line ripple origin', () => {
 
 test('touchstart on the input sets the line ripple origin', () => {
   const {foundation, mockAdapter} = setupTest();
-  const mockEvt = {
-    target: {
-      getBoundingClientRect: () => {
-        return {};
+  const mockTouchStartEvent = {
+    touches: [{
+      target: {
+        getBoundingClientRect: () => {
+          return {};
+        },
       },
-    },
-    clientX: 200,
-    clientY: 200,
+      clientX: 200,
+      clientY: 200,
+    }],
   };
 
   let clickHandler;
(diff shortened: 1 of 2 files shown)


test/unit/mdc-textfield/foundation.test.js
```diff
@@ -772,14 +772,16 @@ test('mousedown on the input sets the line ripple origin', () => {
 
 test('touchstart on the input sets the line ripple origin', () => {
   const {foundation, mockAdapter} = setupTest();
-  const mockEvt = {
-    target: {
-      getBoundingClientRect: () => {
-        return {};
+  const mockTouchStartEvent = {
```
kfranqueiro: Should we update the `verify` call to expect a specific value passed to `setLineRippleTransformOrigin` in order to better test this fix?

abhiomkar: Done.

PR #4086, opened 2018-11-14
chore: Remove unnecessary package-lock

kfranqueiro:
This was added by accident in #4082 after I approved it.


kfranqueiro:
Dialog and Drawer rely on third-party dependencies (focus-trap and tabbable), which is why the package-locks exist in those folders.

The rest of the packages only have internal dependencies which are managed via lerna during development.

PR #4089, opened 2018-11-15
fix(select): Remove style customization for native select > option

### LINKED ISSUE #4088: mdc-select default browser styling - Edge
0.41.0

We got an issue in Edge, but the styling is set for all browsers.

### What OS are you using?
Windows 10

Open the mdc-select in Edge and select an value. Reopen the select and you will see that the selected value can't be seen because the background color and text color are white.

I've created an codepen based on the provided template:
[https://codepen.io/anon/pen/NEjeYw](https://codepen.io/anon/pen/NEjeYw)

The selected option should be readable.

Option background-color & color are the same color.

/

abhiomkar:
Edge browser allows customizing only the background color of `<option>` element but not the text color.

[example on jsfiddle](http://jsfiddle.net/sL3boua4/16).

This fix removes any styles applied to native `<option>` element.

Fixes #4088


packages/mdc-select/_mixins.scss
```diff
@@ -272,7 +272,7 @@
     }
 
     > option {
-      @include mdc-theme-prop(background-color, surface);
+      @include mdc-theme-prop(background-color, surface, $edgeOptOut: true);
```
abhiomkar: @williamernest Do you know why we're applying styles to native elements here? select / option native elements doesn't allow style customization in most of the browser as far as I know.

williamernest:
In browsers that do allow customizations our native select needs to use a
different background color.

PR #4090, opened 2018-11-16
feat(snackbar): Accessibility: Announce label text to screen readers

### LINKED ISSUE #4063: Snackbar: Accessibility
### In scope

- [x] Announce label text to screen reader users
    - [x] Use a "polite" live region to avoid interrupting the user's work
- [x] Hide from accessibility tree while closed
    - [x] Determine whether to use `display: none` or `aria-hidden="true"`
- [x] Prevent focus while closed
- [x] Dedicated "X" icon to dismiss
    If a snackbar gets "stuck", the user needs a way to manually dismiss it. See #3981.

### Out of scope

* Indefinite timeouts
* Disabling timeouts for assistive technology users ([not possible on the Web](https://www.nomensa.com/blog/2011/how-do-you-detect-screen-reader))
* Exposing action buttons to assistive technology users
    - See [ARIA `role="alert"` documentation](https://www.w3.org/TR/wai-aria-1.1/#alert) and [WCAG 2.0 — SC 2.2.3 — No Timing](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-no-exceptions.html)
* Focus management
    - See [ARIA `role="alert"` best practices](https://www.w3.org/TR/wai-aria-practices/#alert) for rationale
* Keyboard shortcuts (mnemonics) to focus or dismiss the snackbar
* Gestures (e.g., swipe down to dismiss)

acdvorak:
Fixes #4063

### DIFF
--- a/packages/mdc-snackbar/util.js
+++ b/packages/mdc-snackbar/util.js
@@ -30,7 +30,7 @@ const {ARIA_LIVE_LABEL_TEXT_ATTR} = strings;
  * @param {!HTMLElement} ariaEl
  * @param {!HTMLElement=} labelEl
  */
-export function announce(ariaEl, labelEl = ariaEl) {
+function announce(ariaEl, labelEl = ariaEl) {
   const priority = ariaEl.getAttribute('aria-live');
   const labelText = labelEl.textContent.trim(); // Ignore `&nbsp;` (see below)
   if (!labelText) {
@@ -86,3 +86,5 @@ export function announce(ariaEl, labelEl = ariaEl) {
     labelEl.textContent = labelText;
   }, ARIA_LIVE_DELAY_MS);
 }
+
+export {announce};
--- a/packages/mdc-snackbar/util.js
+++ b/packages/mdc-snackbar/util.js
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright 2018 Google Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+import {numbers, strings} from './constants';
+
+const {ARIA_LIVE_DELAY_MS} = numbers;
+const {ARIA_LIVE_LABEL_TEXT_ATTR} = strings;
+
+export function announce(rootEl, labelEl) {
+  const priority = rootEl.getAttribute('aria-live');
+  const labelText = labelEl.textContent.trim(); // Ignore `&nbsp;` (see below)
+  if (!labelText) {
+    return;
+  }
+
+  // Temporarily disable `aria-live` to prevent JAWS+Firefox from announcing the message twice.
+  rootEl.setAttribute('aria-live', 'off');
+
+  // Temporarily clear `textContent` to force a DOM mutation event that will be detected by screen readers.
+  // `aria-live` elements are only announced when the element's `textContent` *changes*, so snackbars
+  // sent to the browser in the initial HTML response won't be read unless we clear the element's `textContent` first.
+  // Similarly, displaying the same snackbar message twice in a row doesn't trigger a DOM mutation event,
+  // so screen readers won't announce the second message unless we first clear `textContent`.
+  //
+  // We have to clear the label text two different ways to make it work in all browsers and screen readers:
+  //
+  //   1. `textContent = ''` is required for IE11 + JAWS
+  //   2. `innerHTML = '&nbsp;'` is required for Chrome + JAWS and NVDA
+  //
+  // All other browser/screen reader combinations support both methods.
+  //
+  // The wrapper `<span>` visually hides the space character so that it doesn't cause jank when added/removed.
+  // N.B.: Setting `position: absolute`, `opacity: 0`, or `height: 0` prevents Chrome from detecting the DOM change.
+  //
+  // This technique has been tested in:
+  //
+  //   * JAWS 18.0 & 2019:
+  //       - Chrome 70
+  //       - Firefox 60 (ESR)
+  //       - IE 11
+  //   * NVDA 2017 & 2018:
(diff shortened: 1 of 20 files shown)


packages/mdc-snackbar/util.js
```diff
@@ -0,0 +1,88 @@
...
+  labelEl.textContent = '';
+  labelEl.innerHTML = '<span style="display: inline-block; width: 0; height: 1px;">&nbsp;</span>';
+
+  // Prevent visual jank by temporarily displaying the label text in the ::before pseudo-element.
+  // CSS generated content is normally announced by screen readers
+  // (except in IE 11; see https://tink.uk/accessibility-support-for-css-generated-content/);
+  // however, `aria-live` is turned off, so this DOM update will be ignored by screen readers.
+  labelEl.setAttribute(ARIA_LIVE_LABEL_TEXT_ATTR, labelText);
+
+  setTimeout(() => {
```
kfranqueiro: I'm a little wary that we're not storing / returning the handle on this timeout, but maybe I'm being paranoid. Is there a chance of the snackbar being dismissed before this timeout even elapses and causing weird behavior? Or will it simply not announce because the snackbar will become display: none?

test/screenshot/spec/mdc-snackbar/classes/wide.html
```diff
@@ -50,7 +50,6 @@
            id="test-snackbar--wide"
            role="status"
            aria-live="polite"
-           aria-hidden="true"
```
kfranqueiro: You've been moving the role/aria-live attributes to the label but you didn't do that here?

(I'm actually curious why we want these attributes only on the label - does that mean the action/dismiss buttons won't be announced? How will a user relying on screen readers know that they exist?)

packages/mdc-snackbar/util.js
```diff
@@ -0,0 +1,88 @@
...
+import {numbers, strings} from './constants';
+
+const {ARIA_LIVE_DELAY_MS} = numbers;
+const {ARIA_LIVE_LABEL_TEXT_ATTR} = strings;
+
+/**
+ * @param {!HTMLElement} ariaEl
+ * @param {!HTMLElement=} labelEl
+ */
+export function announce(ariaEl, labelEl = ariaEl) {
```
kfranqueiro: What's the purpose of labelEl being a separate parameter? We don't seem to be using it that way.

test/screenshot/spec/mdc-snackbar/classes/baseline-with-action.html
```diff
@@ -47,12 +47,11 @@
...
-           id="test-snackbar--with-action"
-           role="status"
-           aria-live="polite"
-           aria-hidden="true">
+           id="test-snackbar--with-action">
         <div class="mdc-snackbar__surface">
-          <div class="mdc-snackbar__label">Can't send photo. Retry in 5 seconds.</div>
+          <div class="mdc-snackbar__label"
+               role="status"
+               aria-live="polite">Can't send photo. Retry in 5 seconds.</div>
```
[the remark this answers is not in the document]
acdvorak: Great question!

Yes, I think we want to keep it here in the markup so that clients can easily change it to `assertive` if necessary (some screen readers don't announce the text unless it's `assertive`).

I experimented with having `announce()` take `message` and `priority` args, but I wasn't able to make it work in practice. Screen readers are _really_ picky about where you put the ARIA attributes and do the DOM manipulation, and sometimes that needs to happen on different elements.

I also tried doing something like [`goog.a11y.aria.Announcer#getLiveRegion_()`](https://chromium.googlesource.com/chromium/src/third_party/+/master/google_input_tools/third_party/closure_library/closure/goog/a11y/aria/announcer.js?autodive=0%2F#101), dynamically inserting an anonymous element into the `<body>` using JS. But unfortunately some screen reader/browser combinations refuse to announce the text when I do that. Others needed a much bigger timeout value between clearing and resetting the text.

I've arrived at the present solution after many hours of painful experimentation. It seems to work reliably in 6 of the most commonly-used browser/screen reader combinations: JAWS and NVDA in Chrome, IE 11 💩, and Firefox.

It doesn't seem to work with Safari + VoiceOver on macOS 10.13 High Sierra, unfortunately.
I'll document that limitation in the README (next PR).

packages/mdc-snackbar/util.js
acdvorak: If it's `display: none` when the timeout fires, the browsers I tested (Chrome/IE11/Firefox in NVDA/JAWS) don't announce it.

The timeout is responsible for resetting the `textContent` back to its original value, so I'm not sure we'd want to clear it. But there could be a use case I'm not thinking of.

I made sure to test for non-empty `textContent` at the top of the function, so if you call it twice in a row, the second call is a noop.

From my testing, Windows screen readers queue the message and read it fully as long as the `textContent` was reset to its original value _before_ the snackbar was set to `display: none`.

test/screenshot/spec/mdc-snackbar/classes/wide.html
acdvorak: Thanks, fixed!

You are correct, the action/dismiss buttons won't be announced.

According to [WCAG 2.0 — SC 2.2.3 — No Timing](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-no-exceptions.html), UI controls should not be timing-dependent. We don't want to announce a button that won't be there by the time the user navigates to it.

The snackbar a11y spec also says that apps should provide a redundant means of achieving the same action elsewhere in the app.

So for AT users, this is simply a passive notification.

packages/mdc-snackbar/util.js
acdvorak: Every time the DOM or CSS changes (e.g., adding `display: none`), it breaks in at least one AT/browser combination. I've gone back and forth between using separate elements or the same element, depending on the moods of the Screen Reader Gods.

For example, if we decided to announce the action button to screen reader users (which I advise against—see my comment above), we would need to move the ARIA attributes back to the root element and add `aria-atomic="true"`. Then they would be two separate elements again.

I'm mainly trying to make future changes easier.

test/screenshot/spec/mdc-snackbar/classes/baseline-with-action.html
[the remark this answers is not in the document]
kfranqueiro: In my testing it seemed to work with assertive rather than polite, but I can re-test when we get to the readme.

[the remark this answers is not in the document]
abhiomkar: Thanks for explaining! The announce adapter method seem to be coupled with what util.announce() method does which has its own business logic. It has to follow certain sequence of DOM operations to implement it.

For example, some wrappers have their own announcer utilities which makes it hard for them to reuse it or do they've to reimplement this util function?

acdvorak:
Added unit tests for `announce()`. Snackbar has 100% code coverage.

test/screenshot/spec/mdc-snackbar/classes/baseline-with-action.html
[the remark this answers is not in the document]
acdvorak: I expect internal teams to use [`goog.a11y.aria.Announcer#say()`](https://chromium.googlesource.com/chromium/src/third_party/+/master/google_input_tools/third_party/closure_library/closure/goog/a11y/aria/announcer.js?autodive=0%2F#80), and all other wrappers to use `util.announce()`.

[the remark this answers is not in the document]
abhiomkar: As discussed offline, as long as if the screen reader doesn't end up start pronouncing it twice (or jank) because of `aria-live` on child element and from announcer service, I guess this works for everyone.

test/unit/mdc-snackbar/mdc-snackbar.test.js
```diff
@@ -66,8 +66,9 @@ function setupTest(fixture = getFixture()) {
   const actions = fixture.querySelector('.mdc-snackbar__actions');
   const actionButton = fixture.querySelector('.mdc-snackbar__action-button');
   const actionIcon = fixture.querySelector('.mdc-snackbar__action-icon');
-  const component = new MDCSnackbar(root);
-  return {component, root, surface, label, actions, actionButton, actionIcon};
+  const announce = td.func('announce');
```
abhiomkar: Should `aria-live` attribute on the label element in your test setup?

PR #4106, opened 2018-11-21
fix(top-app-bar): Move scroll target initialization; improve test

kfranqueiro:
This resolves the test failure discovered in #4046.

The `getViewportScrollY` adapter unit test was coincidentally passing because the page's scroll offset and the scroll offset of the content element were typically both 0. This only ended up not being the case when I tried running our unit tests via headless Firefox on Travis.

The test as written actually should never have passed, because `getDefaultFoundation` in the Top App Bar component had the side effect of resetting the scroll target, effectively undoing the test's setup.

This fixes the component to initialize the scroll target in `initialize` rather than `getDefaultFoundation`, and fixes the test to use a mock object to guarantee a distinguishable expected value.

### DIFF
--- a/test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
+++ b/test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
@@ -290,10 +290,10 @@ test('adapter#getViewportScrollY returns scroll distance', () => {
 });
 
 test('adapter#getViewportScrollY returns scroll distance when scrollTarget_ is not window', () => {
-  const {component, fixture} = setupTest();
-  const content = {addEventListener: () => {}, scrollTop: 20};
-  component.setScrollTarget(content);
-  assert.equal(component.getDefaultFoundation().adapter_.getViewportScrollY(), content.scrollTop);
+  const {component} = setupTest();
+  const mockContent = {addEventListener: () => {}, scrollTop: 20};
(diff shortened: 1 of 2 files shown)


test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
```diff
@@ -292,8 +291,8 @@ test('adapter#getViewportScrollY returns scroll distance', () => {
 
 test('adapter#getViewportScrollY returns scroll distance when scrollTarget_ is not window', () => {
   const {component, fixture} = setupTest();
-  const content = fixture.querySelector('.content');
-  component.scrollTarget_ = content;
+  const content = {addEventListener: () => {}, scrollTop: 20};
```
abhiomkar: nit: mockContent*

test/unit/mdc-top-app-bar/mdc-top-app-bar.test.js
```diff
@@ -292,8 +291,8 @@ test('adapter#getViewportScrollY returns scroll distance', () => {
 
 test('adapter#getViewportScrollY returns scroll distance when scrollTarget_ is not window', () => {
   const {component, fixture} = setupTest();
```
abhiomkar: Remove unused variable `fixture`.

PR #4116, opened 2018-11-27
fix(shape): Add noflip comments, fix RTL for categories

kfranqueiro:
This has 3 changes/fixes:

* Don't bother emitting separate RTL styles when we can easily tell it's identical to LTR
* Add `@noflip` annotations to both the LTR and RTL emitted styles when both are emitted
* Fix RTL emitted styles to resolve shape category names (this caused a bug in drawer's RTL styles previously)

### DIFF
--- a/packages/mdc-shape/_mixins.scss
+++ b/packages/mdc-shape/_mixins.scss
@@ -24,11 +24,19 @@
 @import "./functions";
 
 @mixin mdc-shape-radius($radius, $rtl-reflexive: false) {
+  // Even if $rtl-reflexive is true, only emit RTL styles if we can't easily tell that the given radius is symmetrical
+  $needs-flip: $rtl-reflexive and length($radius) > 1;
+
+  @if ($needs-flip) {
+    /* @noflip */
+  }
+
   border-radius: mdc-shape-prop-value($radius);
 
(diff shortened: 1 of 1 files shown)


packages/mdc-shape/_mixins.scss
```diff
@@ -24,11 +24,19 @@
 @import "./functions";
 
 @mixin mdc-shape-radius($radius, $rtl-reflexive: false) {
+  // Even if $rtl-reflexive is true, only emit RTL styles if we can't easily tell that the given radius is symmetrical
+  $needs-flip: $rtl-reflexive and length($radius) > 1;
+
+  @if ($needs-flip) {
+    /* @noflip */
+  }
+
```
abhiomkar: is annotation with new line acceptable?

this generates

```
/* @noflip */

border-radius: mdc-shape-prop-value($radius);
```

instead of

```
/* @noflip */
border-radius: mdc-shape-prop-value($radius);
```

kfranqueiro: I was wary of that too, but when I looked at the diff between the all-in-one CSS before and after, I didn't see any empty newlines generated by these changes.

PR #4124, opened 2018-11-28
fix(list): Accept array of index for selectedIndex API

### LINKED ISSUE #3611: MDCList: Initially focused item still has tab stop while moving focus within list
1. `npm run dev` + load demos/list.html, or `npm start` and load spec/mdc-dialog/classes/baseline-confirmation.html
2. Focus the list (the list will already be focused in the latter example)
3. Arrow down to the next list item
4. Shift+tab

Expected behavior: Focus moves to the previous element w/ tab stop outside of the list

Actual behavior: Focus moves back to the initially focused list item

This affects both single-selection list (1st example) and not (2nd example).

### LINKED ISSUE #4122: MDC List: Accept multiples indexes for selectedIndex for checkbox based list.
MDC List component has `selectedIndex` setter which updates the selection in a list, this is a proxy to foundation's `setSelectedIndex(index)` method. But this only works for single selection list and radio list. Checkbox based list is a multi-selectable list and it needs to accept multiple indexes to update selection.

Also need to add getter method for `selectedIndex` in MDC List component. `selectedIndex`: that accepts and return value type `number|!Array<number>`.

abhiomkar:
Before this implementation `selectedIndex` component API (that proxies foundation's `setSelectedIndex`) use to accept only single number index. This change adds support for multiple indexes that sets the state of checkboxes.

**New behaviour for checkbox based list**

```
[ ] option 1
[ ] option 2
[ ] option 3
[ ] option 4

selectedIndex = [1, 3];

[ ] option 1
[x] option 2
[ ] option 3
[x] option 4

selectedIndex = [2, 3];

[ ] option 1
[ ] option 2
[x] option 3
[x] option 4

selectedIndex = [];

[ ] option 1
[ ] option 2
[ ] option 3
[ ] option 4

selectedIndex = 2;

// Throws error

```

**Other changes:**

- getter API for `selectedIndex` that returns the current state of selection for all variants.
- Tabindex is set to `0` for currently focused element and will be reverted to first selected element on focus out of list root.
- new adapter method `isFocusInsideList`.

BREAKING CHANGE: Introduced new adapter `isFocusInsideList` for MDC List for improved accessibility.

Fixes #4122 
Fixes #3611


packages/mdc-list/foundation.js
```diff
@@ -371,6 +412,38 @@ class MDCListFoundation extends MDCFoundation {
...
+   * @return {boolean}
+   */
+  isIndexValid_(index) {
+    const listSize = this.adapter_.getListItemCount();
+
+    if (index instanceof Array) {
+      if (!this.hasCheckboxAtIndex_(index)) {
+        throw new Error('MDCListFoundation: Array of index is only supported for checkbox based list');
+      }
+      return index.some((i) => this.isIndexInRange_(i));
```
kfranqueiro: Should this be .every instead of .some? What happens if one index is in range but another isn't?

packages/mdc-list/foundation.js
```diff
@@ -371,6 +412,38 @@ class MDCListFoundation extends MDCFoundation {
...
+    } else {
+      return index >= 0 && index < listSize;
```
kfranqueiro: Should we be checking that there is _not_ a checkbox at the index here?

Currently if you pass a single number (not an array) in the case of a checkbox list, it checks the indicated index _in addition_ to what is already checked... but then `list.selectedIndex` reports just that index, not all of the selected indices. I think it'd be safer to prevent this behavior (and insist that set/get are symmetrical) than try to solve for it.

I suspect this logic is also being depended upon for clicks/keydowns. Clearly if I interact with a checkbox list to select an item, `selectedIndex` then reports only that item regardless of what else is already selected, which matches the behavior I'm describing here. However, I can also uncheck via list item interaction, in which case `selectedIndex` _still_ reports the last interacted item - even though it's now _deselected_.

packages/mdc-list/foundation.js
```diff
@@ -371,6 +412,38 @@ class MDCListFoundation extends MDCFoundation {
...
+   * @param {number} index
+   */
+  isIndexInRange_(index) {
+    const listSize = this.adapter_.getListItemCount();
+
+    if (index >= 0 && index < listSize) {
+      return true;
+    } else {
+      const lastIndex = listSize - 1;
+      throw new Error(`MDCListFoundation: Index is out of range.
```
kfranqueiro: Do we really want to be throwing an error? This method actually never returns false - it either returns true or throws.

Previously, IIUC, we would simply ignore an out-of-bounds index?

packages/mdc-list/foundation.js
kfranqueiro: Could this also use `this.isIndexInRange_`? (See also my comment in that method - currently this code path will never throw an error, but that code path will.)

packages/mdc-list/foundation.js
```diff
@@ -110,83 +120,27 @@ class MDCListFoundation extends MDCFoundation {
     this.useActivatedClass_ = useActivated;
   }
 
-  /** @param {number} index */
+  /** @return {!Index} */
+  getSelectedIndex() {
+    return this.selectedIndex_;
```
kfranqueiro: This is currently out-of-sync on init if a list already has a selected or checked item(s), because we initialize this property to -1 statically.

abhiomkar: Yes, that is true. This happens with current implementation as well. We should move pre-selection logic to foundation. I'll create a new issue for this.

packages/mdc-list/foundation.js
abhiomkar: I removed the error throwing when index is not in range, so it should ignore them when out of range and select rest.

packages/mdc-list/foundation.js
kfranqueiro: There's no issue in the current implementation because there's no getter, right?

packages/mdc-list/foundation.js
abhiomkar: `selectedIndex` is now synced with keyboard / mouse interacted selected state change. Thanks for spotting this out.

Also, Checkbox based list only accepts array of indexes.

packages/mdc-list/foundation.js
abhiomkar: Fixed list component to pre-populated selected state for singleSelection, radio and checkbox during initialization.

packages/mdc-list/foundation.js
```diff
@@ -207,6 +162,16 @@ class MDCListFoundation extends MDCFoundation {
     if (listItemIndex >= 0) {
       this.adapter_.setTabIndexForListItemChildren(listItemIndex, -1);
     }
+
+    /**
+     * Between Focusout & Focusin some browsers do not have focus on any element. Setting a delay to wait till the focus
+     * is moved to next element.
+     */
+    setTimeout(() => {
```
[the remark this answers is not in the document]
abhiomkar: I found a bug where checkbox based list end up having multiple tab stops.

Steps to reproduce:

1. Add `tabindex=0` to selected checkbox item.
2. programmatically set selected index. (Shouldn't change the tabindex state)
3. Check other list items. Since there is no trace of list item that had `tabindex=0` during render, it ends up adding second tab stop on last selected / checked list item.

packages/mdc-list/foundation.js
```diff
@@ -110,81 +124,22 @@ class MDCListFoundation extends MDCFoundation {
...
+  }
+
+  /** @param {!Index} index */
   setSelectedIndex(index) {
-    if (index < 0 || index >= this.adapter_.getListItemCount()) return;
+    if (!this.isIndexValid_(index)) return;
 
-    if (this.adapter_.hasCheckboxAtIndex(index)) {
-      this.setAriaAttributesForCheckbox_(index);
+    if (this.hasCheckboxAtIndex_(index)) {
```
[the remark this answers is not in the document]
abhiomkar: That's a good point. I made changes to foundation such that it only checks the first list item in the list to determine the list type so we don't need to loop through the list additionally.

packages/mdc-list/foundation.js
```diff
@@ -333,44 +309,197 @@ class MDCListFoundation extends MDCFoundation {
...
+        }
+
+        this.adapter_.setCheckedCheckboxOrRadioAtIndex(i, isChecked);
+        this.adapter_.setAttributeForElementIndex(i, strings.ARIA_CHECKED, isChecked ? 'true' : 'false');
+      }
+
+      this.selectedIndex_ = index;
+    } else {
+      let isChecked = this.adapter_.isCheckboxCheckedAtIndex(index);
+      if (this.toggleCheckbox_) {
```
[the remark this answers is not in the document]
abhiomkar: I had the same idea of passing down the `toggleIndex` via selectedIndex private method so that it takes a different route.

But didn't want to expose optional parameter to public API that is meant for internal use. Instead, I do want to remove the need of passing `toggleIndex` and only send mouse event object similar to `handleKeydown` handler. Then the foundation can determine if it needs a toggle whether it is clicked on checkbox/radio input or on label. WDYT?

I merged above two conditions.

packages/mdc-list/foundation.js
```diff
@@ -333,44 +309,197 @@ class MDCListFoundation extends MDCFoundation {
...
+      const isFirstItemHasCheckbox = this.adapter_.getListItemCount() > 0 && this.adapter_.hasCheckboxAtIndex(0);
+      if (index.length === 0 && isFirstItemHasCheckbox) {
+        return true;
+      }
+
+      if (!this.hasCheckboxAtIndex_(index)) {
+        throw new Error('MDCListFoundation: Array of index is only supported for checkbox based list');
+      }
+      return index.some((i) => this.isIndexInRange_(i));
+    } else if (typeof index === 'number' && index !== -1) {
```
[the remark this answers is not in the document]
abhiomkar: Since we've removed `hasCheckboxAtIndex_` logic from next condition we no longer need this here.

packages/mdc-list/README.md
```diff
@@ -498,7 +501,7 @@ Method Signature | Description
 `setWrapFocus(value: Boolean) => void` | Sets the list to allow the up arrow on the first element to focus the last element of the list and vice versa.
 `setVerticalOrientation(value: Boolean) => void` | Sets the list to an orientation causing the keys used for navigation to change. `true` results in the Up/Down arrow keys being used. `false` results in the Left/Right arrow keys being used.
 `setSingleSelection(value: Boolean) => void` | Sets the list to be a selection list. Enables the `enter` and `space` keys for selecting/deselecting a list item.
-`setSelectedIndex(index: Number) => void` | Toggles the `selected` state of the list item at index `index`.
+`setSelectedIndex(index: Index) => void` | Sets the selection state to given index or list of indexes if it is checkbox based list. See [constants.js](./constants.js) for `Index` type definition.
```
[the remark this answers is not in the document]
abhiomkar: I prefer keeping this in-sync with JSDoc's param type. Linked it to `constants.js` file for reference to type declaration.

packages/mdc-list/foundation.js
```diff
@@ -333,43 +325,185 @@ class MDCListFoundation extends MDCFoundation {
...
+   */
+  setTabindexToSelectedItem_() {
+    if (this.isSelectableList_()) {
+      let targetIndex = -1;
+      if (typeof this.selectedIndex_ === 'number' && this.selectedIndex_ !== -1) {
+        targetIndex = this.selectedIndex_;
+      } else if (this.selectedIndex_ instanceof Array && this.selectedIndex_.length > 0) {
+        targetIndex = this.selectedIndex_.reduce((currentIndex, minIndex) => Math.min(currentIndex, minIndex));
+      } else {
+        targetIndex = 0;
```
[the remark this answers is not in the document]
abhiomkar: Good call. Initializing it to `0` instead.

Will add a test for this.

Regarding other comment
>  is this PR also fixing #3611?

Yes, it fixes #3611 issue.

Before: `tabindex=0` is set only to selected list item.
After: `tabindex=0` is set to currently focused list item and reverted back to first selected list item when user leaves list focus. This'll fix having double tab stops.

packages/mdc-list/foundation.js
```diff
@@ -333,43 +325,185 @@ class MDCListFoundation extends MDCFoundation {
...
+        throw new Error('MDCListFoundation: Array of index is only supported for checkbox based list');
+      }
+
+      if (index.length === 0) {
+        return true;
+      } else {
+        return index.some((i) => this.isIndexInRange_(i));
+      }
+    } else if (typeof index === 'number') {
+      if (this.programmaticSelection_ && this.isCheckboxList_) {
```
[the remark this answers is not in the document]
abhiomkar: This was required earlier when checkbox based list accepting single number input. It is no longer required. Removed.

Regarding, `programmaticSelection_` comment:
I removed the need for boolean private property and replaced it with private method named `setSelectedIndexOnAction_` which is specific for user actions that always accepts single number argument.

packages/mdc-list/foundation.js
```diff
@@ -303,6 +291,7 @@ class MDCListFoundation extends MDCFoundation {
   /**
    * Focuses the next element on the list.
    * @param {number} index
+   * @return {number}
```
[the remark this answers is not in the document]
abhiomkar: See my other comment.

packages/mdc-list/foundation.js
```diff
@@ -301,7 +296,7 @@ class MDCListFoundation extends MDCFoundation {
         nextIndex = 0;
       } else {
         // Return early because last item is already focused.
-        return;
+        return nextIndex;
```
kfranqueiro: This is incorrect and will return an out-of-bounds index. `index` is the old index, which is what should be returned, as I mentioned in my previous comment.

Same for `focusPrevElement` and `return prevIndex`.

This PR has 100% coverage but I have to assume we're not verifying these branches...

packages/mdc-list/foundation.js
```diff
@@ -505,6 +473,43 @@ class MDCListFoundation extends MDCFoundation {
...
+  }
+
+  /**
+   * @param {number} index
+   * @private
+   */
+  toggleCheckboxAtIndex_(index) {
+    let isChecked = this.adapter_.isCheckboxCheckedAtIndex(index);
+
+    if (this.toggleCheckbox_) {
```
kfranqueiro: It seems like at this point we should be able to pass this through `setSelectedIndexAction_` to here, since the invocation chain between `handleClick` and this point is private. That should be easier to follow than flipping an instance flag from 2 levels up the call stack.

packages/mdc-list/foundation.js
```diff
@@ -505,6 +473,43 @@ class MDCListFoundation extends MDCFoundation {
...
+      isChecked = !isChecked;
+      this.adapter_.setCheckedCheckboxOrRadioAtIndex(index, isChecked);
+    }
+
+    this.adapter_.setAttributeForElementIndex(index, strings.ARIA_CHECKED, isChecked ? 'true' : 'false');
+
+    if (this.selectedIndex_ === -1) {
```
kfranqueiro: When does this case happen?

packages/mdc-list/foundation.js
```diff
@@ -197,6 +194,8 @@ class MDCListFoundation extends MDCFoundation {
    * @param {number} listItemIndex
    */
   handleKeydown(evt, isRootListItem, listItemIndex) {
+    if (this.adapter_.getListItemCount() === 0) return;
```
kfranqueiro: When does this case happen? (Is it even possible for keyboard events to happen if there are no list items to focus?)

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -285,12 +314,19 @@ test('singleSelection calls foundation setSingleSelection with the provided valu
   td.verify(mockFoundation.setSingleSelection(true), {times: 1});
 });
 
-test('selectedIndex calls setSelectedIndex on foundation', () => {
+test('selectedIndex pro setSelectedIndex on foundation', () => {
```
kfranqueiro: I suspect this change was unintentional?

packages/mdc-list/foundation.js
```diff
@@ -333,43 +325,185 @@ class MDCListFoundation extends MDCFoundation {
       }
     }
     this.adapter_.focusItemAtIndex(prevIndex);
+
+    return prevIndex;
   }
 
+  /**
+   * @return {number}
```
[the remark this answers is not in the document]
kfranqueiro: In that case we have code paths that are never used, but validating at the deeper level seems like the safer choice anyway.

packages/mdc-list/foundation.js
```diff
@@ -66,16 +67,42 @@ class MDCListFoundation extends MDCFoundation {
...
+
+    /** @private {boolean} */
+    this.isCheckboxList_ = false;
+
+    /** @private {boolean} */
+    this.isRadioList_ = false;
+  }
+
+  init() {
+    if (this.adapter_.getListItemCount() === 0) return;
```
kfranqueiro: For some reason I can't respond to the original comment thread anymore (thanks GitHub?)...

> Regarding lazy list update, I've added `.init()` call to component's `layout()` method. Otherwise, `.init()` is called during initialization anyway?

I know it's called during initialization, my point is that if there are no list items during initialization it's never going to detect the list type. Adding a `foundation.init` call to the component's `layout` method seems like an unorthodox band-aid. However, looking at the existing component logic, I guess we would already have this problem with `initializeListType`, so it's probably wrong for me to bring that up in this PR.

It also seems odd that we're doing some list-type-detection logic in the component and other logic here in the foundation, which makes me wonder if we could manage to combine the two in the foundation, but that's probably enough effort to spin out into a separate task.

packages/mdc-list/foundation.js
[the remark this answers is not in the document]
abhiomkar: Yes, the foundation was already doing that. But earlier it wasn't changing the tabindex on focus change, it remains at selected item which makes it easier to receive focus back at selected item.

This change fixes #3611

packages/mdc-list/foundation.js
abhiomkar: This happens when none of the checkbox items are selected and selectedIndex is not yet initialized then we provide a default value (i.e., `[]`).

packages/mdc-list/foundation.js
abhiomkar: Removed this condition.

packages/mdc-list/foundation.js
abhiomkar: It is required for foundation to know when the list count or list type is changed. I renamed `init` to `layout` to avoid confusion and this pattern is also followed by other components when layout is changed.

Earlier foundation use to detect the list type on each event handle callback on target list item, after this change foundation does this once during initialization. It is mandatory to manually call `layout` method when list variant is changed (i.e., singlseSelection => checkbox variant - which is I think unusual case).

I had the same idea of moving component logic - `initializeListType` to foundation. But, this section of logic is replaced by props / states in other frameworks like React.

Example:

```<List selectedIndex="8" singleSelection=true />```

File: test/unit/mdc-list/foundation.test.js [hunk 20]
```diff
@@ -640,6 +632,66 @@ test('#handleKeydown bail out early if event origin doesnt have a mdc-list-item
...
+});
+
+test('#focusNextElement retains the focus on last item when wrapFocus=false and returns that index', () => {
+  const {foundation, mockAdapter} = setupTest();
+
+  td.when(mockAdapter.getListItemCount()).thenReturn(4);
+  foundation.setWrapFocus(false);
+
+  assert.equal(3, foundation.focusNextElement(3));
+  td.verify(mockAdapter.focusItemAtIndex(3), {times: 0});
```
kfranqueiro: Change 3 to `td.matchers.isA(Number)`

(I suspect either the `3` here or in the equivalent `focusPrevElement` test below is a copypasta)

test/unit/mdc-list/foundation.test.js
```diff
@@ -640,6 +632,66 @@ test('#handleKeydown bail out early if event origin doesnt have a mdc-list-item
...
+});
+
+test('#focusPrevElement retains the focus on first list item when wrapFocus=false and returns that index', () => {
+  const {foundation, mockAdapter} = setupTest();
+
+  td.when(mockAdapter.getListItemCount()).thenReturn(4);
+  foundation.setWrapFocus(false);
+
+  assert.equal(0, foundation.focusPrevElement(0));
+  td.verify(mockAdapter.focusItemAtIndex(3), {times: 0});
```
kfranqueiro: Change 3 to `td.matchers.isA(Number)`

packages/mdc-list/README.md
```diff
@@ -393,7 +395,7 @@ The MDCList JavaScript component implements the WAI-ARIA best practices for
 [Listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox). This includes overriding the default tab behavior
 within the list component.
 
-The `tabindex` should be set to `0` for first list item element or selected list item element, remaining list item elements should not have `tabindex` set.
+The first item in the list should always have `tabindex` set to `0` on render, remaining list item elements should not have `tabindex` set so that entire list will have only one tab stop.
```
kfranqueiro: Why did we change this advice (and the examples above) away from setting the tabindex on the selected item? Does the code rely on the first item having tabindex now?

packages/mdc-list/README.md
```diff
@@ -498,13 +501,13 @@ Method Signature | Description
 `setWrapFocus(value: Boolean) => void` | Sets the list to allow the up arrow on the first element to focus the last element of the list and vice versa.
 `setVerticalOrientation(value: Boolean) => void` | Sets the list to an orientation causing the keys used for navigation to change. `true` results in the Up/Down arrow keys being used. `false` results in the Left/Right arrow keys being used.
 `setSingleSelection(value: Boolean) => void` | Sets the list to be a selection list. Enables the `enter` and `space` keys for selecting/deselecting a list item.
-`setSelectedIndex(index: Number) => void` | Toggles the `selected` state of the list item at index `index`.
+`setSelectedIndex(index: Index) => void` | Sets the selection state to given index or list of indexes if it is checkbox based list. See [constants.js](./constants.js) for `Index` type definition.
```
kfranqueiro: Need to also add `getSelectedIndex` to docs

packages/mdc-list/README.md
abhiomkar: Reverted this. This is not mandatory but user can still prefer to set tabindex to first element whether it is selectable list or not.

kfranqueiro: Should we also revert the changed examples then? If someone sets tabindex="0" on the first item but there is a selected item elsewhere in the list, will the selected item not have focus? (It should...)

kfranqueiro: I think I just found the answer to this question when testing the test pages, which also have this issue. Setting tabindex="0" on the first item when something else is selected will indeed send focus to the first item when the list is first tabbed into... then when you tab out and in again, it will properly focus the first selected item. So the examples and test pages should definitely initialize tabindex on the first selected element for the best UX.

abhiomkar: I agree. The `tabindex=0` should be set to selected item or first item (if list is not selectable). All examples already have tabindex=0 set to appropriate item. Are there any other examples that I left out having incorrect tabindex?

kfranqueiro: There are 3 examples in the readme which you changed from having tabindex on the selected item to having tabindex on the first item. When you changed the docs under these examples back, you didn't also change the examples back themselves.

The checkbox, radio, and single-select screenshot test pages all have tabindex on the first item rather than on the selected item.

abhiomkar: Done.

kfranqueiro:
FYI it looks like there are some closure failures to address, and there's a seemingly spurious unit test failure in the last run... Screenshot tests should pass once this is synced with master.

PR #4141, opened 2018-12-03
docs(list): Document disabled modifier class

### LINKED ISSUE #3679: MDC Menu: Update Menu & List docs for disabled item state.
## Bugs

It seems like the possibility to disabled menu-items was lost in the 0.39.0 rebuild of mdc-menu? The documentation for how to disable a menu-item has been removed, and following the old documentation does not disable the menu-item when using a later version of MDC Web.

The changelog does not cover this as a breaking change (nor in any other way), which makes me suspect that either the removal of this feature was accidental, or the feature still exists, but the new way to access it has not been documented.

0.40.0

Chrome 69.0.3497.92 (Official Build) (64-bit)
UA-string: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.92 Safari/537.36`

### What OS are you using?

macOS 10.13.6 ("High Sierra")

> Please write the steps which need to be taken in order to reproduce the bug. These steps should be
> as detailed as possible, e.g.
>
> 1. Run the demo server
> 2. Go to localhost:8080/textfield.html
> 3. Tab-focus on the first text field
> 4. Observe the component's behavior
>
> We also have a [template on CodePen](https://codepen.io/williamernest/pen/deyxyp) which we encourage you to use to create a reproduction of
> the issue. The less time it takes for us to reproduce the issue, the less time it takes to verify and
> fix it!

I tried using the CodePen linked to above, but it didn't work correctly, even with the supplied example. And since this is more of a report of a potentially removed feature than a bug appearing in a specific circumstance, I won't put too much time into trying to reproduce it on CodePen, unless you believe that doing so would be helpful in this specific case.

There should be a way to disable specific menu-items in an otherwise enabled menu.

Following the documentation for MDCMenu <0.39.0 does not disable the menu-item when using MDCMenu >=0.39.0, and there is no new documentation on how this can be achieved in another way.

I can't think of anything right now, but just ask and I'll be happy to do what I can to deliver.

kfranqueiro:
Fixes #3679.

Adds mention of `--disabled` modifier class to List docs for CSS classes and a11y, and also mentions the modifier class in the Menu docs.

Also improves wording around selected and activated states in MDC List docs.

### DIFF
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -385,13 +385,11 @@ CSS Class | Description
 `mdc-list-divider--padded` | Optional, leaves gaps on each side of divider to match padding of `list-item__meta`.
 `mdc-list-divider--inset` | Optional, increases the leading margin of the divider so that it does not intersect the avatar column.
 
-> NOTE: `mdc-list-divider` class can be used between list items (example 1) *OR* between two lists (example 2).
+> NOTE: The `mdc-list-divider` class can be used between list items *OR* between two lists (see respective examples under [List Dividers](#list-dividers)).
 
-> NOTE: the difference between selected and activated states:
-> * *Selected* state should be implemented on the `.mdc-list-item` when it is likely to frequently change due to user choice. Eg., selecting one or more photos to share in Google Photos.
-> * *Selected* state may apply to one or multiple items in a list at the same time.
+> NOTE: In Material Design, the selected and activated states apply in different, mutually-exclusive situations:
+> * *Selected* state should be applied on the `.mdc-list-item` when it is likely to frequently change due to user choice. E.g., selecting one or more photos to share in Google Photos.
 > * *Activated* state is more permanent than selected state, and will **NOT** change soon relative to the lifetime of the page. Common examples are navigation components such as the list within a navigation drawer.
-> * *Activated* state only applies to one item in a list at a given time.
 
 ### Sass Mixins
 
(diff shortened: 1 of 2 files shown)


packages/mdc-list/README.md
```diff
@@ -368,11 +369,10 @@ CSS Class | Description
 > NOTE: `mdc-list-divider` class can be used between list items (example 1) *OR* between two lists (example 2).
 
 > NOTE: the difference between selected and activated states:
-
-* *Selected* state should be implemented on the `.mdc-list-item` when it is likely to change soon. Eg., selecting one or more photos to share in Google Photos.
-* Multiple items can be selected at the same time when using the *selected* state.
-* *Activated* state is similar to selected state, however should only be implemented once within a specific list.
-* *Activated* state is more permanent than selected state, and will **NOT** change soon relative to the lifetime of the page.
+> * *Selected* state should be implemented on the `.mdc-list-item` when it is likely to frequently change due to user choice. Eg., selecting one or more photos to share in Google Photos.
+> * *Selected* state may apply to one or multiple items in a list at the same time.
```
abhiomkar: We do not support multiple item selection for the list. Only selection mode available right now is single selection mode via `singleSelection` API.

Select & Active looks and behave similarly. Only semantic difference is "--selected" is likely to change during its life cycle, where as, `--activated` does not change until page is navigated away or refreshed.

For example, list where user needs to make a choice and may change in the same session. But activated is not changed until the next page refresh or navigation.

abhiomkar: didn't notice that our docs already had a note on multiple selection.

I do not see any business logic related to multiple selection (except in checkbox based list). We may need to fix the docs.

kfranqueiro: This was the only mention I found of multiple selection, so I think this should be sufficient for now.

Issue #2, opened 2016-12-05
Implement multi-select component


traviskaufman:
Extricated from https://github.com/google/material-design-lite/issues/4475
> Copied from https://github.com/google/material-design-lite/issues/4948

amsheehan:
Here is a first cut at multi select.

I think it should be included in the mdc-select package, since currently there is only an HTML/CSS implementation of multi-select on the roadmap. As a result, to turn a select into multi select, the user must only set the `multiple` attribute.

Thing to consider:

- Should dividers always span the width of the select box (regardless of nesting)?
- Should dividers have vertical padding?


[Codepen](https://codepen.io/amsheehan/pen/xRyNdo)
[MDC Select README changes](https://gist.github.com/amsheehan/81ec49d3f692faf5527a830f24662fc9)

traviskaufman:
@amsheehan LGTM on the whole. Only thing is on the Codepen, I can focus the list dividers. I think it needs a `disabled` attribute in addition to `role="presentation"`, since it's an option. Also, I think there's a blank final option but that's nbd.

PR #4144, opened 2018-12-05
feat(list): Add notifyAction adapter for action on list item.

abhiomkar:
`notifyAction` adapter method is introduced to notify on user action including keyboard and mouse. This could've been `notifySelected` but the a list can also be a checkbox variant where it can have multiple selections and an action could also cause unselect.

BREAKING CHANGE: Introduced new adapter method `getAttributeForElementIndex` to determine if target list item has `href` attribute and removed `followHref` adapter API.

Changes required for #3481


packages/mdc-list/foundation.js
```diff
@@ -265,7 +267,11 @@ class MDCListFoundation extends MDCFoundation {
         }
 
         // Explicitly activate links, since we're preventing default on Enter, and Space doesn't activate them.
-        this.adapter_.followHref(currentIndex);
+        if (this.adapter_.getAttributeForElementIndex(currentIndex, 'href')) {
```
kfranqueiro: Do we really need this adapter API? followHref basically already performs this check, and I'm not sure whether these two things should be exclusive (e.g. if the href is not actually causing page navigation, shouldn't it still cause `notifyAction`?)

abhiomkar: I separated this logic here since browsers trigger synthetic MouseEvent on enter key for anchor elements so that foundation doesn't end up triggering action event twice.

But on a second thought, I removed the need of `followHref` since browsers implicitly trigger `MouseEvent` which is handled by `handleClick` and programmatic `.click()` is unnecessary. For users who want to also handle `Space` key would need to make use of `MDCList:action` custom event. So the new implementation is more aligned to how browsers work with anchor element.

Our drawer is good example to test this. Where <kbd>Enter</kbd> & <kbd>Space</kbd> both triggers `MDCList:action`. Former automatically follows the href, later just triggers the custom event.

kfranqueiro: Interesting...I thought we were calling `preventDefault` on enter, which iirc is half the reason I added `followHref`. I guess if it's valid to assume that lists of links are mutually exclusive with selection/checkbox/radio lists (which _seems_ reasonable), then this is safe.

packages/mdc-list/foundation.js
```diff
@@ -249,6 +249,9 @@ class MDCListFoundation extends MDCFoundation {
       this.focusLastElement();
     } else if (isEnter || isSpace) {
       if (isRootListItem) {
+        // Return early if enter key is pressed on anchor element which triggers synthetic MouseEvent event.
+        if (evt.target.tagName === 'A' && isEnter) return;
```
kfranqueiro: I suspect we shouldn't be doing this in the foundation for the sake of framework wrappers. The adapter API that's removed in this PR was also separating that from the foundation.

packages/mdc-list/README.md
```diff
@@ -504,11 +504,11 @@ Method Signature | Description
 `removeClassForElementIndex(index: Number, className: String) => void` | Removes the `className` class to the list item at `index`.
 `focusItemAtIndex(index: Number) => void` | Focuses the list item at the `index` value specified.
 `setTabIndexForListItemChildren(index: Number, value: Number) => void` | Sets the `tabindex` attribute to `value` for each child button or anchor element in the list item at the `index` specified.
-`followHref(element: Element) => void` | If the given element has an href, follows the link.
 `hasRadioAtIndex(index: number) => boolean` | Returns true if radio button is present at given list item index.
 `hasCheckboxAtIndex(index: number) => boolean` | Returns true if checkbox is present at given list item index.
 `isCheckboxCheckedAtIndex(index: number) => boolean` | Returns true if checkbox inside a list item is checked.
 `setCheckedCheckboxOrRadioAtIndex(index: number, isChecked: boolean) => void` | Sets the checked status of checkbox or radio at given list item index.
+`notifyAction(index: number) => void` | Notifies user action on list item including keyboard and mouse actions.
```
kfranqueiro: We seem to be missing a section for the MDCList component (we skip directly to Usage within Web Frameworks). We should have a section documenting the component's public methods/properties and also document this event there.

packages/mdc-list/foundation.js
kfranqueiro: I still have questions/reservations about removing followHref in this PR...

* Does it not seem unnatural for links to process under different circumstances than the action event fires? The other reason `followHref` existed was so that both space and enter would activate links within MDC List, in keeping with e.g. the ARIA menu spec for menu item activation. (One could argue the drawer isn't exactly a menu, but then what is it? We're making the keyboard behave menu-like for it...)
* Does `followHref` _need_ to be removed for this PR? Was it ever incorrectly triggering (or triggering twice? Its removal strikes me as a tangential side effect which is why I've been asking about this.

packages/mdc-list/README.md
kfranqueiro: On second thought it might make sense to tackle the missing section after both this and #4124 are merged...

packages/mdc-list/foundation.js
abhiomkar: As you mentioned, I think `followHref` was introduced to bring consistent behaviour with <kbd>Enter</kbd> & <kbd>Space</kbd> for anchor elements which I think unnecessary hack.

ARIA navigation menu button [example](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html) doesn't even consider <kbd>Space</kbd> key behaviour.

I do not think it was triggering `followHref` twice earlier.

packages/mdc-list/README.md
abhiomkar: will update README in last PR once other PRs are merged to avoid merge conflict.

packages/mdc-list/foundation.js
```diff
@@ -264,8 +267,7 @@ class MDCListFoundation extends MDCFoundation {
           this.setSelectedIndex(currentIndex);
         }
 
-        // Explicitly activate links, since we're preventing default on Enter, and Space doesn't activate them.
-        this.adapter_.followHref(currentIndex);
+        this.adapter_.notifyAction(currentIndex);
```
kfranqueiro: Related to https://github.com/material-components/material-components-web/pull/4151#pullrequestreview-189079533, I am wondering if we should do a couple of things here:

1. Call `preventDefaultEvent` in all cases rather than just the `isSingleSelectionList_` case?
2. Only call `notifyAction` if `ELEMENTS_KEY_ALLOWED_IN.indexOf(tagName) === -1`, since e.g. pressing space while a text input is focused shouldn't count?

abhiomkar: This seems like out of scope of this PR. Can we track this as a separate issue?

We also only call `setSelectedIndex` if `ELEMENTS_KEY_ALLOWED_IN.indexOf(tagName) === -1`

abhiomkar: Fixing #1

kfranqueiro: I think we should also fix #2 because otherwise we're going to be firing an event that users may react to, in cases where it really isn't appropriate...

I don't really see how it's out of scope, when these PRs are creating these issues... these are complex changes that I'm trying to prevent new bugs from arising out of.

abhiomkar: I looked at 2nd point: The block of code is executed only if the keydown event is received on list root element (i.e., `if (isRootListItem)`). So the `notifyAction` will never be called if keydown received on list item child elements.

Regarding the 1st point: the scroll behaviour on space key already exists without this PR. I would still prefer addressing this issue in separate PR.

kfranqueiro: Yeah, you're right about the 2nd point.

However, the 1st issue does not exist without this and the menu PR, because these PRs together remove the `preventDefault` logic that currently exists in Menu.

(Non-)Repro:

1. `git checkout master`
2. `git checkout list_custom_event_menu_update_issue3481 -- test/screenshot/spec/mdc-menu` to get your menu screenshot test page against master
3. `npm start`
4. Browse to http://localhost:8080/spec/mdc-menu/classes/baseline.html
5. Make sure the window size is small enough that the body has a scrollbar.
6. Use space key to select an item

Against master: Scroll doesn't happen.
On menu branch (which is based on this branch): Scroll happens.

abhiomkar: I think I know what is happening.

This is indeed issue created with next PR where we removed preventDefault code from menu foundation. I thought preventDefault is only intended for selectable list, I changed it to preventDefault for all list variants to avoid complexity for menu.

kfranqueiro: I just re-tested this with the update and it looks like it's resolved now 👍

test/unit/mdc-list/foundation.test.js
```diff
@@ -47,7 +47,7 @@ test('defaultAdapter returns a complete adapter implementation', () => {
     'removeAttributeForElementIndex', 'addClassForElementIndex', 'removeClassForElementIndex',
     'focusItemAtIndex', 'setTabIndexForListItemChildren', 'followHref', 'hasRadioAtIndex',
     'hasCheckboxAtIndex', 'isCheckboxCheckedAtIndex', 'setCheckedCheckboxOrRadioAtIndex',
-    'isFocusInsideList',
+    'notifyAction', 'isFocusInsideList',
```
kfranqueiro: You added notifyAction but you need to remove followHref - the unit test is failing because of this

kfranqueiro:
PS: I think this should probably be considered a `feat` in the commit subject, along with the breaking change in the commit description.

PR #4160, opened 2018-12-10
fix(text-field): Don't move caret when value has not changed

developit:
This fixes an issue where the cursor on Mobile Safari will be reset to its end position any time `.setValue()` is called.


johnhaitas:
> This fixes an issue where the cursor on Mobile Safari will be reset to its end position any time `.setValue()` is called.

This bug also occurs with MacOS Safari

PR #4165, opened 2018-12-11
fix(select): apperance none not working in chromium

j-o-d-o:
I have to add:
```
-webkit-appearance: none;
```
In order to get rid of the default triangle of the select in chromium (65.0.3325.181).


j-o-d-o:
Ah okay, that explains why it worked with the css import but not using sass. Will do, thx for the tip!

Issue #4143, opened 2018-12-05
MDCTextField: Full width textarea should disable horizontal resize


williamernest:
Discovered on the catalog page. Changing the width of the full width textarea causes the outline to remain full width but the textarea inside to become smaller. Horizontal resize should be disabled in this instance.

PR #4167, opened 2018-12-12
fix(text-field): Restrict resize to vertical for full width text area

williamernest:
fixes: #4143 
fixes: #4168 
The text area by default allows horizontal resize, but the full width modifier class causes the container and notched outline to remain at full width.


kfranqueiro:
New changes also LGTM

Issue #4170, opened 2018-12-12
MDCNotchedOutline: New version allows label to overflow outline


williamernest:
The previous version causes the component to expand to fit and truncated the label. The new version does not truncate or expand and allows the label to overflow the text field/text area.

PR #4171, opened 2018-12-12
fix(notched-outline): Fix label overflow

williamernest:
fixes: #4170 
The new notched outline allowed the label to overflow the parent elements, so this causes the label to ellipse when down and to clip when floating. 

BREAKING CHANGE: Adds a new adapter API, removeNotchWidthProperty. Also changes the vanilla implementation of MDCFloatingLabelAdapter#getWidth to return scrollWidth rather than offsetWidth.


packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -69,13 +70,20 @@
...
     .mdc-floating-label {
       display: inline-block;
       position: relative;
       top: 17px;
       bottom: auto;
+      max-width: 100%;
+    }
+
+    .mdc-floating-label--float-above {
+      max-width: calc(100% / .75);
```
kfranqueiro: This seems incorrect and seems to be causing a FOUC. If I comment out the text field instantiation in fixture.js, the floated labels are the wrong width. If I set this to simply 100%, they're the correct length.

williamernest: I moved this so it only applied when the `--upgraded` class is added.

packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -69,13 +70,20 @@
...
+      text-overflow: clip;
```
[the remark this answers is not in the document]
williamernest: The ellipses started really early and it results in even more of the label being hidden. Switching to clipped allows more of the label to be seen (compared to the non-floating state)

test/unit/mdc-notched-outline/mdc-notched-outline.test.js
```diff
@@ -55,13 +55,14 @@ test('adapter#addClass adds a class to the root element', () => {
 test('adapter#removeClass removes a class to the root element', () => {
   const {root, component} = setupTest();
   component.getDefaultFoundation().adapter_.removeClass('foo');
-  assert.isFalse(root.classList.contains('foo'));
+  component.getDefaultFoundation().adapter_.removeNotchWidthProperty();
```
kfranqueiro: Is the width property actually set on this element before these statements run? Otherwise we're not really verifying anything.

packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -68,14 +69,22 @@
...
       display: inline-block;
       position: relative;
       top: 17px;
       bottom: auto;
+      max-width: 100%;
+    }
+
+    &--upgraded .mdc-floating-label--float-above {
+      max-width: calc(100% / .75);
+      text-overflow: clip;
```
kfranqueiro: (Discussed in DM) Is it safe to apply this even when not upgraded, to avoid the ellipsis appearing then disappearing once JS runs?

packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -68,14 +69,25 @@
...
 
     .mdc-floating-label {
       display: inline-block;
       position: relative;
       top: 17px;
       bottom: auto;
+      max-width: 100%;
+    }
+
+    //.mdc-floating-label--float-above {
```
kfranqueiro: Is this supposed to be removed, or uncommented?

williamernest: Removed. I changed the floating label to use scrollWidth and adjusted the max-width of the notch so it should work better now (label shows ellipsis but the notch shows the rest of the label if you adjust the text area).

Issue #4169, opened 2018-12-12
MDCMenu: Selection group shows all check icons since extra specificity was added to the list item graphic


williamernest:
refs: #4021
The selection group icons are all visible due to the extra specificity on [L98](https://github.com/material-components/material-components-web/blob/fa415797bed1126073fe7fb6aa421b179b51a4a4/packages/mdc-list/mdc-list.scss#L98) of mdc-list.scss.

PR #4172, opened 2018-12-12
fix(menu): Add specificity to selection group class

williamernest:
fixes: #4169


packages/mdc-menu/mdc-menu.scss
```diff
@@ -63,13 +63,14 @@
   .mdc-list-item {
     @include mdc-rtl-reflexive-box(padding, left, 56px);
   }
-}
 
-.mdc-menu__selection-group-icon {
-  @include mdc-rtl-reflexive-position(left, 16px);
+  // Extra specificity required because list-item graphic needed to increase specificity.
```
acdvorak: I don't see `.mdc-list-item__graphic` anywhere on the screenshot test page. How is this related?

kfranqueiro:
I think this is referencing the wrong bug?

Also, should I be able to repro the bug on the new screenshot test page if I un-nest the selector? The page seems to look identical to me in both cases.

kfranqueiro:
Here's the report to update goldens from:

https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/12/13/21_10_43_629/report/report.html?utm_source=cli&utm_medium=test_results

I can now repro the bug by reverting the CSS change on the branch.

kfranqueiro:
I noticed the checkmark is misaligned in IE but this seems to be a pre-existing issue to this PR, so I logged #4184.

packages/mdc-menu/mdc-menu.scss
acdvorak: Nit: Can you update this comment to something like:

```js
// Extra specificity required to override `display` property on `mdc-list-item__graphic`.
```

PR #4189, opened 2018-12-17
chore(infrastructure): Don't kill active screenshot tests

### LINKED ISSUE #4186: Screenshot tests fail when CBT returns invalid response
### Problem

Since at least 2018-10-10, screenshot tests occasionally fail with the following error:

```
ERROR: TypeError: Cannot read property 'length' of undefined
    at CbtApi.killZombieSeleniumTests (/home/travis/build/material-components/material-components-web/test/screenshot/infra/lib/cbt-api.js:421:71)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

https://travis-ci.com/material-components/material-components-web/jobs/165042539

### Root cause

Here's the corresponding line of code:

https://github.com/material-components/material-components-web/blob/716da5a1739c6122704a4efca1b26e606967f4cb/test/screenshot/infra/lib/cbt-api.js#L421

This would seem to suggest that one of the `GET` requests to CBT to query Selenium session status returned an invalid response.

And in fact, I've seen CBT return `{}` on occasion, so that's probably what's causing it.

### LINKED ISSUE #4187: Screenshot tests incorrectly kill other users' tests
### Background

When a user cancels a Travis CI job, it immediately terminates the Linux VM without giving our test code a chance to stop any running Selenium sessions.

As a result, the Selenium sessions continue to run idle on CBT for 24–48 hours, consuming valuable concurrency resources and reducing the number of browsers available to other users.

To work around this, at the beginning of every screenshot test run, we check to see if there are any "zombie" sessions (those that have not received any Selenium commands in at least 2 minutes), and if so, we kill them.

### Problem

The logic to terminate stalled "zombie" tests evidently has an error in it, because I've been seeing a lot of tests fail partway through with errors like this:

```
WebDriverError: Session no longer available: 907e0ad9-835b-4e00-a515-be8b7dedb630
```

For example, this job:

https://travis-ci.com/material-components/material-components-web/jobs/165269435

terminated session `907e0ad9-835b-4e00-a515-be8b7dedb630` in this job:

https://travis-ci.com/material-components/material-components-web/jobs/165267194

It should not have done that.

### Examples

Here's a list of all the test failures I've seen today:

* https://travis-ci.com/material-components/material-components-web/jobs/165082668
* https://travis-ci.com/material-components/material-components-web/jobs/165257990
* https://travis-ci.com/material-components/material-components-web/jobs/165266919
* https://travis-ci.com/material-components/material-components-web/jobs/165267194
* https://travis-ci.com/material-components/material-components-web/jobs/149675421
* https://travis-ci.com/material-components/material-components-web/jobs/165265499
* https://travis-ci.com/material-components/material-components-web/jobs/165269641
* https://travis-ci.com/material-components/material-components-web/jobs/165272126
* https://travis-ci.com/material-components/material-components-web/jobs/165274099

acdvorak:
This PR:

* Increases the zombie Selenium session timeout from `2` minutes to `20` minutes.
    - This should ensure that our tests have enough time to finish, even if CBT doesn't return any Selenium commands in the HTTP response. (We reported this issue to CBT and they are currently investigating, but it might not be fixed for several weeks.)
* Prevents tests from dying when CBT returns an empty JSON response.
    - Instead of throwing an error, we simply log a warning to the console and continue.
* Queries Selenium tests by _test_ ID instead of _session_ ID.
    - For some reason, CBT occasionally returns an empty JSON response when we query by _session_ ID (a standard Selenium/WebDriver identifier). Querying by _test_ ID (a CBT-specific identifier) seems to return a proper response.
* Logs all active Selenium tests to the console to aid in future debugging.

Fixes #4186
Fixes #4187

Terminal screenshot


test/screenshot/infra/lib/cbt-api.js
```diff
@@ -527,6 +545,45 @@ https://crossbrowsertesting.com/account
...
+        selenium_test_id: session.selenium_test_id,
+        selenium_session_id: session.selenium_session_id,
+        start_date: session.start_date,
+        startup_finish_date: session.startup_finish_date,
+        finish_date: session.finish_date,
+        active: session.active,
+        state: session.state,
+        show_result_web_url: session.show_result_web_url,
+        show_result_public_url: session.show_result_public_url,
+        caps: session.caps,
```
kfranqueiro: What is this? Judging by the externs change it was just added? I don't see it referenced anywhere else in this PR.

test/screenshot/infra/lib/cbt-api.js
```diff
@@ -400,24 +401,39 @@ https://crossbrowsertesting.com/account
...
 
-    const stalledSeleniumSessionIds = [];
+    this.logSeleniumSessions_(infoResponses, 'Selenium sessions info');
+
+    const stalledSeleniumTestIds = [];
+
+    for (let i = 0; i < infoResponses.length; i++) {
+      const infoResponse = infoResponses[i];
+      if (!infoResponse.commands) {
+        this.warnMalformedResponse_(infoResponse, 'info', activeSeleniumTestIds[i]);
```
kfranqueiro: What effect does this end up having vs. the script dying on length of undefined before? Does it basically just skip checking one active session?

test/screenshot/infra/lib/cbt-api.js
```diff
@@ -400,24 +401,39 @@ https://crossbrowsertesting.com/account
...
+    const activeSeleniumTestIds = listResponse.selenium.map((session) => session.selenium_test_id);
 
     /** @type {!Array<!CbtSeleniumInfoResponse>} */
     const infoResponses = await Promise.all(activeSeleniumTestIds.map((seleniumTestId) => {
       const infoStackTrace = getStackTrace('killZombieSeleniumTests');
       return this.sendRequest_(infoStackTrace, 'GET', `/selenium/${seleniumTestId}`);
     }));
 
-    const stalledSeleniumSessionIds = [];
+    this.logSeleniumSessions_(infoResponses, 'Selenium sessions info');
```
kfranqueiro: I see two calls to `logSeleniumSessions_` (here and above). IIUC this one only logs detected active sessions. What does the previous one log? Do we need both of these or is this a lot of added noise? Do the responses contain anything sensitive we should avoid outputting?

test/screenshot/infra/lib/cbt-api.js
acdvorak: It's always been present in the HTTP response, but I didn't add it to the externs file before for some reason.

I found it helpful to have this in the console when I was trying to match up Session IDs with Test IDs in the CBT UI.

test/screenshot/infra/lib/cbt-api.js
acdvorak: Yes, exactly

test/screenshot/infra/lib/cbt-api.js
acdvorak: Annoyingly, the "list all active Selenium sessions" endpoint returns everything we need _except_ the list of commands.

So I have to loop through the active sessions, pull out their test IDs, and query each session individually to get the commands.

I updated the PR description with a screenshot of an example log message. As you can see, they're nearly identical, but I thought it would be useful to print both in case something goes wrong and we need to debug. It does add some noise, but at this point I'd rather have more information than less, so that we can track down future issues more easily.

test/screenshot/infra/types/cbt-externs.js
```diff
@@ -127,6 +128,7 @@
  *   resolution: !CbtSeleniumResolution,
  *   os: !CbtSeleniumOs,
  *   browser: !CbtSeleniumBrowser,
+ *   caps: !Object<string, string>,
```
kfranqueiro: Revert this file?

acdvorak: As discussed offline, we'll keep the changes to this file because they're valid regardless of whether we reference the properties or not.

PR #4195, opened 2018-12-18
chore: Declare all sass variables as !default (#3708)

### LINKED ISSUE #3708: Add !default to all Sass variables
See #3635 for motivation

Document the fact that backward compatibility is not guaranteed: "You should not do this, but if you need to, you're on your own."

levibotelho:
Continuing what was done to checkbox in #3635 and fulfilling #3708, I've added `!default` to all variables such that they can be overridden in client applications.

Fixes #3708


levibotelho:
> I signed it!

levibotelho:
Could someone please take a look at this? Happy to make changes if necessary.

levibotelho:
@abhiomkar good idea, how does that look to you?

PR #4203, opened 2018-12-20
fix(snackbar): Rename action/dismiss classes and revise docs/tests

kfranqueiro:
This revises the names of the action and dismiss elements to make them potentially more indicative and less bound to specifically being a button or icon, for potential future-proofing.

I've also revised our docs' wording a bit after talking to our designer again, who expressed serious reservations about featuring the dismiss icon as prominently as we were, at least until further discussion RE IxD and accessibility on his end.

(I'm not including a breaking change note, because this hasn't been released yet. All the better to revise classnames now.)


kfranqueiro:
I got a LGTM from our designer on the README update.

PR #4223, opened 2019-01-04
Fix typo in notched-outline mixins

brennenberkley:
I noticed `mdc-shape-prop-value_($radius)` was mistyped and causing errors


brennenberkley:
I signed it!

PR #4224, opened 2019-01-04
fix(notched-outline): fix missing shape functions import

abhiomkar:


abhiomkar:
That's correct. The functions was indirectly imported from `shape/mixins`.
There was no error messages so far. Found this after closed PR #4223

Issue #4227, opened 2019-01-04
Allow finer-grained control of emitted CSS


kfranqueiro:
In the past, we've received requests (such as #3357) for ways to better control what CSS is output by MDC Web. Being able to more granularly control what CSS we output also has the potential to help prospective framework wrappers that also want to govern output styles.

To this end, @mmalerba has worked on a prototype which would allow selectively including MDC Web's styles either all at once or piece-by-piece, with the ability to request the following subsets:

* base ("structure") - anything that doesn't fall into one of the more specific subsets
* mdc-theme-dependent ("color")
* typography
* animation

Miles' POC, along with a bit of fine-tuning and infrastructure work, is in #4228.

The end result of this is that existing usage of MDC Web will behave exactly the same - no existing mixin APIs change, nor does the output CSS if you continue to import `@material/foo/mdc-foo`. But you will also have the option of `@include mdc-foo(...)` and specifying a query of one or more subsets of the component's styles.

Since these changes are intended to be fully backwards-compatible, we can merge these PRs to master one by one. We should also be able to check CSS output on master vs. on the branch and verify that the results are the same (run `npm run build` on each one and move the folder somewhere to be diffed afterwards).

The following components have emitted styles that will need to be converted:

* [x] mdc-button (@mmalerba + @kfranqueiro, #4228)
* [x] mdc-checkbox (@mmalerba, #4258)
* [x] mdc-ripple (partially in #4228; emitted styles in #4383)
* [x] mdc-menu-surface (@crisbeto, #4279)
* [x] mdc-menu (@crisbeto, #4278)
* [x] mdc-radio (@mmalerba, #4270)
* [x] mdc-switch (@crisbeto, #4275)
* [x] mdc-card (@mmalerba, #4301)
* [x] mdc-typography (mixins done by @mmalerba in #4305; all emitted styles covered via mixins)
* [x] mdc-elevation (mixins done by @mmalerba in #4308; emitted styles in #4383)
* [x] mdc-theme (`mdc-theme-prop` not accepting `$query` on purpose; emitted styles in #4383)
* [x] mdc-shape (mixins only; @mmalerba, #4384)
* [x] mdc-form-field (@mmalerba, #4521)
* [x] mdc-dialog (@kfranqueiro, #4524)
* [x] mdc-fab (@kfranqueiro, #4526)
* [x] mdc-grid-list (@kfranqueiro, #4534)
* [x] mdc-image-list (@kfranqueiro, #4535)
[...]

PR #4228, opened 2019-01-04
feat: Add feature targeting support and apply to mdc-button

kfranqueiro:
This adds a `mdc-feature-targeting` package to support the effort summarized in #4227, and converts the mdc-button package (making necessary adjustments to mdc-ripple to allow it).

Many thanks to @mmalerba for spearheading the initial design and implementation.

The most important thing to note about the feature-targeting infrastructure is that it should be used as little as necessary to provide the granularity outlined in #4227 - i.e., `$query` is only passed down as far as it needs to be passed to ensure this granularity. It is not passed to every single mixin.

This PR also adds a new test script to verify that nothing is missed when feature-targeting. All this requires to work is that as we restructure each package's SCSS, we add an import for it in `test/scss/feature-targeting.scss`.


packages/mdc-feature-targeting/_functions.scss
```diff
@@ -0,0 +1,170 @@
...
+
+  @return (
+    available: $available-features,
+    query: map-get(nth($feature-targets, 1), query)
+  );
+}
+
+// Creates a feature query that is satisfied iff all of its sub-queries are satisfied.
+@function mdc-feature-all($feature-queries...) {
+  $valid: mdc-feature-verify-query_($feature-queries...);
```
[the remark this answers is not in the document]
mmalerba: I think I wrote this originally. I remember it either stripping it or failing some kind of lint check because nothing was being done with the result. Can't remember which...

packages/mdc-feature-targeting/_functions.scss
```diff
@@ -0,0 +1,170 @@
...
+  $valid: mdc-feature-verify-target_($feature-targets...);
+  $available-features: ();
+
+  @each $target in $feature-targets {
+    $available-features: append($available-features, map-get($target, target));
+  }
+
+  @return (
+    available: $available-features,
+    query: map-get(nth($feature-targets, 1), query)
```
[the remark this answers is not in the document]
mmalerba: The list is 1-indexed: http://sass-lang.com/documentation/Sass/Script/Functions.html#nth-instance_method

packages/mdc-feature-targeting/_functions.scss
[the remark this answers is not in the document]
kfranqueiro: I gave these the hairy eyeball at first as well. It's presumably related to Sass functions being required to return values, but these function calls are used purely in an assertion sense - they throw errors if they fail.

packages/mdc-feature-targeting/_functions.scss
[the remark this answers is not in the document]
moog16: Had no idea! Thanks

packages/mdc-feature-targeting/_functions.scss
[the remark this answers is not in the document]
moog16: gotcha...its just some sass issue

kfranqueiro:
Regarding CSS file size before/after, the difference should ideally be zero. The reason this is structured the way it is is to be able to easily verify that we're outputting the same styles before and after these changes.

I wonder if your other idea RE organization occurred to @mmalerba (I suspect he was struck with a few ideas as he prototyped this). The main drawback I can think of for the purposes of the conversion itself is that it'd make it harder to glance at the generated CSS before/after to confirm they're the same, because it would likely reorder a bunch of styles (in a way that is harmless, but would require a lot more careful inspection of diffs, especially once we get into e.g. text field). I think @mmalerba purposely aimed for keeping the output CSS identical to before.

I definitely agree that the feature-targeting package can be hard to grok at first - and that makes me realize it doesn't yet have a README here. I'd like to get this in first so that people can start pushing forward with more conversions (the Angular team is interested in helping out), but I'm happy to take on writing the README (perhaps with a review from Miles) separately.

mmalerba:
@abhiomkar Yes, I did consider just having separate mixins for each feature, but I felt their were a number of advantages this approach had over it:
- It doesn't require breaking changes to the current mixin API. If we wanted to use the separate mixin approach we would need to split up the current mixins
- It can be done in a way that results in zero changes to the output CSS. Refactoring the CSS can be really tricky, as even just reordering lines can change what rules get applied in the end.
- It allows us to group together all styles that apply to a single class. This might be a personal preference thing, but I think it's nice to be able to see all of the styles that are being applied to `.mdc-button` in one place, rather than having to go look for the selector in each of the separate feature mixins.
- It's more powerful than just having separate mixins because it allows combinations of features (e.g. get all styles that apply to either color or typography, get all styles that apply to both color and typography, get all styles that apply to color but not animations, etc)

packages/mdc-button/_mixins.scss
```diff
@@ -163,39 +273,62 @@
...
 
-  &:active {
-    @include mdc-elevation(8);
+  @include mdc-feature-targets($feat-animation) {
+    transition: mdc-elevation-transition-value();
   }
 
-  &:disabled {
-    @include mdc-elevation(0);
+  @include mdc-feature-targets($feat-color) {
```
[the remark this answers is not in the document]
kfranqueiro: This is a little weird since the _default_ value that mdc-elevation uses does not depend on any mdc-theme color "slot", but we may want to err on the safe side in case it eventually does.

packages/mdc-feature-targeting/_mixins.scss
```diff
@@ -0,0 +1,12 @@
+@import "./functions";
+
+// Mixin that annotates the contained styles as applying to specific cross-cutting features
+// indicated by the given list of feature targets.
+@mixin mdc-feature-targets($feature-targets...) {
+  $valid: mdc-feature-verify-target_($feature-targets...);
```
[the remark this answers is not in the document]
kfranqueiro: Nice catch. Yeah, it'll catch it in parse anyway, there'll just be one more line in the stack which I don't think is a big deal since there's 4-5 of them anyway.

packages/mdc-button/_mixins.scss
```diff
@@ -30,6 +32,99 @@
...
+    @include mdc-feature-targets($feat-structure) {
+      @include mdc-button-outline-width(2px);
+    }
+
+    @include mdc-feature-targets($feat-color) {
+      @include mdc-button-outline-color(primary);
+    }
+  }
+
+  @include mdc-feature-targets($feat-structure) {
```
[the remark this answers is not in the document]
kfranqueiro: Looking at this particular case that's a reasonable point, but I can explain why it's written this way. The goal with this restructure is to use `mdc-feature-targets` as high up the stack as possible and pass `$query` as infrequently as possible. The `--raised`, `--filled`, and `--outlined` mixins include styles that target different feature targets, while the `--dense` mixin only involves styles that target structure.

packages/mdc-ripple/_mixins.scss
```diff
@@ -21,68 +21,123 @@
...
+    }
   }
 
   &::before {
-    // Also transition background-color to avoid unnatural color flashes when toggling activated/selected state
-    transition:
-      opacity $mdc-states-wash-duration linear,
-      background-color $mdc-states-wash-duration linear;
-    z-index: 1; // Ensure that the ripple wash for hover/focus states is displayed on top of positioned child elements
+    @include mdc-feature-targets($feat-animation) {
```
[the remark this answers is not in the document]
kfranqueiro: Sass is smart enough to omit the surrounding selector if there is no content.

e.g., this input:

```scss
.test {
  $query: mdc-feature-any(color, structure);
  $structure: mdc-feature-create-target($query, structure);
  $animation: mdc-feature-create-target($query, animation);
  
  @include mdc-feature-targets($structure) {
    position: relative;
  }
  @include mdc-feature-targets($animation) {
    transition: opacity 150ms;
  }
}
```

outputs:

```css
.test {
  position: relative;
}
```

But if you change `$query` to just `mdc-feature-any(color)` then it outputs nothing.

packages/mdc-ripple/_mixins.scss
```diff
@@ -21,68 +21,123 @@
...
-    animation: $mdc-ripple-fade-out-duration mdc-ripple-fg-opacity-out;
-    // Retain transform from mdc-ripple-fg-radius-in activation
-    transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
+    @include mdc-feature-targets($feat-animation) {
+      animation: $mdc-ripple-fade-out-duration mdc-ripple-fg-opacity-out;
+    }
+
+    @include mdc-feature-targets($feat-structure) {
+      // Retain transform from mdc-ripple-fg-radius-in activation
+      transform: translate(var(--mdc-ripple-fg-translate-end, 0)) scale(var(--mdc-ripple-fg-scale, 1));
```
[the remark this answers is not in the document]
kfranqueiro: I don't think so. This is `transform`, not `transition`.

packages/mdc-ripple/_mixins.scss
```diff
@@ -121,35 +179,52 @@
   #{$cssOnlyFocusSelector},
   &.mdc-ripple-upgraded--background-focused::before {
     // Note that this duration is only effective on focus, not blur
-    transition-duration: 75ms;
-    opacity: $opacity;
+    @include mdc-feature-targets($feat-animation) {
```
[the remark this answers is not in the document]
kfranqueiro: I've answered this on the earlier comment; Sass will not emit the selector if no styles are emitted within it.

packages/mdc-feature-targeting/_functions.scss
```diff
@@ -0,0 +1,170 @@
...
+//   the structure that is intended to be passed to the `@mdc-feature-targets` mixin.
+//   (e.g. `(target: color, query: (op: any, queries: (color, typography))`).
+
+//
+// Public
+//
+
+// Creates a feature target from the given feature query and targeted feature.
+@function mdc-feature-create-target($feature-query, $targeted-feature) {
+  $feature-target: (target: $targeted-feature, query: $feature-query);
```
abhiomkar: Reverse this in order of function's arguments.

`(query: $feature-query, target: $targeted-feature)`

packages/mdc-feature-targeting/_functions.scss
```diff
@@ -0,0 +1,170 @@
...
+    $available-features: append($available-features, map-get($target, target));
+  }
+
+  @return (
+    available: $available-features,
+    query: map-get(nth($feature-targets, 1), query)
+  );
+}
+
+// Creates a feature query that is satisfied iff all of its sub-queries are satisfied.
```
abhiomkar: nit: iff -> if 

through out this file.

packages/mdc-feature-targeting/_functions.scss
```diff
@@ -0,0 +1,170 @@
...
+    $valid: mdc-feature-verify-feature_($targeted-feature) and mdc-feature-verify-query_($feature-query);
+  }
+
+  @return true;
+}
+
+// Checks whether the given feature query is satisfied by the given list of available features.
+@function mdc-feature-is-query-satisfied_($feature-query, $available-features) {
+  $valid: mdc-feature-verify-query_($feature-query);
+  $valid: mdc-feature-verify-feature_($available-features...);
```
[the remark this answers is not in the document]
kfranqueiro: Matt asked this too. See https://github.com/material-components/material-components-web/pull/4228#discussion_r247659569

packages/mdc-button/_mixins.scss
```diff
@@ -30,6 +32,99 @@
...
+      @include mdc-button-container-fill-color(transparent);
+    }
+
+    @include mdc-feature-targets($feat-color) {
+      @include mdc-button-ink-color(primary);
+    }
+
+    @include mdc-states(primary, false, $query);
+
+    @include mdc-feature-targets($feat-structure) {
```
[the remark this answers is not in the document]
kfranqueiro: Probably. This was written in a way that avoids any CSS diffs whatsoever, but we could make a few exceptions after the initial 0-diff verification.

packages/mdc-feature-targeting/_functions.scss
kfranqueiro: I think this is intentional, referring to "if and only if"

packages/mdc-button/_mixins.scss
```diff
@@ -30,6 +32,98 @@
...
+        @include mdc-button__icon_;
+      }
+    }
+
+    @include mdc-feature-targets($feat-color) {
+      @include mdc-button-ink-color(primary);
+    }
+
+    @include mdc-states(primary, false, $query);
+
```
abhiomkar: nit: Remove empty line at the end of block.

packages/mdc-button/_mixins.scss
```diff
@@ -92,48 +186,63 @@
...
+    &:active {
+      outline: none;
+    }
+
+    &:hover {
+      cursor: pointer;
+    }
   }
 
   &:disabled {
```
abhiomkar: Curious, wouldn't the below test fail when all the feature are turned off and this emits empty style rule? Unless we're looking for only the existence of CSS properties.

```
$:disabled {
}
```

packages/mdc-button/_mixins.scss
```diff
@@ -163,39 +272,62 @@
...
 
-  &:active {
-    @include mdc-elevation(8);
+  @include mdc-feature-targets($feat-animation) {
+    transition: mdc-elevation-transition-value();
   }
 
-  &:disabled {
-    @include mdc-elevation(0);
+  @include mdc-feature-targets($feat-color) {
```
abhiomkar: Same here. merge these styles with above `feat-color` block.

packages/mdc-feature-targeting/_functions.scss
```diff
@@ -0,0 +1,170 @@
+@import "./variables";
```
abhiomkar: Missing `@license` in all of the files in `mdc-feature-targeting`.

packages/mdc-ripple/_mixins.scss
```diff
@@ -21,68 +21,123 @@
...
-      background-color $mdc-states-wash-duration linear;
-    z-index: 1; // Ensure that the ripple wash for hover/focus states is displayed on top of positioned child elements
+    @include mdc-feature-targets($feat-animation) {
+      // Also transition background-color to avoid unnatural color flashes when toggling activated/selected state
+      transition:
+        opacity $mdc-states-wash-duration linear,
+        background-color $mdc-states-wash-duration linear;
+    }
+
+    @include mdc-feature-targets($feat-structure) {
```
abhiomkar: I think it is safe to merge these structure style to above feat-structure block.

Comment applies across this file.

packages/mdc-button/_mixins.scss
kfranqueiro: Matt also asked this question. Sass is smart enough to not emit the containing selector if there are no rules. See https://github.com/material-components/material-components-web/pull/4228#discussion_r247692856

packages/mdc-feature-targeting/_functions.scss
kfranqueiro: Good catch, I'll add those.

packages/mdc-ripple/_mixins.scss
```diff
@@ -21,68 +21,123 @@
...
+  @include mdc-feature-targets($feat-structure) {
+    --mdc-ripple-fg-size: 0;
+    --mdc-ripple-left: 0;
+    --mdc-ripple-top: 0;
+    --mdc-ripple-fg-scale: 1;
+    --mdc-ripple-fg-translate-end: 0;
+    --mdc-ripple-fg-translate-start: 0;
+
+    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+    will-change: transform, opacity;
```
abhiomkar: `will-change` should be moved to `animation` feature set? or is it used in both `structure` & `animation` feature sets?

kfranqueiro: This one seems like it could go in either/or, but I've moved it to animation.

packages/mdc-ripple/_mixins.scss
```diff
@@ -224,12 +302,22 @@
...
-  @include mdc-states-hover-opacity(mdc-states-opacity($color, hover) + $opacity-modifier);
+@mixin mdc-states-interactions_($color, $has-nested-focusable-element, $opacity-modifier: 0, $query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-feature-targets($feat-color) {
+    @include mdc-states-base-color($color);
+  }
+
+  @include mdc-feature-targets($feat-structure) {
```
moog16: should this be color?

kfranqueiro: I suspect it's structure because the emitted style isn't color related, but the logic is color-dependent... @mmalerba, thoughts on this? (This would apply to press/focus as well)

packages/mdc-ripple/_mixins.scss
kfranqueiro: Pat made a good point that `will-change` can affect appearance because it causes the content to render on a different layer, so I've moved this back to structure. We wouldn't want e.g. screenshot tests having differences from real use due to styles suppressed with the intent of skipping animations.

packages/mdc-ripple/_mixins.scss
kfranqueiro: Discussed offline and yes, this should be changed because it's color-dependent in typical use. Pushed a commit to fix it.

Issue #4225, opened 2019-01-04
Convert packages to TypeScript


kfranqueiro:
### Overview

One of our Q1 2019 goals is to convert our packages' JS to TypeScript, as an increasing number of our consumers have TypeScript codebases and would benefit from this change.

The primary goal is to convert the packages themselves - converting the unit tests would be a plus (and would likely help test the conversion) but ideally we should be able to first convert packages one-by-one and continue to test them with the existing JS unit tests to confirm no unintended changes.

The intent of this conversion is ideally to involve no breaking changes, and to continue to provide both the ES6 and ES5 formats that we currently distribute in addition to TS.

This is a master issue for tracking all packages during the conversion, rather than creating 20 separate issues. I'm listing the components with JS below in a rough chronological order of approach (generally prioritizing dependencies ahead of dependents).

### Packages to convert

* [x] `base` (dependency of everything) (@moog16, #4241 along with infrastructure)
* [x] `dom` (@acdvorak, #4269)
* [x] `animation` (@acdvorak, #4271)
* [x] `linear-progress` (@acdvorak, #4272)
* [x] `ripple` (@moog16, #4300)
* [x] `menu-surface` (@acdvorak, #4273)
* [x] `switch` (@mmalerba)
* [x] `form-field` (@moog16, #4327)
* [x] `selection-control` (exists solely for closure annotations; dependency of checkbox/radio) (@moog16, #4327)
* [x] `checkbox` (depends on `form-field`) (@moog16 and @acdvorak, #4319)
* [x] `radio` (@moog16 and @acdvorak, #4329)
* [x] `icon-button` (@abhiomkar and @acdvorak, #4325)
* [x] `list` (@moog16 and @acdvorak, #4334)
* [x] `menu` (depends on `menu-surface` and `list`) (@acdvorak, #4342)
* [x] `dialog` (note: `focus-trap` has its own `.d.ts`) (@moog16 and @acdvorak, #4333)
* [x] `chips` (@moog16 and @acdvorak, #4332)
* [x] `grid-list` (@moog16 and @acdvorak, #4337)
* [x] `snackbar` (@acdvorak, #4363)
* [x] `slider` (@acdvorak, #4378)
* [x] `line-ripple` (@acdvorak, #4373)
* [x] `floating-label` (@acdvorak, #4374)
* [x] `notched-outline` (depends on `floating-label`) (@acdvorak, #4375)
* [x] `tab-indicator` (@acdvorak, #4391)
[...]

PR #4241, opened 2019-01-10
feat(infrastructure): add tsconfig, tslint, and build setup

moog16:
Refs #4225

- added a tsconfig and tslint
- added ts-loader to webpack to get npm start/dev/build to work
- removed babylon from package.json and upgraded to @babel/parser
  - babylon is deprecated
- renamed `rewrite-declaration-statements-for-closure.js` -> `rewrite-declaration-statements-for-typescript.js`
  - no longer editting `export` statements
  - `imports` are not relative to the `.typescript-tmp/packages` directory (as per internal google requirements)
- renamed `closure-rewrite.js` -> `typescript-rewrite.js`

Tested this by running:
```
npm run lint
npm run start # hit a few pages to see if there were errros
npm run clean
npm run dist
node scripts/cp-pkgs.js
npm run dev # still needs to run (i assumed)
npm test
```

TODO: need to update webpack to build ES6 and ES5 dist files


CONTRIBUTING.md
```diff
@@ -99,9 +99,9 @@ npm run fix # Runs both of the above commands in parallel
 
 npm run test:watch # Runs karma on Chrome, re-running when source files change
 
-npm test # Lints all files, runs karma, runs closure tests, and then runs coverage enforcement checks.
+npm test # Lints all files, runs karma, runs typescript tests, and then runs coverage enforcement checks.
 npm run test:unit # Only runs the karma tests
-npm run test:closure # Runs closure build tests against all closurized files
+npm run test:typescript # Runs typescript build tests against all closurized files
```
kfranqueiro: Change "closurized"?

package.json
```diff
@@ -215,7 +220,7 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
-  "closureWhitelist": [
+  "typescriptWhitelist": [
```
kfranqueiro: Is this even necessary? I find it odd that this works as-is without changing its contents, given that only mdc-base should actually contain typescript in this PR.

packages/mdc-base/component.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016 Google Inc.
+ * Copyright 2019 Google Inc.
```
kfranqueiro: Leave copyright years as-is since we're not changing the actual innards of these modules.

packages/mdc-base/component.ts
```diff
@@ -21,55 +21,48 @@
...
     // Note that we initialize foundation here and not within the constructor's default param so that
     // this.root_ is defined and can be used within the foundation class.
-    /** @protected {!F} */
     this.foundation_ = foundation === undefined ? this.getDefaultFoundation() : foundation;
     this.foundation_.init();
     this.initialSyncWithDOM();
   }
 
-  initialize(/* ...args */) {
+  // tslint:disable-next-line:no-any a component can pass in anything it needs to the constructor
```
kfranqueiro: `the constructor` -> `init`

packages/mdc-base/component.ts
```diff
@@ -21,55 +21,48 @@
  * THE SOFTWARE.
  */
 
-import MDCFoundation from './foundation';
+import MDCFoundation from "./foundation";
```
kfranqueiro: We should absolutely not be using double quotes. AFAICT Google's style guides always promote single quotes for JS / TS.

The double quotes lint rule seems to be coming from [tslint:recommended](https://github.com/palantir/tslint/blob/master/src/configs/recommended.ts#L134). I'm surprised there's none in gts/tslint.json...

We should probably go with `'single'` to match our use.

packages/mdc-base/foundation.ts
```diff
@@ -21,44 +21,35 @@
...
-  static get numbers() {
+  static get numbers(): {} {
     // Classes extending MDCFoundation should implement this method to return an object which exports all
     // of its semantic numbers as constants. e.g. {ANIMATION_DELAY_MS: 350}
     return {};
   }
 
-  /** @return {!Object} */
-  static get defaultAdapter() {
+  static get defaultAdapter(): {} {
```
kfranqueiro: Should this be typed as AdapterType? (I might be wrong here but it stands out to me)

packages/mdc-base/foundation.ts
```diff
@@ -21,44 +21,35 @@
...
  */
 
-/**
- * @template A
- */
-class MDCFoundation {
-  /** @return enum{cssClasses} */
-  static get cssClasses() {
+class MDCFoundation<AdapterType extends {} = {}> {
+  static get cssClasses(): {} {
```
kfranqueiro: Should it be possible for us to type these more specifically using index signatures? i.e. `cssClasses` and `strings` are always `{[key: string]: string}`, and `numbers` is always `{[key: string]: number}`.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-string-index-signatures

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "alwaysStrict": true,
+    "pretty": true,
+    "preserveConstEnums": true,
+    "stripInternal": true,
+    "sourceMap": true,
+    "target": "es5",
+    "newLine": "lf",
+    "types": []
+  }
+}
```
kfranqueiro: Add newline at EOF for this and tslint.json

package.json
moog16: closureWhitelist is something unique to MDC Web (not a closure thing). I believe this is for the `typescript-rewrite.sh` file.

packages/mdc-base/component.ts
moog16: Ya, you're right. I'm surprised that gts doesn't have that. I will add that to our tslint file.

packages/mdc-base/foundation.ts
moog16: Since defaultAdapter is static, there is no way to know what AdapterType is before an instance of MDCFoundation is created.

https://github.com/Microsoft/TypeScript/issues/24018

moog16:
@kfranqueiro figured out the istanbul loader...depressing to see that getting this into the react repo took almost 3 days of experimentation, and it was just a few line changes...

package.json
moog16: It seems like we will need this when we rewrite the other components, since it uses this list to know which ones to convert to relative paths for internal usages.

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+/**
+ * @fileoverview Rewrites JS to match relative path format. That means
```
abhiomkar: Did you mean `declaration`?

`rewrite-declation-statements-for-typescript.js` => `rewrite-declaration-statements-for-typescript.js`

scripts/typescript-rewrite.sh
```diff
@@ -25,26 +25,26 @@
...
+for pkg in $TYPESCRIPT_PKGS; do
+  cp -r "packages/$pkg" $TYPESCRIPT_PKGDIR
 done
-rm -fr $CLOSURE_PKGDIR/**/{node_modules,dist}
+rm -fr $TYPESCRIPT_PKGDIR/**/{node_modules,dist}
 
-log "Rewriting all import statements to be closure compatible"
-node scripts/rewrite-decl-statements-for-closure-test.js $CLOSURE_PKGDIR
+log "Rewriting all import statements to be internal typescript compatible"
+node scripts/rewrite-declation-statements-for-typescript.js $TYPESCRIPT_PKGDIR
```
abhiomkar: typo: `declaration`

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const {default: traverse} = require('babel-traverse');
+const parser = require('@babel/parser');
+const camelCase = require('camel-case');
+const glob = require('glob');
+const recast = require('recast');
+const resolve = require('resolve');
+const t = require('babel-types');
```
abhiomkar: can we name this `types`? `t` seems very short.

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+ * import someDefaultExport from 'third-party-lib';
+ * ```
+ *
+ * becomes
+ *
+ * ```js
+ * import {default: someDefaultExport} from 'mdc.thirdparty.thirdPartyLib'
+ * ```
+ *
+ * Note that for third-party modules, they must be defined in closure_externs.js. See that file for more info.
```
kfranqueiro: This is talking about closure_externs. Is this stale?

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+      extensions: ['.ts', '.js'],
+    });
+    const srcDirectory = path.dirname(srcFile);
+    resolvedSource = path.relative(srcDirectory, fileDir);
+
+    const packageStr = resolvedSource
+      .replace('.js', '')
+      .replace('.ts', '')
+      .replace('.index', '');
+    return `./${packageStr}`;
```
kfranqueiro: I noticed there's an extra leading `./` on converted relative imports that doesn't need to be there in the output. The leading `./` here might be the cause. I'm not sure if this is necessary; if an import is already relative (and thus already has a leading `./` or `../`, would it even hit this path?

scripts/travis-env-vars.sh
```diff
@@ -78,22 +78,22 @@ print_all_changed_files
...
 if [[ "$TEST_SUITE" == 'build' ]]; then
   # Only run build if package JS/Sass files changed
-  check_for_testable_files '^packages/.+\.(js|css|scss)$'
+  check_for_testable_files '^packages/.+\.(js|ts|css|scss)$'
 fi
 
-if [[ "$TEST_SUITE" == 'closure' ]]; then
-  # Only run closure test if package JS files changed
-  check_for_testable_files '^packages/.+\.js$'
+if [[ "$TEST_SUITE" == 'typescript' ]]; then
```
kfranqueiro: Should this branch be removed? (See also comment RE removing the task in .travis.yml)

.travis.yml
```diff
@@ -54,10 +55,9 @@ matrix:
 
   - node_js: 8
     env:
-      - TEST_SUITE=closure
-      - CLOSURE=1
+      - TEST_SUITE=typescript
     script:
-      - if has_testable_files; then npm run test:closure; else log_untestable_files; fi
+      - if has_testable_files; then npm run test:typescript; else log_untestable_files; fi
```
kfranqueiro: This npm script does not exist. Should we remove this? We're already testing building the TS/JS in the build task.

scripts/webpack/js-bundle-factory.js
```diff
@@ -51,6 +51,7 @@ class JsBundleFactory {
     {
       bundleName,
       chunks,
+      extensions = ['.ts', '.js', '.json'],
```
kfranqueiro: Why `.json`?

packages/mdc-base/foundation.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016 Google Inc.
+ * Copyright 2019 Google Inc.
```
kfranqueiro: Revert year

packages/mdc-base/index.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016 Google Inc.
+ * Copyright 2019 Google Inc.
```
kfranqueiro: Revert year

scripts/check-pkg-for-release.js
```diff
@@ -212,7 +212,8 @@ function checkComponentImportedAddedInMDCPackage(ast) {
     'ImportDeclaration'({node}) {
       if (node.source) {
         const source = node.source.value;
-        if (source === pkg.name + '/index') {
+        const pkgFile = pkg.name + '/index';
+        if (source === pkgFile || source === pkgFile + '.ts') {
```
kfranqueiro: Should we be able to re-simplify this again once everything's converted to TS? If so I'd add a TODO comment here (and anywhere else you can think of, e.g. where we're currently accounting for JS + TS).

package.json
```diff
@@ -224,5 +224,8 @@
     "hooks": {
       "commit-msg": "validate-commit-msg"
     }
+  },
+  "dependencies": {
+    "karma-coverage-istanbul-reporter": "^2.0.4"
```
acdvorak: This should be in `devDependencies` (I forget to pass `--save-dev` at least 70% of the time I run `npm i`)

package.json
```diff
@@ -36,11 +38,11 @@
     "start": "npm-run-all --parallel screenshot:serve screenshot:watch",
     "test:watch": "karma start --auto-watch",
     "test:unit": "karma start --single-run",
-    "test:closure": "./scripts/closure-test.sh",
     "test:dependency": "./scripts/dependency-test.sh",
     "test:site": "npm run clean:site && ./scripts/site-generator-test.sh"
   },
   "devDependencies": {
+    "@babel/parser": "^7.2.3",
```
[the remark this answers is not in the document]
moog16: I believe the @babel/parser is separate from babel-core, but is used in babel-core v7. Also the parser is mainly used for the file import/export rewrites, and not the components so I think its safe to use this.

Looked at the upgrade https://babeljs.io/docs/en/v7-migration and it looks like we could do it fairly quickly...but don't think we should group it with the typescript updates.

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+      basedir,
+      extensions: ['.ts', '.js'],
+    });
+    const srcDirectory = path.dirname(srcFile);
+    resolvedSource = path.relative(srcDirectory, fileDir);
+
+    const packageStr = resolvedSource
+      .replace('.js', '')
+      .replace('.ts', '')
+      .replace('.index', '');
```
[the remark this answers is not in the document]
moog16: This might be a relic from the previous. I'm going to remove.

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+ *
+ *
+ * This script rewrites import statements such that:
+ *
+ * ```js
+ * import [<SPECIFIERS> from] '@material/$PKG[/files...]';
+ * ```
+ * becomes
+ * ```js
+ * import [<SPECIFIERS> from] '$PKG.<RESOLVED_FILE_NAMESPACE>';
```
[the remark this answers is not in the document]
moog16: correct...i wasn't very thorough with the comments up here

scripts/rewrite-declation-statements-for-typescript.js
moog16: removed

scripts/rewrite-declation-statements-for-typescript.js
```diff
@@ -0,0 +1,214 @@
...
+ * and
+ *
+ * ```js
+ * import someDefaultExport from 'third-party-lib';
+ * ```
+ *
+ * becomes
+ *
+ * ```js
+ * import {default: someDefaultExport} from 'mdc.thirdparty.thirdPartyLib'
```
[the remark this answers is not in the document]
moog16: let me check with rachel what they are expecting for 3rd party

scripts/rewrite-declation-statements-for-typescript.js
moog16: I put it there, since I'm more used to seeing it like that. I don't think it is necessary. I can remove.

scripts/webpack/js-bundle-factory.js
moog16: I couldn't find what resolve.extensions defaults to, but it is in the code sample of https://webpack.js.org/configuration/resolve/#resolve-extensions. It didn't seem to hurt. Although we do not have any json files we depend on in the component/foundation/adapter, I don't see why it isn't a possibility.

scripts/check-pkg-for-release.js
moog16: ya good idea...want me to also create an issue?

moog16:
@acdvorak @kfranqueiro @abhiomkar addressed all comments. Thanks for looking :)

scripts/check-pkg-for-release.js
kfranqueiro: I added a couple of post-conversion items to #4225.

karma.conf.js
```diff
@@ -72,8 +72,9 @@ module.exports = function(config) {
     concurrency: USING_SL ? 4 : Infinity,
     customLaunchers: customLaunchers,
 
-    coverageReporter: {
+    coverageIstanbulReporter: {
       dir: 'coverage',
+      skipFilesWithNoCoverage: true,
```
kfranqueiro: What's the reason for adding this? Would this risk not reporting files under `packages` that are somehow not tested at all?

karma.conf.js
```diff
@@ -98,14 +99,17 @@ module.exports = function(config) {
         // Cover source files when not debugging tests. Otherwise, omit coverage instrumenting to get
         // uncluttered source maps.
         rules: webpackConfig.module.rules.concat([config.singleRun ? {
-          test: /\.js$/,
+          enforce: 'post',
+          test: /\.(js|ts)?$/,
```
kfranqueiro: Why is there a `?` here? We should only be running on js or ts files, and nothing should end with just `.`

scripts/webpack/js-bundle-factory.js
kfranqueiro: IIRC Node.js can import JSON files so that's probably what inspires it. I'd be inclined to leave json out since I don't think we should ever have a reason to be importing one.

karma.conf.js
moog16: docs aren't super clear here `// Omit files with no statements, no functions and no branches from the report`.

I think I can remove it, since none of the test files are empty anyways.

karma.conf.js
moog16: you're right...I had an `x?` there for jsx/tsx

CONTRIBUTING.md
```diff
@@ -99,9 +99,9 @@ npm run fix # Runs both of the above commands in parallel
 
 npm run test:watch # Runs karma on Chrome, re-running when source files change
 
-npm test # Lints all files, runs karma, runs closure tests, and then runs coverage enforcement checks.
+npm test # Lints all files, runs karma, runs typescript tests, and then runs coverage enforcement checks.
 npm run test:unit # Only runs the karma tests
-npm run test:closure # Runs closure build tests against all closurized files
+npm run test:typescript # Runs typescript build tests against all typescript files
```
kfranqueiro: I should've realized this before, but this task doesn't exist, so remove this line

package.json
kfranqueiro: We discussed this offline with the rest of the team and confirmed that closureWhitelist was only necessary for the closure compilation tests within this repository, to avoid trying to compile packages that hadn't yet been tested / updated for closure compilation. Thus it's safe to remove.

scripts/rewrite-declation-statements-for-typescript.js
kfranqueiro: Turns out I was wrong about this and it was also being used for imports within the same directory, so we had to put this back.

kfranqueiro:
Matt is currently investigating an issue causing incorrect branch/function coverage reporting. (It's reporting uncovered branches that don't exist.) This is why unit tests are failing (the unit tests pass but coverage falls below threshold).

moog16:
Figured out the incorrect branch/function coverage reporting. The side effect is that `istanbul` package no longer works. This package only reported to the CLI, so it is non-essential.

I moved the statements/branches/functions/lines test into the karma.config.js, which fails if the tests limits are not met.

test/unit/index.js
```diff
@@ -33,4 +33,4 @@ afterEach(() => {
 
 // https://github.com/webpack/docs/wiki/context#requirecontext
 const testsContext = require.context(/* directory */ '.', /* useSubdirectories */ true, /\.test\.js$/);
-testsContext.keys().forEach(testsContext);
+testsContext.keys().forEach(testsContext);
```
kfranqueiro: Revert this change

karma.conf.js
```diff
@@ -50,6 +50,17 @@ const SAUCE_LAUNCHERS = {
 
 const customLaunchers = Object.assign({}, USING_SL ? SAUCE_LAUNCHERS : {}, HEADLESS_LAUNCHERS);
 const browsers = USING_TRAVISCI ? Object.keys(customLaunchers) : ['Chrome'];
+const istanbulInstrumentLoader = {
```
kfranqueiro: Nit: `istanbulInstrumenterLoader`

Seems like we could move `include` into here as well?

karma.conf.js
```diff
@@ -50,6 +50,17 @@ const SAUCE_LAUNCHERS = {
...
+  use: [{
+    loader: 'istanbul-instrumenter-loader',
+    options: {esModules: true},
+  }],
+  exclude: [
+    /node_modules/,
+    /adapter.js/,
```
kfranqueiro: Would these need to be `[jt]s` to continue ignoring these as they're converted?

Also, would a `$` at the end for the adapter/constants regexps be a good idea?

karma.conf.js
```diff
@@ -72,13 +83,20 @@ module.exports = function(config) {
...
-        {type: 'html'},
-      ],
+    coverageIstanbulReporter: {
+      'dir': 'coverage',
+      'reports': ['html', 'lcovonly', 'json'],
+      'report-config': {
+        lcovonly: {subdir: '.'},
+        json: {subdir: '.', file: 'coverage.json'},
+      },
+      'emitWarning': false,
```
kfranqueiro: Add a comment explaining this causes failing on coverage below threshold. (I had to go look this up, this is not a well-phrased option name)

scripts/webpack/js-bundle-factory.js
```diff
@@ -62,9 +65,18 @@ class JsBundleFactory {
...
       plugins = [],
+      tsConfigFilePath = path.resolve(__dirname, '../../tsconfig.json'),
     }) {
     chunks = chunks || this.globber_.getChunks({inputDirectory, filePathPattern});
 
+    const babelLoader = {
+      loader: 'babel-loader',
+      options: {
+        cacheDirectory: true,
+        presets: [['es2015', {modules: false}]],
```
kfranqueiro: Isn't this picked up from the `babel` property in our `package.json`? We have other babel configuration there as well which makes me concerned whether all of it is being picked up correctly...

karma.conf.js
```diff
@@ -93,20 +111,17 @@ module.exports = function(config) {
...
     webpack: Object.assign({}, webpackConfig, {
-      devtool: 'inline-source-map',
       module: Object.assign({}, webpackConfig.module, {
         // Cover source files when not debugging tests. Otherwise, omit coverage instrumenting to get
         // uncluttered source maps.
-        rules: webpackConfig.module.rules.concat([config.singleRun ? {
-          test: /\.js$/,
+        rules: webpackConfig.module.rules.concat(config.singleRun ? [Object.assign({
+          enforce: 'post',
+          test: /\.(ts)$/,
```
kfranqueiro: The parens are no longer necessary in these rules.

scripts/webpack/js-bundle-factory.js
moog16: I removed the options here. package.json's options do no have cacheDirectory. Definition reads:

```
cacheDirectory: Default false. When set, the given directory will be used to cache the results of the loader. Future webpack builds will attempt to read from the cache to avoid needing to run the potentially expensive Babel recompilation process on each run. If the value is blank (loader: 'babel-loader?cacheDirectory') or true (loader: 'babel-loader?cacheDirectory=true'), the loader will use the default cache directory in node_modules/.cache/babel-loader or fallback to the default OS temporary file directory if no node_modules folder could be found in any root directory.```

I think this can be applied to any of the builds/tests/servers...so I will add it to the package.json config.

PR #4243, opened 2019-01-11
feat(chips): Move logic for calculating chip bounding rect into a foundation method

rlfriedman:
This change makes adding ripple easier for clients wrapping chips.

BREAKING CHANGE: Adds 3 new chips adapter methods: hasLeadingIcon, getRootBoundingClientRect, and getCheckmarkBoundingClientRect. Also adds a new foundation method: getDimensions.

### DIFF
--- a/packages/mdc-chips/chip/foundation.js
+++ b/packages/mdc-chips/chip/foundation.js
@@ -60,6 +60,9 @@ class MDCChipFoundation extends MDCFoundation {
       notifyRemoval: () => {},
       getComputedStyleValue: () => {},
       setStyleProperty: () => {},
+      hasLeadingIcon: () => {},
+      getBoundingClientRect: () => {},
+      getCheckmarkBoundingClientRect: () => {},
     });
   }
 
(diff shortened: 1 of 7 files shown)


packages/mdc-chips/chip/foundation.js
```diff
@@ -109,6 +112,24 @@ class MDCChipFoundation extends MDCFoundation {
...
   }
 
+  /**
+   * Computes and returns the bounding rect for the chip.
+   * @return {!ClientRect}
+   */
+  computeBoundingRect() {
+    // When a chip has a checkmark and not a leading icon, the bounding rect changes in size depending on the current
+    // size of the checkmark.
+    if (!this.adapter_.hasLeadingIcon() && this.adapter_.getCheckmarkBoundingClientRect()) {
```
moog16: `this.adapter.getCheckmarkBoundingClientRect()` returns an object...I think it is more precise to check that the function call returns an object, instead of just calling the function.

packages/mdc-chips/chip/foundation.js
```diff
@@ -60,6 +60,9 @@ class MDCChipFoundation extends MDCFoundation {
       notifyRemoval: () => {},
       getComputedStyleValue: () => {},
       setStyleProperty: () => {},
+      hasLeadingIcon: () => {},
+      getBoundingClientRect: () => {},
```
moog16: I think adding `root` in the method makes it more clear. `getRootBoundingClientRect`

packages/mdc-chips/chip/index.js
```diff
@@ -71,24 +73,12 @@ class MDCChip extends MDCComponent {
...
-          const width = this.root_.getBoundingClientRect().width + checkmarkEl.getBoundingClientRect().height;
-          return {height, width};
-        },
-      });
-      this.ripple_ = rippleFactory(this.root_, new MDCRippleFoundation(adapter));
-    } else {
-      this.ripple_ = rippleFactory(this.root_);
-    }
+    const adapter = Object.assign(MDCRipple.createAdapter(this), {
+      computeBoundingRect: () => this.foundation_.computeBoundingRect(),
```
moog16: what about `this.foundation_.computeBoundingRectForRipple()`?

packages/mdc-chips/chip/foundation.js
rlfriedman: Changed it to check to confirm it isn't null, does that work?

packages/mdc-chips/chip/index.js
rlfriedman: Updated to getDimensions

packages/mdc-chips/chip/foundation.js
rlfriedman: Done

PR #4261, opened 2019-01-18
chore(infrastructure): Detect already-running screenshot build process

### LINKED ISSUE #3930: Screenshot tests rebuild assets even when a build is already running
## Bug

PR #3732 inadvertently broke the ability of screenshot tests to detect that a build is already running.

### Steps to reproduce:

1. Run `npm start` in one terminal
2. Run `npm run screenshot:test` in another terminal

### Expected behavior:

The screenshot test should...

1. Detect that a build is already running in another process
2. Skip building JS/CSS assets

### Actual behavior:

The screenshot test...

1. Fails to detect that a build is already running in another process
2. Rebuilds JS/CSS assets every time

## Root cause

To detect if a build is already running, we use the cross-platform [`ps-node` library](https://www.npmjs.com/package/ps-node) to scan for a `node` process that has `run.js` and `build` arguments:

https://github.com/material-components/material-components-web/blob/75cf88c5607aae69817133dca9277ff46da8870b/test/screenshot/infra/commands/build.js#L166-L175

However, users can run multiple builds simultaneously in different cloned repos (e.g., during a release or while testing someone's PR), so we need to make sure that we only look for processes running in the same `$PWD` as the current process.

Unfortunately, `ps-node` doesn't expose `$PWD`—all it gives us are PID, command, and arguments:

```json
{
  "pid": 67507,
  "command": "node",
  "arguments": [ "test/screenshot/run.js", "build", "--watch" ],
  "ppid": 67500
}
```

I noticed that `arguments` contained the _relative_ path to `run.js`, so I added `$PWD` directly before the relative paths in `package.json`:

https://github.com/material-components/material-components-web/blob/54ccf8953c6aa21663ede68aa50754173540d4ea/package.json#L26-L34

This gave me _absolute_ paths from `ps-node`:

```json
{
  "pid": 67507,
  "command": "node",
  "arguments": [ "/Users/advorak/dev/mdc-web/test/screenshot/run.js", "build", "--watch" ],
  "ppid": 67500
}
```

I could then test whether the first argument starts with the current `node` process's `$PWD`:

https://github.com/material-components/material-components-web/blob/75cf88c5607aae69817133dca9277ff46da8870b/test/screenshot/infra/lib/process-manager.js#L116

This was inadvertently broken when PR #3732 removed `$PWD` from `package.json`.

## Potential solutions

1. Write to a `.pidfile` instead of scanning running processes with `ps-node`
2. See if any NPM modules return the PWD of another process (ideally w/ Windows support)

Option 1 seems like the simplest and most robust solution, and it would support Windows.

acdvorak:
Fixes #3930


test/screenshot/infra/types/node-externs.js
```diff
@@ -81,8 +81,8 @@
 
 /**
  * @typedef {{
- *   pid: number,
- *   ppid: number,
+ *   pid: string,
```
acdvorak: FYI: The `ps-node` API actually returns a `string` value for both `pid` and `ppid`.

I was previously calling `parseInt()` and overwriting the `string` values with `number`, but that's now removed.

test/screenshot/infra/lib/process-manager.js
```diff
@@ -25,9 +25,18 @@
...
+const LocalStorage = require('../lib/local-storage');
 const {ExitCode} = require('../lib/constants');
 
 class ProcessManager {
+  constructor() {
+    /**
+     * @type {!LocalStorage}
+     * @private
+     */
+    this.locaStorage_ = new LocalStorage();
```
kfranqueiro: Typo, supposed to be `localStorage_`?

test/screenshot/infra/commands/build.js
```diff
@@ -59,6 +59,7 @@ class BuildCommand {
       return;
     }
 
+    await this.setRunningProcessId_();
```
kfranqueiro: When is this ever cleared? If I'm running `npm start`, the file gets created, then when I kill it, the file never gets deleted. Wouldn't that cause a false positive?

acdvorak: The file never gets deleted, but I avoid false positives by checking if the process is actually still running on startup.

I could conceivably delete the file within a process `exit` handler, with the caveat that it obviously won't get called if the node process abends. WDYT?

test/screenshot/infra/lib/process-manager.js
acdvorak: Done

test/screenshot/infra/commands/build.js
kfranqueiro: I missed the ps check before. Is it possible for us to verify that the running process is actually what we think it is and not coincidentally another process that spawned with the same ID?

acdvorak: I tried to do that by making sure the running process is `node` with an argument of `build`:

https://github.com/material-components/material-components-web/blob/0b46c6b7a057e12fb1caab08c718fd281afb01c1/test/screenshot/infra/commands/build.js#L166-L168

https://github.com/material-components/material-components-web/blob/0b46c6b7a057e12fb1caab08c718fd281afb01c1/test/screenshot/infra/lib/process-manager.js#L108-L113

kfranqueiro: Oh that's what commandName and argumentPattern end up being. OK.

acdvorak: I added a bit of code to delete the PID file when the process exits.

PR #4264, opened 2019-01-18
chore(infrastructure): Dynamic screenshot test viewport size

### LINKED ISSUE #4263: Screenshot test viewport size should be dynamic
### TL;DR

Instead of hard-coding a list of components with "small" viewports, we should check each individual HTML file for a `test-viewport--mobile` class at runtime.

We should use [`jsdom`](https://www.npmjs.com/package/jsdom) to `querySelector()` the page locally (on the host machine) instead of doing unreliable `string.indexOf()` checks or making async Selenium requests.

### Background

Screenshot tests have two viewport sizes:

1. Mobile (aka "small")
2. Desktop (aka "large")

Taking screenshots at the fullscreen desktop resolution of `1350x636` (or whatever) is expensive.
Large images take longer to send over the network (1–2 seconds per image), which can easily double the test time.

To avoid unnecessary slowness, we use small, "mobile"-sized viewports for as many components as possible (button, checkbox, text-field, etc.). A few components _do_ need to be tested at desktop resolutions, though (dialog, drawer, etc.).

Resizing the browser window is also expensive: The RTT is 1–2 seconds. To avoid doubling the test time, we batch up all "small" component URLs and capture those, then resize the browser and capture all the "large" component URLs.

### Problem

You might expect that adding the `test-viewport--mobile` class to a screenshot test page would cause the page to be rendered in a "small" browser window:

https://github.com/material-components/material-components-web/blob/9d46c200c4e205ebcad2ff226b29be216b4043bb/test/screenshot/spec/mdc-button/classes/baseline-button-with-icons.html#L42-L49

But no! The "logic" to determine whether a given URL is a "small" or "large" component is just a hard-coded list that will inevitably get out of sync with what's actually in the HTML:

https://github.com/material-components/material-components-web/blob/9d46c200c4e205ebcad2ff226b29be216b4043bb/test/screenshot/infra/lib/selenium-api.js#L664-L671

Adding the `test-viewport--mobile` class to a screenshot test page simply displays a green mobile-viewport-sized border around the components. This tells you how many components you can fit in a mobile viewport before you run the test, but it doesn't affect Selenium browser size.

This is counterintuitive, not discoverable, and easy to get out of date.

### Solution

Instead of hard-coding a list of URLs for "small" components, we should just check the page to see if it has a `test-viewport--mobile` class. Many pages already use it.

We could potentially do a simple `html.indexOf('test-viewport--mobile') > -1` check, but I worry that it would generate false positives for things like commented-out code:

```html
 <body class="test-container"> 
   <!-- TODO: Should we add test-viewport--mobile here? -->
   <main class="test-viewport"> 
     <div class="test-layout"> 
```

Instead, we should use a DOM library like [`jsdom`](https://www.npmjs.com/package/jsdom) to run a `querySelector('test-viewport--mobile')` check on each page, locally (on the host machine). This will be much faster than remote async calls to Selenium and more robust than an `indexOf()` check.

acdvorak:
Fixes #4263

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "istanbul": "^0.4.4",
     "istanbul-instrumenter-loader": "^3.0.0",
     "jimp": "^0.2.28",
+    "jsdom": "^13.1.0",
     "json-stable-stringify": "^1.0.1",
     "karma": "^3.0.0",
     "karma-chrome-launcher": "^2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -287,6 +287,12 @@
         "through": ">=2.2.7 <3"
       }
     },
+    "abab": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.0.tgz",
+      "integrity": "sha512-sY5AXXVZv4Y1VACTtR11UJCPHHudgY5i26Qj5TypE6DKlIApbwb5uqhXcJ5UUGbvZNRh7EeIoW+LrJumBsKp7w==",
+      "dev": true
+    },
     "abbrev": {
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -89,6 +89,7 @@
     "istanbul": "^0.4.4",
     "istanbul-instrumenter-loader": "^3.0.0",
     "jimp": "^0.2.28",
+    "jsdom": "^13.1.0",
```
kfranqueiro: Is it possible to use cheerio for this instead of jsdom? I suspect it will be much faster/lighter.

kfranqueiro: Andy tested cheerio and the difference was far smaller than I expected - about 4 seconds across all files. So it probably doesn't matter which one we use. Andy also pointed out that jsdom also executes scripts which might be useful if we ever end up adding the class dynamically (e.g. due to using templates).

PR #4265, opened 2019-01-18
fix(drawer): Use parentNode DOM API when selecting scrim to make it work with Shadow DOM

### LINKED ISSUE #4254: MDCDrawer can't find the scrim when initialising the element inside the Shadow DOM
Hi,

While trying to implement the modal version of the MDCDrawer, I hit the following exception:

screen shot 2019-01-16 at 17 29 32

I've created a demonstration [here](https://indigo-hunter.glitch.me), you can find the source [here](https://glitch.com/edit/#!/indigo-hunter?path=src/main-app.js:80:83).

Any idea?
Thanks.

abhiomkar:
Fixes #4254


packages/mdc-drawer/index.js
```diff
@@ -110,7 +110,7 @@ class MDCDrawer extends MDCComponent {
 
     if (this.root_.classList.contains(MODAL)) {
       const {SCRIM_SELECTOR} = MDCDismissibleDrawerFoundation.strings;
-      this.scrim_ = /** @type {!Element} */ (this.root_.parentElement.querySelector(SCRIM_SELECTOR));
+      this.scrim_ = /** @type {!Element} */ (this.root_.parentNode.querySelector(SCRIM_SELECTOR));
```
acdvorak: Is there any way we can add a unit test for this use case?
E.g., detach `root_` from the DOM and append it to a `DocumentFragment`?

abhiomkar: Shadow DOM is not fully supported in all the browsers. Setting up test environment with all the polyfills to render Shadow DOM and initializing custom element is not worth of effort to test this?

acdvorak: We actually don't need to test in a "real" Shadow DOM; just move the element into a `DocumentFragment`:

```js
// Fake Shadow DOM
const frag = document.createDocumentFragment();

// This will automatically remove the element from its parent;
// no need to remove it manually.
frag.appendChild(drawer.root_);
```

Shadow DOM roots are just `DocumentFragment` objects, so the above is functionally equivalent to testing in a real Shadow DOM.

abhiomkar: `drawer.root_` doesn't include scrim element so I went ahead and changed the fixture setup to include `DocumentFragment` element to use this in test case.

Thanks for the suggestion!

PR #4267, opened 2019-01-18
refactor(tab-scroller): Fix a Closure type declaration in the Tab Scroller util.

bwobrien:
getMatchesProperty returns Array.pop() from an Array of strings, so the return value should be string.


bwobrien:
Thanks for pointing out MDC DOM, I will switch to that in the future.

PR #4269, opened 2019-01-18
feat(dom): Convert JS to TypeScript

acdvorak:
Refs #4225

### DIFF
--- a/packages/mdc-dom/index.js
+++ b/packages/mdc-dom/index.js
@@ -21,6 +21,6 @@
  * THE SOFTWARE.
  */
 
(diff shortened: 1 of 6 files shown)


packages/mdc-dom/index.js
```diff
@@ -21,6 +21,6 @@
  * THE SOFTWARE.
  */
 
-import * as ponyfill from './ponyfill';
```
moog16: this file should be switched to a .ts file as well

acdvorak: Done

PR #4273, opened 2019-01-19
feat(menu-surface): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-menu-surface/index.ts
```diff
@@ -146,74 +122,43 @@ class MDCMenuSurface extends MDCComponent {
...
-   * focusLastElement: function(),
-   * }}
-   * @private
-   */
-  getFocusAdapterMethods_() {
-    return {
       isFocused: () => document.activeElement === this.root_,
       saveFocus: () => {
-        this.previousFocus_ = document.activeElement;
+        this.previousFocus_ = document.activeElement as HTMLElement;
```
moog16: what if `previousFocus` is a union. `Element | HTMLElement | SVGElement`? If that's the case, then I think firstFocusableElement/lastFocustableElement can also be that type.

packages/mdc-menu-surface/util.ts
```diff
@@ -21,16 +21,12 @@
...
+let storedTransformPropertyName_: string | undefined;
 
 /**
  * Returns the name of the correct transform property to use on the current browser.
- * @param {!Window} globalObj
- * @param {boolean=} forceRefresh
- * @return {string}
  */
-function getTransformPropertyName(globalObj, forceRefresh = false) {
+function getTransformPropertyName(globalObj: Window, forceRefresh = false): string {
```
moog16: The return type should probably just be `transform | webkitTransform`, right? What are the other options?

packages/mdc-menu-surface/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+  /** Emits an event when the menu surface is closed. */
+  notifyClose(): void;
+
+  /** Emits an event when the menu surface is opened. */
+  notifyOpen(): void;
+}
+
+// TODO(acdvorak): Should properties be readonly?
+
+interface MDCMenuSurfacePositionInput {
```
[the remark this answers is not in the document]
mmalerba: If you leave the off the `readonly` here, you can decide based on the context where you're using it by just using `MDCMenuSurfacePositionInput` when you want read/write and using `Readonly<MDCMenuSurfacePositionInput>` when you want read only.

packages/mdc-menu-surface/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+ * for more details.
+ * https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md
+ */
+interface MDCMenuSurfaceAdapter {
+  addClass(className: string): void;
+  removeClass(className: string): void;
+  hasClass(className: string): boolean;
+  hasAnchor(): boolean;
+
+  isElementInContainer(el: Element): boolean;
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-menu-surface/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+  focusLastElement(): void;
+
+  /** Emits an event when the menu surface is closed. */
+  notifyClose(): void;
+
+  /** Emits an event when the menu surface is opened. */
+  notifyOpen(): void;
+}
+
+// TODO(acdvorak): Should properties be readonly?
```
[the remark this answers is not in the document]
acdvorak: I removed all `readonly` keywords, as the existing logic overwrites values in several places.

packages/mdc-menu-surface/adapter.ts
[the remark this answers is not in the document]
acdvorak: I removed all `readonly` keywords, as the existing logic overwrites values in several places.

Thanks for the tip, @mmalerba!

packages/mdc-menu-surface/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+
+interface MDCMenuSurfacePositionInput {
+  /*readonly*/ top?: number | string;
+  /*readonly*/ right?: number | string;
+  /*readonly*/ bottom?: number | string;
+  /*readonly*/ left?: number | string;
+
+  // TODO(acdvorak): Remove this and find a better way than for...in with hasOwnProperty().
+  [key: string]: number | string;
```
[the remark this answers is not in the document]
acdvorak: Removed

packages/mdc-menu-surface/foundation.ts
```diff
@@ -21,120 +21,112 @@
...
+  MDCMenuSurfacePositionInput,
+  MDCMenuSurfacePositionOutput,
+  MDCMenuSurfaceSize,
+} from './adapter';
+import {Corner, CornerBit, cssClasses, numbers, strings} from './constants';
+
+// TODO(acdvorak): Remove this alias
+type AnchorMargin = MDCMenuSurfacePositionOutput;
+
+interface AutoLayoutMeasurements {
```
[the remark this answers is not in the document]
acdvorak: This type is only used internally by the foundation in private methods, so it probably doesn't need to be exported 😀

packages/mdc-menu-surface/foundation.ts
```diff
@@ -21,120 +21,112 @@
...
-   * types.
-   * @return {!MDCMenuSurfaceAdapter}
+   * {@see MDCMenuSurfaceAdapter} for typing information on parameters and return types.
    */
-  static get defaultAdapter() {
-    return /** @type {!MDCMenuSurfaceAdapter} */ ({
+  static get defaultAdapter(): MDCMenuSurfaceAdapter {
+    // tslint:disable:no-empty
+    return {
       addClass: () => {},
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-menu-surface/foundation.ts
```diff
@@ -158,72 +150,111 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
...
+    this.isOpen_ = true;
+  }
 
   /**
-   * Handle keys that close the surface.
-   * @param {!Event} evt
+   * Closes the menu surface.
    */
-  handleKeydown(evt) {
+  close() {
```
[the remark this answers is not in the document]
acdvorak: I had to reorder several methods to satisfy tslint's rule about public methods coming first, which screwed up the diff.

packages/mdc-menu-surface/foundation.ts
```diff
@@ -372,61 +400,64 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
...
-    if (this.hoistedElement_ || this.isFixedPosition_) {
-      position = this.adjustPositionForHoistedElement_(position);
+    if (this.isHoistedElement_ || this.isFixedPosition_) {
+      this.adjustPositionForHoistedElement_(position);
     }
 
     for (const prop in position) {
       if (position.hasOwnProperty(prop) && position[prop] !== '0') {
-        position[prop] = `${parseInt(position[prop], 10)}px`;
+        // tslint:disable-next-line:ban
```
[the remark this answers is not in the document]
acdvorak: `parseInt()`. Luckily, I was able to remove all the `parseInt()` calls by modifying several private methods to pass `number` values to each other instead of `string`s with `"px"` at the end.

packages/mdc-menu-surface/index.ts
```diff
@@ -21,44 +21,48 @@
  * THE SOFTWARE.
  */
 
+// TODO(acdvorak): Should we disable object-literal-sort-keys globally?
```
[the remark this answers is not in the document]
acdvorak: SG. Internal teams use `tslint:disable` comments, so I'll do the same (sparingly).

packages/mdc-menu-surface/index.ts
```diff
@@ -70,8 +74,8 @@ class MDCMenuSurface extends MDCComponent {
       this.setFixedPosition(true);
     }
 
-    this.handleKeydown_ = (evt) => this.foundation_.handleKeydown(evt);
-    this.handleBodyClick_ = (evt) => this.foundation_.handleBodyClick(evt);
+    this.handleKeydown_ = (evt) => this.foundation_.handleKeydown(evt as KeyboardEvent);
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-menu-surface/index.ts
```diff
@@ -146,74 +122,43 @@ class MDCMenuSurface extends MDCComponent {
...
     this.foundation_.setAnchorCorner(corner);
   }
 
-  /**
-   * @param {!AnchorMargin} margin
-   */
-  setAnchorMargin(margin) {
-    this.foundation_.setAnchorMargin(margin);
+  setAnchorMargin(margin: AnchorMargin) {
+    this.foundation_.setAnchorMargin(margin);  //
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-menu-surface/index.ts
```diff
@@ -146,74 +122,43 @@ class MDCMenuSurface extends MDCComponent {
...
   }
 
-  /** @param {boolean} quickOpen */
-  set quickOpen(quickOpen) {
-    this.foundation_.setQuickOpen(quickOpen);
-  }
+  getDefaultFoundation(): MDCMenuSurfaceFoundation {
+    // TODO(acdvorak): Should we change the type of root_ to HTMLElement? What other kind of element could it be?
```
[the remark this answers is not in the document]
acdvorak: The only situation I can think of where we might have a non-HTML element is ripples on SVG icons. However, `SVGElement` appears to have most of the same properties and methods as `HTMLElement` (`style`, `querySelector()`, etc.), so it's _probably_ safe to update `MDCComponent`'s `root_` property type to `HTMLElement`.

WDYT?

packages/mdc-menu-surface/index.ts
acdvorak: Done

packages/mdc-menu-surface/util.ts
acdvorak: Done

packages/mdc-menu-surface/index.ts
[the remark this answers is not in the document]
kfranqueiro: I remember when we were annotating for Closure compiler, HTMLElement was more problematic, possibly due to many DOM APIs applying more generically to (and returning) Elements. I'd be curious if we end up running into similar problems here or not.

packages/mdc-menu-surface/adapter.ts
```diff
@@ -21,7 +21,7 @@
  * THE SOFTWARE.
  */
 
-/* eslint no-unused-vars: [2, {"args": "none"}] */
+import {MenuDimensions, MenuDistance, MenuPoint} from './constants';
 
 /**
  * Adapter for MDCMenuSurface. Provides an interface for managing
```
moog16: remove 
```
* Additionally, provides type information for the adapter to the Closure
 * compiler.
 *	
```

We've also been removing the list of 

```
* - classes
 * - dom
 * - focus
 * - position
 * - dimensions
 * - event handlers
```

but I dont think its necessary

packages/mdc-menu-surface/constants.ts
```diff
@@ -21,7 +21,8 @@
  * THE SOFTWARE.
  */
 
-/** @enum {string} */
+// tslint:disable:object-literal-sort-keys no-bitwise
```
moog16: does this disable the entire file? Do we need this?

packages/mdc-menu-surface/index.ts
```diff
@@ -0,0 +1,207 @@
...
+    return new MDCMenuSurface(root);
+  }
+
+  anchorElement!: Element | null;
+
+  private previousFocus_!: HTMLElement | SVGElement | null;
+  private firstFocusableElement_!: HTMLElement | SVGElement | null;
+  private lastFocusableElement_!: HTMLElement | SVGElement | null;
+
+  private handleKeydown_!: EventListener;
```
moog16: Use `SpecificEventListener<'keydown'>` from `@material/dom/index` for both handleKeydown and `SpecificEventListener<'click'>` for handleBodyClick

packages/mdc-menu-surface/index.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.anchorElement = hasParentAnchor ? parentEl : null;
+    this.previousFocus_ = null;
+    this.firstFocusableElement_ = null;
+    this.lastFocusableElement_ = null;
+
+    if (this.root_.classList.contains(cssClasses.FIXED)) {
+      this.setFixedPosition(true);
+    }
+
+    this.handleKeydown_ = (evt) => this.foundation_.handleKeydown(evt as KeyboardEvent);
```
moog16: if you use `SpecificEventListener<'keydown'>`, you can remove the cast and just have `(evt) => this.foundation_.handleKeydown(evt)`

This does require us to define `root_!: HTMLElement`

packages/mdc-menu-surface/index.ts
```diff
@@ -0,0 +1,207 @@
...
+  setAnchorCorner(corner: Corner) {
+    this.foundation_.setAnchorCorner(corner);
+  }
+
+  setAnchorMargin(margin: Partial<MenuDistance>) {
+    this.foundation_.setAnchorMargin(margin);
+  }
+
+  getDefaultFoundation(): MDCMenuSurfaceFoundation {
+    const rootEl = this.root_ as HTMLElement;
```
moog16: if we define `root_!: HTMLElement`, then this case can be removed

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+  static get numbers() {
+    return numbers;
+  }
+
+  static get Corner() {
+    return Corner;
+  }
+
+  /**
+   * {@see MDCMenuSurfaceAdapter} for typing information on parameters and return types.
```
moog16: is this a closure comment?

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+
+  destroy() {
+    clearTimeout(this.openAnimationEndTimerId_);
+    clearTimeout(this.closeAnimationEndTimerId_);
+    // Cancel any currently running animations.
+    cancelAnimationFrame(this.animationRequestId_);
+  }
+
+  /**
+   * @param corner Default anchor corner alignment of top-left menu surface corner.
```
moog16: remove closure annotation. Can keep the rest of the comment though.

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+   */
+  setAnchorCorner(corner: Corner) {
+    this.anchorCorner_ = corner;
+  }
+
+  /**
+   * @param margin Set of margin values from anchor.
```
moog16: remove closure annotation. Can keep the rest of the comment though.

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+  private measurements_!: AutoLayoutMeasurements; // initialized in open()
+  private isQuickOpen_: boolean;
+  private isHoistedElement_: boolean;
+  private isFixedPosition_: boolean;
+  private position_: MenuPoint;
+
+  constructor(adapter: MDCMenuSurfaceAdapter) {
+    super(Object.assign(MDCMenuSurfaceFoundation.defaultAdapter, adapter));
+
+    this.isOpen_ = false;
```
moog16: these can all be defined inline, since they are primitives. See lines: https://github.com/material-components/material-components-web/pull/4319/files#diff-66c8b5f3b2e270459927908be746faeeR56

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+      this.adjustPositionForHoistedElement_(position);
+    }
+
+    this.adapter_.setTransformOrigin(`${horizontalAlignment} ${verticalAlignment}`);
+    this.adapter_.setPosition(position);
+    this.adapter_.setMaxHeight(maxMenuSurfaceHeight ? maxMenuSurfaceHeight + 'px' : '');
+  }
+
+  /**
+   * @return Measurements used to position menu surface popup.
```
moog16: remove closure comments

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+        (avoidHorizontalOverlap && !isAlignedRight && leftOverflow < 0) ||
+        (rightOverflow > 0 && leftOverflow < rightOverflow)) {
+      corner |= CornerBit.RIGHT;
+    }
+
+    return corner;
+  }
+
+  /**
+   * @param corner Origin corner of the menu surface.
```
moog16: remove closure comments

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+      if (this.anchorCorner_ & CornerBit.BOTTOM) {
+        maxHeight -= this.measurements_.anchorSize.height;
+      }
+    }
+
+    return maxHeight;
+  }
+
+  /**
+   * @param corner Origin corner of the menu surface.
```
moog16: remove closure comments

packages/mdc-menu-surface/foundation.ts
```diff
@@ -0,0 +1,500 @@
...
+      }
+
+      return rightOffset;
+    }
+
+    return avoidHorizontalOverlap ? anchorSize.width - this.anchorMargin_.right : this.anchorMargin_.left;
+  }
+
+  /**
+   * @param corner Origin corner of the menu surface.
```
moog16: remove closure comments

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -26,7 +26,7 @@ import bel from 'bel';
 import domEvents from 'dom-events';
 import td from 'testdouble';
 
-import {MDCMenuSurface, MDCMenuSurfaceFoundation} from '../../../packages/mdc-menu-surface/index';
+import {MDCMenuSurface, MDCMenuSurfaceFoundation} from '../../../packages/mdc-menu-surface';
```
moog16: revert change

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -33,7 +33,7 @@ import {MDCSelect} from '../../../packages/mdc-select/index';
 import {cssClasses, strings} from '../../../packages/mdc-select/constants';
 import {MDCNotchedOutline} from '../../../packages/mdc-notched-outline/index';
 import {MDCMenu, MDCMenuFoundation} from '../../../packages/mdc-menu/index';
-import {MDCMenuSurfaceFoundation} from '../../../packages/mdc-menu-surface/index';
+import {MDCMenuSurfaceFoundation} from '../../../packages/mdc-menu-surface';
```
moog16: revert this change. It probably needs to be `index.ts`

packages/mdc-menu-surface/foundation.ts
mmalerba: You can (and probably should) still use doc comments like this, you just don't need to specify the type like you would for JS. You can use [TSDoc](https://github.com/Microsoft/tsdoc) to generate docs from it.

packages/mdc-menu-surface/adapter.ts
acdvorak: Done

packages/mdc-menu-surface/constants.ts
acdvorak: Removed file-level `tslint:disable` comments in favor of block- and line-level comments.

packages/mdc-menu-surface/index.ts
acdvorak: Done

packages/mdc-menu-surface/index.ts
acdvorak: Done

packages/mdc-menu-surface/index.ts
acdvorak: Done

packages/mdc-menu-surface/foundation.ts
acdvorak: This comment already existed in `foundation.js`. Updated it to use proper [JSDoc syntax](http://usejsdoc.org/tags-see.html).

packages/mdc-menu-surface/foundation.ts
acdvorak: This is actually [TypeDoc](https://typedoc.org/guides/doccomments/) syntax 😀

packages/mdc-menu-surface/foundation.ts
acdvorak: This is actually [TypeDoc](https://typedoc.org/guides/doccomments/) syntax 😀

packages/mdc-menu-surface/foundation.ts
acdvorak: This is actually [TypeDoc](https://typedoc.org/guides/doccomments/) syntax 😀

packages/mdc-menu-surface/foundation.ts
acdvorak: This is actually [TypeDoc](https://typedoc.org/guides/doccomments/) syntax 😀

packages/mdc-menu-surface/foundation.ts
acdvorak: This is actually [TypeDoc](https://typedoc.org/guides/doccomments/) syntax 😀

packages/mdc-menu-surface/foundation.ts
acdvorak: This is actually [TypeDoc](https://typedoc.org/guides/doccomments/) syntax 😀

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
acdvorak: The tests pass as-is, so why does it matter?

test/unit/mdc-select/mdc-select-enhanced.test.js
acdvorak: The tests pass as-is, so why does it matter?

packages/mdc-menu-surface/foundation.ts
acdvorak: Nice! Good call. It looks like you can assign object literal values too:

```js
  private anchorCorner_: Corner = Corner.TOP_START;
  private anchorMargin_: MenuDistance = {top: 0, right: 0, bottom: 0, left: 0};
  private position_: MenuPoint = {x: 0, y: 0};
```

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
kfranqueiro: Without `/index`, any node toolchain will resolve this to our UMD code under `dist` (due to `main` in package.json) which is probably not what they want.

acdvorak: Yeah but this is a unit test, not a source file, and we're pointing to `packages/mdc-menu-surface`, not `@material/menu-surface` (so there is no `dist`). Is it really still a problem?

I see your point for source files, though 🙂

packages/mdc-menu-surface/adapter.ts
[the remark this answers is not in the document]
acdvorak: We talked offline, and decided to stick with `Element` for our public APIs.

packages/mdc-menu-surface/README.md
```diff
@@ -152,8 +152,10 @@ Mixin | Description
 
 Constant Name | Description
 --- | ---
-`Corner` | Enum for representing an element corner for positioning the menu-surface. See [constants.js](./constants.js).
-`AnchorMargin` | Margin values representing the distance from anchor point that the menu surface should be shown. See [foundation.js](./foundation.js).
+`Corner` | Enum for representing an element corner for positioning the menu-surface. See [constants.js](constants.ts).
```
kfranqueiro: On all 4 lines here, the link was updated to .ts but the text still says .js

packages/mdc-menu-surface/README.md
```diff
@@ -204,25 +206,25 @@ Method Signature | Description
...
+`setPosition(position: Partial<MenuDistance>) => void` | Sets the position of the menu surface element.
 `setMaxHeight(value: string) => void` | Sets `max-height` style for the menu surface element.
 
 ### `MDCMenuSurfaceFoundation`
 
 Method Signature | Description
 --- | ---
-`setAnchorCorner(corner: Corner) => void` | Sets the corner that the menu surface will be anchored to. See [constants.js](./constants.js)
-`setAnchorMargin(margin: AnchorMargin) => void` | Sets the distance from the anchor point that the menu surface should be shown.
+`setAnchorCorner(corner: Corner) => void` | Sets the corner that the menu surface will be anchored to. See [constants.js](constants.ts)
```
kfranqueiro: .js -> .ts here too

packages/mdc-menu-surface/constants.ts
```diff
@@ -31,55 +30,74 @@ const cssClasses = {
...
+  TOP_RIGHT = CornerBit.RIGHT,
+  BOTTOM_LEFT = CornerBit.BOTTOM,
+  BOTTOM_RIGHT = CornerBit.BOTTOM | CornerBit.RIGHT, // tslint:disable-line:no-bitwise
+  TOP_START = CornerBit.FLIP_RTL,
+  TOP_END = CornerBit.FLIP_RTL | CornerBit.RIGHT, // tslint:disable-line:no-bitwise
+  BOTTOM_START = CornerBit.BOTTOM | CornerBit.FLIP_RTL, // tslint:disable-line:no-bitwise
+  BOTTOM_END = CornerBit.BOTTOM | CornerBit.RIGHT | CornerBit.FLIP_RTL, // tslint:disable-line:no-bitwise
+}
+
+interface MenuDimensions {
```
kfranqueiro: Is `constants.ts` where we've been putting this sort of thing? It seems like an odd idea for interfaces to be here. Would putting them in `adapter.ts` make more sense?

acdvorak: I went back and forth on that myself. @moog16 and I decided to move them from `adapter.ts` to `constants.ts` because the constants file already has enums, which feel "type-y".

We were thinking that any type whose name beings with `MDC` (i.e., index, foundation, and adapter) should have its own file; everything else would go in `constants.js`. It feels simple and straightforward to me.

We could conceivably create an entirely separate file just for interfaces and `type` definitions, but that might be overkill.

WDYT?

kfranqueiro: IMO interfaces feel more type-y than enums (which are much more value-y than type-y), which is why I thought of `adapter.ts`. How many components have we done this with so far?

acdvorak: As discussed offline, we're going to create a new `types.ts` file for interfaces and other type definitions that don't belong in `constants.ts` or `adapter.ts`.

packages/mdc-menu-surface/index.ts
```diff
@@ -0,0 +1,206 @@
...
+class MDCMenuSurface extends MDCComponent<MDCMenuSurfaceFoundation> {
+  static attachTo(root: Element): MDCMenuSurface {
+    return new MDCMenuSurface(root);
+  }
+
+  anchorElement: Element | null = null;
+
+  protected root_!: HTMLElement; // assigned in MDCComponent constructor
+
+  private previousFocus_: HTMLElement | SVGElement | null = null;
```
kfranqueiro: Is there a reason we're using `HTMLElement | SVGElement` for these rather than just `Element`?

packages/mdc-menu/README.md
```diff
@@ -204,7 +204,7 @@ Property | Value Type | Description
 Method Signature | Description
 --- | ---
 `setAnchorCorner(Corner) => void` | Proxies to the menu surface's `setAnchorCorner(Corner)` method.
-`setAnchorMargin(AnchorMargin) => void` | Proxies to the menu surface's `setAnchorMargin(AnchorMargin)` method.
+`setAnchorMargin(MenuPosition) => void` | Proxies to the menu surface's `setAnchorMargin(MenuPosition)` method.
```
kfranqueiro: Is `MenuPosition` the right thing here?

packages/mdc-menu/index.js
```diff
@@ -24,8 +24,8 @@
 import MDCComponent from '@material/base/component';
 import {MDCMenuFoundation} from './foundation';
 import {strings, cssClasses} from './constants';
-import {MDCMenuSurface, Corner} from '@material/menu-surface/index';
-import {MDCMenuSurfaceFoundation, AnchorMargin} from '@material/menu-surface/foundation';
+import {MDCMenuSurface, Corner} from '@material/menu-surface/index.ts';
+import {MDCMenuSurfaceFoundation, MenuPosition} from '@material/menu-surface/foundation.ts';
```
kfranqueiro: AFAICT `MenuPosition` doesn't exist?

packages/mdc-menu/index.js
```diff
@@ -120,7 +120,7 @@ class MDCMenu extends MDCComponent {
   }
 
   /**
-   * @param {!AnchorMargin} margin
+   * @param {!MenuPosition} margin
```
kfranqueiro: I suspect this is going to need updating... and will this even pass closure compiler tests right now given that menu-surface is converted but this isn't?

packages/mdc-menu-surface/index.ts
acdvorak: `Element` lacks `.focus()`, which we call in the adapter.

packages/mdc-menu/README.md
acdvorak: Woops. Updated to `Partial<MenuDistance>`. Good catch!

packages/mdc-menu/index.js
acdvorak: Fixed

packages/mdc-menu/index.js
acdvorak: No idea, but we removed the Closure Compiler test from the `feat/typescript` branch.

kfranqueiro: I had forgotten there are no longer closure compiler tests on this branch, so this is fine. I was mainly concerned about CI being able to pass.

packages/mdc-menu-surface/README.md
```diff
@@ -204,25 +209,25 @@ Method Signature | Description
...
-`setAnchorMargin(margin: AnchorMargin) => void` | Sets the distance from the anchor point that the menu surface should be shown.
+`setAnchorCorner(corner: Corner) => void` | Sets the corner that the menu surface will be anchored to. See [constants.ts](constants.ts)
+`setAnchorMargin(margin: Partial<MenuDistance>) => void` | Sets the distance from the anchor point that the menu surface should be shown.
 `setIsHoisted(isHoisted: boolean) => void` | Sets whether the menu surface has been hoisted to the body so that the offsets are calculated relative to the page and not the anchor.
 `setFixedPosition(isFixed: boolean) => void` | Sets whether the menu surface is using fixed positioning.
 `setAbsolutePosition(x: number, y: numnber) => void` | Sets the absolute x/y position of the menu. Should only be used when the menu is hoisted or using fixed positioning.
-`handleBodyClick(event: Event) => void` | Method used as the callback function for the `click` event.
-`handleKeydown(event: Event) => void` | Method used as the callback function for the `keydown` events.
+`handleBodyClick(event: MouseEvent) => void` | Method used as the callback function for the `click` event.
```
kfranqueiro: There's a typo of `number` on the line before this one in the existing docs

packages/mdc-menu-surface/index.ts
```diff
@@ -0,0 +1,207 @@
...
+        return {width: window.innerWidth, height: window.innerHeight};
+      },
+      getBodyDimensions: () => {
+        return {width: document.body.clientWidth, height: document.body.clientHeight};
+      },
+      getWindowScroll: () => {
+        return {x: window.pageXOffset, y: window.pageYOffset};
+      },
+      setPosition: (position) => {
+        this.root_.style.left = 'left' in position ? `${position.left}px` : null;
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-menu-surface/README.md
acdvorak: Done

packages/mdc-menu-surface/index.ts
```diff
@@ -0,0 +1,207 @@
...
+        return {width: window.innerWidth, height: window.innerHeight};
+      },
+      getBodyDimensions: () => {
+        return {width: document.body.clientWidth, height: document.body.clientHeight};
+      },
+      getWindowScroll: () => {
+        return {x: window.pageXOffset, y: window.pageYOffset};
+      },
+      setPosition: (position) => {
+        this.root_.style.left = 'left' in position ? `${position.left}px` : '';
```
mmalerba: FYI, you might run into trouble here if closure decides to minify these properties, I would recommend this pattern instead:

```ts
this.root_.style.left = position.left == null ? `{position.left}px` : '';
```

(using `==` intentionally as lets `0` through but changes `null` and `undefined` to `''`)

acdvorak: IIRC, Closure Compiler munges the property names of objects (unless you explicitly prevent it from doing so with an annotation or w/e), but it _doesn't_ rename `string` values like `'left'`.

So the object will get minified/obfuscated to something like `{a, b, c, d}`, whereas this line of code would still be testing for the existence of a property named `'left'` that no longer exists.

Using `.dotNotation` avoids this problem by allowing Closure to rename the property everywhere.

kfranqueiro: Oh I see, I was thinking about this backwards then (thinking about avoiding it being obfuscated whatsoever rather than being obfuscated consistently).

I'm still a little unnerved about `!=` and am wondering if we can be consistent on one of `null` or `undefined` specifically?

acdvorak: If this is indeed an issue, it is present in several places:

https://github.com/material-components/material-components-web/blob/86cfdadd4fc999f696d0b057cb80afa0e3061120/packages/mdc-menu-surface/foundation.ts#L260-L261

https://github.com/material-components/material-components-web/blob/86cfdadd4fc999f696d0b057cb80afa0e3061120/packages/mdc-menu-surface/foundation.ts#L266-L269

https://github.com/material-components/material-components-web/blob/86cfdadd4fc999f696d0b057cb80afa0e3061120/packages/mdc-menu-surface/foundation.ts#L449-L450

https://github.com/material-components/material-components-web/blob/86cfdadd4fc999f696d0b057cb80afa0e3061120/packages/mdc-menu-surface/foundation.ts#L459-L463

kfranqueiro: It's also worth noting that this code already existed in this fashion both externally and internally prior to this conversion, so it would seem it hasn't caused an issue.

acdvorak: After talking to Ken offline, we're not too worried about this because `MDCMenuFoundation` is already running with prod jscompiler minification/obfuscation settings.

Thanks for bringing this up, though!

PR #4275, opened 2019-01-19
feat(switch): add feature targeting for styles

crisbeto:
Sets up style feature targeting for the switch component.


packages/mdc-switch/_mixins.scss
```diff
@@ -23,18 +23,100 @@
...
+      .mdc-switch__native-control {
+        @include mdc-switch__native-control-disabled_
+      }
+    }
+  }
+
+  // postcss-bem-linter: end
+}
+
+@mixin mdc-switch-toggled-on-color($color, $query: mdc-feature-all()) {
```
mmalerba: @kfranqueiro should we add feature targeting to only the main public mixin, or these ones too? (If we are adding it here too, we should add these mixins to the test that verifies nothing is emitted for empty query, I think it would fail currently)

mmalerba:
It looks like the order of some of the rules got swapped around. I'm not sure if it matters or not (potentially could due to specificity reasons). Here's the before and after for diffing:

[switch-before.css](https://github.com/material-components/material-components-web/files/2784007/switch-before.css.txt)
[switch-after.css](https://github.com/material-components/material-components-web/files/2784008/switch-after.css.txt)

crisbeto:
I don't think there's much overlap between the different style declarations so we shouldn't run into any issues with ordering.

packages/mdc-switch/_mixins.scss
kfranqueiro: That's a good point, this mixin is not invoked by the base styles, so we don't necessarily need to cover it, if the goal is specifically to add feature targeting to the base emitted styles. If you're fine with that, then I think we can omit `$query` from `mdc-switch-toggled-on-color` and `mdc-switch-toggled-off-color`.

packages/mdc-switch/_mixins.scss
```diff
@@ -76,3 +155,142 @@
...
+    cursor: pointer;
+    pointer-events: auto;
+  }
+}
+
+@mixin mdc-switch__thumb_($query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-feature-targets($feat-structure) {
+    @include mdc-elevation(2);
```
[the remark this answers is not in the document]
kfranqueiro: Replying to this comment to note this hasn't been fixed yet.

packages/mdc-switch/_mixins.scss
mmalerba: I could also see doing it, just for the sake of consistency on public mixins. If we always do it then people don't have to think about whether this is one of the ones that supports it.

Also for things like disabling animations, you probably do want to have the ability to exclude the `animation` feature from _all_ mixins, not just the base one.

crisbeto:
@kfranqueiro I've rebased the PR and double-checked the output changes. It seems like a couple of unrelated styles got swapped which didn't have a visual effect.

crisbeto:
Looks like the changes that catch nested feature target mixins caught a couple in this PR. I moved a couple of styles around to fix the errors.

packages/mdc-switch/_mixins.scss
kfranqueiro: If we're parameterizing `$query` here, then I think we need to enclose the first 2 mixin invocations within a color feature-targets block (same for `mdc-switch-toggled-off-color`), and invoke these in `test/scss/feature-targeting.scss` as Miles suggested.

packages/mdc-switch/_mixins.scss
```diff
@@ -71,8 +154,146 @@
...
+      mdc-switch-transition(border-color);
+  }
+}
+
+@mixin mdc-switch__thumb-underlay_($query: mdc-feature-all()) {
+  $feat-animation: mdc-feature-create-target($query, animation);
+  $feat-color: mdc-feature-create-target($query, color);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-states($mdc-switch-baseline-theme-color, false, $query);
```
kfranqueiro: Move `mdc-states` after `mdc-ripple-surface` (`mdc-ripple-surface` first is a common pattern across the packages that use it)

packages/mdc-switch/_mixins.scss
```diff
@@ -23,18 +23,101 @@
...
+  // postcss-bem-linter: define switch
+
+  $feat-animation: mdc-feature-create-target($query, animation);
+  $feat-color: mdc-feature-create-target($query, color);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-ripple-common($query);
+
+  .mdc-switch {
+    @include mdc-switch-toggled-off-ripple-color($mdc-switch-toggled-off-ripple-color, $query);
```
kfranqueiro: Move this to the end of this block to keep it near the other related color styles (and reduce diff)

packages/mdc-switch/_mixins.scss
```diff
@@ -23,18 +23,101 @@
...
+
+  .mdc-switch__track {
+    @include mdc-switch__track_($query);
+  }
+
+  .mdc-switch__thumb-underlay {
+    @include mdc-switch__thumb-underlay_($query);
+  }
+
+  .mdc-switch__native-control {
```
kfranqueiro: Move this above `.mdc-switch__track` to reduce diff

crisbeto:
@kfranqueiro I've addressed the latest set of feedback. Can you take a look at `feature-targeting.scss` in particular? I'm not sure that I completely understood what you meant with your comment.

packages/mdc-switch/_mixins.scss
```diff
@@ -23,56 +23,287 @@
...
+}
+
+@mixin mdc-switch__thumb-underlay_($query: mdc-feature-all()) {
+  $feat-animation: mdc-feature-create-target($query, animation);
+  $feat-color: mdc-feature-create-target($query, color);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-ripple-surface($query);
+  @include mdc-states($mdc-switch-baseline-theme-color, false, $query);
+  @include mdc-ripple-radius-unbounded(100%, $query);
```
kfranqueiro: Move this above `mdc-states`

kfranqueiro:
The update to `feature-targeting.scss` looked fine except that it also needed to invoke the baseline `mdc-switch` mixin to test feature targeting in that. I added that and fixed lint issues in the mixins (check with `npm run lint:css` locally to catch them early).

packages/mdc-switch/_mixins.scss
kfranqueiro: Whoops. Ignore this comment; it was hanging around as a pending comment but I already resolved it on the branch when I linted it.

PR #4277, opened 2019-01-21
Problems related to invalid style properties resolved

amalichev:
I used laravel 5.7  from the box. So I got the following problem when compiling:  

screen

### DIFF
--- a/packages/mdc-layout-grid/_mixins.scss
+++ b/packages/mdc-layout-grid/_mixins.scss
@@ -153,7 +153,7 @@
     align-self: flex-start;
 
     @supports (display: grid) {
-      align-self: start;
+      align-self: flex-start;
     }
   }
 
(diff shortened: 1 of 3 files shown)


amalichev:
I signed it!

packages/mdc-layout-grid/_mixins.scss
```diff
@@ -153,7 +153,7 @@
     align-self: flex-start;
 
     @supports (display: grid) {
-      align-self: start;
+      align-self: flex-start;
```
kfranqueiro: I don't think `flex-start` and `flex-end` make sense when `align-self` is used in the context of CSS Grid (not flexbox). This file should probably be reverted.

The warning is ignorable in this case because there are separate fallback styles for browsers that don't support CSS Grid, which do use `flex-*`.

kfranqueiro:
Closing this given that the warnings under mdc-layout-grid seem invalid due to the fallback styles present for IE, and mdc-toolbar is deprecated in favor of mdc-top-app-bar which doesn't have this issue.

PR #4285, opened 2019-01-23
feat(text-field): Added support for text field without label

### LINKED ISSUE #4239: MDCTextField: Add support for text fields without labels
From the [Design Guidelines](https://material.io/design/components/text-fields.html#filled-text-field):

image

abhiomkar:
**Features included**:

- Adds support for text field without label by making floating label element optional for all text field variants.
- Adds support for placeholder with or without floating label element where placeholder is visible only on focused state for text field that has label.

Fixes #4239

**Screenshot**:


packages/mdc-textfield/README.md
```diff
@@ -129,6 +129,44 @@ To disable the text field, add the `disabled` attribute to the `<input>` element
 </div>
 ```
 
+### Text Field without label
+
+A text field doesn’t require a label if a separate but clear label indicator is already displayed adjacent to the text field.
+Add class name `mdc-text-field--no-label` and remove the label element from the structure.
+
+**Filled**
```
kfranqueiro: Nit: Is there a reason for bold rather than level-4 headings?

test/screenshot/spec/mdc-textfield/classes/focused-outlined-placeholder.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
```
kfranqueiro: Given we have other test pages that show multiple focus states at once, it seems like we could combine this into focused-baseline-placeholder.html (and just call it focused-placeholder.html)? Save ourselves one more page load and screenshot capture.

abhiomkar: The reason I separated these tests out is that I specifically want to test if placeholder text and label are not overlapping. Since we can only have one element in focus state per page I had to create separate test for these 2 variants.

test/screenshot/spec/mdc-textfield/classes/focused-baseline-placeholder.html
```diff
@@ -0,0 +1,65 @@
...
+  </head>
+
+  <body class="test-container">
+    <main class="test-viewport test-viewport--mobile">
+      <div class="test-layout">
+
+        <div class="test-cell test-cell--textfield">
+          <div class="mdc-text-field mdc-text-field--focused">
+            <input type="text" id="filled-text-field" class="mdc-text-field__input test-text-field__input" placeholder="Placeholder text" autofocus>
+            <label class="mdc-floating-label" for="filled-text-field">Label</label>
```
kfranqueiro: Should this also have the `mdc-floating-label--float-above` class?

test/screenshot/spec/mdc-textfield/classes/focused-outlined-placeholder.html
kfranqueiro: Given the recent update to the styles to select on the `--focused` class instead of `:focus`, I think we can actually rely on that instead of `autofocus`. That will allow us to combine these, and should also avoid flakes due to the blinking text cursor.

abhiomkar:
Updated golden screenshots.

PR #4286, opened 2019-01-23
fix: add strict true to tsconfig

moog16:


packages/mdc-linear-progress/foundation.ts
```diff
@@ -48,9 +48,10 @@ export default class MDCLinearProgressFoundation extends MDCFoundation<MDCLinear
     };
   }
 
-  private isDeterminate_: boolean;
-  private isReversed_: boolean;
-  private progress_: number;
+  // https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html
```
acdvorak: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization

packages/mdc-dom/ponyfill.ts
```diff
@@ -30,12 +30,12 @@ interface MsElement extends Element {
   msMatchesSelector(selector: string): boolean;
 }
 
-function closest(element: Element, selector: string): Element | null {
+function closest(element: HTMLElement, selector: string): HTMLElement | null {
```
acdvorak: This brings up a broader question about how we want to type `Element` variables in general:

Should we assume that _everything_ is an `HTMLElement` for simplicity? Or should we be technically correct (the best kind of correct!) and use `Element` everywhere, _except_ in cases where we're 100% sure that the element will _never_ be an `SVGElement` (e.g., in `MDCMenuSurface`)?

What's more, if we use `HTMLElement` everywhere, will it end up forcing us (or our clients) to cast lots of things to from `Element` to `HTMLElement`? If so, will that be too confusing or burdensome (especially in cases where we know that something _might_ be an `SVGElement`)?

E.g.:

```js
// Might be an SVGElement
const iconEl = this.root_.querySelector('.mdc-button__icon');

if (iconEl) {
  // Cast SVGElement to HTMLElement...
  const buttonEl = ponyfill.closest(iconEl as HTMLElement, '.mdc-button');

  // ...
}
```

Realistically, there are only two subclasses of `Element` that we will ever encounter: `HTMLElement` and `SVGElement`.

From my testing in IE 11 (the most restrictive browser we support), `SVGElement` lacks `parentElement`, `classList`, `focus()`, and `blur()`. It appears to support all other APIs that we use (`querySelector()`, `msMatchesSelector()`, `addEventListener()`, etc.).

I'm leaning heavily toward "make most things an `HTMLElement`" because I think it will make our lives easier in many cases, but I'd like to get your take on it.

WDYT?

packages/mdc-linear-progress/foundation.ts
```diff
@@ -48,9 +48,10 @@ export default class MDCLinearProgressFoundation extends MDCFoundation<MDCLinear
     };
   }
 
-  private isDeterminate_: boolean;
-  private isReversed_: boolean;
-  private progress_: number;
+  // https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization
```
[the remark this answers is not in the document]
moog16: true...I usually end up deleting this type of comment 3 weeks later lol

packages/mdc-dom/ponyfill.ts
mmalerba: I would suggest going with the most general class that you don't consider to be a usage error. (e.g. for the native control on the checkbox `HTMLInputElement`, for most things probably `HTMLElement`, for icons `Element`). If you need to do something that's only supported on the more specific interface you can use an `instanceof` check

moog16: Ya don't need to cast...I'll check in what @acdvorak has stated above

moog16: @mmalerba just so that I understand, in general we should do the most general type? In this case it is Element. This is because HTMLInputElement and HTMLElement are all of type Element. So those 2 cases would work when passing `element`?

mmalerba: Yeah in this case it would be an error on the user's part to pass a `Node` since `Node` doesn't support `matches`. The most general class that isn't a user error would be `Element`, that will allow users to pass `Element` or `HTMLElement`, etc.

I mentioned `HTMLInputElement` as an example of when you would actually want to use a pretty specific type (when you're giving the type for the native control in the `MDCCheckbox`). It would be an error on the user's part to try to pass a `HTMLDivElement` as the native control for the checkbox. `HTMLInputElement`, though pretty specific, happens to be the most general class that isn't a user error for that case.

mmalerba: Hmmm, I don't know, it technically makes the return type a lie. I think it's better to have the user of this API explicitly cast if that's what they want (or use `instanceof` if they really need to make sure it is that type.)

tsconfig.json
```diff
@@ -1,12 +1,24 @@
...
+    "noFallthroughCasesInSwitch": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
+    "noErrorTruncation": true,
+    "noEmitOnError": false,
+    "importHelpers": true,
+    "noStrictGenericChecks": true,
+    "strictNullChecks": true,
     "alwaysStrict": true,
+    "strict": true,
```
kfranqueiro: We don't need `strictNullChecks`, `alwaysStrict`, `strictPropertyInitialization`, and `strictFunctionTypes` when `strict` is true, as it already includes them. You can remove those others.

PR #4287, opened 2019-01-24
refactor(animation): Replace enums with string literal type aliases

acdvorak:
Refs #4225

IMO this makes the code a little easier to read/write.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types


moog16:
Sleeping on this one - string literals might be cleaner, but the enums also provide an easy avenue to refactoring. In this case I don't think it matters as much, since the naming of `animationend` or `transform` are going to change :)

kfranqueiro:
I for one vastly prefer this approach... but it looks like it's failing tests in its current state.

moog16:
My only reason for using enums were for autocomplete, and avoiding spelling mistakes. I think in this case though, these can already be autocomplete'd. If you have a stronger opinion here then lets go with yours :+1:

packages/mdc-animation/index.ts
```diff
@@ -21,31 +21,18 @@
...
+    '-webkit-animation' | '-webkit-transform' | '-webkit-transition'
+);
+type StandardJsEventType = (
+    'animationend' | 'animationiteration' | 'animationstart' | 'transitionend'
+);
+type PrefixedJsEventType = (
+    'webkitAnimationEnd' | 'webkitAnimationIteration' | 'webkitAnimationStart' | 'webkitTransitionEnd'
+);
 
 interface CssVendorPropertyMap {
```
moog16: I know this isn't part of the change, but I think `[key: string]` is loose typing. Since `CssVendorPropertyMap` and `JsVendorPropertyMap` are just used in this file, can you change it to:

```
type CssVendorPropertyMap = {
  [key in StandardCssPropertyName]: CssVendorProperty;
}

type JsVendorPropertyMap = {
  [key in StandardJsEventType]: JsVendorProperty;
}```

acdvorak: Done

PR #4291, opened 2019-01-24
fix(ripple): Fixed the order of css animation shorthand values

### LINKED ISSUE #4283: Ordering of animation shorthand breaks css-module build
## Bugs

I am attempting to compose mdc text field classes into a css-module file. When building this file with local names turned on, the css-module parser is unable to parse some of the animation shorthand used. This is because the css-module parser requires animation shorthand to specify the animation name before the timing / duration etc.

It appears that this could be rectified by updating two scss mixins to return animation styles with the animation name at the front.

These two files are:

- `packages/mdc-checkbox/_mixins.scss`
- `packages/mdc-ripple/_mixins.scss`

Making this change would allow developers using css-modules to style their components using the mdc library and should not diversely effect anyone else.

0.42.1

### Example of proposed change

Source: https://github.com/material-components/material-components-web/blob/master/packages/mdc-ripple/_mixins.scss#L124

#### Before

``` scss
@include mdc-feature-targets($feat-animation) {
	&.mdc-ripple-upgraded--foreground-activation::after {
		animation:
		$mdc-ripple-translate-duration mdc-ripple-fg-radius-in forwards,
		$mdc-ripple-fade-in-duration mdc-ripple-fg-opacity-in forwards;
	}
}
```

#### After

``` scss
@include mdc-feature-targets($feat-animation) {
	&.mdc-ripple-upgraded--foreground-activation::after {
		animation:
		mdc-ripple-fg-radius-in $mdc-ripple-translate-duration forwards,
		mdc-ripple-fg-opacity-in $mdc-ripple-fade-in-duration forwards;
	}
}
```

abhiomkar:
Fixes #4283

### DIFF
--- a/packages/mdc-ripple/_mixins.scss
+++ b/packages/mdc-ripple/_mixins.scss
@@ -124,8 +124,8 @@
   @include mdc-feature-targets($feat-animation) {
     &.mdc-ripple-upgraded--foreground-activation::after {
       animation:
-        $mdc-ripple-fg-radius-in $mdc-ripple-translate-duration forwards,
-        $mdc-ripple-fg-opacity-in $mdc-ripple-fade-in-duration forwards;
+        mdc-ripple-fg-radius-in $mdc-ripple-translate-duration forwards,
+        mdc-ripple-fg-opacity-in $mdc-ripple-fade-in-duration forwards;
     }
   }
 
--- a/packages/mdc-ripple/_mixins.scss
+++ b/packages/mdc-ripple/_mixins.scss
@@ -124,8 +124,8 @@
   @include mdc-feature-targets($feat-animation) {
     &.mdc-ripple-upgraded--foreground-activation::after {
       animation:
-      mdc-ripple-fg-radius-in $mdc-ripple-translate-duration forwards,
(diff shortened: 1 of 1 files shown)


packages/mdc-ripple/_mixins.scss
```diff
@@ -124,8 +124,8 @@
   @include mdc-feature-targets($feat-animation) {
     &.mdc-ripple-upgraded--foreground-activation::after {
       animation:
-        $mdc-ripple-translate-duration mdc-ripple-fg-radius-in forwards,
-        $mdc-ripple-fade-in-duration mdc-ripple-fg-opacity-in forwards;
+      mdc-ripple-fg-radius-in $mdc-ripple-translate-duration forwards,
```
kfranqueiro: It looks like these lines lost some indentation in this change.

kfranqueiro:
Also, FYI, I believe this PR is actually changing our code from following the CSS animation spec correctly, to actually breaking spec. See https://github.com/material-components/material-components-web/issues/4283#issuecomment-457278827

abhiomkar:
Fixed in #4297

Issue #2258, opened 2018-02-15
Changing input label on a MDCTextField from Javascript?


rkkoszewski:
Hi, by using...
`node.querySelector('.mdc-text-field').MDCTextField.value = "Text"`
I can change the value of the text input. But how can I change the label? In the documentation it says that MDCTextFieldFoundation supports multiple sub-elements inclusive label, but its not stated how to do that, and if it is even possible with AutoInit. Anybody can shed some light here on how to make this work?

P.D: I know that I can change the label by using JQuery and changing the text directly, but I'm trying to do this with the official MDCTextField Class.

kfranqueiro:
Thanks for raising this question instead of immediately taking "the jQuery route" :)

Can you give us some info on your use case for needing to change the text field's label dynamically? This is something that we haven't expected people to need to do (bearing in mind that there is also a separate helper text element), and if we were to add this, we'd probably also want to talk to our designers about if/how they'd want to frame guidance around this.

rkkoszewski:
Hi,
I have actually two use cases for it. One is where I'm building a realtime website, which may have its content changed (I update only what is changed, not the whole page). Another use case, and which has even higher priority for me, is that I'm building a property based widget editor, which includes MDC Web components. The user can change some predefined properties and these will reflect in realtime on the widgets (The properties are of kind "label", "value", "help-text" and so on).

It would be just nice to have an official API to perform these tasks. Also having an official API to change these optional parts could avoid JQuery workarounds which may break whenever any html or class changes are made in MDC Web.

moog16:
@rkkoszewski I've marked this as `help wanted`, so if you want to try taking a stab at it please feel free and open a PR. It should follow a similar pattern to the helperText. There is a `helperTextContent` method in the mdc-textfield/index.js file.

PR #4299, opened 2019-01-25
Adding label setter to MDCTextField

thebrianchen:
Taking a stab at #2258
* Added `setLabelContent()` to MDCTextfield and foundation
* Added `setContent()` to MDCFloatingLabel
* Fixed bug in `getFoundationMap_()` where `foundation_` was missing the trailing underscore, resulting in `undefined` fields

### DIFF
--- a/packages/mdc-textfield/foundation.js
+++ b/packages/mdc-textfield/foundation.js
@@ -30,7 +30,6 @@ import MDCFloatingLabelFoundation from '@material/floating-label/index';
 import {MDCTextFieldAdapter, NativeInputType, FoundationMapType} from './adapter';
 import {cssClasses, strings, numbers, VALIDATION_ATTR_WHITELIST, ALWAYS_FLOAT_TYPES} from './constants';
 
-
 /**
  * @extends {MDCFoundation<!MDCTextFieldAdapter>}
  * @final
@@ -107,7 +106,6 @@ class MDCTextFieldFoundation extends MDCFoundation {
    */
   constructor(adapter, foundationMap = /** @type {!FoundationMapType} */ ({})) {
     super(Object.assign(MDCTextFieldFoundation.defaultAdapter, adapter));
-
     /** @type {!MDCTextFieldHelperTextFoundation|undefined} */
     this.helperText_ = foundationMap.helperText;
     /** @type {!MDCFloatingLabelFoundation|undefined} */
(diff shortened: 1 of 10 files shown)


packages/mdc-textfield/index.js
```diff
@@ -494,9 +502,10 @@ class MDCTextField extends MDCComponent {
    */
   getFoundationMap_() {
     return {
-      helperText: this.helperText_ ? this.helperText_.foundation : undefined,
-      leadingIcon: this.leadingIcon_ ? this.leadingIcon_.foundation : undefined,
-      trailingIcon: this.trailingIcon_ ? this.trailingIcon_.foundation : undefined,
+      helperText: this.helperText_ ? this.helperText_.foundation_ : undefined,
```
kfranqueiro: the helper-text and icon sub-components have public `foundation` getters so this was correct as-is.

packages/mdc-textfield/foundation.js
```diff
@@ -105,9 +106,10 @@ class MDCTextFieldFoundation extends MDCFoundation {
    */
   constructor(adapter, foundationMap = /** @type {!FoundationMapType} */ ({})) {
     super(Object.assign(MDCTextFieldFoundation.defaultAdapter, adapter));
-
     /** @type {!MDCTextFieldHelperTextFoundation|undefined} */
     this.helperText_ = foundationMap.helperText;
+    /** @type {!MDCFloatingLabelFoundation|undefined} */
```
kfranqueiro: I'm not sure there's a need to add this to the foundation. The foundation itself does not depend on `setLabelContent` anywhere, so it can go directly in the component, in which case it can simply reference `label_` on the component instance (which already exists).

packages/mdc-textfield/index.js
```diff
@@ -369,6 +369,14 @@ class MDCTextField extends MDCComponent {
     this.foundation_.setUseNativeValidation(useNativeValidation);
   }
 
+  /**
+   * Sets the label element content.
+   * @param {string} content
+   */
+  set labelContent(content) {
+    this.foundation_.setLabelContent(content);
```
kfranqueiro: I suspect we need to update the notched outline's width if it is present and open. Otherwise the label size will update but the notch probably won't by itself. We should at least manually test this to be sure.

kfranqueiro:
I'm closing this since there are outstanding items we'd need to test and there's been no movement on it. (Also we're currently converting the text field components to TypeScript and it probably makes more sense to tackle this post-conversion at this point.)

PR #4300, opened 2019-01-25
feat(ripple): convert to ts

moog16:
related #4225


packages/mdc-ripple/adapter.ts
```diff
@@ -42,75 +40,54 @@
...
+
+  isSurfaceDisabled(): boolean;
+
+  addClass(className: string): void;
+
+  removeClass(className: string): void;
+
+  containsEventTarget(target: EventTarget | null): boolean;
+
+  registerInteractionHandler<K extends keyof GlobalEventHandlersEventMap>(
```
mmalerba: ooh fancy! I didn't know this was a thing :)

I'd suggest putting some aliases in a common place somewhere to make the types more readable:

```ts
type EventType = keyof GlobalEventHandlersEventMap;
type SpecificEventListener<K extends EventType> = (evt: GlobalEventHandlersEventMap[K]) => void;
```
Then these would look like:

```ts
registerInteractionHandler<E extends EventType>(evtType: E, handler: SpecificEventListener<E>): void;
registerResizeHandler(handler: SpecificEventListener<'resize'>): void;
```

moog16: yaa I like that :+1:

packages/mdc-ripple/util.ts
```diff
@@ -20,24 +20,30 @@
...
- * Stores result from supportsCssVariables to avoid redundant processing to detect CSS custom variable support.
- * @private {boolean|undefined}
- */
-let supportsCssVariables_;
+interface CssObject {
+  supports(prop: string, value?: string): boolean;
+}
+
+interface CssWindow extends Window {
+  CSS?: CssObject;
```
mmalerba: TS has a type `CSS` which is equivalent to your `CssObject`, it's just not on the `Window` for some reason. You might be able to make a file of type fixes somewhere, like this:

```ts
declare interface Window { CSS: CSS; }
```

You could also add a fix for the `msMatchesSelector` thing, might be nicer than having to invent your own types and cast them everywhere

moog16: ya we need to invent a place for these global types. Maybe the correct place is in mdc-dom package. I'm going to add it to the `index.ts` for now. We can decide for a better place as this PR evolves.

moog16: Also I found that this works:
```
declare global {
  interface Window {
    CSS: CSS;
  }
}

```

globals seems like a last ditch effort. Is that a valid pattern? Or is there a better way of accomplishing this?

packages/mdc-menu-surface/adapter.ts
```diff
@@ -0,0 +1,131 @@
+/**
```
moog16: don't know how this ended up here...removing

packages/mdc-ripple/adapter.ts
```diff
@@ -42,75 +40,46 @@
...
+
+  registerInteractionHandler<K extends EventType>(evtType: K, handler: SpecificEventListener<K>): void;
+
+  deregisterInteractionHandler<K extends EventType>(evtType: K, handler: SpecificEventListener<K>): void;
+
+  registerDocumentInteractionHandler<K extends EventType>(evtType: K, handler: SpecificEventListener<K>): void;
+
+  deregisterDocumentInteractionHandler<K extends EventType>(evtType: K, handler: SpecificEventListener<K>): void;
+
+  registerResizeHandler(handler: (evt: GlobalEventHandlersEventMap['resize']) => void): void;
```
mmalerba: `SpecificEventListener<'resize'>`

moog16:
I signed it

packages/mdc-ripple/util.ts
```diff
@@ -154,16 +154,17 @@ function getNormalizedEventCoords(ev, pageOffset, clientRect) {
   let normalizedY;
   // Determine touch point relative to the ripple container.
   if (ev.type === 'touchstart') {
-    ev = /** @type {!TouchEvent} */ (ev);
-    normalizedX = ev.changedTouches[0].pageX - documentX;
-    normalizedY = ev.changedTouches[0].pageY - documentY;
+    const e = ev as TouchEvent;
```
kfranqueiro: Can we give this variable a more descriptive name like `touchEvent` rather than an even more abbreviated name? Single-letter variables are notoriously harder to search usages of.

packages/mdc-ripple/util.ts
```diff
@@ -154,16 +154,17 @@ function getNormalizedEventCoords(ev, pageOffset, clientRect) {
...
+    normalizedY = e.changedTouches[0].pageY - documentY;
   } else {
-    ev = /** @type {!MouseEvent} */ (ev);
-    normalizedX = ev.pageX - documentX;
-    normalizedY = ev.pageY - documentY;
+    const e = ev as MouseEvent;
+    normalizedX = e.pageX - documentX;
+    normalizedY = e.pageY - documentY;
+  }
+  if (normalizedX === undefined || normalizedY === undefined) {
```
kfranqueiro: Is this actually possible to happen?

packages/mdc-ripple/index.ts
```diff
@@ -26,88 +26,80 @@ import MDCRippleAdapter from './adapter';
...
+      removeClass: (className) => instance.root_.classList.remove(className),
+      updateCssVariable: (varName, value) => (instance.root_ as HTMLElement).style.setProperty(varName, value),
     };
   }
 
-  /** @return {boolean} */
-  get unbounded() {
-    return this.unbounded_;
+  disabled = false;
+  root_: Element = this.root_;
```
kfranqueiro: Possibly dumb question: is the assignment here actually doing anything or is it just assigning `this.root_` to itself?

packages/mdc-ripple/util.ts
```diff
@@ -55,82 +52,87 @@ function detectEdgePseudoVarBug(windowObj) {
...
- */
-
-function supportsCssVariables(windowObj, forceRefresh = false) {
-  let supportsCssVariables = supportsCssVariables_;
+/** Checks whether the browser supports Css Variables. */
+export function supportsCssVariables(windowObj: Window, forceRefresh = false): boolean {
+  let supportsCssVars = supportsCssVariables_;
   if (typeof supportsCssVariables_ === 'boolean' && !forceRefresh) {
-    return supportsCssVariables;
+    return Boolean(supportsCssVariables_);
```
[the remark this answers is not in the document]
moog16: True...I could use a guard here. TS Compiler was complaining with the cast.

packages/mdc-ripple/util.ts
```diff
@@ -55,82 +52,87 @@ function detectEdgePseudoVarBug(windowObj) {
...
   }
 
   const supportsFunctionPresent = windowObj.CSS && typeof windowObj.CSS.supports === 'function';
   if (!supportsFunctionPresent) {
-    return;
+    return false;
   }
 
-  const explicitlySupportsCssVars = windowObj.CSS.supports('--css-vars', 'yes');
+  const {CSS} = windowObj;
```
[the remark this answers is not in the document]
moog16: Correct if we don't define `files` or `includes` in the tsconfig.json file, TS defaults to finding all `.d.ts` files within the tsconfig's root directory.

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html
```
If the "files" and "include" are both left unspecified, the compiler defaults to including all TypeScript (.ts, .d.ts and .tsx) files in the containing directory and subdirectories except those excluded using the "exclude" property. JS files (.js and .jsx) are also included if allowJs is set to true. If the "files" or "include" properties are specified, the compiler will instead include the union of the files included by those two properties. Files in the directory specified using the "outDir" compiler option are excluded as long as "exclude" property is not specified.
```

packages/mdc-ripple/util.ts
```diff
@@ -55,82 +52,87 @@ function detectEdgePseudoVarBug(windowObj) {
...
+  return supportsPassive_ ? {passive: true} as EventListenerOptions : false;
 }
 
-/**
- * @param {!Object} HTMLElementPrototype
- * @return {string}
- */
-function getMatchesProperty(HTMLElementPrototype) {
+/** Gets the matches function from an element. */
+export function getMatchesFunction(htmlElementPrototype: HTMLElement): (selector: string) => boolean {
```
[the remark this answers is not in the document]
moog16: Also something I picked up from Rachel's CL...I didn't check it for usages. Good catch

packages/mdc-ripple/util.ts
```diff
@@ -139,13 +141,11 @@ function getMatchesProperty(HTMLElementPrototype) {
...
 
-/**
- * @param {!Event} ev
- * @param {{x: number, y: number}} pageOffset
- * @param {!ClientRect} clientRect
- * @return {{x: number, y: number}}
- */
-function getNormalizedEventCoords(ev, pageOffset, clientRect) {
+export function getNormalizedEventCoords(
+    ev: Event | undefined, pageOffset: Point, clientRect: ClientRect): Point {
```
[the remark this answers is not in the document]
moog16: in foundation.ripple the activationEvent is defaulted to `undefined`. It then trickled to this function. Not sure what else to default the `defaultActivationState`.

PR #4303, opened 2019-01-27
feat(list): add feature targeting for styles

crisbeto:
Sets up feature targeting for `mdc-list`.


packages/mdc-list/_mixins.scss
```diff
@@ -21,12 +21,278 @@
...
+        @include mdc-typography-baseline-top($dense-item-primary-text-baseline-height);
+        @include mdc-typography-baseline-bottom($item-secondary-text-baseline-height);
+      }
+      // stylelint-enable plugin/selector-bem-pattern
+    }
+  }
+
+  .mdc-list-item__secondary-text {
+    @include mdc-feature-targets($feat-structure) {
+      @include mdc-typography(body2);
```
kfranqueiro: Wouldn't `mdc-typography(...)` target the `typography` feature?

@mmalerba I assume that other `mdc-typography` mixins can be included under structure and that it's only important for styles pertaining to typographical scale levels to be under the typography feature?

mmalerba: I think we should do it the same as theme, styles that depend on the typography global variables that users can fiddle with should be targeted at the typography feature (so in this case, yeah I would think this should be targeted at typography).  If the other typography mixins don't depend on the global variables then I would say, yes we can put them under structure

crisbeto:
@kfranqueiro I've addressed the feedback. Note that I didn't move the styles from `mdc-typography-overflow-ellipsis`, `mdc-typography-baseline-top` and `mdc-typography-baseline-bottom` under the `typography` target, because the styles themselves seem more structural (`margin`, `overflow`, `vertical-align` etc.).

packages/mdc-list/_mixins.scss
```diff
@@ -21,62 +21,397 @@
...
+    @include mdc-feature-targets($feat-structure) {
+      pointer-events: none;
+    }
+  }
+
+  .mdc-list-item__primary-text {
+    @include mdc-typography-overflow-ellipsis($query);
+    @include mdc-typography-baseline-top($item-primary-text-baseline-height, $query);
+    @include mdc-typography-baseline-bottom($item-secondary-text-baseline-height, $query);
+    @include mdc-typography-baseline-top($dense-item-primary-text-baseline-height, $query);
```
kfranqueiro: This line is in the wrong place. This belongs inside `.mdc-list--dense & {` below.

packages/mdc-list/_mixins.scss
```diff
@@ -21,62 +21,397 @@
...
+  }
+}
+
+// Ripple styles for an interactive list item (one that is enabled and inside an interactive list).
+@mixin mdc-list-item-interactive-ripple_($query: mdc-feature-all()) {
+  @include mdc-ripple-surface($query);
+  @include mdc-states(mdc-theme-prop-value(on-surface), false, $query);
+  @include mdc-states-activated(primary, false, $query);
+  @include mdc-states-selected(primary, false, $query);
+  @include mdc-ripple-radius-bounded($query: $query);
```
kfranqueiro: Move this above `mdc-states`

packages/mdc-list/_mixins.scss
```diff
@@ -21,62 +21,397 @@
...
 // Private
 //
 
+@mixin mdc-list-base_($query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+  $feat-structure: mdc-feature-create-target($query, structure);
+  $feat-typography: mdc-feature-create-target($query, typography);
+
+  @include mdc-typography(subtitle1, $query);
+  @include mdc-feature-targets($feat-typography) {
```
kfranqueiro: Add newline before this line for lint

packages/mdc-list/_mixins.scss
```diff
@@ -21,62 +21,397 @@
...
+  }
+
+  .mdc-list-group__subheader {
+    $mdc-list-subheader-virtual-height: 3rem;
+    $mdc-list-subheader-leading: map-get(map-get($mdc-typography-styles, body1), line-height);
+    $mdc-list-subheader-margin:
+      ($mdc-list-subheader-virtual-height - $mdc-list-subheader-leading) / 2;
+
+    @include mdc-typography(subtitle1, $query);
+    @include mdc-feature-targets($feat-structure) {
```
kfranqueiro: Add newline before this line for lint

crisbeto:
@kfranqueiro I've addressed the latest set of feedback.

kfranqueiro:
No screenshot diffs in list or menu: https://storage.googleapis.com/mdc-web-screenshot-tests/kfranqueiro/2019/02/04/15_49_03_549/report/report.html?utm_source=cli&utm_medium=test_results

PR #4305, opened 2019-01-29
feat(typography): add feature targeting for styles

mmalerba:


packages/mdc-button/_mixins.scss
```diff
@@ -164,10 +164,16 @@
 }
 
 @mixin mdc-button-shape-radius($radius, $rtl-reflexive: false) {
-  @include mdc-shape-radius(mdc-shape-resolve-percentage-radius($mdc-button-height, $radius), $rtl-reflexive);
+  @include mdc-shape-radius(
```
abhiomkar: Store first parameter into a local variable instead?

```
$resolved-radius: mdc-shape-resolve-percentage-radius($mdc-button-height, $radius);
@include mdc-shape-radius($resolved-radius, $rtl-reflexive);
```

Same below.

packages/mdc-typography/_mixins.scss
```diff
@@ -20,53 +20,78 @@
...
 
-@mixin mdc-typography-baseline-top($distance) {
-  display: block;
-  margin-top: 0;
-  /* @alternate */
-  line-height: normal;
+@mixin mdc-typography-baseline-top($distance, $query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-feature-targets($feat-structure) {
```
abhiomkar: why not include all the styles inside single `@include` block?

packages/mdc-button/_mixins.scss
kfranqueiro: TBH I'm not even sure why these lines changed, since we typically fit to 120 width. (Though now that I'm checking that's only in our eslintrc, not stylelint)

mmalerba: whoops sorry about that, was experimenting with using prettier and I guess it went crazy

packages/mdc-typography/_mixins.scss
mmalerba: I remember @kfranqueiro mentioning on another PR (can't remember where exactly) that he preferred this:

```scss
.some-selector {
    ...
  }
}

.some-other-selector {
    ...
  }
}
```
over this:

```scss
  .some-selector {
    ...
  }

  .some-other-selector {
    ...
  }
} 
```

I agree that this makes it easier to read the file because you can follow the selectors down first, and then worry about what feature you're targetting at the very end

mmalerba:
I made a PR to update the stylelint config: https://github.com/material-components/material-components-web/pull/4307

packages/mdc-typography/_mixins.scss
abhiomkar: Ah ok. Thanks for explaining!

Yes, packaging feature related styles (ie, `structure`) into one block might be good for readability. But if you think it is consistent with other files this looks good.

WDYT @kfranqueiro ?

kfranqueiro: Yeah, this was based on a comment I made on one of the other PRs. IIRC I specifically commented in a case where writing selectors nested within feature-targets blocks ended up causing redundant selectors in output compared to before, so this also serves to discourage that from happening, as well as just more closely mirroring how you'd write CSS if you were to ignore the feature-targets stuff.

PR #4306, opened 2019-01-29
refactor: 💡 Added CSS prefix for card component

BrsJsk:
Added CSS prefix for card component for better browser compatibility

### DIFF
--- a/packages/mdc-card/_mixins.scss
+++ b/packages/mdc-card/_mixins.scss
@@ -52,15 +52,27 @@
 //
 
 @mixin mdc-card-container-layout_ {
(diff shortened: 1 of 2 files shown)


BrsJsk:
I signed it!

kfranqueiro:
Thanks for submitting a pull request to MDC Web!

Looking at this PR, this is adding vendor prefixing for features that are supported unprefixed within the set of browsers we support (most recent 2 versions, and only back to IE 11 for MS browsers). We also use `autoprefixer` within our build to output necessary vendor prefixes. So this change should not be necessary.

PR #4307, opened 2019-01-29
chore: set max-line-length in stylelint so auto-formatters pick it up

mmalerba:


test/screenshot/.stylelintrc.yaml
```diff
@@ -3,6 +3,7 @@ rules:
   # CSS
   custom-media-pattern: "^(?:test|custom|report)-.+"
   custom-property-pattern: "^(?:test|custom|report)-.+"
+  max-line-length: 120
```
kfranqueiro: I think you probably meant to add this to the root-level `.stylelintrc.yaml`, not just the one under `test/screenshot`?

mmalerba: whoops, yes

mmalerba:
Oh this apparently causes some lint failures due to existing lines that are too long, I'll fix those and add them to this PR

kfranqueiro:
Fixed a few nits and updated from master. Running screenshot tests locally right now. Diffed CSS and only diffs were whitespace introduced inside the wrapped `calc` expressions in toolbar.

PR #4309, opened 2019-01-29
 feat(ripple): add feature targeting to remaining public mixins

mmalerba:
depends on #4308

### DIFF
--- a/packages/mdc-ripple/_mixins.scss
+++ b/packages/mdc-ripple/_mixins.scss
@@ -91,7 +91,9 @@
   &::before {
     @include mdc-feature-targets($feat-animation) {
       // Also transition background-color to avoid unnatural color flashes when toggling activated/selected state
-      transition: opacity $mdc-states-wash-duration linear, background-color $mdc-states-wash-duration linear;
+      transition:
+        opacity $mdc-states-wash-duration linear,
+        background-color $mdc-states-wash-duration linear;
     }
 
     @include mdc-feature-targets($feat-structure) {
--- a/packages/mdc-ripple/_mixins.scss
+++ b/packages/mdc-ripple/_mixins.scss
@@ -75,9 +75,11 @@
 
     -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
     will-change: transform, opacity;
+  }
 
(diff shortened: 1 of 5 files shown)


test/scss/feature-targeting.scss
```diff
@@ -4,35 +4,54 @@
 @import "@material/feature-targeting/functions";
 @import "@material/menu/mixins";
 @import "@material/radio/mixins";
+@import "@material/ripple/mixins";
 @import "@material/typography/mixins";
 
 // Button
-
-@include mdc-button($query: mdc-feature-any());
+.mdc-button {
```
kfranqueiro: I know we talked about this elsewhere, but looking at this, it occurs to me that it's incorrect (and thus misleading if anyone were to take this as reference, although they shouldn't) to invoke a component's main `mdc-foo` mixin within a selector. (This only goes for components, e.g. not elevation/typography)

If we're doing this per-component, perhaps we should reserve that treatment only for individual public mixins, not the main one capable of emitting all the base styles (which would emit this selector itself)?

Alternatively we could just have a `.test` selector around the entire set...

packages/mdc-ripple/_mixins.scss
```diff
@@ -89,9 +91,7 @@
   &::before {
     @include mdc-feature-targets($feat-animation) {
       // Also transition background-color to avoid unnatural color flashes when toggling activated/selected state
-      transition:
-        opacity $mdc-states-wash-duration linear,
-        background-color $mdc-states-wash-duration linear;
+      transition: opacity $mdc-states-wash-duration linear, background-color $mdc-states-wash-duration linear;
```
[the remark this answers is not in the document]
mmalerba: Yep... it's really nice for making quick changes without having to worry about getting the indentation right. I just wish it handled a few things differently

test/scss/feature-targeting.scss
mmalerba: yeah that's a good point, I'll just put `.test` around everything

mmalerba: went with `.mdc-test` in the end due to the linter

PR #4310, opened 2019-01-29
fix(text-field): Fix textfield placeholder & outline stroke animation

abhiomkar:
**Before**

textfield-anim-before

**After**

textfield-anim-after-40ms

### DIFF
--- a/packages/mdc-notched-outline/mdc-notched-outline.scss
+++ b/packages/mdc-notched-outline/mdc-notched-outline.scss
@@ -54,6 +54,7 @@
       border-top: 1px solid;
       border-bottom: 1px solid;
       pointer-events: none;
+      transition: border $mdc-notched-outline-transition-duration $mdc-animation-standard-curve-timing-function;
     }
 
     &__leading {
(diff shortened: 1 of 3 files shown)


packages/mdc-notched-outline/mdc-notched-outline.scss
```diff
@@ -51,6 +51,7 @@
     &__trailing {
       box-sizing: border-box;
       height: 100%;
+      transition: border $mdc-notched-outline-transition-duration $mdc-animation-standard-curve-timing-function;
```
kfranqueiro: This ends up having somewhat janky behavior in some browsers e.g. Chrome. In particular, border-width doesn't seem to transition very well and tends to just "snap" in the middle of the transition, which IMO looks jarring compared to the border snapping immediately when the user focuses or blurs the field.

abhiomkar: As discussed offline, it not very noticeable when running animation at 100% speed. But it is more prominent only when hover & focus happens immediately.

PR #4312, opened 2019-01-29
feat(ripple): updated ripple ts rewrite to remove extraneous code/types

moog16:


packages/mdc-ripple/index.ts
```diff
@@ -88,6 +88,10 @@ class MDCRipple extends MDCComponent<MDCRippleFoundation> implements
   }
 
   disabled = false;
+  /**
+   * this line may seem redundent, but without this the TS compiler does not
+   * think that `this` is of type RippleCapableSurface
+   */
```
mmalerba: does `root_!: Element;` not work?

mmalerba: Oh I see why this is happening, it's not just TS being dumb. It's because `MDCComponent` says `root_` is protected and `RippleCapableSurface` says its public, so you need to redeclare it as public for it to fulfill the interface

PR #4319, opened 2019-01-30
feat(checkbox): typescript conversion

moog16:
refs #4225


packages/mdc-checkbox/constants.ts
```diff
@@ -21,36 +21,32 @@
...
-/** @enum {string} */
-const cssClasses = {
-  UPGRADED: 'mdc-checkbox--upgraded',
-  CHECKED: 'mdc-checkbox--checked',
-  INDETERMINATE: 'mdc-checkbox--indeterminate',
-  DISABLED: 'mdc-checkbox--disabled',
-  ANIM_UNCHECKED_CHECKED: 'mdc-checkbox--anim-unchecked-checked',
-  ANIM_UNCHECKED_INDETERMINATE: 'mdc-checkbox--anim-unchecked-indeterminate',
-  ANIM_CHECKED_UNCHECKED: 'mdc-checkbox--anim-checked-unchecked',
+const cssClasses: {[key: string]: string} = {
```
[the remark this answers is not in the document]
moog16: I'll also open a PR to update mdc-base.

packages/mdc-checkbox/foundation.ts
```diff
@@ -194,6 +164,7 @@ class MDCCheckboxFoundation extends MDCFoundation {
     } = MDCCheckboxFoundation.cssClasses;
 
     switch (oldState) {
+    // @ts-ignore:no-switch-case-fall-through already existing fallthrough
```
acdvorak: Re-enable linter after this `switch()` block
(or use `@tslint-ignore-line` / `@tslint-ignore-next-line` instead).

packages/mdc-checkbox/index.ts
```diff
@@ -21,78 +21,90 @@
...
+import {MDCSelectionControl} from '@material/selection-control/index';
+
 import MDCCheckboxFoundation from './foundation';
-import {MDCRipple, MDCRippleFoundation} from '@material/ripple/index';
-import {getMatchesProperty} from '@material/ripple/util';
 
-/** @const {!Array<string>} */
-const CB_PROTO_PROPS = ['checked', 'indeterminate'];
+const {getMatchesProperty} = util;
+const CB_PROTO_PROPS: string[] = ['checked', 'indeterminate'];
```
acdvorak: Can't `string[]` be inferred?

packages/mdc-checkbox/index.ts
```diff
@@ -21,78 +21,90 @@
...
-    super(...args);
-
-    /** @private {!MDCRipple} */
-    this.ripple_ = this.initRipple_();
-    /** @private {!Function} */
-    this.handleChange_;
-    /** @private {!Function} */
-    this.handleAnimationEnd_;
-  }
+  root_!: Element;
```
acdvorak: This is redundant with `MDCComponent` and missing the `protected` visibility modifier.

Either remove this line or change the type to `protected HTMLElement` and remove the cast below (`(this.root_ as HTMLElement).offsetWidth`).

packages/mdc-checkbox/index.ts
```diff
@@ -21,78 +21,90 @@
...
+  private ripple_: MDCRipple = this.initRipple_();
+  private handleChange_!: () => void;
```
acdvorak: Replace `() => void;` with `EventListener; // assigned in initialSyncWithDOM()`

packages/mdc-checkbox/index.ts
```diff
@@ -21,78 +21,90 @@
...
+  private handleAnimationEnd_!: () => void;
```
acdvorak: Replace `() => void;` with `EventListener; // assigned in initialSyncWithDOM()`

packages/mdc-checkbox/index.ts
kfranqueiro: I think we've been opting to keep `root_` as `Element` and cast as necessary so far, right? It looks like we avoid re-declaring this in other components we've converted.

acdvorak: Correct. I'm guessing Matt was overriding this as an `HTMLElement` at some point, and then changed it back to `Element` instead of removing it. We should be able to just remove this line.

mmalerba: I think this is actually because `root_` is `protected` in the super class, but it must be `public` to fulfill the `RippleCapableSurface` interface. Should definitely add a comment why this is being done though

acdvorak: I was wrong: public visibility for this property is required by `RippleCapableSurface`.

I'll add a comment to clarify this.

packages/mdc-checkbox/index.ts
acdvorak: Yes it can

packages/mdc-checkbox/index.ts
acdvorak: Done

packages/mdc-checkbox/index.ts
acdvorak: Done

packages/mdc-checkbox/index.ts
```diff
@@ -21,187 +21,165 @@
...
-    super.destroy();
-  }
 }
 
-/**
- * @param {ObjectPropertyDescriptor|undefined} inputPropDesc
- * @return {boolean}
- */
-function validDescriptor(inputPropDesc) {
+function validDescriptor(inputPropDesc: PropertyDescriptor | undefined): inputPropDesc is PropertyDescriptor {
```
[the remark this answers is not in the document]
acdvorak: This is actually a [Type Guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards).

If the function returns `true`, the TypeScript compiler knows that the value passed to `inputPropDesc` is a `PropertyDescriptor` type (and _not_ `undefined`), thus avoiding a bunch of `Object is possibly 'undefined'` errors and `if(... !== undefined)` checks.

PR #4292, opened 2019-01-24
feat(form-field): typescript conversion

moog16:
related https://github.com/material-components/material-components-web/issues/4225

dependent on https://github.com/material-components/material-components-web/pull/4286

PR #4320, opened 2019-01-30
 feat(switch): Convert JS to TypeScript

mmalerba:
depends on #4292


packages/mdc-base/component.ts
```diff
@@ -114,4 +114,4 @@ class MDCComponent<FoundationType extends MDCFoundation> {
   }
 }
 
-export default MDCComponent;
+export {MDCComponent as default, MDCComponent};
```
moog16: just curious why you decided to export both ways.

packages/mdc-switch/foundation.ts
```diff
@@ -21,47 +21,51 @@
...
-      addClass: (/* className: string */) => {},
-      removeClass: (/* className: string */) => {},
-      setNativeControlChecked: (/* checked: boolean */) => {},
-      setNativeControlDisabled: (/* disabled: boolean */) => {},
-    });
+  /** The default Adapter for the switch. */
+  static get defaultAdapter(): MDCSwitchAdapter {
+    return {
+      // tslint:disable:no-empty
+      addClass: () => {},
```
moog16: I've been changing these to `() => undefined`, then removing the tslint:disables

packages/mdc-base/component.ts
mmalerba: This way it doesn't break people using the default export, but it allows them to use the named one if they want. At some point you could deprecate the default one and eventually remove it.

moog16: Ok, if we do this then we should be doing this across all components then, correct?

mmalerba: If eliminating default exports is something you want to do eventually, then yeah I would do this kind of thing to all the ones that have default exports

packages/mdc-form-field/adapter.ts
```diff
@@ -21,39 +21,15 @@
...
-  deactivateInputRipple() {}
+interface MDCFormFieldAdapter {
+  activateInputRipple(): void;
+  deactivateInputRipple(): void;
+  deregisterInteractionHandler<K extends keyof GlobalEventHandlersEventMap>(
+    type: K, handler: (evt: GlobalEventHandlersEventMap[K],
+  ) => void): void;
+  registerInteractionHandler<K extends keyof GlobalEventHandlersEventMap>(
+    type: K, handler: (evt: GlobalEventHandlersEventMap[K],
+  ) => void): void;
```
moog16: update the export default line

moog16:
LGTM..but i think you are waiting for the form-field/selection-control PR to merge first, correct?

mmalerba:
rebased now that dependencies are merged

test/unit/mdc-switch/mdc-switch.test.js
```diff
@@ -27,7 +27,7 @@ import td from 'testdouble';
 
 import {supportsCssVariables} from '../../../packages/mdc-ripple/util';
 import {install as installClock} from '../helpers/clock';
-import {MDCSwitchFoundation, MDCSwitch} from '../../../packages/mdc-switch/index';
+import {MDCSwitchFoundation, MDCSwitch} from '../../../packages/mdc-switch';
```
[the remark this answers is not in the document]
mmalerba: I thought I saw this in another one of these conversion PRs which is why I did it here, will undo it though

PR #4321, opened 2019-01-30
fix(infrastructure): remove babel loader

moog16:
related https://github.com/material-components/material-components-web/issues/4225


moog16:
we should have travis run this step. It looks like `npm run dist` covers both build/build:min. But on 2nd thought it would be nicer to see both `build` and `build:min` separately. I can open a PR to update.

kfranqueiro:
Unit tests are also failing in IE11 due to a syntax error. I don't think we can completely remove babel-loader while we still have ES6-formatted unit tests...

It also occurs to me that even for purposes of just the typescript sources, we probably can't actually remove babel-loader until we remove all occurrences of `Object.assign`, which probably means we need to wait until after all of the modules are converted to typescript.

moog16:
I added back the babel loader for `js` files. I believe it to be unnecessary to have it in both places (ts files and js files`

moog16:
closing this PR...more to come in #4409

PR #4322, opened 2019-01-30
chore: Update golden screenshots after Edge upgrade

abhiomkar:


kfranqueiro:
We're repeatedly encountering a lot of flakes with this update (in addition to flakes we've worked around previously), so after discussing with the team we're opting to stop running screenshot tests on Edge (especially since Edge will eventually be Chromium-based), unless/until we find a convincing reason to do so in the future. See #4323.

kfranqueiro:
Superseded by #4323.

PR #4325, opened 2019-01-30
feat(icon-button): Convert to TypeScript

abhiomkar:
Refs #4225


packages/mdc-icon-button/index.ts
```diff
@@ -22,34 +22,25 @@
...
-  static attachTo(root) {
+class MDCIconButtonToggle extends MDCComponent<MDCIconButtonToggleFoundation> {
+  static attachTo(root: Element) {
     return new MDCIconButtonToggle(root);
   }
 
+  private ripple_: MDCRipple;
+  // TODO: type for event handler?
+  // tslint:disable-next-line:no-empty
+  private handleClick_ = (() => {});
```
moog16: try this: 
```

import {SpecificEventListener} from '@material/dom/index';

...
private handleClick_: SpecificEventListener<'click'>;
```

packages/mdc-icon-button/index.ts
```diff
@@ -22,34 +22,25 @@
...
+
+  // tslint:disable-next-line:no-any a component can pass in anything it needs to the constructor
```
moog16: this comment is not disabling the next line for my ide

moog16:
Don't forget to also update `js-bundle-factory.js` to 
```
iconButton: getAbsolutePath('/packages/mdc-icon-button/index.ts'),
```

and

packages/mdc-icon-button/adapter.ts
```diff
@@ -38,38 +36,22 @@
  * the component in your framework of choice. See architecture documentation
  * for more details.
  * https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md
- *
```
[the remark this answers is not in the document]
kfranqueiro: You mean remove the 2 lines above about closure compiler, right?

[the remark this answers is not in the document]
abhiomkar: Thanks for spotting this. I'm also considering to remove _**...managing below things...**_ part.

packages/mdc-icon-button/index.ts
abhiomkar: that didn't work. used `EventListenerOrEventListenerObject` type instead.

packages/mdc-icon-button/adapter.ts
```diff
@@ -21,55 +21,30 @@
  * THE SOFTWARE.
  */
 
-/* eslint no-unused-vars: [2, {"args": "none"}] */
-
 /**
- * Adapter for MDC Icon Button Toggle. Provides an interface for managing
- * - classes
- * - dom
```
moog16: I think we can leave this list

packages/mdc-icon-button/foundation.ts
```diff
@@ -22,37 +22,30 @@
...
+import {MDCIconButtonToggleAdapter} from './adapter';
 import {cssClasses, strings} from './constants';
 
-/**
- * @extends {MDCFoundation<!MDCIconButtonToggleAdapter>}
- */
-class MDCIconButtonToggleFoundation extends MDCFoundation {
-  static get cssClasses() {
+class MDCIconButtonToggleFoundation extends MDCFoundation<MDCIconButtonToggleAdapter> {
+  static get cssClasses(): {[key: string]: string} {
```
moog16: Let the type be inferred 

https://github.com/material-components/material-components-web/pull/4319#discussion_r252748427

packages/mdc-icon-button/index.ts
```diff
@@ -22,34 +22,27 @@
...
-  static attachTo(root) {
+class MDCIconButtonToggle extends MDCComponent<MDCIconButtonToggleFoundation> {
+  static attachTo(root: Element) {
     return new MDCIconButtonToggle(root);
   }
 
-  constructor(...args) {
-    super(...args);
+  private ripple_: MDCRipple;
+  private handleClick_: EventListenerOrEventListenerObject;
```
moog16: Did the @material/Dom import not work?

kfranqueiro:
Do we also need to update `index` to `index.ts` for icon-button in `material-components-web/index.js`? It looks like we've been doing that for the others as they've been converted.

packages/mdc-icon-button/index.ts
```diff
@@ -22,34 +22,27 @@
...
-    /** @private {!MDCRipple} */
-    this.ripple_ = this.initRipple_();
-    /** @private {!Function} */
-    this.handleClick_;
-  }
+  constructor(
+    root: Element,
+    foundation?: MDCIconButtonToggleFoundation,
+    // tslint:disable-next-line:no-any a component can pass in anything it needs to the constructor
+    ...args: any[]
```
moog16: I don't think we event need `...args`. Its only used in `mdc-base/component`'s `initialize` method. And there is not `initialize method.

packages/mdc-icon-button/index.ts
moog16: if you add the line below you can replace `EventListenerOrEventListenerObject ` --> `private handleClick_: SpecificEventListener<'click'>;`


```
root_!: HTMLElement;
```

moog16:
@kfranquiero I think we may have missed a few component declarations in `material-components-web/index.js`. We can do it here to be consistent, but we will eventually remove all the `.ts` exts anyways during the clean up phase of https://github.com/material-components/material-components-web/issues/4225.

kfranqueiro:
@moog16 afaict we have done it for every package that's been hit so far, with the exception of animation which doesn't appear in material-components-web/index.js at all. I was assuming it was necessary for the all-in-one build to work?

moog16:
Don't think it is necessary since webpack resolves either `.js` or `.ts`.

packages/mdc-icon-button/index.ts
acdvorak: We probably shouldn't override the constructor. Instead, initialize the properties in `initialSyncWithDOM()` and mark them with `!`:

```js
  private ripple_!: MDCRipple;
  private handleClick_!: EventListenerOrEventListenerObject;
```

PR #4334, opened 2019-02-01
feat(list): ts conversion

moog16:
related #4225

### DIFF
--- a/packages/mdc-list/foundation.ts
+++ b/packages/mdc-list/foundation.ts
@@ -301,8 +301,8 @@ class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
     }
 
     if (this.selectedIndex_ >= 0 && this.selectedIndex_ !== index) {
-      this.adapter_.removeClassForElementIndex((this.selectedIndex_ as number), selectedClassName);
-      this.adapter_.setAttributeForElementIndex((this.selectedIndex_ as number), strings.ARIA_SELECTED, 'false');
+      this.adapter_.removeClassForElementIndex(this.selectedIndex_ as number, selectedClassName);
+      this.adapter_.setAttributeForElementIndex(this.selectedIndex_ as number, strings.ARIA_SELECTED, 'false');
     }
 
     this.adapter_.addClassForElementIndex(index, selectedClassName);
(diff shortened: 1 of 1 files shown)


packages/mdc-base/component.ts
```diff
@@ -98,7 +98,7 @@ class MDCComponent<FoundationType extends MDCFoundation> {
    * Fires a cross-browser-compatible custom event from the component root of the given type,
    * with the given data.
    */
-  emit(evtType: string, evtData: object, shouldBubble = false) {
+  emit(evtType: string, evtData: number | object, shouldBubble = false) {
```
mmalerba: I believe you can use `unknown` instead of `number | object` here. It will allow you to accept number, object, string, etc. but you won't be able to do things to it (e.g. access properties), which makes it safe, unlike `any`. [(read more here)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type)

acdvorak: Thanks for the suggestion!

Ken and I thought it might be better to make this method generic, so that the caller can express the exact type of `evtData`. WDYT?

mmalerba: Yeah you could do that too. I'm not sure specifying the type really does anything for you though, it doesn't get saved on the instance or used in the return value, it just gets handed off to browser APIs

moog16: @acdvorak, you mean `emit<T = object>(evtType: string, evtData: T, shouldBubble = false)`? I think it is better than just unknown, but to @mmalerba's point its not "safer" than it being unknown.

acdvorak: Fair point 😄 

However, `emit<T extends object>()` has the advantage that it prevents primitive types like `number` from being passed to `evtData`, which we were previously doing in `MDCList` (see #4356).

To avoid backward compatibility issues in the future, we should always pass some kind of `object` so that we can easily add more properties.

OOC, what's the difference between `emit<T extends object>()` and `emit<T = object>()`?

moog16:
That's true...when I first saw that we were just passing a number, I was
wondering if that worked correctly (I'm assuming it does).

`emit<T = object>` defaults `T` to `object`. `<T extends object` ensures
that `T` is an `object`. You could also write `<T extends object = object>`.

packages/mdc-base/component.ts
mmalerba: If you want to disallow primitives, you can just do `eventData: object`. It looked from the existing code like you wanted to be able to pass objects or numbers (or I assumed anything else which is why I suggested `unknown`)

acdvorak: True, but making it generic will also help protect us against accidentally passing the wrong data. E.g.:

```ts
interface Foo {
  foo: number;
}

interface Bar {
  bar: string;
}

this.emit<Foo>('whatever', {bar: '...'}); // Error
this.emit('whatever', {bar: '...'}); // "Valid", but incorrect
```

PR #4338, opened 2019-02-01
feat(layout-grid): Add order device type

Jelmer7:
mdc-layout-grid__cell--order-{INDEX}-{TYPE_OF_DEVICE}


packages/mdc-layout-grid/mdc-layout-grid.scss
```diff
@@ -79,6 +79,17 @@
       @include mdc-layout-grid-cell-order($i);
     }
   }
+    
+  //Order per device.
+  @each $size in map-keys($mdc-layout-grid-breakpoints) {
```
abhiomkar: Can you please provide some context about this change?

Thanks!

Jelmer7: This extends the current `mdc-layout-grid__cell--order-<INDEX>` with `-<TYPE_OF_DEVICE>`.
It makes it possible to have a different order of elements on different types of devices.

In my current project I'm using this when on desktop I want to have two cells with span 6. On phone I want to display both cells with a span of 4, but the second cell first.

Using  the current `mdc-layout-grid__cell--order-<INDEX>` would mean the second cell would also show first (left) on desktop, or last on phone (The same as no order). Wich is not what I want.

Thanks!

SampsonCrowley: +1 for this, having an order class doesn't make much sense when you can't even use it to change the layout based on available size

aeromac:
I can't think of a reason why mdc-layout-grid__cell--order would be needed except for wanting to use it to target a device type

SampsonCrowley:
@aeromac There are accessibility reasons. Content and visible layout order aren't always the same. It's just much more necessary for media size reasons

anlexN:
can we solve PR?

PR #4351, opened 2019-02-04
feat(menu-surface): Update setPosition adapter API to use numeric values

acdvorak:
Accepting `string` values implies that we support all valid CSS units (`em`, `rem`, `%`, etc.), but in truth we only support `px`.

Changing the type of the properties from `string` to `number` makes this clearer, and also simplifies converting the package to TypeScript in PR #4273.

BREAKING CHANGE: `MDCMenuSurfaceAdapter.setPosition()` now expects an object with properties of type `number` rather than `string`. E.g., `setPosition({top: '5px', left: '10px'})` is now `setPosition({top: 5, left: 10})`.


packages/mdc-menu-surface/foundation.js
```diff
@@ -425,8 +425,8 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
     }
 
     for (const prop in position) {
-      if (position.hasOwnProperty(prop) && position[prop] !== '0') {
-        position[prop] = `${parseInt(position[prop], 10)}px`;
+      if (position.hasOwnProperty(prop)) {
+        position[prop] = parseInt(position[prop], 10);
```
kfranqueiro: Shouldn't parseInt go away in this PR?

packages/mdc-menu-surface/foundation.js
```diff
@@ -425,8 +425,8 @@ class MDCMenuSurfaceFoundation extends MDCFoundation {
     }
 
     for (const prop in position) {
-      if (position.hasOwnProperty(prop) && position[prop] !== '0') {
-        position[prop] = `${parseInt(position[prop], 10)}px`;
+      if (position.hasOwnProperty(prop)) {
```
kfranqueiro: This is changing behavior such that `position[prop]` is set even if `position[prop]` is 0. Do we want that?

packages/mdc-menu-surface/index.js
```diff
@@ -258,10 +258,10 @@ class MDCMenuSurface extends MDCComponent {
         return {x: window.pageXOffset, y: window.pageYOffset};
       },
       setPosition: (position) => {
-        this.root_.style.left = 'left' in position ? position.left : null;
-        this.root_.style.right = 'right' in position ? position.right : null;
-        this.root_.style.top = 'top' in position ? position.top : null;
-        this.root_.style.bottom = 'bottom' in position ? position.bottom : null;
+        this.root_.style.left = 'left' in position ? `${position.left}px` : '';
```
kfranqueiro: Why did this change from `null` to `''` here, if we're not doing that in the TypeScript PR? (It occurred to me to suggest this there, but I figured we were preserving existing behavior.)

packages/mdc-menu-surface/foundation.js
acdvorak: The existing code ensured that we didn't add a `"px"` suffix to values of `0`.

That's no longer necessary, so I removed this entire `for` loop.

packages/mdc-menu-surface/foundation.js
acdvorak: Yep—not sure how I missed these.

packages/mdc-menu-surface/index.js
acdvorak: The Closure Compiler test suddenly started complaining that these values can't be `null` 🤷‍♂️

acdvorak: CSS specs seem to indicate that `null` is _not_ a valid value. Empty string is preferred.
So Closure Compiler is [technically correct](https://i.imgur.com/PpV0evc.png); [TypeScript is wrong](https://github.com/Microsoft/TypeScript/blob/2f9218f346bbc606cccc4aebfca1c2c4f3167fee/lib/lib.dom.d.ts#L2435-L2436).

* https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty

    > `value` _Optional_ is a `DOMString` containing the new property value. If not specified, treated as the **empty string**.

* https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue

    > Return value is a `DOMString` containing the value of the property. If not set, returns the **empty string**.

* https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface
* https://drafts.csswg.org/cssom/#cssomstring
* https://heycam.github.io/webidl/#idl-DOMString

    > Note: Note also that `null` is not a value of type `DOMString`. To allow `null`, a nullable `DOMString`, written as `DOMString?` in IDL, needs to be used.

packages/mdc-menu-surface/foundation.js
```diff
@@ -26,7 +26,7 @@
  *   top: number,
  *   right: number,
  *   bottom: number,
- *   left: number
+ *   left: number,
```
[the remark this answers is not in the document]
acdvorak: It's Closure, so probably not. Removed. Thanks!

PR #4356, opened 2019-02-05
fix(list): Update default notifyAction impl to emit object

### LINKED ISSUE #4355: MDCList notifyAction() emits primitive event detail
### Problem

We're currently emitting a `number` value for the `CustomEvent.detail` property in
`MDCListAdapter.notifyAction()`:

https://github.com/material-components/material-components-web/blob/af950fcc87a08db99ade9a8a100fe9d3e9de9f60/packages/mdc-list/index.js#L279-L281

This is [technically allowed](https://i.imgur.com/PpV0evc.png) by TypeScript's `CustomEvent<T>` interface:

https://github.com/Microsoft/TypeScript/blob/2f9218f346bbc606cccc4aebfca1c2c4f3167fee/lib/lib.dom.d.ts#L3415-L3422

```ts
interface CustomEvent<T = any> extends Event {
    /**
     * Returns any custom data event was created with.
     * Typically used for synthetic events.
     */
    readonly detail: T;
    initCustomEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T): void;
}
```

However, emitting a primitive value for an event `detail` is not future-resistant. If we ever need to pass _more_ data, we'll be forced to break backward compatibility for a large number of customers.

### Solution

Rather than break backward compatibility _later_—after lots of downstream consumers have already wrapped the new v0.44.0 `MDCList` API from PR #4144—we should break it _now_.

This will also simplify PR #4334 and PR #4342.

acdvorak:
Fixes #4355

BREAKING CHANGE: The default `MDCListAdapter.notifyAction()` implementation now emits an object of type `{index: number}` rather than a primitive `number` directly.


acdvorak:
Updated unit test to `assert.deepEqual(event.detail, {index: 3})`.

PR #4358, opened 2019-02-05
feat(base): Add generic type signatures to listen/unlisten/emit

acdvorak:
Also moves `mdc-dom/events.ts` to `mdc-base/types.ts`
and adds `CustomEventListener<E extends Event>` type.

### DIFF
--- a/packages/mdc-checkbox/index.ts
+++ b/packages/mdc-checkbox/index.ts
@@ -23,7 +23,7 @@
 
 import {getCorrectEventName} from '@material/animation/index';
 import MDCComponent from '@material/base/component';
-import {EventType, SpecificEventListener} from '@material/base/types';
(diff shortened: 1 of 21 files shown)


packages/mdc-checkbox/index.ts
```diff
@@ -23,7 +23,7 @@
 
 import {getCorrectEventName} from '@material/animation/index';
 import MDCComponent from '@material/base/component';
-import {EventType, SpecificEventListener} from '@material/dom/index';
+import {EventType, SpecificEventListener} from '@material/base/types';
```
moog16: I think we should also be importing/exporting from @material/base/index

acdvorak: Done

PR #4363, opened 2019-02-07
feat(snackbar): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-snackbar/index.ts
```diff
@@ -21,61 +21,48 @@
...
-    this.announce_ = announceFactory();
+  initialize(announcerFactory: AnnouncerFactory = () => util.announce) {
+    this.announce_ = announcerFactory();
   }
 
   initialSyncWithDOM() {
-    this.surfaceEl_ = /** @type {!HTMLElement} */ (this.root_.querySelector(SURFACE_SELECTOR));
-    this.labelEl_ = /** @type {!HTMLElement} */ (this.root_.querySelector(LABEL_SELECTOR));
-    this.actionEl_ = /** @type {!HTMLElement} */ (this.root_.querySelector(ACTION_SELECTOR));
+    this.surfaceEl_ = this.root_.querySelector<HTMLElement>(SURFACE_SELECTOR)!;
```
kfranqueiro: We're casting to `HTMLElement` here but defining these as `Element` above? If they were already safely `HTMLElement` in the original implementation, then we can probably preserve that... either way, should we be consistent?

packages/mdc-snackbar/util.ts
```diff
@@ -77,8 +76,12 @@ function announce(ariaEl, labelEl = ariaEl) {
   labelEl.setAttribute(ARIA_LIVE_LABEL_TEXT_ATTR, labelText);
 
   setTimeout(() => {
-    // Allow screen readers to announce changes to the DOM again.
-    ariaEl.setAttribute('aria-live', priority);
+    if (priority === null) {
```
kfranqueiro: Is this fixing a bug at the same time as the conversion? We should probably carry this over to master so it gets noted...

kfranqueiro: Discussed w/ Andy; it's technically fixing a bug, but for an edge case that won't occur in valid usage of the component. This is probably fine to just fix in the TS branch.

packages/mdc-snackbar/foundation.ts
```diff
@@ -173,62 +145,44 @@ class MDCSnackbarFoundation extends MDCFoundation {
...
+    if (isEscapeKey && this.getCloseOnEscape()) {
       this.close(REASON_DISMISS);
     }
   }
 
-  /**
-   * @param {!MouseEvent} evt
-   */
-  handleActionButtonClick(evt) {
+  handleActionButtonClick(_evt: MouseEvent) {
```
moog16: can you remove this argument since you're not using it?

packages/mdc-snackbar/index.ts
moog16: I also notice you're `!` to each querySelector call. Should we throw an error if they don't exist?

packages/mdc-snackbar/index.ts
```diff
@@ -21,61 +21,48 @@
  * THE SOFTWARE.
  */
 
+import {SpecificEventListener} from '@material/base';
 import {MDCComponent} from '@material/base/component';
-import MDCSnackbarFoundation from './foundation';
+import * as ponyfill from '@material/dom/ponyfill';
 import {strings} from './constants';
+import {MDCSnackbarFoundation} from './foundation';
+import {Announcer, AnnouncerFactory} from './types';
```
moog16: is this something we also want to export from `index`

packages/mdc-snackbar/index.ts
acdvorak: Removed `HTMLElement` generic type param

packages/mdc-snackbar/util.ts
acdvorak: Nope, it's just to make the TypeScript compiler happy.
`getAttribute()` can technically return `null`.
It's not a use case we support, however; the `aria-live` attribute should _always_ be present.

packages/mdc-snackbar/foundation.ts
acdvorak: I considered doing removing it, but I thought it would be more future-proof to keep it.

If we end up needing to inspect the event object in the future—e.g., to avoid taking an action if the shift/meta keys are pressed—we won't need to introduce a breaking change, because we'll already have access to the event.

Removing the param would technically be a breaking change too, so we'd have to create a separate PR to make the change in `master` as well.

kfranqueiro: Yeah... from the perspective of this being a straight conversion, might as well leave it, since I guess it was unused even on master then, right?

PR #4367, opened 2019-02-07
docs(text-field): Add breaking change to changelog for helper-line

kfranqueiro:
#4244 didn't include a BREAKING CHANGE note when it was merged. This is super important because all existing helper-text usages are broken by its changes.

Thanks @jamesmfriedman for pointing this out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 
 ### BREAKING CHANGES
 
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
^ You might want to update the Textfield Helper Text docs. They have no mention of needing to be nested.

kfranqueiro:
RE @jamesmfriedman's comment, I checked with him on Discord about this. The latest docs do mention the helper-line element, but they haven't been mirrored to material.io yet. I have a ticket open with the site team now to do that.

PR #4370, opened 2019-02-07
docs(dialog): Expand wording RE layout-sensitive children

### LINKED ISSUE #4352: MDCCheckbox initial ripple hover not centred when initialisation of JS in a MDCDialog
## Bugs

There seems to be an issue with the initial hover ripple with Checkboxes when they are in dialogs. (See CodePen below) This issue seems to happen when initialising the Checkbox JS to the checkboxes both before, or after the dialog has opened. If you comment the `let checkbox = new mdc.checkbox.MDCCheckbox(eve);` line in the CodePen, then the hover state fixes its self.

When viewing the CodePen, don't resize the viewing window as doing that fixes the issue for some reason.

> 0.43.1

> Chrome

### What OS are you using?

> MacOS 10.14.2

> Open [CodePen](https://codepen.io/anon/pen/wNqdLq)
> **DON'T** resize the CodePen window, as that seems to fix the issue

> When the dialog is opened, the ripple hover is centred on the checkboxes

> The initial hover for the ripple is positioned in the top left corner of each checkbox 

Image of when the dialog is opened:
screen shot 2019-02-05 at 4 33 49 pm

Image of when the CodePen window is resized:
screen shot 2019-02-05 at 4 34 01 pm
Gif of the issue:
bug5

kfranqueiro:
Resolves #4352.

### DIFF
--- a/packages/mdc-dialog/README.md
+++ b/packages/mdc-dialog/README.md
@@ -89,8 +89,9 @@ const dialog = new MDCDialog(document.querySelector('.mdc-dialog'));
 > See [Importing the JS component](../../docs/importing-js.md) for more information on how to import JavaScript.
 
 MDC Dialog makes no assumptions about what will be added to the `mdc-dialog__content` element. Any List, Checkboxes,
-etc. must also be instantiated. Additionally, call `layout` on any applicable components within the content when
-`MDCDialog:opened` is emitted.
+etc. must also be instantiated. If your dialog contains any layout-sensitive components, you should wait until
+`MDCDialog:opened` is emitted to instantiate them (or call `layout` on them) so that the dialog's transition finishes
+first.
 
 For example, to instantiate an MDC List inside of a Simple or Confirmation Dialog:
 
--- a/packages/mdc-dialog/README.md
+++ b/packages/mdc-dialog/README.md
@@ -104,6 +104,9 @@ dialog.listen('MDCDialog:opened', () => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


packages/mdc-dialog/README.md
```diff
@@ -89,8 +89,9 @@ const dialog = new MDCDialog(document.querySelector('.mdc-dialog'));
 > See [Importing the JS component](../../docs/importing-js.md) for more information on how to import JavaScript.
 
 MDC Dialog makes no assumptions about what will be added to the `mdc-dialog__content` element. Any List, Checkboxes,
-etc. must also be instantiated. Additionally, call `layout` on any applicable components within the content when
-`MDCDialog:opened` is emitted.
+etc. must also be instantiated. If your dialog contains any layout-sensitive components, you should wait until
```
acdvorak: It might be clearer if we include an example of what we mean by "layout-sensitive". Something like:

```markdown
If your dialog contains any layout-sensitive components
(e.g., some elements appear mispositioned or incorrectly sized),
you should wait until the `MDCDialog:opened` event is emitted to instantiate them
(or call `layout()` on them) so that the dialog's transition finishes first.
```

or:

```markdown
If your dialog contains any layout-sensitive components,
you should wait until the `MDCDialog:opened` event is emitted to instantiate them
(or call `layout()` on them) so that the dialog's transition finishes first.
Mispositioned or incorrectly-sized elements (e.g., ripples or floating labels)
are a strong indication that child components are being instantiated
before the dialog has finished opening.
```

kfranqueiro: Good idea. I added it as a NOTE after the code snippet instead to avoid elongating the prose here.

PR #4371, opened 2019-02-07
fix(MenuSurface): Corrects open animation issue

jamesmfriedman:
The opening animation is no longer present the first time the menu opens, as seen on the MDC docs site, as well as causing issues in my React wrapper.

https://material-components.github.io/material-components-web-catalog/#/component/menu


jamesmfriedman:
Well, I failed linting... if one of the maintainers could help me out with that part, I'm pretty sure this still fixes the issue.

PR #4380, opened 2019-02-08
fix(chips): Flip leading icon margin when used in RTL contexts (#4379)

### LINKED ISSUE #4379: MDC Chips: leading icons not spaced correctly for RTL

> @material/chips @ 0.43.0

> I believe this to affect all browsers. Tested on Chrome (Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36)

### What OS are you using?

> MacOS 10.14.1

> 1. Go to the Chips [demo page](https://material-components.github.io/material-components-web-catalog/#/component/chips).
> 2. Add `dir="rtl"` to the root `<html>` element.
> 3. Observe incorrect spacing for chips with leading icons.

> The margin on the leading icon should flip from right to left edge.

> The margin remains on the right edge of the leading icon, causing the icon to butt up against the chip label.

screenshot 2019-02-08 at 09 25 50

WillsB3:
This fixes the margin of leading icons inside Chips from being applied to the incorrect side of the icon when the Chip appears in a page with RTL writing mode.

Fixes #4379.

### DIFF
--- a/packages/mdc-chips/_mixins.scss
+++ b/packages/mdc-chips/_mixins.scss
@@ -22,6 +22,7 @@
 
 @import "@material/elevation/mixins";
 @import "@material/ripple/mixins";
+@import "@material/rtl/mixins";
 @import "@material/theme/functions";
 @import "@material/theme/mixins";
 @import "@material/shape/mixins";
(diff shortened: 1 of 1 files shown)


kfranqueiro:
Nice catch and thanks for the PR! This is also reproducible with our screenshot test pages in RTL mode (which our screenshot test pages support, but we don't capture yet due to how much time it would add to test runs).

Before fix:

image

After fix:

image

PR #4383, opened 2019-02-08
feat(feature-targeting): target top-level styles for elevation, ripple,

mmalerba:
theme, and typography (part of #4227)

### DIFF
--- a/packages/mdc-elevation/_mixins.scss
+++ b/packages/mdc-elevation/_mixins.scss
@@ -25,6 +25,23 @@
 @import "@material/theme/variables";
 @import "./variables";
 
+@mixin mdc-elevation-top-level($query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @for $z-value from 0 through 24 {
+    .mdc-elevation--z#{$z-value} {
+      @include mdc-elevation($z-value, $query: $query);
+    }
+  }
+
+  .mdc-elevation-transition {
+    @include mdc-feature-targets($feat-structure) {
+      transition: mdc-elevation-transition-value();
+      will-change: $mdc-elevation-property;
+    }
+  }
+}
+
 // Applies the correct CSS rules to an element to give it the elevation specified by $z-value.
 // The $z-value must be between 0 and 24.
 // If $color has an alpha channel, it will be ignored and overridden. To increase the opacity of the shadow, use
--- a/packages/mdc-elevation/_mixins.scss
+++ b/packages/mdc-elevation/_mixins.scss
@@ -26,6 +26,7 @@
 @import "./variables";
 
 @mixin mdc-elevation-top-level($query: mdc-feature-all()) {
+  $feat-animation: mdc-feature-create-target($query, animation);
   $feat-structure: mdc-feature-create-target($query, structure);
 
   @for $z-value from 0 through 24 {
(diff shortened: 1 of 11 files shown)


packages/mdc-elevation/_mixins.scss
```diff
@@ -25,6 +25,23 @@
 @import "@material/theme/variables";
 @import "./variables";
 
+@mixin mdc-elevation-top-level($query: mdc-feature-all()) {
```
mmalerba: For a couple of them I had to do `mdc-<package>-top-level` since `mdc-<package>` was taken. Should I go back and change them all to match for consistency?

packages/mdc-elevation/_mixins.scss
```diff
@@ -25,6 +25,23 @@
...
+
+  @for $z-value from 0 through 24 {
+    .mdc-elevation--z#{$z-value} {
+      @include mdc-elevation($z-value, $query: $query);
+    }
+  }
+
+  .mdc-elevation-transition {
+    @include mdc-feature-targets($feat-structure) {
+      transition: mdc-elevation-transition-value();
```
kfranqueiro: Wouldn't this be under animation?

packages/mdc-elevation/_mixins.scss
kfranqueiro: That's a fair point; it'd probably be less confusing that way (esp. for stuff like mdc-theme).

Also, do you think `mdc-foo-baseline` would work for naming rather than `mdc-foo-top-level`?

packages/mdc-theme/_mixins.scss
```diff
@@ -20,8 +20,47 @@
...
 
+@import "@material/feature-targeting/functions";
+@import "@material/feature-targeting/mixins";
 @import "./variables";
 
+@mixin mdc-theme($query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+
+  :root {
+    @include mdc-feature-targets($feat-color) {
```
kfranqueiro: Is there a reason this feature-targets is outside the `@each` but the ones below are inside?

packages/mdc-ripple/mdc-ripple.scss
```diff
@@ -20,36 +20,5 @@
 // THE SOFTWARE.
 //
 
-@import "@material/animation/functions";
-@import "@material/theme/mixins";
-@import "./common";
```
kfranqueiro: I don't think you want to remove `common` here... all of the emitted keyframes styles disappeared.

packages/mdc-elevation/_mixins.scss
mmalerba: That would work, though in the case of typography people might think it has something to do with baseline alignment

packages/mdc-elevation/_mixins.scss
mmalerba: Whoops, I think I read that as `transform` in my head

packages/mdc-theme/_mixins.scss
mmalerba: This one doesn't have any selectors inside of it, just properties

packages/mdc-elevation/_mixins.scss
mmalerba: per slcak discussion, going with `mdc-foo-core-styles`

packages/mdc-theme/_mixins.scss
```diff
@@ -20,8 +20,47 @@
 // THE SOFTWARE.
 //
 
+@import "@material/feature-targeting/functions";
```
kfranqueiro: I think we need to add feature-targeting as a package dependency

PR #4377, opened 2019-02-08
feat(text-field): Convert JS to TypeScript

acdvorak:
Refs #4225

PR #4386, opened 2019-02-09
feat(select): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-select/types.ts
```diff
@@ -0,0 +1,50 @@
...
+  detail: SelectEventDetail;
+}
+
+export interface SelectEventDetail {
+  value: string;
+  index: number;
+}
+
+export type LineRippleFactory = (el: Element) => MDCLineRipple;
+export type HelperTextFactory = (el: Element) => MDCTextFieldHelperText;
```
acdvorak: Depends on #4377

packages/mdc-select/package.json
```diff
@@ -25,6 +25,7 @@
     "@material/ripple": "^0.44.0",
     "@material/rtl": "^0.42.0",
     "@material/shape": "^0.43.0",
+    "@material/textfield": "^0.44.0",
```
moog16: what is this for?

kfranqueiro: I am also curious about this, as we have thus far avoided having text field or select rely on each other (otherwise we would've avoided duplicating some sub-components here)

packages/mdc-list/index.ts
```diff
@@ -33,8 +33,8 @@ class MDCList extends MDCComponent<MDCListFoundation> {
     this.foundation_.setVerticalOrientation(value);
   }
 
-  get listElements(): Element[] {
-    return [].slice.call(this.root_.querySelectorAll(strings.ENABLED_ITEMS_SELECTOR));
+  get listElements(): HTMLElement[] {
```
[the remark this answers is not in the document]
acdvorak: I think it was because we're calling `.focus()` on these items in a few places, and I assumed `.mdc-list-item` would always be an `HTMLElement` anyway, so I thought we could avoid type assertions by just declaring it an `HTMLElement`.

However, it looks like it's pretty easy to change this back to `Element` if you prefer (take a look at commit 91f5975bfc68af5c21a30ad34653fee6a080683e).

I figured it's probably OK to _return_ a more specific subtype, as long as our _arguments_ are still `Element` for Closure users.

WDYT?

packages/mdc-select/icon/foundation.ts
```diff
@@ -22,66 +22,59 @@
...
+  private savedTabIndex_!: string | null; // assigned in initialSyncWithDOM()
+  private readonly interactionHandler_!: EventListener; // assigned in initialSyncWithDOM()
 
-    /** @private {string?} */
-    this.savedTabIndex_ = null;
+  constructor(adapter?: Partial<MDCSelectIconAdapter>) {
+    super({...MDCSelectIconFoundation.defaultAdapter, ...adapter});
 
-    /** @private {function(!Event): undefined} */
+    this.savedTabIndex_ = null;
```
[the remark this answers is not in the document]
acdvorak: Correct.

The compilation error is only an issue for classes that extend `MDCComponent`, because the `MDCComponent` constructor calls instance methods (`initialize()` and `initialSyncWithDOM()`) _before_ the subclass constructor runs, so any values set by those instance methods get overwritten with `null` in the subclass constructor 🤦‍♂️.

`MDCFoundation` doesn't do that, thankfully.

packages/mdc-select/package.json
acdvorak: Woops, removed

packages/mdc-select/foundation.ts
```diff
@@ -22,85 +22,78 @@
...
+   * See {@link MDCSelectAdapter} for typing information on parameters and return types.
    */
-  static get defaultAdapter() {
-    return /** @type {!MDCSelectAdapter} */ ({
-      addClass: (/* className: string */) => {},
-      removeClass: (/* className: string */) => {},
+  static get defaultAdapter(): MDCSelectAdapter {
+    // tslint:disable:object-literal-sort-keys
+    return {
+      addClass: (/* className: string */) => undefined,
```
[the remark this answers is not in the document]
acdvorak: Woops, removed

packages/mdc-menu/types.ts
```diff
@@ -24,6 +24,14 @@
 import {MDCList} from '@material/list/index';
 import {MDCMenuSurface} from '@material/menu-surface/index';
 
+export interface MenuItemEvent extends Event {
+  detail: MenuItemEventDetail;
+}
+
+export interface DefaultMenuItemEvent extends Event {
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-menu/types.ts
```diff
@@ -24,6 +24,14 @@
 import {MDCList} from '@material/list/index';
 import {MDCMenuSurface} from '@material/menu-surface/index';
 
+export interface MenuItemEvent extends Event {
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-select/icon/foundation.ts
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-select/types.ts
acdvorak: This should have been `MDCSelectHelperText`. Fixed.

Now `mdc-select` doesn't need to depend on `mdc-textfield` package anymore.

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+
+  private menu_!: MDCMenu | null; // assigned in enhancedSelectSetup_()
+  private menuOpened_ = false;
+
+  // Exactly one of these fields must be non-null.
+  private nativeControl_!: HTMLSelectElement | null;
+  private selectedText_!: HTMLElement | null;
+
+  private hiddenInput_!: HTMLInputElement | null;
+  private leadingIcon_!: MDCSelectIcon | null;
```
kfranqueiro: Why is this `!` and `| null` when AFAICT it's either assigned a value or not assigned at all?

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+  private helperText_!: MDCSelectHelperText | null;
+  private menuElement_!: Element | null;
+  private ripple!: MDCRipple | null;
+  private lineRipple_!: MDCLineRipple | null;
+  private label_!: MDCFloatingLabel | null;
```
kfranqueiro: Similar comment here to one I left on #4377 - do all of these need to be `!` and `| null` or can some or all of them simply be left as undefined if the element isn't present?

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+  private handleFocus_!: SpecificEventListener<'focus'>; // assigned in initialize()
+  private handleBlur_!: SpecificEventListener<'blur'>; // assigned in initialize()
+  private handleClick_!: SpecificEventListener<'click'>; // assigned in initialize()
+  private handleKeydown_!: SpecificEventListener<'keydown'>; // assigned in initialize()
+  private handleMenuOpened_!: EventListener; // assigned in initialize()
+  private handleMenuClosed_!: EventListener; // assigned in initialize()
+  private handleMenuSelected_!: CustomEventListener<MenuItemEvent>; // assigned in initialize()
+  private validationObserver_!: MutationObserver; // assigned in initialize()
+
+  get value(): string {
```
kfranqueiro: I'm a little confused why the getters/setters are before initialize/destroy/etc. this time? I guess this might be consistent with the original but if git is going to consider it a rewrite anyway, we might as well make this more consistent with the others.

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+    return selectedIndex;
+  }
+
+  set selectedIndex(selectedIndex: number) {
+    this.foundation_.setSelectedIndex(selectedIndex);
+  }
+
+  get disabled(): boolean {
+    return this.root_.classList.contains(cssClasses.DISABLED) ||
+        (this.nativeControl_ ? this.nativeControl_.disabled : false);
```
kfranqueiro: Extra indentation?

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+    if (this.validationObserver_) {
+      this.validationObserver_.disconnect();
+    }
+
+    super.destroy();
+  }
+
+  getDefaultFoundation(): MDCSelectFoundation {
+    return new MDCSelectFoundation({
+          ...(this.nativeControl_ ? this.getNativeSelectAdapterMethods_() : this.getEnhancedSelectAdapterMethods_()),
```
kfranqueiro: Indentation?

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+
+  private getNativeSelectAdapterMethods_() {
+    // tslint:disable:object-literal-sort-keys
+    return {
+      getValue: () => this.nativeControl_!.value,
+      setValue: (value: string) => this.nativeControl_!.value = value,
+      openMenu: () => undefined,
+      closeMenu: () => undefined,
+      isMenuOpen: () => false,
+      setSelectedIndex: (index: number) => this.nativeControl_!.selectedIndex = index,
```
kfranqueiro: More adapter methods that are supposed to have no return value

packages/mdc-select/types.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export type SelectEvent = CustomEvent<SelectEventDetail>;
+
+export interface SelectEventDetail {
+  value: string;
+  index: number;
+}
+
+// TODO(acdvorak): Every component should export its own factory and event types.
```
kfranqueiro: Define "every"? This is pretty particular to text field and select sub-components.

packages/mdc-base/component.ts
```diff
@@ -103,16 +103,13 @@ class MDCComponent<FoundationType extends MDCFoundation> {
    * Fires a cross-browser-compatible custom event from the component root of the given type,
    * with the given data.
    */
-  emit<T extends object>(evtType: string, evtData: T, shouldBubble = false) {
+  emit<T extends object>(evtType: string, detail: T, bubbles = false) {
```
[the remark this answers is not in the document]
moog16: I think it is to inline the argument in the object below.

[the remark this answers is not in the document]
kfranqueiro: That still seems like a random thing to do in base, in a PR for one specific other package...

packages/mdc-select/helper-text/index.ts
```diff
@@ -22,45 +22,31 @@
...
       addClass: (className) => this.root_.classList.add(className),
       removeClass: (className) => this.root_.classList.remove(className),
       hasClass: (className) => this.root_.classList.contains(className),
       setAttr: (attr, value) => this.root_.setAttribute(attr, value),
       removeAttr: (attr) => this.root_.removeAttribute(attr),
-      setContent: (content) => {
-        this.root_.textContent = content;
-      },
-    })));
+      setContent: (content) => this.root_.textContent = content,
```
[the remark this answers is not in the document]
moog16: I think it would be undefined for all of these, correct? Does `void` ==` `undefined` in TS? It seems like it must if the compiler doesn't complain.

packages/mdc-menu/types.ts
```diff
@@ -24,6 +24,9 @@
 import {MDCList} from '@material/list/index';
 import {MDCMenuSurface} from '@material/menu-surface/index';
 
+export type MenuItemEvent = CustomEvent<MenuItemEventDetail>;
+export type DefaultMenuItemEvent = CustomEvent<DefaultMenuItemEventDetail>;
```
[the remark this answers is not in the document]
acdvorak: I had a hard time figuring out what to call this type, so I'm definitely open to suggestions.

* `MenuItemEvent` is only used by foundation and adapter
* `DefaultMenuItemEvent` is only used by the component, and has an additional property on its `detail` object

What would you prefer to call it?

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+
+    // One of these elements must be non-null.
+    const element = this.nativeControl_ || this.selectedText_!;
+
+    element.addEventListener('change', this.handleChange_);
+    element.addEventListener('focus', this.handleFocus_);
+    element.addEventListener('blur', this.handleBlur_);
+
+    MOUSEDOWN_TOUCHSTART_EVENTS.forEach((evtType) => {
+      element.addEventListener(evtType, this.handleClick_ as EventListener);
```
moog16: since element can potentially be null, I think we need a guard here.

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+  destroy() {
+    // One of these elements must be non-null.
+    const element = this.nativeControl_ || this.selectedText_!;
+
+    element.removeEventListener('change', this.handleChange_);
+    element.removeEventListener('focus', this.handleFocus_);
+    element.removeEventListener('blur', this.handleBlur_);
+    element.removeEventListener('keydown', this.handleKeydown_);
+    MOUSEDOWN_TOUCHSTART_EVENTS.forEach((evtType) => {
+      element.removeEventListener(evtType, this.handleClick_ as EventListener);
```
moog16: same with here. I think we need a guard, since this.selectedText_ can be null

packages/mdc-select/helper-text/index.ts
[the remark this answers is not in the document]
acdvorak: IIUC, in TS, a return type of `void` doesn't guarantee a return value, but doesn't prohibit one, either. I can wrap it in `{}` if you prefer.

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+
+  private initRipple_(): MDCRipple {
+    // At least one of the elements must be non-null.
+    const element = this.nativeControl_ || this.selectedText_!;
+
+    // tslint:disable:object-literal-sort-keys
+    const foundation = new MDCRippleFoundation({
+      ...MDCRipple.createAdapter(this),
+      registerInteractionHandler: <E extends EventType>(evtType: E, handler: SpecificEventListener<E>) => {
+        element.addEventListener(evtType, handler);
```
moog16: do we need a guard for these method calls too?

packages/mdc-select/index.ts
acdvorak: They can be left undefined, but the `if` statements are a little more verbose than a ternary. It's really just a style choice.

packages/mdc-select/index.ts
```diff
@@ -0,0 +1,596 @@
...
+      getLabelWidth: () => this.label_ ? this.label_.getWidth() : 0,
+    };
+  }
+
+  /**
+   * Calculates where the line ripple should start based on the x coordinate within the component.
+   */
+  private getNormalizedXCoordinate_(evt: MouseEvent): number {
+    const targetClientRect = (evt.target as Element).getBoundingClientRect();
+    const xCoordinate = evt.clientX; // TODO(acdvorak): How should this be typed?
```
moog16: isn't this inferred to be a number already?

packages/mdc-select/helper-text/index.ts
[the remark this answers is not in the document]
kfranqueiro: From what I'm reading in the TS handbook, `void` can only be assigned `null` or `undefined`. This, however, returns `content`, as written.

packages/mdc-select/types.ts
acdvorak: A number of components need to instantiate ripples, icons, etc. I figured we might as well just make it a standard pattern to always export a factory type for every component so that we don't end up with duplicate factory definitions in lots of different components.

packages/mdc-select/index.ts
acdvorak: Done

packages/mdc-select/index.ts
acdvorak: Done

packages/mdc-select/index.ts
acdvorak: Done

packages/mdc-select/index.ts
acdvorak: Done

packages/mdc-select/index.ts
acdvorak: Done

packages/mdc-select/icon/foundation.ts
```diff
@@ -22,66 +22,58 @@
  */
 
 import {MDCFoundation} from '@material/base/foundation';
-import MDCSelectIconAdapter from './adapter';
+import {EventType} from '@material/base/index';
+import {MDCSelectIconAdapter} from './adapter';
 import {strings} from './constants';
 
+const CLICK_KEYDOWN_EVENTS: EventType[] = ['click', 'keydown'];
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-select/icon/README.md
```diff
@@ -114,8 +114,8 @@ Method Signature | Description
 `setAttr(attr: string, value: string) => void` | Sets an attribute with a given value on the icon element.
 `removeAttr(attr: string) => void` | Removes an attribute from the icon element.
 `setContent(content: string) => void` | Sets the text content of the icon element.
-`registerInteractionHandler(evtType: string, handler: EventListener) => void` | Registers an event listener for a given event.
-`deregisterInteractionHandler(evtType: string, handler: EventListener) => void` | Deregisters an event listener for a given event.
+`registerInteractionHandler(evtType: EventType, handler: EventListener) => void` | Registers an event listener for a given event.
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-list/index.ts
```diff
@@ -135,7 +135,7 @@ class MDCList extends MDCComponent<MDCListFoundation> {
         }
       },
       focusItemAtIndex: (index) => {
-        const element = this.listElements[index] as HTMLElement;
+        const element = this.listElements[index] as HTMLElement | null;
```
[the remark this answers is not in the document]
acdvorak: You are technically correct! 😀  Fixed.

There's a subtle but important difference between a type assertion with and without `| null` (or `| undefined`):

```ts
// This tells the TypeScript compiler that `element` will
// NEVER be null or undefined, so it won't yell at you
// if you try to access a property on `element`, which is dangerous.
const element = this.listElements[index] as HTMLElement;

// This tells tsc that `element` might not have a value, so it should
// yell at you if you try to access a property without a truthy check.
const element = this.listElements[index] as HTMLElement | undefined;
```

In this specific case, it doesn't matter quite as much because we already have a truthy check below (`if (element) { ... }`), but if we ever refactor this code, we could easily forget that `element` might be `undefined`.

Personally, I'd rather be semantically accurate by declaring that the type might be `undefined`.

packages/mdc-base/component.ts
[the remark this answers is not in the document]
acdvorak: Reverted

packages/mdc-select/index.ts
kfranqueiro: The component is only valid if nativeInput or selectedText is present, and the code above declares that selectedText will be defined if nativeInput isn't (there's also a comment above it).

packages/mdc-select/helper-text/index.ts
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-select/index.ts
acdvorak: `element` should never be null:

```ts
// One of these elements must be non-null.
const element = this.nativeControl_ || this.selectedText_!;
```

packages/mdc-select/index.ts
acdvorak: `element` should never be null:

```ts
// One of these elements must be non-null.
const element = this.nativeControl_ || this.selectedText_!;
```

packages/mdc-select/index.ts
acdvorak: `element` should never be null:

```ts
// One of these elements must be non-null.
const element = this.nativeControl_ || this.selectedText_!;
```

packages/mdc-select/index.ts
moog16: Should we then throw an error if both are missing? It seems like a helpful user error to know why the component wouldn't be working. I guess it is obvious if you don't have a select/list, just throwing out ideas.

packages/mdc-select/index.ts
acdvorak: The problem is that this method actually takes a `MouseEvent | TouchEvent`, and `clientX` isn't defined on `TouchEvent`. I'm actually not sure how this code works at all right now.

```ts
type PointerEventType = 'mousedown' | 'touchstart';

const POINTER_EVENTS: PointerEventType[] = ['mousedown', 'touchstart'];

class MDCSelect {
  private handleClick_!: SpecificEventListener<PointerEventType>; // assigned in initialize()

  initialSyncWithDOM() {
    this.handleClick_ = (evt) => {
      if (this.selectedText_) {
        this.selectedText_.focus();
      }
      this.foundation_.handleClick(this.getNormalizedXCoordinate_(evt));
    };

    POINTER_EVENTS.forEach((evtType) => {
      element.addEventListener(evtType, this.handleClick_ as EventListener);
    });
  }

  private getNormalizedXCoordinate_(evt: MouseEvent | TouchEvent): number {
    const targetClientRect = (evt.target as Element).getBoundingClientRect();
    const xCoordinate = evt.clientX; // TODO(acdvorak): How should this be typed?
    return xCoordinate - targetClientRect.left;
  }
}
```

moog16: OH! Wait I think there is a bug for this...I don't think it works.

moog16: Having a hard time finding it, but I do believe this was an issue in either React or Web or both.

moog16: https://github.com/material-components/material-components-web-react/issues/275#issuecomment-424272414

kfranqueiro: Didn't we fix this somewhere to dig out `touches[0]`, which does have clientX?

moog16: The above issue I linked to is text field. So I think we also need to fix for select.

packages/mdc-select/index.ts
acdvorak: Yeah, that sounds like a good idea. I'll do that and add a unit test.

packages/mdc-select/index.ts
acdvorak: Good call! A [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) should do the trick:

```ts
class MDCSelect {
  private getNormalizedXCoordinate_(evt: MouseEvent | TouchEvent): number {
    const targetClientRect = (evt.target as Element).getBoundingClientRect();
    const xCoordinate = this.isTouchEvent_(evt) ? evt.touches[0].clientX : evt.clientX;
    return xCoordinate - targetClientRect.left;
  }

  private isTouchEvent_(evt: MouseEvent | TouchEvent): evt is TouchEvent {
    return Boolean((evt as TouchEvent).touches);
  }
}
```

acdvorak: Done

packages/mdc-select/index.ts
acdvorak: Done

packages/mdc-select/index.ts
```diff
@@ -336,14 +239,111 @@ class MDCSelect extends MDCComponent<MDCSelectFoundation> implements RippleCapab
     super.destroy();
   }
 
+  get value(): string {
```
moog16: wow I wish Github would just move the initialize method lol

packages/mdc-select/index.ts
```diff
@@ -59,8 +59,10 @@ class MDCSelect extends MDCComponent<MDCSelectFoundation> implements RippleCapab
   private menuOpened_ = false;
 
   // Exactly one of these fields must be non-null.
-  private nativeControl_!: HTMLSelectElement | null;
-  private selectedText_!: HTMLElement | null;
+  private nativeControl_!: HTMLSelectElement | null; // assigned in initialize()
+  private selectedText_!: HTMLElement | null; // assigned in initialize()
+
+  private targetElement_!: HTMLElement; // assigned in initialize()
```
moog16: good clean up idea!

packages/mdc-select/index.ts
```diff
@@ -502,10 +502,14 @@ class MDCSelect extends MDCComponent<MDCSelectFoundation> implements RippleCapab
    */
   private getNormalizedXCoordinate_(evt: MouseEvent | TouchEvent): number {
     const targetClientRect = (evt.target as Element).getBoundingClientRect();
-    const xCoordinate = (evt as MouseEvent).clientX; // TODO(acdvorak): How does this work for TouchEvent?
+    const xCoordinate = this.isTouchEvent_(evt) ? evt.touches[0].clientX : evt.clientX;
     return xCoordinate - targetClientRect.left;
   }
 
+  private isTouchEvent_(evt: MouseEvent | TouchEvent): evt is TouchEvent {
```
moog16: what if we changed this to just `getXCoordinateFromEvent`? Then we just encapsulate this `this.isTouchEvent_(evt) ? evt.touches[0].clientX : evt.clientX;` logic...cause we still end up having to cast it anyways inside this method.

acdvorak: As discussed offline, it would need to look like this:

```ts
  private getNormalizedXCoordinate_(evt: MouseEvent | TouchEvent): number {
    const targetClientRect = (evt.target as Element).getBoundingClientRect();
    return this.getClientXFromEvent_(evt) - targetClientRect.left;
  }

  private getClientXFromEvent_(evt: MouseEvent | TouchEvent): number {
    const touchEvt = evt as TouchEvent;
    const mouseEvt = evt as MouseEvent;
    if (touchEvt.touches) {
      return touchEvt.touches[0].clientX;
    }
    return mouseEvt.clientX;
  }
```

Matt and I prefer the `isTouchEvent_()` style, so we're going to leave it as-is.

PR #4372, opened 2019-02-07
fix(ripple): Use mdc-dom.matches everywhere instead of a custom one (#4340)

julien1619:
Fixes #4340

`mdc-dom` implements a `matches()` method that wasn't used everywhere. These custom implementations prevented some SSR usage of the material-components library. This commit removes all the custom implementations.

I removed some tests that were testing a now removed method, and I handled to make the other pass.

PS: This is my first contribution here, I hope I did everything right!

BREAKING CHANGE: `getMatchesProperty()` has been removed from `@material/ripple/util`. Use `matches()` from `@material/dom/ponyfill` instead.

Issue #4340, opened 2019-02-01
Update ripple and tab-scroller to use mdc-dom/ponyfill.matches


kfranqueiro:
We added an `mdc-dom` package with a common `matches` function, but we still have 2 packages using their own version: Ripple and Tab Scroller.

Additionally, these packages use a version which is run eagerly ahead-of-time, which poses problems in SSR scenarios. (See https://github.com/material-components/material-components-web-react/issues/654)

**Edit:** Only Ripple actually runs the function eagerly, although both work in the same manner.

We should update these components to use `mdc-dom` and remove the duplicate functions.

PR #4389, opened 2019-02-09
chore: Remove Object.assign from TS files

acdvorak:
Refs #4225

This PR improves consistency across TS files:

* Replaces `Object.assign()` with `{...spread}` operator
* Changes foundation constructor arg types to `Partial<MDCFooAdapter>`
* Uses `ponyfill.matches()` from `mdc-dom` instead of `getMatchesProperty()` from `mdc-ripple`
    - `getMatchesProperty()` still exists to avoid breaking backward compatibility, but it is no longer used in TS
* Standardizes naming: `evt`, `evtType`, and `<K extends EventType>`
* Always import `/index` (e.g., `@material/foo/index` instead of `@material/foo`)

### DIFF
--- a/packages/mdc-base/component.ts
+++ b/packages/mdc-base/component.ts
@@ -51,6 +51,7 @@ class MDCComponent<FoundationType extends MDCFoundation> {
     this.initialSyncWithDOM();
   }
 
+  /* istanbul ignore next: method param only exists for typing purposes; it does not need to be unit tested */
   // tslint:disable-next-line:no-any a component can pass in anything it needs to initialize
   initialize(..._args: any[]) {
     // Subclasses can override this to do any additional setup work that would be considered part of a
--- a/packages/mdc-grid-list/foundation.ts
+++ b/packages/mdc-grid-list/foundation.ts
@@ -44,7 +44,7 @@ class MDCGridListFoundation extends MDCFoundation<MDCGridListAdapter> {
   private readonly resizeHandler_: EventListener;
   private resizeFrame_ = 0;
 
-  constructor(adapter: Partial<MDCGridListAdapter> = {}) {
+  constructor(adapter?: Partial<MDCGridListAdapter>) {
     super({...MDCGridListFoundation.defaultAdapter, ...adapter});
 
     this.resizeHandler_ = this.alignCenter.bind(this);
--- a/packages/mdc-grid-list/foundation.ts
+++ b/packages/mdc-grid-list/foundation.ts
@@ -44,6 +44,7 @@ class MDCGridListFoundation extends MDCFoundation<MDCGridListAdapter> {
   private readonly resizeHandler_: EventListener;
   private resizeFrame_ = 0;
 
+  /* istanbul ignore next: optional argument is not a branch statement */
   constructor(adapter?: Partial<MDCGridListAdapter>) {
     super({...MDCGridListFoundation.defaultAdapter, ...adapter});
 
--- a/packages/mdc-grid-list/foundation.ts
+++ b/packages/mdc-grid-list/foundation.ts
@@ -44,8 +44,9 @@ class MDCGridListFoundation extends MDCFoundation<MDCGridListAdapter> {
   private readonly resizeHandler_: EventListener;
   private resizeFrame_ = 0;
 
-  constructor(adapter: MDCGridListAdapter) {
-    super(Object.assign(MDCGridListFoundation.defaultAdapter, adapter));
+  constructor(adapter: Partial<MDCGridListAdapter> = {}) {
+    super({...MDCGridListFoundation.defaultAdapter, ...adapter});
+
     this.resizeHandler_ = this.alignCenter.bind(this);
   }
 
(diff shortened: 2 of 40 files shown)


kfranqueiro:
Using `ponyfill.matches` everywhere will end up fixing #4340, which someone submitted #4372 for against master originally (I suggested waiting until the TS conversion).

packages/mdc-grid-list/foundation.ts
```diff
@@ -44,8 +44,10 @@ class MDCGridListFoundation extends MDCFoundation<MDCGridListAdapter> {
   private readonly resizeHandler_: EventListener;
   private resizeFrame_ = 0;
 
-  constructor(adapter: MDCGridListAdapter) {
-    super(Object.assign(MDCGridListFoundation.defaultAdapter, adapter));
+  /* istanbul ignore next: optional argument is not a branch statement */
```
kfranqueiro: Is `ignore next` still necessary without a default assignment? I don't see it on the other constructors using an optional argument.

packages/mdc-radio/index.ts
```diff
@@ -81,18 +80,17 @@ class MDCRadio extends MDCComponent<MDCRadioFoundation> implements RippleCapable
...
+      deregisterInteractionHandler: <K extends EventType>(evtType: K, handler: SpecificEventListener<K>) =>
+          this.nativeControl_.removeEventListener(evtType, handler),
       // Radio buttons technically go "active" whenever there is *any* keyboard interaction. This is not the
       // UI we desire.
       isSurfaceActive: () => false,
       isUnbounded: () => true,
-      registerInteractionHandler: <K extends EventType>(type: K, handler: SpecificEventListener<K>) =>
-        this.nativeControl_.addEventListener(type, handler),
+      registerInteractionHandler: <K extends EventType>(evtType: K, handler: SpecificEventListener<K>) =>
+          this.nativeControl_.addEventListener(evtType, handler),
```
kfranqueiro: excess indentation?

packages/mdc-ripple/index.ts
```diff
@@ -38,41 +39,35 @@ class MDCRipple extends MDCComponent<MDCRippleFoundation> implements RippleCapab
...
-    const MATCHES = util.getMatchesProperty(HTMLElement.prototype);
-
     return {
       addClass: (className) => instance.root_.classList.add(className),
       browserSupportsCssVars: () => util.supportsCssVariables(window),
       computeBoundingRect: () => instance.root_.getBoundingClientRect(),
       containsEventTarget: (target) => instance.root_.contains(target as Node),
       deregisterDocumentInteractionHandler: (evtType, handler) =>
-        document.documentElement.removeEventListener(evtType, handler, util.applyPassive()),
+          document.documentElement.removeEventListener(evtType, handler, util.applyPassive()),
```
kfranqueiro: Excess indentation here and below?

test/unit/mdc-ripple/mdc-ripple.test.js
```diff
@@ -118,12 +119,17 @@ test('adapter#isUnbounded delegates to unbounded getter', () => {
...
-  root[MATCHES] = matches;
-  assert.isOk(component.getDefaultFoundation().adapter_.isSurfaceActive());
+  const {component} = setupTest();
+  const realMatches = ponyfill.matches;
+  const fakeMatches = td.func('.matches');
+  td.when(fakeMatches(td.matchers.isA(Element), ':active')).thenReturn(true);
+  ponyfill.matches = fakeMatches;
+
+  try {
+    assert.isOk(component.getDefaultFoundation().adapter_.isSurfaceActive());
```
kfranqueiro: `isTrue`

packages/mdc-base/component.ts
```diff
@@ -51,6 +51,7 @@ class MDCComponent<FoundationType extends MDCFoundation> {
     this.initialSyncWithDOM();
   }
 
+  /* istanbul ignore next: method param only exists for typing purposes; it does not need to be unit tested */
```
kfranqueiro: What is this comment skipping? There's no default assignment and I don't see any conditionals in the output code... shouldn't this end up hit during our unit tests anyway?

packages/mdc-grid-list/foundation.ts
acdvorak: I'm honestly not sure why it's needed in some places but not others.

I've noticed that it's _usually_ needed when a base foundation is subclassed by another foundation that doesn't have its own constructor (e.g., tab-scroller), but this case I have no idea.

My only guess is that it's a quirk in Istanbul.

packages/mdc-radio/index.ts
acdvorak: IntelliJ did this for me automatically, and it passed the linter, so I assumed the Google tslint settings preferred it.

However, I can't find any indentation or continuation settings in https://github.com/google/ts-style/blob/8f81766/tslint.json or https://github.com/palantir/tslint/blob/8a11d05/src/configs/recommended.ts, so I'll reduce these back to 2 spaces and update my IntelliJ format settings.

packages/mdc-base/component.ts
acdvorak: Welcome to Istanbul 😛 If you remove this comment, code coverage drops on certain components. Don't ask me to explain it, cuz I gave up trying to figure it out after 30 min of banging my head against a wall.

PR #4391, opened 2019-02-09
feat(tab-indicator): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-tab-indicator/index.ts
```diff
@@ -23,55 +23,35 @@
 
 import {MDCComponent} from '@material/base/component';
 
-import MDCTabIndicatorAdapter from './adapter';
-import MDCTabIndicatorFoundation from './foundation';
+import {MDCTabIndicatorAdapter} from './adapter';
```
moog16: Don't forget to export this too

packages/mdc-tab-indicator/index.ts
```diff
@@ -81,10 +61,7 @@ class MDCTabIndicator extends MDCComponent {
     return new MDCSlidingTabIndicatorFoundation(adapter);
   }
 
-  /**
-   * @param {!ClientRect=} previousIndicatorClientRect
-   */
-  activate(previousIndicatorClientRect) {
+  activate(previousIndicatorClientRect?: ClientRect | null) {
```
moog16: Does this have to be both `null` or `undefined` as denoted by the `?`? Can it just be `previousIndicatorClientRect?: ClientRect`?

packages/mdc-tab-indicator/index.ts
```diff
@@ -93,4 +70,9 @@ class MDCTabIndicator extends MDCComponent {
   }
 }
 
-export {MDCTabIndicator, MDCTabIndicatorFoundation, MDCSlidingTabIndicatorFoundation, MDCFadingTabIndicatorFoundation};
+export {MDCTabIndicator as default, MDCTabIndicator};
+export * from './adapter';
+export * from './constants';
```
kfranqueiro: We haven't been exporting constants from index - probably because they're already exposed off of the foundation, usually?

test/unit/mdc-tab-indicator/sliding-foundation.test.js
```diff
@@ -56,7 +56,9 @@ test('#activate does not modify transform and does not transition if no client r
 });
 
 test(`#deactivate removes the ${MDCSlidingTabIndicatorFoundation.cssClasses.ACTIVE} class`, () => {
-  const {foundation, mockAdapter} = setupTest();
+  const foundation = new MDCSlidingTabIndicatorFoundation();
```
kfranqueiro: Why was this test changed to use a real adapter with one mocked method?

acdvorak: Code coverage

packages/mdc-tab-indicator/index.ts
acdvorak: Woops, removed

PR #4392, opened 2019-02-09
feat(tab-scroller): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-tab-scroller/foundation.ts
```diff
@@ -124,12 +111,12 @@ class MDCTabScrollerFoundation extends MDCFoundation {
 
   /**
    * Handles the transitionend event
-   * @param {!Event} evt
    */
-  handleTransitionEnd(evt) {
+  handleTransitionEnd(evt: Event) {
```
moog16: This type should be TransitionEvent.

packages/mdc-tab-scroller/util.ts
```diff
@@ -54,14 +50,20 @@ function computeHorizontalScrollbarHeight(documentObj, shouldCacheResult = true)
...
-function getMatchesProperty(HTMLElementPrototype) {
-  return [
-    'msMatchesSelector', 'matches',
-  ].filter((p) => p in HTMLElementPrototype).pop();
-}
+export type VendorMatchesFunctionName = 'webkitMatchesSelector' | 'msMatchesSelector';
+export type MatchesFunctionName = VendorMatchesFunctionName | 'matches';
 
-export {computeHorizontalScrollbarHeight, getMatchesProperty};
+export function getMatchesProperty(htmlElementPrototype: {}): MatchesFunctionName {
```
moog16: shouldn't we be using the `@material/dom/matches` function here instead?

moog16: probably can be a different PR actuallly... nvm

packages/mdc-tab-scroller/adapter.ts
```diff
@@ -0,0 +1,114 @@
...
+   */
+  computeScrollContentClientRect(): ClientRect;
+
+  /**
+   * Returns the height of the browser's horizontal scrollbars (in px).
+   */
+  computeHorizontalScrollbarHeight(): number;
+}
+
+export {MDCTabScrollerAdapter as default, MDCTabScrollerAdapter};
```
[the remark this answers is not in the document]
acdvorak: No, this is migrated from `adapter.js`, but there were enough changes that it confused git into thinking it's a new file instead of a rename.

packages/mdc-tab-scroller/util.ts
acdvorak: I actually already updated `index.ts` to use `ponyfill.matches()`, but I'm leaving `util.getMatchesProperty()` here to avoid breaking backward compatibility for any clients who might be using it. We'll remove `getMatchesProperty()` from the various packages that provide it (including ripple) after all the TS stuff is merged, with a BREAKING CHANGE commit.

packages/mdc-tab-scroller/foundation.ts
acdvorak: Done

packages/mdc-tab-scroller/adapter.ts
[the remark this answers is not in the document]
acdvorak: Done

PR #4393, opened 2019-02-09
feat(tab): Convert JS to TypeScript

acdvorak:
Refs #4225

PR #4394, opened 2019-02-09
feat(tab-bar): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-tab-bar/types.ts
```diff
@@ -0,0 +1,25 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+export type TabFactory = (el: Element) => MDCTab;
```
acdvorak: Depends on PR #4391, PR #4392 and PR #4393

packages/mdc-tab-bar/index.ts
```diff
@@ -0,0 +1,175 @@
...
+      getScrollPosition: () => this.tabScroller_!.getScrollPosition(),
+      getScrollContentWidth: () => this.tabScroller_!.getScrollContentWidth(),
+      getOffsetWidth: () => (this.root_ as HTMLElement).offsetWidth,
+      isRTL: () => window.getComputedStyle(this.root_).getPropertyValue('direction') === 'rtl',
+      setActiveTab: (index) => this.foundation_.activateTab(index),
+      activateTabAtIndex: (index, clientRect) => this.tabList_[index].activate(clientRect),
+      deactivateTabAtIndex: (index) => this.tabList_[index].deactivate(),
+      focusTabAtIndex: (index) => this.tabList_[index].focus(),
+      getTabIndicatorClientRectAtIndex: (index) => this.tabList_[index].computeIndicatorClientRect(),
+      getTabDimensionsAtIndex: (index) => this.tabList_[index].computeDimensions(),
```
moog16: is it possible that we have `!tabList_.length`? Should this adapter method also return undefined?

acdvorak: Good question. It looks like `getTabDimensionsAtIndex()` only gets called when there's at least one tab, so I don't think this should be a problem.

packages/mdc-tab-bar/index.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  destroy() {
+    super.destroy();
+    this.unlisten(MDCTabFoundation.strings.INTERACTED_EVENT, this.handleTabInteraction_);
+    this.unlisten('keydown', this.handleKeyDown_);
+    this.tabList_.forEach((tab) => tab.destroy());
+
+    if (this.tabScroller_) {
```
kfranqueiro: Is this `if` necessary? We defined it with `!` above.

PR #4395, opened 2019-02-09
feat(auto-init): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-auto-init/index.ts
```diff
@@ -41,23 +43,25 @@ function _emit(evtType, evtData, shouldBubble = false) {
...
     }
 
-    const Ctor = registry[ctorName];
+    const Ctor = registry[ctorName]; // tslint:disable-line:variable-name
     if (typeof Ctor !== 'function') {
       throw new Error(
         `(mdc-auto-init) Could not find constructor in registry for ${ctorName}`);
     }
 
+    // @ts-ignore
```
moog16: if we have a ts-ignore, we should also have the reason why we need to ignore it - please comment.

packages/mdc-auto-init/index.ts
```diff
@@ -78,7 +82,8 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
   return components;
 }
 
-mdcAutoInit.register = function(componentName, Ctor, warn = CONSOLE_WARN) {
+// tslint:disable-next-line:variable-name
+mdcAutoInit.register = function(componentName: string, Ctor: Function, warn = CONSOLE_WARN) {
```
moog16: `Function` --> `() => any`

kfranqueiro: Wouldn't something like `new (...args: any[]) => any` be more appropriate for a constructor? (I don't suppose there's a way to avoid using `any` here...)

packages/mdc-auto-init/package.json
```diff
@@ -7,5 +7,8 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/material-components/material-components-web.git"
+  },
+  "dependencies": {
+    "@material/base": "^0.41.0"
```
kfranqueiro: I'm probably missing something obvious but what does this need base for now?

packages/mdc-auto-init/index.ts
moog16: actually we should be using `unknown` according to our best practices

acdvorak:
```ts
interface ComponentClass {
  // tslint:disable-next-line:no-any a component can pass in anything it needs to the constructor
  new<F extends MDCFoundation>(root: Element, foundation?: F, ...args: any[]): MDCComponent<F>;
  attachTo<F extends MDCFoundation>(root: Element): MDCComponent<F>;
}

mdcAutoInit.register = function(componentName: MDCAutoInitKey, Ctor: ComponentClass, warn = CONSOLE_WARN) {
  // ...
}
```

packages/mdc-auto-init/package.json
acdvorak: Added base imports to index.ts 😄

packages/mdc-auto-init/index.ts
```diff
@@ -78,7 +97,8 @@ export default function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
   return components;
 }
 
-mdcAutoInit.register = function(componentName, Ctor, warn = CONSOLE_WARN) {
+// tslint:disable-next-line:variable-name
```
kfranqueiro: Which name requires this comment?

packages/mdc-auto-init/types.ts
```diff
@@ -0,0 +1,55 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+export type MDCAutoInitKey = (
```
kfranqueiro: Does this make autoinit completely inextensible to third-parties using TS? i.e. will this cause `register` with anything besides these values to fail to compile?

(Not to mention this adds one more thing we need to remember to populate any time we add a new component...)

packages/mdc-auto-init/index.ts
```diff
@@ -21,16 +21,32 @@
  * THE SOFTWARE.
  */
 
-const registry = Object.create(null);
+// tslint:disable:only-arrow-functions
 
-const CONSOLE_WARN = console.warn.bind(console);
+import {MDCComponent} from '@material/base/component';
```
moog16: can we condense this to 
```
import {MDCComponent, MDCFoundation} from '@material/base/index;
```

packages/mdc-auto-init/index.ts
```diff
@@ -21,16 +21,32 @@
...
+
+type Registry = {
+  [K in MDCAutoInitKey]?: ComponentClass;
+};
+
+const registry: Registry = {};
+
+const CONSOLE_WARN = console.warn.bind(console); // tslint:disable-line:no-console
+
+function _emit<T extends object>(evtType: string, evtData: T, shouldBubble = false) {
```
moog16: did we change shouldBubble --> bubble in another PR?

packages/mdc-auto-init/index.ts
acdvorak: `Ctor`. I renamed it to `Constructor` and added a comment explaining why we're disabling this tslint rule.

packages/mdc-auto-init/types.ts
acdvorak: `tsc` will yell at us if we forget to populate it when we add a new component.

But you're right that this would make it inextensible to third parties.

Should I remove `MDCAutoInitKey` and change the param to accept a `string` instead?

packages/mdc-auto-init/index.ts
acdvorak: Done

packages/mdc-auto-init/index.ts
acdvorak: Yeah, but Ken wasn't fond of the change, so I reverted it. Good eye tho! 👀

packages/mdc-auto-init/types.ts
kfranqueiro: I think we should do that, yes. I'm pretty sure I've heard of clients registering additional components with autoinit.

packages/mdc-auto-init/index.ts
kfranqueiro: We could rename shouldBubble across the board later in a separate PR if we want, the problem was it was renamed in base in a PR not otherwise involving base, without also updating comments across the repo that mention it as `shouldBubble`.

packages/mdc-auto-init/types.ts
acdvorak: Done

packages/mdc-auto-init/index.ts
```diff
@@ -40,36 +54,39 @@ function _emit(evtType, evtData, shouldBubble = false) {
...
 
-    if (node[ctorName]) {
+    if (Object.getOwnPropertyDescriptor(node, ctorName)) {
       warn(`(mdc-auto-init) Component already initialized for ${node}. Skipping...`);
       continue;
     }
 
     // TODO: Should we make an eslint rule for an attachTo() static method?
-    const component = Ctor.attachTo(node);
+    // See https://github.com/Microsoft/TypeScript/issues/14600 for discussion of static interface support in TS
```
kfranqueiro: Is the TODO above this obsolete? (Like, would our build actually fail without an `attachTo` given the typings above? Plus we're using tslint now to reuse the Google configuration.)

acdvorak: Unfortunately TypeScript won't catch that because it doesn't do static type checking.
See https://github.com/Microsoft/TypeScript/issues/14600.

`attachTo()` in the `Component` interface is only there to make TS happy when we try to call `Constructor.attachTo()` in `mdcAutoInit()`. It doesn't verify that `static attachTo()` exists on classes passed to `register()`.

It _does_, however, verify that the required constructor signature exists (via `new<F...>()`) on classes passed to `register()`.

PR #4396, opened 2019-02-09
feat(package): Convert JS to TypeScript

acdvorak:
Refs #4225

Removes deprecated package `mdc-icon-toggle` from the all-in-one JS bundle.


kfranqueiro:
This probably won't pass CI until we convert everything else.

moog16:
Why don't you think so?

moog16:
I think the `createMainJsCombined` in js-bundle-factory.js needs to be converted.

acdvorak:
Updated all `.js` paths to `.ts` in `js-bundle-factory.js` (except for `icon-toggle` and `toolbar`, which haven't been converted to TS)

moog16:
Oh ya - Ken might be right. It might error on all of the packages that don't yet have types that we depend on currently. Other than that I approve

packages/material-components-web/index.ts
```diff
@@ -107,13 +102,12 @@ export {
...
   slider,
-  switchControl,
   snackbar,
+  switchControl,
   tab,
   tabBar,
   tabIndicator,
   tabScroller,
   textField,
-  toolbar,
```
moog16: looks like you opened a PR for toolbar...i think you mean to revert this.

acdvorak: Done

moog16:
I think there is a `check-pkg-for-release.js` script that is also missing. Run `node scripts/check-pkg-for-release.js` to verify

PR #4404, opened 2019-02-11
 chore: rename main mixins to end with `-core-styles`

mmalerba:
Depends on #4383


mmalerba:
Thanks, prettier still makes some weird choices in some cases, must've forgot to revert them. I should probably just turn it off

PR #4407, opened 2019-02-12
refactor: Make TS easier to wrap

acdvorak:
### Goal

Make it easier for frameworks to wrap our TypeScript code.

Refs #4225

### How

**Isolate framework-specific code from framework-agnostic code.**

Specifically, this PR:

* Maintains full backward compatibility with existing code
* Moves all component definitions from `index.ts` to `component.ts`
    - `index.ts` is now purely re-exporting other files
* Moves component-dependent types from `types.ts` to `component.ts`
    - Framework-related types are now quarantined in a single file: `component.ts`
    - All other files are now purely framework-agnostic, and can be safely wrapped by frameworks that don't use our components
* Makes `import` paths more specific by switching from `@material/foo/index` to e.g. `@material/foo/foundation` or `@material/foo/types`
    - The only exception is `component.ts` files: Since they're basically our default "framework", they can safely import _other_ "framework" types via `@material/foo/index`
* Updates class & interface export syntax to isolate the `default` export line for future removal if necessary:
    ```ts
    export class MDCFooFoundation {
      // ...
    }
    export default MDCFooFoundation;
    ```
* Combines `typings/custom.d.ts` and `typings/dom.ie.d.ts` into `packages/mdc-dom/externs.d.ts` so the types will be publicly visible
* Adds `MDCFooFactory` types for components that need them (e.g., `MDCRipple`, `MDCList`)

### Packages

* [x] `animation`
* [x] `checkbox`
* [x] `chips`
* [x] `dialog`
* [x] `dom`
* [x] `drawer`
* [x] `floating-label`
* [x] `form-field`
* [x] `grid-list`
* [x] `icon-button`
* [x] `icon-toggle` _(deprecated)_
* [x] `line-ripple`
* [x] `linear-progress`
* [x] `list`
* [x] `menu`
* [x] `menu-surface`
* [x] `notched-outline`
* [x] `radio`
* [x] `ripple`
* [x] `select`
* [x] `selection-control`
* [x] `slider`
* [x] `snackbar`
* [x] `switch`
* [x] `tab`
* [x] `tab-bar`
* [x] `tab-indicator`
* [x] `tab-scroller`
* [x] `tab-tabs` _(deprecated)_
* [x] `textfield`
* [x] `toolbar` _(deprecated)_
* [x] `top-app-bar`


packages/mdc-chips/chip-set/component.ts
```diff
@@ -0,0 +1,142 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+import {MDCComponent} from '@material/base/component';
+import {MDCChip, MDCChipFactory, MDCChipFoundation} from '../chip';
```
kfranqueiro: I instinctively wanted to suggest `chip/index` here... but I guess this is safe because the subdirectory shouldn't be subject to package.json `main` resolution?

packages/mdc-chips/chip-set/component.ts
```diff
@@ -0,0 +1,142 @@
...
+import {MDCChipInteractionEvent, MDCChipRemovalEvent, MDCChipSelectionEvent} from '../chip';
+import {MDCChipSetFoundation} from './foundation';
+
+let idCounter = 0;
+
+export type MDCChipSeFactory = (root: Element, foundation?: MDCChipSetFoundation) => MDCChipSet;
```
kfranqueiro: Typo in name?

packages/mdc-dialog/component.ts
```diff
@@ -0,0 +1,186 @@
...
+  private focusTrap_!: FocusTrapLib.FocusTrap; // assigned in initialSyncWithDOM()
+
+  private handleInteraction_!: SpecificEventListener<'click'|'keydown'>; // assigned in initialSyncWithDOM()
+  private handleDocumentKeydown_!: SpecificEventListener<'keydown'>; // assigned in initialSyncWithDOM()
+  private handleLayout_!: EventListener; // assigned in initialSyncWithDOM()
+  private handleOpening_!: EventListener; // assigned in initialSyncWithDOM()
+  private handleClosing_!: () => void; // assigned in initialSyncWithDOM()
+
+  initialize(
+      focusTrapFactory?: FocusTrapFactory,
```
kfranqueiro: The default assignment was taken off of this argument. Is it never relied upon?

packages/mdc-dialog/index.ts
```diff
@@ -21,167 +21,11 @@
...
-  }
-}
-
-export {MDCDialog as default, MDCDialog, util};
+export {MDCDialog as default} from './component';
 export * from './adapter';
+export * from './component';
 export * from './foundation';
 export * from './types';
+export {util};
```
kfranqueiro: Should this be moved up? Seems weird to have `export *`s sandwiched between two more specific exports. Or is the idea here that the non-`from` export is last?

packages/mdc-dom/externs.d.ts
```diff
@@ -24,3 +24,7 @@
 declare interface Element {
   msMatchesSelector?: (selector: string) => boolean;
 }
+
+declare interface Window {
```
kfranqueiro: Does moving this from `typings` require more things to rely on `dom`? Are we sure we have the dependency everywhere it's needed? Does this now need to be included explicitly?

packages/mdc-chips/chip-set/component.ts
acdvorak: Correct, it should be safe. But we can keep `/index` for consistency if you prefer. WDYT?

packages/mdc-chips/chip-set/component.ts
acdvorak: Good catch!

packages/mdc-dialog/component.ts
acdvorak: Not directly, no. `util.createFocusTrapInstance()` specifies the exact same default value, so it seemed pointless to duplicate that here.

packages/mdc-dom/externs.d.ts
acdvorak: Good question. In our repo, TypeScript doesn't seem to care where this file is. Some wrapper frameworks might run into problems though, and would need to add a dep on `mdc-dom`. Would you rather put this in `mdc-base`? (I actually considered doing that, but `mdc-dom` seemed more semantically correct. But maybe convenience is more important than semantics.)

@moog16 @mmalerba Is it OK that these types have the same names as existing types in `lib.dom.d.ts`? Does TypeScript somehow merge the two type definitions together or something? Or should we create a new interface with a different name that `extends Window`?

packages/mdc-chips/chip-set/component.ts
kfranqueiro: I'm slightly inclined towards consistency so it doesn't trip up any one else reading it, but I dunno if that's just me.

packages/mdc-checkbox/component.ts
```diff
@@ -0,0 +1,185 @@
...
+  set value(value: string) {
+    this.nativeCb_.value = value;
+  }
+}
+
+function validDescriptor(inputPropDesc: PropertyDescriptor | undefined): inputPropDesc is PropertyDescriptor {
+  return !!inputPropDesc && typeof inputPropDesc.set === 'function';
+}
+
+export default MDCCheckbox;
```
acdvorak: Remove all `default` exports from componentfiles. The existing JS code in `master` only has `default` exports for foundations.

packages/mdc-chips/chip-set/component.ts
moog16: I think if we're questioning it, we should lean towards being more explicit. Its not going to hurt to add the `/index`.

packages/mdc-floating-label/foundation.ts
```diff
@@ -22,11 +22,11 @@
  */
 
 import {MDCFoundation} from '@material/base/foundation';
-import {SpecificEventListener} from '@material/base/index';
+import {SpecificEventListener} from '@material/base/types';
```
moog16: if we export the types from `base/index.ts` we don't need to change this.

packages/mdc-dom/externs.d.ts
mmalerba: Exactly, TS merges interfaces: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

packages/mdc-chips/chip/types.ts
```diff
@@ -1,42 +0,0 @@
-/**
```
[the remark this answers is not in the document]
acdvorak: `types.ts` is for definitions that are used by the adapter and/or foundation.

Types that refer to the component class, like `type MDCFooFactory = (el) => MDCFoo`, stay in `component.ts`.

This ensures that we can safely sync all non-component files without getting errors about missing references to `MDCFoo`.

Does that make sense? Would a different pattern be better?

packages/mdc-floating-label/foundation.ts
acdvorak: Foundation and adapter files need to use more-specific import paths instead of `/index`, because we might not end up mirroring all `index.ts` files, which would cause broken references when syncing.

However, `index.ts` files _can_ import from other `index.ts` files, since we don't sync them.

packages/mdc-animation/index.ts
```diff
@@ -111,12 +90,3 @@ function getCorrectEventName(windowObj: Window, eventType: StandardJsEventType):
   }
   return eventType;
 }
-
-export {
```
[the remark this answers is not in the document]
acdvorak: Done

packages/mdc-dom/externs.d.ts
acdvorak: Done. Moved from `mdc-dom` to `mdc-base`.

packages/mdc-chips/chip-set/component.ts
acdvorak: Done

packages/mdc-checkbox/component.ts
acdvorak: Yes, none of our existing JS components have default exports, so we shouldn't add them.

acdvorak:
@moog16:

> Are you going to do the rest of the components after we iron out the details here? Or will you open a new PR after this?

I talked to Ken offline and we decided to do everything in one _giant_ PR, because it'll be easier to ensure consistency across all packages. It's the lesser of two evils 🙈🙊

packages/mdc-chips/chip/types.ts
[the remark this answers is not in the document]
moog16: ya that makes sense...I wasn't sure what the pattern was. Can we add a segment in the best_practices.md? It wasn't totally obvious just from looking at the code.

packages/mdc-checkbox/component.ts
```diff
@@ -0,0 +1,183 @@
...
+  destroy() {
+    this.ripple_.destroy();
+    this.nativeCb_.removeEventListener('change', this.handleChange_);
+    this.unlisten(getCorrectEventName(window, 'animationend'), this.handleAnimationEnd_);
+    this.uninstallPropertyChangeHooks_();
+    super.destroy();
+  }
+
+  getDefaultFoundation(): MDCCheckboxFoundation {
+    return new MDCCheckboxFoundation({
```
acdvorak: Extract variable for all adapter implementations.

The foundation argument is a `Partial<MDCFooAdapter>`, so `tsc` won't complain if we omit a method; it'll just merge it with the default implementation, which does nothing.

Extracting a variable and explicitly declaring it of type `MDCFooAdapter` ensures that we'll get a compiler error if we forget a field.

packages/mdc-menu/index.ts
```diff
@@ -21,192 +21,8 @@
...
-        index: evtData.index,
-        item: this.items[evtData.index],
-      }),
-    });
-    // tslint:enable:object-literal-sort-keys
-  }
-}
-
-export {MDCMenu as default, MDCMenu, Corner};
+export {Corner} from '@material/menu-surface/constants'; // for backward compatibility
```
kfranqueiro: I guess AnchorMargin no longer exists, but again it's just typings so probably not critical to lose it here.

packages/mdc-drawer/component.ts
```diff
@@ -0,0 +1,156 @@
...
+    if (this.scrim_ && this.handleScrimClick_ && this.root_.classList.contains(MODAL)) {
+      this.scrim_.removeEventListener('click', this.handleScrimClick_);
+      // Ensure drawer is closed to hide scrim and release focus
+      this.open = false;
+    }
+  }
+
+  getDefaultFoundation() {
+    // tslint:disable:object-literal-sort-keys
+    const adapter: MDCDrawerAdapter = {
```
kfranqueiro: This looks right (has separate typed variable) but is missing the "do not inline" comment seen elsewhere.

Also in this boat: tab-indicator, tab-scroller, tabs/tab-bar, tabs/tab-bar-scroller, toolbar, top-app-bar

packages/mdc-ripple/types.ts
```diff
@@ -21,15 +21,15 @@
  * THE SOFTWARE.
  */
 
-export interface Point {
+export interface MDCRipplePoint {
```
kfranqueiro: Can you explain why you wanted to leave RippleCapableSurface alone "for backcompat", given that it's a type and meanwhile there are other types that no longer exist (some of which I've pointed out)?

.gitignore
```diff
@@ -7,7 +7,6 @@ out
 *.log
 .idea
 *.sw*
-.closure-tmp
+.rewrite-tmp
```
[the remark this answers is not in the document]
acdvorak: Correct, it's only used by the now-internal sync scripts. I added a comment explaining what it's for.

packages/mdc-animation/index.ts
```diff
@@ -21,102 +21,8 @@
...
-  PrefixedCssPropertyName,
-  StandardCssPropertyName,
-  PrefixedJsEventType,
-  StandardJsEventType,
-  getCorrectEventName,
-  getCorrectPropertyName,
-};
+export {util}; // New namespace
+export * from './types';
+export * from './util'; // Old namespace for backward compatibility
```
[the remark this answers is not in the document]
acdvorak: Oh yeah, I guess that's technically a breaking change...

Should I mirror this with a PR in master to ensure it ends up in the changelog?

packages/mdc-tab-scroller/component.ts
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Scrolls to the given pixel position
+   * @param scrollX The pixel value to scroll to
+   */
+  scrollTo(scrollX: number) {
+    this.foundation_.scrollTo(scrollX);
+  }
+}
+
+export default MDCTabScroller; // For backward compatibility
```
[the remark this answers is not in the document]
acdvorak: Good catch. It looks like only `MDCTabScrollerRTL` and its subclasses have a default export, so I removed this one.

https://github.com/material-components/material-components-web/blob/55e4229071be44decbfcdf2079c1fddffe391fe2/packages/mdc-tab-scroller/index.js#L147

https://github.com/material-components/material-components-web/blob/f310e7364f6f7a7ae162908ca27808f0e4619184/packages/mdc-tab-scroller/rtl-scroller.js#L70

https://github.com/material-components/material-components-web/blob/f310e7364f6f7a7ae162908ca27808f0e4619184/packages/mdc-tab-scroller/rtl-default-scroller.js#L104

packages/mdc-chips/chip/foundation.ts
```diff
@@ -185,4 +185,4 @@ class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
   }
 }
 
-export {MDCChipFoundation as default, MDCChipFoundation};
+export default MDCChipFoundation;
```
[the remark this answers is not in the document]
acdvorak: Yeah, I don't think we need to worry about it. We don't use any of these types internally, so it shouldn't break anybody.

packages/mdc-menu/index.ts
acdvorak: Yep, this shouldn't break anybody. (`AnchorMargin` was renamed to `MDCMenuDistance`.)

packages/mdc-drawer/component.ts
acdvorak: Good catch! Not sure how I missed this. Fixed.

packages/mdc-ripple/types.ts
acdvorak: I assumed it was more likely to be used by Closure clients, but it looks like I was wrong. Renamed to `MDCRippleCapableSurface`.

packages/mdc-animation/index.ts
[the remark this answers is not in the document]
kfranqueiro: Is it possible and reasonable to do that in master? (I guess it comes down to being able to make a similar change to linear-progress in master.)

packages/mdc-tab-bar/index.ts
```diff
@@ -21,155 +21,6 @@
...
-    if (tabScrollerElement) {
-      return tabScrollerFactory(tabScrollerElement);
-    }
-    return null;
-  }
-}
-
-export {MDCTabBar as default, MDCTabBar};
 export * from './adapter';
+export * from './component';
```
kfranqueiro: Missing `export * from './types';` ?

packages/mdc-linear-progress/foundation.ts
```diff
@@ -21,12 +21,12 @@
...
  */
 
-import {getCorrectPropertyName} from '@material/animation/index';
+import {getCorrectPropertyName} from '@material/animation/util';
 import {MDCFoundation} from '@material/base/foundation';
 import {MDCLinearProgressAdapter} from './adapter';
 import {cssClasses, strings} from './constants';
 
-class MDCLinearProgressFoundation extends MDCFoundation<MDCLinearProgressAdapter> {
+export class MDCLinearProgressFoundation extends MDCFoundation<MDCLinearProgressAdapter> {
```
kfranqueiro: Apparently not in this PR, but somehow this foundation lost the constructor which mixes the default adapter with the passed adapter.

packages/mdc-radio/foundation.ts
```diff
@@ -25,7 +25,7 @@ import {MDCFoundation} from '@material/base/foundation';
 import {MDCRadioAdapter} from './adapter';
 import {cssClasses, strings} from './constants';
 
-class MDCRadioFoundation extends MDCFoundation<MDCRadioAdapter> {
+export class MDCRadioFoundation extends MDCFoundation<MDCRadioAdapter> {
```
kfranqueiro: FYI this class is missing a constructor that calls super while mixing the default adapter plus the passed adapter. It appears to have always lacked it, even in JS.

acdvorak: Yikes on bikes! Added.

I did a quick grep of all packages and I don't think we're missing any other constructors.

However, I noticed that `MDCTabScrollerRTL` (which is used by `MDCTabScrollerFoundation`) does not extend `MDCFoundation`, nor does it accept partial adapters in the constructor:

https://github.com/material-components/material-components-web/blob/56a065035c0f3f9e3f4a3fff3f3b445f9f1ffd3b/packages/mdc-tab-scroller/rtl-scroller.ts#L27-L32

It's already like that in `master`:

https://github.com/material-components/material-components-web/blob/70672f12ff19591c53fef6876c059c854f9cb816/packages/mdc-tab-scroller/rtl-scroller.js#L33-L38

Is that OK?

packages/mdc-linear-progress/foundation.ts
acdvorak: Good catch! Added.

packages/mdc-radio/foundation.ts
kfranqueiro: That's not a foundation class, so N/A.

There are also a few foundation classes which extend other (non-base) foundation classes which already deal with the constructor and all share a common adapter, so it's N/A in those cases 2. The 2 cases I pointed out are the only 2 that seem problematic.

PR #4409, opened 2019-02-13
feat(infrastructure): add esmodules support for treeshaking

moog16:
related #4225


packages/mdc-animation/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.js
+!*.ts
+# must come after !*.ts
+*.d.ts
+*.js.map
```
acdvorak: Can any of these rules go in the root `.gitignore` / `.npmignore` file? (Or `packages/.gitignore`)

We can always override them for specific packages if necessary (e.g., `mdc-base` or `mdc-dom` have a `.d.ts` file, or packages that haven't been converted to TypeScript yet).

packages/mdc-animation/package.json
```diff
@@ -9,6 +9,9 @@
     "animation"
   ],
   "main": "dist/mdc.animation.js",
+  "module": "./index.js",
```
acdvorak: Nit: Is the `./` necessary?

scripts/cp-pkgs.js
```diff
@@ -77,9 +90,24 @@ function cpAsset(asset) {
   return cpFile(asset, destDir).then(() => console.log(`cp ${asset} -> ${destDir}`));
 }
 
+function cpDeclarationAsset(asset) {
+  const assetPkg = path.parse(asset.split('build/')[1]).dir;
+  if (!fs.existsSync(assetPkg)) {
+    Promise.reject(new Error(`Non-existent asset package path ${assetPkg} for ${asset}`));
+  }
+  const destFileName = toCamelCase(path.parse(asset).name.replace(/^mdc-|\.d$/gm, ''));
```
acdvorak: Why is the `/m` ([multiline](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Syntax)) switch needed on this RegExp?

scripts/webpack/js-bundle-factory.js
```diff
@@ -91,6 +91,7 @@ class JsBundleFactory {
       },
       devtool: 'source-map',
       module: {
+        noParse: ['/*.ts/'],
```
acdvorak: A few questions:

1. Why is this needed? It sounds like `noParse` is normally used to prevent Webpack from parsing large external third party libraries (https://webpack.js.org/configuration/module/#modulenoparse)
2. Is this going to break/disable source maps?
3. Doesn't this effectively negate `rules.test = /\.ts$/` below?
4. Why are there string quotes around what appears to be a RegExp literal?

tsconfig.json
```diff
@@ -21,6 +21,6 @@
     "target": "es5",
     "newLine": "lf",
     "types": [],
-    "declaration": true
+    "declaration": true,
```
acdvorak: Unfortunately this is invalid JSON syntax and will break many JSON parsers (including Node's default `JSON.parse()` implementation IIRC)

scripts/cp-pkgs.js
```diff
@@ -77,9 +90,24 @@ function cpAsset(asset) {
...
+  const destDir = path.join(assetPkg, 'dist', `mdc.${destFileName}.d.ts`);
+  return cpFile(asset, destDir).then(() => console.log(`cp ${asset} -> ${destDir}`));
```
acdvorak: We should probably handle the exception case here too, and at least print out the error

packages/mdc-animation/.gitignore
kfranqueiro: For `.gitignore` we should be able to make it common, e.g. `packages/**/*.js`. I am inclined to think this will _not_ work for `.npmignore` though.

packages/mdc-animation/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.js
+!*.ts
```
kfranqueiro: I'm confused, why does this even need to be here? Shouldn't we just need to ignore *.d.ts and be done?

For that matter, why does this `gitignore` include things only present under `dist`? The root `gitignore` already ignores `packages/*/dist`, which makes me wonder if this `gitignore` is completely overriding that, in which case to Andy's point we should definitely be working with the existing root `gitignore` instead.

packages/mdc-animation/.gitignore
moog16: moved to global gitignore

scripts/cp-pkgs.js
moog16: removed

packages/mdc-animation/.gitignore
moog16: This is now
```
packages/**/*.d.ts
```

scripts/webpack/js-bundle-factory.js
moog16: I think I had that from a little test from before. Removing it. SideNote: it doesn't seem to have any effect as it still parses the ts files...

.gitignore
```diff
@@ -3,6 +3,9 @@ node_modules
 /build
 /coverage
 packages/*/dist
+packages/**/*.js
```
acdvorak: Do we really need this line? Won't it be covered by the `*/dist` rule above?

.gitignore
```diff
@@ -3,6 +3,9 @@ node_modules
...
+packages/**/*.d.ts
```
acdvorak: Do we really need this line? Won't it be covered by the `*/dist` rule above?

package.json
```diff
@@ -4,20 +4,21 @@
   "license": "MIT",
   "scripts": {
     "build": "npm run clean && mkdirp build && webpack --progress --colors",
+    "build:esmodules": "node ./scripts/build/esmodule-builder.js",
     "build:demos": "npm run clean && mkdirp build && webpack --config=demos/webpack.config.js --progress --colors",
     "build:min": "mkdirp build && cross-env MDC_ENV=production webpack -p --progress --colors",
     "changelog": "standard-changelog -i CHANGELOG.md -k packages/material-components-web/package.json",
-    "clean": "del-cli build/** build .typescript-tmp/** .typescript-tmp",
+    "clean": "del-cli build/** build packages/**/*.js .typescript-tmp/** .typescript-tmp",
```
acdvorak: Is `packages/**/*.js` going to nuke stuff inside `node_modules`?

.gitignore
moog16: the build:esmodules task actually builds `.js` files in the root of each package/component.

.gitignore
moog16: .d.ts files also are built into the package directories

packages/mdc-base/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.js
```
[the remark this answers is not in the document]
moog16: actually this entire fist shouldn't be here :|

.travis.yml
```diff
@@ -39,9 +39,9 @@ matrix:
 
   - node_js: 8
     env:
-      - TEST_SUITE=build
+      - TEST_SUITE=dist
```
kfranqueiro: I would suggest leaving this name as-is. Otherwise you also need to update travis-env-vars.sh.

packages/material-components-web/package.json
```diff
@@ -8,6 +8,9 @@
     "material design"
   ],
   "main": "dist/material-components-web.js",
+  "module": "index.js",
+  "sideEffects": false,
+  "types": "dist/mdc.materialComponentsWeb.d.ts",
```
kfranqueiro: it seems odd that this filename is inconsistent with `main`?

packages/mdc-auto-init/package.json
```diff
@@ -3,6 +3,9 @@
   "description": "Declarative, easy-to-use auto-initialization for Material Components for the web",
   "version": "0.41.0",
   "main": "dist/mdc.autoInit.js",
+  "types": "dist/mdc.autoInit.d.ts",
```
kfranqueiro: Make the order of these 3 added properties consistent across files (they seem shuffled in a few of them, in most they were added as modules, sideEffects, then types)

packages/mdc-dialog/package-lock.json
```diff
@@ -1,25 +1,25 @@
...
-      "resolved": "https://registry.npmjs.org/tabbable/-/tabbable-3.1.2.tgz",
-      "integrity": "sha512-wjB6puVXTYO0BSFtCmWQubA/KIn7Xvajw0x0l6eJUudMG/EAiJvIUnyNX6xO4NpGrJ16lbD0eUseB9WxW0vlpQ=="
-    },
-    "xtend": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/xtend/-/xtend-4.0.1.tgz",
-      "integrity": "sha1-pcbVMr5lbiPbgg77lDofBJmNY68="
-    }
-  }
+	"requires": true,
```
kfranqueiro: Remove this file from the PR, it looks like all that changed was indentation for some reason (and it changed to hard tabs, which I didn't think npm ever uses??)

packages/mdc-grid-list/package.json
```diff
@@ -4,6 +4,9 @@
   "description": "The Material Components for the web grid list component",
   "license": "MIT",
   "main": "dist/mdc.gridList.js",
+  "module": "index.js",
+  "sideEffects": false,
+  "types": "dist/gridList.d.ts",
```
kfranqueiro: From this point on in the PR it seems like the `types` entries are missing the `mdc.` prefix to be consistent with `main`

.gitignore
acdvorak: Ahh ok. Would it be possible to generate the `.js` files in `dist/` or some other subdirectory so that we can more easily isolate them?

.gitignore
acdvorak: Same comment as above: Can we build `.d.ts` files in `dist/` or some other subdirectory?

.gitignore
```diff
@@ -3,6 +3,9 @@ node_modules
...
+packages/**/*.js.map
```
acdvorak: Can we ignore _all_ `.map` files? E.g.: `packages/**/*.map`

Presumably we'll need to negate this rule in `.npmignore` so we still publish `.map` files to npm.

.gitignore
moog16: well there are 2 versions of the .d.ts and .js files. There is the UMD and the ES6 versions. All the ES6 versioned code I've placed in the root, so when users want to import it will look like:

```
import Foo from @material/foo/index
```

But maybe we can put it in a `/esm` directory? So the entire package will look like:
```
/dist
/esm
package.json
...
```

WDYT?

.gitignore
moog16: We actually only `.npmignore` `.ts` files, so this won't be an issue for `npm`.

And yes I will change to just ignore all `.map` files

package.json
```diff
@@ -4,20 +4,21 @@
   "license": "MIT",
   "scripts": {
     "build": "npm run clean && mkdirp build && webpack --progress --colors",
+    "build:esmodules": "node ./scripts/build/esmodule-builder.js",
     "build:demos": "npm run clean && mkdirp build && webpack --config=demos/webpack.config.js --progress --colors",
     "build:min": "mkdirp build && cross-env MDC_ENV=production webpack -p --progress --colors",
     "changelog": "standard-changelog -i CHANGELOG.md -k packages/material-components-web/package.json",
-    "clean": "del-cli build/** build .typescript-tmp/** .typescript-tmp",
+    "clean": "del-cli build/** build .typescript-tmp/** .typescript-tmp && git clean -fX",
```
acdvorak: Please don't use `git clean -fX`.

I have a bunch of log data from travis-ci.org (before we migrated to .com) sitting in directories that are manually .gitignore'd on my machine. I also have a bunch of custom TS migration and test scripts too. This command would nuke all of those files, without asking, and it's not possible to recover the log data since Travis no longer has it.

Arguably I shouldn't be keeping important stuff in a gitignore'd directory (and this is a good opportunity for me to move it elsewhere), but `git clean -fX` is still a really dangerous command that should only be used with great care.

It might be better to output generated files (like `.d.ts` and `.js`) in a subdirectory of each package, such as `dist/`, so that we can be more surgical with `clean`.

packages/material-components-web/package.json
acdvorak: I think the `.d.ts` filenames are generated automatically by `toCamelCase()` in `scripts/cp-pkgs.js`.

But I agree it feels a little odd; maybe we should add an exception for this case to keep the filenames consistent.

packages/mdc-tabs/package.json
```diff
@@ -4,6 +4,9 @@
   "description": "The Material Components for the web tabs component",
   "license": "MIT",
   "main": "dist/mdc.tabs.js",
+  "module": "index.js",
+  "types": "dist/mdc.tabs.d.ts",
```
acdvorak: It looks like `md-tabs` is missing a `.npmignore` file

acdvorak: Make sure all other packages have one too 😀

.gitignore
kfranqueiro: The idea was to keep the ESM JS files where they currently are, so that we don't break literally everybody currently using them. That's why the JS and corresponding d.ts files are directly under each package folder.

I knew it'd be a little rough to set up, but it looks like at the very least the gitignore and clean script are handling it here.

kfranqueiro: I guess to follow that up: I'm assuming that d.ts should generally appear where the js files appear... if it's permissible to do otherwise, we could put the d.ts files for individual ESM files under dist or something (given that it seems like typescript just looks for .d.ts anywhere under the project), but that might seem confusing?

.gitignore
kfranqueiro: npmignore _overrides_ gitignore. You shouldn't need to go out of your way to override gitignore in npmignore - npm only checks one or the other.

On the other hand, if there's anything that we _should_ be ignoring that's in our gitignore and not our npmignores, we should fix that...

.gitignore
moog16: I think it is convention that the `.d.ts` files appear sibling to the `.js` files, so I think it'll be confusing for people. I think we should stick to convention.

So it sounds like we need to keep the gitignore like this to keep backwards compatibility.

packages/material-components-web/package.json
moog16: Ya, it is a bit odd that we don't follow the pattern of the rest of the components, but it should match the `main` property. I'll make it an exception.

package.json
moog16: oh no! I hope I didn't just delete all your files. It was just the easiest command to use...the del-cli package doesn't respect node-glob patterns. So it is going to be very verbose.

kfranqueiro: I guess the alternative is add something under `scripts` ourselves?

packages/mdc-tabs/package.json
kfranqueiro: I added it to tabs, removed it from selection-control (because #4442 deletes that), and checked everything else.

moog16: Thanks!

package.json
```diff
@@ -4,23 +4,24 @@
...
     "fix:css": "stylelint --fix \"packages/**/*.scss\"; stylelint --fix --config=test/screenshot/.stylelintrc.yaml \"test/screenshot/**/*.scss\"",
     "fix": "npm-run-all --parallel fix:*",
     "lint:css": "stylelint \"packages/**/*.scss\" && stylelint --config=test/screenshot/.stylelintrc.yaml \"test/screenshot/**/*.scss\"",
     "lint:js": "eslint test scripts webpack.config.js demos/webpack.config.js karma.conf.js",
-    "lint:ts": "tslint --exclude \"test/**/*.d.ts\" \"packages/**/*.ts\" \"test/**/*.ts\" \"scripts/**/*.ts\"",
+    "lint:ts": "tslint --exclude \"**/*.d.ts\" \"packages/**/*.ts\" \"test/**/*.ts\" \"scripts/**/*.ts\"",
     "lint:html": "find test/screenshot/spec/ -name '*.html' | grep -v 'index.html$' | xargs htmllint",
     "lint:imports": "node scripts/check-imports.js",
-    "lint": "npm-run-all --parallel lint:*",
+    "lint": "npm run clean && npm-run-all --parallel lint:*",
```
kfranqueiro: Why was `clean` added to this?

scripts/build/esmodule-builder.js
```diff
@@ -0,0 +1,13 @@
+process.env.NODE_ENV = process.env.NODE_ENV || 'production';
```
kfranqueiro: Is there a reason we default this to production? (I'm not sure whether it's ever not defined or what the default behavior is otherwise.)

It seems like for the most part this script could probably be replaced with a single shell command? Do we even need this script?

moog16: I see what you're saying, since we're only ever going to run this for 'prod'. Also I can't find any documentation for `tsc` running differently with a different NODE_ENV, nor does it matter when I manually change it from the command line.

I will inline this in package.json

scripts/build/clean.js
```diff
@@ -0,0 +1,45 @@
...
+    if (err) {
+      throw err;
+    }
+
+    console.log(`deleted ${directory} directory`);
+  });
+}
+
+function removeFilesOfType(type) {
+  const fileGlob = `packages/*!(node_modules|dist)/*.${type}`;
```
kfranqueiro: I'm pretty sure the `!(...)` here is pointless because there is no `node_modules` or `dist` directly under `packages`. I get the same number of items out of this with or without it.

scripts/build/clean.js
```diff
@@ -0,0 +1,45 @@
+
```
kfranqueiro: I think this script needs a comment explaining its goal. In particular I'm confused why we're seemingly leaving `dist` intact? Your previous approach of `git clean -fX` would have removed that, right?

scripts/build/clean.js
```diff
@@ -0,0 +1,45 @@
...
+      throw err;
+    }
+
+    console.log(`deleted ${directory} directory`);
+  });
+}
+
+function removeFilesOfType(type) {
+  const fileGlob = `packages/*!(node_modules|dist)/*.${type}`;
+  const nestedFileGlob = `packages/*/!(node_modules|dist)/*.${type}`;
```
kfranqueiro: `packages/*/!(node_modules|dist)/**/*.${type}` would future-proof this in case we ever end up with more than one level of subdirectories for some reason.

scripts/cp-pkgs.js
```diff
@@ -48,6 +52,18 @@ if (!isValidCwd) {
   process.exit(1);
 }
 
+function toCamelCase(name) {
```
acdvorak: This function can be simplified to:

```js
/**
 * @param {string} dashedName A dash-separated package name. E.g., "mdc-linear-progress".
 * @return {string} dashedName converted to camelCase. E.g., "mdcLinearProgress".
 */
function toCamelCase(dashedName) {
  return dashedName.replace(/-([a-z])/g, (_, letter) => letter.toUpperCase());
}
```

scripts/verify-pkg-main.js
```diff
@@ -36,31 +36,61 @@ if (!isValidCwd) {
...
   const packageInfo = JSON.parse(fs.readFileSync(jsonPath));
   if (!packageInfo.main) {
     return;
   }
+  verifyPath(packageInfo, jsonPath, 'main');
+  verifyPath(packageInfo, jsonPath, 'module');
+  verifyPath(packageInfo, jsonPath, 'types');
+});
+
+if (invalidMains > 0 || invalidModules > 0 || invalidTypes > 0) {
```
acdvorak:
```js
function logErrorAndExit(message) {
  console.error(message);
  process.exit(1);
}

if (invalidMains > 0) {
  logErrorAndExit(`${invalidMains} incorrect main property values found; please fix.`);
}
if (invalidModules > 0) {
  logErrorAndExit(`${invalidModules} incorrect module property values found; please fix.`);
}
if (invalidTypes > 0) {
  logErrorAndExit(`${invalidTypes} incorrect types property values found; please fix.`);
}

console.log('Success: All packages with main, module, types properties reference appropriate files!');
```

scripts/cp-pkgs.js
```diff
@@ -74,12 +90,67 @@ function cpAsset(asset) {
...
+      throw err;
+    }
+  );
+}
+
+function main() {
+  cleanPkgDistDirs();
+
+  Promise.all(globSync('build/*.{css,js,map}').map(cpAsset)).catch((err) => {
+    console.error(`Error encountered copying assets: ${err}`);
```
acdvorak:
```js
console.error('Error encountered copying assets:', err);
```

scripts/cp-pkgs.js
```diff
@@ -74,12 +90,67 @@ function cpAsset(asset) {
...
+  /**
+   * this method builds the files that the next lines copy to each package
+   */
+  dtsBundler();
+
+  Promise.all(
+    globSync(`build/packages/**/{${DECLARATION_FILE_PREFIX},${ALL_IN_ONE_PACKAGE}}*.d.ts`)
+      .map(cpDeclarationAsset)
+  ).catch((err) => {
+    console.error(`Error encountered copying assets: ${err}`);
```
acdvorak:
```js
console.error('Error encountered copying assets:', err);
```

package.json
```diff
@@ -4,20 +4,21 @@
   "license": "MIT",
   "scripts": {
     "build": "npm run clean && mkdirp build && webpack --progress --colors",
+    "build:esmodules": "$(npm bin)/tsc --project ./tsconfig.json --module esnext",
```
acdvorak: I don't think you need `$(npm bin)/`. Isn't it automatically added to the `$PATH` for all `scripts` in `package.json`? E.g., we call `standard-changelog` directly below. The `$(...)` syntax is not Windows-compatible.

packages/mdc-base/package.json
```diff
@@ -4,8 +4,14 @@
...
   "main": "dist/mdc.base.js",
+  "module": "index.js",
+  "sideEffects": false,
+  "types": "dist/mdc.base.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/material-components/material-components-web.git"
+  },
+  "dependencies": {
+    "tslib": "^1.9.3"
```
acdvorak: Does greenkeeper update dependencies in `packages/*/`? Or will we need to remember to update them all manually?

packages/mdc-dom/package.json
```diff
@@ -4,6 +4,9 @@
   "version": "0.41.0",
   "license": "MIT",
   "main": "dist/mdc.dom.js",
+  "module": "index.js",
+  "sideEffects": false,
+  "types": "dist/mdc.dom.d.ts",
```
acdvorak: Why do other packages depend on `tslib` while this one doesn't?

scripts/build/clean.js
```diff
@@ -0,0 +1,46 @@
...
+  removeDirectory('build');
+  removeDirectory('.rewrite-tmp');
+  removeFilesOfType('css');
+  removeFilesOfType('js');
+  removeFilesOfType('d.ts');
+  removeFilesOfType('map');
+}
+
+function removeDirectory(directory) {
+  exec(`rm -rf ${directory}`, (err) => {
```
acdvorak: Do we care about Windows support? You might want to add `del-cli` back to the root `package.json` and use it here instead of executing a shell command.

scripts/cp-pkgs.js
```diff
@@ -74,12 +90,67 @@ function cpAsset(asset) {
...
+  const packageName = path.parse(asset).name.replace(/^mdc-|\.d$/g, '');
+  const isAllInOne = packageName === ALL_IN_ONE_PACKAGE;
+  const destFileName = isAllInOne ? packageName : `mdc.${toCamelCase(packageName)}`;
+  return path.join(assetPkg, 'dist', `${destFileName}.d.ts`);
+}
+
+function cpDeclarationAsset(asset) {
+  const assetPkg = path.parse(asset.split('build/')[1]).dir;
+  if (!fs.existsSync(assetPkg)) {
+    Promise.reject(new Error(`Non-existent asset package path ${assetPkg} for ${asset}`));
```
acdvorak: `return Promise.reject(...)`

scripts/build/clean.js
```diff
@@ -0,0 +1,46 @@
...
+    if (err) {
+      throw err;
+    }
+  });
+}
+
+function removeFilesOfType(type) {
+  const fileGlob = `packages/*/*.${type}`;
+  const nestedFileGlob = `packages/*/!(node_modules)/**/*.${type}`;
```
acdvorak: Remove the negation and pass `{ignore: ['**/node_modules/**']}` to `globSync()` as the second param.

scripts/cp-pkgs.js
```diff
@@ -74,12 +90,67 @@ function cpAsset(asset) {
...
-  process.exit(1);
-});
+function getDeclarationFileName(asset, assetPkg) {
+  const packageName = path.parse(asset).name.replace(/^mdc-|\.d$/g, '');
+  const isAllInOne = packageName === ALL_IN_ONE_PACKAGE;
+  const destFileName = isAllInOne ? packageName : `mdc.${toCamelCase(packageName)}`;
+  return path.join(assetPkg, 'dist', `${destFileName}.d.ts`);
+}
+
+function cpDeclarationAsset(asset) {
```
acdvorak: Mark this function `async` and add a JSDoc annotation so we know it returns a `Promise<whatever>`

packages/mdc-dom/package.json
kfranqueiro: Probably because it doesn't use any TS syntax requiring helper functions. I think we should probably add tslib to all packages that contain TS though, because this is very likely to silently break the moment we unwittingly add something that needs it.

package.json
```diff
@@ -4,20 +4,21 @@
...
     "build:demos": "npm run clean && mkdirp build && webpack --config=demos/webpack.config.js --progress --colors",
     "build:min": "mkdirp build && cross-env MDC_ENV=production webpack -p --progress --colors",
     "changelog": "standard-changelog -i CHANGELOG.md -k packages/material-components-web/package.json",
-    "clean": "del-cli build/** build .rewrite-tmp/** .rewrite-tmp",
-    "clean:site": "del-cli .site-generator-tmp/** .site-generator-tmp",
-    "dist": "npm run build && npm run build:min",
+    "clean": "node -e \"require('./scripts/build/clean').main()\"",
```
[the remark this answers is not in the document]
moog16: I export main() and removeDirectory() from the clean.js file. I believe this is the correct syntax for the npm CLI.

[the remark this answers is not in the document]
kfranqueiro: What you can do in addition is this:

```js
if (require.main === module) {
  main();
}
```

Then you should be able to just do `node scripts/build/clean`.

scripts/cp-pkgs.js
moog16: fancy! :)

scripts/verify-pkg-main.js
moog16: should I also update the `verifyPath` method to use this function as well?

package.json
moog16: ah ya you're right - don't need it in this context, and it is not windows compat

packages/mdc-base/package.json
moog16: Not sure - will have to look into it. If greenkeeper does not keep track of it, is there something you recommend?

packages/mdc-dom/package.json
moog16: The `lint:imports` check will actually catch this. Otherwise we'll add a package that isn't being used.

scripts/build/clean.js
moog16: ah...definitely

scripts/build/clean.js
moog16: Oh wow - that's so much better. I was thinking I made this too complicated.

scripts/build/clean.js
```diff
@@ -0,0 +1,44 @@
...
+ * @fileoverview Cleans top-level build directories (for OSS and internal processes),
+ * dist subfolders under packages, and .js[.map] files in package directories.
+ */
+
+const {exec} = require('child_process');
+const fs = require('fs');
+
+const {sync: globSync} = require('glob');
+
+if (require.main === module) {
```
kfranqueiro: Nit: move to end of file

scripts/cp-pkgs.js
```diff
@@ -68,18 +88,96 @@ function getAssetEntry(asset) {
...
+  Promise.all(
+    globSync(`build/packages/**/{${DECLARATION_FILE_PREFIX},${ALL_IN_ONE_PACKAGE}}*.d.ts`)
+      .map(cpDeclarationAsset)
+  ).catch((err) => {
+    console.error('Error encountered copying assets:', err);
+    process.exit(1);
+  });
+}
+
+main();
```
kfranqueiro: Nit: Is there even a reason for `main` to be a function encapsulating this code if we're just immediately executing it anyway?

scripts/verify-pkg-main.js
```diff
@@ -36,31 +36,73 @@ if (!isValidCwd) {
...
+  if (invalidMains > 0) {
+    logErrorAndExit(`${invalidMains} incorrect main property values found; please fix.`);
+  }
+  if (invalidModules > 0) {
+    logErrorAndExit(`${invalidModules} incorrect module property values found; please fix.`);
+  }
+  if (invalidTypes > 0) {
+    logErrorAndExit(`${invalidTypes} incorrect types property values found; please fix.`);
+  }
+} else {
```
kfranqueiro: You can remove the if/else now (as illustrated in Andy's original comment) since this line will only be reached if none of the above `if`s were true.

scripts/verify-pkg-main.js
```diff
@@ -36,31 +36,73 @@ if (!isValidCwd) {
...
+  console.log('Success: All packages with main, module, types properties reference appropriate files!');
+}
 
-  const mainPath = path.join(path.dirname(jsonPath), packageInfo.main);
+/**
```
kfranqueiro: Nit: Move these functions up above the immediately-executing code that uses them

package.json
```diff
@@ -4,20 +4,21 @@
...
     "build": "npm run clean && mkdirp build && webpack --progress --colors",
+    "build:esmodules": "tsc --project ./tsconfig.json --module esnext",
     "build:demos": "npm run clean && mkdirp build && webpack --config=demos/webpack.config.js --progress --colors",
     "build:min": "mkdirp build && cross-env MDC_ENV=production webpack -p --progress --colors",
     "changelog": "standard-changelog -i CHANGELOG.md -k packages/material-components-web/package.json",
-    "clean": "del-cli build/** build .rewrite-tmp/** .rewrite-tmp",
-    "clean:site": "del-cli .site-generator-tmp/** .site-generator-tmp",
-    "dist": "npm run build && npm run build:min",
+    "clean": "node ./scripts/build/clean",
+    "clean:site": "node -e \"require('./scripts/build/clean').removeDirectory('.site-generator-tmp')\"",
```
acdvorak: Can we pass a flag to `scripts/build/clean` instead of inlining JS code here?

E.g.:

```
"clean:site": "node ./scripts/build/clean --site-generator-only",
```

package.json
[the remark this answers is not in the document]
acdvorak: Can we pass a CLI flag instead? Or have two separate JS files? Inline JS in `package.json` doesn't feel right.

scripts/verify-pkg-main.js
acdvorak: It looks like `verifyPath()` is tallying up the number of invalid package properties, so we probably don't want to call `process.exit()` inside of it. Am I understanding that right?

scripts/build/clean.js
```diff
@@ -0,0 +1,44 @@
...
+  removeDirectory('build');
+  removeDirectory('.rewrite-tmp');
+  removeFilesOfType('css');
+  removeFilesOfType('js');
+  removeFilesOfType('d.ts');
+  removeFilesOfType('map');
+}
+
+function removeDirectory(directory) {
+  exec(`del-cli ${directory}/ ${directory}`, (err) => {
```
acdvorak: My bad - it looks like we want https://github.com/sindresorhus/del instead. That'll give us a programmatic API so we don't have to invoke a subshell.

scripts/build/clean.js
```diff
@@ -0,0 +1,44 @@
...
+    if (err) {
+      throw err;
+    }
+  });
+}
+
+function removeFilesOfType(type) {
+  const fileGlob = `packages/*/*.${type}`;
```
acdvorak: Should this be `packages/**/*.${type}` to include sub-directories (e.g., `mdc-drawer/dismissible/`)?

scripts/verify-pkg-main.js
kfranqueiro: Actually... Andy has a more valid point here, that we shouldn't be using logErrorAndExit at all, so ignore my comment here

scripts/build/clean.js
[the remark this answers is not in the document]
acdvorak: I vaguely recall running into a situation where `del-cli` would delete all subdirectories and files, but not the parent directory itself, on macOS. Not sure if that's still an issue tho.

scripts/verify-pkg-main.js
kfranqueiro: Doesn't that mean we shouldn't use it here either? Or simplify the message to just "problems found" in general, since iirc we're already spitting out more specific error messages for each problem encountered?

acdvorak: Oh yeah, good point. So maybe `logErrorAndExit()` needs to be updated to something like:

```js
function logError(message) {
  console.error(message);
  process.exitCode = 1;
}
```

packages/mdc-base/package.json
acdvorak: No idea lol

kfranqueiro: I wouldn't be too inclined to worry about it - we have typescript at top-level, right? Any time we update that we could look at whether this has any updates.

scripts/build/clean.js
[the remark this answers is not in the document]
moog16: Ya I was copying what we had before in the package.json. I remember you (ken) saying that we needed it because of Windows. But again not sure if its an issue either. I can test it out on windows and mac and see if just ${directory} works.

scripts/cp-pkgs.js
moog16: I thought it cleaned it up by putting it in a function. I can move it out.

scripts/build/clean.js
moog16: yes i just noticed that too when testing text-field. Thanks!

scripts/build/clean.js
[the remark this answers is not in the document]
kfranqueiro: Yes, but we had `del-cli dir/** dir` before, not `dir/ dir` which AFAIK is exactly the same thing twice.

scripts/verify-pkg-main.js
```diff
@@ -75,14 +40,14 @@ function verifyPath(packageJson, jsonPath, packagePropertyKey) {
   let isInvalid = false;
   if (!isAtRoot && packageJsonPropPath.indexOf('dist') === -1) {
     isInvalid = true;
-    console.error(`${jsonPath} ${packagePropertyKey} property does not reference a file under dist`);
+    logError(`${jsonPath} ${packagePropertyKey} property does not reference a file under dist`);
```
kfranqueiro: This is now going to exit as soon as a single invalid path is found, which is less helpful than printing all issues out at once so that someone can fix them all at once before running it again.

kfranqueiro: Nevermind, I just now realized the change made to `logError`, which is ridiculously subtle...

scripts/build/clean.js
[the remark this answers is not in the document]
kfranqueiro: Going to resolve this thread given we're now using `del` and I haven't seen the old problem come up.

packages/mdc-dom/package.json
kfranqueiro: You're referring to running `lint:imports` specifically on the generated JS? Weren't we thinking of updating that to run on the `ts` itself? Otherwise `lint:imports` is a no-op without running `npm run build:esmodules` first (which for example our lint task on Travis CI won't do). But if we convert it to TS it will no longer catch this.

packages/mdc-base/package.json
merlinnot: Renovate has a great support for monorepos.

packages/mdc-dom/package.json
kfranqueiro: After talking with Andy this morning, I've added tslib to the animation and dom packages, which are the only individual packages containing TS that didn't already have it. I think it's safe to leave it out of material-components-web since we only ever add to the existing code without adding entirely new structures. This way we can eventually convert the import check script to look at the TS files along with what's in #4449.

packages/mdc-base/package.json
acdvorak: Thanks for the suggestion @merlinnot! We'll look into that 😀

scripts/cp-pkgs.js
```diff
@@ -68,18 +88,92 @@ function getAssetEntry(asset) {
...
+  const destDir = getDeclarationFileName(asset, assetPkg);
+  return cpFile(asset, destDir).then(
+    () => console.log(`cp ${asset} -> ${destDir}`),
+    (err) => {
+      throw err;
+    }
+  );
+}
+
+/**
```
acdvorak: Nit: Remove second `*` here and add a newline after `*/`. This comment describes everything below it; it's not a JSDoc comment for `cleanPkgDistDirs()` specifically.

scripts/verify-pkg-main.js
```diff
@@ -27,6 +27,51 @@ const {sync: globSync} = require('glob');
 
 const isValidCwd = fs.existsSync('packages') && fs.existsSync(path.join('packages', 'material-components-web', 'dist'));
 
+/**
+ * Verifies that a file exists at the `packagePropertyKey`. If it doesnot
```
acdvorak: Nit: `does not`

scripts/verify-pkg-main.js
```diff
@@ -27,6 +27,51 @@ const {sync: globSync} = require('glob');
...
+ * this function will console an error.
```
acdvorak: Nit: `this function will log an error to the console.`

PR #4410, opened 2019-02-13
feat(toolbar): Convert JS to TypeScript

acdvorak:
Refs #4225


packages/mdc-toolbar/component.ts
```diff
@@ -23,38 +23,28 @@
...
-    return this.fixedAdjustElement_;
-  }
+  private ripples_!: MDCRipple[]; // assigned in initialize()
+  private fixedAdjustElement_?: HTMLElement;
 
   initialize() {
-    this.ripples_ = [].map.call(this.root_.querySelectorAll(MDCToolbarFoundation.strings.ICON_SELECTOR), (icon) => {
+    this.ripples_ = [];
+
+    [].forEach.call(this.root_.querySelectorAll(strings.ICON_SELECTOR), (icon: HTMLElement) => {
```
kfranqueiro: Did the existing `[].map` not work?

packages/mdc-toolbar/foundation.ts
```diff
@@ -220,11 +255,13 @@ export default class MDCToolbarFoundation extends MDCFoundation {
...
+  private updateElementStylesDefaultBehavior_(flexibleExpansionRatio: number) {
+    const maxTitleSize = numbers.MAX_TITLE_SIZE;
+    const minTitleSize = numbers.MIN_TITLE_SIZE;
     const currentTitleSize = (maxTitleSize - minTitleSize) * flexibleExpansionRatio + minTitleSize;
 
     this.adapter_.setStyleForTitleElement('font-size', `${currentTitleSize}rem`);
   }
 }
+
+export default MDCToolbarFoundation;
```
moog16: change to export {...as default, ...}

packages/mdc-toolbar/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+export {MDCToolbar as default} from './component';
```
moog16: should we have a default? Is this for backwards compatibility?

packages/mdc-toolbar/component.ts
acdvorak: It technically worked, but it was more cumbersome and less type safe. TS can't infer the return type of the array, so I had to cast it to `MDCRipple[]`:

```ts
this.ripples_ = [].map.call(this.root_.querySelectorAll(strings.ICON_SELECTOR), (icon: HTMLElement) => {
  const ripple = MDCRipple.attachTo(icon);
  ripple.unbounded = true;
  return ripple;
}) as MDCRipple[];
```

If we return `"foo"` here instead of `ripple`, TypeScript won't yell at us 🙁

packages/mdc-toolbar/adapter.ts
```diff
@@ -0,0 +1,53 @@
...
+  getOffsetHeight: () => number;
+  getFirstRowElementOffsetHeight: () => number;
+  notifyChange: (evtData: ToolbarEventDetail) => void;
+  setStyle: (property: string, value: string) => void;
+  setStyleForTitleElement: (property: string, value: string) => void;
+  setStyleForFlexibleRowElement: (property: string, value: string) => void;
+  setStyleForFixedAdjustElement: (property: string, value: string) => void;
+}
+
+export default MDCToolbarAdapter;
```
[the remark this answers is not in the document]
acdvorak: I figured we should make all of our exports consistent, so that we don't need to remember which ones need a legacy `default` export and which ones don't.

packages/mdc-toolbar/component.ts
```diff
@@ -65,8 +55,25 @@ export class MDCToolbar extends MDCComponent {
...
+
+  get fixedAdjustElement(): HTMLElement | undefined {
+    return this.fixedAdjustElement_;
+  }
+
+  private get firstRowElement_(): HTMLElement {
+    return this.root_.querySelector<HTMLElement>(strings.FIRST_ROW_SELECTOR)!;
+  }
+
+  private get titleElement_(): HTMLElement {
```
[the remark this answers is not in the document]
acdvorak: Nope! Good call.

packages/mdc-toolbar/component.ts
```diff
@@ -89,5 +96,8 @@ export class MDCToolbar extends MDCComponent {
         }
       },
     });
+    // tslint:enable:object-literal-sort-keys
   }
 }
+
+export default MDCToolbar;
```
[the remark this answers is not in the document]
acdvorak: I actually switched to this split export syntax to make it easier to disable the default export in the future. See #4407 PR description for an explanation.

packages/mdc-toolbar/index.ts
acdvorak: Not sure if it's needed for backward compatibility, but all of our packages should be consistent. It's really weird for clients to have to use a default export for one component and a named export for another component.

packages/mdc-toolbar/foundation.ts
acdvorak: I actually switched to this split export syntax to make it easier to disable the default export in the future. See #4407 PR description for an explanation.

packages/mdc-toolbar/index.ts
kfranqueiro: Come to think of it, I don't think any of our `index` files have default exports. The other modules do.

acdvorak: Good point! I added https://github.com/material-components/material-components-web/pull/4407#discussion_r256948622 to remind myself to remove these in all `index.ts` and `component.ts` files.

acdvorak: Removed `default` export from `index.ts` and `component.ts`

packages/mdc-toolbar/component.ts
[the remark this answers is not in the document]
acdvorak: As discussed offline, we're going to remove the `default` exports from `component.ts` and `index.ts` because they do not currently exist in `index.js` on `master`.

Removed.

packages/mdc-toolbar/component.ts
```diff
@@ -65,8 +56,25 @@ export class MDCToolbar extends MDCComponent {
...
+  set fixedAdjustElement(element: HTMLElement | null) {
+    this.fixedAdjustElement_ = element;
+    this.foundation_.updateAdjustElementStyles();
+  }
+
+  get fixedAdjustElement(): HTMLElement | null {
+    return this.fixedAdjustElement_;
+  }
+
+  private get firstRowElement_(): HTMLElement {
```
moog16: isn't it also accurate that this could return both `HTMLElement | null`?

packages/mdc-toolbar/foundation.ts
```diff
@@ -37,65 +74,63 @@ export default class MDCToolbarFoundation extends MDCFoundation {
...
+    toolbarRowHeight: 0,
+  };
+
+  private readonly resizeHandler_: SpecificEventListener<'resize'>;
+  private readonly scrollHandler_: SpecificEventListener<'scroll'>;
+
+  constructor(adapter?: Partial<MDCToolbarAdapter>) {
+    super({...MDCToolbarFoundation.defaultAdapter, ...adapter});
+
     this.resizeHandler_ = () => this.checkRowHeight_();
```
moog16: I think these can actually be defined outside the constructor

packages/mdc-toolbar/component.ts
acdvorak: From the README, it looks like a row element is always present. Is that not the case?

packages/mdc-toolbar/foundation.ts
acdvorak: Done

packages/mdc-toolbar/component.ts
moog16: It should be...I was thinking we should throw another error if it doesn't exist, but is that the only required element in this component?

acdvorak: Yeah it looks like this is the only required subelement

moog16: let's throw an error then if the firstRowElement is not supplied. Then we can definitely say that this getter will always return an HTMLElement.

acdvorak: Good call. Done. I also made this a normal property rather than a getter for consistency with other packages.

Issue #4173, opened 2018-12-13
Enhanced select menu doesn't stay open


MarcoTroost:
When using the the enhanced version of the mdc-select menu (v. 0.42.0), the menu-list immediately disappears. One would expect the menu to stay open until it loses focus.

This behavior can also be seen @ https://material-components.github.io/material-components-web-catalog/#/component/select , at least on my computer (mac os mojave, chrome v 71).

williamernest:
Thanks for filing this issue. It looks like any long-click on the enhanced select will cause the menu to close when you release. Definitely a bug.

shelart:
I think it's caused by the same reason as the enhanced MDCSelect doesn't work on touch screens. It seems like Chrome emits both events `touchstart` and `mousedown` when a laptop trackpad used, as a touch screen on mobile devices.
I actually don't understand why MDC have to listen for `...down` events instead of listening just `click`-s, which are working well on all types of devices: mouse, touchpad, touch screen (tested).

PR #4413, opened 2019-02-14
fix(MDCSelect): issue #4173

shelart:
Fix Enhanced MDCSelect behaviour on touch screens/desktops with touchpad

#4173

### DIFF
--- a/packages/mdc-select/index.js
+++ b/packages/mdc-select/index.js
@@ -358,7 +358,7 @@ class MDCSelect extends MDCComponent {
     element.addEventListener('focus', this.handleFocus_);
     element.addEventListener('blur', this.handleBlur_);
 
-    ['mousedown', 'touchstart'].forEach((evtType) => {
+    ['click'].forEach((evtType) => {
       element.addEventListener(evtType, this.handleClick_);
     });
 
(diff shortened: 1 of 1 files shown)


kfranqueiro:
Can you describe what this is fixing? We purposely listen to both touch and mouse events, so this should already be working...

I'm also cognizant that we most likely had a specific reason for listening to down events rather than the click event.

shelart:
Looking at Travis CI build log, I suspect the reason for listening to `...down` events was to correspond with the ripple effect. The `handleClick_()` method (which was invoked by `...down` events) calls the `handleClick()` Foundation's method which is responsible for the ripple, and the latter also calls the adapter's `openMenu()` method.

By other words, `...down` events invokes both actions - the ripple effect & the menu opening. Obviously, a ripple should show on touchdown, not complete click (which is currently broken by my PR). My assumption is that's only reason to listen to `...down` events.

kfranqueiro:
I checked with Will who worked on this component earlier and he suggested there may have been something related to focus behavior as to why we listened to down events. I'd have to dig through history and test it to be sure.

moog16:
I just checked with Will that the issue he saw before isn't an issue any longer (it was probably fixed with MenuSurface changes). We checked FF and Chrome. We just need to check IE11/Edge before we approve this PR.

moog16:
Closing for #4413

PR #4414, opened 2019-02-14
fix(tab): Fix closure test errors

abhiomkar:

### DIFF
--- a/packages/mdc-tab-scroller/rtl-scroller.js
+++ b/packages/mdc-tab-scroller/rtl-scroller.js
@@ -33,8 +33,8 @@ import {MDCTabScrollerAdapter, MDCTabScrollerAnimation} from './adapter';
 class MDCTabScrollerRTL {
   /** @param {!MDCTabScrollerAdapter} adapter */
   constructor(adapter) {
-    /** @private */
-    this.adapter_ = adapter;
+    /** @package */
+    this.adapter = adapter;
   }
 
   /**
--- a/packages/mdc-tab-scroller/rtl-scroller.js
+++ b/packages/mdc-tab-scroller/rtl-scroller.js
@@ -33,7 +33,7 @@ import {MDCTabScrollerAdapter, MDCTabScrollerAnimation} from './adapter';
 class MDCTabScrollerRTL {
   /** @param {!MDCTabScrollerAdapter} adapter */
   constructor(adapter) {
-    /** @package */
+    /** @package {!MDCTabScrollerAdapter} */
     this.adapter = adapter;
   }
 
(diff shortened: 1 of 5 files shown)


packages/mdc-tab-scroller/rtl-scroller.js
```diff
@@ -33,8 +33,8 @@ import {MDCTabScrollerAdapter, MDCTabScrollerAnimation} from './adapter';
 class MDCTabScrollerRTL {
   /** @param {!MDCTabScrollerAdapter} adapter */
   constructor(adapter) {
-    /** @private */
-    this.adapter_ = adapter;
+    /** @package {!MDCTabScrollerAdapter} */
```
kfranqueiro: I forget, does closure not have `@protected`? (We're changing this to `protected readonly` on the TypeScript branch.)

packages/mdc-tab-scroller/rtl-scroller.js
```diff
@@ -33,8 +33,8 @@ import {MDCTabScrollerAdapter, MDCTabScrollerAnimation} from './adapter';
...
+    this.adapter = adapter;
```
kfranqueiro: Can we not still keep the underscore? (We're keeping it on the TS branch even though we're switching it to protected)

abhiomkar: Closure doesn't like the underscore suffix for non-private properties.

packages/mdc-tab-scroller/rtl-scroller.js
abhiomkar: `@protected` sounds relevant here.

packages/mdc-tab-scroller/rtl-scroller.js
kfranqueiro: I'm starting to wonder i we actually ran into this before...

I'd really rather avoid making `adapter` in this _one_ module when it's `adapter_` absolutely everywhere else in the repo if we could avoid it... especially if we can get away with `protected adapter_` when we convert to TypeScript. I'm not sure if we've tested importing / building tab packages yet. @acdvorak?

acdvorak: I imported all tab-related packages from #4407 (commit 07bc5ea) and it worked _with_ the underscore (in TypeScript). I'm guessing Closure doesn't complain about generated code?

acdvorak: Can we selectively disable the Closure warning for this line instead of renaming the property?

kfranqueiro: I'm wondering if we should abandon this PR if it will work in TypeScript and the SCSS changes are already covered in the other PR.

acdvorak: @abhiomkar Is this PR critical?
Or can we wait until TypeScript gets merged in #4407 and #4409?

abhiomkar: It isn't critical. This change just avoids lint errors in closure context. Since we'll be merging TS branch sooner, we may no longer need this change?

PR #4420, opened 2019-02-15
chore(infrastructure): Fix GitHub API authentication bug

acdvorak:
A [recent update to the GitHub Octokit API](https://github.com/octokit/rest.js/issues/1228) requires a new auth token format:

`"token ABC..."` instead of `"ABC..."`.

This PR fixes errors like this:

https://travis-ci.com/material-components/material-components-web/jobs/178167993#L6363

```
API rate limit exceeded for 35.224.112.202. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
HttpError: API rate limit exceeded for 35.224.112.202. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
    at response.text.then.message (/home/travis/build/material-components/material-components-web/node_modules/@octokit/request/lib/request.js:55:27)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
↳  called from:
Failed to get pull request number for branch "feat/typescript":
Error: GitHubApi.getPullRequestNumber()
    at GitHubApi.getPullRequestNumber (/home/travis/build/material-components/material-components-web/test/screenshot/infra/lib/github-api.js:138:20)
    at DiffBaseParser.parseDiffBase (/home/travis/build/material-components/material-components-web/test/screenshot/infra/lib/diff-base-parser.js:105:46)
```

In addition, this PR changes our custom `GitHubApi` wrapper class so that it no longer throws an exception when Octokit returns an error, so screenshot tests will no longer fail on Travis CI if posting a status check or PR comment fails, for example.


test/screenshot/infra/lib/diff-base-parser.js
```diff
@@ -83,7 +83,7 @@ class DiffBaseParser {
       if (process.env.TRAVIS_BRANCH) {
         baseBranch = `origin/${process.env.TRAVIS_BRANCH}`;
       } else {
-        baseBranch = await this.gitHubApi_.getPullRequestBaseBranch(prNumber);
+        baseBranch = await this.gitHubApi_.getPullRequestBaseBranch(prNumber) || baseBranch;
```
kfranqueiro: Wouldn't this cause incorrect diffs against base in cases where we're PR'ing against feat/typescript?

Also, this line suggests that getPullRequestBaseBranch just returns falsy in the problem case (it doesn't actually throw)?

acdvorak: Yes, it would cause incorrect base diffs for PRs merging into non-master branches if the GitHub API returns an error (e.g., due to rate limiting on an unauthenticated account).

I removed all the `throw new Error()` statements from the `GitHubApi` class because GitHub integration isn't totally critical to screenshot tests, so we probably shouldn't cause Travis CI to fail when a status check `POST` fails.

test/screenshot/infra/commands/test.js
```diff
@@ -462,8 +462,6 @@ ${CliColor.bold.red('Skipping screenshot tests.')}
    */
   logComparisonResults_(reportData) {
     this.logger_.foldStart('screenshot.diff_results', 'Diff results');
-    this.logComparisonResultSet_('Skipped', reportData.screenshots.skipped_screenshot_list);
```
kfranqueiro: Is there a reason to remove Skipped? It shouldn't show up in practice on Travis but might be useful locally, IIUC...

test/screenshot/infra/lib/github-api.js
```diff
@@ -167,16 +167,32 @@ class GitHubApi {
...
+  /**
+   * @param {string} message
+   * @param {string} stackTrace
+   * @param {(Error|string)=} unsafeErr
+   * @private
+   */
+  error_(message, stackTrace, unsafeErr = '') {
+    let redactedError = nodeUtil.inspect(unsafeErr, /* showHidden */ false, /* depth */ 5);
+    while (this.authToken_ && redactedError.indexOf(this.authToken_) > -1) {
+      redactedError = redactedError.replace(this.authToken_, '[REDACTED]');
```
kfranqueiro: Is there anything else we need to redact? Should we really be spitting out this entire object?

test/screenshot/infra/lib/report-builder.js
```diff
@@ -996,20 +996,7 @@ class ReportBuilder {
   logRunParameters_(verb, screenshots) {
     const count = screenshots.length;
     const plural = count === 1 ? '' : 's:';
-
     console.log(`${verb} ${count} screenshot${plural}`);
-    if (count > 0) {
```
kfranqueiro: What output was this responsible for?

test/screenshot/infra/commands/test.js
acdvorak: Good point. Reverted.

test/screenshot/infra/lib/report-builder.js
acdvorak: This gets called at the beginning of the test. It prints out the list of files that will be compared, added/removed, and skipped.

I was trying to clean up some log spam, but I'll revert this change for now so we can get the GitHub fix in.

test/screenshot/infra/lib/github-api.js
acdvorak: I didn't see anything else. I figured it would be better to print the full error message for debugging purposes in case something like this happens again, but I could be persuaded otherwise.

For example, GitHub returns an HTTP 404 for an invalid auth token (instead of HTTP 401/403), and I wouldn't have been able to figure that out had I not been able to see the full error message.

WDYT?

Here's the raw error output:

```
Failed to set commit status::
Error: GitHubApi.setPullRequestStatus()
    at GitHubApi.setPullRequestStatus (/Users/advorak/dev/mdc-web/test/screenshot/infra/lib/github-api.js:84:20)
    at process._tickCallback (internal/process/next_tick.js:68:7)

 { HttpError: Not Found
    at response.text.then.message (/Users/advorak/dev/mdc-web/node_modules/@octokit/request/lib/request.js:55:27)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  name: 'HttpError',
  status: 404,
  headers:
   { 'access-control-allow-origin': '*',
     'access-control-expose-headers':
      'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type',
     connection: 'close',
     'content-encoding': 'gzip',
     'content-security-policy': 'default-src \'none\'',
     'content-type': 'application/json; charset=utf-8',
     date: 'Wed, 20 Feb 2019 18:24:28 GMT',
     'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
     server: 'GitHub.com',
     status: '404 Not Found',
     'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
     'transfer-encoding': 'chunked',
     'x-content-type-options': 'nosniff',
     'x-frame-options': 'deny',
     'x-github-media-type': 'github.v3; format=json',
     'x-github-request-id': '8D00:62C7:34C324:813403:5C6D9B5C',
     'x-ratelimit-limit': '60',
     'x-ratelimit-remaining': '57',
     'x-ratelimit-reset': '1550690663',
     'x-xss-protection': '1; mode=block' },
  request:
   { method: 'POST',
     url:
      'https://api.github.com/repos/material-components/material-components-web/statuses/a917eb8ee10d077ff748f4a046da8aab32e7e43a',
     headers:
      { accept: 'application/vnd.github.v3+json',
        'user-agent':
         'octokit.js/16.15.0 Node.js/10.14.2 (macOS High Sierra; x64)',
        authorization: '[REDACTED]',
        'content-type': 'application/json; charset=utf-8' },
     body:
      '{"state":"pending","target_url":"https://github.com/material-components/material-components-web","description":"0 of 1 (0.0%) - 0 diffs","context":"screenshot-test/butter-bot"}',
     request:
      { validate:
         { context: { type: 'string' },
           description: { type: 'string' },
           owner: { required: true, type: 'string' },
           repo: { required: true, type: 'string' },
           sha: { required: true, type: 'string' },
           state:
            { enum: [ 'error', 'failure', 'pending', 'success' ],
              required: true,
              type: 'string' },
           target_url: { type: 'string' } } } },
  documentation_url:
   'https://developer.github.com/v3/repos/statuses/#create-a-status' }
```

acdvorak:
This PR was inadvertently included in #4407 which got merged first.

PR #4421, opened 2019-02-15
feat(tab): Implement a ripple color mixin for Tab

bwobrien:


packages/mdc-tab/_mixins.scss
```diff
@@ -36,6 +37,17 @@
   }
 }
 
+@mixin mdc-tab-ripple-color($color) {
```
kfranqueiro: Should we call this mdc-tab-states-color given that it's invoking mdc-states?

patrickrodee: That's a good idea. 

Just an aside: the mixed naming between "ripple" and "states" keeps tripping me up.

bwobrien: SGTM, done.

bwobrien:
Yeah sorry about that. I will update the docs.

bwobrien:
Added docs for the mixins.

packages/mdc-tab/_mixins.scss
```diff
@@ -36,6 +37,17 @@
   }
 }
 
+@mixin mdc-tab-states-color($color) {
+  .mdc-tab__ripple {
+    @include mdc-states($color);
+  }
+}
+
+@mixin mdc-tab-ink-color($color) {
```
kfranqueiro: Since this is here... should this also set mdc-tab-icon-color so that everything is set together? (I forget why text and icon color were originally separate, maybe @patrickrodee can weigh in on this.)

patrickrodee: We had separate mixins for text and icon in other components when I added these mixins. Setting `icon` color in the `ink` mixin makes sense to me.

bwobrien: Makes sense, my definition of ink was outdated. Done.

PR #4422, opened 2019-02-15
chore: Add @alternate annotation to fix closure sass errors

abhiomkar:


packages/mdc-tab/mdc-tab.scss
```diff
@@ -46,7 +46,9 @@
   outline: none;
   background: none;
   text-align: center;
+  /* @alternate */
```
kfranqueiro: I'm actually not sure why we have `text-decoration: none` and `text-transform: uppercase` here... we're including `mdc-typography(button)` which sets the exact same thing. Could we just remove these styles instead or is there a reason I'm not seeing as to why they're here (i.e. to intentionally hard-code them even if the button typography settings are changed)?

patrickrodee: We can probably ditch them. I can't remember why they were included tbh

kfranqueiro:
@acdvorak are those not the same 2 lines that are removed in this PR for being redundant?

PR #4427, opened 2019-02-18
docs(tab-bar): Recommend Material Icons

dfabulich:
This recommendation fixes the example, which implicitly uses Material Icons.


kfranqueiro:
Thanks for spotting this omission!

PR #4428, opened 2019-02-19
docs: Recommend Self-Hosting Material Icons in Getting Started Guide

dfabulich:
A number of examples in the documentation rely on Material Icons implicitly. The Getting Started guide should recommend using them. (This would have saved me a lot of time today.)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -18,6 +18,14 @@ To try Material Components for the web with minimal setup, load the precompiled 
 </head>
 ```
 
+We also recommend using [Material Icons](https://material.io/tools/icons/) from Google Fonts. (Some of our examples implicitly use Material Icons.)
+
+```html
+<head>
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+</head>
+```
+
 Then include MDC markup...
 
 ```html
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -12,8 +12,10 @@ path: /docs/getting-started/
 To try Material Components for the web with minimal setup, load the precompiled all-in-one CSS and JS bundles from unpkg:
 
 ```html
-https://unpkg.com/material-components-web@latest/dist/material-components-web.min.css
-https://unpkg.com/material-components-web@latest/dist/material-components-web.min.js
+<head>
+  <link href="https://unpkg.com/material-components-web@latest/dist/material-components-web.min.css" rel="stylesheet">
+  <script src="https://unpkg.com/material-components-web@latest/dist/material-components-web.min.js"></script>
+</head>
 ```
 
 Then include MDC markup...
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -365,7 +365,65 @@ Now run `npm start` again and open http://localhost:8080. You should see a Mater
 
 <img src="button_with_ripple.png" alt="Button with Ripple" width="90" height="36">
 
-### Step 5: Build Assets for Production
+### Step 5: Add Material Icons
+
+See the [Material Icons Guide](https://google.github.io/material-design-icons/) for detailed instructions on self-hosting Material Icons.
+
+To self-host the icon font, first install the Material Icons as a Node dependency:
+
+```
+npm install --save-dev material-design-icons
+```
+
+Now you can use SASS to `@include` the Material Icon CSS file. In this example, we'll use an [Icon Button](https://material.io/develop/web/components/buttons/icon-buttons/).
+
+```
+npm install --save-dev @material/icon-button
+```
+
+Update your `index.html` to include the MDC Icon Button markup, including the `material-icons` class.
+
+```html
+<body>
+  <button class="mdc-icon-button material-icons">favorite</button>
+  <script src="bundle.js" async></script>
+</body>
+```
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -355,7 +365,65 @@ Now run `npm start` again and open http://localhost:8080. You should see a Mater
 
 <img src="button_with_ripple.png" alt="Button with Ripple" width="90" height="36">
 
-### Step 5: Build Assets for Production
+### Step 5: Add Material Icons
```
kfranqueiro: I'm OK with the edits to the top of this guide, but I'd be inclined against the self-hosting advice for now...mostly because the material-design-icons package hasn't had an update published in 3 years.

Also, we could very easily add an icon into the button added in the earlier step rather than introducing a 3rd component, to keep this section more brief.

dfabulich: From my perspective, the core problem is that many component demos implicitly rely on Material Icons. If you follow the steps in the getting started guide and then naively try to use the demos without installing Material Icons, the demos break in a hard-to-understand way.

If Material Icons are not recommended, then the demos should stop implicitly relying on them. But, on the contrary, the demos use them and their documentation says, "We recommend using Material Icons."

And it turns out that installing Material Icons when following this Getting Started guide is pretty tricky! It took me the better part of a day to figure out how to get this to work in webpack. So I'd like that information to be here for the benefit of others.

Would it help if I marked the "Add Material Icons" step as "(Optional)"?

kfranqueiro: I'm not saying Material Icons aren't recommended. I'm specifically suggesting against _self-hosting_ the icons using the npm package. Using the icons off of google fonts is more likely to provide up-to-date resources at this time.

dfabulich: It's hard for me to imagine the sort of organization that wants to self host material components but _not_ material icons.

Whatever reason you want to self-host, all of the reasons you want to self-host your JS apply equally to CSS/icons.

Some people want to self-host for performance reasons, embedding CSS directly in the document head. Some people want to self-host for security reasons. https://jakearchibald.com/2018/third-party-css-is-not-safe/ Some people just want to self-host for stability reasons. (I don't want my site to change just because someone releases a new version of material icons.)

And, anyway, I don't see what's so exciting about "up-to-date resources" for a project that hasn't had a release in three years. :-)

kfranqueiro: I will see if I can get more information from the fonts and icon teams, but I suspect that the hosted icon font may have been updated more recently than the npm package.

dfabulich: I've made a separate PR #4501 including just the edits to the top of the guide. Hopefully we can merge that, and then this PR would just discuss adding self-hosting instructions for Material Icons.

kfranqueiro: Thanks, splitting that out was a good idea. I'll merge that once CI finishes again.

Regarding the npm package, I got word back from someone on the fonts team with knowledge on the icon font, and he confirmed that the icon font is being maintained on Google Fonts but there is currently no activity on the npm package, so I think we should leave this out.

I'm still open to the alternative idea of adding an example of an [icon inside the button](https://github.com/material-components/material-components-web/tree/v1.0.1/packages/mdc-button#icons) in the section of the getting started guide where we add a button.

dfabulich: Maybe you can suggest a better way to solve this problem?

The problem I really wanted to solve is to copy and paste the example from https://material.io/develop/web/components/tabs/tab-bar/ and have it "just work" out of the box if you've correctly followed the Getting Started guide.

As it stood yesterday, if you followed the Getting Started guide and followed the demo instructions on the tab bar, you'd see this, with no error in the JS console:

Screen Shot 2019-03-12 at 2 54 14 PM

That's what this PR is all about. I'm glad to see PR #4501 getting merged in. But in fact PR #4501 wouldn't have helped me, precisely because I _did_ follow the self-hosting instructions, and I, personally, would want to self-host any critical app UI icons I need.

Note that the Tab Bar documentation is all dependent on the user following the self-hosting instructions. It asks the user to `npm install @material/tab-bar` and then `import {MDCTabBar} from '@material/tab-bar'` which you can only do if you have some kind of build tool, presumably webpack, converting npm dependencies into libraries that your JS can import.

PR #4501 would not help users who followed the self-hosting instructions Getting Started guide, so I'd want _some_ kind of instructions in the Getting Started guide that will just DTRT for self-hosting users. I claim that The Right Thing in this case is to self-host the icons themselves, and the only public documentation I see on how to do that is to depend on `material-design-icons`.

WDYT? Is there a good way to get the Tab Bar documentation to work right out of the box in a self-hosted environment, including self-hosting the icon?

kfranqueiro: There is unfortunately not really a recommendable self-hosting option right now that I would feel comfortable putting in the getting started guide.

Your options are:

* Use the npm package (which hasn't been updated in a long time)
* Pull down https://fonts.googleapis.com/icon?family=Material+Icons and the associated font resource yourself to self-host (I checked; the latest woff2 is definitely considerably larger than the one provided in the npm package)

At any rate, I've certainly made it known to the fonts team that this is something people are interested in, and I would certainly like to be able to document a more recommendable option in the future.

abhiomkar:
Closing since inactive.

PR #4434, opened 2019-02-19
fix: add missing import to mdc-floating-label/mixins

ALMaclaine:
add missing @material/theme/mixins.scss `@import` to @material/floating-label/mixins.scss


kfranqueiro:
Thanks for spotting this!

PR #4435, opened 2019-02-19
chore: replace JS for TS

moog16:
related #4225

I'm fairly certain I found all instances of `.js` that needed to be replaced to `.ts`. I ignored all demos/tests since we aren't rewriting those just yet.

TODO:
- revisit the `check-imports.js` script


docs/authoring-components.md
```diff
@@ -738,7 +733,7 @@ class MDCNewComponentFoundation extends MDCFoundation {
 ```
 
 The primary purpose of this is so that our components can interoperate with aspects of Google's
-front-end infrastructure, such as Closure Stylesheets' [CSS Class Renaming](https://github.com/google/closure-stylesheets#renaming) mechanism. In addition, it
```
kfranqueiro: This comment is not made irrelevant by the TypeScript conversion.

docs/code/architecture.md
```diff
@@ -29,6 +29,10 @@ Foundation and Adapter. This lets us reuse Foundation code across multiple web
 platforms, e.g. React and Angular, by re-implementing only the Adapter. For now
 we've only implemented a vanilla JavaScript version of the Adapter.
 
+### TypeScript
+
+MDC Web has built our components using TypeScript to increase developer communication and spped. It's proven to catch bugs before releasing new versions of our components, and provides another layer of safety via types. TypeScript helps the Open Source Contributors and maintainers alike with adding features and fixing bugs.
```
kfranqueiro: MDC Web components are built using ...

docs/contribute/code_review.md
```diff
@@ -24,7 +24,8 @@ of having too many open pull requests.
 ### Check The Obvious
 
 *  Do tests pass?
-*  Does the demo work?
+*  Are there coresponding screenshot tests?
```
kfranqueiro: corresponding

docs/contribute/code_review.md
```diff
@@ -38,6 +39,7 @@ APIs:
 *  CSS classes
 *  DOM structure
 *  Utility JavaScript Functions
+*  TypeScript Declaration Files
```
kfranqueiro: Why was this added to this list? Everything else in this list is manually engineered; aren't d.ts automatically generated?

docs/getting-started.md
```diff
@@ -149,7 +149,7 @@ And open http://localhost:8080 in a browser. You should see a blue “Hello Worl
 Now that you have webpack configured to compile Sass into CSS, let's include the Sass files for the Material Design button. First install the Node dependency:
 
 ```
-npm install --save-dev @material/button
```
kfranqueiro: Don't remove `--save-dev`. I'm pretty sure I stopped someone from doing this before (or had them revert it).

docs/getting-started.md
```diff
@@ -230,38 +230,69 @@ Now run `npm start` again and open http://localhost:8080. You should see a Mater
...
 - [babel-preset-es2015](https://www.npmjs.com/package/babel-preset-es2015): Preset for compiling es2015
 - [babel-plugin-transform-object-assign](http://npmjs.org/package/babel-plugin-transform-object-assign), for IE 11 support
 
+> NOTE: You can decide to leave out `ts-loader` and `typescript` if you do not want to use TypeScript. However we do recommend you using TypeScript as it provides all the features of type safety. You can also use other TypeScript loaders such as [Awesome TypeScript Loader](https://www.npmjs.com/package/awesome-typescript-loader). We chose ts-loader because it is what we use to build our components.
+
 You can install all of them by running this command:
 
 ```
-npm install --save-dev babel-core@6 babel-loader@7 babel-preset-es2015 babel-plugin-transform-object-assign
+npm install --save-dev babel-core@6 babel-loader@7 babel-preset-es2015 babel-plugin-transform-object-assign ts-loader@3.5.0
```
kfranqueiro: Should typescript also be installed locally? It's not in this list...

Also, babel-plugin-transform-object-assign should no longer be needed (we should test that - I'm pretty sure all instances of Object.assign should no longer exist on the feat/typescript branch)

docs/getting-started.md
```diff
@@ -230,38 +230,69 @@ Now run `npm start` again and open http://localhost:8080. You should see a Mater
...
+
+We also recommend installing typescript globally to run `tsc` commands from your CLI:
+
+```
+npm install -g typescript # this allows you to run `tsc --init` in a subsequent step
+```
+
+### Step 4: TypeScript setup
+
+Now we need to setup TypeScript through your `tsconfig.json` file. The easiest way is to use the `TypeScript CLI` known as `tsc`. If you install it globally found in the [Step 3](#step-3-installing-javascript-packages), you can run the follow:
```
kfranqueiro: follow -> following

docs/getting-started.md
```diff
@@ -230,38 +230,69 @@ Now run `npm start` again and open http://localhost:8080. You should see a Mater
...
+
+
```
kfranqueiro: Remove extra newline

docs/getting-started.md
moog16: Why are we telling people to add their non-build step deps to devDeps?  `dependencies` are modules which are also required at runtime.

docs/getting-started.md
moog16: Hmm...ya i suppose it should. Below I also recommend installing globally, but it should be included in their deps in package.json

docs/contribute/code_review.md
moog16: true - removing!

docs/authoring-components.md
moog16: ok I'll add this back

docs/closure-compiler.md
```diff
@@ -1,179 +0,0 @@
-<!--docs:
```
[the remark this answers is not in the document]
moog16: Well if we want to keep this as a best practices for writing foundations/adapter/components with TypeScript that makes sense. But a lot of this seems to be covered in the `authoring-components.md`. What do you think?

docs/importing-js.md
```diff
@@ -26,12 +26,20 @@ of your built assets, you will want to explicitly reference the package's `index
...
+
 Note that in this case, you must ensure your build toolchain is configured to process and transpile MDC Web's modules
 as well as your own. You will also need to include babel's
 [`transform-object-assign`](https://www.npmjs.com/package/babel-plugin-transform-object-assign) plugin for IE 11 support.
 
 See the [Getting Started guide](getting-started.md) for more details on setting up an environment.
 
+#### TypeScript
+
+If you are using TypeScript, MDC Web's packages also include `.d.ts` files for your consumption. Most of the time you shouldn't need to reference these, as the TypeScript compiler should automatically pick them up. MDC Web has also defined the `types` property found in `package.json`, for your convinence. There is a bundled `.d.ts` file found under the `dist` directory that maps to the respective UMD module. There are corresponding `.d.ts` files for each foundation/component/adapter/etc. within the package.
```
acdvorak: Spelling: `convinence`. Also, what does "for your convenience" mean? Why is the `types` property important; what does it do?

docs/importing-js.md
```diff
@@ -26,12 +26,20 @@ of your built assets, you will want to explicitly reference the package's `index
...
+
+> NOTE: Intentionally we did not include the `.ts` source files in our packages. Everyone can use the `.d.ts` files and transpiled `.js` (in ES5 or ES2015+ format) as `.js` files are universally accepted.
```
acdvorak: Reword first sentence and add an explanation:

```markdown
NOTE: We intentionally omit `.ts` source files in our packages because <REASON>. ...
```

docs/contribute/code_review.md
```diff
@@ -24,7 +24,8 @@ of having too many open pull requests.
 ### Check The Obvious
 
 *  Do tests pass?
-*  Does the demo work?
+*  Are there corresponding screenshot tests?
```
acdvorak: Link to `test/screenshot/` directory

docs/code/architecture.md
```diff
@@ -29,6 +29,10 @@ Foundation and Adapter. This lets us reuse Foundation code across multiple web
 platforms, e.g. React and Angular, by re-implementing only the Adapter. For now
 we've only implemented a vanilla JavaScript version of the Adapter.
 
+### TypeScript
+
+MDC Web components are built using TypeScript to increase developer communication and spped. It's proven to catch bugs before releasing new versions of our components, and provides another layer of safety via types. TypeScript helps the Open Source Contributors and maintainers alike with adding features and fixing bugs.
```
acdvorak:
```markdown
MDC Web components are written in [TypeScript](https://www.typescriptlang.org/)
to increase developer velocity and reduce errors. Our npm releases include
ES5 and ES2015+ JavaScript bundles, as well as `.d.ts` typing declaration files
for TypeScript users. See [Importing JS](../importing-js.md) for more information.
```

docs/authoring-components.md
```diff
@@ -760,16 +755,11 @@ Concretely:
...
-
-#### Closure Compatibility
 
-> NOTE: Our currently existing components are in the process of being made compatible with closure.
+#### TypeScript Compatibility
 
-All core MDC Web components must be fully compatible with the Google Closure Compiler using its
-advanced compilation mechanisms. We've provided a thorough explanation of this, as well as
-conventions, examples, and common closure patterns you may not be used to, in our [closure compiler documentation](./closure-compiler.md).
+All core MDC Web components must be fully compatible with the [TypeScript](https://www.typescriptlang.org/index.html). We've provided a list of requirements and further explanation of meeting the TypeScript compatibility, as well as
```
acdvorak:
```markdown
All core MDC Web components must be fully compatible with strict-mode [TypeScript](https://www.typescriptlang.org/).
```

docs/authoring-components.md
```diff
@@ -738,7 +733,7 @@ class MDCNewComponentFoundation extends MDCFoundation {
 ```
 
 The primary purpose of this is so that our components can interoperate with aspects of Google's
-front-end infrastructure, such as Closure Stylesheets' [CSS Class Renaming](https://github.com/google/closure-stylesheets#renaming) mechanism. In addition, it
+front-end infrastructure, such as Closure Stylesheets' [CSS Class Renaming](https://github.com/google/closure-stylesheets#renaming) mechanism.. In addition, it
```
acdvorak: Remove duplicate `.`

docs/authoring-components.md
```diff
@@ -478,9 +472,10 @@ A typical component within our codebase looks like so:
 packages
   ├── mdc-component
       ├── README.md # The component's README
-      ├── constants.js # The component's cssClasses/strings/numbers constants
-      ├── foundation.js # The component's foundation class
-      ├── index.js # The file that contains the vanilla component class, as well as exports the foundation
+      ├── adapter.ts # The adapter interface implemented by the component
+      ├── constants.ts # The component's cssClasses/strings/numbers constants
+      ├── foundation.ts # The component's foundation class
+      ├── index.ts # The file that contains the vanilla component class, as well as exports the foundation
```
acdvorak: Add `component.ts` and update this comment to reflect that `index.ts` only re-exports the other files.

docs/authoring-components.md
```diff
@@ -250,39 +246,35 @@ be used to build out lint tools to enforce proper adapter shape. This example sh
...
   static get defaultAdapter() {
     return {
       getAttr: (/* attr: string */) => /* string */ '',
-      setAttr: (/* attr: string, value: string */) => {},
-      addClass: (/* className: string */) => {},
-      removeClass: (/* className: string */) => {},
-      setToggleColorTextContent: (/* textContent: string */) => {},
-      registerInteractionHandler: (/* type: string, handler: EventListener */) => {},
-      deregisterInteractionHandler: (/* type: string, handler: EventListener */) => {}
+      setAttr: (/* attr: string, value: string */) => undefined,
```
acdvorak: Replace Closure Doc comments with TypeScript

docs/authoring-components.md
```diff
@@ -231,14 +232,9 @@ methods. We can now take these fake methods and transform them into an adapter i
...
-| SOMEHOW_REGISTER_INTERACTION_HANDLER(type: string, handler: EventListener) | registerInteractionHandler(type: string, handler: EventListener) |
-| SOMEHOW_DEREGISTER_INTERACTION_HANDLER(type: string, handler: EventListener) | deregisterInteractionHandler(type: string, handler: EventListener) |
 
-> Note: It is a convention in our code base to use the terms `registerInteractionHandler` and
-`deregisterInteractionHandler` as adapter methods for adding/removing generic event listeners. We do
-this because we feel it more semantic in that most components are only interested in interactivity.
-However, feel free to call these methods `{add,remove}EventListener` or anything else you would
-like.
+> Note: We have changed our convention of using `registerInteractionHandler` and
+`deregisterInteractionHandler` as adapter methods for adding/removing generic event listeners. Instead please manage events in the component layer. This is because event systems can change from framework/environment (ie. React Synthetic Events).
```
acdvorak: Change this:

```markdown
Instead please manage events in the component layer.
This is because event systems can change from framework/environment
```

to this:

```markdown
Event handling varies significantly across frameworks
(e.g., [React Synthetic Events](https://reactjs.org/docs/events.html)),
so we recommend managing events in the component layer.
```

docs/authoring-components.md
```diff
@@ -231,14 +232,9 @@ methods. We can now take these fake methods and transform them into an adapter i
...
 | SOMEHOW_UPDATE_TOGGLE_COLOR_TEXT_CONTENT(textContent: string) | setToggleColorTextContent(textContent: string) |
-| SOMEHOW_REGISTER_INTERACTION_HANDLER(type: string, handler: EventListener) | registerInteractionHandler(type: string, handler: EventListener) |
-| SOMEHOW_DEREGISTER_INTERACTION_HANDLER(type: string, handler: EventListener) | deregisterInteractionHandler(type: string, handler: EventListener) |
 
-> Note: It is a convention in our code base to use the terms `registerInteractionHandler` and
-`deregisterInteractionHandler` as adapter methods for adding/removing generic event listeners. We do
-this because we feel it more semantic in that most components are only interested in interactivity.
-However, feel free to call these methods `{add,remove}EventListener` or anything else you would
-like.
+> Note: We have changed our convention of using `registerInteractionHandler` and
```
acdvorak: It would be clearer to phrase this more strongly:

```markdown
> Note: We no longer recommend using `registerInteractionHandler` and
```

docs/authoring-components.md
acdvorak: Nit: Change "ie." to "e.g.," ([Grammar Girl](https://www.quickanddirtytips.com/education/grammar/ie-versus-eg) to the rescue!)

docs/importing-js.md
moog16: I read this doc again: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

It reads, "Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so."

I've been reading this incorrectly, and its actually required in our case. I'll update the wording.

docs/authoring-components.md
moog16: I replaced it with:
```
      getAttr: (attr: string) => '',
      setAttr: (attr: string, value: string) => undefined,
      addClass: (className: string) => undefined,
      removeClass: (className: string) => undefined,
      setToggleColorTextContent: (textContent: string) => undefined,
      registerInteractionHandler: (type: string, handler: EventListener) => undefined,
      deregisterInteractionHandler: (type: string, handler: EventListener) => undefined
```

Although in our actual implementation we don't have unused arguments. Should I just remove them to look like:

```
      getAttr: () => '',
      setAttr: () => undefined,
      addClass: () => undefined,
      removeClass: () => undefined,
      setToggleColorTextContent: () => undefined,
      registerInteractionHandler: () => undefined,
      deregisterInteractionHandler: () => undefined,
```

acdvorak:
We can probably update `check-imports.js` in PR #4449.

docs/authoring-components.md
acdvorak: I would probably leave them to make it clear that they exist, even if we're not using them.

docs/importing-js.md
```diff
@@ -26,12 +26,20 @@ of your built assets, you will want to explicitly reference the package's `index
 import {MDCFoo, MDCFooFoundation} from '@material/foo/index';
 ```
 
+Certain build tools will often pick up on the use of `package.json`'s `module` property, which points to the ES2015+ `index.js` source code. If you are using [Webpack](https://webpack.js.org/) or [Rollup](https://rollupjs.org/guide/en), then you will not need to reference the `index.js` file, and can continue to use the `@material/foo` import paths.
```
acdvorak:
```markdown
Certain build tools will detect the `module` property in `package.json`,
which points to ES2015+ source code in an `index.js` file.
If you're using [Webpack](https://webpack.js.org/) or
[Rollup](https://rollupjs.org/guide/en), you do not need to
reference `/index` directly, and can continue to use the shorter
`@material/foo` import path syntax.
```

(Reformat the above into one line)

docs/authoring-components.md
```diff
@@ -316,8 +310,16 @@ Since this component is a vanilla component, it should be modeled after the vani
 favors getters and setters to implement its functionality (think `checked`, `disabled`, etc.). Our
 adapter is extremely straightforward as we can simply repurpose the methods we started out with.
 
-```js
+```ts
 class RedblueToggle extends MDCComponent {
+  initialize() {
+    this.root.addEventListener('click', this.foundation_.handleClick);
```
kfranqueiro: These can use `this.listen` / `this.unlisten` here now that the class extends MDCComponent

docs/authoring-components.md
```diff
@@ -478,9 +474,11 @@ A typical component within our codebase looks like so:
 packages
   ├── mdc-component
       ├── README.md # The component's README
-      ├── constants.js # The component's cssClasses/strings/numbers constants
-      ├── foundation.js # The component's foundation class
-      ├── index.js # The file that contains the vanilla component class, as well as exports the foundation
+      ├── adapter.ts # The adapter interface implemented by the component
```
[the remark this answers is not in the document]
kfranqueiro: Were some of Andy's comments not resolved?

docs/closure-compiler.md
```diff
@@ -1,179 +0,0 @@
...
-  parameterized by `MDCRippleFoundation`, which - as shown above - is itself parameterized by
-  `MDCRippleAdapter`. Thus, the `MDCRipple` class is declared as
-  `class MDCRipple extends MDCComponent<MDCRippleFoundation>`.
-
-## MDC Web Closure Conventions
-
-The following guidelines outline the general conventions for writing closure-friendly code for MDC Web.
-This section should contain most - if not all - of what you need to get up and running writing components in our codebase.
-
-### All `import` statements must _not_ use re-exported modules.
```
kfranqueiro: This should effectively be moved to best_practices.md in the form of a heading explaining that in the context of component packages, only the index and component files are allowed to reference `index`, because wrappers shouldn't need to pull in `index` or `component`

docs/closure-compiler.md
```diff
@@ -1,179 +0,0 @@
...
-interface EventDataType {
-  foo: string;
-  bar: number;
-}
-```
-<!--{% endraw %} -->
-
-Using this convention allows us to write tooling around handling these expressions, such as lint rule exceptions.
-
-### For key/value maps, use index signatures where possible
```
kfranqueiro: This should be moved to best_practices.md; I added this specifically for TypeScript

docs/importing-js.md
kfranqueiro: It looks like this comment hasn't been applied.

Furthermore, index.js isn't ES2015+ source anymore. It's ES5 source in an ES module.

acdvorak: How would you phrase "ES5 source in an ES module" in context?

docs/authoring-components.md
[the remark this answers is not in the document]
moog16: I'm just about to fix these. I didn't see this yesterday.

docs/importing-js.md
moog16: The 2 index.js files of the npm package are:

```js
dist/index.js (ES5) -- UMD
index.js (ES2015+) -- ESModule
```

Isn't what Andy wrote correct?

kfranqueiro: The distinction is that prior to the TS conversion, *.js was ES modules _with ES2015+ syntax_, thus requiring babel and additionally the object assign polyfill. After the TS conversion, *.js are ES modules _with ES5 syntax_. I believe that means you can get away with just webpack and not babel at all?

In terms of phrasing, something like: "which points to an ES module otherwise containing only ES5 syntax". We could link something to Rollup's pkg.module definition at https://github.com/rollup/rollup/wiki/pkg.module if we want.

docs/closure-compiler.md
moog16: Is this line still accurate? 

This is an unfortunate side-effect of how [closure's module naming mechanism works](https://github.com/google/closure-compiler/issues/2257).

docs/closure-compiler.md
moog16: Re both closure-compiler comments: I ended up adding this entire block:
```

#### All `import` statements must _not_ use re-exported modules

Only the `index.ts` or `component.ts` files are allowed to reference from other component packages' `index.ts`. This is because wrapping libraries only use `foundation` and `adapter`, so we should decouple the `component`.

```ts
// BAD
import {MDCFoundation} from '@material/base';
// GOOD
```

#### All adapters must be defined as interfaces

Each adapter must be defined within an `adapter.ts` file in the component's package directory.
All methods should contain a summary of what they should do. This summary should be
copied over to the adapter API documentation in our README. This will facilitate future endeavors
to potentially automate the generation of our adapter API docs. _Note that this replaces the
inline comments present in the methods within `defaultAdapter`_.

 ```ts
// adapter.ts
export interface MDCComponentAdapter {
  /**
   * Adds a class to the root element.
   */
  addClass(className: string): void;
   /**
   * Removes a class from the root element.
   */
  removeClass(className: string): void;
}
```

#### Foundation classes must extend `MDCFoundation`

 Foundations must extend `MDCFoundation` parameterized by their respective adapter. The
`defaultAdapter` must return an object with the correct adapter shape.

```ts
// foundation.ts
import MDCComponentAdapter from './adapter';
export class MDCComponentFoundation extends MDCFoundation<MDCComponentAdapter> {
  static get defaultAdapter(): MDCComponentAdapter {
    return {
      addClass: (className: string) => undefined,
      removeClass: (className: string) => undefined,
    };
  }
}
```

#### Component classes must extend `MDCComponent`

Components must extend `MDCComponent` parameterized by their respective foundation.

```ts
// index.ts
import MDCComponentFoundation from './foundation';
export class MDCAwesomeComponent extends MDCComponent<MDCComponentFoundation> {
  getDefaultFoundation(): MDCComponentFoundation {
    return new MDCComponentFoundation({
      addClass: (className: string) => this.root_.classList.add(className),
      removeClass: (className: string) => this.root_.classList.remove(className),
    });
  }
}
```

### For key/value maps, use index signatures where possible

Index signatures are useful for homogeneous maps of key/value pairs, while interfaces are useful for specific object signatures.
Both are more specific than just typing something as `object`, and are thus preferable for their respective use cases.

```ts
static get strings(): {[key: string]: string} {
  ...
}
```

docs/importing-js.md
moog16: @kfranqueiro I just checked by running `npm run build:esmodules`, so I do think you're correct.

Also I just noticed that we don't use `Object.assign` anywhere in our `packages/` directory. I do not think you need babel for either of our UMD or ESM packages.

kfranqueiro: RE your last point, that's what I mean. Previously we were relying on the object assign polyfill to generate our UMD (you were never expected to need babel to consume those). Now we don't need to use that polyfill for our build process, and neither does anyone else using us within their own.

moog16: In getting-started.md there is this dep:
```
- [babel-plugin-transform-object-assign](http://npmjs.org/package/babel-plugin-transform-object-assign), for IE 11 support
```

I am going to remove it

PR #4437, opened 2019-02-20
docs: update all imports after UMD change

ALMaclaine:
update all ES6 import statments to include the index to make all docs and consistent with UMD dist update


ALMaclaine:
TS update makes much of this obsolete.

PR #4443, opened 2019-02-23
chore: Fix regression in `MDCChipFoundation.getDimensions()`

acdvorak:
PR #4332 inadvertently broke short-circuit logic and evaluation order.

The change caused incorrect ripple size calculations on filter chips without leading icons, which was caught by internal screenshot tests.

### DIFF
--- a/packages/mdc-chips/chip/foundation.ts
+++ b/packages/mdc-chips/chip/foundation.ts
@@ -94,19 +94,24 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
   }
 
   getDimensions(): ClientRect {
-    const rootRect = this.adapter_.getRootBoundingClientRect();
-    const checkmarkRect = this.adapter_.getCheckmarkBoundingClientRect();
+    const getRootRect = () => this.adapter_.getRootBoundingClientRect();
+    const getCheckmarkRect = () => this.adapter_.getCheckmarkBoundingClientRect();
 
     // When a chip has a checkmark and not a leading icon, the bounding rect changes in size depending on the current
     // size of the checkmark.
-    if (!this.adapter_.hasLeadingIcon() && checkmarkRect !== null) {
-      // The checkmark's width is initially set to 0, so use the checkmark's height as a proxy since the checkmark
-      // should always be square.
-      const width = rootRect.width + checkmarkRect.height;
-      return {...rootRect, width};
-    } else {
-      return rootRect;
+    if (!this.adapter_.hasLeadingIcon()) {
+      const checkmarkRect = getCheckmarkRect();
+      if (checkmarkRect) {
+        const rootRect = getRootRect();
+        const height = rootRect.height;
+        // The checkmark's width is initially set to 0, so use the checkmark's height as a proxy since the checkmark
+        // should always be square.
+        const width = rootRect.width + checkmarkRect.height;
+        return {...rootRect, width, height};
+      }
     }
+
+    return getRootRect();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/mdc-chips/chip/foundation.ts
```diff
@@ -94,19 +94,24 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
...
+    if (!this.adapter_.hasLeadingIcon()) {
+      const checkmarkRect = getCheckmarkRect();
+      if (checkmarkRect) {
+        const rootRect = getRootRect();
+        const height = rootRect.height;
+        // The checkmark's width is initially set to 0, so use the checkmark's height as a proxy since the checkmark
+        // should always be square.
+        const width = rootRect.width + checkmarkRect.height;
+        return {...rootRect, width, height};
+      }
```
patrickrodee: There's a missing return in the case of checkmarkRect being falsy. However, tests still seem to be passing. Do we still need the falsy check for `checkmarkRect`?

packages/mdc-chips/chip/foundation.ts
```diff
@@ -94,19 +94,24 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
...
-      const width = rootRect.width + checkmarkRect.height;
-      return {...rootRect, width};
-    } else {
-      return rootRect;
+    if (!this.adapter_.hasLeadingIcon()) {
+      const checkmarkRect = getCheckmarkRect();
+      if (checkmarkRect) {
+        const rootRect = getRootRect();
+        const height = rootRect.height;
+        // The checkmark's width is initially set to 0, so use the checkmark's height as a proxy since the checkmark
```
patrickrodee: I'm confused by what "use the checkmark's height as a proxy" means. Is this comment still relevant?

packages/mdc-chips/chip/foundation.ts
acdvorak: The missing `return` statement is logically equivalent to the original syntax; I just split a single `if(...)` statement into two for better readability:

https://github.com/material-components/material-components-web/blob/70672f12ff19591c53fef6876c059c854f9cb816/packages/mdc-chips/chip/foundation.js#L116-L129

`checkmarkRect` can be `null` because `adapter.getCheckmarkBoundingClientRect()` may return `null`, so I think we probably need to keep the falsy check (unless I'm misunderstanding).

patrickrodee: I GET IT NOW. My bad... It's been a long day.

packages/mdc-chips/chip/foundation.ts
acdvorak: I think it has to do with how the checkmark animates from `width: 0` to `width: auto` (or w/e).

Since the checkmark starts at `width: 0`, we can't query its `width` to figure out how wide it's eventually going to be when it's done animating.

However, we _can_ assume that the checkmark will have square dimensions, so querying its `height` is the same as querying its (future) `width`.

Does that make sense?

packages/mdc-chips/chip/foundation.ts
acdvorak: I hear that 😴

packages/mdc-chips/chip/foundation.ts
patrickrodee: Well I understand now. I do think the comment could be clarified though.

>The checkmark is a square meaning the client rect's width and height are equivalent. However, the checkbox is initially hidden by setting the width to 0. To account for a potential 0 width checkbox, we use the checkbox's height when adding it to the root client rect's width.

patrickrodee: Use your comment. That's perfect.

PR #4448, opened 2019-02-25
style: Add explanations to `tslint:disable` comments

acdvorak:
All `tslint:disable:object-literal-sort-keys` comments now include a description that explains why the linter rule needs to be disabled. For the most part, it's to ensure consistent ordering of methods between adapter interfaces and their implementations in the foundation and component.

This PR also reorders a few properties in `MDCCheckbox` and `MDCRadio` for consistency.

Refs #4225

### DIFF
--- a/packages/mdc-radio/component.ts
+++ b/packages/mdc-radio/component.ts
@@ -59,15 +59,6 @@ export class MDCRadio extends MDCComponent<MDCRadioFoundation> implements MDCRip
     return this.ripple_;
   }
 
-  private get nativeControl_(): HTMLInputElement {
-    const {NATIVE_CONTROL_SELECTOR} = MDCRadioFoundation.strings;
-    const el = this.root_.querySelector<HTMLInputElement>(NATIVE_CONTROL_SELECTOR);
-    if (!el) {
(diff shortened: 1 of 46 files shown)


packages/mdc-radio/component.ts
```diff
@@ -64,6 +59,20 @@ export class MDCRadio extends MDCComponent<MDCRadioFoundation> implements MDCRip
...
+  private get nativeControl_(): HTMLInputElement {
+    const {NATIVE_CONTROL_SELECTOR} = MDCRadioFoundation.strings;
+    const el = this.root_.querySelector<HTMLInputElement>(NATIVE_CONTROL_SELECTOR);
+    if (!el) {
+      throw new Error(`Radio component requires a ${NATIVE_CONTROL_SELECTOR} element`);
+    }
+    return el;
+  }
+
+  // Public visibility for this property is required by MDCRippleCapableSurface.
```
kfranqueiro: Why is this down below something that's private now?

acdvorak: I was trying to make radio and checkbox consistent.
Moved `private get nativeControl_()` to the bottom of both files.

PR #4449, opened 2019-02-26
chore: Rewrite lint:imports & add component dep check

acdvorak:
This PR refactors `npm run lint:imports` and adds a new check for component-dependent imports in non-component files.

Example error messages:

```
ERROR: mdc-textfield/foundation.ts: Import path '@material/base/index'
is only allowed in component.ts and index.ts files. Please import a more specific file
(e.g., 'adapter', 'foundation', 'types').
This keeps byte sizes smaller for wrapper libraries that only use our foundations.

ERROR: mdc-textfield/component.ts: Import path '@material/base'
should point directly to a specific TypeScript file.
E.g.: import '@material/base/index'

ERROR: mdc-textfield/index.ts: Import path './character-counter'
should point directly to a specific TypeScript file.
E.g.: import './character-counter/index'

ERROR: mdc-textfield/component.ts: Import path '@material/dom/ponyfill'
references NPM module '@material/dom', but that module
is not listed in package.json dependencies.

ERROR: mdc-textfield/index.ts: Import path '../mdc-textfield/helper-text/index'
appears to be a relative path to an MDC package. Please use a @material import instead.
E.g.: import '@material/foo/types' instead of import '../mdc-foo/types'.

ERROR: mdc-textfield/foundation.ts: Import path
'/Users/advorak/dev/mdc-web/packages/mdc-textfield/adapter'
appears to be an absolute path. Please use a relative path instead.
```

screenshot


scripts/custom-linter.js
```diff
@@ -0,0 +1,167 @@
...
+    let relativePath = path.relative(inputFileDir, resolvedPath);
+    if (!relativePath.startsWith('./') && !relativePath.startsWith('../')) {
+      relativePath = `./${relativePath}`;
+    }
+    const pointsToTSFile = relativePath === `${importPath}.ts`;
+    if (!pointsToTSFile) {
+      logLinterViolation(
+        inputFilePath,
+        // eslint-disable-next-line max-len
+        `Import path '${importPathAnsi}' should point directly to a specific TypeScript file. E.g.: import '${relativePath.replace('.ts', '')}'`
```
kfranqueiro: I'm confused... doesn't this block get executed when `relativePath` _doesn't_ have `.ts`? So what does replacing `.ts` with nothing accomplish?

scripts/custom-linter.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  const importPathIsComponent = importPath.endsWith('/index') || importPath.endsWith('/component');
+  const inputFileIsComponent = inputFilePath.endsWith('/index.ts') || inputFilePath.endsWith('/component.ts');
+  if (importPathIsComponent && !inputFileIsComponent) {
+    logLinterViolation(
+      inputFilePath,
+      // eslint-disable-next-line max-len
+      `Import path '${importPathAnsi}' is only allowed in component.ts and index.ts files. Please import a more specific file (e.g., adapter.ts, foundation.ts, or types.ts). This keeps bite sizes smaller for wrapper libraries that only use our foundations.`
```
kfranqueiro: bite -> byte

(don't code hungry)

scripts/custom-linter.js
acdvorak: `resolvedPath` is the full, absolute path to the TS file, including `.ts` extension.

`relativePath` is the relative version of `resolvedPath`, so it still has a `.ts` extension.

I renamed the variables to make this clearer. Sorry for the confusion!

scripts/custom-linter.js
acdvorak: cookie monster eating

scripts/custom-linter.js
kfranqueiro: My point was this statement is within an `if (!pointsToTSFile)` block, which I had interpreted (based on the previous line of code before the condition) as meaning that the filename doesn't end with `.ts`, which means there's no `'.ts'` to replace with `''` anyway. Did I misinterpret the line before this condition?

acdvorak: Gotcha. I refactored the code and was able to remove this entire `if` block. A different `if` statement now handles the same use case in a more general way.

PR #4454, opened 2019-02-26
fix(checkbox): move checkbox ripple styles into separate mixin

mmalerba:


packages/mdc-checkbox/mdc-checkbox.scss
```diff
@@ -20,5 +20,8 @@
 // THE SOFTWARE.
 //
 
+@import "@material/ripple/mixins";
 @import "./mixins";
+@include mdc-ripple-common;
 @include mdc-checkbox-core-styles;
+@include mdc-checkbox-ripple;
```
mmalerba: Is it ok to have this file include 3 mixins? An alternative approach would be to have mixins like: `mdc-checkbox-without-ripple` `mdc-checkbox-ripple` and then `mdc-checkbox-core-styles` like this:

```scss
@mixin mdc-checkbox-core-styles($query: mdc-feature-all()) {
  @include mdc-ripple-common($query);
  @include mdc-checkbox-without-ripple($query);
  @include mdc-checkbox-ripple($query);
}
```

moog16: I think this is a good change, as this doesn't look like it will break any current users using `mdc-checkbox/mdc-checkbox.scss`.

With Mat-Angular, I anticipate that we wouldn't use this `mdc-checkbox.scss` file, but instead use the 3 mixins you have here. Is that correct? We'd also then have to remove `mdc-ripple-common` from button, checkbox, list, menu, radio, etc.

mmalerba: Good the way I currently have it or splitting it as mentioned in the comment above?

Yeah Angular Material will just use the mixins directly to avoid including parts we don't need (i.e. all of the ripple-related stuff).

That's a good point about all of the other components, should I refactor them in this PR or separate ones?

moog16: I think it is good the way you have it now. Would we want to have a checkbox without ripple? Ripple is not only a UI indicator, but an indicator for a11y as well - and want to promote having it.

I think we should just start here and build up as we go. We'll want to develop the pattern first. I just want to see where these changes sit in the bigger picture.

mmalerba: Yeah I guess my question is really: If someone just does `@include mdc-checkbox-core-styles` in their app, should that include ripple styles? The way I have it set up right now it doesn't. Though you could say that using the mixins is already "advanced mode" compared to the "simple mode" of just importing the main file, so maybe its ok to expect users of the mixins to know that they have to `@include` all three.

moog16: I think we have a wide range of users in terms of their ability. Mixins definitely fall in the "advanced users". I think its ok to expect that advanced users know that they have to `@include` the 3 mixins, but it should document that in the Readme. We do that in several components. 

One day though I would love to have autogenerated docs (eg. jsdoc).

moog16: I was thinking more about the separating the other components. Maybe we should just create a feature branch and do them all at once. @kfranqueiro any thoughts?

kfranqueiro: I'll preface this by saying that the vast majority of our basic clients will just import `mdc-checkbox.scss` and so the mixins won't matter.

However, I do think that it becomes inconsistent if suddenly some components' `-core-styles` mixins are no longer actually emitting _all_ of the baseline styles for the given MDC Web component. As far as MDC Web is concerned, ripple styles _are_ a part of its components' baseline styles.

I realize that the alternative probably makes for a slightly more obtuse API. The `-without-ripple` and `-ripple` are borderline-internal but need to be exposed specifically for framework wrappers interested in substituting in their own existing ripple implementation. We might want to comment as such above these two mixins to advocate just using `-core-styles` in typical use cases.

RE one or separate PRs... I'm inclined to think one PR for all, from the perspective of trying this out on all affected components out of the ones converted, so that we can work out any kinks before merging some but not others and then possibly needing to backtrack on them (our next release should be Monday). I think this would include button, card, checkbox, radio, switch, and list.

moog16: What if we named them `-core-styles` and `-core-styles-without-ripple`. `-core-styles` would include the `-core-styles-without-ripple` and include `ripple` mixins? And then like you said, advocate for using `core-styles`

packages/mdc-card/_mixins.scss
```diff
@@ -34,6 +34,13 @@
 //
 
 @mixin mdc-card-core-styles($query: mdc-feature-all()) {
+  @include mdc-card-without-ripple($query);
```
mmalerba: it looks like card was not already including `mdc-ripple-common` is that intentional?

packages/mdc-checkbox/mdc-checkbox.scss
mmalerba: Ok, updated to maintain the current behavior or the `*-core-styles` mixins. Also added changes for the other components mentioned above.

packages/mdc-card/_mixins.scss
kfranqueiro: That was probably technically an oversight that went unnoticed because users were already using something else that ended up including it (e.g. button).

test/scss/feature-targeting.scss
```diff
@@ -24,19 +24,25 @@
   @include mdc-button-shape-radius(0, $query: mdc-feature-any());
   @include mdc-button-horizontal-padding(0, $query: mdc-feature-any());
   @include mdc-button-outline-width(0, $query: mdc-feature-any());
+  @include mdc-button-ripple($query: mdc-feature-any());
```
kfranqueiro: IIUC these don't need to be here because `*-core-styles` already exercise them indirectly. Do you have a strong opinion on them being here for consistency's sake?

packages/mdc-button/_mixins.scss
```diff
@@ -33,11 +33,17 @@
 @import "./variables";
 
 @mixin mdc-button-core-styles($query: mdc-feature-all()) {
+  @include mdc-ripple-common($query);
```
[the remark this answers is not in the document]
mmalerba: Yeah that makes sense, I'll update them

test/scss/feature-targeting.scss
mmalerba: I've been adding all public mixins, that way we don't have to think about which ones are included by other public mixins already. It doesn't hurt since it should all boil down to no styles emitted in the end

mmalerba:
updated, PTAL.

Should also mention that we should be sure to run screenshot tests on this one. It moves the ripple styles to after the non-ripple styles instead of intermixed. We should verify that it doesn't cause any unexpected changes

kfranqueiro:
Yeah, I took a look at a couple of the CSS diffs too, to make sure I could identify everything.

Report came back with no changes: https://storage.googleapis.com/mdc-web-screenshot-tests/kfranqueiro/2019/03/06/17_36_49_241/report/report.html

However, we don't have tests for e.g. pressed state. We'll run this against further tests internally after the current in-progress sync with the first feature-targeting changes is finished.

PR #4462, opened 2019-02-28
fix(menu-surface): Fix anchorElement initialization

kfranqueiro:
When Menu Surface was converted to TypeScript, its `anchorElement` field was initialized to `null`. This caused a regression, because any proper setup that occurred in `initialSyncWithDOM` was being steamrolled by this `null` initialization, because that occurs in `constructor` _after_ `super` is called.

### DIFF
--- a/test/unit/mdc-menu-surface/mdc-menu-surface.test.js
+++ b/test/unit/mdc-menu-surface/mdc-menu-surface.test.js
@@ -45,12 +45,28 @@ function getFixture(open, fixedPosition = false) {
   `;
 }
 
-function setupTest(open = false, fixedPosition = false) {
+function getAnchorFixture(menuFixture) {
+  const anchorEl = bel`
+    <div class="mdc-menu-surface--anchor">
+      <button class="mdc-button">Open</button>
+    </div>
+  `;
+  anchorEl.appendChild(menuFixture);
+  return anchorEl;
+}
+
+function setupTest(open = false, fixedPosition = false, withAnchor = false) {
   const root = getFixture(open, fixedPosition);
   const MockFoundationConstructor = td.constructor(MDCMenuSurfaceFoundation);
   const mockFoundation = new MockFoundationConstructor();
-  const component = new MDCMenuSurface(root, mockFoundation);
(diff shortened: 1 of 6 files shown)


test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -45,12 +45,28 @@ function getFixture(open, fixedPosition = false) {
...
+  const anchorEl = bel`
+    <div class="mdc-menu-surface--anchor">
+      <button class="mdc-button">Open</button>
+    </div>
+  `;
+  anchorEl.appendChild(menuFixture);
+  return anchorEl;
+}
+
+function setupTest(open = false, fixedPosition = false, withAnchor = false) {
```
acdvorak: Can we make these arguments an object so we don't have a nebulous call to `setupTest(false, false, true)`?

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -45,12 +45,28 @@ function getFixture(open, fixedPosition = false) {
...
   const mockFoundation = new MockFoundationConstructor();
-  const component = new MDCMenuSurface(root, mockFoundation);
-  return {root, component, mockFoundation};
+
+  const fixtures = {root, mockFoundation};
+  if (withAnchor) {
+    fixtures.anchor = getAnchorFixture(root);
+  }
+
+  fixtures.component = new MDCMenuSurface(root, mockFoundation);
```
acdvorak: It might be a little easier to read if we have a single object literal containing all properties. E.g.:

```js
const anchor = withAnchor ? getAnchorFixture(root) : undefined;
const component = new MDCMenuSurface(root, mockFoundation);
return {root, mockFoundation, anchor, component};
```

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
kfranqueiro: If I modify about 24 other unrelated lines of the file in this PR, technically yes

PR #4467, opened 2019-03-01
fix(text-field): Fixed text field to float when input is autofilled

### LINKED ISSUE #4447: [TextField] bug(text-field, notched-outline): Chrome autofill doesn't work with outline variant
Hi,

there is still a problem with autofilling of credentials in Chrome, when the outlined variant of textfields is used:

grafik

There was already an issue (#137) for the non-outlined variant, which was solved in #180:

https://github.com/material-components/material-components-web/blob/70672f12ff19591c53fef6876c059c854f9cb816/packages/mdc-textfield/mdc-text-field.scss#L134-L138
_(Please note that the filename has changed after the PR.)_

As in the outline variant more classes and combinations of classes are involved for the filled state, there'll be more CSS selectors with `* &:-webkit-autofill`, as well as more JS needed, as far as I can guess without having a deeper look into the library/SCSS files.

0.44.1

It's only a problem with Chrome, Firefox still sends a change event. I can't say anything about the other browsers at the moment as I haven't tested them.

Create a form with a textfield for username and password while using the outlined variant. Then open it in Chrome, type in some credentials, save them with Chrome and reload the page.

Hints:
- Chrome doesn't ask for saving the credentials when it's a local path.
- Do not click around in the page as this will cause the labels to be updated.

abhiomkar:
Fixes #4447

### DIFF
--- a/packages/mdc-textfield/adapter.ts
+++ b/packages/mdc-textfield/adapter.ts
@@ -97,6 +97,9 @@ export interface MDCTextFieldInputAdapter {
    * Deregisters an event listener on the native input element for a given event.
    */
   deregisterInputInteractionHandler<K extends EventType>(evtType: K, handler: SpecificEventListener<K>): void;
+
+  /** Returns true if input element has browser auto-filled value. */
+  hasAutofillValue(): boolean;
 }
 
 export interface MDCTextFieldLabelAdapter {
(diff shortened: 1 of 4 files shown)


abhiomkar:
Superseded by #4469

PR #4470, opened 2019-03-01
feat(theme): Add support for CSS vars with fallback

patrickrodee:
**What this does**
Add the ability to specify a CSS custom property name and fallback for color mixins. Allows clients that can support CSS custom property theming to use that when desired without breaking browsers that do not support CSS custom properties.

**What this does not do**
This does not establish naming conventions for CSS custom properties. Those names are entirely up to the developers who use this feature.

**Usage**
```
.my-custom-button {
  @include mdc-button-fill((
    varname: --my-custom-button-fill,
    fallback: purple,
  ));
}
```

The generated CSS is as follows:
```
.my-custom-button {
  background-color: purple;
  background-color: var(--my-custom-button-fill, purple);
}
```

This CSS allows for browsers that do not support CSS custom properties to default to the fallback (purple) while still supporting an overridable CSS custom property `--my-custom-button-fill`.


abhiomkar:
I think this is a great addition to add support for CSS custom property in overriding component styles! 👍 

I've a suggestion in the way we use the mixin, instead of changing the style argument structure I added a way to auto detect the presence of CSS custom property and extract the fallback color automatically.

```scss

// theme/_mixins.scss

@mixin mdc-theme-css-var($property, $var, $important: false, $edgeOptOut: false) {
  @if is-css-var($var) {
    $fallback-style: get-fallback-style($var); // Extracts fallback color from css custom property.
    @include mdc-theme-prop($property, $fallback-style, $important: $important, $edgeOptOut: $edgeOptOut);
    @include mdc-theme-prop($property, $var, $important: $important, $edgeOptOut: $edgeOptOut);
  } @else {
    @include mdc-theme-prop($property, $var, $important: $important, $edgeOptOut: $edgeOptOut)
  }
}

// button/_mixins.scss

@mixin mdc-button-container-fill-color($color) {
  &:not(:disabled) {
    @include mdc-theme-css-var(background-color, $color, $edgeOptOut: true);
  }
}

// override.scss

.my-custom-button {
  @include mdc-button-container-fill-color(var(--my-custom-button-color, purple));
}
```

WDYT?

patrickrodee:
@abhiomkar I originally thought about that but felt the results were a little too "magic." My intention was to make it very clear that a fallback value had to be specified and what role it served. The suggested change also has side-effects for anyone already passing `var(--custom-prop, purple)` to a mixin. I was hoping to avoid any side-effects with this change and make it purely opt-in.

packages/mdc-theme/_mixins.scss
```diff
@@ -67,7 +68,15 @@
 // $edgeOptOut controls whether to feature-detect around Edge to avoid emitting CSS variables for it,
 // intended for use in cases where interactions with pseudo-element styles cause problems due to Edge bugs.
 @mixin mdc-theme-prop($property, $style, $important: false, $edgeOptOut: false) {
-  @if mdc-theme-is-valid-theme-prop-value_($style) {
+  @if mdc-theme-is-var-with-fallback_($style) {
```
abhiomkar: Please update README including the new type for `$style` argument.

packages/mdc-theme/_variables.scss
```diff
@@ -129,6 +129,10 @@ $mdc-theme-property-values: (
 //
 // NOTE: This function must be defined in _variables.scss instead of _functions.scss to avoid circular imports.
 @function mdc-theme-prop-value($style) {
+  @if mdc-theme-is-var-with-fallback_($style) {
```
abhiomkar: Returning the new style type as is causes an issue to other mixins that uses `mdc-theme-prop-value()` function. For example, `mdc-card-outline` uses this function to set border-color. [mdc-card/_mixins.scss:L303](https://github.com/material-components/material-components-web/blob/b9c5fc6c684d82f63d007761214edf3ca21f99d4/packages/mdc-card/_mixins.scss#L303)

patrickrodee: Can we update that line to be as follows?

```
@include mdc-theme-prop(border-color, $color);
```

patrickrodee: List divider uses the approach described above. I'll include that update in this PR.

https://github.com/material-components/material-components-web/blob/b9c5fc6c684d82f63d007761214edf3ca21f99d4/packages/mdc-list/_mixins.scss#L360

kfranqueiro: I'm confused though, what issue was this causing? Do we need to be concerned for existing valid cases of mdc-theme-prop-value being used?

abhiomkar: It is still an issue if component needs to mix colors or apply opacity using `rgba()` sass function.

For example, dialog uses `rgba()` to apply opacity to color.

**dialog/_mixins.scss:**
```
background-color: rgba(mdc-theme-prop-value($color), $opacity);
```
Where, `mdc-theme-prop-value` returns new color format which rgba function doesn't support.

This is another advanced use case where text field uses `mix()` function two mix two colors:

**textfield/_variables.scss:**
```
$mdc-text-field-background:
    mix(mdc-theme-prop-value(on-surface), mdc-theme-prop-value(surface), 4%);
```

patrickrodee:
@abhiomkar I think that would only work for the existing CSS custom properties (`primary`, `secondary`). I'm intending to support arbitrary CSS custom properties.

packages/mdc-theme/_variables.scss
abhiomkar: @kfranqueiro It shouldn't break existing valid usage. This is an issue only for users who send new color format to component's mixins.

```
mdc-dialog-title-ink-color('purple'); // works
mdc-dialog-title-ink-color({
  varname: '--custom-css-var-ink-color',
  fallback: purple
}); // breaks
```

patrickrodee: `mdc-theme-prop-value` has been updated to always return the fallback value.

https://github.com/material-components/material-components-web/blob/e58979ecd8a1cac81f1cbf2d3501d5ce19b76d20/packages/mdc-theme/_variables.scss#L131-L138

packages/mdc-card/_mixins.scss
```diff
@@ -301,7 +301,7 @@
   }
 
   @include mdc-feature-targets($feat-color) {
-    border-color: mdc-theme-prop-value($color);
+    @include mdc-theme-prop(border-color, $color);
```
abhiomkar: You can revert this now since `theme-prop-value` always returns `fallback` value. We can modify this in all other places in later PRs.

packages/mdc-theme/README.md
```diff
@@ -110,6 +110,38 @@ Property Name | Description
 `on-secondary` | A text/iconography color that is usable on top of secondary color
 `on-surface` | A text/iconography color that is usable on top of surface color
 
+#### `mdc-theme-prop` with CSS Custom Properties
+
+> **Note** The Sass map `$style` argument is intended *only* for use with color mixins.
```
abhiomkar: I don't see why we can't use it for other CSS properties. We aren't doing anything specific with color in our implementation?

packages/mdc-theme/README.md
```diff
@@ -110,6 +110,38 @@ Property Name | Description
...
+```
+
+...will produce the following CSS...
+
+```
+.foo {
+  color: red;
+  color: var(--foo-color, red);
+}
+``
```
abhiomkar: nit: Three ticks to close the block.

packages/mdc-theme/README.md
```diff
@@ -110,6 +110,38 @@ Property Name | Description
...
+
+The above output CSS will apply the fallback color for all supported browsers (including IE11) while allowing for CSS custom property use as a progressive enhancement. This argument type is intended for clients who need custom color application outside of the existing theme properties.
```
abhiomkar: The fallback terminology here might get confused with [Custom property fallback values](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Custom_property_fallback_values). Can this be more specific?

packages/mdc-theme/README.md
```diff
@@ -110,6 +110,38 @@ Property Name | Description
...
+Fields | Description
+--- | ---
+`varname` | The name of a CSS custom property
+`fallback` | A fallback value for the CSS custom property
+
+For example, the following Sass...
+
+```
+.foo {
+  @include mdc-theme-prop(color, (
```
abhiomkar: If I understand this correctly, this is equivalent to applying same mixin with CSS custom property and regular CSS value?

```
mdc-button-fill-color(purple);
mdc-button-fill-color(var(--my-custom-button, purple));
```

packages/mdc-theme/README.md
patrickrodee: I haven't tested it for other CSS custom properties. I'd rather scope this to what it was intended for. We can always remove this line from the docs at a later date.

packages/mdc-theme/README.md
patrickrodee: The output CSS is pretty different. Calling both mixins includes the full selector twice. Below is an approximation of the output CSS:

```
.mdc-button {
  background-color: purple;
}
.mdc-button {
  background-color: var(--my-custom-button, purple);
}
```

Issue #390, opened 2017-03-08
Babel config should follow browsers specification


sgomes:
Using `babel-preset-es2015` is now considered bad practice, and `babel-preset-env` is preferred.

This is because the `env` plugin allows for specifying a list of target browsers, which enables babel to automatically determine which plugins and polyfills to use, resulting in less code. https://github.com/babel/babel-preset-env

It will also enable code size to go down as we remove supported browsers in the future (looking at you, IE 11).

Tasks:
- [ ] Replace `babel-preset-es2015` by `babel-preset-env` in `package.json`
- [ ] Set up an `env` configuration in `package.json` matching our supported browsers (validate string at http://browserl.ist/)
- [ ] Ensure the output includes no polyfills, adjusting options if needed (see `useBuiltIns` and `exclude` options)
- [ ] Ensure output remains sane and works across all supported browsers
- [ ] (Optional) Measure generated JS size delta and brag about it on this issue :) Note that this may actually be small because of IE11 :(

BBosman:
I've done some experimentation and these are my findings.

Before:
```
-rw-r--r--  1 User  staff  248529 Mar  9 15:40 build/material-components-web.js
-rw-r--r--  1 User  staff   99880 Mar  9 15:40 build/material-components-web.min.js
```

After:
```
-rw-r--r--  1 User  staff  248529 Mar  9 15:40 build/material-components-web.js
-rw-r--r--  1 User  staff   99880 Mar  9 15:40 build/material-components-web.min.js
```

So that's a _whopping_ difference of 0 bytes. 

As an experiment I removed IE 11 and got this:
```
-rw-r--r--  1 User  staff  247917 Mar  9 15:43 build/material-components-web.js
```

A difference of 612 bytes, but with a **big caveat**. It will fail on creating the `.min.js` file, as `webpack` uses `uglifyjs` internally to minify the code and it doesn't yet support es2016 (or later) code.

So switching from `babel-preset-es2015` to `babel-preset-env` isn't viable until `webpack`/`uglifyjs` support es2016 (and newer) imho.

sgomes:
Thanks for looking into this, @BBosman! uglify not supporting ES2015 is a known issue, yes, good point :-/

If we are to this we'll need to add investigating a replacement, such as [babili](https://github.com/babel/babili) to the list of tasks.

traviskaufman:
[...]

PR #4471, opened 2019-03-01
chore(infrastructure): update from babel-preset-es2015 to env

moog16:
fixes https://github.com/material-components/material-components-web/issues/390


moog16:
I thought we wanted to keep getting-started to 2015, since users might want to use class and whatever else 2015 comes with. However if they are just using our API, they might not be using class. I can update getting-started too if you agree with that.

moog16:
Ya as far as users are concerned it won't matter to them as much.

PR #4473, opened 2019-03-01
fix(list): Remove unused adapter.removeAttributeForElementIndex

moog16:
While implementing react list, I noticed this method is actually not used in the foundation.


kfranqueiro:
I think this seems reasonable, but it occurs to me we're going to need to start noting even adapter API removals as BREAKING CHANGE because it will break anyone using the adapter type in TypeScript if they're implementing a superset.

cc @abhiomkar if he wants to take a look at this too.

moog16:
Ya - I also thought this is a breaking change even if its not being used. Users should at least be aware via the CHANGELOG.

PR #4477, opened 2019-03-04
docs: Replace stackblitz link with glitch for quick start demo

abhiomkar:


README.md
```diff
@@ -72,7 +72,7 @@ Sample usage of text field component. Please see [MDC Textfield](packages/mdc-te
 ```html
 <div class="mdc-text-field">
   <input type="text" id="my-text-field" class="mdc-text-field__input">
-  <label class="mdc-floating-label" for="my-text-field">Hint text</label>
+  <label class="mdc-floating-label" for="my-text-field">Label</label>
```
moog16: there is also another `<label class="mdc-floating-label" for="my-text-field">Hint text</label>` above. Do you also want to change that label?

abhiomkar: Good catch. Changed it.

PR #4478, opened 2019-03-05
fix(text-field): Fix for input alignment in textfield with trailing icon

### LINKED ISSUE #4416: Filled text fields floating label doesn't align with trailing icon

Chome

### What OS are you using?

ChomeOS

Have a boxed input with a floating label and a trailing icon, input some text. The floated label will not align with the input.

The floated label and the text will align.

abhiomkar:
Made changes to private mixin of `text-field/icon`: `mdc-text-field-icon-horizontal-position_` by adding a new argument to send default text field input padding for trailing icon variant.

In previous implementation it assumes the leading input padding and the icon position always equals, but it is not the case in filled variant since the icon position is 4px lesser (`12px`) in filled variant compared to outlined variant (`16px`).

**Before**

**After**

Fixes #4416


packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -42,7 +42,7 @@
     }
   } @else {
     .mdc-text-field__input {
-      @include mdc-rtl-reflexive-property(padding, $position /* left-value */, $padding  /* right-value */);
+      @include mdc-rtl-reflexive-property(padding, $input-padding /* left-value */, $padding  /* right-value */);
```
kfranqueiro: I'm a little bit confused as to why this change is only needed within the `else`?

abhiomkar: Used `$input-padding` to `left` position too to make it semantically correct. Icon's `position` value and text field's `input-padding` are same anyway.

PR #4479, opened 2019-03-06
feat(list): Automatically use appropriate aria attribute for single selection list.

### LINKED ISSUE #4128: List: Use appropriate role for selected/activated list items based on parent role
From #3403:

We may be able to auto-detect the presence of roles in which `aria-selected` is valid, and in other cases opt for `aria-current` instead (which is the most semantically-correct attribute to use for e.g. drawer navigation).

Since this role is applied dynamically within the List foundation based on user interaction, it would require JS updates, in addition to documentation updates to the relevant components.

abhiomkar:
With this change List detects the presence of `aria-selected` or `aria-current` ARIA attributes and uses it to set to next selected list item. `aria-current` value can either be any of the token value (Eg, `page`, `step` etc) or `true`.

This change also fixes accessibility audit error:
`[aria-*] attributes do not match their roles`

**Screenshot** of annotation by Mac's Voice Over on Safari:
image

Fixes #4128

BREAKING CHANGE: Added new adapter method `getAttributeForElementIndex` in MDC List for improved accessibility.


packages/mdc-list/foundation.ts
```diff
@@ -295,29 +301,49 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
...
     this.selectedIndex_ = index;
   }
 
+  /**
+   * Sets aria attribute for single selection at given index.
+   */
+  private setAriaForSingleSelectionAtIndex_(index: number) {
+    if (this.selectedIndex_ === numbers.UNSET_INDEX) {
+      this.ariaCurrentAttrValue_ =
+            this.adapter_.getAttributeForElementIndex(index as number, strings.ARIA_CURRENT);
```
kfranqueiro: I'm confused how this method works, since this is reading the attribute value for the element that we haven't set the attribute on yet. Wouldn't this always yield `'false'` or `null` (given that this method is never called when `selectedIndex_ === index`)? Shouldn't this be doing exactly the opposite of what it's doing - checking the attribute's value at `selectedIndex_` if `selectedIndex_` is set?

I don't see any tests verifying that `aria-current` is ever actually set to a value other than `null`, so I highly suspect a bug here.

Meanwhile, I take it this method will always fall back to using `aria-selected` if the markup had nothing selected initially? Is it not possible to detect based on the list's role instead? I don't remember whether we _required_ that an item be initially selected before.

Also, why does `index` need `as number`? It's already declared as `number`...

abhiomkar: To clarify your first question, `setSelectedIndex()` method is called on initialization of the component to set pre-selected items where `this.selectedIndex_` is still "unset". That's when we detect the presence of one of the aria attribute. See [component.ts:L126](https://github.com/material-components/material-components-web/blob/7357170958916857fb7f47b077496403795ad861/packages/mdc-list/component.ts#L126).

I've considered adding a new foundation method to set pre-selected index (For example, `setInitialSelectedIndex` or `setDefaultSelectedIndex`) which just sets the state - `this.selectedIndex_`, I dropped this idea since having two foundation setter methods might create ambiguity for wrapper libraries.

WDYT?

kfranqueiro: So I thought I must've been missing something, and after talking with Abhinay, I realized I was.

Firstly, `this.ariaCurrentAttrValue_` is an instance variable, and we expect to set it exactly once here, then reuse it for all subsequent updates.

Secondly, the first call to this method (where this condition will be true) happens during the component's `initializeListType` method, at which point it calls this method with the index that _already has_ the attribute set. This is in direct contrast to subsequent calls which pass an index to select, rather than one that's already selected.

abhiomkar: Added a code comment to make it more clearer. Also, added new unit tests.

For records, here are the alternatives I've considered to get around detecting something from DOM during initialization to keep the state in sync with foundation:

- Move pre-selected logic from component to adapters which will be eventually called from foundation. This might need two new additional adapters - `getInitialSelectedIndex` and `setUseActivatedClass` call can be moved inside provided we've `hasClassAtIndex` that checks existence of `LIST_ITEM_ACTIVATED_CLASS` class name.
- New foundation method to set pre-selected index (For example, setInitialSelectedIndex or setDefaultSelectedIndex) which just sets the state `this.selectedIndex_`.

packages/mdc-list/README.md
```diff
@@ -418,6 +418,9 @@ Use `role="listbox"` only for single selection list, without this role the `ul`
 Do not use `aria-orientation` attribute for standard list (i.e., `role="list"`), use component's `vertical` property to set the orientation
 to vertical.
 
+Single selection list supports `aria-selected` and `aria-current` attributes. List automatically detects the presence of these attributes
```
kfranqueiro: The "Pre-selected list item" and "Setup in `singleSelection()`" sections still mention `aria-selected` exclusively... should their wording be broadened to mention `aria-current`? (Maybe not in the single-selection case if that's expected to only ever apply for roles where `aria-selected` is valid?)

packages/mdc-drawer/README.md
```diff
@@ -34,7 +34,7 @@ npm install @material/drawer
 <aside class="mdc-drawer">
   <div class="mdc-drawer__content">
     <nav class="mdc-list">
-      <a class="mdc-list-item mdc-list-item--activated" href="#" aria-selected="true">
+      <a class="mdc-list-item mdc-list-item--activated" href="#" aria-current="page">
```
kfranqueiro: Should we also modify the fixture in `mdc-drawer.test.js`?

test/unit/mdc-list/foundation.test.js
```diff
@@ -843,13 +852,41 @@ test('#setSelectedIndex removes selected/activated class name and sets aria-sele
...
+  td.when(mockAdapter.getAttributeForElementIndex(2, strings.ARIA_CURRENT)).thenReturn(null);
+  td.when(mockAdapter.getAttributeForElementIndex(3, strings.ARIA_CURRENT)).thenReturn(null);
   foundation.setSelectedIndex(2);
 
   foundation.setSelectedIndex(3);
   td.verify(mockAdapter.removeClassForElementIndex(2, cssClasses.LIST_ITEM_SELECTED_CLASS), {times: 1});
   td.verify(mockAdapter.setAttributeForElementIndex(2, strings.ARIA_SELECTED, 'false'), {times: 1});
 });
 
+test('#setSelectedIndex sets aria-current="false" to previously selected index and sets aria-current without any token'
```
kfranqueiro: Do we really need both of these tests? Should just the `'page'` test suffice?

Also, should we have a test somewhere for verifying that aria-current vs. aria-selected is properly detected on initialization?

kfranqueiro:
One other thing I forgot to comment on - should we update demos/list.html for consistency?

packages/mdc-list/README.md
abhiomkar: I modified "Pre-selected list item" and "Setup in singleSelection()" section cross-linking to "Accessibility" section.

packages/mdc-drawer/README.md
abhiomkar: Good catch! Changed it to aria-current in fixture.

abhiomkar:
Thanks for the review! Resolved all the review comments. PTAL.

We could add few more screenshot tests to cover for `demos/list.html` is there anything particular you need a demo for?

I thought we no longer update or maintain `demos/` folder.

test/unit/mdc-list/foundation.test.js
```diff
@@ -843,13 +851,65 @@ test('#setSelectedIndex removes selected/activated class name and sets aria-sele
...
+
+  td.when(mockAdapter.getListItemCount()).thenReturn(5);
+  td.when(mockAdapter.getAttributeForElementIndex(2, strings.ARIA_CURRENT)).thenReturn('page');
+  foundation.setSelectedIndex(2);
+
+  td.verify(mockAdapter.setAttributeForElementIndex(2, strings.ARIA_CURRENT, 'false'), {times: 0});
+  td.verify(mockAdapter.setAttributeForElementIndex(2, strings.ARIA_CURRENT, 'page'), {times: 1});
+});
+
+test('#setSelectedIndex should set aria-selected as default option in the absence of aria-selected on pre-selected '
```
kfranqueiro: I'm confused looking at these two tests because both seem to have identical setup, different verifications, but both...pass? If aria-current is found, shouldn't aria-selected _not_ be set?

abhiomkar: The next test was fixed in immediate next commit:
https://github.com/material-components/material-components-web/pull/4479/commits/9b7f617ae2cc7ded0f0e704a86aaccf4696eabf2

You'll not see the same test setup in full PR diff. Sorry for the confusion.

kfranqueiro: Nevermind, I overlooked the `'page'` / `null` difference.

abhiomkar:
Thanks for the clarification! Updated `demos/list.html` with navigation example that has `aria-current` attribute.

PR #4485, opened 2019-03-08
fix: Don't import * from focus-trap to avoid default export confusion

### LINKED ISSUE #4484: bug(MDCDialog) Uncaught TypeError: e is not a function at Object.i [as createFocusTrapInstance]

1.0.0

Chrome version: 72.0.3626.119 (64 bits)

### What OS are you using?

Windows 10

1. Clone: https://github.com/Jelmer7/mdc-dialog-test
2. Run: npm install
3. Run: npm run build 
4. Open index.html

**or**

1. Install locally following Getting started guide on material.io
    **Using MDC Web with Sass and ES2015**
2. Update packages to 1.0.0
3. Create page with mdc dialog (literally copied from material.io)

No errors initilializing dialog

Console gives:
Uncaught TypeError: e is not a function...

This only happens locally, using CDN this problem does not occur
for demonstration clone: https://github.com/Jelmer7/mdc-dialog-test

kfranqueiro:
Fixes #4484.


packages/mdc-drawer/component.ts
```diff
@@ -25,7 +25,8 @@ import {MDCComponent} from '@material/base/component';
 import {SpecificEventListener} from '@material/base/types';
 import {MDCList, MDCListFactory} from '@material/list/component';
 import {MDCListFoundation} from '@material/list/foundation';
-import * as createFocusTrap from 'focus-trap';
+import createFocusTrap from 'focus-trap';
```
abhiomkar: make this change in dialog's `component.ts` too?

packages/mdc-dialog/util.ts
```diff
@@ -21,18 +21,19 @@
  * THE SOFTWARE.
  */
 
-import * as createFocusTrap from 'focus-trap';
+import createFocusTrap from 'focus-trap';
+import {FocusTarget, FocusTrap, Options} from 'focus-trap';
```
acdvorak: I think these lines can be combined:

```ts
import createFocusTrap, {FocusTarget, FocusTrap, Options} from 'focus-trap';
```

Ditto other files

acdvorak:
I think `fix` is correct because this fixes bug #4484

kfranqueiro:
I think the screenshot tests are diffing because of the same reason the unit tests are failing. The interesting part is that this error sounds eerily similar to the error that #4484 is experiencing in 1.0.0...

kfranqueiro:
So a few things happened to lead to where this branch is at now.

* The absence of `esModuleInterop` is what caused this to show up in some environments but not others... and then invert which cases passed/failed when I attempted to fix it.
* Adding `esModuleInterop` caused marked increases in the size of our UMD bundles when combined with `importHelpers`, because basically every module now ends up requiring the helpers, and webpack pulls in the full `tslib` import whereas `tsc` only emits the necessary functions.
* `importHelpers` _is_ desirable for individual ES modules, because otherwise you end up with various subsets of tslib emitted in every individual module.
* TL;DR: this PR turns off `importHelpers` in the tsconfig to avoid UMD bundle bloat, but adds it via CLI for `build:esmodules` specifically.

gertcuykens:
https://github.com/material-components/material-components-web/issues/4507

hmm could it be something is still not right?

PR #4486, opened 2019-03-08
chore: Add custom mdc linter rules for TS

acdvorak:
### New linter rules

TODO: Provide code examples

* Require all top-level exported symbols to be documented in the package README
* Require all top-level exported types to start with `MDC`
* Require all exports to be inline; disallow named object exports in all files except `index.ts` (this makes it easier to determine if a symbol is exported because you can simply walk up the AST until you find an `export` node)
* Require an "assigned in METHOD()" comment for definite assignment assertions (and auto-generate the suggested method name). E.g.: `private progress_!: number; // assigned in init()`
* Prohibit type definitions in `index.ts` files (only imports and re-exports allowed)
* Prohibit leading underscores on all node types
* Prohibit trailing underscores on all node types except class properties/methods and local variables
* Require `foundation.ts` files to export at least one class ending in `Foundation`
* Require `adapter.ts` files to export at least one interface ending in `Adapter`
* Require all types in `adapter.ts` files to be publicly-exported interfaces ending with `Adapter`


scripts/lint-mdc.ts
```diff
@@ -80,16 +83,24 @@ interface PackageJson {
 
 const PACKAGES_DIR_ABSOLUTE = path.resolve(__dirname, '../packages');
 
+// TODO(acdvorak): Remove Corner, CornerBit, ALWAYS_FLOAT_TYPES, VALIDATION_ATTR_WHITELIST
```
kfranqueiro: When you say remove... do you mean these should be mentioned in the READMEs so they don't need to be excluded here?

scripts/lint-mdc.ts
```diff
@@ -347,16 +377,117 @@ function visitOneAssignmentExpression(
   });
 }
 
-function checkAllReturnAnnotations(inputFilePath: string, inputCode: string) {
+function checkAllAnnotations(inputFilePath: string, inputCode: string) {
```
kfranqueiro: Can we comment above functions like this to explain what "All" is?

scripts/lint-mdc.ts
```diff
@@ -496,28 +757,234 @@ function checkOneClassMemberIsInReadme(
...
+  if (babelTypes.isClassDeclaration(node)) {
+    return 'class';
+  }
+  if (babelTypes.isTSEnumDeclaration(node)) {
+    return 'enum';
+  }
+  if (babelTypes.isVariableDeclaration(node)) {
+    return 'variable';
+  }
+  if (babelTypes.isVariableDeclaration(node)) {
```
kfranqueiro: Repeat of the previous condition?

scripts/lint-mdc.ts
```diff
@@ -496,28 +757,234 @@ function checkOneClassMemberIsInReadme(
...
+}
+
+function collectPubliclyExportedClassNames(nodePath: NodePath<ClassLikeNode>, classNames: string[]) {
+  const id = nodePath.node.id;
+  if (id && isInlineExportedSymbol(nodePath)) {
+    classNames.push(id.name);
+  }
+}
+
+function getIdentifierInfo(node: babelTypes.Node): NamedIdentifier {
```
kfranqueiro: This seems like a lot of code to normalize across different things in the AST... I guess there's no built in convenience for anything like this?

scripts/lint-mdc.ts
```diff
@@ -496,28 +757,234 @@ function checkOneClassMemberIsInReadme(
...
-  if (babelTypes.isClassMethod(node) || babelTypes.isClassProperty(node) ||
-      babelTypes.isObjectMethod(node) || babelTypes.isObjectProperty(node) ||
+function checkAllAdaptersAndFoundations(inputFilePath: string, inputCode: string) {
+  const isAdapterFile = inputFilePath.endsWith('adapter.ts');
+  const isFoundationFile = inputFilePath.endsWith('foundation.ts');
+  const publiclyExportedClassNames: string[] = [];
+  const ast = getAstFromCodeString(inputCode);
+
+  babelTraverse(ast, {
+    ClassDeclaration(nodePath: NodePath<babelTypes.ClassDeclaration>) {
```
kfranqueiro: Don't we only ever have `interface` Adapter types in adapter modules? What are the other two traversals for? (Also, what's InterfaceDeclaration vs. TSInterfaceDeclaration?)

kfranqueiro: Wait, I guess this happens not just for adapter files. I guess I find this confusing because both this and the below adapter-only traversal push into the same `publiclyExportedClassNames` array which seems error-prone in principle even if it's not right now.

(I don't suppose it'd be reasonable to have that function return an array rather than append to one...)

scripts/lint-mdc.ts
```diff
@@ -496,28 +757,234 @@ function checkOneClassMemberIsInReadme(
...
+        ],
     );
     return;
   }
 }
 
-function getNameAndLocation(node: babelTypes.Node): NamedIdentifier {
-  if (babelTypes.isClassMethod(node) || babelTypes.isClassProperty(node) ||
-      babelTypes.isObjectMethod(node) || babelTypes.isObjectProperty(node) ||
+function checkAllAdaptersAndFoundations(inputFilePath: string, inputCode: string) {
```
kfranqueiro: Would this actually make more sense to be 2 separate functions, one for adapters, one for foundations? I'm honestly finding it confusing why there's an extra traversal for the adapter case anyway.

abhiomkar:
Closing since inactive.

PR #4503, opened 2019-03-13
docs: Add placeholder for circular progress component

lucasecdb:


kfranqueiro:
Thanks for thinking ahead to this. I'm going to check with a couple of people though, because I think the plan is to get rid of the placeholder directories, because they end up causing confusion when looking at the directory listing in the repository under `packages`.

kfranqueiro:
I've gotten confirmation that we'll be able to remove the placeholder folders in a week or two, so I'm going to close this.

Note that removing the placeholders has no bearing on our plans of implementing components!

PR #4516, opened 2019-03-19
feat(drawer): Make list instance publicly accessible

### LINKED ISSUE #4364: Make List public in MDC Drawer
Make `this.list_` public in `MDCDrawer` in order to programmatically set properties on the `MDCList`, such as `selectedIndex`.

[Example use case on Glitch](https://glitch.com/~material-tabs-to-drawer)

kfranqueiro:
Fixes #4364.

### DIFF
--- a/packages/mdc-drawer/component.ts
+++ b/packages/mdc-drawer/component.ts
@@ -68,6 +68,10 @@ export class MDCDrawer extends MDCComponent<MDCDismissibleDrawerFoundation> {
   private handleKeydown_!: SpecificEventListener<'keydown'>; // initialized in initialSyncWithDOM()
   private handleTransitionEnd_!: SpecificEventListener<'transitionend'>; // initialized in initialSyncWithDOM()
 
+  get list(): MDCList | undefined {
+    return this.list_;
+  }
+
   initialize(
       focusTrapFactory: MDCDrawerFocusTrapFactory = createFocusTrap as unknown as MDCDrawerFocusTrapFactory,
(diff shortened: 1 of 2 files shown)


packages/mdc-drawer/component.ts
```diff
@@ -68,6 +68,10 @@ export class MDCDrawer extends MDCComponent<MDCDismissibleDrawerFoundation> {
   private handleKeydown_!: SpecificEventListener<'keydown'>; // initialized in initialSyncWithDOM()
   private handleTransitionEnd_!: SpecificEventListener<'transitionend'>; // initialized in initialSyncWithDOM()
 
+  get list(): MDCList | undefined {
```
abhiomkar: May not be related to this PR. Looking at the implementation of drawer component, `list_` should not be `undefined`, drawer component should throw error when list is not found instead. Drawer foundation breaks at `focusActiveNavigationItem` adapter method if list is not found.

kfranqueiro: I had wondered about this too, but we have the `list_` member currently declared as `MDCList?` and it looks like we're tolerating it being absent in `destroy`. So maybe there's indeed a separate bug to be had here...

kfranqueiro: Upon further inspection/discussion, it looks like the adapter APIs also tolerate the list not existing, so in the interests of maintaining current behavior I'll keep this PR as-is.

tychenjiajun:
It will be better if I can listen to MDCList:action in the drawer.

PR #4517, opened 2019-03-19
fix(menu): Fix selection group list item spacing

abhiomkar:
Before:

After:


packages/mdc-menu/_mixins.scss
```diff
@@ -80,7 +80,7 @@
 
     .mdc-list-item {
       @include mdc-feature-targets($feat-structure) {
-        @include mdc-rtl-reflexive-box(padding, left, 56px);
+        @include mdc-rtl-reflexive-property(padding, 56px, 16px);
```
kfranqueiro: Reuse `$mdc-list-side-padding` for the `16px` (import `mdc-list/variables`)

PR #4518, opened 2019-03-19
chore: Test integrating bundlesize

abhiomkar:


abhiomkar:
This needs to be integrated with Travis and also good to have a trending graph with bundle size information for each component.

PR #4519, opened 2019-03-19
fix(card): Use on-surface color for action icons

### LINKED ISSUE #4238: MDCCard: text and background colours
According to the `mdc-card` [stylesheet](https://github.com/material-components/material-components-web/blob/d4f77e1a25bc2f78cf575ad7c69bd55ee3d26941/packages/mdc-card/mdc-card.scss#L33), the fill colour of a card is `surface`. But the colour for icons is [defined below](https://github.com/material-components/material-components-web/blob/d4f77e1a25bc2f78cf575ad7c69bd55ee3d26941/packages/mdc-card/mdc-card.scss#L133) as `text-icon-on-background`. As surface and background colours can differ, it seems like a bug.

Moreover, why there are no predefined variables like `text-<TEXT-STYLE>-on-surface`?
There are some for `background`, `light`, `dark`, but no for `surface`. It seems that text on surface is quite common. Is it missing on purpose or by accident?

0.43.0

kfranqueiro:
Fixes #4238.


kfranqueiro:
I checked with design on this while I was asking about #4520, and this should be on-surface with 87% opacity (high emphasis).

test/screenshot/golden.json
```diff
@@ -160,11 +160,11 @@
     }
   },
   "spec/mdc-card/classes/media.html": {
-    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2018/11/02/14_50_15_747/spec/mdc-card/classes/media.html?utm_source=golden_json",
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/kfranqueiro/2019/03/19/21_19_18_527/spec/mdc-card/classes/media.html?utm_source=golden_json",
```
abhiomkar: The action icon buttons looks more black than it should comparing to spec. We might need to adjust the opacity.

kfranqueiro: This is correct, I mistyped above. The icons are supposed to be medium emphasis. Fixed in latest commits.

PR #4520, opened 2019-03-19
fix(menu): Use on-surface color for graphic/meta content

### LINKED ISSUE #4238: MDCCard: text and background colours

kfranqueiro:
Fixes #4238 as it pertains to Menu.

### DIFF
--- a/packages/mdc-menu/_mixins.scss
+++ b/packages/mdc-menu/_mixins.scss
@@ -35,8 +35,9 @@
 
   // Customize the menu-surface and contained list to match the mdc-menu styles.
   .mdc-menu {
-    @include mdc-list-item-meta-ink-color(on-surface, $query);
-    @include mdc-list-item-graphic-ink-color(on-surface, $query);
+    @include mdc-list-item-primary-text-ink-color($mdc-menu-ink-color, $query);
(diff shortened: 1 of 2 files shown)


kfranqueiro:
It occurs to me that the current state of this PR leads to menu list item text still being on-background but icons are on-surface, which is probably wrong. I'm going to see if I can double-check with design on whether everything should be on-surface.

kfranqueiro:
Update from design: both the icons and text should be on-surface at 87% opacity (high emphasis).

PR #4522, opened 2019-03-20
feat(tab): Add Tab Sass mixins targeting active state colors

bwobrien:


packages/mdc-tab/_mixins.scss
```diff
@@ -49,10 +49,27 @@
   @include mdc-tab-text-label-color($color);
 }
 
+@mixin mdc-tab-active-text-label-color($color) {
+  &.mdc-tab--active .mdc-tab__text-label {
```
kfranqueiro: Can these mixins just do what was being done in tab-bar.scss before and reuse other mixins within `&.mdc-tab--active`? That way there's no risk of updating one mixin and forgetting to update these other mixins in the future.

bwobrien: Do you mean omitting `&.mdc-tab--active` here and requiring developers to specify it themselves?

bwobrien: Oh nevermind, I get it, reuse `mdc-tab-text-label-color`. Of course, will make the change.

bwobrien: Done.

PR #4526, opened 2019-03-20
feat(fab): Add feature targeting for styles

kfranqueiro:
Refs #4227.

There are CSS diffs due to reordering, but the same styles should still be present.

cc @mmalerba


packages/mdc-fab/_mixins.scss
```diff
@@ -21,124 +21,220 @@
...
-  @include mdc-fab-container-color($container-color);
+  @include mdc-fab-container-color($container-color, $query: $query);
 
-  @if ($fill-tone == "dark") {
-    @include mdc-fab-ink-color(text-primary-on-dark);
-    @include mdc-states(text-primary-on-dark);
-  } @else {
-    @include mdc-fab-ink-color(text-primary-on-light);
-    @include mdc-states(text-primary-on-light);
+  @include mdc-feature-targets($feat-color) {
```
mmalerba: I think you can remove this one since you're passing the query through below

kfranqueiro: Good point... I had originally forgotten to forward the query through the includes below, and then forgot about the include here.

...Meanwhile, now I'm spooked because this somehow didn't fail building? Seems like it should have...

mmalerba: Yeah that is strange, I can see how the feature-targeting test would miss it. We gave it a query that selects nothing, so it fails the first feature-targets check and Sass doesn't bother evaluating what's inside.

But is this getting built as part of the dev app? https://github.com/material-components/material-components-web/blob/f310e7364f6f7a7ae162908ca27808f0e4619184/demos/fab.scss#L93 I would have expected that to result in an error

mmalerba: Oh actually it did fail when I checked out your PR, reverted the fix commit and tried to `npm run dev`. Got the following error:

```
ERROR in ./demos/fab.scss
    Module build failed: 
        @error "mdc-feature-targets must not be used inside of another mdc-feature-targets block";
```

I guess maybe this isn't built as part of your CI?

mmalerba: I made a PR (https://github.com/material-components/material-components-web/pull/4528) to run all the test mixins twice, once with none of the features and once with all of them. In the all of them case, we don't need to do anything with the output, just make sure that it compiles. I think this would have caught your issue here

PR #4529, opened 2019-03-21
chore: Add ability to cherry-pick commits for minor releases

kfranqueiro:
This renames `cherry-pick-patch-commits.js` to just `cherry-pick-commits.js`, and it now expects to see either a `--patch` or `--minor` argument.

`--patch` continues to behave as this script behaved before.

`--minor` seeks to the latest major/minor (not patch) tag, then cherry-picks all non-breaking-change commits.

The end result is similar in both cases - you end up detached from any branch in your local clone and can follow the rest of the release process from there.

(I'll send a separate PR to update the rest of our release process doc.)

This also adds a `--skip-tests` flag to the script, which really shouldn't be used for actual releases, but can be useful when testing updates to this script since it takes a while to build our assets and run our tests, and the final results don't get output until after they run.

### DIFF
--- a/scripts/cherry-pick-commits.js
+++ b/scripts/cherry-pick-commits.js
@@ -22,7 +22,10 @@
  */
 
 /**
- * @fileoverview Identifies and cherry-picks commits that are not features and do not contain breaking changes.
+ * @fileoverview Identifies and cherry-picks commits for minor or patch releases.
+ *
+ * Minor releases skip commits that do not contain breaking changes.
+ * Patch releases skip commits that are not features and that do not contain breaking changes.
  *
  * Defaults to operating against the latest non-pre-release tag, but a tag can be specified via command-line argument.
  * This will automatically attempt to cherry-pick appropriate commits since the tag, but will abort and skip any
@@ -52,10 +55,13 @@ const logFormat = {
 const logSplitter = ';;;';
 
 // Resolves to the most recent non-pre-release git tag.
-async function getMostRecentTag() {
+async function getMostRecentTag(mode) {
   const tags = await simpleGit.tags();
-  // Filter old independent-versioned tags and pre-releases
-  const filteredTags = tags.all.filter((tag) => /^v/.test(tag) && !/-/.test(tag));
+  // Filter old independent-versioned tags and pre-releases.
+  // If cherry-picking for a minor release, ignore patch releases to work from the latest major or minor release.
+  const filteredTags = tags.all.filter((tag) => /^v/.test(tag) && !/-/.test(tag) &&
+      (mode !== 'minor' || /\.0$/.test(tag)));
+
   return filteredTags[filteredTags.length - 1];
 }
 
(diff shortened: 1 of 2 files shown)


scripts/cherry-pick-commits.js
```diff
@@ -52,10 +55,13 @@ const logFormat = {
...
 // Resolves to the most recent non-pre-release git tag.
-async function getMostRecentTag() {
+async function getMostRecentTag(mode) {
   const tags = await simpleGit.tags();
-  // Filter old independent-versioned tags and pre-releases
-  const filteredTags = tags.all.filter((tag) => /^v/.test(tag) && !/-/.test(tag));
+  // Filter old independent-versioned tags and pre-releases.
+  // If cherry-picking for a minor release, ignore patch releases to work from the latest major or minor release.
+  const filteredTags = tags.all.filter((tag) => /^v/.test(tag) && !/-/.test(tag) &&
+      (mode !== 'minor' || /\.0$/.test(tag)));
```
moog16: can `mode !== minor` be changed to the affirmative? `mode === 'patch'`

scripts/cherry-pick-commits.js
```diff
@@ -172,15 +185,24 @@ async function run() {
...
+  let buildSucceeded;
+  let testsSucceeded;
+  const skipTests = args.includes('--skip-tests');
 
-  console.log('');
-  console.log('Running unit tests...');
-  const testsSucceeded = checkSpawnSuccess('npm run test:unit');
+  if (skipTests) {
+    console.log('Skipping npm run build and npm run test:unit.');
+    console.log('You should _really_ run these unless you\'re just testing this script.');
```
moog16: Is the _really_ supposed to appear italicized in the terminal?

kfranqueiro: I think that depends on the terminal emulator being used. I didn't feel like going crazy into formatting codes...

scripts/cherry-pick-commits.js
kfranqueiro: Do you think that's easier to read than this? I wanted to express "restrict to .0 versions for minor releases" which led me to include minor in the condition.

scripts/cherry-pick-commits.js
```diff
@@ -52,10 +55,13 @@ const logFormat = {
 const logSplitter = ';;;';
 
 // Resolves to the most recent non-pre-release git tag.
-async function getMostRecentTag() {
+async function getMostRecentTag(mode) {
```
moog16: name changee getMostRecentTag --> getStartingTag (or better name)

scripts/cherry-pick-commits.js
moog16: could you pull this into a variable name. I think that would be clearer.

kfranqueiro: We decided naming a variable for this would be even harder, and the other comments I added improved this.

Issue #4523, opened 2019-03-20
Incorrect / confusing README info regarding applying text style classes in themes


traviskaufman:
When looking at the [css custom properties section of the mdc-theme README](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#css-custom-properties), it says:

> **A note about <TEXT_STYLE>,** <TEXT_STYLE> represents the lowercase name of the text styles listed above, e.g. `hint`

However, AFAICT there don't seem to be any "text styles" listed above, and when I look in the table of modifiers I do not see "hint" anywhere on the page:

![mdc-theme-no-textstyles](https://user-images.githubusercontent.com/1185269/54718040-01aae080-4b30-11e9-8701-3aab5138d19e.png)

I was able to find it by looking in [the code](https://github.com/material-components/material-components-web/blob/master/packages/mdc-theme/_variables.scss#L51-L66); perhaps it was removed from the docs at some point and needs to be re-added back in?

FWIW even though it's a bit more work listing out the text styles explicitly would create less work for me as an end user :smiley: 

Thanks for all your hard work on the library!

moog16:
We might have remove this thinking we would add them back. We will try to digging in history and put that table back in place.

kfranqueiro:
This might've been intentionally removed while work was being done on mdc-theme last year... but it also might've been unintentionally removed when we cleaned up formatting in our readme files.

We've gradually been moving towards 3 text emphasis levels (see `$mdc-theme-text-emphasis` below the code you highlighted), which we tend to combine with a color via `rgba` in Sass (though this is admittedly less css-custom-property friendly), but there are still some styles (and the existing mdc-theme variables) referencing the original 5-level map that went missing from the docs. For the time being, at least, it seems like we'll need to document both.

PR #4531, opened 2019-03-21
docs(theme): add back the text styles and primary colors

moog16:
fixes https://github.com/material-components/material-components-web/issues/4523


packages/mdc-theme/README.md
```diff
@@ -11,6 +11,21 @@ path: /catalog/theme/
 
 The Material Design color system can be used to create a color scheme that reflects your brand or style.
 
+The colors in this module are derived from three theme colors:
```
abhiomkar: I think the OP is referring to missing CSS variable names in our docs that were removed in this commit (https://github.com/material-components/material-components-web/commit/9394b9dd1e735b29e390a92a00d9ce08b2f50cfb)

For example, documentation for `--mdc-theme--text-primary-on-light` CSS custom property. We should at least list all possible replacement for `TEXT_STYLE` in the table every time when we mention `TEXT_STYLE` (ie., primary, secondary, hint, disabled and icon).

kintz09: Does background replace surface here? Or are both still used?

kfranqueiro: Surface was introduced later, but I agree that restoring this part of the docs would be confusing. The main focus of this PR should be the five text styles.

moog16: I believe the different options are in this PR between lines 22 and 26. Is there one that I'm missing?

moog16: Gotcha - I will remove this part.

abhiomkar: I was referring to these diffs specifically: https://github.com/material-components/material-components-web/commit/9394b9dd1e735b29e390a92a00d9ce08b2f50cfb#diff-7f578dcc2f8f072f080e3144e80a3e60L96

> A note about <TEXT_STYLE>, <TEXT_STYLE> represents the lowercase name of the text styles listed above, e.g. hint.

We've a note about `TEXT_STYLE` in our docs but it is not very clear which section we're referring to when we say **_listed above_**. It would be easier if we specify possible options for `TEXT_STYLE` next it it.

moog16: Yes you're right...I updated it. WDYT?

PR #4534, opened 2019-03-22
feat(grid-list): Add feature targeting for styles

kfranqueiro:
Refs #4227.

cc @mmalerba


packages/mdc-grid-list/_mixins.scss
```diff
@@ -0,0 +1,249 @@
...
+@import "@material/theme/mixins";
+@import "@material/typography/mixins";
+@import "./variables";
+
+@mixin mdc-grid-list-core-styles($query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  // postcss-bem-linter: define grid-list
+  @at-root {
```
mmalerba: is the `@at-root` necessary?

packages/mdc-grid-list/_mixins.scss
```diff
@@ -0,0 +1,249 @@
...
+
+  .mdc-grid-tile__title {
+    @include mdc-typography-overflow-ellipsis($query: $query);
+
+    @include mdc-feature-targets($feat-structure) {
+      display: block;
+      margin: 0;
+      padding: 0;
+      font-size: 1rem;
+      font-weight: #{map-get($mdc-typography-font-weight-values, medium)};
```
mmalerba: should this one be under typography?

packages/mdc-grid-list/_mixins.scss
kfranqueiro: Probably not, good point. I think we...uh..."rooted" out a couple of these elsewhere during feature targeting updates as well.

packages/mdc-grid-list/_mixins.scss
kfranqueiro: Whoops, looks like it. Thanks for catching that.

packages/mdc-grid-list/_mixins.scss
```diff
@@ -0,0 +1,251 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+@import "@material/feature-targeting/functions";
```
abhiomkar: Please update package.json with feature-targeting dependency.

packages/mdc-grid-list/_mixins.scss
```diff
@@ -0,0 +1,251 @@
...
+
+    @include mdc-feature-targets($feat-structure) {
+      display: block;
+      margin: 0;
+      padding: 0;
+      font-size: 1rem;
+      line-height: 1rem;
+    }
+
+    @include mdc-feature-targets($feat-typography) {
```
abhiomkar: Curious what does the typography includes? Should font-size and line-height be part of typography?

packages/mdc-grid-list/_mixins.scss
kfranqueiro: Oh crud, I forgot this for the last few as well, just went and fixed those on this branch. Thanks!

packages/mdc-grid-list/_mixins.scss
kfranqueiro: It's primarily intended to include things that depend on the mdc-typography scale variables. Thinking on this particular line again (which @mmalerba brought up), I'm not even sure it's absolutely necessary for this, since this is just font-weight... Thoughts, Miles? What's the rule of thumb for things falling under the typography feature aside from direct uses of the `mdc-typography` mixin?

mmalerba: Like you said, I think if it depends on variables that we want the user to configure then it should be in the typography feature. That way users can do something like this:

```sass
@include mdc-grid-list-core-styles(mdc-without(typography));

.normal-typography-section {
  @include mdc-grid-list-core-styles(typography);
}

.fancy-typography-section {
  $mdc-typography-font-weight-values: $fancy-typography-weights !global;
  @include mdc-grid-list-core-styles(typography);
}
```

Basically when they call the mixin with the `typography` feature as the query, they're asking "give me all of the style rules that might have changed as a result of my messing with the typography variables"

PR #4540, opened 2019-03-26
fix(tab): Fix tab icon color mixin to support SVG icons.

### LINKED ISSUE #4538: MDC Tabs: Icons turning black instead of purple.

> 1.0.1

> All browsers

### What OS are you using?

> Windows 10

1. Open [this codepen](https://codepen.io/Richienb/full/ZPwXMo).
2. Start changing tabs.

### What is the expected behaviour?

> The tab icons should change in accordance with the selected tab.

### What is the actual behaviour?

> The tab icons turn black instead of purple when selected.

Richienb:
This simple patch correctly sets the colour of the icon in an MDC Tab. This pull request fixes https://github.com/material-components/material-components-web/issues/4538.

There are probably better ways of doing this but this is mine.


Richienb:
I've fixed the build error - it was related to whitespace.
We are now ready to merge!

PR #4546, opened 2019-03-27
chore: Upgraded focus-trap in dialog & drawer that improves performance

stof:
See https://github.com/davidtheclark/focus-trap/blob/5.0.0/CHANGELOG.md for the upstream changelog.

The optimization applies when the dialog contains lots of tabbable nodes. In my case, this happened because the dialog contained a long list with checkboxes in each item, resulting in ~1300 checkboxes.
Opening the dialog blocked the main thread for 1.8s to activate the focus trap, which is down to 15ms for the same dialog with the updated version.

This results from 2 upstream optimizations:

- https://github.com/davidtheclark/focus-trap/issues/76 which avoids using tabbable twice during initialization (dividing the time by 2). MDC is not affected by the small BC break related to that change, as it does not rely on the `onActivate` callback of focus-trap
- https://github.com/davidtheclark/tabbable/pull/37 which changes the way to detect whether elements are displayed on screen or hidden, relying on the browser knowing that already rather than computing it for each element


stof:
is there anything blocking the review here ?

Issue #4140, opened 2018-12-03
[Shape] Text Field and Select's outline-shape-radius functions don't support separate corner values


kfranqueiro:
(This is true in 0.41.x and will still be true after #4035.)

The `mdc-text-field-outline-shape-radius` and `mdc-select-outline-shape-radius` mixins perform checks which assume that the given radius is a single value. However, most `*-shape-radius` mixins support multiple values to assign different sizes to different corners.

If you try to pass multiple radius values to one of these mixins, you'll get an error. For example, passing `8px 4px`:

On 0.41.x: `Error: 12px-8px 4px-1.2 is not a number.`

On the #4035 branch: `Error: Undefined operation "8px 4px > 12px".`

We should investigate whether we can support separate corner values in these mixins, or in the worst case, document that they only support one.

kfranqueiro:
I have WIP for this on [fix/outlined-shape-radius](https://github.com/material-components/material-components-web/compare/fix/outlined-shape-radius) (which should probably be split into 2 feature branches to PR, actually). I'm waiting on the outcome of #4171 to continue, though, as it might affect this, and I also realized we probably need to account for ensuring width on the trailing side as well (currently the code in master only accounts for this on the leading side).

kfranqueiro:
One of our design advocates has pointed out that it's rather disruptive to encounter this when setting values for entire shape categories at once, so in the nearer term, we should probably handle this with a warning and a graceful fallback to applying just one of the radius values.

Meanwhile, I've moved my branch to `feat/textfield-select-shape-radius`, updated it against master, added more tests, and fixed some issues I found. Still need to fix an issue of interoperability with trailing icon styles.

kfranqueiro:
Unfortunately, my shorter-term fix in #4547 has now been further set back by #4548, which causes another error due to a value being expanded earlier, resulting in an unexpected nested list. This has, however, caused me to step back and seek to confirm whether assigning multiple radius values to a single shape category is intended/valid. If so, there are probably some things our shape mixins/functions are currently doing (or, more particularly, the order they're being done in) that should be amended.

kfranqueiro:
I just updated the `feat/textfield-select-shape-radius` branch and resolved any conflicts or oversights from merging with master after #4547.

asyncLiz:
This has been fixed in https://github.com/material-components/material-components-web/pull/6163. Outlined textfield/select components now support setting individual corner radii values. Note that filled variants still only support top-left and top-right corners per spec.

PR #4547, opened 2019-03-27
fix(shape): Fix errors related to multi-value shape categories

kfranqueiro:
Refs #4140.

This PR changes the behavior from throwing an error and failing compilation, to warning about each encountered case and automatically falling back to using the first radius value. This is done in such a way as to avoid this problem even when multiple values are assigned at the shape variable level (i.e. centrally for small/medium/large).

I have a separate WIP branch to actually support multiple corner radii for notched-outline-related components (see the bug).

To test, assign multiple radii to the small/medium `$mdc-shape-...-component-radius` variable(s) in `mdc-shape/_variables.scss`.

### DIFF
--- a/packages/mdc-select/_mixins.scss
+++ b/packages/mdc-select/_mixins.scss
@@ -112,7 +112,7 @@
 
   @if (length(mdc-shape-prop-value($radius)) > 1) {
     // stylelint-disable-next-line max-line-length
-    @warn "mdc-select-outline-shape-radius currently only supports a single radius value. Follow https://github.com/material-components/material-components-web/issues/4140 for updates.";
+    @warn "mdc-select-outline-shape-radius only supports a single radius; see https://github.com/material-components/material-components-web/issues/4140";
   }
 
   .mdc-notched-outline {
--- a/packages/mdc-select/_mixins.scss
+++ b/packages/mdc-select/_mixins.scss
@@ -108,7 +108,12 @@
 }
 
 @mixin mdc-select-outline-shape-radius($radius, $rtl-reflexive: false) {
-  $resolved-radius: mdc-shape-resolve-percentage-radius($mdc-select-height, $radius);
+  $resolved-radius: nth(mdc-shape-resolve-percentage-radius($mdc-select-height, mdc-shape-prop-value($radius)), 1);
+
+  @if (length(mdc-shape-prop-value($radius)) > 1) {
+    // stylelint-disable-next-line max-line-length
+    @warn "mdc-select-outline-shape-radius currently only supports a single radius value. Follow https://github.com/material-components/material-components-web/issues/4140 for updates.";
+  }
 
   .mdc-notched-outline {
     @include mdc-notched-outline-shape-radius($resolved-radius, $rtl-reflexive);
@@ -118,13 +123,11 @@
     @include mdc-shape-radius($resolved-radius, $rtl-reflexive);
   }
 
-  $radius-pixels: mdc-shape-prop-value($resolved-radius);
-
-  @if ($radius-pixels > $mdc-notched-outline-leading-width) {
+  @if ($resolved-radius > $mdc-notched-outline-leading-width) {
     .mdc-select__native-control {
       @include mdc-rtl-reflexive-property(
         padding,
-        $radius-pixels + $mdc-notched-outline-padding,
+        $resolved-radius + $mdc-notched-outline-padding,
         $mdc-select-arrow-padding
       );
     }
@@ -132,7 +135,7 @@
     + .mdc-select-helper-text {
       @include mdc-rtl-reflexive-property(
         margin,
-        $radius-pixels + $mdc-notched-outline-padding,
+        $resolved-radius + $mdc-notched-outline-padding,
         $mdc-select-outline-label-offset
       );
     }
--- a/packages/mdc-shape/_functions.scss
+++ b/packages/mdc-shape/_functions.scss
@@ -57,6 +57,8 @@
 // mdc-shape-resolve-percentage-radius(36px, 50%) => `18px` (i.e., 36px / 2)
 //
 @function mdc-shape-resolve-percentage-radius($component-height, $radius) {
+  $radius: mdc-shape-prop-value($radius);
+
   @if type-of($radius) == "list" {
     $radius-value: ();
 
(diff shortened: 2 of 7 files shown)


kfranqueiro:
Looks like I missed replacing one instance of an old variable name. That should be fixed now.

packages/mdc-select/_mixins.scss
```diff
@@ -108,7 +108,12 @@
 }
 
 @mixin mdc-select-outline-shape-radius($radius, $rtl-reflexive: false) {
-  $resolved-radius: mdc-shape-resolve-percentage-radius($mdc-select-height, $radius);
+  $resolved-radius: nth(mdc-shape-resolve-percentage-radius($mdc-select-height, mdc-shape-prop-value($radius)), 1);
+
+  @if (length(mdc-shape-prop-value($radius)) > 1) {
+    // stylelint-disable-next-line max-line-length
+    @warn "mdc-select-outline-shape-radius currently only supports a single radius value. Follow https://github.com/material-components/material-components-web/issues/4140 for updates.";
```
moog16: Is there a way to put this on a new line? Its difficult to see this on my terminal:
Screen Shot 2019-03-28 at 12 09 13 PM

packages/mdc-textfield/_mixins.scss
```diff
@@ -217,21 +217,24 @@
 }
 
 @mixin mdc-text-field-outline-shape-radius($radius, $rtl-reflexive: false) {
-  $resolved-radius: mdc-shape-resolve-percentage-radius($mdc-text-field-height, $radius);
+  $resolved-radius: nth(mdc-shape-resolve-percentage-radius($mdc-text-field-height, mdc-shape-prop-value($radius)), 1);
+
+  @if (length(mdc-shape-prop-value($radius)) > 1) {
+    // stylelint-disable-next-line max-line-length
+    @warn "mdc-text-field-outline-shape-radius currently only supports a single radius value. Follow https://github.com/material-components/material-components-web/issues/4140 for updates.";
```
moog16: same with text field 
Screen Shot 2019-03-28 at 12 10 18 PM

packages/mdc-select/_mixins.scss
kfranqueiro: I've shortened the message as much as possible. I originally tried 2 separate `@warn` statements but that ends up printing the stack trace each time, which makes it effectively more confusing to read. I also tried `\n` which doesn't work as a newline.

kfranqueiro:
After merging from master after #4548 was merged, I encountered new errors. I've now resolved those, but it involved some surgery to the order of operations performed in the mdc-shape functions. The end result is that mdc-shape now better supports categories having multi-radius values (whereas our current code was clearly assuming they'd be only a single radius value).

packages/mdc-shape/_functions.scss
```diff
@@ -70,39 +72,10 @@
   }
 }
 
-@function mdc-shape-resolve-percentage-for-corner_($component-height, $radius) {
```
kfranqueiro: I moved this function lower in the file verbatim, to group internal functions together.

packages/mdc-shape/_functions.scss
```diff
@@ -70,39 +72,10 @@
...
-
-//
-// Strips unit from number. This is accomplished by dividing the value by itself to cancel out units, while resulting
-// in a denominator of 1.
-//
-// Examples:
-//
-// 50% => 50
-//
-@function mdc-shape-strip-unit_($number) {
```
kfranqueiro: This function was apparently not being used whatsoever, so I removed it.

packages/mdc-shape/_functions.scss
```diff
@@ -114,15 +87,26 @@
       @error "Invalid radius: '#{$radius}' is more than 4 values";
     }
 
-    $radius-value: ();
+    $radius-values: ();
 
-    @each $corner in $radius {
-      $radius-value: append($radius-value, mdc-shape-prop-corner-value_($corner));
```
kfranqueiro: I ended up merging the logic of `mdc-shape-prop-corner-value_` into this function after some of the finagling I did led to this function otherwise pretty much not doing anything important in itself besides validating list length.

packages/mdc-shape/_functions.scss
```diff
@@ -114,15 +87,27 @@
...
+    @for $i from 1 through length($radius) {
+      $corner: nth($radius, $i);
+
+      @if map-has-key($mdc-shape-category-values, $corner) {
+        // If a category is encountered within a list of radii, apply the category's value for the corresponding corner
```
kfranqueiro: This behavior maximizes compatibility with the existing behavior which clearly expected shape categories to be applicable to individual corners (e.g. `small small 0 0`). This will effectively behave identically to the existing behavior in cases where a shape category is a single radius, while making list-of-radii category values actually functional.

packages/mdc-shape/_functions.scss
```diff
@@ -147,34 +132,50 @@
...
+    // Converts the percentage to number without unit. Example: 50% => 50.
+    $percentage: $radius / ($radius * 0 + 1);
+
+    @return $component-height * ($percentage / 100);
+  } @else {
+    @return $radius;
+  }
+}
+
+@function mdc-shape-validate-radius-value_($radius) {
```
kfranqueiro: This function has been rewritten to emit an error, akin to the `verify` functions in `feature-targeting`, so that there's less code to write in places that call this.

packages/mdc-shape/_functions.scss
```diff
@@ -114,15 +87,27 @@
...
 
-    $radius-value: ();
+    $radius-values: ();
 
-    @each $corner in $radius {
-      $radius-value: append($radius-value, mdc-shape-prop-corner-value_($corner));
+    @for $i from 1 through length($radius) {
+      $corner: nth($radius, $i);
+
+      @if map-has-key($mdc-shape-category-values, $corner) {
```
abhiomkar: What happens if `small` value is `4px 0 4px 0` and I apply this mixin with this value - `mdc-button-shape-radius(small 0 small 0)`?

I'm OK not supporting something like `mdc-button-shape-radius(small 0 small 0)` if that makes the implementation complex.

packages/mdc-shape/_functions.scss
```diff
@@ -147,34 +132,50 @@
...
 }
 
-@function mdc-shape-prop-corner-value_($radius) {
-  @if map-has-key($mdc-shape-category-values, $radius) {
-    @return map-get($mdc-shape-category-values, $radius);
-  } @else if mdc-shape-is-valid-radius-value_($radius) {
+//
+// Unpacks shorthand values for border-radius (i.e. lists of 1-3 values).
+// If a list of 4 values is given, it is returned as-is.
+// TODO: This is private for purposes of getting it into a patch; make it public for a future minor/major release.
```
abhiomkar: Please add examples in the description.

packages/mdc-shape/_functions.scss
```diff
@@ -70,39 +72,10 @@
...
-  }
-
-  @return $number;
-}
-
 //
 // Returns $radius value of shape category - `large`, `medium` or `small`.
 // Otherwise, it returns the $radius itself if valid.
-// $radius can be a single value or list of up to 4.
+// $radius can be a single value, or a list of up to 4 values.
```
abhiomkar: Update examples section in the description with new behaviour.

packages/mdc-shape/_functions.scss
kfranqueiro: That would result in `4px 0 4px 0`. It maps each respective corner from the shape category.

packages/mdc-shape/_functions.scss
```diff
@@ -114,15 +87,27 @@
       @error "Invalid radius: '#{$radius}' is more than 4 values";
     }
 
-    $radius-value: ();
+    $radius-values: ();
 
-    @each $corner in $radius {
-      $radius-value: append($radius-value, mdc-shape-prop-corner-value_($corner));
+    @for $i from 1 through length($radius) {
```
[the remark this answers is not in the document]
kfranqueiro: I need the index to correlate between the given radius list and the full-expanded list of radii for any categories that are referenced.

PR #4549, opened 2019-03-28
chore: Update golden screenshots after browser upgrade

abhiomkar:


kfranqueiro:
That's interesting... a couple of these diffs show ligatures no longer applying. Think I'll point that out to someone on fonts.

On the bright side, it looks like this also fixes a flake we saw Chrome having in some of our dialog cases, so that's great!

PR #4558, opened 2019-04-01
feat(list): Add disabled class name to constants

abhiomkar:


packages/mdc-list/constants.ts
```diff
@@ -24,6 +24,7 @@
 const cssClasses = {
   LIST_ITEM_ACTIVATED_CLASS: 'mdc-list-item--activated',
   LIST_ITEM_CLASS: 'mdc-list-item',
+  LIST_ITEM_DISABLED_CLASS: 'mdc-list-item--disabled',
```
kfranqueiro: Is there a place we should be using this? Hard to sell me on adding a constant if it's unused...

PR #4563, opened 2019-04-02
fix(list): Add cursor: pointer for interactive list items

### LINKED ISSUE #4557: mcd-list does not use a hover pointer

1.1.0

Chrome, Edge, IE, and Firefox.

### What OS are you using?

Windows 10

It's reproducible on the material demo site, and I created a codepen as well:
* https://material-components.github.io/material-components-web-catalog/#/component/list
* https://codepen.io/seriema/pen/gybOxe

1. Load a list of items
2. Hover over any list item

The mouse cursor should become a pointer, as described in Material Guidelines:
https://material.io/design/interaction/states.html#hover
expected

The mouse cursor is either an arrow or caret, depending on if I'm hovering on the item text or other part of the item.

joyzhong:
Fixes #4557

### DIFF
--- a/packages/mdc-list/_mixins.scss
+++ b/packages/mdc-list/_mixins.scss
@@ -78,6 +78,11 @@
     }
   }
 
+  // Only enable mouse hover for interactive list items which are not disabled.
+  :not(.mdc-list--non-interactive) > .mdc-list-item:not(.mdc-list-item--disabled) {
+    cursor: pointer;
+  }
+
   // "Selected" is ephemeral and likely to change soon. E.g., selecting one or more photos to share in Google Photos.
   // "Activated" is more permanent. E.g., the currently highlighted navigation destination in a drawer.
   .mdc-list-item--selected,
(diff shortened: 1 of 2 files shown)


packages/mdc-list/_mixins.scss
```diff
@@ -78,6 +78,11 @@
     }
   }
 
+  // Only enable mouse hover for interactive list items which are not disabled.
+  :not(.mdc-list--non-interactive) > .mdc-list-item:not(.mdc-list-item--disabled) {
```
kfranqueiro: I'd be inclined to match the selector we use in the `mdc-list-ripple` mixin here, which also has `.mdc-list-item` on the last compound selector (to avoid applying this to every child of the list, which could include other elements such as dividers).

packages/mdc-list/_mixins.scss
```diff
@@ -78,6 +78,11 @@
     }
   }
 
+  // Only enable mouse hover for interactive list items which are not disabled.
```
kfranqueiro: Nit: enable mouse hover -> change mouse cursor

packages/mdc-list/_mixins.scss
joyzhong: Changed the order in the other selector instead, as discussed.

joyzhong: Switched back to original order (and updated here to be consistent) with .mdc-list-item last, since we use &--activated, etc. in `mdc-list-ripple`, forcing .mdc-list-item to go last.

PR #4565, opened 2019-04-03
fix(chips): Fix incorrect ripple effect on filter chip

### LINKED ISSUE #2812: Incorrect ripple effect on filter chip with no leading icon
## Bugs

The filter chip with no leading icon has incorrect ripple when clicked. Instead of starting the ripple where clicked the ripple center is always in the same spot.

> 0.35.2

> All

### What OS are you using?

> Mac OS

> 1. Run the demo server
> 2. Go to localhost:8080/chips.html
> 3. Click on a Filter chip with no leading icon 
> 4. Observe the component's behavior
> 5. Helps to see if you set animations to 10%

> The ripple should start where the mouse is clicked on the chip. 

> The ripple effect is starting in the same spot and not where the chip is clicked.

Incorrect behavior
bad-chip

Correct behavior
correct-ripple

TroyTae:
Fixes #2812

### DIFF
--- a/packages/mdc-chips/chip/foundation.ts
+++ b/packages/mdc-chips/chip/foundation.ts
@@ -114,12 +114,12 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
         // Object.assign combine properties of object.
         // But ClientRect has only getters, so I added code below.
         const rect = {
-          width: rootRect.width,
+          bottom: rootRect.bottom,
           height: rootRect.height,
-          top: rootRect.top,
           left: rootRect.left,
           right: rootRect.right,
-          bottom: rootRect.bottom,
+          top: rootRect.top,
+          width: rootRect.width,
         };
         return {...rect, width, height};
       }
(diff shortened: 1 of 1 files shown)


packages/mdc-chips/chip/foundation.ts
```diff
@@ -110,7 +110,18 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
         // when adding it to the root client rect's width.
         const checkmarkWidth = checkmarkRect.height;
         const width = rootRect.width + checkmarkWidth;
-        return {...rootRect, width, height};
+        // Spread expression translate to Object.assign in Typescript.
```
abhiomkar: Good catch!

You can directly return the object with calculated height & width, so we don't need to use spread operator anymore.

```
return rect = {
  bottom: rootRect.bottom,
  height: rootRect.height,
  left: rootRect.left,
  right: rootRect.right,
  top: rootRect.top,
  width: width,
};
```

abhiomkar: We don't need this line anymore

```
const height = rootRect.height;
```

TroyTae: Is there any special meaning this syntax? ```return rect = {};```
I guess you wanna talk this: ```return {}```
So I reflect it :)

TroyTae: Thank you for your review!

abhiomkar: Sorry for the typo. Yes I meant `return {}`.

PR #4568, opened 2019-04-03
fix(list): Include disabled list items in keyboard navigation and allow focus

### LINKED ISSUE #4561: Disabled menu items should be focusable
Per [the menu README](https://github.com/material-components/material-components-web/blob/797ebc89e97348d5aac80616f70581c7b0f76a13/packages/mdc-menu/README.md):

> ## Disabled Menu Items
> Menu items can be disabled by adding the `mdc-list-item--disabled` modifier class (from MDC List). Disabled menu items will be excluded from keyboard navigation.`

However, the [ARIA practices documentation](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) states that disabled menu items must be focusable (i.e. keyboard navigable), though not activatable (see the first point under the green "NOTE" section).

See also [further elaborations on ARIA-practices on handling disabled elements](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls).

abhiomkar:
Fixes #4561 

As per ARIA practice recommendation disabled menu / list items are focusable but cannot be activated. It should only be focused but should not have other states such as activated / pressed.

Please see [ARIA Menu > Keyboard Interactions](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) & [Focusability of disabled controls](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls)

BREAKING CHANGE: MDCList's `listElements` component API now includes disabled list items which previously returned only enabled list items.


packages/mdc-list/component.ts
```diff
@@ -37,7 +37,7 @@ export class MDCList extends MDCComponent<MDCListFoundation> {
   }
 
   get listElements(): Element[] {
-    return [].slice.call(this.root_.querySelectorAll(strings.ENABLED_ITEMS_SELECTOR));
```
moog16: remove this constant since its not used anywhere

PR #4569, opened 2019-04-04
fix(textfield): Fix leading/trailing IconButton position in TextField

### LINKED ISSUE #4059: MDC Textfield: Add support for leading / trailing icon button in text field
The actionable leading / trailing icons needs to ripple. Using icon button seems appropriate. Position is bit off when rendering icon button inside text field as leading / trailing icons.

[Demo here.](https://stackblitz.com/edit/mdc-textfield-demo-wcc1cx?file=index.html)

image

sdomenet:
Fixes #4059


sdomenet:
CLA signed

PR #4575, opened 2019-04-08
fix(icon-button): Fix position when used as TextFieldIcon

### LINKED ISSUE #4059: MDC Textfield: Add support for leading / trailing icon button in text field

ghost:
Fixes #4059 

This PR fixes the position issue that occurs when an IconButton is used as a TextFieldIcon.
It does so by adding negative margin equal to the padding to the IconButton in this particular situation.


packages/mdc-icon-button/_mixins.scss
```diff
@@ -87,6 +87,10 @@
     height: $height + $padding * 2;
     padding: $padding;
     font-size: max($width, $height);
+
+    &.mdc-text-field__icon {
```
abhiomkar: This is applying the styles based on the context. It should be other way around. That is, parent component needs to apply styles for sub component.

Also, I feel the leading / trailing icon should be center aligned no matter the size of the icon.

For example,

```css
.mdc-text-field__icon {
  ...
  top: 50%;
  transform: translateY(-50%);
}
```

But we also need to consider adding enough (left / right based on the position of icon) padding to input element to avoid input overlapping with icon.

sdomenet: > This is applying the styles based on the context. It should be other way around. That is, parent component needs to apply styles for sub component.
This is what i wanted to do initially. But because i chose to fix the issue using negative margin, i needed the padding value from the icon button and switched things around. Should I 

> Also, I feel the leading / trailing icon should be center aligned no matter the size of the icon.
Isn't the size of the icon strictly defined in the spec?

> But we also need to consider adding enough (left / right based on the position of icon) padding to input element to avoid input overlapping with icon.
That's something i am looking at

abhiomkar: IMO, it is not safe to assume if the icon size will be fixed. For example, in future icon button may introduce dense variant which is smaller than standard variant.

abhiomkar:

@sdomenet we can always reopen it though! :)

sdomenet:
I would like to submit another commit to solve that issue, can I do that here or should I create a new PR?

abhiomkar:
Please commit to same PR. Thanks!

moog16:
@sdomenet would you be able to add a screenshot test for this?

moog16 (changes requested):
I created some screenshot tests for you. The dense-leading-mdc-icon.html is failing. Please a search for `mdc-textfield__icon` under `packages/mdc-textfield`. You should also add cases for mdc-icon-button in most/all of those areas. 


[classes.tar.gz](https://github.com/material-components/material-components-web/files/3203835/classes.tar.gz)

anlexN:
can we solve PR?

PR #4582, opened 2019-04-09
fix(menu): Incorrect menu positioning when first click

TroyTae:
스크린샷 2019-04-09 오후 12 20 34
> When you first click on the menu, dropdown is positioning incorrect.

스크린샷 2019-04-09 오후 12 20 47
> But it works find after second click.

스크린샷 2019-04-09 오후 12 27 32
The position of the dropdown goes inline style.
But when first click, there is no top/left value.
In this case dropdown make inner scroll and position calculated incorrect.
So how about give initial position?

### DIFF
--- a/packages/mdc-menu-surface/_mixins.scss
+++ b/packages/mdc-menu-surface/_mixins.scss
@@ -95,6 +95,8 @@
   @include mdc-feature-targets($feat-structure) {
     display: none;
     position: absolute;
+    top: 0;
+    left: 0;
     box-sizing: border-box;
     max-width: calc(100vw - #{$mdc-menu-surface-min-distance-from-edge});
     max-height: calc(100vh - #{$mdc-menu-surface-min-distance-from-edge});
--- a/packages/mdc-menu-surface/_mixins.scss
+++ b/packages/mdc-menu-surface/_mixins.scss
@@ -95,8 +95,6 @@
   @include mdc-feature-targets($feat-structure) {
     display: none;
     position: absolute;
-    top: 0;
-    left: 0;
     box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


packages/mdc-menu-surface/_mixins.scss
```diff
@@ -95,6 +95,8 @@
   @include mdc-feature-targets($feat-structure) {
     display: none;
     position: absolute;
+    top: 0;
```
abhiomkar: The positions may change based on where the menu surface is anchored and the orientation of menu surface based on viewport size. So I think, it is not safe to hard code the values only for `top` & `left`.

TroyTae: Oh I see maybe I need to get position of menu before applying className "--animating-open".

TroyTae: I executed autoPosition_ before applying class.
And I moved dimensions setter into autoautoPosition_.
Check my code again please :)

abhiomkar: Thanks for the changes! Somehow I'm not able to reproduce the original issue on master branch.

Here is the demo:
https://dist-qw5tsda9o.now.sh/

But I'm still able to reproduce this on catalog site with latest version. I'm curious, why it calculates the position correctly outside the catalog.

TroyTae: @abhiomkar 
I guess this issue is fired when there is inner scroll.
Because select's menu surface appends to body.

스크린샷 2019-04-17 오전 2 46 15
And catalog website has inner scroll.

TroyTae: @abhiomkar 
Can you check this please?

PR #4584, opened 2019-04-09
chore: Remove repeated license comments from minified source code

### LINKED ISSUE #4220: Include copyright notice only once in final build file
Currently the `@license` annotation is repeated in compiled & minified file [material-components-web.min.js](https://unpkg.com/material-components-web@0.42.1/dist/material-components-web.min.js). The copyright notice should only be included once per file in minified and un-minified file.

abhiomkar:
**File size comparison**

Name | Before | After
-- | -- | --
material-components-web.js | 840K | 708K
material-components-web.min.js | 464K | 296K

**File size comparison (w/ gzip compression)**

Name | Before | After
-- | -- | --
material-components-web.js | 96.64KB | 72.68KB
material-components-web.min.js | 53.01KB | 48.18KB

Fixes #4220

### DIFF
--- a/scripts/webpack/copyright-banner-plugin.js
+++ b/scripts/webpack/copyright-banner-plugin.js
@@ -30,15 +30,15 @@
 const webpack = require('webpack');
 
 class CopyrightBannerPlugin extends webpack.BannerPlugin {
-  constructor({projectName, authorName, licenseName}) {
+  constructor() {
     super({
-      banner: [
-        '/*!',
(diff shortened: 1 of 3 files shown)


moog16:
The `material-components-web/dist/material-components-web.js` file still contains 157 instances of `@license` in this file. I think to actually fix this bug we need to remove those. I removed all the @licenses from that file and it reduces the filesize from 861kb to 684kb.

Issue #3922, opened 2018-10-16
MDCMenu: Parameterize whether menu automatically focuses the first item


KNawm:
### What MDC Web Version are you using?

0.40.1

### What browser(s) is this bug affecting?

Tested on Chrome, Chrome for Android and Firefox

### What OS are you using?

Windows 10

### What are the steps to reproduce the bug?

1. Run the demo server
2. Go to localhost:8080/menu.html
3. Open any menu
4. Observe the component's behavior

### What is the expected behavior?

In Material examples of behavior it doesn't selects the first item in a menu on opening.
![](https://i.imgur.com/o83ucou.gif)

### What is the actual behavior?

It automatically selects the first item
![](https://i.imgur.com/w1L3352.gif)

### Any other information you believe would be useful?

I don't know if this is intended behavior but I tested opening a menu and navigating with the keyboard arrows and if it is intended it should be blank and only select the first item if a user uses the keys or other form of navigating to select the appropriate option.

MichaelTamm:
Related bug: Although the focus can be moved via arrow keys, once the menu is open, pressing the [Tab] key closes the menu. (Expected behaviour: The next menu item should be focused.)

acdvorak:
@KNawm: Thanks for filing this issue! We'll reach out to Design to get clarification.

@MichaelTamm: The behavior you're seeing—where pressing the <kbd>Tab</kbd> key closes the menu—is actually intentional. Menus are only supposed to have one tab stop, and the <kbd>↑</kbd> and <kbd>↓</kbd> arrow keys are used to focus the previous/next menu item. This mirrors how menus behave in Google Docs, macOS, and Windows, for example.

AlexZarivayskiy:
Perhaps, this is a temporary solution, until that issue will be solved.
```
HTML
...
<!-- Add my-menu-class -->
<div class="mdc-menu mdc-menu-surface my-menu-class" tabindex="-1">
...


CSS
.my-menu-class :not(.mdc-list--non-interactive)>:not(.mdc-list-item--disabled).mdc-list-item::before, 
.my-menu-class :not(.mdc-list--non-interactive)>:not(.mdc-list-item--disabled).mdc-list-item::after
{
	background-color: #fff;
}
.my-menu-class .mdc-list-item:hover {
	background-color: #f0f0f0;
}
```

kfranqueiro:
[...]

PR #4587, opened 2019-04-09
feat(menu): Focus management features & accessibility improvements

### LINKED ISSUE #3760: [List] Not able to arrow through the list when focus is on list root.
Steps to reproduce:
1. Render interactive list.
2. Focus on list root element.
3. Press DOWN arrow.

Actual result:
Nothing gets focused.

Expected result:
First list item needs to be focus when DOWN key is pressed. & Last list item needs to be focused on UP key.

[Codepen](https://codepen.io/abhiomkar/pen/dgZJNX).

### LINKED ISSUE #4513: MDC Menu: Implement option to disable focus grabbing
Follow-on from https://github.com/material-components/material-components-web/issues/3922#issuecomment-465618844.

MDC Menu can now be parameterized to affect focus behavior, and the default is now to focus the root of the menu rather than a specific item. However, there is currently no way to avoid focusing the menu altogether.

(Note that this should be used VERY sparingly, and is intended for cases where a menu is opened from another control that should retain focus, e.g. a text field.)

abhiomkar:
## Changes

- Focus on list root element instead of menu/menusurface root element when menu is opened. List root has appropriate ARIA attributes such as role="menu" and aria-label.
- Moved focus logic from menu to list since list root may have focus on menu open.
- Introducing new enum (TypeScript) to set default focus state every time when the menu is opened.

```ts
enum DefaultFocusState {
  NONE = 0,
  LIST_ROOT = 1,
  FIRST_ITEM = 2,
  LAST_ITEM = 3,
}
```

> See Accessibility section of Menu README file for intended focus behaviour on menu open.

Fixes #4513 
Fixes #3760
Refs #3922 

BREAKING CHANGE: New adapter methods to MDC List: `isRootFocused`. MDC Menu: Replaced adapter methods `isRootFocused`, `focusRoot` with `focusListRoot`. When using MDC List inside MDC Menu `tabindex` should be set on list root element where `role="menu"` is assigned.


packages/mdc-list/_mixins.scss
```diff
@@ -409,6 +409,10 @@
     margin: 0;
     padding: 8px 0;
     list-style-type: none;
+
+    &:focus {
```
moog16: curious why we didn't need this before?

packages/mdc-list/component.ts
```diff
@@ -237,10 +238,7 @@ export class MDCList extends MDCComponent<MDCListFoundation> {
   private handleKeydownEvent_(evt: KeyboardEvent) {
     const index = this.getListItemIndex_(evt);
     const target = evt.target as Element;
-
-    if (index >= 0) {
```
moog16: why do you not need the check anymore?

packages/mdc-list/foundation.ts
```diff
@@ -177,9 +178,20 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
     const isEnter = evt.key === 'Enter' || evt.keyCode === 13;
     const isSpace = evt.key === 'Space' || evt.keyCode === 32;
 
+    if (this.adapter_.isRootFocused()) {
+      if (arrowUp || isEnd) {
```
moog16: side comment: why aren't the arrow* --> isArrow*?

packages/mdc-list/foundation.ts
```diff
@@ -177,9 +178,20 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
...
+        this.focusFirstElement();
+      }
+
+      return;
+    }
+
     let currentIndex = this.adapter_.getFocusedElementIndex();
-    let nextIndex = numbers.UNSET_INDEX;
-    if (currentIndex === numbers.UNSET_INDEX) {
+    if (currentIndex === -1) {
```
moog16: why not just leave the -1 to `UNSET_INDEX`? Should we just get rid of `UNSET_INDEX` from constants?

packages/mdc-menu/README.md
```diff
@@ -198,9 +198,16 @@ With focus on the menu button:
...
   * <kbd>Up Arrow</kbd> opens the menu and moves focus to the last menu item.
-  * The focus is set to menu root element when clicked or touched. Menu handles the keyboard handling once it is opened and focused on root element.
+  * The focus is set to list root element (where `role="menu"` is set) when clicked or touched. MDC List handles the keyboard navigation once it receives the focus.
+
+Use `setDefaultFocusState` method to set default focus state that will be focused every time when menu is opened.
 
-Use `setDefaultFocusItemIndex` method to set the index of the menu item that will be focused every time the menu opens. Set it to `numbers.FOCUS_ROOT_INDEX` to focus on menu root element.
+Focus state | Description
+--- | ---
+`DefaultFocusState.FIRST_ITEM` | Focuses the first menu item. Set this when menu button receives <kbd>Enter</kbd>, <kbd>Space</kbd>, <kbd>Down Arrow</kbd>.
```
moog16: is kbd the same as the backticks (`)?

packages/mdc-menu/foundation.ts
```diff
@@ -120,30 +106,30 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
...
+        this.adapter_.focusListRoot();
+        break;
+      case DefaultFocusState.FIRST_ITEM:
+        this.adapter_.focusItemAtIndex(0);
+        break;
+      case DefaultFocusState.LAST_ITEM:
+        this.adapter_.focusItemAtIndex(this.adapter_.getMenuItemCount() - 1);
+        break;
+      case DefaultFocusState.NONE:
+      default:
```
moog16: do you need this default since it does nothing? Or should the `this.adapter_.focusListRoot();` go here instead?

test/screenshot/spec/mdc-menu/fixture.js
```diff
@@ -29,7 +31,7 @@ window.mdc.testFixture.fontsLoaded.then(() => {
   menu.open = true;
 
   buttonEl.addEventListener('click', () => {
-    menu.setDefaultFocusItemIndex(mdc.menu.MDCMenuFoundation.numbers.FOCUS_ROOT_INDEX);
+    menu.setDefaultFocusState(DefaultFocusState.LIST_ROOT);
```
moog16: Why do you need this if it is the default?

test/unit/mdc-list/foundation.test.js
```diff
@@ -215,6 +215,43 @@ test('#handleFocusOut sets tabindex=0 to first selected index when focus leaves
...
+  const preventDefault = td.func('preventDefault');
+  const event = {key: 'ArrowUp', preventDefault};
+
+  td.when(mockAdapter.isRootFocused()).thenReturn(true);
+  td.when(mockAdapter.getListItemCount()).thenReturn(5);
+  foundation.handleKeydown(event, false, -1);
+
+  td.verify(preventDefault(), {times: 1});
+  td.verify(mockAdapter.focusItemAtIndex(4), {times: 1});
+  td.verify(mockAdapter.getFocusedElementIndex(), {times: 0});
```
moog16: this verify is covered in the above test case.

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -256,38 +224,46 @@ test('handleItemAction item action event inside of a child element of a selectio
...
-test('setDefaultFocusItemIndex focuses the root element on menu open when set to numbers.FOCUS_ROOT_INDEX', () => {
+test('handleMenuSurfaceOpened focuses the list root element when DefaultFocusState is set to LIST_ROOT', () => {
   const {foundation, mockAdapter} = setupTest();
 
-  td.when(mockAdapter.getMenuItemCount()).thenReturn(5);
-  foundation.setDefaultFocusItemIndex(menuNumbers.FOCUS_ROOT_INDEX);
+  foundation.setDefaultFocusState(DefaultFocusState.LIST_ROOT);
   foundation.handleMenuSurfaceOpened();
-  td.verify(mockAdapter.focusRoot());
+  td.verify(mockAdapter.focusListRoot());
```
moog16: I think you want a {times: 1} here

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -256,38 +224,46 @@ test('handleItemAction item action event inside of a child element of a selectio
...
 });
 
-test('handleMenuSurfaceOpened menu focuses the root element by default on menu surface opened', () => {
+test('handleMenuSurfaceOpened menu focuses the list root element by default on menu surface opened', () => {
   const {foundation, mockAdapter} = setupTest();
 
-  td.when(mockAdapter.getMenuItemCount()).thenReturn(5);
   foundation.handleMenuSurfaceOpened();
-  td.verify(mockAdapter.focusRoot());
+  td.verify(mockAdapter.focusListRoot());
```
moog16: add a {times: 1} here

moog16:
Looks good overall - just some minor comments/questions

packages/mdc-list/_mixins.scss
abhiomkar: We didn't have `tabindex` on the list root before to be on focus.

packages/mdc-list/component.ts
abhiomkar: Now that list handles the keyboard events on the root, we no longer need to limit the keyboard handle to list items.

packages/mdc-list/foundation.ts
abhiomkar: `UNSET_INDEX` is only relevant for `selectedIndex`, whereas `currentIndex` refers to the current list item where the keydown is being handled.

We still need `UNSET_INDEX` constant to reference as the list has no selection.

packages/mdc-menu/README.md
abhiomkar: `kbd` is for keyboard keys, where as backticks is used when referencing code.

image

packages/mdc-menu/foundation.ts
abhiomkar: Good catch. Done!

test/screenshot/spec/mdc-menu/fixture.js
abhiomkar: The previous `handleMenuSurfaceOpened` calls in `keydown` callback below overrides the default focus state. It is required every time before menu is opened.

test/unit/mdc-list/foundation.test.js
abhiomkar: Added this to make sure if it returns early.

packages/mdc-list/foundation.ts
```diff
@@ -168,18 +169,29 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
    * Key handler for the list.
    */
   handleKeydown(evt: KeyboardEvent, isRootListItem: boolean, listItemIndex: number) {
-    const arrowLeft = evt.key === 'ArrowLeft' || evt.keyCode === 37;
-    const arrowUp = evt.key === 'ArrowUp' || evt.keyCode === 38;
-    const arrowRight = evt.key === 'ArrowRight' || evt.keyCode === 39;
-    const arrowDown = evt.key === 'ArrowDown' || evt.keyCode === 40;
+    const isArrowLeft = evt.key === 'ArrowLeft' || evt.keyCode === 37;
```
moog16: nice - i was going to comment on this, but it wasn't code you wrote :)

PR #4590, opened 2019-04-11
fix(select): Fix enhanced select issue where it does not stay open on long press #4173

### LINKED ISSUE #4173: Enhanced select menu doesn't stay open

moog16:
continuation of https://github.com/material-components/material-components-web/pull/4413
fixes #4173

### DIFF
--- a/packages/mdc-select/component.ts
+++ b/packages/mdc-select/component.ts
@@ -37,9 +37,6 @@ import {MDCSelectHelperText, MDCSelectHelperTextFactory} from './helper-text/ind
 import {MDCSelectIcon, MDCSelectIconFactory} from './icon/index';
 import {MDCSelectEventDetail, MDCSelectFoundationMap} from './types';
 
-type PointerEventType = 'click';
-
-const POINTER_EVENTS: PointerEventType[] = ['click'];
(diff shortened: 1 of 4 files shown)


packages/mdc-select/component.ts
```diff
@@ -37,9 +37,9 @@ import {MDCSelectHelperText, MDCSelectHelperTextFactory} from './helper-text/ind
 import {MDCSelectIcon, MDCSelectIconFactory} from './icon/index';
 import {MDCSelectEventDetail, MDCSelectFoundationMap} from './types';
 
-type PointerEventType = 'mousedown' | 'touchstart';
+type PointerEventType = 'click';
```
abhiomkar: We not longer need `POINTER_EVENTS` & `PointerEventType` since this is only single event, we can directly add register it to handleClick foundation method without looping through `POINTER_EVENTS`.

moog16:
@shelart could you comment here and say "I signed it" please

shelart:
I signed it

Issue #4571, opened 2019-04-05
MDC Ripple before/after background color should be set to currentColor


moog16:
Ripple's before/after background color is set via `mdc-states-interactions_` sass mixin. This is used 3 times in the baseline, activated, and selected mixins. This has reportedly has caused incorrect ripple colors when theming. 

The proposed solution is to replace `@include mdc-theme-prop(background-color, $color, $edgeOptOut: true);` found in the `mdc-states-base-color` mixin with `background-color: currentColor`. 

An extra win for us is that it will decrease the bundled css size of ripple, and therefore decreasing the css of all components using ripple.

jamesmfriedman:
Thanks for recording this Matt. I can verify that the `currentColor` method works in every situation I've come across. In my own downstream apps, I now just have it set as a global override with `!important`.

abhiomkar:
I just realized that not all block elements have `color` CSS property to inherit to ripple `background-color`. 

Block elements that don't have text or font icons not necessarily has `color` CSS property, for example the thumb overlay in switch component.

Current implementation:

![image](https://user-images.githubusercontent.com/47312/55997629-2a367e00-5c89-11e9-8baa-78ce50414c2d.png)

with `currentColor`:

![image](https://user-images.githubusercontent.com/47312/55997599-19860800-5c89-11e9-906e-072f0b815655.png)

The ripple color is grey instead of primary color because the currentColor is inherited from its parent.

jamesmfriedman:
That seems trivially easy to fix by just applying color in css to these elements.

moog16:
Yes, but ripple won't just "work out of the box" then. Although you can consider the unbounded ripple on another surface element (ie radio, checkbox) advanced usage. If we go this route, we would require the users with ripple on another surface to use another mixin.

jamesmfriedman:
Maybe I’m missing something, but why don’t you just specify the color property on the radio, checkbox, and switch? Then you don’t need any special behavior. 

Or, make the mixin where it defaults to currentColor but accepts an override for situations like this (I know there are default argument values on SASS). 

I don’t think the fact that they’re unbounded is what makes them special since icon buttons can have colored icons which the ripple should respond to.

moog16:
Ya I'm sure there is a good way around that. I will figure something out :)

bonniezhou:
Closing this, infeasible due to comments above.

jamesmfriedman:
[...]

Issue #4527, opened 2019-03-21
[TopAppBar] Icon buttons in Card have the class `mdc-icon-button` while the buttons in Top App Bar don't


tychenjiajun:
It's strange and confusing.
[Card](https://material.io/develop/web/components/cards/)
```html
<div class="mdc-card__action-icons">
   <button class="material-icons mdc-icon-button mdc-card__action mdc-card__action--icon" title="Share">share</button>
    <button class="material-icons mdc-icon-button mdc-card__action mdc-card__action--icon" title="More options">more_vert</button>
  </div>
```

[Top App Bar](https://material.io/develop/web/components/top-app-bar/)
```html
<section class="mdc-top-app-bar__section mdc-top-app-bar__section--align-end" role="toolbar">
      <a href="#" class="material-icons mdc-top-app-bar__action-item" aria-label="Download" alt="Download">file_download</a>
      <a href="#" class="material-icons mdc-top-app-bar__action-item" aria-label="Print this page" alt="Print this page">print</a>
      <a href="#" class="material-icons mdc-top-app-bar__action-item" aria-label="Bookmark this page" alt="Bookmark this page">bookmark</a>
    </section>
```

moog16:
@tychenjiajun the Card is just a wrapper component, which means we didn't build in icon styles. This means we have to use mdc-icon-button in the Card. 

However looking at `@mixin mdc-top-app-bar-icon_` is very similar to the `@mixin mdc-icon-button-base_`. And this might be a good opprotunity to combine the 2 mixins and use mdc-icon-button instead and remove the `mdc-top-app-bar-icon` mixin. I will wait for others on the team to chime in.

moog16:
We will look into making top app bar support icon button, and possibly deprecate the `mdc-top-app-bar-icon_` mixin in the future.

moog16:
Closing. Merged #4745

PR #4592, opened 2019-04-11
fix(ripple): default to using background-color: currentColor for before/after ripple

moog16:
fixes https://github.com/material-components/material-components-web/issues/4571


moog16:
The ripple color in button should be the same as the text color as shown in this glitch: https://glitch.com/edit/#!/various-ocelot?path=index.html:9:0. In the first example if we change the CSS var, --mdc-theme-primary, it updates both text and ripple color. In the second example, If we just change the ink color it doesn't update the ripple color.

moog16:
@abhiomkar while fixing #4527 I'm finding that I need to customize the ripple. If I have background: currentColor, I wouldn't have to customize the ripple color. Can we pull this one out of the icebox?

packages/mdc-ripple/_mixins.scss
```diff
@@ -184,18 +184,22 @@
...
+        } @else {
+          // If a color with 0 alpha is specified, don't render the ripple pseudo-elements at all.
+          // This avoids unnecessary transitions and overflow.
+          content: none;
+        }
       } @else {
-        // If a color with 0 alpha is specified, don't render the ripple pseudo-elements at all.
-        // This avoids unnecessary transitions and overflow.
-        content: none;
+        background-color: currentColor;
```
abhiomkar: Seems like unsafe assumption to use inherited color for ripple where the inherited color is unknown.

moog16: In the majority of our components the background-color is currentColor. We will provide the mixin to override this. Also this IFF the $color argument is not provided, which was not an option before. Meaning this is backwards compatible.

moog16:
@abhiomkar confirmed that on the icon-button, if we change the wrapping element's color to a different color with `!important`, this does not cascade down to the element. Example from screenshot tests:
```
<div style='color: orange !important'>
  <button class="mdc-icon-button mdc-icon-button--on mdc-theme--primary" aria-label="Add to favorites" aria-pressed="true">
    :before
    <i class="material-icons mdc-icon-button__icon mdc-icon-button__icon--on">favorite</i>
    <i class="material-icons mdc-icon-button__icon">favorite_border</i>
  </button>
</div>
```
The :before element's background-color is still button's color via the `currentColor` prop.

moog16:
Closing - stale PR

Issue #4579, opened 2019-04-08
All packages /index should export ./constants


moog16:
Example in mdc-ripple/index.ts:

```js
import * as util from './util';

export {util};
export * from './adapter';
export * from './component';
export * from './foundation';
export * from './types';
```

We should also have a line to export constants, `export * from './constants'`. This should happen in all packages.

PR #4593, opened 2019-04-11
chore: export constants from index

moog16:
fixes https://github.com/material-components/material-components-web/issues/4579


packages/mdc-chips/chip-set/index.ts
```diff
@@ -23,4 +23,8 @@
 
 export * from './adapter';
 export * from './component';
+export {
```
mmalerba: Another idea would be for all of them do something like: `export * as chipSetConstants from './constants';`. That way you don't have to rename all the individual symbols.

packages/mdc-tabs/tab-bar/index.ts
```diff
@@ -23,5 +23,9 @@
 
 export * from './adapter';
 export * from './component';
+export {
+  strings as tabBartrings,
```
mmalerba: tabBarStrings

packages/mdc-chips/chip-set/index.ts
moog16: ah I like that idea...I"ll try that out :)

moog16: I don't think that syntax actually works.

mmalerba: Oh interesting looks like you're right. I assumed it would since that style syntax works for import. In that case I think you could accomplish the same thing like this:

```ts
import * as chipSetConstants from './constants';
export {chipSetConstants};
```

moog16: Ah - I did try that but I forgot the `*`...Not as clean as your original, but better than manually naming everything.

moog16:
With this change as is, you should already be able to do something like:
```
mdc.menuSurface.cssClasses;
```

Is that what you were proposing?

packages/mdc-checkbox/index.ts
```diff
@@ -23,4 +23,5 @@
 
 export * from './adapter';
 export * from './component';
+export * from './constants';
```
mmalerba: for consistency, doesn't it make sense to do the `import * as` thing for all of them?

moog16: ya we should be consistent. Good call!

moog16: @abhiomkar actually makes a good point that the constants should be prefixed with `MDCx` like `MDCSelectFoundation`, etc. For consistency I will update these to `MDCSelectConstants`. So the result will be:

```
mdc.select.MDCSelectConstants.cssClasses;
```

abhiomkar: Curious, does re-exporting named alias might affect tree-shaking? For example, `MDCSelectConstants` bundles everything including cssClass, strings, numbers, etc?

abhiomkar: If it does bundle everything, I think it makes sense to export them as-is as @mmalerba suggested for simplicity.

moog16: @abhiomkar what @mmalerba said in [this comment](https://github.com/material-components/material-components-web/pull/4593#discussion_r276853644), is to name space all of them. Is that what you're proposing to do?

mmalerba: I just tested this out, it does look like merging them all into a single object prevents unused constants from being tree-shaken. I guess its kind of a similar problem as with tacking it all on to the Foundation class. 

To a certain degree it's a problem even if we don't merge them (e.g. using one of the `cssClasses` means none of the others can be tree shaken). I guess the ideal from a tree-shaking perspective would be if every constant was a separate export

abhiomkar: I agree that ideally every constant needs to be separate export in tree-shaking perspective.

Also it's worth noting that since we were not re-exporting separate constants from index file we ended up exposing them via Foundation's static methods. We might deprecate these static methods from Foundation for clean up.

moog16: Hey picking this back up since I have less on my plate. If we were to export every constant individually, would we get rid of 'string', 'numbers', 'cssClasses'? And the entire constants file would be flat? I agree if we go through with this change, we should also remove the static members from foundation. This however will be a breaking change, so I think we'd have to keep them around with a deprecation notice.

moog16:
We've decided to go the route of exporting all constants as `cssClasses` or `strings`, etc. unless it comes from within a subcomponent (ie. chips, chipSet). These will be prepended with the subcomponent name: `chipCssClasses` or `chipSetCssClasses`.

packages/mdc-chips/chip-set/index.ts
```diff
@@ -21,6 +21,9 @@
  * THE SOFTWARE.
  */
 
+import {cssClasses as chipSetCssClasses, strings as chipSetStrings} from './constants';
+
 export * from './adapter';
 export * from './component';
 export * from './foundation';
+export {chipSetCssClasses, chipSetStrings};
```
abhiomkar: You can directly export without additional import statement.

```ts
export {cssClasses as chipSetCssClasses, strings as chipSetStrings} from './constants';
```

abhiomkar:
Also, Add `BREAKING CHANGE:` line to PR description.

moog16:
I dont think this is a breaking change. This is a new feature, so its not going to break anyone currently using our components.

packages/mdc-select/index.ts
```diff
@@ -21,9 +21,12 @@
...
 
+import * as selectConstants from './constants';
+
 export * from './adapter';
 export * from './component';
 export * from './foundation';
 export * from './types';
 export * from './helper-text/index';
 export * from './icon/index';
+export {selectConstants};
```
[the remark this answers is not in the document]
moog16: I believe this is an outdated commit

Issue #4585, opened 2019-04-09
MDCChecbkbox component.ts should not import index file


ALMaclaine:
In the MDCCheckbox component.ts file [here](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/component.ts#L26) and [here](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/component.ts#L27).

The import statement should not import the /index but rather the package name.

```js
// From
import {ponyfill} from '@material/dom/index';
import {MDCRipple, MDCRippleAdapter, MDCRippleCapableSurface, MDCRippleFoundation} from '@material/ripple/index';

// To
import {ponyfill} from '@material/dom';
import {MDCRipple, MDCRippleAdapter, MDCRippleCapableSurface, MDCRippleFoundation} from '@material/ripple';
```

As per [importing-js.md](https://github.com/material-components/material-components-web/blob/master/docs/importing-js.md#es-modules) and #4509

abhiomkar:
Thanks for reporting this issue and PR!

There are many places where we're referencing `/index` when importing modules. It is intentional to explicitly import from `/index` file where all pre-compiled modules (including foundation, component, etc) are exported. We might need to be even more explicit about where we want to import the modules from: For example, `mdc-foo/foundation`, `mdc-foo/constants` instead of importing everything from `mdc-foo/index` for consistency.

Can you please provide more details or context as to why do you think it should point to package name without `/index`?

ALMaclaine:
There's a pretty long discussion between @moog16 and @kfranqueiro in #4509 about whether/when `/index` is needed.

Also now that I look at it, I'm pretty sure the second line is wrong because foundations are supposed to be imported directly.

https://github.com/material-components/material-components-web/blob/master/docs/code/best_practices.md#foundation-import-statements-must-not-use-re-exported-modules.

Honestly I'm just trying to help fix inconsistencies in the code base where I can. However it's somewhat hard because at times what documentation says conflicts with what different contributors are saying. Two weeks ago I was ADDING the index in because that's what the importing-js readme said, but after discussion in #4509 it was stated that importing index is unnecessary because of the package.json module field.

moog16:
@ALMaclaine gotcha - we're going to comb through and replace all instances to follow out best practices. Thanks!

PR #4594, opened 2019-04-11
chore: normalize imports by removing @material/**/index.js for foundation, types, adapter, etc

moog16:
fixes https://github.com/material-components/material-components-web/issues/4585


ALMaclaine:
There's quite a few in material-components-web as well.

https://github.com/material-components/material-components-web/blob/master/packages/material-components-web/index.ts

abhiomkar:
@ALMaclaine `material-components-web/index.ts` exports everything from other component's index files. It is intentional to use /index.ts path in this case.

PR #4595, opened 2019-04-12
fix(text-field): Fix TextField padding with leading icon

### LINKED ISSUE #4294: Text field/Select padding should be 12px if icon is present

Padding between edge of text field and icon should be 12px for both leading and trailing icons

https://material-components.github.io/material-components-web-catalog/#/component/text-field
- Left padding with leading icon is 16px for both outlined and filled text fields
- Right padding with trailing icon is 16px for outlined text fields

sdomenet:
Refs #4294 

Padding is now constant across text field variants and icon position at 12px.
I have a question about a variable initialized 2 times with different values, please see comment.

### DIFF
--- a/packages/mdc-textfield/icon/_variables.scss
+++ b/packages/mdc-textfield/icon/_variables.scss
@@ -22,4 +22,5 @@
 
 $mdc-text-field-icon-position: 12px !default;
 $mdc-text-field-icon-padding: 48px !default;
+$mdc-text-field-dense-icon-padding: 44px !default;
 $mdc-text-field-dense-icon-padding: 38px !default;
--- a/packages/mdc-textfield/icon/_variables.scss
+++ b/packages/mdc-textfield/icon/_variables.scss
@@ -22,5 +22,4 @@
 
 $mdc-text-field-icon-position: 12px !default;
 $mdc-text-field-icon-padding: 48px !default;
-$mdc-text-field-dense-icon-padding: 44px !default;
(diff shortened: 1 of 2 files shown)


williamernest:
The first variable can be removed since it's ignored by the second assignment. 

I duplicated this PR in #4639 so that I could run the screenshot tests. 

This does fix the issue in question but we should hold off on merging it until the select is updated as well for consistency across like components.

packages/mdc-textfield/icon/_variables.scss
```diff
@@ -22,5 +22,4 @@
 
 $mdc-text-field-icon-position: 12px !default;
 $mdc-text-field-icon-padding: 48px !default;
-$mdc-text-field-dense-icon-padding: 44px !default;
```
abhiomkar: I think, they way `!default` works in sass it assigns a value only if it is not already initialized. So the currently value of `$mdc-text-field-dense-icon-padding` is `44px` and the next assignment with value `38px` is ignored.

We need to remove second assignment to keep current behaviour intact.

As @williamernest mentioned in his previous comment can you please add this fix to select component for consistency?

Thanks!

sdomenet: I reverted the changes to keep `dense-icon-padding` at 44px.

abhiomkar:
Designer approved these pixel changes. I'm merging this PR to `fix/leading_icon_padding` to make this change in select component as well for consistency.

PR #4599, opened 2019-04-15
chore: Updated golden screenshots after FF upgrade 62 => 65

abhiomkar:


abhiomkar:
It seems the previous build failure caused by the flakiness of FF browser. This change is not required anymore.

PR #4601, opened 2019-04-15
fix: export constants from index.ts

### LINKED ISSUE #4579: All packages /index should export ./constants

mmalerba:
@moog16 fixes #4579


mmalerba:
My bad, didn't see that

PR #4602, opened 2019-04-15
fix(checkbox): Fixed disabled checkbox styles for Edge browser

### LINKED ISSUE #4596: Checkbox disabled styles aren't applied until hover
Checkbox isn't getting the `:disabled` styles immediately on Edge. Checkbox only gets the disabled style after hover or some kind of interaction that forces layout.

Switching the Scss styles to `[disabled]` seems to fix this issue.

mmalerba:
@moog16 

Fixes #4596

### DIFF
--- a/packages/mdc-checkbox/_mixins.scss
+++ b/packages/mdc-checkbox/_mixins.scss
@@ -424,8 +424,11 @@
 }
 
 @mixin mdc-checkbox-unmarked-background-selector-disabled_ {
+  // Note: we must use `[disabled]` instead of `:disabled` below because Edge does not always recalculate the style
+  // property when the `:disabled` pseudo-class is followed by a sibling combinator. See:
+  // https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11295231/
   // stylelint-disable-next-line selector-max-specificity
-  .mdc-checkbox__native-control:disabled:not(:checked):not(:indeterminate) ~ .mdc-checkbox__background {
+  .mdc-checkbox__native-control[disabled]:not(:checked):not(:indeterminate) ~ .mdc-checkbox__background {
     @content;
   }
 }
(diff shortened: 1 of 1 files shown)


packages/mdc-checkbox/_mixins.scss
```diff
@@ -141,7 +141,7 @@
       @include mdc-checkbox__native-control_;
     }
 
-    &:disabled {
+    &[disabled] {
```
abhiomkar: If this is edge bug in styling checkbox can we add attribute selector only to mdc-checkbox?

`:disabled` CSS3 pseudo selector seems to be working for all other components in all major browsers.

WDYT?

mmalerba: I didn't test the other components, I just figured they could have similar bugs. In order to really test it, you have to have a way to change its disabled state without interacting with the element itself. (e.g. put a button on the page that on click programmatically changes the disabled state of the component). If the component just starts off disabled, or you disable it by clicking on it, the bug won't show.

If we're confident other components don't have the issue I can switch them back, but it seems safer to just change them

moog16: I'll test them out, to verify

abhiomkar: I just tested this on Edge browser programmatically toggling the button / checkbox and also rendering widgets with disabled state. I'm not able to reproduce the original issue?

checkbox_disabled_edge

https://marshy-tiger.glitch.me/

This MS Edge bug seems related:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11295231/

mmalerba: Good find, I've updated it to only change places where `:disabled` was being used together with the sibling combinator (which were all in checkbox).

packages/mdc-checkbox/_mixins.scss
```diff
@@ -424,8 +424,11 @@
 }
 
 @mixin mdc-checkbox-unmarked-background-selector-disabled_ {
+  // Note: we must use `[disabled]` instead of `:disabled` below because Edge does not recalculate
+  // the style properly when the `:disabled` pseudo-class is followed by a sibling combinator. See:
```
abhiomkar: typo: `properly` => `property`*

Same below.

PR #4608, opened 2019-04-16
fix(text-field): add classes constant

TroyTae:
There are some classes which is defined in README.md.
But there is no definition in constant.ts.
So I added it!

### DIFF
--- a/packages/mdc-textfield/constants.ts
+++ b/packages/mdc-textfield/constants.ts
@@ -34,7 +34,7 @@ const cssClasses = {
   DENSE: 'mdc-text-field--dense',
   DISABLED: 'mdc-text-field--disabled',
   FOCUSED: 'mdc-text-field--focused',
-  FULL_WIDTH: 'mdc-text-field--fullwidth',
+  FULLWIDTH: 'mdc-text-field--fullwidth',
   HELPER_LINE: 'mdc-text-field-helper-line',
   INVALID: 'mdc-text-field--invalid',
   NO_LABEL: 'mdc-text-field--no-label',
@@ -42,7 +42,7 @@ const cssClasses = {
   ROOT: 'mdc-text-field',
   TEXTAREA: 'mdc-text-field--textarea',
   WITH_LEADING_ICON: 'mdc-text-field--with-leading-icon',
-  WITH_TRAILING_ICON: 'mdc-text-field--with-leading-icon',
+  WITH_TRAILING_ICON: 'mdc-text-field--with-trailing-icon',
 };
 
 const numbers = {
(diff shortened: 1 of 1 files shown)


TroyTae:
related to https://github.com/material-components/material-components-web-react/pull/812

packages/mdc-textfield/constants.ts
```diff
@@ -34,12 +34,15 @@ const cssClasses = {
   DENSE: 'mdc-text-field--dense',
   DISABLED: 'mdc-text-field--disabled',
   FOCUSED: 'mdc-text-field--focused',
+  FULL_WIDTH: 'mdc-text-field--fullwidth',
   HELPER_LINE: 'mdc-text-field-helper-line',
   INVALID: 'mdc-text-field--invalid',
+  NO_LABLE: 'mdc-text-field--no-label',
```
moog16: LABEL

TroyTae: @moog16
I guess mdc constant follows below rule...

1. remove "mdc-text-field--"
2. apply snakecase
3. apply uppercase

Why LABLE is suitable in this case?

moog16: I don't think I understand what you're trying to say. I'm saying that LABLE is misspelled.

TroyTae: Oh I'm so sorry.
I understood you want to change fron NO_LABEL to LABEL.
I'll change it!

packages/mdc-textfield/constants.ts
```diff
@@ -34,12 +34,15 @@ const cssClasses = {
...
   FOCUSED: 'mdc-text-field--focused',
+  FULL_WIDTH: 'mdc-text-field--fullwidth',
   HELPER_LINE: 'mdc-text-field-helper-line',
   INVALID: 'mdc-text-field--invalid',
+  NO_LABEL: 'mdc-text-field--no-label',
   OUTLINED: 'mdc-text-field--outlined',
   ROOT: 'mdc-text-field',
   TEXTAREA: 'mdc-text-field--textarea',
   WITH_LEADING_ICON: 'mdc-text-field--with-leading-icon',
+  WITH_TRAILING_ICON: 'mdc-text-field--with-leading-icon',
```
moog16: This should be --with-trailing-icon

packages/mdc-textfield/constants.ts
```diff
@@ -34,12 +34,15 @@ const cssClasses = {
   DENSE: 'mdc-text-field--dense',
   DISABLED: 'mdc-text-field--disabled',
   FOCUSED: 'mdc-text-field--focused',
+  FULL_WIDTH: 'mdc-text-field--fullwidth',
```
moog16: if the class name is `--fullwidth`, lets change the key to `FULLWIDTH`

packages/mdc-textfield/constants.ts
TroyTae: Thank you!

PR #4610, opened 2019-04-16
fix(top-app-bar): Move comment line to appropriate section

### LINKED ISSUE #4609: MDCTopAppBar has private mixin in public section
One of the mixins in MDCTopAppBar [here](https://github.com/material-components/material-components-web/blob/master/packages/mdc-top-app-bar/_mixins.scss#L73), is private (I believe as indicated by the trailing underscore), but is in the public section.

Alternatively, if this member is supposed to be public, it has a trailing underscore and is not documented.

ALMaclaine:
Move MDCTopAppBar mixin that is private from public to private section

fixes #4609


packages/mdc-top-app-bar/_mixins.scss
```diff
@@ -103,32 +132,3 @@
     }
   }
 }
-
```
abhiomkar: It was intentional to keep all private mixins at the bottom of the page. Any reason for change of its location?

ALMaclaine:
It's in the public section.

https://github.com/material-components/material-components-web/blob/master/packages/mdc-top-app-bar/_mixins.scss#L73

I moved it to the private section.

ALMaclaine:
@abhiomkar There's a private mixin in the public section.

packages/mdc-top-app-bar/_mixins.scss
abhiomkar: In that case, doesn't it make sense to move `// private` comment lines above `mdc-top-app-bar-mobile-breakpoint_` mixin? This'll avoid touching git history.

ALMaclaine: Yea, that would be better, I'll adjust it.

ALMaclaine:
@abhiomkar Okay, I believe I adjusted it correctly.

PR #4614, opened 2019-04-17
feat(text-field): define icon's cssClasses

TroyTae:
There is one question!
"mdc-text-field__icon" or "mdc-text-field-icon", which one is correct?
Almost code is written "mdc-text-field__icon".

스크린샷 2019-04-17 오후 5 20 19
But definition of cssClasses is written "mdc-text-field-icon" in README.
And other module(helper-text/counter) follows this convention.


TroyTae:
related to https://github.com/material-components/material-components-web-react/pull/812

PR #4615, opened 2019-04-17
fix(tab-bar): fix the activation of tab without previous active tab

### LINKED ISSUE #4226: [TabBar] MDCTabBar: At least one tab must have mdc-tab--active class
There must be one tab element in the tab bar component with a class of mdc-tab--active or an error is thrown in the console: Cannot read property 'deactivate' of undefined when switching tabs

Latest from master branch.

All

See this [CodePen](https://codepen.io/robjacobs/pen/Kboapr?editors=1011) example.  Notice the error in the console that results from trying to use the tab bar activateTab method.  Notice the error is also thrown when clicking on tabs.  Uncommenting the last tab in the example fixes the issue as it has a mdc-tab--active class.

Using the mdc-tab--active class to initially set the active tab should not be required.  You should be able to use the mdc-tab activateTab method to set the initial active tab.

You must use the mdc-tab--active on at least one child mdc-tab element.

stof:
If there is no previous active tab, the index will be `-1`, which will be trigger errors in the adapter of the component when trying to call methods on `this.tabList_[index]`. This is now handled gracefully for the case when there is no previously active tab.

Fixes #4226

### DIFF
--- a/packages/mdc-tab-bar/adapter.ts
+++ b/packages/mdc-tab-bar/adapter.ts
@@ -74,7 +74,7 @@ export interface MDCTabBarAdapter {
    * @param index The index of the tab to activate
    * @param clientRect The client rect of the previously active Tab Indicator
    */
-  activateTabAtIndex(index: number, clientRect: ClientRect): void;
+  activateTabAtIndex(index: number, clientRect?: ClientRect): void;
 
   /**
    * Deactivates the tab at the given index
--- a/packages/mdc-tab-bar/foundation.ts
+++ b/packages/mdc-tab-bar/foundation.ts
@@ -97,8 +97,13 @@ export class MDCTabBarFoundation extends MDCFoundation<MDCTabBarAdapter> {
       return;
     }
 
-    this.adapter_.deactivateTabAtIndex(previousActiveIndex);
-    this.adapter_.activateTabAtIndex(index, this.adapter_.getTabIndicatorClientRectAtIndex(previousActiveIndex));
+    let previousClientRect;
+    if (previousActiveIndex !== -1) {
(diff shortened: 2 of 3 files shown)


packages/mdc-tab-bar/adapter.ts
```diff
@@ -74,7 +74,7 @@ export interface MDCTabBarAdapter {
    * @param index The index of the tab to activate
    * @param clientRect The client rect of the previously active Tab Indicator
    */
-  activateTabAtIndex(index: number, clientRect: ClientRect): void;
+  activateTabAtIndex(index: number, clientRect?: ClientRect): void;
```
stof: the implementation in the component adapter defers to a MDCTab method which already uses an optional parameter. That's why it does not involve other changes.

packages/mdc-tab-bar/foundation.ts
```diff
@@ -97,8 +97,13 @@ export class MDCTabBarFoundation extends MDCFoundation<MDCTabBarAdapter> {
       return;
     }
 
-    this.adapter_.deactivateTabAtIndex(previousActiveIndex);
-    this.adapter_.activateTabAtIndex(index, this.adapter_.getTabIndicatorClientRectAtIndex(previousActiveIndex));
+    let previousClientRect;
+    if (-1 !== previousActiveIndex) {
```
patrickrodee: Let's reverse this to `if (previousActiveIndex !== -1)`

packages/mdc-tab-bar/adapter.ts
patrickrodee: Nice!

stof:
@patrickrodee done. For the test, I added a new test covering that behavior instead of changing the one you linked, as it was covering a different thing.

Issue #4525, opened 2019-03-20
MDC Menu: Implement API to change selected item in selection group


abhiomkar:
Implement API to change selected item in selection group. 

For example,

```
// API Signature
menu.setSelectedIndex(index: number, selectionGroupIndex = 0)

// Usage
menu.setSelectedIndex(3)
```

![image](https://user-images.githubusercontent.com/47312/54722748-af6fbc80-4b3b-11e9-8a49-e6d4d040e49d.png)

Plausible ARIA attributes for selectable menu items

- `role="menuitemradio"` for each selectable menu item.
- `aria-checked="true"` for selected menu item.

moog16:
We should also be updating the aria tags with regards to the checkbox.

moog16:
Closing for #4620

PR #4620, opened 2019-04-17
feat(menu): add setSelectedIndex to set selected item in menu selection group

moog16:
fixes https://github.com/material-components/material-components-web/issues/4525

BREAKING CHANGE: Replaced adapter methods `getParentElement`, `getSelectedElementIndex` with `getSelectedSiblingOfItemAtIndex`, `isSelectableItemAtIndex`.


packages/mdc-menu/foundation.ts
```diff
@@ -131,19 +131,33 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
...
     // De-select the previous selection in this group.
     const selectedIndex = this.adapter_.getSelectedElementIndex(selectionGroup);
     if (selectedIndex >= 0) {
       this.adapter_.removeAttributeFromElementAtIndex(selectedIndex, strings.ARIA_SELECTED_ATTR);
       this.adapter_.removeClassFromElementAtIndex(selectedIndex, cssClasses.MENU_SELECTED_LIST_ITEM);
     }
     // Select the new list item in this group.
-    this.adapter_.addClassToElementAtIndex(index, cssClasses.MENU_SELECTED_LIST_ITEM);
-    this.adapter_.addAttributeToElementAtIndex(index, strings.ARIA_SELECTED_ATTR, 'true');
+    const normalizedIndex = this.adapter_.getListItemIndexOfSelectionGroup(index, selectionGroup);
```
abhiomkar: Alternatively the `setSelectedIndex` API can accept the root list indexing. This approach simplifies our implementation and also avoids confusion using two separate indexing (ie., indexing for main list and indexing for each selection group). It makes the component consumer easier to just use list indexing everywhere. 

Majority of the time that I've seen Menu has single selection group. `setSelectedIndex` will work even though there are multiple selection groups as expected.


API signature:

```ts
setSelectedIndex(index: number)
```

Example:

```ts
setSelectedIndex(5) // index '0' starts from first list item always.
```

WDYT?

packages/mdc-menu/foundation.ts
```diff
@@ -134,12 +133,15 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
   /**
    * Selects the list item at `index` witnin the specified selection group.
    * @param index Index of list item within the selection group.
-   * @param selectionGroupIndex Index of selection group within the menu.
    */
-  setSelectedIndex(index: number, selectionGroupIndex = 0) {
-    const selectionGroup = this.adapter_.getSelectionGroupAtIndex(selectionGroupIndex);
+  setSelectedIndex(index: number) {
+    const listItem = this.adapter_.getListItemByIndex(index);
```
abhiomkar: `getListItemAtIndex`* to be consistent with List adapter APIs

packages/mdc-menu/foundation.ts
```diff
@@ -134,12 +133,15 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
...
+    if (!listItem) {
+      throw new Error('No list item at specified index.');
```
abhiomkar: Add `MDCMenuFoundation: ` at the beginning of the error message. This'll help easily debug the error. Same below.

packages/mdc-menu/foundation.ts
```diff
@@ -134,12 +133,15 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
   /**
    * Selects the list item at `index` witnin the specified selection group.
    * @param index Index of list item within the selection group.
-   * @param selectionGroupIndex Index of selection group within the menu.
    */
-  setSelectedIndex(index: number, selectionGroupIndex = 0) {
-    const selectionGroup = this.adapter_.getSelectionGroupAtIndex(selectionGroupIndex);
+  setSelectedIndex(index: number) {
+    const listItem = this.adapter_.getListItemByIndex(index);
+    if (!listItem) {
```
abhiomkar: We should do validation on the `index` number instead similar to List's setSelectedIndex foundation API.

packages/mdc-menu/component.ts
```diff
@@ -40,7 +40,7 @@ export class MDCMenu extends MDCComponent<MDCMenuFoundation> {
   }
 
   private menuSurfaceFactory_!: MDCMenuSurfaceFactory; // assigned in initialize()
-  private listFactory_!: MDCListFactory; // assigned in initialize()
+  private listFactory_!: MDCListFactory; // ` assigned in initialize()
```
abhiomkar: revert.

packages/mdc-menu/component.ts
```diff
@@ -140,6 +140,10 @@ export class MDCMenu extends MDCComponent<MDCMenuFoundation> {
     this.menuSurface_.setAnchorMargin(margin);
   }
 
+  setSelectedIndex(index: number) {
```
abhiomkar: Add TS doc for param.

packages/mdc-menu/component.ts
```diff
@@ -205,6 +209,9 @@ export class MDCMenu extends MDCComponent<MDCMenuFoundation> {
         const selectedListItem = selectionGroup.querySelector(`.${cssClasses.MENU_SELECTED_LIST_ITEM}`);
         return selectedListItem ? this.items.indexOf(selectedListItem) : -1;
       },
+      getListItemByIndex: (index: number) => {
```
abhiomkar: `return` statement is not required if your remove curly brackets.

packages/mdc-menu/foundation.ts
```diff
@@ -99,9 +100,7 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
     // Wait for the menu to close before adding/removing classes that affect styles.
     this.closeAnimationEndTimerId_ = setTimeout(() => {
       const selectionGroup = this.getSelectionGroup_(listItem);
-      if (selectionGroup) {
-        this.handleSelectionGroup_(selectionGroup, index);
-      }
+      this.handleSelectionGroup_(selectionGroup, index);
```
abhiomkar: I feel this shouldn't even call `handleSelectionGroup_` if the current item is not in selection group.

test/screenshot/spec/mdc-menu/classes/multiple-menu-selection-group.html
```diff
@@ -0,0 +1,165 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2018 Google Inc.
```
abhiomkar: 2019*

packages/mdc-menu/README.md
```diff
@@ -265,6 +267,7 @@ Method Signature | Description
 `handleItemAction(listItem: Element) => void` | Event handler for list's action event.
 `handleMenuSurfaceOpened() => void` | Event handler for menu surface's opened event.
 `setDefaultFocusState(focusState: DefaultFocusState) => void` | Sets default focus state where the menu should focus every time when menu is opened. Focuses the list root (`DefaultFocusState.LIST_ROOT`) element by default.
+`setSelectedIndex(index: number) => void` | Selects the list item at `index` witnin the menu.
```
abhiomkar: "Selects the list item at given `index`."

packages/mdc-menu/foundation.ts
moog16: are you referring to 
```
  private isIndexValid_(index: MDCListIndex) {
    if (index instanceof Array) {
      if (!this.isCheckboxList_) {
        throw new Error('MDCListFoundation: Array of index is only supported for checkbox based list');
      }

      if (index.length === 0) {
        return true;
      } else {
        return index.some((i) => this.isIndexInRange_(i));
      }
    } else if (typeof index === 'number') {
      if (this.isCheckboxList_) {
        throw new Error('MDCListFoundation: Expected array of index for checkbox based list but got number: ' + index);
      }
      return this.isIndexInRange_(index);
    } else {
      return false;
    }
  }
```

If so I can make this a public static method, or transform it to a util.

abhiomkar: Yes, but we just need to validate if index is in range (similar to what `isIndexInRange_` method does). Since we can't share the adapters between two components I would just add separate implementation in Menu as of now with custom error message catered to `MDCMenuFoundation`.

packages/mdc-menu/adapter.ts
```diff
@@ -65,6 +65,11 @@ export interface MDCMenuAdapter {
    */
   getElementIndex(element: Element): number;
 
+  /**
+   * @return List item element.
+   * @param index Index of list item within menu.
+   */
+  getListItemAtIndex(index: number): Element | null;
```
abhiomkar: Foundation will always send valid index number to adapter that makes return type non-nullable.

packages/mdc-menu/component.ts
```diff
@@ -143,6 +143,14 @@ export class MDCMenu extends MDCComponent<MDCMenuFoundation> {
     this.menuSurface_.setAnchorMargin(margin);
   }
 
+  /**
+   * Sets the list item as the selected row at the specified index.
+   * @param index Index of list item within menu.
+   */
+  setSelectedIndex(index: number) {
```
abhiomkar: Do you think component should also support `getSelectedIndex` API? Seems like a common use case to retrieve current state.

`getSelectedIndex` will always return array of selected index to support multiple selection group. We can track this as separate issue.

packages/mdc-menu/foundation.ts
```diff
@@ -99,8 +100,8 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
     // Wait for the menu to close before adding/removing classes that affect styles.
     this.closeAnimationEndTimerId_ = setTimeout(() => {
       const selectionGroup = this.getSelectionGroup_(listItem);
-      if (selectionGroup) {
-        this.handleSelectionGroup_(selectionGroup, index);
+      if (!selectionGroup) {
```
abhiomkar: Revert this change?

packages/mdc-menu/foundation.ts
```diff
@@ -131,6 +132,22 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
     this.defaultFocusState_ = focusState;
   }
 
+  /**
+   * Selects the list item at `index` witnin the specified selection group.
```
abhiomkar: Update TS Doc based on new behaviour.

packages/mdc-menu/foundation.ts
```diff
@@ -131,6 +132,22 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
...
+   * @param index Index of list item within the selection group.
+   */
+  setSelectedIndex(index: number) {
+    const listItem = this.adapter_.getListItemAtIndex(index);
+    if (!listItem) {
+      throw new Error('MDCMenuFoundation:No list item at specified index.');
+    }
+    const selectionGroup = this.getSelectionGroup_(listItem);
+    if (!selectionGroup) {
+      throw new Error('MDCMenuFoundation:No selection group at specified index.');
```
abhiomkar: nit: space after colon.

Same above.

packages/mdc-menu/adapter.ts
moog16: What should i have for the defaultFoundation value then? I could also switch this to undefined.

packages/mdc-menu/component.ts
moog16: https://github.com/material-components/material-components-web/issues/4634

packages/mdc-menu/foundation.ts
moog16: are you saying we should support multi select for the menu? Otherwise the isIndexValid method should probably just look like:
```
  private isIndexInRange_(index: number) {
    const listSize = this.adapter_.getListItemCount();
    return index >= 0 && index < listSize;
  }

```

We'll also need to add this listItemCount adapter method.

abhiomkar: Multi-selection is out of scope for Menu. 

Yes, that's the validation I'm referring to, you can make use of existing adapter `getMenuItemCount` to get the count instead.

packages/mdc-menu/adapter.ts
abhiomkar: `null as Element`? This keeps the TSLint happy. We don't want to create a dummy element just to provide value to default adapter.

WDYT?

moog16: ya I don't think we should create a dummy element. However isn't `Element | null` technically correct, since at one point it is `null`?

packages/mdc-menu/foundation.ts
```diff
@@ -131,6 +132,27 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
...
 
+  /**
+   * Selects the list item at `index` within the menu.
+   * @param index Index of list item within the menu.
+   */
+  setSelectedIndex(index: number) {
+    if (!this.isIndexInRange_(index)) {
+      throw new Error('MDCMenuFoundation: No list item at specified index.');
+    }
+    const listItem = this.adapter_.getListItemAtIndex(index);
```
abhiomkar: I understand that this adapter API was introduced to be consistent with other adapter APIs in this foundation. IMO foundation should avoid accessing DOM directly and this needs to be moved over to adapter. For example, `getSelectionGroup_` logic should be moved to adapter.

I created a new branch ([feat/menu/selected-item-selection-group_cleanup_techdebt](https://github.com/material-components/material-components-web/compare/feat/menu/selected-item-selection-group_cleanup_techdebt?expand=1)) based off of your branch to clean up the tech debt that we had earlier with menu foundation. PTAL, Feel free to merge this into your branch and make required changes if it looks good.

Thanks!

moog16: It sounds like I don't need to merge your branch into mine. I can just work off of your branch, since you branched off of this one. Is that correct? I think your changes make sense. And yes we shouldn't be referencing the DOM in the foundation.

abhiomkar: You can use the same PR if you merge my branch into yours.

packages/mdc-dialog/package-lock.json
```diff
@@ -1,8 +1,6 @@
 {
-  "name": "@material/dialog",
-  "version": "1.1.0",
-  "lockfileVersion": 1,
   "requires": true,
+  "lockfileVersion": 1,
```
abhiomkar: Revert all package-lock.json changes

packages/mdc-menu/README.md
```diff
@@ -250,12 +251,12 @@ Method Signature | Description
 `elementContainsClass(element: Element, className: string) => boolean` | Returns true if the `element` contains the `className` class.
 `closeSurface() => void` | Closes the menu surface.
 `getElementIndex(element: Element) => number` | Returns the `index` value of the `element`.
-`getParentElement(element: Element) => Element \| null` | Returns the `.parentElement` element of the `element` provided.
-`getSelectedElementIndex(element: Element) => number` | Returns the `index` value of the element within the selection group provided, `element` that contains the `mdc-menu-item--selected` class.
 `notifySelected(index: number) => void` | Emits a `MDCMenu:selected` event for the element at the `index` specified.
 `getMenuItemCount() => number` | Returns the menu item count.
 `focusItemAtIndex(index: number)` | Focuses the menu item at given index.
 `focusListRoot() => void` | Focuses the list root element.
+`getSelectedSiblingOfItemAtIndex(index: number) => number` | Returns the closest selected sibling of item specified at `index`.
```
abhiomkar: Returns selected list item index within the same selection group which is a sibling of item at given `index`.

Also update this description in code TS doc.

packages/mdc-menu/foundation.ts
```diff
@@ -132,41 +130,36 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
...
+   * @param index Index of list item within the menu.
    */
-  private handleSelectionGroup_(selectionGroup: Element, index: number) {
-    // De-select the previous selection in this group.
-    const selectedIndex = this.adapter_.getSelectedElementIndex(selectionGroup);
-    if (selectedIndex >= 0) {
-      this.adapter_.removeAttributeFromElementAtIndex(selectedIndex, strings.ARIA_SELECTED_ATTR);
-      this.adapter_.removeClassFromElementAtIndex(selectedIndex, cssClasses.MENU_SELECTED_LIST_ITEM);
+  setSelectedIndex(index: number) {
+    if (!this.isIndexInRange_(index)) {
```
abhiomkar: Since this function is validating the index and throws error when index is invalid, let's call this `validateIndex_`. The return statement will never be executed with current structure.

```
this.isIndexValid_(index);
```

test/screenshot/spec/mdc-menu/classes/menu-selection-group.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!--
-  Copyright 2018 Google Inc.
+  Copyright 2019 Google Inc.
```
abhiomkar: Revert this change.

test/screenshot/spec/mdc-menu/classes/multiple-menu-selection-group.html
```diff
@@ -0,0 +1,165 @@
...
+  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  THE SOFTWARE.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Menu Selection Group - MDC Web Screenshot Test</title>
```
abhiomkar: Multiple Menu Selection Group

test/screenshot/spec/mdc-menu/fixture.js
```diff
@@ -26,6 +26,7 @@ import {DefaultFocusState} from '../../../../packages/mdc-menu/constants';
 window.mdc.testFixture.fontsLoaded.then(() => {
   const buttonEl = document.querySelector('.test-menu-button');
   const menuEl = document.querySelector('.mdc-menu');
+  const multipleSelectionGroupMenuEl = document.getElementById('multiple-selection-group-menu');
```
abhiomkar: Usually class names / ids related to tests starts with `test-`

test/unit/mdc-menu/mdc-menu.test.js
```diff
@@ -39,10 +39,37 @@ function getFixture(open) {
         <li tabIndex="-1" class="mdc-list-item" role="menuitem">Item</a>
         <li role="separator"></li>
         <li tabIndex="-1" class="mdc-list-item" role="menuitem">Another Item</a>
-        <ul class="mdc-menu__selection-group mdc-list" role="menu">
-          <li tabIndex="-1" class="mdc-list-item" role="menuitem">Item</a>
-          <li tabIndex="-1" class="mdc-list-item mdc-menu-item--selected" role="menuitem">Another Item</a>
-        </ul>
+        <li>
```
abhiomkar: Do you think we should create new fixture for single selection group as well?

test/unit/mdc-menu/mdc-menu.test.js
```diff
@@ -209,6 +236,12 @@ test('setAnchorMargin', () => {
   td.verify(menuSurface.setAnchorMargin({top: 0, right: 0, bottom: 0, left: 0}));
 });
 
+test('setSelectedIndex', () => {
```
abhiomkar: setSelectedIndex calls foundation method setSelectedIndex with given index.

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -158,9 +157,10 @@ test('handleItemAction item action event inside of a selection group with anothe
   const itemEl = document.createElement('li');
   td.when(mockAdapter.elementContainsClass(itemEl, listClasses.LIST_ITEM_CLASS)).thenReturn(true);
   td.when(mockAdapter.getElementIndex(itemEl)).thenReturn(0);
-  td.when(mockAdapter.getParentElement(itemEl)).thenReturn(itemEl);
   td.when(mockAdapter.elementContainsClass(itemEl, cssClasses.MENU_SELECTION_GROUP)).thenReturn(true);
-  td.when(mockAdapter.getSelectedElementIndex(itemEl)).thenReturn(0);
+  td.when(mockAdapter.isSelectableItemAtIndex(0)).thenReturn(true);
+  td.when(mockAdapter.getSelectedSiblingOfItemAtIndex(0)).thenReturn(0);
```
abhiomkar: `getSelectedSiblingOfItemAtIndex` will return sibling index that is different index?

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -266,96 +267,58 @@ test('handleMenuSurfaceOpened does not focus anything when DefaultFocusState is
...
-  td.when(mockAdapter.elementContainsClass(itemEl, listClasses.LIST_ITEM_CLASS)).thenReturn(true);
-  td.when(mockAdapter.getElementIndex(itemEl)).thenReturn(0);
-
-  foundation.handleItemAction(itemEl);
-
-  td.verify(mockAdapter.closeSurface(), {times: 1});
+  const listItemEl = document.createElement('div');
+  td.when(mockAdapter.isSelectableItemAtIndex(0)).thenReturn(true);
+  td.when(mockAdapter.elementContainsClass(listItemEl, cssClasses.MENU_SELECTION_GROUP)).thenReturn(true);
+  td.when(mockAdapter.getSelectedSiblingOfItemAtIndex(0)).thenReturn(false);
```
abhiomkar: This should return a valid sibling index.

test/unit/mdc-menu/menu.foundation.test.js
```diff
@@ -266,96 +267,58 @@ test('handleMenuSurfaceOpened does not focus anything when DefaultFocusState is
...
-  foundation.handleItemAction(itemEl);
-  clock.tick(numbers.TRANSITION_CLOSE_DURATION);
+test('setSelectedIndex throws error if index is not in range', () => {
+  const {foundation} = setupTest();
 
-  td.verify(mockAdapter.removeClassFromElementAtIndex(td.matchers.isA(Number), cssClasses.MENU_SELECTED_LIST_ITEM),
-    {times: 0});
-  td.verify(mockAdapter.addClassToElementAtIndex(0, cssClasses.MENU_SELECTED_LIST_ITEM), {times: 1});
+  try {
+    foundation.setSelectedIndex(5);
```
abhiomkar: You'll need to use assertThrow here.

packages/mdc-menu/foundation.ts
moog16: sorry i'm confused. Do you want to call it `isIndexValid_` or `validateIndex_`?

abhiomkar: I'm suggesting, `this.validateIndex_()`.


As per current implementation `isIndexValid_()` does not always return boolean.

test/unit/mdc-menu/mdc-menu.test.js
moog16: That is what the `getFixture` method already has. Or did you want to isolate the selection group?

abhiomkar: Ah ok. Yes we can keep this change if selection group is already in the fixture.

packages/mdc-menu/foundation.ts
moog16: Ah sorry I was confused...current implementation is `isIndexInRange_`. I'll change to validateIndex and return boolean.

abhiomkar: Sorry for not being clearer. `isIndexInRange_` seems perfectly fine if the function is just returning a boolean without any side-effects (i.e., throwing an error).

My earlier comment was that this if condition will never fail, because the method throws JS error instead of returning `false` value. (I noticed this in [coverage report](https://codecov.io/gh/material-components/material-components-web/src/d2ea93dcfa07e43620615f960242687fc66514d6/packages/mdc-menu/foundation.ts#L94)).

moog16: OH - I understood what you meant after I implemented this. I think I changed it correctly. I just need to update tests.

moog16:
This Pr requires https://github.com/material-components/material-components-web/pull/4679 to be merged first

PR #4632, opened 2019-04-22
Update _mixins.scss button fixes #4631

### LINKED ISSUE #4631: [Button] Add "fill: color" to disabled button
## Feature Requests

When using SVG icons in Buttons the CSS attribute `color` doesn't change the actual color of the SVG icon. Therefore an icon is not greyed out correctly when used with a disabled button. See demo below:

image

(there is a white icon on the `Next` button that should be greyed out)

I suggest that additionally to the `color` attribute that there is the `fill` attribute that would add the same color.

So instead of this:
image

There would be this CSS:
image

This would produce my expected output for the disabled button icon:
image

BhavyaSingh2611:
fix #4631 of white coloured icon

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -279,7 +279,7 @@
   &:disabled {
     @include mdc-feature-targets($feat-color) {
       @include mdc-theme-prop(background-color, transparent);
-
+      fill: $mdc-button-disabled-ink-color;
       color: $mdc-button-disabled-ink-color;
     }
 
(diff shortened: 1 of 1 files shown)


maracuja-juice:
Build is failing because of this:
image

packages/mdc-button/_mixins.scss
```diff
@@ -279,7 +279,7 @@
   &:disabled {
     @include mdc-feature-targets($feat-color) {
       @include mdc-theme-prop(background-color, transparent);
-
+      fill: $mdc-button-disabled-ink-color;
```
abhiomkar: `fill` is only applicable for icon block elements, so it needs to be moved to `mdc-button__icon_` mixin.

Also, this can use `fill: currentColor` instead.

abhiomkar:

@BhavyaSingh2611 Please feel free to re-open! :)

Thanks!

PR #4635, opened 2019-04-23
fix(text-field): fix overlapping icon and text field (#4617)

### LINKED ISSUE #4617: Text field input overlaps icon when there is no label

> v1.1.1

> Chrome Version 73.0.3683.86

### What OS are you using?

> Windows 10

> Add a text field with a leading (or trailing) icon and the no-label class
> https://codepen.io/anon/pen/NmyXYp

> The input should not overlap the icon.

> The input overlaps the icon

rbare1:
Set only top and bottom padding to vertically center without affecting
side padding that is set when an icon is present.

fixes #4617


abhiomkar:
Closing this PR for #4637

Issue #2263, opened 2018-02-16
Dialog / Select: overflow: hidden on body doesn't prevent scroll on iOS


kfranqueiro:
(Split from #2246)

### What browser(s) is this bug affecting?

TODO: need to nail down exact Safari or WebView versions

### What OS are you using?

iOS

### What are the steps to reproduce the bug?

Load the [select demo page](http://material-components-web.appspot.com/select.html) and open one of the selects, then tap+drag to scroll

### What is the expected behavior?

The body shouldn't scroll behind the open select

### What is the actual behavior?

The body does scroll

### Any other information you believe would be useful?

When the select menu is open, we set a class on the body that sets overflow: hidden. We do something similar for dialog, so I presume the same thing would happen in that demo as well (if there's enough on the page to scroll).

mohammadrafigh:
the same applies to dialogs and drawers in Android Webview.

mohammadrafigh:
It seems this issue was related to chromium engine and WebView on Android and I see it's now fixed. I cannot regenerate this anymore on any app. without any change.

PR #4636, opened 2019-04-23
fix: prevent overscroll for dialog and modal drawer

mohammadrafigh:
Fix dialog and modal drawer overscroll issue mentioned in #2263 for Webviews on Android and iOS


mohammadrafigh:
oops I missed something I'm going to rebase and push again

packages/mdc-drawer/component.ts
```diff
@@ -146,6 +146,8 @@ export class MDCDrawer extends MDCComponent<MDCDismissibleDrawerFoundation> {
       notifyOpen: () => this.emit(strings.OPEN_EVENT, {}, true /* shouldBubble */),
       trapFocus: () => this.focusTrap_!.activate(),
       releaseFocus: () => this.focusTrap_!.deactivate(),
+      addBodyClass: (className) => document.body.classList.add(className),
```
abhiomkar: We may have use cases where the inner elements of body are scrollable instead of the body itself.

We may be able to solve this problem by preventDefault'ing `touchmove` event on scrim element.

WDYT?

packages/mdc-drawer/adapter.ts
```diff
@@ -85,4 +85,14 @@ export interface MDCDrawerAdapter {
    * and restores focus to where it was prior to calling `trapFocus`.
    */
   releaseFocus(): void;
+
```
abhiomkar: Since this is a breaking change please move target branch (base branch) to `develop`.

mohammadrafigh: yes sure

packages/mdc-drawer/component.ts
mohammadrafigh: I doubt that preventDefault on touchmove prevents overscrolling behavior since it's a css thing that browser relies on it. but maybe I'm wrong. so please give me tomorrow to test it too.

mohammadrafigh: sorry for delay I was busy with our project. I tried it. putting touchmove on scrim doesn't work but it works for mdc-dialog itself because the scrim is a sibling to main dialog content but mdc-dialog is the parent so the overscroll could be prevented from mdc-dialog root. if you prefer this way I'm going to apply changes. we should do this in many places I think, in select menu etc.

mohammadrafigh: @abhiomkar I applied the changes in a new commit please review and let me know if I should rebase and squash these 2 commits.

packages/mdc-dialog/README.md
```diff
@@ -400,17 +406,21 @@ Method Signature | Description
 `setScrimClickAction(action: string)` | Sets the action reflected when the scrim is clicked. Setting to `''` disables closing the dialog via scrim click.
 `getAutoStackButtons() => boolean` | Returns whether stacked/unstacked action button layout is automatically handled during layout logic.
 `setAutoStackButtons(autoStack: boolean) => void` | Sets whether stacked/unstacked action button layout is automatically handled during layout logic.
-`handleInteraction(event: Event)` | Handles `click` and `keydown` events on or within the dialog's root element
-`handleDocumentKeydown(event: Event)` | Handles `keydown` events on or within the document while the dialog is open
+`handleClick(event: MouseEvent)` | Handles `click` events on or within the dialog's root element.
+`handleKeydown(event: KeyboardEvent)` | Handles `keydown` events on or within the dialog's root element.
+`handleDocumentKeydown(event: Event)` | Handles `keydown` events on or within the document while the dialog is open.c
```
abhiomkar: nit: Remove `c`

packages/mdc-dialog/foundation.ts
```diff
@@ -176,33 +177,53 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
...
+
+  /** Handles keydown on the document. */
   handleDocumentKeydown(evt: KeyboardEvent) {
     const isEscape = evt.key === 'Escape' || evt.keyCode === 27;
     if (isEscape && this.escapeKeyAction_ !== '') {
       this.close(this.escapeKeyAction_);
     }
   }
 
+  handleOverScroll(evt: TouchEvent) {
```
abhiomkar: It should prevent scrolling only when dialog is open.

packages/mdc-dialog/component.ts
```diff
@@ -121,15 +123,17 @@ export class MDCDialog extends MDCComponent<MDCDialogFoundation> {
       document.removeEventListener('keydown', this.handleDocumentKeydown_);
     };
 
-    this.listen('click', this.handleInteraction_);
-    this.listen('keydown', this.handleInteraction_);
+    this.listen('click', this.handleClick_);
+    this.listen('keydown', this.handleKeydown_);
+    this.listen('touchmove', this.handleOverScroll_);
```
abhiomkar: We need to attach this event only to `scrim` element. Then this method can be named as `handleScrimTouchmove`.

packages/mdc-checkbox/README.md
```diff
@@ -122,6 +122,7 @@ Mixin | Description
 `mdc-checkbox-container-colors($unmarked-stroke-color, $unmarked-fill-color, $marked-stroke-color, $marked-fill-color, $generate-keyframes)` | Generates CSS classes to set and animate the stroke color and/or container fill color of a checkbox
 `mdc-checkbox-ink-color($color)` | Sets the ink color of the checked and indeterminate icons
 `mdc-checkbox-focus-indicator-color($color)` | Sets the color of the focus indicator
+`mdc-checkbox-touch-dimension($touch-dimension)` | Sets the touch dimension of the checkbox.
```
abhiomkar: Rebase your branch with `develop`

packages/mdc-dialog/component.ts
mohammadrafigh: I tried that it doesn't work if we add this only to scrim since it's not a parent of dialog. so overscrolling dialog will scroll the elements on back.

packages/mdc-dialog/README.md
mohammadrafigh: oops! thanks

packages/mdc-checkbox/README.md
mohammadrafigh: sure

mohammadrafigh:
It seems this issue was related to chromium engine and WebView on Android and I see it's now fixed there. I cannot regenerate this anymore on any app. so I close this.

PR #4646, opened 2019-04-25
fix(text-field): Fix TextField padding with label and fullwidth

Krivega:
Before
Выделение_033
After
Выделение_034

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -487,8 +487,10 @@
 
     display: block;
 
-    .mdc-text-field__input {
-      padding: 0;
+    &.mdc-text-field--no-label {
+      .mdc-text-field__input {
+        padding: 0;
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/mdc-textfield/_mixins.scss
```diff
@@ -486,15 +486,15 @@
...
-    .mdc-text-field__input {
-      padding: 0;
-    }
   }
 
   &.mdc-text-field--textarea .mdc-text-field__input {
     resize: vertical;
   }
+
+  &.mdc-text-field--no-label:not(.mdc-text-field--textarea) .mdc-text-field__input {
```
abhiomkar: Full width text fields does not already have label. Can you please provide more context about this change?

As per your screenshot, seems like you are rendering floating label inside full width text field? I'm not sure if we support full width text field with label.

abhiomkar:

@Krivega we can always reopen it though! :)

Thanks!

PR #4647, opened 2019-04-25
fix(text-field): Padding between Textfield and Icon

### LINKED ISSUE #4296: Text field spacing between icon and label/input should be 16px

Space between icons and text should be 16px for both leading or trailing icons.

https://material-components.github.io/material-components-web-catalog/#/component/text-field
- Space between leading icon and text is currently 8px, for filled and outlined text fields
- space between trailing icon and text is 12px for filled text fields and 8px for outlined

Note that the horizontal (both left and right) padding of the whole text field component should be
- 16px if no icon is present
- 12px if icon is present, see #4294 #4148

sdomenet:
Together with my PR #4595, this should fix #4296 
`$mdc-text-field-icon-padding` was set to `48px`.
But when an icon is present, the input padding should be the sum of:
- the icon padding: `12px`,
- the icon width: `24px`,
- the input padding: `16px`
Therefore `52px`.

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -398,7 +398,10 @@
     $mdc-text-field-icon-padding,
     $mdc-text-field-input-padding
   );
-  @include mdc-notched-outline-floating-label-float-position($mdc-text-field-outlined-label-position-y, 32px);
+  @include mdc-notched-outline-floating-label-float-position(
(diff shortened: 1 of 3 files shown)


abhiomkar:
The input padding value seems correct. But it's breaking floating label float position on text field focus.

image

sdomenet:
I'm on it.

PR #4654, opened 2019-04-26
fix(tab): Explicitly set margin to 0 on tabs for Safari

### LINKED ISSUE #4653: MDCTabs: Safari shows space between individual tabs
We were able to override this by explicitly setting the margin to 0 on the mdc-tab class.

Screen Shot 2019-04-26 at 12 57 12 PM

williamernest:
fixes #4653

### DIFF
--- a/packages/mdc-tab/mdc-tab.scss
+++ b/packages/mdc-tab/mdc-tab.scss
@@ -42,6 +42,7 @@
   box-sizing: border-box;
   height: $mdc-tab-height;
   padding: 0 24px;
+  margin: 0;
   border: none;
   outline: none;
   background: none;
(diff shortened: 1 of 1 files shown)


packages/mdc-tab/mdc-tab.scss
```diff
@@ -42,6 +42,7 @@
   box-sizing: border-box;
   height: $mdc-tab-height;
   padding: 0 24px;
+  margin: 0;
```
abhiomkar: Add a comment here for safari.

williamernest: :+1:

PR #4663, opened 2019-05-01
fix(text-field): Update character counter to update when value is set.

williamernest:


packages/mdc-textfield/foundation.ts
```diff
@@ -289,8 +289,8 @@ export class MDCTextFieldFoundation extends MDCFoundation<MDCTextFieldAdapter> {
     // Prevent Safari from moving the caret to the end of the input when the value has not changed.
     if (this.getValue() !== value) {
       this.getNativeInput_().value = value;
-      this.setCharacterCounter_(value.length);
     }
+    this.setCharacterCounter_(value.length);
```
abhiomkar: What's the context of this change? any reason for setting this when value is same as native input value?

abhiomkar: Please add a test case.

williamernest: For frameworks that update the value property on the input, there is no way to trigger the character counter to resynchronize with the value. Since the character counter doesn't have any public apis in the component, this is a way to trigger that update. 

Also, this line doesn't apply to the Safari comment so it seems like there wasn't an explicit reason to include this line in the value check.

williamernest:
Reverted

PR #4664, opened 2019-05-01
docs(typography): Update setting font-family to use unquote

williamernest:
Updates the example of setting the global font-family to use unquote.

### DIFF
--- a/packages/mdc-typography/README.md
+++ b/packages/mdc-typography/README.md
@@ -139,7 +139,7 @@ $mdc-typography-styles-button: (
 
 Example: Overriding the global `font-family` property. 
 ```scss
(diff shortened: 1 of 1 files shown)


packages/mdc-typography/README.md
```diff
@@ -139,7 +139,7 @@ $mdc-typography-styles-button: (
 
 Example: Overriding the global `font-family` property. 
 ```scss
-$mdc-typography-font-family: "Arial, Helvetica, sans-serif";
+$mdc-typography-font-family: unquote("Arial, Helvetica, sans-serif");
```
moog16: This is probably not the only place we have this in sass...I'll check and possibly create an issue.

moog16: I'm not finding any other usages like this. No issue.

PR #4665, opened 2019-05-01
fix(typography): Use unquote for setting font-family in variables. 

williamernest:
While this did technically work, it semantically doesn't match the way we recommend users set the font-family.


packages/mdc-typography/_variables.scss
```diff
@@ -22,7 +22,7 @@
 
 @import "./functions";
 
-$mdc-typography-font-family: Roboto, sans-serif !default;
```
abhiomkar: Does our override mixins break if we set without quotes?

williamernest: Which override mixins? The `mdc-typography` mixins ?

abhiomkar: Yes, mixins / functions that use this variable.

williamernest: This variable is used below in the font styles map, but I would assume that our screenshot tests would break if the mixins/functions would break with this change.

abhiomkar: `unquote("Roboto, sans-serif")` seems to be same as `Roboto, sans-serif`. I looked at our typography mixins / functions, we don't seem to validate this type anywhere.

```sass
$fonts: unquote("Arial, sans-serif");
@debug type-of($fonts); // Return string

$fonts: Arial, sans-serif;
@debug type-of($fonts); // Return list

font-family: $fonts; // font-family can accept both list & string type.
```

Having said that this change looks good to me. But sass automatically type casts the list to string when applied to CSS property as mentioned above.

WDYT?

moog16:
@williamernest did you want to merge this?

packages/mdc-typography/_variables.scss
williamernest: I think either way works for me, as long as we're consistent in both what we recommend users use (via docs) and what we use.

williamernest:
Yes.

PR #4666, opened 2019-05-01
fix(tab-indicator): Show border for high contrast

patrickrodee:


packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -30,6 +30,7 @@
 @mixin mdc-tab-indicator-underline-color($color) {
   > .mdc-tab-indicator__content--underline {
     @include mdc-theme-prop(background-color, $color);
```
moog16: I think you could remove this background-color as well. I think it's a good opportunity to add a screenshot test to confirm that we didn't change anything. I can help you with that. :)

patrickrodee: Moved to using just border to make it consistent between high-contrast mode and normal.

patrickrodee:
I'll open a separate PR for screenshot tests.

packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -29,14 +29,13 @@
 
 @mixin mdc-tab-indicator-underline-color($color) {
   > .mdc-tab-indicator__content--underline {
-    @include mdc-theme-prop(background-color, $color);
```
moog16: I thought this was still required? (per convo from yesterday)

patrickrodee: I updated the height mixin to set the border-top-width instead of the actual height prop.

https://github.com/material-components/material-components-web/blob/1f1bf2c8b2afb051c431e200c2451e4d7362f3a9/packages/mdc-tab-indicator/_mixins.scss#L36-L40

moog16: Ok that makes sense. This works for me.

patrickrodee:
Dependent on #4673

PR #4667, opened 2019-05-02
docs: Integrated with TypeDoc for auto generating component docs.

abhiomkar:
## TypeDoc

Integrated with TypeDoc to auto generate documentation for Adapter / Foundation / Component APIs. Component README will provide reference to generated API doc in `Design & API Documentation` section as shown below.

_Example of Text Field's `Design & API Documentation` section linking API document page_

~Going forward all component API TSDoc serve as single source of truth for all API documentation. This'll avoid duplicating API documentation in [README file](https://github.com/material-components/material-components-web/blob/v2.0.0/packages/mdc-textfield/README.md) that we currently maintain in markdown table format.~

### Usage

Run this command after releasing a version of MDC Web to generate and publish to `gh-pages` branch.

```bash
npm run publish:docs
```

### Demo

- Homepage for [MDC Web component docs](https://material-components.github.io/material-components-web)
- [MDCTextField component doc](https://material-components.github.io/material-components-web/classes/_mdc_textfield_component_.mdctextfield.html)

### TODO

- Auto generate docs for sass APIs too using [sassdoc](http://sassdoc.com/).

**EDIT**

No plans to edit existing README files. This PR is a step forward to link API docs in current component README file.


williamernest:
The only issue I see with moving to this style of producing docs is the material.io team uses the readme file to produce the documentation for material.io (https://material.io/develop/web/components/input-controls/text-field/)

abhiomkar:
I think that is OK for not including API documentation on the component page of mio site. I feel, including everything in main README of component page is also overwhelming to consumers. But we'll be linking it on the page so consumer can still access latest API documentation.

We've few challenges with component documentation that auto-generating documentation solves:

- Component author have to manually create them to support markdown format duplicating the method / param description. This comes with maintenance cost and we easily forget to update in our docs when we make changes in component. We had few components with discrepancies in documentation.
- Having documentation in HTML format makes it easy to navigate between modules / classes by clicking on the target token. (For example, clicking on `MDCMenuSurfaceCorner` from foundation takes you to the respective interface definition).
- It makes it easier to link source code on GitHub from component documentation.

williamernest:
I don't think it makes sense for one platform to not have documentation on the material.io site when every other platform is represented. In addition to that, there's work in progress that this would drastically affect. I believe it's a good idea for building and maintaining documentation, but I don't foresee the readme going away and this will be in addition to.

abhiomkar:
Temporarily closing this.

PR #4668, opened 2019-05-02
feat(button): cssClasses constants

TroyTae:
related to https://github.com/material-components/material-components-web-react/issues/827


williamernest:
The button component doesn't require JS, so it doesn't require a constants file. While other frameworks may need these class names in a constant, I would argue that if it's not being used in MDC-Web it shouldn't be included in MDC-Web

TroyTae:
@williamernest 
Ah I see, then I declare it in the framework I want to use!

PR #4675, opened 2019-05-03
feat(tab): Improved mixins

patrickrodee:


packages/mdc-tab/mdc-tab.scss
```diff
@@ -129,13 +127,12 @@
 }
 
 .mdc-tab--active {
-  @include mdc-tab-text-label-color(primary);
```
moog16: This breaks CSS Vars for label and icon color. We might need to revert this change.

patrickrodee: There's a PR to fix this I think: https://github.com/material-components/material-components-web/pull/4688

PR #4676, opened 2019-05-03
fix(tab-indicator): Remove child selector

patrickrodee:


moog16:
@patrickrodee did you do this change on a non-google account? If it was - we can override the cla.

patrickrodee:
Nope, I've done it all on my corp machine. Let's override.

PR #4677, opened 2019-05-03
fix(infrastructure): revert stylelint change

moog16:


moog16:
These changes are unnecessary. The screenshots are a diff from Chrome 74.

PR #4685, opened 2019-05-06
feat(drawer): update icon-button cssClasses constant

TroyTae:
related to https://github.com/material-components/material-components-web-react/issues/827

Maybe we need to notify breaking change that cssClasses.ICON_BUTTON_ON changes to cssClasses.ON.

### DIFF
--- a/packages/mdc-icon-button/constants.ts
+++ b/packages/mdc-icon-button/constants.ts
@@ -22,7 +22,9 @@
  */
 
 export const cssClasses = {
-  ICON_BUTTON_ON: 'mdc-icon-button--on',
+  ICON: 'mdc-icon-button__icon',
+  ICON_ON: 'mdc-icon-button__icon--on',
+  ON: 'mdc-icon-button--on',
   ROOT: 'mdc-icon-button',
 };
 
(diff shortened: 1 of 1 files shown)


packages/mdc-icon-button/constants.ts
```diff
@@ -22,7 +22,10 @@
  */
 
 export const cssClasses = {
-  ICON_BUTTON_ON: 'mdc-icon-button--on',
+  ICON: 'mdc-icon-button__icon',
+  ICON_BUTTON_ON: 'mdc-icon-button--on', // deprecated
```
abhiomkar: I like it marked it as a deprecated for backward compat. We can remove this later in major release.

TroyTae: I tried to change code like this for notifying to developer...
```
/**
   * @deprecated
   * Use cssClasses.ON instead
   */
  get ICON_BUTTON_ON() {
    console.warn('MDCWeb: Use cssClasses.ON instead of cssClasses.ICON_BUTTON_ON');
    return 'mdc-icon-button--on';
  },
```
But console is blocked by tslint 😢

PR #4688, opened 2019-05-06
fix(tab): Fix tab color variables to use color literals

### LINKED ISSUE #4686: Regression: CSS runtime variable theming for tabs is broken
I think I've tracked it down to this line which no longer respects css variables, only the SASS.

https://github.com/material-components/material-components-web/pull/4675/files#diff-bebfa5920069adc8cf8c7d36390da5f6R130

abhiomkar:
This should fix #4686 to some extent.

- Used theme literal (`primary`) when defining override variables, where as, `mdc-theme-prop-value` is a function that returns a color value of `primary`.
- Set `!default` to all sass variables defined in `_variable.scss` for full customization.


packages/mdc-tab/_variables.scss
```diff
@@ -28,5 +28,5 @@ $mdc-tab-text-label-opacity: .6 !default;
 $mdc-tab-icon-opacity: .54 !default;
 $mdc-tab-text-label-color-default: rgba(mdc-theme-prop-value(on-surface), $mdc-tab-text-label-opacity) !default;
```
patrickrodee: Should these `mdc-theme-prop-value` calls be updated, too?

abhiomkar: Since we're mixing the color with alpha here we've to set color value to this variable.

PR #4689, opened 2019-05-07
chore: Removed unused scripts based on new release process

abhiomkar:


scripts/post-release.sh
```diff
@@ -28,27 +28,12 @@ function log() {
   echo '\033[36m[post-release]\033[0m' "$@"
 }
 
-if [[ $(git diff --cached CHANGELOG.md) ]]; then
```
moog16: when/how are we supposed to generate the changelog instead?

abhiomkar: Deleted this script & cherry pick script that are unused based on new release process documented internally.

We would need to run `npm run changelog` log command to generate changelog content as per new release process documentation.

moog16: Do you also want to delete this repo's release process doc in this PR then?

Also I don't see running the post-release script in the new doc.

abhiomkar: Updated this PR to delete cherry pick script and release process doc.

Issue #4495, opened 2019-03-11
Auto-init: Enable it to run twice and track what has been initialized


williamernest:
If the user updates the DOM and adds more components, the auto-init function should be able to be run again only initializing the components that were added.

abhiomkar:
This might be resolved by setting `data-` attribute (Eg, data-mdc-auto-init-state="registered") to DOM elements after a component is initialized. So that, next time when we run `.mdcAutoInit()` it'll only process DOM elements that do not have this `data-` attribute.

On related note, MDC also needs to be able to provide a way to retrieve component instance via DOM element. But we can track it as separate issue.

PR #4691, opened 2019-05-08
feat(auto-init): initialize components once with multiple mdc.autoInit() calls

moog16:
fixes https://github.com/material-components/material-components-web/issues/4495


packages/mdc-auto-init/index.ts
```diff
@@ -91,6 +91,7 @@ export function mdcAutoInit(root = document, warn = CONSOLE_WARN) {
       writable: false,
     });
     components.push(component);
+    node.setAttribute(AUTO_INIT_STATE_ATTR, INTIALIZED_STATE);
```
abhiomkar: I feel, we can take this further by moving it to MDC base component (`MDCComponent`). Auto init function then look for elements that do not have this attribute.

So this'll also avoid problem of initialing a component twice.

WDYT?

moog16: The only component that will be using this is the auto-init package, correct? What will other components do with the `AUTO_INIT_STATE_ATTR`?

abhiomkar: Right, we need to name this attribute more generic.

`data-mdc-init-state="initialized"`?

abhiomkar: MDCComponent can throw an error when component is initialized twice on same DOM element. Where as, autoInit picks up the elements that do not have this initialized data attribute and initializes them.

moog16: I see - that's a good idea

moog16: Well actually - MDC Component should early exit if its already been initialized. Right? I don't think we need to throw an error.

abhiomkar: IMO, the component should not be forgiving when it is initialized more than once. Throwing error explicitly makes developer discover the bug easier.

abhiomkar: Also, currently there is no way to get component instance from DOM element reference. I think It would be great addition if we can also access it via MDCComponent API (example, `MDCComponent.getInstance(menuEl)`).

Suggestion: MDCComponent can store component instance globally (example, `window.__mdcComponentInstanceXXXX` where `XXXX` is uniq id), and store this string to data attribute of DOM element where the component is initialized.

```html
data-mdc-component-id="__mdcComponentInstanceXXXX"
```

This can potentially replace `init-status` data attribute.

WDYT?

moog16: What about a warning instead of an error? Throwing an error might cause the user's app to crash or become in a bad state. 

RE storing DOM Node on component: I think is a feature separate from this. I think we could add whatever you want on top of what we're currently building in this PR.

abhiomkar: Idea is to not allow multiple instance of same component on same DOM element.

```ts
new MDCFoo(fooEl); // Always expected to return an instance of `MDCFoo`
```

Silently ignoring it or throwing a warning does not look semantic to me.

moog16: Gotcha - ok that makes sense. I'll throw a warning. 

So you're ok with storing DOM Node in separate PR?

moog16: I just tried setting the attribute in the constructor of mdc-base/component like so:

```
  constructor(
      root: Element,
      foundation?: FoundationType,
      ...args: Array<unknown>
  ) {
    this.root_ = root;
    debugger
    if (this.root_.getAttribute(INIT_STATE_ATTR) === INTIALIZED_STATE) {
      throw new Error('MDCComponent: component is already initialized.');
    }
    this.initialize(...args);
    this.root_.setAttribute(INIT_STATE_ATTR, INTIALIZED_STATE);
```

This however causes an issue. In checkbox (and others), we instantiate both the ripple and checkbox on the same .mdc-checkbox element. So when the checkbox initializes the ripple, it throws the error. I think we could get around this using a boolean when instantiating the ripple within the checkbox (and others). I wanted to run this by you before moving forward.

moog16: created https://github.com/material-components/material-components-web/issues/4715

moog16:
ready for rereview

packages/mdc-auto-init/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+export const strings = {
+  AUTO_INIT_ATTR: 'data-mdc-auto-init',
+  AUTO_INIT_STATE_ATTR: 'data-mdc-auto-init-state',
+  INTIALIZED_STATE: 'initialized',
```
abhiomkar: typo* `INITIALIZED_STATE`

moog16:
The `:not()` selector within the mdcAutoInit method is not supported by IE11. 

`const nodes: Element[] = [].slice.call(root.querySelectorAll(
    `[${AUTO_INIT_ATTR}]:not([${AUTO_INIT_STATE_ATTR}="${INITIALIZED_STATE}"])`));`

Looking for other solutions.

PR #4695, opened 2019-05-09
fix(ripple): Fixes issue where Chrome v74 shows black artifact on ripple surface on hover

### LINKED ISSUE #4629: Black hover over list item with checkbox

1.1.1

Chrome (Chromium) 74.0.3729.61

### What OS are you using?

Ubuntu 18.04

1. Go to [CodePen](https://codepen.io/anon/pen/qwKvbW) with Chrome 74+
2. Hover over items
3. You can see the black hover style

The background should be transparent.

The background is black.

Screenshot from 2019-04-22 00-01-13

This issue only happens on Chrome 74+. I'm not sure if it's a bug in Chrome 74+ or not since it's still in beta, but removing the z-index from this selector fixes the problem:

```
:not(.mdc-list--non-interactive)>:not(.mdc-list-item--disabled).mdc-list-item::before {
    transition: opacity 15ms linear,background-color 15ms linear;
    z-index: 1;
}
```

removing "will-change" in this selector fixes this issue too:

```
:not(.mdc-list--non-interactive)>:not(.mdc-list-item--disabled).mdc-list-item {
    cursor: pointer;
    --mdc-ripple-fg-size: 0;
    --mdc-ripple-left: 0;
    --mdc-ripple-top: 0;
    --mdc-ripple-fg-scale: 1;
    --mdc-ripple-fg-translate-end: 0;
    --mdc-ripple-fg-translate-start: 0;
    -webkit-tap-highlight-color: transparent;
    will-change: transform,opacity;
```

abhiomkar:
Removed will-change CSS property from Ripple.

Didn't notice any drastic performance changes with & without will-change CSS property.

* [Prototype: with will-change](https://mdc-list-will-change-test-qf5w6fen7.now.sh/checkbox.html) - _Current Behaviour_  
* [Prototype: without will-change](https://mdc-list-will-change-test-jcjev60xb.now.sh/checkbox.html)

As per documentation from MDN & other resources:

> Don't apply will-change to too many elements. The browser already tries as hard as it can to optimize everything. Some of the stronger optimizations that are likely to be tied to will-change end up using a lot of a machine’s resources, and when overused like this can cause the page to slow down or consume a lot of resources. ([MDN will-change](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change))

> Use sparingly...

> Don't apply will-change to elements to perform premature optimization...

> Warning: Do not promote elements unnecessarily. ([Ref](https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count))

IMO, Using will-change CSS property anticipating performance degradation is premature optimization.

We also need to consider removing will-change in other places until and unless it is really required.

EDIT: May 10th
This change resulted in so many screenshot test failures because of changes in text rendering in all the browsers. The text looks more sharp after this change since browser takes it out from composite layer. Golden screenshots are updated accordingly.

Fixes #4629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+<a name="2.1.0"></a>
+# [2.1.0](https://github.com/material-components/material-components-web/compare/v2.0.0...v2.1.0) (2019-05-06)
+
+
+### Bug Fixes
+
+* **tab:** Update horizontal padding mixin ([#4678](https://github.com/material-components/material-components-web/issues/4678)) ([d3ce9c9](https://github.com/material-components/material-components-web/commit/d3ce9c9))
+* **tab-indicator:** Remove child selector ([#4676](https://github.com/material-components/material-components-web/issues/4676)) ([edbe0ba](https://github.com/material-components/material-components-web/commit/edbe0ba))
+* **tab-indicator:** Show border for high contrast ([#4666](https://github.com/material-components/material-components-web/issues/4666)) ([5a52847](https://github.com/material-components/material-components-web/commit/5a52847))
+* **text-field:** Fixes overlapping input with leading icon in absence of label ([#4637](https://github.com/material-components/material-components-web/issues/4637)) ([64e459e](https://github.com/material-components/material-components-web/commit/64e459e))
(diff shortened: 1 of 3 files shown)


moog16:
Ya I also took a look at the First meaningful paint between your two prototypes. No noticible difference. Although one thing to consider is it is only 80 checkboxes.

PR #4701, opened 2019-05-10
refactor(top-app-bar): Remove [de]registerEventHandler methods from adapters

moog16:
related https://github.com/material-components/material-components-web/issues/4527 & https://github.com/material-components/material-components-web/issues/2813

BREAKING CHANGE: remove adapter methods registerNavigationIconInteractionHandler, deregisterNavigationIconInteractionHandler, registerScrollHandler, deregisterScrollHandler, registerResizeHandler, deregisterResizeHandler.


packages/mdc-top-app-bar/component.ts
```diff
@@ -57,19 +61,49 @@ export class MDCTopAppBar extends MDCComponent<MDCTopAppBarBaseFoundation> {
     this.scrollTarget_ = window;
   }
 
+  initialSyncWithDOM() {
+    this.handleNavigationIconClick_ = this.foundation_.handleNavigationClick.bind(this.foundation_);
```
abhiomkar: `this.handleNavigationIconClick_`* to be same as foundation's callbak method name.

packages/mdc-top-app-bar/component.ts
```diff
@@ -57,19 +61,49 @@ export class MDCTopAppBar extends MDCComponent<MDCTopAppBarBaseFoundation> {
...
+    this.handleResize_ = this.foundation_.handleResize && this.foundation_.handleResize.bind(this.foundation_);
```
abhiomkar: This needs to be even more explicite by detecting the type of TopAppBar before assigning callbacks.

Currently it is checking for existence of methods in foundation (for example, `handleResize`), instead it can detect presense of class name similar to how we detect the type of top app bar before instantiating appropriate TopAppBar foundation in `getDefaultFoundation` method. This makes easier for us to discover which TopAppBar variant handles respective events.

packages/mdc-top-app-bar/README.md
```diff
@@ -234,8 +228,9 @@ All foundations provide the following methods:
 
 Method Signature | Description
 --- | ---
-`initScrollHandler(handler: EventListener) => void` | Registers a scroll handler on a specific target element.
-`destroyScrollHandler(handler: EventListener) => void` | Deregisters the current scroll handler set by the foundation.
+`handleScroll() => void` | Handles `scroll` event on specified scrollTarget (default to `window`).
+`handleResize() => void` | Handles `resize` event on window.
```
abhiomkar: `handleWindowResize` since the target is not the root.

packages/mdc-top-app-bar/README.md
```diff
@@ -234,8 +228,9 @@ All foundations provide the following methods:
 
 Method Signature | Description
 --- | ---
-`initScrollHandler(handler: EventListener) => void` | Registers a scroll handler on a specific target element.
-`destroyScrollHandler(handler: EventListener) => void` | Deregisters the current scroll handler set by the foundation.
+`handleScroll() => void` | Handles `scroll` event on specified scrollTarget (default to `window`).
```
abhiomkar: `handleTargetScroll` since this is handling scroll on outside the component.

packages/mdc-top-app-bar/short/foundation.ts
```diff
@@ -46,16 +46,13 @@ export class MDCShortTopAppBarFoundation extends MDCTopAppBarBaseFoundation {
     }
 
     if (!this.adapter_.hasClass(cssClasses.SHORT_COLLAPSED_CLASS)) {
-      this.scrollHandler_ = () => this.shortAppBarScrollHandler_();
-      this.adapter_.registerScrollHandler(this.scrollHandler_);
+      this.handleScroll = () => this.shortAppBarScrollHandler_();
```
abhiomkar: This logic needs to be moved to `handleScroll` callback itself and exit early when short collapsed class is found.

```ts
handleScroll() {
  if (this.adapter_.hasClass(cssClasses.SHORT_COLLAPSED_CLASS)) {
    return;
  }
  // ...
}
```

this.handleScroll = () => this.shortAppBarScrollHandler_

packages/mdc-top-app-bar/foundation.ts
```diff
@@ -50,64 +50,23 @@ export class MDCTopAppBarBaseFoundation extends MDCFoundation<MDCTopAppBarAdapte
...
       getViewportScrollY: () => 0,
       getTotalActionItems: () => 0,
     };
     // tslint:enable:object-literal-sort-keys
   }
 
-  protected scrollHandler_?: SpecificEventListener<'scroll'>;
-  protected resizeHandler_?: SpecificEventListener<'resize'>;
-  private readonly navClickHandler_: SpecificEventListener<'click'>;
+  handleScroll?: SpecificEventListener<'scroll'>;
```
abhiomkar: Ideally these methods should be overriden in respective child classes of TopAppBar foundation.

In current model each child foundation of TopAppBar is overwriting this property by reassigning new function, but each foundation should override these functions instead.

packages/mdc-top-app-bar/component.ts
moog16: Having the detection of class names in the component is business logic. I thought we were trying to avoid more business logic in the component layer. The foundations set/unset the handleScroll and handleResize value (function or undefined). There is a bit a magic that is happening, but I believe the foundation should work this way.

abhiomkar: This business logic is very specific to component that determines appropriate foundation to initialize. See `mdc-drawer/component.ts` for example.

We can avoid having this logic in component by creating separate Class for each component variant, but we didn't want to go that route yet.

moog16: I believe it will look like this:
```
  initialSyncWithDOM() {
    const isFixed = this.root_.classList.contains(cssClasses.FIXED_CLASS);
    const isShort = this.root_.classList.contains(cssClasses.SHORT_CLASS);

    this.handleNavigationClick_ = this.foundation_.handleNavigationClick.bind(this.foundation_);
    this.handleWindowResize_
      = this.foundation_.handleWindowResize && this.foundation_.handleWindowResize.bind(this.foundation_);
    this.handleTargetScroll_
      = this.foundation_.handleTargetScroll && this.foundation_.handleTargetScroll.bind(this.foundation_);

    this.scrollTarget_.addEventListener('scroll', this.handleTargetScroll_ as EventListener);

    if (this.navIcon_) {
      this.navIcon_.addEventListener('click', this.handleNavigationClick_ as EventListener);
    }

    if (!isShort && !isFixed) {
      window.addEventListener('resize', this.handleWindowResize_ as EventListener);
    }
  }

```

Does that look ok to you?

abhiomkar: We might not need this condition if we provide abstract method in base Class of TopAppBar.

```ts
this.foundation_.handleWindowResize && ...
```

It'll always return a Function type.

moog16: You mean the base class should effectively be a noop?

So we can just do:
```
  initialSyncWithDOM() {
          window.addEventListener('resize', this.handleWindowResize_ as EventListener);
          this.navIcon_.addEventListener('click', this.handleNavigationClick_ as EventListener);
          this.scrollTarget_.addEventListener('scroll', this.handleTargetScroll_ as EventListener);
  }
```

abhiomkar: Yes.

moog16: That seems inefficient. We could potentially have 3 event listeners doing nothing.

abhiomkar: Not if we debounce window `resize` & `scroll` events. But the way foundations implemented is that it assumes component attaches event listeners to specific TopAppBar variants. So, we might've to bind these events to only specific variants.

This logic could potentially be moved to Foundation but it is out of scope of this PR.

moog16: `So, we might've to bind these events to only specific variants.` --> You mean as it is right now? If not, what does this look like? 

Even if we do debounce, we're still creating noop eventListeners, which can be avoided by keeping this as is.

packages/mdc-top-app-bar/foundation.ts
```diff
@@ -50,64 +50,23 @@ export class MDCTopAppBarBaseFoundation extends MDCFoundation<MDCTopAppBarAdapte
...
       getViewportScrollY: () => 0,
       getTotalActionItems: () => 0,
     };
     // tslint:enable:object-literal-sort-keys
   }
 
-  protected scrollHandler_?: SpecificEventListener<'scroll'>;
-  protected resizeHandler_?: SpecificEventListener<'resize'>;
-  private readonly navClickHandler_: SpecificEventListener<'click'>;
+  handleTargetScroll?: SpecificEventListener<'scroll'>;
```
abhiomkar: Again couldn't comment on previous thread.

`handleTargetScroll` needs to be a abstract method which can be overriden in child classes.

abhiomkar: Please see `mdc-drawer/dismissible/foundation.ts` for example where `opened()` & `closed()` are abstract methods in Dismissible variant where Modal foundation overrides it.

packages/mdc-top-app-bar/short/foundation.ts
```diff
@@ -46,20 +46,17 @@ export class MDCShortTopAppBarFoundation extends MDCTopAppBarBaseFoundation {
     }
 
     if (!this.adapter_.hasClass(cssClasses.SHORT_COLLAPSED_CLASS)) {
```
abhiomkar: Remove this check since it is already attached to target scroll event in component.

packages/mdc-top-app-bar/component.ts
```diff
@@ -57,19 +61,55 @@ export class MDCTopAppBar extends MDCComponent<MDCTopAppBarBaseFoundation> {
...
+    const isFixed = this.root_.classList.contains(cssClasses.FIXED_CLASS);
+    const isShort = this.root_.classList.contains(cssClasses.SHORT_CLASS);
+
+    this.handleNavigationClick_ = this.foundation_.handleNavigationClick.bind(this.foundation_);
+    this.handleWindowResize_
+      = this.foundation_.handleWindowResize && this.foundation_.handleWindowResize.bind(this.foundation_);
+    this.handleTargetScroll_
+      = this.foundation_.handleTargetScroll && this.foundation_.handleTargetScroll.bind(this.foundation_);
+
+    this.scrollTarget_.addEventListener('scroll', this.handleTargetScroll_ as EventListener);
```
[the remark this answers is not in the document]
moog16: You mean this check? `this.foundation_.handleTargetScroll && this.foundation_.handleTargetScroll.bind(this.foundation_);` I will remove this check in the next commit

packages/mdc-top-app-bar/foundation.ts
moog16: I do not see any abstract methods in the dismissible/foundation. I see `protected`, but the top-app-bar/component needs to call the `handleX` methods.

Also if they are abstract, that means they technically don't always exist, which means I'll need the check I am removing RE [this comment](https://github.com/material-components/material-components-web/pull/4701#discussion_r284412039), correct?

packages/mdc-top-app-bar/fixed/foundation.ts
```diff
@@ -31,17 +30,10 @@ export class MDCFixedTopAppBarFoundation extends MDCTopAppBarFoundation {
...
-    super(adapter);
-
-    this.scrollHandler_ = () => this.fixedScrollHandler_();
-  }
-
   /**
    * Scroll handler for applying/removing the modifier class on the fixed top app bar.
    */
-  private fixedScrollHandler_() {
+  handleTargetScroll = () => {
```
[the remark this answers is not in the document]
moog16: We have a lint rule `only-arrow-functions`. Without the arrow we I can't retain the type `SpecificEventListener<x>`. Any suggestions?

packages/mdc-top-app-bar/foundation.ts
abhiomkar: abstract method is just an empty function (which is a truthy value), in typescript we use `abstract` method to enforce child classes to implement.

packages/mdc-top-app-bar/fixed/foundation.ts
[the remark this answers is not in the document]
abhiomkar: I do not see any lint errors with this being a named function?

packages/mdc-top-app-bar/foundation.ts
moog16: I'm getting error `Abstract methods can only appear within an abstract class.`. I think I just make them empty methods.

moog16: spoke offline...will not have abstract methods, but instead will just have noops

packages/mdc-top-app-bar/short/foundation.ts
```diff
@@ -45,21 +45,16 @@ export class MDCShortTopAppBarFoundation extends MDCTopAppBarBaseFoundation {
...
-    if (!this.adapter_.hasClass(cssClasses.SHORT_COLLAPSED_CLASS)) {
-      this.scrollHandler_ = () => this.shortAppBarScrollHandler_();
-      this.adapter_.registerScrollHandler(this.scrollHandler_);
-      this.shortAppBarScrollHandler_();
-    }
-  }
-
-  destroy() {
-    super.destroy();
+    this.handleTargetScroll();
```
abhiomkar: Add test for `isCollapsed()`

packages/mdc-top-app-bar/component.ts
```diff
@@ -82,21 +119,7 @@ export class MDCTopAppBar extends MDCComponent<MDCTopAppBarBaseFoundation> {
...
-        if (this.navIcon_) {
-          this.navIcon_.addEventListener(evtType, handler);
-        }
-      },
-      deregisterNavigationIconInteractionHandler: (evtType, handler) => {
-        if (this.navIcon_) {
-          this.navIcon_.removeEventListener(evtType, handler);
-        }
-      },
       notifyNavigationIconClicked: () => this.emit(strings.NAVIGATION_EVENT, {}),
```
abhiomkar: Add test for `notifyNavigationIconClicked()`

packages/mdc-top-app-bar/short/foundation.ts
abhiomkar: foundation calling the handle method by itself looks weird to me. Add a comment if this is expected?

moog16: there is already a test here: 
```test('short top app bar: #adapter.removeClass called when page is scrolled to the top', () => {
```

packages/mdc-top-app-bar/short/foundation.ts
```diff
@@ -45,21 +45,19 @@ export class MDCShortTopAppBarFoundation extends MDCTopAppBarBaseFoundation {
...
-    if (!this.adapter_.hasClass(cssClasses.SHORT_COLLAPSED_CLASS)) {
-      this.scrollHandler_ = () => this.shortAppBarScrollHandler_();
-      this.adapter_.registerScrollHandler(this.scrollHandler_);
-      this.shortAppBarScrollHandler_();
-    }
-  }
-
-  destroy() {
-    super.destroy();
+    // this was a private method call before, but we decided to override the base class method instead.
```
abhiomkar: First line of this code comment is not required. I think we don't have to refer to the history.

packages/mdc-top-app-bar/component.ts
```diff
@@ -57,19 +61,52 @@ export class MDCTopAppBar extends MDCComponent<MDCTopAppBarBaseFoundation> {
...
+      window.removeEventListener('resize', this.handleWindowResize_ as EventListener);
+    }
     super.destroy();
   }
 
   setScrollTarget(target: EventTarget) {
     // Remove scroll handler from the previous scroll target
-    this.foundation_.destroyScrollHandler();
+    if (this.handleTargetScroll_) {
+      this.scrollTarget_.removeEventListener('scroll', this.handleTargetScroll_ as EventListener);
```
abhiomkar: `this.handleTargetScroll_` is non-nullable so we don't need this check.

Same below.

packages/mdc-top-app-bar/fixed/foundation.ts
```diff
@@ -31,17 +30,10 @@ export class MDCFixedTopAppBarFoundation extends MDCTopAppBarFoundation {
...
-    super(adapter);
-
-    this.scrollHandler_ = () => this.fixedScrollHandler_();
-  }
-
   /**
    * Scroll handler for applying/removing the modifier class on the fixed top app bar.
    */
-  private fixedScrollHandler_() {
+  handleTargetScroll() {
```
abhiomkar: Add `@override` annotation all `handleTargetScroll` and `handleWindowScroll` methods in subclasses.

packages/mdc-top-app-bar/foundation.ts
```diff
@@ -50,64 +49,23 @@ export class MDCTopAppBarBaseFoundation extends MDCFoundation<MDCTopAppBarAdapte
...
-      this.adapter_.deregisterScrollHandler(this.scrollHandler_);
-    }
-  }
-
-  private initResizeHandler_() {
-    if (this.resizeHandler_) {
-      this.adapter_.registerResizeHandler(this.resizeHandler_);
-    }
-  }
+  handleTargetScroll() {} // tslint:disable-line:no-empty
```
abhiomkar: `/** Other variants of TopAppBar foundation overrides this method */`

Same below.

packages/mdc-top-app-bar/fixed/foundation.ts
```diff
@@ -21,7 +21,6 @@
  * THE SOFTWARE.
  */
 
-import {MDCTopAppBarAdapter} from '../adapter';
```
abhiomkar: Seems like the coverage score went down, can we aim to have 100% test coverage?

image

packages/mdc-top-app-bar/fixed/foundation.ts
moog16: I don't see @override anywhere else. Is that a JSDoc thing?

Issue #1892, opened 2018-01-08
Add prefix and suffix for text fields


ribrdb:
## Feature Request

Please add support for prefixes and suffixes as shown here:
https://material.io/guidelines/components/text-fields.html#text-fields-input-types

This seems like it's probably pretty similar to the leading and trailing icons you already support.

mikehans9:
jesus......almost one year and the most basic feature for textfields still not done? a simple prefix/suffix? crazy.

jamesmfriedman:
This just came up for RMWC. Thinking about adding it in myself, but it isn't actually easy to support. This is a relatively complex issue due to the fact that the size of the prefix or suffix is unknown so it can't just be added like the icons are with padding. The solution then would be something like making the container display flex, but having the input only be a small piece of it, and having the other elements outside of it, then deferring focus to the input if anything else is clicked.

ChainsDD:
I've created a mostly working solution. I still need to support leading and trailing icons, and convert it to a material component. Once that's done I'll post it here to get some more feedback. 

Basically I'm positioning a span on top of the input, and adjusting the padding of the input to account for it. It's working pretty well, I'll try and throw something up on codepen to demonstrate it.

patrickrodee:
@asyncliz

PR #4718, opened 2019-05-15
feat(text-field): Prefix/suffix support

ChainsDD:
Addresses #1892.

### DIFF
--- a/demos/index.html
+++ b/demos/index.html
@@ -180,6 +180,14 @@
             </span>
           </a>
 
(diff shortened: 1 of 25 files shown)


ChainsDD:
I signed it!

benebsiny:
So... do textfield now support prefix/suffix ?

asyncLiz:
Thanks for the PR! We're working on our implementation of prefix and suffix, which should be coming really soon. Subscribe to #1892 for updates!

PR #4726, opened 2019-05-16
fix(select): Fixes arrow direction on select focused state

### LINKED ISSUE #4618: [Select] After choosing an element in mdc-select the arrow still points to the top

1.1.1

All

### What OS are you using?

Ubuntu

> 1. Go to https://codepen.io/anon/pen/mgxvKR
> 2. Choose an item
> 3. You can see the arrow is still pointing to the top

the arrow should be depended to the visibility of the menu (options) and not the focus status of the select element.

The arrow is pointing to the down.
Screenshot from 2019-04-17 21-28-10

abhiomkar:
#### Before
image

#### After

image

Introduced `handleMenuOpened` & `handleMenuClosed` foundation methods to listen to menu lifecycle events.

Fixes #4618


packages/mdc-select/README.md
```diff
@@ -372,6 +372,7 @@ well as interaction targets. See [here](icon/) for more information on using ico
 | `mdc-select__menu` | Mandatory when using the enhanced select. This class should be placed on the `mdc-menu` element within the `mdc-select` element. |
 | `mdc-select__dropdown-icon` | Mandatory. Should be placed on an `i` element within the `mdc-select` element. Used for the dropdown arrow svg and animation.
 | `mdc-select__icon` | Optional. Should be placed on an `i` or `svg` element within the `mdc-select` element. Used for the leading icon.
+| `mdc-select--activated` | Optional. Styles the activated state of select. This class will be added automatically when menu is opened.
```
moog16: I think the correct term here according to the [spec](https://material.io/design/interaction/states.html#selected) is --selected instead of --activated.

packages/mdc-select/component.ts
```diff
@@ -156,6 +156,8 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
     this.handleKeydown_ = (evt) => this.foundation_.handleKeydown(evt);
     this.handleMenuSelected_ = (evtData) => this.selectedIndex = evtData.detail.index;
     this.handleMenuOpened_ = () => {
+      this.foundation_.handleMenuOpened();
```
moog16: This logic should be within the menu foundation. This seems like business logic that the select should not be aware of. Same with below.

packages/mdc-select/foundation.ts
```diff
@@ -132,6 +132,14 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
     this.notchOutline(openNotch);
   }
 
+  handleMenuOpened() {
+    this.adapter_.addClass(cssClasses.ACTIVATED);
```
moog16: I think the adapter methods should move to the menu, and should instead be adding `mdc-menu--selected`.

packages/mdc-select/README.md
abhiomkar: I do not see the example of select component using --selected term?

I named this class name after the spec from [Menus](https://material.io/design/components/menus.html):
image

Where activated indicates that the select dropdown is opened.

packages/mdc-select/foundation.ts
abhiomkar: This class name is added when select / menu dropdown is opened which is out of the scope of Menu component.

packages/mdc-select/component.ts
abhiomkar: Did you mean the focus logic in `handleMenuOpened_` should be moved to select foundation? I feel it is out of the scope of this PR. There are many places in this file we do include business logic in component which should've been delegated to foundation.

packages/mdc-select/README.md
moog16: hmm yeah you're right. The definition of activated vs selected is confusing to me, but whatever spec says...

packages/mdc-select/component.ts
moog16: Ignore comment as it is out of scope for the PR

packages/mdc-select/README.md
```diff
@@ -462,6 +463,8 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
 | `handleFocus() => void` | Handles a focus event on the `select` element. |
 | `handleBlur() => void` | Handles a blur event on the `select` element. |
 | `handleClick(normalizedX: number) => void` | Sets the line ripple center to the normalizedX for the line ripple. |
+| `handleMenuOpened() => void` | Handles menu or menu surface opened event.
```
moog16: Don't think you need to mention menu surface:
```
Handles menu opened event.
```

same with below.

abhiomkar: I think this depends on framework where the parent component encapsulates the child component's events. I want give option here that this callback can be bound to menu or menu surface event. 

In vanilla implementation we listen to Menu Surface opened event.

PR #4735, opened 2019-05-17
fix(checkbox): screenshot test golden update

moog16:
Fixing checkbox screenshot test. SVG has changed on Firefox.


abhiomkar:
This test was flaky for me earlier. Do you want to remove autofocus?

abhiomkar:
Or let’s move autofocus to unchecked checkbox.

moog16:
@abhiomkar how will moving autofocus fix the issue?

abhiomkar:
@moog16 Seems like the ripple z position is flaky with the check mark icon. We can avoid this flaky test by moving the focus to unchecked checkbox.

PR #4736, opened 2019-05-17
feat(chips): Add setAttr adapter method

patrickrodee:
Add `MDCChipAdapter#setAttr` method. Update screenshot tests with appropriate roles.

BREAKING CHANGE: Add the `setAttr` method to the chip adapter.

### DIFF
--- a/test/screenshot/spec/mdc-chips/classes/choice.html
+++ b/test/screenshot/spec/mdc-chips/classes/choice.html
@@ -44,16 +44,16 @@
       <div class="test-layout">
         <div class="test-cell test-cell--chip">
           <div class="mdc-chip-set mdc-chip-set--choice">
-            <button class="mdc-chip">
+            <button class="mdc-chip" role="radio" aria-checked="false">
               <span class="mdc-chip__text">Chip One</span>
             </button>
-            <button class="mdc-chip">
+            <button class="mdc-chip" role="radio" aria-checked="false">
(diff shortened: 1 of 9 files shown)


test/screenshot/spec/mdc-chips/classes/choice.html
```diff
@@ -44,33 +44,33 @@
...
+            <button class="mdc-chip" role="radio" aria-checked="false">
               <span class="mdc-chip__text">Chip Four</span>
             </button>
           </div>
         </div>
 
         <div class="test-cell test-cell--chip">
           <div class="mdc-chip-set mdc-chip-set--choice">
-            <button class="mdc-chip mdc-chip--selected">
+            <button class="mdc-chip mdc-chip--selected" role="radio" aria-checked="true">
```
abhiomkar: If this is a radio type selection update this demo to have single selection.

patrickrodee: Choice chips are single selection by default. I don't think there's any change to be done for that.

abhiomkar: Right. All radio buttons have `aria-checked="true"` which is not intended for radio group. Am I missing something?

patrickrodee: Maybe I'm missing something. Can you make a commit or open a PR with the change you want? I'm not following.

Issue #4460, opened 2019-02-27
[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event


gertcuykens:
chrome example: https://output.jsbin.com/fotedog

https://jsbin.com/fotedog/edit?html,output

probably related to https://github.com/material-components/material-components-web/issues/629

Not sure how to fix it

![image](https://user-images.githubusercontent.com/863731/53515281-e406d000-3ac9-11e9-955d-1f7572d0800c.png)

![image](https://user-images.githubusercontent.com/863731/53520192-5df08680-3ad5-11e9-865e-8cf7d67c0f65.png)

acdvorak:
Thanks for bringing this to our attention!

We'll need to update the component's default adapter implementation to pass `{passive: true}` as the third parameter to `addEventListener()` in [supported browsers](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener):

https://github.com/material-components/material-components-web/blob/ad5743ae665d138c38a23ba36a3d2f63b5c388d5/packages/mdc-textfield/component.ts#L381-L390

It looks like a number of components listen to `touchstart` events: `ripple`, `select`, `slider`, `textfield`, and `tab-bar-scroller`. So we might want to extract the `applyPassive()` function from ripple util and move it to `mdc-dom`:

https://github.com/material-components/material-components-web/blob/2025c8b5eb9891ec3f159e2fde03f45d23f6bbed/packages/mdc-ripple/util.ts#L91-L109

moog16:
I'm no longer seeing this issue in our screenshot tests, unit tests, or in the jsbin provided. Is this something we should still fix?

gertcuykens:
When you refresh the page it still happens, why you need to refresh I don't know.

![image](https://user-images.githubusercontent.com/863731/57415935-5b5d8c00-71fe-11e9-9030-6035a8088eeb.png)

abhiomkar:
I'm able to reproduce this, the log level should be set to verbose.

moog16:
I see this with verbose as well.

PR #4747, opened 2019-05-21
fix: move applyPassive to dom package for use in text-field

moog16:
fixes https://github.com/material-components/material-components-web/issues/4460


packages/mdc-ripple/util.ts
```diff
@@ -86,30 +80,6 @@ export function supportsCssVariables(windowObj: Window, forceRefresh = false): b
   return supportsCssVars;
 }
 
-/**
- * Determine whether the current browser supports passive event listeners, and
- * if so, use them.
- */
-export function applyPassive(globalObj: Window = window, forceRefresh = false):
```
abhiomkar: Seems like only IE does not support `{passive: boolean}` option. Doesn't the browser just ignore this option if not supported. Not sure why we've introduced this check in first place. WDYT?

moog16: I looked into it - https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners

> However, in browsers that do not support passive event listeners, the third parameter is a boolean to indicate whether the event should bubble or capture. So, using the syntax above may cause unintended consequences.

I think we need to leave this in here if I understand correctly.

packages/mdc-dom/events.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Stores result from applyPassive to avoid redundant processing to detect
+ * passive event listener support.
+ */
+let supportsPassive_: boolean | undefined;
+
+/**
+ * Determine whether the current browser supports passive event listeners, and
+ * if so, use them.
+ */
+export function applyPassive(globalObj: Window = window, forceRefresh = false):
```
abhiomkar: Can we also apply this for other component that attach touch events (including, TabScroller, Slider?)

moog16:
depends on https://github.com/material-components/material-components-web/pull/4824

PR #4750, opened 2019-05-22
chore(infrastructure): update lerna to latest 3.6.0

moog16:
continuation of https://github.com/material-components/material-components-web/pull/3328


packages/mdc-drawer/package-lock.json
```diff
@@ -1,8 +1,6 @@
 {
-  "name": "@material/drawer",
```
abhiomkar: Revert -lock changes.

moog16:
Yes - but this is reflected in https://github.com/material-components/material-components-web/pull/4751. I separated the PRs.

PR #4761, opened 2019-05-28
feat: Updated constant files with supported cssClasses

abhiomkar:
Refs https://github.com/material-components/material-components-web-react/issues/827

by @TroyTae

### DIFF
--- a/packages/mdc-linear-progress/constants.ts
+++ b/packages/mdc-linear-progress/constants.ts
@@ -22,9 +22,19 @@
  */
 
 export const cssClasses = {
-  CLOSED_CLASS: 'mdc-linear-progress--closed',
-  INDETERMINATE_CLASS: 'mdc-linear-progress--indeterminate',
-  REVERSED_CLASS: 'mdc-linear-progress--reversed',
+  BAR: 'mdc-linear-progress__bar',
+  BAR_INNER: 'mdc-linear-progress__bar-inner',
+  BUFFER: 'mdc-linear-progress__buffer',
+  BUFFERING_DOTS: 'mdc-linear-progress__buffering-dots',
+  CLOSED: 'mdc-linear-progress--closed',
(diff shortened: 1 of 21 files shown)


packages/mdc-linear-progress/constants.ts
```diff
@@ -22,9 +22,19 @@
...
 export const cssClasses = {
-  CLOSED_CLASS: 'mdc-linear-progress--closed',
-  INDETERMINATE_CLASS: 'mdc-linear-progress--indeterminate',
-  REVERSED_CLASS: 'mdc-linear-progress--reversed',
+  BAR: 'mdc-linear-progress__bar',
+  BAR_INNER: 'mdc-linear-progress__bar-inner',
+  BUFFER: 'mdc-linear-progress__buffer',
+  BUFFERING_DOTS: 'mdc-linear-progress__buffering-dots',
+  CLOSED: 'mdc-linear-progress--closed',
+  CLOSED_CLASS: 'mdc-linear-progress--closed', // deprecated
```
moog16: if these are deprecated, shouldn't we just do a rename through the rest of the package?

TroyTae: I want to notice about deprecated constants.
This is my original my code:
```
const cssClasses = {
  get CLOSED_CLASS() {
    console.warn('MDCWeb: CLOSED_CLASS is deprecated, use CLOSED instead');
    return 'mdc-linear-progress--closed';
  }
};
```
But mdc tslint prevent code about console 😢 
If you think it's enough to notice through the braking change, I'll remove it!

abhiomkar: We don't have to throw warning. But we'll have to update existing codebase to use these new constants. :)

TroyTae: Yes it looks to do that.
I'll refactor in the `feat/update_constants` branch separately from this PR :)

Issue #4591, opened 2019-04-11
Implement MVP of Datatable component


abhiomkar:
Implement MVP of datatable component:

- [x] Add base styles for standard variant. Implement sass mixins for customizations. Screenshot test with base styles. (PR #4600)
- [x] Implement interactions and new event to select / deselect table rows. Screenshot test with row selection. (PR #4662)
- [x] Update docs. (PR #4694)
- [ ] Update catalog.

Spec reference: http://material.io/go/design-data-tables

fopsdev:
Just curious when this will land?

moog16:
There is a PR open to merge into master right now.

PR #4764, opened 2019-05-28
feat(data-table): Added separate layout method for async support

abhiomkar:
### Changes

* Added `layoutAsync()` foundation method for cases when registering checkbox callback is asynchronous. Vanilla implementation still uses `layout()` which is synchronous.
* `isCheckboxAtRowIndexChecked()` adapter method to replace event sniffing `event.target.checked`.

Refs #4591

### DIFF
--- a/packages/mdc-data-table/README.md
+++ b/packages/mdc-data-table/README.md
@@ -243,7 +243,8 @@ Method Signature | Description
 
 Method Signature | Description
 --- | ---
-`layout() => void` | Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table.
+`layout() => void` | Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table. Use this if registering checkbox is synchronous.
+`layoutAsync() => void` | Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table. Use this if registering checkbox is asynchronous.
 `getRows() => HTMLElement[]` | Returns array of row elements.
 `setSelectedRowIds(rowIds: string[]) => void` | Sets selected row ids. Overwrites previously selected rows.
 `getSelectedRowIds() => Array<string|null>` | Returns array of selected row ids.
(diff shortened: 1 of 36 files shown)


packages/mdc-data-table/README.md
```diff
@@ -243,7 +243,8 @@ Method Signature | Description
 
 Method Signature | Description
 --- | ---
-`layout() => void` | Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table.
+`layout() => void` | Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table. Use this if registering checkbox is synchronous.
+`layoutAsync() => void` | Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table. Use this if registering checkbox is asynchronous.
```
moog16: does Async mean checkbox is added after the fact? Or do you mean if checkbox is initialized after the datatable after an event has occurred? Just want to know of the use case.

abhiomkar: layoutAsync() itself is asynchronous because registerHeaderRowCheckbox and registerRowCheckboxes can be asynchronous in this case.

This supports use case where registerHeaderRowCheckbox / registerRowCheckboxes does not register checkbox instances synchronously. So the client would've to wait till all promises are resolved.

Example,

```ts
async updateTable() {
  await layout();
}
```

joyzhong: Document that `layoutAsync` returns a Promise, not void. Can you also modify the description to specify the registerCheckbox methods which are sync/async?

abhiomkar: Done.

packages/mdc-data-table/foundation.ts
```diff
@@ -65,10 +71,23 @@ export class MDCDataTableFoundation extends MDCFoundation<MDCDataTableAdapter> {
...
   }
 
+  /**
+   * Re-initializes header row checkbox and row checkboxes when selectable rows are added or removed from table.
+   * Use this if registering checkbox is asynchronous.
+   */
+  async layoutAsync() {
+    if (this.adapter_.isRowsSelectable()) {
+      await this.adapter_.registerHeaderRowCheckbox();
+      await this.adapter_.registerRowCheckboxes();
```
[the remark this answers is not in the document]
abhiomkar: Registering checkbox shouldn't take much time, I think doing in parallel brings complexity to foundation without much gain in performance.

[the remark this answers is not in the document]
joyzhong: SGTM.

PR #4777, opened 2019-05-31
fix(menu): In Windows high contrast mode, decrease opacity of disabled menu items.

joyzhong:
Currently, there's no differentiation between disabled/enabled menu items in Windows high contrast mode.


packages/mdc-menu/_mixins.scss
```diff
@@ -66,6 +66,16 @@
       }
     }
 
+    @media screen and (-ms-high-contrast: active) {
```
abhiomkar: Add code comment referencing that we're targeting IE to fix hight contrast.

PR #4778, opened 2019-05-31
chore: Karma config changes to add babel polyfill in test setup

abhiomkar:
### Changes

* This setup will allow us to use `async/await` and other JS features that are not supported in all browsers.
* Also added reference links that points to respective documentation.


package.json
```diff
@@ -43,7 +43,9 @@
     "test:site": "npm run clean:site && ./scripts/site-generator-test.sh"
   },
   "devDependencies": {
+    "@babel/core": "^7.4.5",
```
moog16: I don't see you using this anywhere? Is core necessary?

abhiomkar: This is required by https://github.com/babel/karma-babel-preprocessor

PR #4779, opened 2019-05-31
fix(menu): Switch from aria-selected to aria-checked for selected menu item.

joyzhong:
aria-checked is advised in ARIA spec for selectable menu items:
- https://www.w3.org/TR/wai-aria-1.1/#menuitemcheckbox
- https://www.w3.org/TR/wai-aria-1.1/#menuitemradio


joyzhong:
> LGTM.
> 
> Please also update examples in README & screenshot test files.

No README/screenshot test updates needed, as there were no references to `aria-selected`.

Issue #4738, opened 2019-05-20
Clean up: demos/deprecated packages


moog16:
- [x] Remove demos
  - [x] Remove package deps for demos
- [x] Remove deprecated packages
   - [x] tabs
   - [x] ~~icon toggle~~ # already removed
   - [x] toolbar

moog16:
waiting for https://github.com/material-components/material-components-web/pull/4752 before moving ahead with remove tabs, icontoggle, toolbar.

moog16:
Once https://github.com/material-components/material-components-web/pull/4687 is merged, this can be closed.

abhiomkar:
Toolbar is removed, we can close this once #4784 is merged.

moog16:
All deprecated packages removed!

PR #4784, opened 2019-06-03
fix(tabs): Remove deprecated package mdc-tabs

moog16:
related https://github.com/material-components/material-components-web/issues/4738

Deprecating mdc-tabs. mdc-tabs, mdc-tab-bar, mdc-tab-scroller, and mdc-tab-indicator are used in its place. 

BREAKING CHANGE: remove mdc-tabs package.


moog16:
blocked by https://github.com/material-components/material-components-web/pull/4824

PR #4785, opened 2019-06-04
feat(chips): Add simple keyboard navigation

patrickrodee:
Add simple support for chip keyboard navigation

BREAKING CHANGE: Requires updates to chip and chip set implementations to meet adapter and foundation updates.


patrickrodee:
I have no clue why changes to linear progress indicator and the wrappers.md file got picked up by a merge with origin/develop.

PR #4789, opened 2019-06-04
fix(menu): Fix bug where TAB does not respect the default browser tab order.

joyzhong:


packages/mdc-menu/adapter.ts
```diff
@@ -57,8 +57,10 @@ export interface MDCMenuAdapter {
 
   /**
    * Closes the menu-surface.
+   * @param skipRestoreFocus Whether to skip restoring focus to the previously
+   *    focused element after the surface has been closed.
    */
-  closeSurface(): void;
+  closeSurface(skipRestoreFocus?: boolean): void;
```
abhiomkar: to avoid double negation can we name this to `restoreFocus` instead?

packages/mdc-menu/foundation.ts
```diff
@@ -82,7 +82,7 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
     const isTab = key === 'Tab' || keyCode === 9;
 
     if (isTab) {
-      this.adapter_.closeSurface();
+      this.adapter_.closeSurface(true);
```
abhiomkar: nit: `this.adapter_.closeSurface(/** restoreFocus */ true);`

This'll document the boolean value.

test/unit/mdc-menu-surface/mdc-menu-surface.test.js
```diff
@@ -84,20 +84,18 @@ test('destroy deregisters key handler on the menu surface', () => {
   td.verify(mockFoundation.handleKeydown(td.matchers.isA(Event)), {times: 0});
 });
 
-test('get/set open', () => {
+test('#isOpen', () => {
```
abhiomkar: Can we also update unit test for mdc-menu?

Thanks!

packages/mdc-menu/adapter.ts
joyzhong: Since this is an optional param (with default false), and restoring focus is the default behavior in most cases, I prefer leaving this as is so we only need to specify `skipRestoreFocus = true` in the small # of use cases it's actually needed (TAB key for now). Wdyt?

abhiomkar: Makes sense.. Thanks!

PR #4797, opened 2019-06-07
refactor(select): Remove support for native select.

joyzhong:
Native select is not present in the Material spec (https://material.io/design/components/menus.html#usage), so we are removing support for it.

BREAKING CHANGE: Support for native select menu was removed. Use the enhanced (now the default) select menu instead.


cintaccs:
fair enough that you kill support for this native select - but...
1. isn't the idea of the native select that it should kick in and be used automatically on mobile devices... (to use the native controls for select)?
2. as long as the MDC enhanced select is throwing DOM elements around out of scope from the origin - and the cleanup process isn't automated / handled ... it kind of leaves no options to use the select component without introducing this kind of issues....

... therefore ... what about fixing and enhancing the component library before deleting too many milestone listings and features...?

joyzhong:
1. This was the original idea, but as it's not part of the Material spec and adds additional complexity when refactoring this component, we are removing support for it.
2. A follow-up change will refactor the enhanced select component such that we no longer hoist the menu to the body element, so cleanup should no longer be an issue.

Note that we'll also be doing some refactoring to move much of the component logic to the foundation, to make this cleaner to use for our wrapper libraries.

cintaccs:
> 1. This was the original idea, but as it's not part of the Material spec and adds additional complexity when refactoring this component, we are removing support for it.
> 2. A follow-up change will refactor the enhanced select component such that we no longer hoist the menu to the body element, so cleanup should no longer be an issue.
> 
> Note that we'll also be doing some refactoring to move much of the component logic to the foundation, to make this cleaner to use for our wrapper libraries.

Thanks for the quick reply! 
Re. 2 - sounds good!!

does this means that the wrapper libraries becomes more and more necessary as you see it? i.e. the MDC library alone will not do much without also depending on wrapper libraries...? A quality from my point of view of the current MDC library is that it is standalone - and I would expect Google supported - and therefore a strong reason for the MDC library to become defacto standard for MDC web controls. However if it is tied too much and depending too much on REACT or Angular or whatever flavour of the year :-) it would not have a long lifetime - and in my opinion that would be a waste of a useful library and a re-use of efforts that everyone can and should benefit from to gain MD success and higher end user satisfaction...

joyzhong:
> does this means that the wrapper libraries becomes more and more necessary as you see it? i.e. the MDC library alone will not do much without also depending on wrapper libraries...?

We do not plan to remove support for the vanilla MDC library components, nor to tie our components to one particular framework. In fact, we'd like to improve our library foundation code to make it easier for more frameworks to wrap, which is one of the motivations for the refactoring of the select component.

packages/mdc-select/component.ts
```diff
@@ -180,11 +173,11 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
       }
     };
 
-    this.targetElement_.addEventListener('change', this.handleChange_);
-    this.targetElement_.addEventListener('focus', this.handleFocus_);
-    this.targetElement_.addEventListener('blur', this.handleBlur_);
+    this.selectedText_.addEventListener('change', this.handleChange_);
```
abhiomkar: Do we still need 'change' event?

packages/mdc-select/component.ts
```diff
@@ -180,11 +173,11 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
...
+    this.selectedText_.addEventListener('focus', this.handleFocus_);
+    this.selectedText_.addEventListener('blur', this.handleBlur_);
 
-    this.targetElement_.addEventListener('click', this.handleClick_ as EventListener);
+    this.selectedText_.addEventListener('click', this.handleClick_ as EventListener);
 
     if (this.menuElement_) {
```
abhiomkar: We might not need to check for this condition since menu will always be present.

packages/mdc-select/component.ts
joyzhong: Good call!

packages/mdc-select/component.ts
```diff
@@ -53,17 +53,13 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
 
   ripple!: MDCRipple | null;
 
-  private menu_!: MDCMenu | null; // assigned in enhancedSelectSetup_()
+  private menu_!: MDCMenu; // assigned in enhancedSelectSetup_()
```
abhiomkar: nit: Remove/update `enhancedSelectSetup_` code comment.

Same below.

packages/mdc-select/component.ts
```diff
@@ -88,20 +84,17 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
       helperTextFactory: MDCSelectHelperTextFactory = (el) => new MDCSelectHelperText(el),
   ) {
     this.isMenuOpen_ = false;
-    this.nativeControl_ = this.root_.querySelector(strings.NATIVE_CONTROL_SELECTOR);
-    this.selectedText_ = this.root_.querySelector(strings.SELECTED_TEXT_SELECTOR);
+    this.selectedText_ = this.root_.querySelector(strings.SELECTED_TEXT_SELECTOR) as HTMLElement;
 
-    const targetElement = this.nativeControl_ || this.selectedText_;
-    if (!targetElement) {
+    if (!this.selectedText_) {
```
abhiomkar: We usually do not verify for existence of supporting DOM elements. We can remove this check.

packages/mdc-select/component.ts
```diff
@@ -376,68 +348,40 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
...
       },
       setValue: (value: string) => {
-        const element = this.menuElement_!.querySelector(`[${strings.ENHANCED_VALUE_ATTR}="${value}"]`);
-        this.setEnhancedSelectedIndex_(element ? this.menu_!.items.indexOf(element) : -1);
+        const element = this.menuElement_.querySelector(`[${strings.ENHANCED_VALUE_ATTR}="${value}"]`);
+        this.setEnhancedSelectedIndex_(element ? this.menu_.items.indexOf(element) : -1);
       },
       openMenu: () => {
-        if (this.menu_ && !this.menu_.open) {
+        if (!this.menu_.open) {
```
abhiomkar: We might not need this check here. We can refactor this in upcoming PRs.

test/unit/mdc-select/mdc-select-enhanced.test.js
```diff
@@ -903,12 +903,6 @@ test(`adapter#setValid applies ${cssClasses.INVALID} properly`, () => {
   assert.isFalse(fixture.classList.contains(cssClasses.INVALID));
 });
 
-test('change event triggers foundation.handleChange()', () => {
```
abhiomkar: Rename this test file to `mdc-select.test.js`?

test/unit/mdc-select/mdc-select.test.js
```diff
@@ -1,686 +0,0 @@
-/**
```
abhiomkar: 🎉

packages/mdc-select/component.ts
joyzhong: Renamed these in a future PR.

packages/mdc-select/component.ts
joyzhong: Noted.

msrumon:
So, any doc on how we can use enhanced select within form?

PR #4800, opened 2019-06-07
fix(infrastructure): Fix failing screenshot tests

patrickrodee:
Appears to be an update to Firefox.


patrickrodee:
Thanks for the fast reviews 💜

PR #4806, opened 2019-06-10
fix(text-field): Fix asterisk color of text field when input is invalid and disabled 

### LINKED ISSUE #4805: [TextField] The color of asterisk stays red when input is invalid and disabled.
### Steps to reproduce

1. Go to [text field demo on Glitch](https://glitch.com/edit/#!/mercury-tuesday?path=index.html:1:0).
2. Notice that textfields that are not initialized and disabled has red asterisk.

### Actual behavior

Asterisk is in red color even though input is disabled

### Expected behavior

Asterisk needs to be in grey color similar to floating label color when input is disabled.

### Your Environment:

| MDC Web          | `v2.3.0`
| Browser          | All
| Operating System | All

### Possible solution

Set asterisk color on invalid excluding `--disabled` text fields.

abhiomkar:
Fixes #4805

### DIFF
--- a/test/screenshot/spec/mdc-textfield/fixture.js
+++ b/test/screenshot/spec/mdc-textfield/fixture.js
@@ -22,7 +22,8 @@
  */
 
 window.mdc.testFixture.fontsLoaded.then(() => {
-  [].forEach.call(document.querySelectorAll('.mdc-text-field'), (el) => {
+  [].forEach.call(document.querySelectorAll('.mdc-text-field:not([data-no-init="true"])'), (el) => {
+    console.log('init');
     mdc.textField.MDCTextField.attachTo(el);
   });
 
--- a/test/screenshot/spec/mdc-textfield/fixture.js
+++ b/test/screenshot/spec/mdc-textfield/fixture.js
@@ -23,7 +23,6 @@
 
 window.mdc.testFixture.fontsLoaded.then(() => {
   [].forEach.call(document.querySelectorAll('.mdc-text-field:not([data-no-init="true"])'), (el) => {
-    console.log('init');
     mdc.textField.MDCTextField.attachTo(el);
   });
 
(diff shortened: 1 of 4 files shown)


test/screenshot/spec/mdc-textfield/fixture.js
```diff
@@ -22,7 +22,7 @@
  */
 
 window.mdc.testFixture.fontsLoaded.then(() => {
-  [].forEach.call(document.querySelectorAll('.mdc-text-field'), (el) => {
+  [].forEach.call(document.querySelectorAll('.mdc-text-field:not([data-no-init="true"])'), (el) => {
```
moog16: Do we have anywhere in our docs that you shouldn't initialize a disabled text field?

abhiomkar: Internal implementation does not initialize the component until interacted. In that case asterisk would still be in red color.

moog16: I'm suggesting that we need to add some text to the Readme so that developers know not to initialize the text field until disabled is removed.

abhiomkar: Clients can definitely initialize text field even if it is disabled, why do you think they don't need to initialize when it is disabled?

moog16: I'm just wondering why you aren't initializing the text field.

moog16: I thought you said this morning that initializing it will remove `--invalid` class.

abhiomkar: Two reasons for this screenshot without initializing text field:

- Validates the styles (Only style without JS) for disabled & invalid state. Our screenshots are captured after JS execution that removes `--invalid` modifier from root. There might be some uses cases where clients do not need to initialize text field for some reason.
- Our internal implementation, we do not instantiate the foundation until the text field is interacted hence the asterisk stays red on page load which is unexpected. This test case should cover this case.

PR #4823, opened 2019-06-11
refactor(select): Move #get/setSelectedIndex component logic to Foundation.

joyzhong:
BREAKING CHANGE: Added to select menu component the following methods: Foundation#getSelectedIndex, Adapter#getSelectedMenuItem, Adapter#getMenuItems, Adapter#setAttributeForElement, Adapter#getMenuItemWithValueAttribute, Adapter#getMenuItemText, Adapter#toggleMenuItemSelectedClass. Also removes support for a hidden input element inside the select.


packages/mdc-select/component.ts
```diff
@@ -465,10 +467,10 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> implements MDCR
     };
   }
 
-  private setEnhancedSelectedIndex_(index: number) {
+  private setSelectedIndex_(index: number) {
```
abhiomkar: We can potentially split this adapter method into smaller adapter methods. `setSelectedIndex` adapter method seems to be doing many DOM operations which is again harder to implement by Frameworks. This also makes harder to support business logic of `setSelectedIndex` adapter method that are internal to Foundation.

WDYT?

joyzhong: Great idea - done.

packages/mdc-select/foundation.ts
```diff
@@ -87,15 +92,41 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+  setSelectedIndex(index: number, closeMenu = false) {
+    const menuItemToSelect = this.adapter_.getMenuItems()[index];
+    this.adapter_.setSelectedText(
+        menuItemToSelect ? this.adapter_.getMenuItemText(menuItemToSelect).trim() : '');
+
+    const previouslySelectedItem = this.adapter_.getSelectedMenuItem();
+    if (previouslySelectedItem) {
+      this.adapter_.toggleMenuItemSelectedClass(previouslySelectedItem, false);
+      // Remove the ARIA_SELECTED attribute.
+      this.adapter_.setAttributeForElement(previouslySelectedItem, strings.ARIA_SELECTED_ATTR);
```
abhiomkar: This adapter method can refer index instead of directly referencing DOM element.

`setAttributeAtIndex(index: number, name: string, value: string)`

See MDCListAdapter for example.

packages/mdc-select/foundation.ts
```diff
@@ -87,15 +92,41 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+    }
+    if (menuItemToSelect) {
+      this.adapter_.toggleMenuItemSelectedClass(menuItemToSelect, true);
```
abhiomkar: Same here. This adapter method can refer Index.

We can use even more generic adapter for future usecases where we need to add another class name:

`addClassAtIndex(index: number, className: string)`

`removeClassAtIndex(index: number, className: string)`

packages/mdc-select/adapter.ts
```diff
@@ -90,47 +85,70 @@ export interface MDCSelectAdapter {
...
+   * Returns true if the menu is currently open.
    */
-  setValid(isValid: boolean): void;
+  isMenuOpen(): boolean;
+
+  /**
+   * Sets the attribute on the menu item at the given index. If `attributeValue`
+   * is undefined, removes the attribute.
+   */
+  setAttributeAtIndex(index: number, attributeName: string, attributeValue?: string): void;
```
abhiomkar: To make this explicite, do you think we need to split this into two separate adapter methods for set and remove?

It'll be consistant with other adapter methods.

packages/mdc-select/foundation.ts
```diff
@@ -87,15 +93,41 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+      this.adapter_.closeMenu();
+    }
+
     const didChange = true;
     this.handleChange(didChange);
   }
 
   setValue(value: string) {
-    this.adapter_.setValue(value);
+    const index = this.adapter_.getIndexOfMenuItemWithAttribute(strings.VALUE_ATTR, value);
```
abhiomkar: We're using DOM as source of truth here, we can store array of values in foundation instead where index serves as item key. This'll be retrieved on init for first time.

Example:
`this.adapter_.getItemValues(): string[]`

packages/mdc-select/foundation.ts
```diff
@@ -87,15 +93,41 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+    this.selectedIndex_ = index;
+
+    this.adapter_.setSelectedText(
+        this.selectedIndex_ === numbers.UNSET_INDEX ?
+            '' : this.adapter_.getMenuItemTextAtIndex(this.selectedIndex_).trim());
+
+    if (previouslySelectedIndex !== numbers.UNSET_INDEX) {
+      this.adapter_.toggleClassAtIndex(previouslySelectedIndex, cssClasses.SELECTED_ITEM_CLASS, false);
+      // Remove the ARIA_SELECTED attribute.
+      this.adapter_.setAttributeAtIndex(previouslySelectedIndex, strings.ARIA_SELECTED_ATTR);
```
abhiomkar: As mentioned in my previous comment, I think having a explicite `remove*` adapter method makes it easier to read, IMO.

joyzhong:
BREAKING CHANGE: Added to select menu component the following methods: Foundation#getSelectedIndex, Adapter#setAttributeAtIndex, Adapter#removeAttributeAtIndex, Adapter#getMenuItemValues, Adapter#getMenuItemTextAtIndex, Adapter#toggleClassAtIndex. Also removes support for a hidden input element inside the select.

PR #4830, opened 2019-06-13
refactor(select): Don't hoist menu to body el on open.

joyzhong:
BREAKING CHANGE: The menu is no longer hoisted to the body element when opening the select menu. This necessitates a restructuring of the HTML structure of the select component: the select anchor is now wrapped in a parent element, and the anchor's sibling is the select menu. See README for examples.


packages/mdc-menu-surface/component.ts
```diff
@@ -93,7 +93,7 @@ export class MDCMenuSurface extends MDCComponent<MDCMenuSurfaceFoundation> {
   }
 
   /**
-   * Removes the menu-surface from it's current location and appends it to the
+   * Removes the menu-surface from its current location and appends it to the
```
abhiomkar: Can we remove this method from menu surface since no longer used?

packages/mdc-select/foundation.ts
```diff
@@ -87,6 +89,12 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
     this.helperText_ = foundationMap.helperText;
   }
 
+  /** Returns the index of the currently selected menu item, or -1 if none. */
+  getSelectedIndex(): number {
+    const selectedMenuItem = this.adapter_.getSelectedMenuItem();
```
abhiomkar: The state of selectedIndex needs to be stored in foundation. Currently seems like we're using DOM as source of truth to retrieve selected index.

See MDCListFoundation for example.

packages/mdc-menu-surface/component.ts
joyzhong: Will do in a follow-up PR, since this one is already huge.

packages/mdc-select/foundation.ts
joyzhong: Will apply changes to PR #4823, since its changes were partially merged into this one (on accident). Sorry for the confusion!

packages/mdc-select/component.ts
```diff
@@ -43,23 +42,24 @@ import {MDCSelectEventDetail, MDCSelectFoundationMap} from './types';
...
-  private menu_!: MDCMenu; // assigned in enhancedSelectSetup_()
+  private menu_!: MDCMenu; // assigned in selectSetup_()
   private isMenuOpen_!: boolean; // assigned in initialize()
 
+  private selectAnchor_!: HTMLElement; // assigned in initialize()
   private selectedText_!: HTMLElement; // assigned in initialize()
 
-  private hiddenInput_!: HTMLInputElement | null; // assigned in enhancedSelectSetup_()
-  private menuElement_!: Element; // assigned in enhancedSelectSetup_()
+  private hiddenInput_!: HTMLInputElement | null; // assigned in selectSetup_()
```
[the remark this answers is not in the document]
joyzhong: That's done in https://github.com/material-components/material-components-web/pull/4823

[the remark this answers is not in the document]
joyzhong: (README had changes from my previous accidental merge with #4823. Leaving as is for now, since I'll be eventually merging the two PR's anyways.)

joyzhong:
Note that screenshot changes are due to renaming of screenshots due to previous PR which removed support for native select (s/enhanced//).

packages/mdc-select/mdc-select.scss
```diff
@@ -37,6 +37,35 @@
...
+      transition: transform $mdc-select-dropdown-transition-duration $mdc-animation-standard-curve-timing-function;
+    }
+  }
+
+  &__native-control {
+    padding-top: 20px;
+  }
+}
+
+.mdc-select__anchor {
```
abhiomkar: All public mixins has to be applied at root class name.

joyzhong: As discussed, will do in a follow-up PR.

PR #4831, opened 2019-06-13
chore: Updated lerna config as per new release process

abhiomkar:
### Description

- Updated `lerna.json` file supported by new lerna version. Set `command.version.conventionalCommits` to `true` which automatically determines the next release version based on our commit conventions.
- Removed CHANGELOG dependencies: `standard-changelog`, `conventional-changelog-angular`, `conventional-commits-parser`, `cz-conventional-changelog` to use lerna provided changelog script.
- Removed check for MDC Web demo site deploy in pre-release script.
- Removed `husky` - commit message validation. Going forward, This commit messages will be validated in PR title using [GitHub bot](https://github.com/probot/semantic-pull-requests)

PR #4834, opened 2019-06-14
chore: Added config file for semantic PR bot

abhiomkar:
See https://github.com/probot/semantic-pull-requests


abhiomkar:
Husky is removed in #4831

PR #4838, opened 2019-06-15
feat: add feature targeting for styles to tab-related packages

crisbeto:
Adds support for style feature targeting to the `tab`, `tab-bar`, `tab-indicator` and `tab-scroller` packages.


crisbeto:
cc @mmalerba

packages/mdc-tab-bar/_mixins.scss
```diff
@@ -1,25 +1,39 @@
...
+  // postcss-bem-linter: define tab-bar
+  .mdc-tab-bar {
+    @include mdc-feature-targets($feat-structure) {
+      @include mdc-tab-bar-width(100%);
+    }
+  }
+  // postcss-bem-linter: end
+}
 
 @mixin mdc-tab-bar-width($width) {
```
abhiomkar: I think every mixin needs optional `$query` parameter.

packages/mdc-tab-bar/_mixins.scss
```diff
@@ -1,25 +1,39 @@
...
+@import "@material/feature-targeting/functions";
+@import "@material/feature-targeting/mixins";
+
+@mixin mdc-tab-bar-core-styles($query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  // postcss-bem-linter: define tab-bar
+  .mdc-tab-bar {
+    @include mdc-feature-targets($feat-structure) {
+      @include mdc-tab-bar-width(100%);
```
mmalerba: We want all the public mixin APIs to support passing a `$query`, so it should be passed through to this one even though it's trivial

packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -1,27 +1,81 @@
...
+  $feat-structure: mdc-feature-create-target($query, structure);
+  $feat-animation: mdc-feature-create-target($query, animation);
+
+  // postcss-bem-linter: define tab-indicator
+
+  .mdc-tab-indicator {
+    @include mdc-tab-indicator-base_($query);
+  }
+
+  @include mdc-feature-targets($feat-structure) {
```
abhiomkar: `mdc-feature-targets()` mixins will need to be included within the CSS selector for consistancy with other feature target refactor. This'll ensure we don't accidentally add another CSS property in same CSS selector which is not related to structure.

```sass
.mdc-tab-indicator__content {
    transform-origin: left;
    opacity: 0;
  }
}
```

See chips for example.

packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -1,27 +1,81 @@
...
+    .mdc-tab-indicator__content {
```
mmalerba: So far we've been following a rule of "don't wrap selectors in `mdc-feature-targets`". The reasoning is that its easier to follow what element you're targeting when you can read down the whole selector tree first before worrying about what features are being targeted. I realize in this case its all 1-liners and it would be annoying to keep writing `@include mdc-feature-targets($feat-structure)` for each one. 

Anyways I just wanted to point this out, I'll let @abhiomkar decide on whether we want to strictly follow the style rule in this case.

packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -1,27 +1,81 @@
...
+    // postcss-bem-linter: ignore
+    .mdc-tab-indicator--active.mdc-tab-indicator--fade .mdc-tab-indicator__content {
+      @include mdc-tab-indicator--active-fade__content_;
+    }
+  }
+
+  // postcss-bem-linter: end
+}
 
 @mixin mdc-tab-indicator-surface {
```
mmalerba: These public mixins should also take a `$query`

packages/mdc-tab/_mixins.scss
```diff
@@ -20,11 +20,89 @@
...
+// from the other tah styles. It is recommended that most users use `mdc-tab-core-styles` instead.
+@mixin mdc-tab-ripple($query: mdc-feature-all()) {
+  @include mdc-ripple-common($query);
+
+  .mdc-tab__ripple {
+    @include mdc-tab__ripple_($query);
+  }
+}
+
 @mixin mdc-tab-horizontal-padding($padding) {
```
mmalerba: public mixins should take `$query`

crisbeto:
@abhiomkar @mmalerba I've reworked it based on the feedback.

mmalerba:
Can you give a diff of the final CSS just to sanity check? Code looks good to me

crisbeto:
Sure, here's the before and after. The main difference is that the order of a few selectors has changed, but from what I could tell it shouldn't have an effect on the end result. Also the `after` styles are a bit smaller, because the current ones in master include an extra license header in the middle of the file.

[before.zip](https://github.com/material-components/material-components-web/files/3299934/before.zip)
[after.zip](https://github.com/material-components/material-components-web/files/3299933/after.zip)

mmalerba:
Most of the look pretty trivial, though mdc.tab.css has some significant movement (maybe that's finem just wanted to point it out: http://www.mergely.com/53ydf5dw/)

packages/mdc-tab-bar/_mixins.scss
```diff
@@ -1,26 +1,40 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
 
-@mixin mdc-tab-bar-width($width) {
-  width: $width;
+@import "@material/feature-targeting/functions";
```
abhiomkar: Update mdc-tab* dependencies in package.json of 3 packages with `@material/feature-targeting`.

packages/mdc-tab-indicator/_mixins.scss
```diff
@@ -1,60 +1,211 @@
...
+    left: 0;
+    width: 100%;
+    height: 100%;
+    pointer-events: none;
+    z-index: 1;
+  }
+}
+
+@mixin mdc-tab-indicator__content_ {
+  transform-origin: left;
```
abhiomkar: As mentioned in my previous comment, please include feature target mixins here. This'll avoid accidentally adding future styles properties in unexpected feature target.

Same with other mixin files.

packages/mdc-tab/_mixins.scss
```diff
@@ -20,64 +20,301 @@
...
+}
+
+@mixin mdc-tab__icon_ {
+  width: $mdc-tab-icon-size;
+  height: $mdc-tab-icon-size;
+  font-size: $mdc-tab-icon-size;
+}
+
+@mixin mdc-tab--stacked_ {
+  height: $mdc-tab-stacked-height;
```
abhiomkar: Same here. Please include feature target mixins for each class name.

packages/mdc-tab-indicator/_mixins.scss
crisbeto: I thought that only applied to public mixins? From what I can tell it isn't being done in any of the other private mixins, unless it involves more than one feature target.

abhiomkar: Can we in-line these styles above? This is essentially `mdc-tab-indicator__content_` with structure only feature target styles. I'm wary about accidentally adding a style here which is not `structure` feature target in future. 

WDYT?

crisbeto: I wouldn't mind inlining them, but we did some of the other components with a similar level of granularity. This actually came in handy for the Angular Material menu since it allowed us to shave off a lot of styles that we didn't need.

abhiomkar: Sorry for not being clearer in my previous comment. We would still render these styles behind `structure` feature-target. I'm suggesting we can remove the need for private mixin and move it at the top.

```sass
.mdc-tab-indicator__content {
    transform-origin: left;
    opacity: 0;
  }
}
```

instead of

```sass
.mdc-tab-indicator__content {
    @include mdc-tab-indicator__content_;
  }
}
```

If we ever need to add additional animation related styles to `.mdc-tab-indicator__content` then we would do so by using animation feature target mixin inside the same selector.

/cc @mmalerba

mmalerba: Yeah makes sense to me, it looks like some of these private mixins are just an alias for one or two properties, so you might as well just inline them. For ones that are bigger / included in multiple places we could either pass the query through or just put a comment at the top of the mixin to indicate that all of the styles belong to whichever feature

crisbeto:
I've reworked it based on the latest set of feedback: I got rid of the smaller private mixins and the ones that are left should have feature targets. Can you take another look @abhiomkar?

PR #4845, opened 2019-06-18
chore: remove webpack-dev-server from package.json

moog16:
related: https://github.com/material-components/material-components-web/issues/4842


abhiomkar:
`fix: ...` commit message appears as bug fixes in changelog, so changed it to `chore: ...`

PR #4847, opened 2019-06-19
chore(infrastructure): update webpack v4.x

moog16:
fixes #4847 

webpack3 (master branch):
- running `npm run dist`:
  - 2:22
  - 2:30
  - 2:17
  - average: 143s

~~webpack4 (with 2 minified file build):~~
~~- running `npm run dist`:~~
  ~~- 1:09~~
  ~~- 1:09~~
  ~~- 1:12~~

webpack4 (with minified and non-minified builds):
- running `npm run dist`:
  - 1:08
  - 1:08
  - 1:11
  - average: 71.7s

**Build time decreased by 49.14%**


abhiomkar:
Why would be remove *.min.js files?

abhiomkar:
Did you see any different in build time compared to Webpack 3?

moog16:
by default they the bundles are all minified. So there's no need to have both as they are the same.

abhiomkar:
We need to continue publishing both minified and un-minified versions like before. Un-minified version is useful for development / debugging.

moog16:
We can provide maps instead of un-minified (which we already actually do). I don't think un-minified is totally necessary.

abhiomkar:
We shouldn't break existing publish structure as part of this upgrade, IMO.

I'm not expert in this area, but it is fairly common practice to publish minified & un-minified.

moog16:
Fair point! I'll make it exactly as it was.

scripts/build/environment.js
```diff
@@ -27,6 +27,8 @@
 
 'use strict';
 
+// TODO: remove this class. this is only used for webpack.
```
moog16: I can add an issue for this, but I didn't wanna do this without chatting with the group about it. This seems to be unnecessary for webpack to work.

moog16:
@abhiomkar ready to go again

scripts/build/environment.js
abhiomkar: yah, we can remove this in upcoming PR.

scripts/cp-pkgs.js
```diff
@@ -110,20 +109,25 @@ async function cpAsset(asset) {
...
-    const destBasename = isAllInOne ? packageDirectory : `mdc.${toCamelCase(packageDirectory.replace(/^mdc-/, ''))}`;
-    const destFilename = path.join(packagePath, 'dist', `${destBasename}.d.ts`);
-
-    console.log(`Writing UMD declarations in ${destFilename.replace(process.cwd() + '/', '')}`);
-    dts.bundle({
-      name,
-      main,
-      out: destFilename,
+    const main = path.join(PACKAGES_DIRECTORY, packageDirectory, './index.d.ts');
+    fs.access(main, fs.constants.F_OK, (err) => {
```
abhiomkar: nit: error*

scripts/cp-pkgs.js
```diff
@@ -110,20 +109,25 @@ async function cpAsset(asset) {
...
+      if (!err) {
```
abhiomkar:
```ts
if (error) {
  return;
}
```

Returns early if error exist. This avoids nested block and makes it easy to read.

scripts/build/environment.js
moog16: created issue https://github.com/material-components/material-components-web/issues/4863

PR #4853, opened 2019-06-21
fix: update TypeScript version to 3.5.x and fix typing errors

### LINKED ISSUE #4852: Upgrade TypeScript version to 3.5.x
Goal: update TS version to 3.5.x.

Opening this as a bug as there are several errors popping up due to `SpecificEventListener`. They are appearing in:

1. form-field/component
1. ripple/component
1. top-app-bar/component

Fixing one of these should fix all instances.

moog16:
fixes #4852 

- Upgrading TS from 3.2.2 to 3.5.2

- Root cause from upgrade is still a bit unknown to me. No [changes](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript) in 3.3, 3.4, or 3.5 should have caused this type error to occur. 

- These type errors should have appeared before. The error occurred on add/removeEventListener calls where the root node was type `Element`. Type `Element` add/removeEventListener calls expected the handler to be of type `EventListenerOrEventListenerObject`, while the `SpecificEventListener<K>` is specifically for node of type `HTMLElement`, and included event handlers like `MouseEvent`, `KeyboardEvent`, etc. These pose problematic as they are a superset of Event. 

- Fix: ~~I changed the `topAppBar`, `formField`, and `ripple` root_ elements to be `HTMLElement` from `Element`. This is more correct as `HTMLElement` is a superset of `Element`. Our components should always be using either a SVG element or HTML element of which type `HTMLElement` covers.~~ I decided to cast the add/removeEventListener node to an `HTMLElement`.


moog16:
Ah sorry it was lost in this commit https://github.com/material-components/material-components-web/pull/4853/commits/7cb4b5f3720839fdc665053aa9d91766f6a105cf. I'm un-reverting

moog16:
updated!

PR #4861, opened 2019-06-26
feat(list): Add mixin for disabled text opacity

patrickrodee:
Set opacity for all items instead of color. This solves problems where high-contrast mode on Windows 7 causes disabled and enabled list items to be indistinguishable.

### DIFF
--- a/packages/mdc-list/README.md
+++ b/packages/mdc-list/README.md
@@ -411,7 +411,7 @@ Mixin | Description
 `mdc-list-item-shape-radius($radius, $rtl-reflexive)` | Sets the rounded shape to list item with given radius size. Set `$rtl-reflexive` to true to flip radius values in RTL context, defaults to false.
 `mdc-list-divider-color($color)` | Sets divider ink color.
 `mdc-list-group-subheader-ink-color($color)` | Sets ink color of subheader text within list group.
-`mdc-list-item-disabled-text-color($opacity`) | Sets the color of the text when the list item is disabled.
+`mdc-list-item-disabled-text-color($color`) | Sets the color of the text when the list item is disabled.
 `mdc-list-item-disabled-text-opacity($opacity`) | Sets the opacity of the text when the list item is disabled.
 
 ### Accessibility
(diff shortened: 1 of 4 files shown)


packages/mdc-list/_mixins.scss
```diff
@@ -393,6 +391,30 @@
   }
 }
 
+@mixin mdc-list-item-disabled-text-opacity($opacity, $query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
```
abhiomkar: This mixin can take disabled color with rgba value instead. We want to move away from controlling opacity.

packages/mdc-list/README.md
```diff
@@ -411,6 +411,8 @@ Mixin | Description
 `mdc-list-item-shape-radius($radius, $rtl-reflexive)` | Sets the rounded shape to list item with given radius size. Set `$rtl-reflexive` to true to flip radius values in RTL context, defaults to false.
 `mdc-list-divider-color($color)` | Sets divider ink color.
 `mdc-list-group-subheader-ink-color($color)` | Sets ink color of subheader text within list group.
+`mdc-list-item-disabled-text-color($opacity`) | Sets the color of the text when the list item is disabled.
```
abhiomkar: $color*

packages/mdc-list/_mixins.scss
```diff
@@ -63,6 +63,8 @@
   @include mdc-list-item-graphic-ink-color(text-icon-on-background, $query);
   @include mdc-list-item-meta-ink-color(text-hint-on-background, $query);
   @include mdc-list-group-subheader-ink-color(text-primary-on-background, $query);
+  @include mdc-list-item-disabled-text-opacity($mdc-list-text-disabled-opacity);
```
abhiomkar: Missing $query

packages/mdc-list/_mixins.scss
abhiomkar: I think this feature target is color for this.

PR #4862, opened 2019-06-27
fix(menu): Vertically center the group icon

patrickrodee:
IE11 does not vertically position the group icon when it is followed by a text item. This fixes that behavior.
UBXZKPFOvPi


moog16:
This can actually be cherry picked into master. This isn't a breaking change.

PR #4873, opened 2019-07-03
chore(infrastructure): automate method description table in README.md

### LINKED ISSUE #4903: Infrastructure for documentation automation
- build out scripts to automatically generate markdown tables
- test this on at least one package (mdc-drawer)
- add `npm run docgen` script to run markdown generation
- markdown should be in same format for material.io

moog16:
related https://app.zenhub.com/workspaces/mdc-web-585809de72a2830c7f7fbaad/issues/material-components/material-components-web/4858

fixes #4903


packages/mdc-drawer/README_BASE.md
```diff
@@ -0,0 +1,444 @@
...
+listEl.addEventListener('click', (event) => {
+  drawer.open = false;
+});
+
+document.body.addEventListener('MDCDrawer:closed', () => {
+  mainContentEl.querySelector('input, button').focus();
+});
+```
+
+<!-- %component-foundation-adapter-spacer% -->
```
moog16: This README_BASE.md file is copied from drawer/README.md, with the table removed from here.

moog16:
Ready for rereview

packages/mdc-drawer/README.md
```diff
@@ -437,55 +437,92 @@ document.body.addEventListener('MDCDrawer:closed', () => {
...
+### MDCDrawerAdapter
 
-Method Signature | Description
---- | ---
-`handleScrimClick() => void` | Handles click event on scrim.
+#### Methods
+
+Name | Signature | Description
+--- | --- | --- 
+notifyOpen | `() => void` | Emits a custom event "MDCDrawer:opened" denoting the drawer has opened.
```
abhiomkar: Can we combine name and signature?

`notifyOpen() => void`

packages/mdc-drawer/README.md
```diff
@@ -437,55 +437,92 @@ document.body.addEventListener('MDCDrawer:closed', () => {
...
+trapFocus | `() => void` | Traps focus on root element and focuses the active navigation element.
+
+
+### MDCModalDrawerFoundation
+
+#### Methods
+
+Name | Signature | Description
+--- | --- | --- 
+init | `() => void` | n/a
```
abhiomkar: Can we not include it in table if TS doc is missing?

packages/mdc-drawer/component.ts
```diff
@@ -34,13 +34,19 @@ import {MDCDrawerFocusTrapFactory} from './util';
 
 const {cssClasses, strings} = MDCDismissibleDrawerFoundation;
 
+/**
+ * @events MDCDrawer:event %-% Emits when the navigation drawer has closed. %-% {}
```
abhiomkar: is this `@event` (Singular)? is this annotation valid for class name? I do not see any examples.

moog16: `@event` for some reason does not work with documentalist. I believe it is a bug - I'll open one against documentalist.

moog16: https://github.com/palantir/documentalist/issues/93

packages/mdc-drawer/README.md
moog16: removed

moog16:
ready for rereview

packages/mdc-drawer/component.ts
```diff
@@ -34,13 +34,18 @@ import {MDCDrawerFocusTrapFactory} from './util';
 
 const {cssClasses, strings} = MDCDismissibleDrawerFoundation;
 
+/**
+ * @events MDCDrawer:closed %-% Emits when the navigation drawer has closed. %-% {}
+ * @events MDCDrawer:opened %-% Emits when the navigation drawer has opened. %-% {}
```
abhiomkar: Follow-up with previous comment.

if `@event` is not supported by documentalist is it possible to still parse this with `@event` as custom annotation tag?

Also, this doesn't have to be a table, it can be a bullet list. I'm sure people surely forget to add this separator for future class names.

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,284 @@
...
+        }))
+        .documentGlobs('packages/**/*') // ← async operation, returns a Promise
+        .then((docs) => {
+          resolve(docs);
+          // TODO - remove when all docs are complete
+          return JSON.stringify(docs, null, 2);
+        })
+        .then((json) => {
+          // TODO: comment out - just for debugging purposes.
+          fs.writeFileSync('docs.json', json);
```
abhiomkar: Remove this line. Also above return statement that stringifies JSON.

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,284 @@
...
+
+  constructor() {
+    this.docData = {};
+    this.markdownBuffer = {};
+  }
+
+  /**
+   * Generates JSON from source files TypeScript documentation.
+   * This contains all the esmodule classes (ie. foundations, adapters, components) in JSON format.
+   * @returns Promise(json)
```
abhiomkar: `Promise<{}>`?

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,284 @@
...
+    this.generateMarkdownFileFromBuffer();
+  }
+
+  /**
+   * Creates documentation for a specified `esmodule`.
+   * @param esmodule module name (ie. MDCSelectIconFoundation)
+   */
+  generateDocsForModule(esmodule: string) {
+    const moduleName = `${markdownHeaderLevel} ${esmodule}\n\n`;
+    const markdownString =
```
abhiomkar: I'm wary about concatenating strings, this makes it hard to reformat the structure later. We need a templating engine, something like Nunjucks. This makes it easier to restructure the markdown output without looking at the code.

packages/mdc-drawer/adapter.ts
```diff
@@ -45,6 +45,7 @@ export interface MDCDrawerAdapter {
   hasClass(className: string): boolean;
 
   /**
+   * Returns true if the an element contains the given class.
```
abhiomkar: nit: ...true if an element contains...

packages/mdc-drawer/README.md
```diff
@@ -437,55 +437,83 @@ document.body.addEventListener('MDCDrawer:closed', () => {
 });
 ```
 
-## `MDCDrawer` Properties and Methods
+<!-- docgen-tsdoc-replacer:start -->
```
abhiomkar: Let's make this obvious by mentioning that this is autogenerated and do not edit.


```html
<!-- docgen-tsdoc-replacer:start **DO NOT EDIT, This section is automatically generated** -->
```

packages/mdc-drawer/component.ts
moog16: Yeah I was thinking we could make it a list (originally i had it that way). 

Also the `@event` tag is being eaten by documentalist...so I had to create this fake tag. What if we use `@fires`?

scripts/documentation/tsdoc.ts
moog16: Ok I will look into Nunjucks

moog16:
@abhiomkar ready for re-review

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,333 @@
...
+    });
+  }
+
+  /**
+   * Generates JSON from source files TypeScript documentation.
+   * This contains all the esmodule classes (ie. foundations, adapters, components) in JSON format.
+   * @returns Promise<{}>
+   */
+  generateJSONFromFiles() {
+    return new Promise((resolve) => {
```
abhiomkar: You can use async/await here.

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,333 @@
...
+      return;
+    }
+    const markdownObject: ModuleMarkdown = {
+      events: this.getDocumentationForModule(esmodule).events,
+      methods: this.getDocumentationForMethods(esmodule),
+      moduleName: esmodule,
+      properties: this.getDocumentationProperties(esmodule),
+      readmeDirectoryPath,
+    };
+    this.addToMarkdownBuffer(readmeDirectoryPath, markdownObject);
```
abhiomkar: It is hard to reason about when buffer is used (side-effect), can we make this method return generated object which can be written into the README instead?

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  /** Sorts modules by the following rules:
+   * 1. Components
+   * 2. Adapters
+   * 3. Foundations
+   * 4. Alphabetized
+   * @param a 1st module of comparison
+   * @param b 2nd module of comparison
+   */
+  private sortByModuleType(a: ModuleMarkdown, b: ModuleMarkdown): number {
```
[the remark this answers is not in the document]
moog16: Do you mean insert into the buffer in this order?

packages/mdc-drawer/component.ts
abhiomkar: `@fires` doesn't seem to be valid annotation either. is `@fires` supported in documentalist?

Let's keep `@events` for now.

moog16: ok sounds good - @fires is supported https://jsdoc.app/tags-fires.html

scripts/documentation/tsdoc.ts
moog16: updated to remove the markdownBuffer object from the class

moog16:
@abhiomkar ready for rereview - might still need to fix the sorting thing. I'm not sure if we can avoid sorting.

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,336 @@
...
+   */
+  async generateJSONFromFiles() {
+    let _docs;
+    await new Documentalist()
+      .use(/\.ts$/, new TypescriptPlugin({
+        excludePaths: ['node_modules'],
+        includeDeclarations: true,
+      }))
+      .documentGlobs('packages/**/*')
+      .then((docs) => _docs = docs)
```
abhiomkar: await will return resolved value (`docs`), you wouldn't need this.

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,336 @@
...
+   */
+  async generateMarkdownFileFromBuffer(markdownBuffer: {[s: string]: ModuleMarkdown[]}) {
+    for (const readmeDirectoryPath in markdownBuffer) {
+      /**
+       * This currently only has been tested on mdc-drawer.
+       * TODO: remove this if condition once all READMEs are generated
+       */
+      const allowList = [
+        'mdc-drawer',
+        // 'mdc-textfield',
```
abhiomkar: Remove commented code and move this list to file level constant.

packages/mdc-drawer/README.md
```diff
@@ -437,55 +437,80 @@ document.body.addEventListener('MDCDrawer:closed', () => {
...
+`isOpening() => boolean` | Returns true if the drawer is animating open.
+`open() => void` | Opens the drawer from the closed state.
+
+
+
+<!-- docgen-tsdoc-replacer:end -->
+
+## Usage within Web Frameworks
+
+If you are using a JavaScript framework, such as React or Angular, you can create a Navigation Drawer for your framework. Depending on your needs, you can use the _Simple Approach: Wrapping MDC Web Vanilla Components_, or the _Advanced Approach: Using Foundations and Adapters_. Please follow the instructions [here](../../docs/integrating-into-frameworks.md).
```
abhiomkar: Can we move this to handlebar template and replace `...create a Navigation Drawer for your framework...` with `...create this component for your framework...` to not break the order.

This section comes after component API table and before adapter / foundation table.

moog16:
@abhiomkar ready for rereview. I'm actually going to use this as the Feature Branch, so I will not be merging this.

moog16:
Spoke offline - since this isn't a breaking change we can merge this into master

scripts/documentation/api-markdown-table-template.hbs
```diff
@@ -0,0 +1,36 @@
...
+
+{{/if}}
+{{#if module.events.length}}
+#### Events
+{{#each module.events}}
+- {{documentation}}
+{{/each}}
+
+{{/if}}
+{{#if @first}}
```
abhiomkar: Instead of flattening the array can we have 2 arrays? One for component and another for framework specific modules.

moog16: Hmm you mean sort by adapter/foundation and component?

abhiomkar: the template data assumes that first item in modules array is always `component.ts` module. It would be nice if we've separate entry to reference component APIs and Framework specific APIs, so that the order in the array doesn't matter. WDYT?

moog16: yeah i agree it is fragile...let me look into that.

moog16:
@abhiomkar ready for re-review!

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,356 @@
...
+    const readmeMarkdownPath = `./packages/${readmeDirectoryPath}/${README_FILE}`;
+    const readmeMd = await readFile(readmeMarkdownPath, 'utf8');
+    const componentModules = markdownBuffer[readmeDirectoryPath].component
+      .filter((module) => module.methods.length || module.properties.length || module.events.length)
+      .sort(this.sortByModuleType);
+    const nonComponentModules = markdownBuffer[readmeDirectoryPath].nonComponent
+      .filter((module) => module.methods.length || module.properties.length || module.events.length)
+      .sort(this.sortByModuleType);
+    const apiMarkdownTable =
+      this.templateFunction({modules: componentModules, showFrameworkUsage: true}) +
```
abhiomkar: wouldn't it be easier if template function takes all template data it needs at once?

```ts
this.templateFunction({
  componentModules,
  nonComponentModules,
});
```

moog16: Well it is the same template, we'd create another template (partial) to combine these together. I don't think it matters either way works. What would you like?

abhiomkar: I think having a single template as a whole makes it easier to reformat later. The current implementation we are not fully leveraging the use of templating engine, we concat two strings generated separately in the code instead.

moog16: Ok sounds good - i'll make the change.

moog16: Updated!

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,361 @@
...
+    this.setupTemplates();
+  }
+
+  /**
+   * Generates JSON from source files TypeScript documentation.
+   * This contains all the esmodule classes (ie. foundations, adapters, components) in JSON format.
+   * @returns Promise<{}>
+   */
+  async generateJSONFromFiles() {
+    if (!this.templateFunction) console.log('AHHHHHH MEOW')
```
abhiomkar: Remove debug lines.

scripts/documentation/tsdoc.ts
```diff
@@ -0,0 +1,361 @@
...
+    };
+  }
+
+  /**
+   * This is higher level documentation from the class.
+   * Currently this should only include events documentation.
+   * @returns documentation of the esmodule.
+   * @param esmodule module name (ie. MDCSelectIconFoundation)
+   */
+  getDocumentationForModule(esmodule: string): ModuleDocumentation {
```
abhiomkar: Can we merge this into `getDocumentationForEvents`?

scripts/documentation/tsdoc.ts
moog16: Embarrassing! lol

scripts/documentation/tsdoc.ts
moog16: I was thinking if we ever wanted another tag off of the module, that it would be easier to split like this. But I can combine this into one for now, since there isn't anything we need.

PR #4876, opened 2019-07-04
feat(snackbar): add feature targeting for styles

crisbeto:
Adds support for feature targeting to the `mdc-snackbar` styles.

Relates to #4227.


crisbeto:
Here's a comparison between the before and after styles. A couple of lines had to be moved around so that they'd fit into the mixins.

[before.css.txt](https://github.com/material-components/material-components-web/files/3360464/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/3360465/after.css.txt)

packages/mdc-snackbar/_mixins.scss
```diff
@@ -20,73 +20,269 @@
...
 @import "./variables";
 
-@mixin mdc-snackbar-fill-color($color) {
+@mixin mdc-snackbar-core-styles($query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+  $feat-animation: mdc-feature-create-target($query, animation);
+
+  // postcss-bem-linter: define snackbar
+  .mdc-snackbar {
+    @include mdc-snackbar-z-index($mdc-snackbar-z-index, $query);
```
abhiomkar: Let's name all `$query` params in this file when including mixins. 

```scss
@include mdc-snackbar-z-index($mdc-snackbar-z-index, $query: $query);
```

packages/mdc-snackbar/_mixins.scss
```diff
@@ -20,73 +20,269 @@
 // THE SOFTWARE.
 //
 
+@import "@material/feature-targeting/functions";
```
abhiomkar: Update mdc-snackbar/package.json with @material/feature-targeting dependency.

crisbeto:
Thank you for the feedback @abhiomkar, it should all be addressed now. Can you take another look?

PR #4877, opened 2019-07-06
feat(drawer): add feature targeting for styles

crisbeto:
Adds support for feature targeting to the `mdc-drawer` styles.

Relates to #4227.


crisbeto:
Here's the output before and after these changes. A few lines got moved around, but it shouldn't make a difference.

[before.css.txt](https://github.com/material-components/material-components-web/files/3364634/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/3364635/after.css.txt)

packages/mdc-drawer/dismissible/mdc-drawer-dismissible.scss
```diff
@@ -20,24 +20,5 @@
...
-    display: flex;
-  }
-}
-
-.mdc-drawer-app-content {
-  @include mdc-rtl-reflexive-box(margin, left, 0);
-
-  position: relative;
-}
+@include mdc-drawer--dismissible-core-styles;
```
abhiomkar: We do not usually follow BEM for mixin names. this can be renamed to `mdc-drawer-dismissible-core-style`.

Same for drawer modal mixin name.

packages/mdc-drawer/_mixins.scss
```diff
@@ -20,29 +20,267 @@
 // THE SOFTWARE.
 //
 
+@import "@material/feature-targeting/functions";
```
abhiomkar: Please update package.json with feature targeting dependency.

crisbeto:
Reworked based on the feedback @abhiomkar.

PR #4878, opened 2019-07-08
fix(chips): Ignore selection events in chip set

patrickrodee:
Emit selection events from the chip but allow them to be ignored in the chip set when triggered by the chip set.

BREAKING CHANGE: `MDCChipAdapter#notifySelection` has been updated as has the `MDCChipSelectionEventDetail`


packages/mdc-chips/chip/component.ts
```diff
@@ -192,8 +192,8 @@ export class MDCChip extends MDCComponent<MDCChipFoundation> implements MDCRippl
         this.emit<MDCChipRemovalEventDetail>(
           strings.REMOVAL_EVENT, {chipId: this.id, root: this.root_}, true /* shouldBubble */);
       },
-      notifySelection: (selected) => this.emit<MDCChipSelectionEventDetail>(
-          strings.SELECTION_EVENT, {chipId: this.id, selected}, true /* shouldBubble */),
+      notifySelection: (selected, shouldHandle) => this.emit<MDCChipSelectionEventDetail>(
```
moog16: Since you use the word `shouldIgnore` everywhere else, I think `shouldHandle` --> `shouldIgnore`

patrickrodee: ah you're right! thanks

packages/mdc-chips/README.md
```diff
@@ -365,7 +365,7 @@ Method Signature | Description
 `removeClassFromLeadingIcon(className: string) => void` | Removes a class from the leading icon element
 `eventTargetHasClass(target: EventTarget, className: string) => boolean` | Returns true if target has className, false otherwise
 `notifyInteraction() => void` | Notifies the Chip Set that the chip has been interacted with\*
-`notifySelection(selected) => void` | Notifies the Chip Set that the chip has been selected or deselected\*\*
+`notifySelection(selected: boolean, shouldIgnore: boolean) => void` | Notifies the Chip Set that the chip has been selected or deselected\*\*
```
abhiomkar: Let's document second boolean parameter. same for `handleChipSelection`.

packages/mdc-chips/chip/foundation.ts
```diff
@@ -334,9 +334,7 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
       this.adapter_.setPrimaryActionAttr(strings.ARIA_CHECKED, 'false');
     }
 
-    if (shouldNotify) {
-      this.adapter_.notifySelection(selected);
-    }
+    this.adapter_.notifySelection(selected, shouldIgnore);
```
abhiomkar: Instead of introducing a boolean flag to decide whether to emit an event, can we modify foundation to notify only when needed? I don't see why we needed this.

moog16: It was this way, however I think @patrickrodee had issues with this design, correct?

patrickrodee: That's correct. We have existing cases in Google that require the selection event to be emitted when the selection changes.

abhiomkar: I might be lacking some context here. At least not of enough context provided in this PR about new boolean option.

Here is my understanding, `setSelectedFromChipSet` method is invoked by chip set (parent). When this method invoked by chip set, the chip doesn't need to notify the selection event to avoid infinite loop.

My suggestion was that, `setSelected_` private method can be split into two separate private methods, one for presentational only changes (`setSelected_`) and another for notifying event (`notifySelection_`). If we do this, `setSelected` calls `setSelected_` & `notifySelection_` where `setSelectedFromChipSet` just calls `setSelected_` (presentational only changes).

WDYT?

patrickrodee: I've update the functionality. The chip set is now conditionally able to block events from bubbling and also label events as ignorable.

patrickrodee: Also updated to use the suggestion above.

packages/mdc-chips/README.md
patrickrodee: Done

patrickrodee:
@abhiomkar ready for re-review

packages/mdc-chips/README.md
```diff
@@ -409,7 +409,7 @@ Method Signature | Description
 --- | ---
 `isSelected() => boolean` | Returns true if the chip is selected
 `setSelected(selected: boolean) => void` | Sets the chip's selected state
-`setSelectedFromChipSet(selected: boolean) => void` | Sets the chip's selected state (called from the chip set)
+`setSelectedFromChipSet(selected: boolean, shouldNotify: boolean) => void` | Sets the chip's selected state (called from the chip set) to the `selected` param. Will emit a selection event if called with `shouldNotify` set to `true`
```
abhiomkar: Can `shouldNotify` be same as param name in `notifySelection` (ie., `chipSetShouldIgnore`)? Boolean flag is reversed in both these methods and it's confusing.

Or Can we rename all these booleans to `isUserSelected` or `selectOnAction`? This boolean will be set to true only when it is user selected. WDYT?

packages/mdc-chips/chip-set/foundation.ts
```diff
@@ -69,19 +69,11 @@ export class MDCChipSetFoundation extends MDCFoundation<MDCChipSetAdapter> {
    * Selects the chip with the given id. Deselects all other chips if the chip set is of the choice variant.
    */
   select(chipId: string) {
```
abhiomkar: Since this is not used anywhere outside foundation can we remove this select and rename selectOnInit to select? it makes it ambiguous to have separate select APIs.

packages/mdc-chips/chip-set/foundation.ts
```diff
@@ -69,19 +69,11 @@ export class MDCChipSetFoundation extends MDCFoundation<MDCChipSetAdapter> {
...
-    if (this.adapter_.hasClass(cssClasses.CHOICE) && this.selectedChipIds_.length > 0) {
-      const previouslySelectedChip = this.selectedChipIds_[0];
-      const previouslySelectedIndex = this.adapter_.getIndexOfChipById(previouslySelectedChip);
-      this.selectedChipIds_.length = 0;
-      this.adapter_.selectChipAtIndex(previouslySelectedIndex, false);
-    }
-    this.selectedChipIds_.push(chipId);
-    const index = this.adapter_.getIndexOfChipById(chipId);
-    this.adapter_.selectChipAtIndex(index, true);
+  selectOnInit(chipId: string) {
```
abhiomkar: Add TSDoc.

packages/mdc-chips/chip-set/foundation.ts
```diff
@@ -194,7 +191,7 @@ export class MDCChipSetFoundation extends MDCFoundation<MDCChipSetAdapter> {
     if (index >= 0) {
       this.selectedChipIds_.splice(index, 1);
       const chipIndex = this.adapter_.getIndexOfChipById(chipId);
-      this.adapter_.selectChipAtIndex(chipIndex, false);
+      this.adapter_.selectChipAtIndex(chipIndex, false, true);
```
abhiomkar: Document these boolean values, It makes it easier to discover what these values are mapped to.

Example:

```ts
foo(/** barParam */ true);
```

packages/mdc-chips/chip-set/foundation.ts
```diff
@@ -217,6 +214,22 @@ export class MDCChipSetFoundation extends MDCFoundation<MDCChipSetAdapter> {
...
+
+  private select_(chipId: string, shouldNotify: boolean) {
+    if (this.selectedChipIds_.indexOf(chipId) >= 0) {
+      return;
+    }
+
+    if (this.adapter_.hasClass(cssClasses.CHOICE) && this.selectedChipIds_.length > 0) {
+      const previouslySelectedChip = this.selectedChipIds_[0];
+      const previouslySelectedIndex = this.adapter_.getIndexOfChipById(previouslySelectedChip);
+      this.selectedChipIds_.length = 0;
```
abhiomkar: Can we set this to empty array instead to make this explicit?

packages/mdc-chips/README.md
patrickrodee: I'll go with the first option. Thanks!

patrickrodee: The `shouldNotify` is for clients. When `setSelectedFromChipSet` is called, the event is ignored by the chip set. I'll update the param to make it clear it's for client notification.

packages/mdc-chips/chip-set/foundation.ts
patrickrodee: Superseded by above comment.

packages/mdc-chips/chip-set/foundation.ts
patrickrodee: Done

packages/mdc-chips/README.md
abhiomkar: We still use different terminology. It would be great if we can align the terminology for this boolean flag across component. (`chipSetShouldIgnore`, `shouldNotify`, `chipSetShouldIgnore` & `shouldNotifyClients`).

patrickrodee: Cool cool. I updated this so only `shouldNotifyClients` is used.

PR #4880, opened 2019-07-08
feat(ripple): Add support for ripple target to mixins.

joyzhong:
This is to support moving the ripple target to an inner element for components with increased touch target (button, chips, etc.)


package-lock.json
```diff
@@ -10342,21 +10342,21 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+          "resolved": false,
```
abhiomkar: Revert package-lock.json file changes.

packages/mdc-ripple/_mixins.scss
```diff
@@ -98,10 +98,12 @@
   }
 }
 
-@mixin mdc-ripple-surface($query: mdc-feature-all()) {
+@mixin mdc-ripple-surface($query: mdc-feature-all(), $ripple-target-opt: null) {
```
abhiomkar: nit: `$ripple-target-opt` -> `$ripple-target`

Or `$ripple-target-selector`?

packages/mdc-ripple/_mixins.scss
```diff
@@ -98,10 +98,12 @@
...
   $feat-animation: mdc-feature-create-target($query, animation);
   $feat-structure: mdc-feature-create-target($query, structure);
 
+  $ripple-target: if($ripple-target-opt != null, " #{$ripple-target-opt}", "");
```
abhiomkar: Can we reverse this condition? :)

#Yoda

packages/mdc-ripple/_mixins.scss
```diff
@@ -140,13 +142,13 @@
 
   // Common styles for upgraded surfaces (some of these depend on custom properties set via JS or other mixins)
 
-  &.mdc-ripple-upgraded::before {
+  &.mdc-ripple-upgraded#{$ripple-target}::before {
```
abhiomkar: This seem like `.mdc-ripple-upgraded` and ripple-target are at the same level without the space between them. To make this explicit we can use nested block and remove space prefix from `$ripple-target`.

See this [gist from SassMeister](https://gist.github.com/abhiomkar/efc3e1131be96a6ed43632986726dce0) for example. `$ripple-target` is set to `&` when it is null to default it to root.

packages/mdc-ripple/_mixins.scss
```diff
@@ -389,14 +423,36 @@
...
-  @include mdc-states-base-color($color, $query);
-  @include mdc-states-hover-opacity(mdc-states-opacity($color, hover) + $opacity-modifier, $query);
+  @include mdc-ripple-target-selector($ripple-target-opt) {
+    @include mdc-states-base-color($color, $query);
+  }
+
+  @include mdc-states-hover-opacity(
+    mdc-states-opacity($color, hover) + $opacity-modifier,
+    $query,
+    $ripple-target-opt);
```
abhiomkar: We need to name $query and $ripple-target for future proofing. 

```scss
@include mdc-states-hover-opacity(
  mdc-states-opacity($color, hover) + $opacity-modifier,
    $query: $query,
    $ripple-target: $ripple-target);
```

This way the argument order doesn't matter.

Same for across this file.

packages/mdc-ripple/_mixins.scss
joyzhong: Love it! Thanks for the great suggestion.

packages/mdc-ripple/_mixins.scss
```diff
@@ -201,38 +214,51 @@
...
   // Focus overrides hover by reusing the ::before pseudo-element.
   // :focus-within generally works on non-MS browsers and matches when a *child* of the element has focus.
   // It is useful for cases where a component has a focusable element within the root node, e.g. text field,
   // but undesirable in general in case of nested stateful components.
   // We use a modifier class for JS-enabled surfaces to support all use cases in all browsers.
   $cssOnlyFocusSelector: if(
     $has-nested-focusable-element,
-    "&:not(.mdc-ripple-upgraded):focus::before, &:not(.mdc-ripple-upgraded):focus-within::before",
-    "&:not(.mdc-ripple-upgraded):focus::before"
+    "&:not(.mdc-ripple-upgraded):focus#{$ripple-target-selector}::before," +
```
abhiomkar: We can refactor this to use nested block selectors here too. :) 

Example: https://www.sassmeister.com/gist/58a8f62bec24b96f4c13f1a5f34a4d5b

packages/mdc-ripple/_mixins.scss
```diff
@@ -98,10 +98,12 @@
   }
 }
 
-@mixin mdc-ripple-surface($query: mdc-feature-all()) {
+@mixin mdc-ripple-surface($query: mdc-feature-all(), $ripple-target: null) {
```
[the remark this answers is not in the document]
joyzhong: Yes, great observation! Done.

packages/mdc-ripple/_mixins.scss
```diff
@@ -201,75 +210,96 @@
...
   // :focus-within generally works on non-MS browsers and matches when a *child* of the element has focus.
   // It is useful for cases where a component has a focusable element within the root node, e.g. text field,
   // but undesirable in general in case of nested stateful components.
   // We use a modifier class for JS-enabled surfaces to support all use cases in all browsers.
-  $cssOnlyFocusSelector: if(
-    $has-nested-focusable-element,
-    "&:not(.mdc-ripple-upgraded):focus::before, &:not(.mdc-ripple-upgraded):focus-within::before",
-    "&:not(.mdc-ripple-upgraded):focus::before"
-  );
+  @if $has-nested-focusable-element {
```
abhiomkar: This looks lot better! 🎉 

Thanks for making this more readable.

packages/mdc-ripple/_mixins.scss
```diff
@@ -286,55 +316,73 @@
...
+      $query: $query,
+      $ripple-target: $ripple-target);
   }
 }
 
 // Simple mixin for selected states which automatically selects opacity values based on whether the ink color is
 // light or dark.
-@mixin mdc-states-selected($color, $has-nested-focusable-element: false, $query: mdc-feature-all()) {
+@mixin mdc-states-selected(
+  $color, $has-nested-focusable-element: false, $query: mdc-feature-all(), $ripple-target: "&") {
```
abhiomkar: Can we break this into separate lines?

packages/mdc-ripple/_mixins.scss
joyzhong: Thanks for the great suggestion!

PR #4888, opened 2019-07-10
feat(data-table): Added .destroy() method to data-table component

abhiomkar:
#### Description

* Added `.destroy()` method to data table component.
* Remove overriding touch dimension of row checkbox.
* Removed testing with mocks in data table component test.

Refs #4591

### DIFF
--- a/packages/mdc-data-table/component.ts
+++ b/packages/mdc-data-table/component.ts
@@ -22,6 +22,7 @@
  */
 
 import {MDCComponent} from '@material/base/component';
+import {SpecificEventListener} from '@material/base/types';
 import {MDCCheckbox, MDCCheckboxFactory} from '@material/checkbox/component';
 import {closest} from '@material/dom/ponyfill';
 import {MDCDataTableAdapter} from './adapter';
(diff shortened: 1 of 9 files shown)


packages/mdc-data-table/component.ts
```diff
@@ -80,6 +88,14 @@ export class MDCDataTable extends MDCComponent<MDCDataTableFoundation> {
     this.foundation_.setSelectedRowIds(rowIds);
   }
 
+  destroy() {
+    this.headerRow_.removeEventListener('change', this.handleHeaderRowCheckboxChange_);
+    this.content_.removeEventListener('change', this.handleRowCheckboxChange_);
+
+    this.headerRowCheckbox_.destroy();
```
moog16: I think you need an `if (this.headerRowCheckbox_)` check

test/screenshot/spec/mdc-data-table/classes/baseline-checkbox.html
```diff
@@ -97,7 +97,7 @@
             </tr>
             <tr data-row-id="u2" class="mdc-data-table__row mdc-data-table__row--selected" aria-selected="true">
               <td class="mdc-data-table__cell mdc-data-table__cell--checkbox">
-                <div class="mdc-checkbox mdc-data-table__row-checkbox">
+                <div class="mdc-checkbox mdc-data-table__row-checkbox mdc-checkbox--selected">
```
moog16: nit: this row is "Offline". Is the Online/Offline & checkbox supposed to be related?

test/unit/mdc-data-table/mdc-data-table.test.js
```diff
@@ -279,76 +255,13 @@ test('adapter#setAttributeAtRowIndex', () => {
...
-    component,
-    adapter,
-  } = setupTestWithMocks();
-
-  const rows = [].slice.call(root.querySelectorAll(strings.ROW_SELECTOR));
-  td.when(mockFoundation.getRows()).thenReturn(rows);
-
-  adapter.registerRowCheckboxes();
-
-  td.verify(mockCheckboxFactory(td.matchers.isA(Element)), {times: 3});
```
moog16: isn't this test still valid with the exception of the component.destroy() call?

packages/mdc-data-table/component.ts
abhiomkar: As per current implementation `headerRowCheckbox_` is non-nullable and guaranteed to be initialized in `initialSyncWithDOM` method.

test/screenshot/spec/mdc-data-table/classes/baseline-checkbox.html
abhiomkar: They're not related. Checkbox selects the table row.

See this for example: https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2019/06/04/16_18_48_988/spec/mdc-data-table/classes/baseline-checkbox.html?utm_source=golden_json

test/unit/mdc-data-table/mdc-data-table.test.js
abhiomkar: It is a valid test but mocking isn't helping much. But, I discovered more bugs without mocking compared to with mocking.

PR #4898, opened 2019-07-12
feat(linear-progress): add feature targeting for styles

crisbeto:
Adds support for feature targeting to the mdc-linear-progress styles.

Relates to #4227.


packages/mdc-linear-progress/_mixins.scss
```diff
@@ -18,28 +18,252 @@
...
+      }
+    }
+
+    &__secondary-bar {
+      @include mdc-feature-targets($feat-structure) {
+        visibility: hidden;
+      }
+    }
+
+    @include mdc-linear-progress-indeterminate_($query: $query);
```
crisbeto: I had to break these out because stylelint was complaining about too much nesting.

crisbeto:
Here's a before/after comparison. A couple of lines got swapped, but they shouldn't matter.

[before.css.txt](https://github.com/material-components/material-components-web/files/3387772/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/3387773/after.css.txt)

Issue #4722, opened 2019-05-15
Export ES6 modules from dist directory


moog16:
related https://github.com/material-components/material-components-web/issues/4492#issuecomment-492226652

Feature request to generate ES6 modules from out TS files from within the /dist directory. This will help with GCC consumption.

rarous:
I'm sorry, I was not clear. This is not related to `/dist` but to the rest of the code in the package. Now it is mix of ES6 module system and classes transpiled to ES5 code. This ES5 code is the issue. Consumer already needs some traspiler to consume ES6 modules, so there is no need to transpile classes to ES5 code, because this is already not consumable by any ES5 interpreter. 

For GCC compatibility it will help to have ES6 classes there.

rarous:
After some investigation updating script `buils:esmodules` in `package.json` with `--target es6` should do the trick:

```
 "build:esmodules": "tsc --project ./tsconfig.json --target es6 --module esnext --importHelpers",
```

moog16:
@rarous I don't think you need to transpile our files in `/dist`. I tried this in my own HTML/CSS setup using checkbox, and just including `@material/checkbox/dist/mdc.checkbox.js`. I also ran use the [`es-check`](https://github.com/dollarshaveclub/es-check) lib on the ES5 file. The transpile ES5 files are UMD, _not_ ES6 so they should work in most browsers including IE11. 

I'm fine with including ES6 modules with the stated `build:esmodules...` script you created, but I don't think your other statement is valid. Unless I am still misunderstanding something?

rarous:
I'm not talking about `/dist` at all. It is bundle.

I'm interested in ES modules like `@material/checkbox/component.js`. Now they are mix of ES next modules (`import .. from`,  `export ...`) with ES5 prototypes. There is AFAIK no need to not having ES6 classes there.

moog16:
I forgot we also have those files (sorry). And I see what you're saying. This will be a breaking change, but it doesn't make sense to have them in ES5 since these files still need to go through transpilation.

I think we're on the same page now :)

rarous:
True :)

It was unnecessary Breaking Change introduced in v1.0

kfranqueiro:
https://github.com/material-components/material-components-web/issues/4222 explains why MDC Web provides ES6 modules with ES5 syntax.

TL;DR:

* Many build toolchains default to _not_ transpiling anything under `node_modules`. Providing ES6 modules with ES6 code requires additional configuration in these environments.
[...]

PR #4899, opened 2019-07-12
feat(infrastructure): add es6 module support

moog16:
fixes https://github.com/material-components/material-components-web/issues/4722

This now builds ES6 with ESModule import/export syntax. These files will be found within each package under the `packages/mdc-foo/es6` directory.

@abhiomkar I think we may want to add some verification check like we have with `scripts/verify-pkg-main.js`. If you want me to add to this let me know and I can modify it.


rarous:
@moog16 thank you for working on this. If I'm reading correctly your changes, there is still missing one thing to work it correctly and it is [`es2015` entry in `package.json`](https://2ality.com/2017/04/setting-up-multi-platform-packages.html#es2015%3A-es6%2C-esm) pointing to ES2015 module.

moog16:
@rarous is that actually a package.json entry? I can't find the webpack/rollup documentation for `es2015` on either of their sites. This leads me to believe that this isn't a real package.json property. We currently have `module`, however this is already taken.

One of the main objectives of this PR is to not break anyone but add functionality.

rarous:
@moog16 Most of the tooling has this configurable. But it should be in `package.json`. My suggestion is to use `es2015` or to use `module` and introduce `browser` for current `module` entry.

moog16:
Can you provide some kind of documentation besides Dr Axel's blog? I respect him and his work, but I would something more official?

rarous:
@moog16 There really no standard for this, best resource I can get is this https://github.com/stereobooster/package.json#package-bundlers.

abhiomkar:
Closing this as obsolete. 

If we're going to introduce ES6 syntax in modules it would the only syntax. Replace existing module which is transpiled to ES5.

PR #4922, opened 2019-07-18
fix(list): added #adapter.listItemAtIndexHasClass to prevent user state change to disabled items

### LINKED ISSUE #4696: MDCList: Disabled list item with disabled checkbox throws JS error
PR #4568 introduced a bug where a list with a disabled list item with a disabled checkbox throws an error when the list is initialized.

2.1.0

Affects all browsers.

### What OS are you using?

macOS 10.14.4

https://codepen.io/anon/pen/OYMZWY

The checkboxes state is set regardless of its enabled/disabled state.

Before this PR and in MDC version 1.1.1, the disabled checkboxes state would not be set and no error was thrown. This was better than in v2.0.0 but not quite what I would expect.

A disabled checkbox always has a checked/unchecked state and is only disabled for user interaction—it should always be set upon initialization.

MDCLists `listElements` now returns all list items including disabled ones and sets the `checked` attribute of its containing **enabled** checkbox input element. If the checkbox is **disabled**, it is not found by the css selector and the error `TypeError: null is not an object (evaluating 'toggleEl.checked = isChecked')` is thrown.

moog16:
fixes #4696

BREAKING CHANGE: New adapter method `listItemAtIndexHasClass`


moog16:
There are 2 different paths of selecting a listitem (w/ checkbox/radio/none) - `programatic` and 
 via the `user`. This PR reflects the idea that

1. all programatic changes should allow for the list item (w/ checkbox/radio) to be toggled
2. all user interactions (click, keydown) should *NOT* allow list item (w/checkbox/radio) to be toggled.

Pre-PR, MDC-List didn't allow any disabled checkboxes or radio buttons to be selected.

packages/mdc-list/foundation.ts
```diff
@@ -444,7 +445,15 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
...
 
+  /**
+   * 1) Toggles checkbox, 2) toggles radio, 3) sets selectedIndex
+   * of list item(s) on **user interaction**. User interaction should not toggle
+   * list item(s) when disabled.
+   */
   private setSelectedIndexOnAction_(index: number, toggleCheckbox = true) {
+    if (this.adapter_.isListItemDisabled(index)) {
+      return;
+    }
```
abhiomkar: nit: empty-line below this.

packages/mdc-list/foundation.ts
```diff
@@ -444,7 +445,15 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
     return index >= 0 && index < listSize;
   }
 
+  /**
+   * 1) Toggles checkbox, 2) toggles radio, 3) sets selectedIndex
```
abhiomkar: _Sets selected index on user action, toggles checkbox / radio based on `toggleCheckbox` value._

packages/mdc-list/adapter.ts
```diff
@@ -75,6 +75,12 @@ export interface MDCListAdapter {
    */
   isRootFocused(): boolean;
 
+  /**
```
abhiomkar: You also need to sync this to internal and implement this adapter before merging it.

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -484,3 +484,19 @@ test('adapter#isRootFocused returns true if list root is on focus', () => {
   assert.isTrue(component.getDefaultFoundation().adapter_.isRootFocused());
   document.body.removeChild(root);
 });
+
+test('adapter#isListItemDisabled returns true list item has disabled class', () => {
+  const {root, component} = setupTest();
+  root.querySelectorAll(`.${cssClasses.LIST_ITEM_CLASS}`)[0]
+    .classList.add(cssClasses.LIST_ITEM_DISABLED_CLASS);
+  document.body.appendChild(root);
+  assert.isTrue(component.getDefaultFoundation().adapter_.isListItemDisabled(0));
```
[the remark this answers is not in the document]
moog16: Are you saying without calling setupTest()?

packages/mdc-list/constants.ts
```diff
@@ -45,6 +45,8 @@ const strings = {
     .${cssClasses.LIST_ITEM_CLASS} button:not(:disabled),
     .${cssClasses.LIST_ITEM_CLASS} a
   `,
+  ENABLED_CHECKBOX_RADIO_SELECTOR: 'input[type="checkbox"], input[type="radio"]',
```
abhiomkar: I think `ENABLED_` prefix should be added to above `CHECKBOX_RADIO_SELECTOR` & `CHECKBOX_SELECTOR` instead since these selectors also selects disabled controls.

allan-chen: Agreed

packages/mdc-list/constants.ts
```diff
@@ -40,19 +40,21 @@ const strings = {
...
+  ENABLED_CHECKBOX_RADIO_SELECTOR: 'input[type="checkbox"]:not(:disabled), input[type="radio"]:not(:disabled)',
+  ENABLED_CHECKBOX_SELECTOR: 'input[type="checkbox"]:not(:disabled)',
   FOCUSABLE_CHILD_ELEMENTS: `
     .${cssClasses.LIST_ITEM_CLASS} button:not(:disabled),
     .${cssClasses.LIST_ITEM_CLASS} a,
     .${cssClasses.LIST_ITEM_CLASS} input[type="radio"]:not(:disabled),
     .${cssClasses.LIST_ITEM_CLASS} input[type="checkbox"]:not(:disabled)
   `,
-  RADIO_SELECTOR: 'input[type="radio"]:not(:disabled)',
+  RADIO_SELECTOR: 'input[type="radio"]',
```
allan-chen: This began as an effort for better naming consistency but actually fixes a bug where if the first radio element of a list is disabled, the remaining radio list breaks. This was due to the following lines in `layout()` of component not realizing a disabled radio at index 0 also makes the whole list a radio list, as a result of the constant above.  

```
 ...
 else if (this.adapter_.hasRadioAtIndex(0)) {
      this.isRadioList_ = true;
 }
```

packages/mdc-list/constants.ts
```diff
@@ -40,19 +40,21 @@ const strings = {
...
-  CHECKBOX_RADIO_SELECTOR: 'input[type="checkbox"]:not(:disabled), input[type="radio"]:not(:disabled)',
-  CHECKBOX_SELECTOR: 'input[type="checkbox"]:not(:disabled)',
+  CHECKBOX_RADIO_SELECTOR: 'input[type="checkbox"], input[type="radio"]',
+  CHECKBOX_SELECTOR: 'input[type="checkbox"]',
   CHILD_ELEMENTS_TO_TOGGLE_TABINDEX: `
     .${cssClasses.LIST_ITEM_CLASS} button:not(:disabled),
     .${cssClasses.LIST_ITEM_CLASS} a
   `,
+  ENABLED_CHECKBOX_RADIO_SELECTOR: 'input[type="checkbox"]:not(:disabled), input[type="radio"]:not(:disabled)',
+  ENABLED_CHECKBOX_SELECTOR: 'input[type="checkbox"]:not(:disabled)',
```
allan-chen: These two ENABLED_* constants are not used anywhere at this point. They are essentially renamed versions of the old `CHECKBOX_RADIO_SELECTOR` and `CHECKBOX_SELECTOR` before this PR, which were too restrictive in terms of what constituted a valid checkbox/radio to be toggled programmatically. Per Matt's philosophy that all programmatic changes to disabled checkboxes are allowed, I think we can remove this restriction and its associated constants?

abhiomkar: We can remove if unused.

packages/mdc-list/adapter.ts
```diff
@@ -75,6 +75,12 @@ export interface MDCListAdapter {
...
+   * @param index list item index.
+   * @return true if root element is disabled.
+   */
+  isListItemDisabled(index: number): boolean;
```
patrickrodee: Left a comment on the CL. Could we make this `listItemAtIndexHasClass(index: number, className: string): boolean` instead? It'll avoid having to add another method if we need to query for another conditional class on list items.

allan-chen: Yes, good idea. This reduces potential interface modifications in the future. Please see updated changes.

patrickrodee:
Make sure you add the breaking change notice to the merge request.

Issue #4881, opened 2019-07-09
[Density] Implement dense mixins for Button


abhiomkar:

abhiomkar:
Fixed in #4944

PR #4944, opened 2019-07-24
feat(button): Added density mixin to button

abhiomkar:
### Description

Density system to customize component visual height for adaptive layout.

* Added `mdc-button-density` mixin to Button component, which validates the sass parameter and supports tokens including default, comfortable and compact.
* Added `mdc-button-height` which is a supporting mixin which can be used to set custom component height.
* Separated baseline theme styles from core styles (Only for density & shape subsystem) to prevent CSS duplicates.

@jelbourn This is preparation change until we fully decide on UX design decision. This can be modified later based on the feedback. All components will have low-level sass mixin APIs (for example, to set component height / size) apart from density mixin for customization. 

#### Screenshot

Refs #4881 

BREAKING CHANGE: Removed `--dense` button variant, Use button density mixin instead.


abhiomkar:
@jelbourn Sent PR #4951 to add density utility package and addressed comments in this PR. PTAL.

PR #4951, opened 2019-07-24
feat(density): Added density utility component.

abhiomkar:
### Description

* Added `mdc-density-prop-value` function to return appropriate component height / size based on given density scale and component's density scale map. The `*-prop-value` name is given to be consistent with other utility functions used within MDC. For example, `mdc-theme-prop-value`.
* Component scale maps (e.g., `$mdc-button-density-scale-map`) use literals (e.g., `compact`) as keys to easily refer baseline / default component (max) height and compact (min) height to easily validate and throw error for unsupported component height. Scale map variable can be overridden by developers.

@jelbourn For review.

### DIFF
--- a/packages/mdc-density/README.md
+++ b/packages/mdc-density/README.md
@@ -1,0 +1,56 @@
+<!--docs:
+title: "Density"
+layout: detail
+section: components
+excerpt: "Density subsystem provides adaptive layout to components."
+path: /catalog/density/
+-->
+
+# Density
+
+Density subsystem provides adaptive layout to components.
+
+## Installation
+
+```
+npm install @material/density
+```
+
+## Basic Usage
+
+### Styles
+
+This package is used as utility for other components' density mixins. Mixin provided by this package is not intended to
+be consumed directly by developers, use component's density mixin instead.
+
+The styles for applying density to button component instance looks like this:
+
+```scss
+@import "@material/button/mixins";
+
+.my-custom-button {
+  @include mdc-button-density(compact);
+}
+```
+
+This would apply `compact` (highest density) to button component instance.
+
+> You would indirectly use the Density API through respective component's mixin which takes care of setting appropriate
+> component height.
+
+## Style Customization
+
+### Sass Variables
+
+Variable | Description
+--- | ---
+`$mdc-density-interval` | Density interval between each dense scale. This interval is used for numbered density scale to calculate dense height based on baseline component height.
+`$mdc-density-minimum-scale` | Minimum scale supported by density subsystem. This scale always maps to highest dense scale.
+`$mdc-density-maximum-scale` | Maximum scale supported by density subsystem. This scale always maps to lowest dense scale.
+`$mdc-density-supported-scales` | Supported density scale when density token or literal is used (For example, `compact`).
+
+### Sass Functions
+
+Function | Description
+--- | ---
+`mdc-density-prop-value($density-config, $density-scale, $property-name)` | Returns component property value based on given density config and density scale.
--- a/packages/mdc-density/_functions.scss
+++ b/packages/mdc-density/_functions.scss
@@ -1,0 +1,108 @@
+//
+// Copyright 2019 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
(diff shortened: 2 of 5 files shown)


packages/mdc-density/_functions.scss
```diff
@@ -0,0 +1,70 @@
...
+/// @example
+///   mdc-density-prop-value(-2, (
+///     default: 36px,
+///     comfortable: 32px,
+///     compact: 24px,
+///   ));
+///   // 28px
+///
+@function mdc-density-prop-value($density-scale, $density-scale-map) {
+  @if (type-of($density-scale) == "string" and
```
jelbourn: Would it make sense to also accept the number values? You could just map them instead of keeping a list
```scss
$mdc-density-supported-scales: (
  default: 0,
  comfortable: -1,
  compact: -2
)
```
Then just do everything with the numeric values

packages/mdc-density/_functions.scss
```diff
@@ -0,0 +1,70 @@
...
+  index($list: $mdc-density-supported-scales, $value: $density-scale) == null) {
+    @error "mdc-density: Supported density scales #{$mdc-density-supported-scales}, but received #{$density-scale}.";
+  }
+
+  $height: null;
```
jelbourn: This is similar to what I was proposing in #4944, but loses the level of indirection for the property being used. What I was suggesting was that each component has a config that captures more than one property, e.g. both `height` and `width`. So something like
```scss
$mdc-button-density (
  height: (
    0: $mdc-button-height, // Original height prior to density system
    -1: 32px,
    -3: $mdc-button-minimum-height
  ),
  width: (
    0: 10px,
    -1: 20px,
    -3: 30px
  )
  interval: 4px
)
```

Then you could apply the values like
```scss
.whatever-class {
  height:  mdc-density-prop-value($mdc-button-density, compact, height);
  width:  mdc-density-prop-value($mdc-button-density, compact, width);
}
```

I imagine different components would have different properties with specific sizes at different levels. Including the interval in the component config also keeps the door open to custom intervals for each component, even if it does end up being `4px` universally. 

Did you have any reasons why this system would be problematic?

abhiomkar: I like the idea of density config where we capture config for more than one property. I made changes in this PR accordingly as per your first point.

Regarding your 2nd point, We decided not to change density interval per component, it is always universal (as of now). This universal interval can be overridden by clients. We can always add interval to config if needed.

Also, related to your numbered scales as keys here are the issues that I see with this:

- We also need mapping for token density scale (e.g., `compact`)

So, other alternative config structure would be:

```scss
$density-config: (
  height: (
    default: $mdc-button-height,
    interval: 4px,
  ),
  scale-map: (
    compact: -3,
    comfortable: -1,
  ),
)
```

But this makes it hard for baseline or clients to configure these values relative to baseline. `24px` or `compact` is much more straight forward than numbered scales, IMO. It gets even more complication when we start introducing more properties.

- If we know the default property value we can programmatically determine other numbered scales based on universal interval, so I see that it is redundant.

packages/mdc-density/_functions.scss
abhiomkar: `$mdc-density-supported-scales` is universal. & It wouldn't be appropriate if we always map `compact` to `-2` which is different for each component.

Density function do accept numbered values in addition to string / token values. Are you saying that we would need to accept this as parameter to sass function? Please see my below comment.

packages/mdc-density/_functions.scss
jelbourn: My thinking is that the "compact", "comfortable", and "cozy" names are great for end-users picking from a menu, but that they're completely opaque to a developer. You the components have a fixed interval for density steps, it's much easier to visualize those intervals as "0, -1, -2" than remember which C-word is which. 

I had also been assuming that each of "compact", "comfortable", and "cozy" maps to a specific step universally. Is that not the case?

abhiomkar: "compact", "comfortable" values do not universally map to same scale. Also, it is not practical to apply same numbered scale "0, -1, -2" across components. For example, the minimum density for button is `-3` whereas the minimum density for checkbox is `-4`. Tokens helps map to appropriate density based on component which makes it easier for developer to use these without hardcoding the numbered scales. i.e., `compact` always refers to lowest supported density scales across components.

jelbourn: I suppose that works for me

packages/mdc-density/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+npm install @material/density
+```
+
+## Basic Usage
+
+### Styles
+
+This package is used as utility for other components' density mixins. Mixin provided by this package is not intended to
+consumer directly by developers, use component's density mixin instead.
```
joyzhong: s/to consumer/to be consumed/

PR #4953, opened 2019-07-26
refactor(chip): Add ripple target as an inner element.

aprigogin:


aprigogin:
BREAKING CHANGE: This changes the structure of the chip element by moving the ripple from the outer element to an inner .mdc-chip__ripple div element.

OLD

<button class="mdc-button">
  <span class="mdc-button__label">Hello World</span>
</button>


NEW

<button class="mdc-chip">
  <div class="mdc-chip__ripple"></div>
  <span class="mdc-chip__label">Hello World</span>
</button>

abhiomkar:
@aprigogin @joyzhong We also need to update the README file. It would be great if we include breaking change tag in commit message when we squash & merge. This will help changelog script to pick up the breaking change line automatically.

PR #4954, opened 2019-07-29
refactor(select): Move disabled logic into foundation

bonniezhou:
Simplify adapter and component by moving `setDisabled` logic into foundation.

BREAKING CHANGE: `MDCSelectAdapter#setDisabled` is removed, and `MDCSelectAdapter#setSelectedTextAttr` is added.

### DIFF
--- a/packages/mdc-select/component.ts
+++ b/packages/mdc-select/component.ts
@@ -186,6 +186,7 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
     // Initially sync floating label
     this.foundation_.handleChange(/* didChange */ false);
 
+    // Sets disabled state in foundation
     this.disabled = this.selectAnchor_.classList.contains(cssClasses.DISABLED);
   }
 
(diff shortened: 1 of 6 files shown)


packages/mdc-select/component.ts
```diff
@@ -186,9 +186,8 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
     // Initially sync floating label
     this.foundation_.handleChange(/* didChange */ false);
 
-    if (this.selectAnchor_.classList.contains(cssClasses.DISABLED)) {
-      this.disabled = true;
-    }
+    // Sets disabled state in foundation
+    this.disabled = this.selectAnchor_.classList.contains(cssClasses.DISABLED);
```
abhiomkar: Can we move disabled state in foundation? 

`get disabled()` static method can use this disabled state from foundation as source of truth.

packages/mdc-select/foundation.ts
```diff
@@ -96,6 +98,7 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
     this.helperText_ = foundationMap.helperText;
 
     this.menuItemValues_ = this.adapter_.getMenuItemValues();
+    this.disabled_ = false;
```
abhiomkar: We can initialize assign default value (`false`) during initialization of this variable.

PR #4956, opened 2019-07-30
refactor(chip): ripple inner element fixup

aprigogin:
Added remainder of the changes that were needed to move ripple element to be internal to the chip.

### DIFF
--- a/packages/mdc-chips/_mixins.scss
+++ b/packages/mdc-chips/_mixins.scss
@@ -262,9 +262,9 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
         position: absolute;
         width: 100%;
         height: 100%;
-        overflow: hidden;
         // Needed for ie11 to position ripple correctly at the top of container
         top: 0;
+        overflow: hidden;
       }
     }
   }
@@ -317,7 +317,7 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
     $rtl-reflexive,
     $query: $query);
 
-    #{$mdc-chip-ripple-target} {
+  #{$mdc-chip-ripple-target} {
     @include mdc-shape-radius(
       mdc-shape-resolve-percentage-radius($mdc-chip-height-default, $radius),
       $rtl-reflexive,
@@ -454,10 +454,10 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
 
   #{$mdc-chip-ripple-target} {
     @include mdc-feature-targets($feat-structure) {
-      border: $width solid transparent;
-      margin-left: $horizontal-margin;
-      margin-right: $horizontal-margin;
       top: $vertical-margin;
+      margin-right: $horizontal-margin;
+      margin-left: $horizontal-margin;
+      border: $width solid transparent;
     }
   }
 }
@@ -473,8 +473,8 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
   $margin: 0 - $padding;
 
   #{$mdc-chip-ripple-target} {
-    margin-left: $margin;
     margin-right: $margin;
+    margin-left: $margin;
   }
 }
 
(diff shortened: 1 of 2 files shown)


aprigogin:
@googlebot I consent

packages/mdc-chips/_mixins.scss
```diff
@@ -429,6 +448,18 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
...
     border-width: $width;
   }
+
+  $horizontal-margin: 0 - $horizontal-padding-value - $width;
+  $vertical-margin: 0 - $width;
+
+  #{$mdc-chip-ripple-target} {
+    @include mdc-feature-targets($feat-structure) {
+      top: $vertical-margin;
+      margin-right: $horizontal-margin;
```
abhiomkar: We do not need negative margin & transparent border if ripple is absolutely positioned inside chip.

We would need to explicitly set top & left position to `0` to start from top-left corner relative to its container.

aprigogin: good point... you do need transparent borders, but do not need negative margins, my mistake.

packages/mdc-chips/_mixins.scss
```diff
@@ -438,6 +465,15 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
     padding-right: $padding;
     padding-left: $padding;
   }
+
+  $margin: 0 - $padding;
+
+  #{$mdc-chip-ripple-target} {
+    @include mdc-feature-targets($feat-structure) {
+      margin-right: $margin;
```
abhiomkar: Can we remove negative margins from here too?

packages/mdc-chips/_mixins.scss
```diff
@@ -429,6 +448,14 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
     padding-left: $horizontal-padding-value;
     border-width: $width;
   }
+
+  #{$mdc-chip-ripple-target} {
+    @include mdc-feature-targets($feat-structure) {
+      top: -$width;
```
abhiomkar: top & left can be set to `0` which is the same thing.

packages/mdc-chips/_mixins.scss
```diff
@@ -438,6 +465,15 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
     padding-right: $padding;
     padding-left: $padding;
   }
+
+  $margin: 0 - $padding;
+
```
joyzhong: nit: -1 newline

packages/mdc-chips/_mixins.scss
joyzhong: I think this is necessary in order to compensate for the chip border. top/left of 0 will position the ripple element inside the border.

abhiomkar: Ah that's right. this is to offset transparent border-width added to ripple.

packages/mdc-chips/_mixins.scss
aprigogin: Argh!!! had it removed from corresponding CL, but missed it on the PR.

packages/mdc-chips/_mixins.scss
```diff
@@ -438,6 +465,15 @@ $mdc-chip-ripple-target: ".mdc-chip__ripple";
     padding-right: $padding;
     padding-left: $padding;
   }
+
+  $margin: 0 - $padding;
```
abhiomkar: nit: `$margin: -$padding;`

aprigogin: That whole chunk should have been removed

packages/mdc-chips/_mixins.scss
aprigogin: This whole chunk should not have been there. Removed from CL, but forgot to update PR.

PR #4958, opened 2019-07-31
refactor(select): Move adapter logic into foundation

bonniezhou:
- Proxy to foundation for `get/set required`
- Move `checkValidity` and `setValid` adapter logic to foundation
- Simplify `openMenu` in adapter
- Remove unused adapter method `getSelectedMenuItem`
- Simplify `toggleClassAtIndex` adapter method into `addClassAtIndex` and `removeClassAtIndex`

### DIFF
--- a/packages/mdc-select/component.ts
+++ b/packages/mdc-select/component.ts
@@ -281,18 +281,14 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
    * Sets the control to the required state.
    */
   set required(isRequired: boolean) {
-    if (isRequired) {
-      this.selectedText_.setAttribute('aria-required', isRequired.toString());
-    } else {
-      this.selectedText_.removeAttribute('aria-required');
-    }
+    this.foundation_.setRequired(isRequired);
   }
 
   /**
    * Returns whether the select is required.
    */
   get required(): boolean {
-    return this.selectedText_.getAttribute('aria-required') === 'true';
+    return this.foundation_.getRequired();
   }
 
   /**
@@ -348,29 +344,14 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
         }
         return '';
       },
-      checkValidity: () => {
-        const classList = this.root_.classList;
-        if (classList.contains(cssClasses.REQUIRED) && !classList.contains(cssClasses.DISABLED)) {
-          // See notes for required attribute under https://www.w3.org/TR/html52/sec-forms.html#the-select-element
-          // TL;DR: Invalid if no index is selected, or if the first index is selected and has an empty value.
-          return this.selectedIndex !== -1 && (this.selectedIndex !== 0 || Boolean(this.value));
-        }
-        return true;
-      },
-      setValid: (isValid: boolean) => {
-        this.selectedText_.setAttribute('aria-invalid', (!isValid).toString());
-        if (isValid) {
-          this.root_.classList.remove(cssClasses.INVALID);
-        } else {
-          this.root_.classList.add(cssClasses.INVALID);
-        }
-      },
       setSelectedText: (text: string) => this.selectedText_.textContent = text,
+      getSelectedTextAttr: (attr: string) => this.selectedText_.getAttribute(attr),
+      setSelectedTextAttr: (attr: string, value: string) => this.selectedText_.setAttribute(attr, value),
+      removeSelectedTextAttr: (attr: string) => this.selectedText_.removeAttribute(attr),
       openMenu: () => {
         if (!this.menu_.open) {
           this.menu_.open = true;
           this.isMenuOpen_ = true;
-          this.selectedText_.setAttribute('aria-expanded', 'true');
         }
       },
       closeMenu: () => {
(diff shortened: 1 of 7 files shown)


bonniezhou:
@abhiomkar this PR is incomplete but I wanted to get an initial review from you to see if these are the changes we want to make. 
In some cases, moving more adapter logic into the foundation may make it more complicated to wrap in other frameworks. For example, I'm guessing that `checkValidity` became an adapter method for wrapping in MDC React, because React [uses props instead of classes](https://github.com/material-components/material-components-web-react/blob/master/packages/select/index.tsx#L219).

packages/mdc-select/foundation.ts
```diff
@@ -307,11 +318,33 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+      return this.selectedIndex_ !== -1 && (this.selectedIndex_ !== 0 || Boolean(this.adapter_.getValue()));
+    }
+    return true;
+  }
+
+  setRequired(isRequired: boolean) {
+    if (isRequired) {
+      this.adapter_.setSelectedTextAttr('aria-required', isRequired.toString());
+    } else {
+      this.adapter_.removeSelectedTextAttr('aria-required');
```
abhiomkar: We can use `setSelectedTextAttr` here, setting attr value to `false`.

packages/mdc-select/component.ts
```diff
@@ -348,29 +344,14 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
...
-        }
-      },
       setSelectedText: (text: string) => this.selectedText_.textContent = text,
+      getSelectedTextAttr: (attr: string) => this.selectedText_.getAttribute(attr),
+      setSelectedTextAttr: (attr: string, value: string) => this.selectedText_.setAttribute(attr, value),
+      removeSelectedTextAttr: (attr: string) => this.selectedText_.removeAttribute(attr),
       openMenu: () => {
         if (!this.menu_.open) {
           this.menu_.open = true;
           this.isMenuOpen_ = true;
```
abhiomkar: `this.isMenuOpen_` - We may not need to store this state in component,  we can use `this.menu_.open` instead.

packages/mdc-select/component.ts
```diff
@@ -348,29 +344,14 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
...
-        } else {
-          this.root_.classList.add(cssClasses.INVALID);
-        }
-      },
       setSelectedText: (text: string) => this.selectedText_.textContent = text,
+      getSelectedTextAttr: (attr: string) => this.selectedText_.getAttribute(attr),
+      setSelectedTextAttr: (attr: string, value: string) => this.selectedText_.setAttribute(attr, value),
+      removeSelectedTextAttr: (attr: string) => this.selectedText_.removeAttribute(attr),
       openMenu: () => {
         if (!this.menu_.open) {
```
abhiomkar: Component should not check for correctness. For example, `if (!this.menu_.open) ...` it should be moved to foundation or it is not required.

packages/mdc-select/component.ts
```diff
@@ -399,17 +379,19 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
         const menuItem = this.menu_.items[index];
         return menuItem && menuItem.textContent ? menuItem.textContent : '';
       },
-      toggleClassAtIndex: (index: number, className: string, toggle: boolean) => {
+      addClassAtIndex: (index: number, className: string) => {
         const menuItem = this.menu_.items[index];
```
abhiomkar: Foundation should guarantee to send valid index, again component does not need to check for correctness of input to adapter. ie, `menu_.items[index]` always returns valid menu item.

packages/mdc-select/component.ts
```diff
@@ -348,29 +344,14 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
...
-      setValid: (isValid: boolean) => {
-        this.selectedText_.setAttribute('aria-invalid', (!isValid).toString());
-        if (isValid) {
-          this.root_.classList.remove(cssClasses.INVALID);
-        } else {
-          this.root_.classList.add(cssClasses.INVALID);
-        }
-      },
       setSelectedText: (text: string) => this.selectedText_.textContent = text,
+      getSelectedTextAttr: (attr: string) => this.selectedText_.getAttribute(attr),
```
abhiomkar: `getValue` need clean-up too. But we can do it in separate PR.

packages/mdc-select/component.ts
```diff
@@ -281,18 +281,14 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
    * Sets the control to the required state.
    */
   set required(isRequired: boolean) {
-    if (isRequired) {
```
abhiomkar: Focus management in `menuSetup_` going to be refactored in later PRs?

abhiomkar:
> @abhiomkar this PR is incomplete but I wanted to get an initial review from you to see if these are the changes we want to make.
> In some cases, moving more adapter logic into the foundation may make it more complicated to wrap in other frameworks. For example, I'm guessing that `checkValidity` became an adapter method for wrapping in MDC React, because React [uses props instead of classes](https://github.com/material-components/material-components-web-react/blob/master/packages/select/index.tsx#L219) [](/material-components/material-components-web-react/blob/HEAD@{2019-07-31T22:09:42Z}/packages/select/index.tsx#L219).

Good question! References to `props` can be delegated to foundation via adapter. (For example, adapter_.isDisabled) which you've done already.

PR #4959, opened 2019-08-01
 fix(top-app-bar): reopen short top-app-bar

### LINKED ISSUE #4928: TopAppBar short never reopens after initial scroll
### Steps to reproduce
1. Go to https://material.io/develop/web/components/top-app-bar/
2. Scroll down to the "short" variant of the top-app-bar
3. Scroll down, notice it shrinks
4. Scroll back to top, notice it does not restore to its original full size

### Expected behavior
The top-app-bar should restore to its original size for the short variant whenever the page is scrolled back to the top

Noticed this while upgrading RMWC to 3.0.0. I don't believe this was an issue with 2.0.0

TroyTae:
fixes #4928

Short top-app-bar isn't reopen intentionally.
So I cannot sure that this is a bug...

And there is one suggestion!
How about use `translateY` instead of `top` and fix top/left to 0?
1. `translateY` has better performance than `top`
2. If top-app-bar has parent element that has padding, top-app-bar is moved to top suddenly when starting scroll. Because top-app-bar is fixed element that isn't set top/left.

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -305,6 +305,9 @@ $mdc-button-ripple-target: ".mdc-button__ripple";
     line-height: inherit;
     user-select: none;
     -webkit-appearance: none;
+    // Even though `visible` is the default, IE 11 computes the property as
+    // `hidden` in some cases, unless it's explicitly defined here.
+    overflow: visible;
     vertical-align: middle;
   }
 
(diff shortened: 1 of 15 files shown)


TroyTae:
If you agree above changes, I'll fix unit test :)

TroyTae:
@abhiomkar 
I fix all of tests but `Semantic Pull Request` is pending continuously.

PR #4965, opened 2019-08-06
fix(snackbar): close snackbar when keydown esc

TroyTae:
Now `closeOnEscape_` property isn't working.
Because event target is the root element of snackbar.
I think we need to add listener to window in this case :)

### DIFF
--- a/packages/mdc-snackbar/component.ts
+++ b/packages/mdc-snackbar/component.ts
@@ -150,7 +150,6 @@ export class MDCSnackbar extends MDCComponent<MDCSnackbarFoundation> {
 
   private deregisterWindowKeyDownHandler_(handler: SpecificEventListener<'keydown'>) {
     window.removeEventListener('keydown', handler);
-    this.unlisten('keydown', handler);
   }
 
   private registerSurfaceClickHandler_(handler: SpecificEventListener<'click'>) {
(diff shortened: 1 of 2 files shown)


TroyTae:
@abhiomkar 
So then what is the role of this handler(https://github.com/material-components/material-components-web/blob/master/packages/mdc-snackbar/foundation.ts#L156-L161)? 🤔

Or this specification is deprecated??
스크린샷 2019-08-07 오전 10 29 02

abhiomkar:
> @abhiomkar
> So then what is the role of this handler([/packages/mdc-snackbar/foundation.ts@`master`#L156-L161](https://github.com/material-components/material-components-web/blob/master/packages/mdc-snackbar/foundation.ts#L156-L161) [](/material-components/material-components-web/blob/HEAD@{2019-08-07T01:33:53Z}/packages/mdc-snackbar/foundation.ts#L156-L161))? 🤔
> 
> Or this specification is deprecated??

Descriptions says that ESC is pressed only when snackbar is in focus state.

bonniezhou:
Working as intended. User must navigate to the snackbar before pressing ESC to close it (when the `closeOnEscape` property is `true`)

TroyTae:
Ah I see, thank you!

PR #4971, opened 2019-08-08
feat(divider): Add MDC Divider

### LINKED ISSUE #2662: Add new `mdc-divider` component
https://material.io/go/design-dividers

Richienb:
Closes #2662.

### DIFF
--- a/packages/mdc-divider/_mixins.scss
+++ b/packages/mdc-divider/_mixins.scss
@@ -1,3 +1,25 @@
+//
+// Copyright 2019 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


packages/mdc-divider/_mixins.scss
```diff
@@ -0,0 +1,7 @@
+@mixin mdc-divider-core-styles {
+  @include mdc-divider-line-color(rgba(0, 0, 0, .117));
+}
+
+@mixin mdc-divider-line-color($color) {
```
abhiomkar: Since the divider style property is specific to context, I think it would make sense to expose divider color in `mdc-theme/_variables.scss`. We don't need separate package for this, IMO.

abhiomkar: Re-opening this since it is still unresolved. I do not see a value in having a separate component for divider, it can reside in mdc-theme instead.

Richienb: Should I move the class into mdc-theme or just the rgba code?

porfirioribeiro: > Since the divider style property is specific to context, I think it would make sense to expose divider color in `mdc-theme/_variables.scss`. We don't need separate package for this, IMO.

I find myself always looking for this variable when i need to implement some kind of divider.
Sometimes end up using `$mdc-list-divider-color-on-light-bg` but it seems it should be on theme base. So it could be used everywhere

packages/mdc-divider/_mixins.scss
```diff
@@ -0,0 +1,27 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+@import "./variables";
```
porfirioribeiro: So where does this file exists with the variable declaration?

zakdances:
Really excited for this to be merged! There's nothing quite like a good divider - powerful, utilitarian, elegant.

packages/mdc-divider/_mixins.scss
abhiomkar: @Richienb Yes, that is correct. This should be stored in mdc-theme package as mentioned in my previous comment.

Please create a new PR to make this change. Thanks!

aeromac:
what happened to this?

PR #4978, opened 2019-08-08
fix(slider): Visual bug when slider value is displayed as "-0"

### LINKED ISSUE #5379: [MDCSlider] Discrete Slider schows value "-0" by default
## Bug report
Discrete Slider schows value "-0" by default. Must only "0"

### Steps to reproduce
1. Go to 'https://material-components.github.io/material-components-web-catalog/#/component/slider'
2. Click on 'Discrete Slider'
3. Drag slider to the left to zero
4. Then pull just a little to the right
4. See error: value "-0"

### Actual behavior
value "-0"

### Expected behavior
value only "0"

### Screenshots
[Slider bug](http://dmitry-bret.de/screenshots/bug-slider.jpg "Slider bug")

### Your Environment:

| MDC Web          | 4.0.0
| Browser          | Chrome
| Operating System | Windows 10

Tester798:
If min or max slider value is set to "0" then there can be visual bug when edge value is displayed as "-0".

Here is how it looks:
image

You can check it here: https://codepen.io/Tester456/pen/GVGJEq

Closes #5379


Tester798:
@googlebot I signed it!

packages/mdc-slider/foundation.ts
```diff
@@ -463,9 +463,9 @@ export class MDCSliderFoundation extends MDCFoundation<MDCSliderAdapter> {
     if (this.step_ && !valueSetToBoundary) {
       value = this.quantize_(value);
     }
-    if (value < min) {
+    if (value <= min) {
```
abhiomkar: Can you please add / update unit tests?

Thanks!

Tester798: Thanks for the reply.
Ok, I added two tests which simulate the bug.

While I was looking into it I found that seems like the bug is caused by `toLocaleString` function call in next line:
https://github.com/material-components/material-components-web/blob/cefcad79644350f05bc1c92efc07f9f1349d0f97/packages/mdc-slider/component.ts#L120
Here is simple demo: https://codepen.io/Tester456/pen/JjPmPmo
So maybe this bug can be fixed there instead of where I tried to fix it initially.

Tester798: Hmm, seems like original fix is only fixing edge cases when "0" value is on the left or right side of the slider, but it doesn't fix the bug when "0" value is located in the middle:
added 3rd slider https://codepen.io/Tester456/pen/oNvawqw

So maybe fixing the bug in the mentioned above line will be a good idea.
Also because of this, probably two tests can be replaced with just one which checks value "0" in the middle.

Tester798: Replaced two tests with only one and changed the fix.

packages/mdc-slider/component.ts
```diff
@@ -117,7 +117,13 @@ export class MDCSlider extends MDCComponent<MDCSliderFoundation> {
         this.thumbContainer_.style.setProperty(propertyName, value);
       },
       setTrackStyleProperty: (propertyName, value) => this.track_.style.setProperty(propertyName, value),
-      setMarkerValue: (value) => this.pinValueMarker_.innerText = value.toLocaleString(),
+      setMarkerValue: (value) => {
+        if (value === 0) {
+          this.pinValueMarker_.innerText = '0';
```
asyncLiz: I'm not sure if this is the appropriate solution. The problem is that setMarkerValue is being provided `-0` by the foundation, and it shouldn't be doing that.

The root problem is the [calculation of `pctComplete`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts#L330), which can be negative if the pageX is negative.

I think a better solution might be to solve the problem there. It's not possible for the slider to be "negative percentage" filled.

```diff
     let pctComplete = xPos / this.rect_.width;
+    if (pctComplete < 0) {
+      pctComplete = 0;
+    }
```

By fixing it in the foundation layer, we won't have to fix similar `-0` bugs in other implementations, such as MWC.

Can you make that change and resolve the conflicting file?

Tester798: Thanks for your reply, updated pull request with resolved conflicting file.

Tried your variant but looks like the bug is still there: the test is failing with `-0` slider value.
Here is the page similar to the one in the test: https://codepen.io/Tester456/pen/povOvEJ

asyncLiz: Ok, I understand. It's possible that "0" on the slider isn't 0%.

I still think that a foundation-layer solution is the appropriate approach. The component shouldn't have to worry about what its getting from the foundation or need to do any data massaging.

You could take advantage of JS's `||` pipe and the fact that both `0` and `-0` evaluate as falsy to coerce `-0` cases to `0`. [setValue_()](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts#L435) would be an appropriate place to fix it.

```diff
+    value = value || 0; // coerce -0 to 0
     this.value_ = value;
     this.adapter_.setAttribute(strings.ARIA_VALUENOW, String(this.value_));
```

Tester798: Yes, I agree with you, that it should be fixed foundation, I just wasn't able to find the place where is should be fixed. Looks like it's working fine now.
Thanks a lot for the fix!

abhiomkar:
Unit tests does not work on forked repos because of Sauce labs credentials.

asyncLiz:
Ah got it, thanks!

asyncLiz:
[cl/290121534](http://cl/290121534)

asyncLiz:
@Tester798 sorry for one more update, but we've been migrating a bunch of our test files. Can you rebase master onto your branch one last time and move the code your wrote in mdc-slider.test.js to [test/component.ts](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/test/component.test.ts)?

You'll also need to change `test(` to `it(` and `assert.equal(pinValueMarker.innerHTML, '0');` to `expect(pinValueMarker.innerHTML).toEqual('0')`

Tester798:
Ok, done.

packages/mdc-slider/test/component.test.ts
```diff
@@ -490,4 +490,29 @@ describe('MDCSlider', () => {
     document.body.removeChild(wrapper);
   });
 
+  it('correct value on pin value marker when value is 0', () => {
+    const root = getFixture();
+    Object.assign(root.style, {
```
asyncLiz: Can you change this to directly set properties instead of using Object.assign? That will fix IE11 issues and [match what we're doing elsewhere](https://github.com/material-components/material-components-web/blob/56a8bfef1b118817060891efd4d050e8d4b5a383/packages/mdc-slider/test/component.test.ts#L108-L111).

```ts
root.style.position = 'absolute';
root.style.left = '0';
root.style.top = '0';
root.style.width = '400px';
```

packages/mdc-slider/test/component.test.ts
```diff
@@ -490,4 +490,29 @@ describe('MDCSlider', () => {
...
+    root.setAttribute('aria-valuemax', '100');
+    const component = new MDCSlider(root);
+    document.body.appendChild(root);
+    component.layout();
+
+    root.dispatchEvent(new MouseEvent('mousedown', {view: window, bubbles: true, cancelable: true, clientX: 199}));
+    root.dispatchEvent(new MouseEvent('mousemove', {view: window, bubbles: true, cancelable: true, clientX: 200}));
+    root.dispatchEvent(new MouseEvent('mouseup', {view: window, bubbles: true, cancelable: true, clientX: 200}));
+
+    const pinValueMarker = root.querySelector(strings.PIN_VALUE_MARKER_SELECTOR);
```
asyncLiz: Cast this `as HTMLElement` to fix a null check compilation error.

```ts
const pinValueMarker = root.querySelector(strings.PIN_VALUE_MARKER_SELECTOR) as HTMLElement;
```

Tester798: Done.

packages/mdc-slider/test/component.test.ts
Tester798: Done.

asyncLiz:
Hi @Tester798, hopefully this will be the last change I ask for!

tl;dr, the test isn't passing in IE11 because of `new MouseEvent()`.

I think we should move this test to the foundation pointer-events test, since the problem is fixed in the foundation layer itself, not the component layer. This also gives us access to our mock handlers for events and we can test all the event types (mouse, touch, and pointer).

Can you remove the code from component.test.ts and add the following to [foundation-pointer-events.test.ts](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/test/foundation-pointer-events.test.ts)?
```ts
    it(`on body ${
          moveEvt} discrete slider avoids "-0" pin value marker`,
      () => {
        const {foundation, mockAdapter, rootHandlers, bodyHandlers} =
            setupTest();

        mockAdapter.computeBoundingRect.and.returnValue({left: 0, width: 100});
        mockAdapter.hasClass.withArgs(cssClasses.IS_DISCRETE)
            .and.returnValue(true);
        foundation.init();
        foundation.setMin(-50);
        foundation.setMax(50);
        jasmine.clock().tick(1);

        // Simulate barely moving the pointer to the left of "0"
        rootHandlers[downEvt](clientXObj(50));
        bodyHandlers[moveEvt]({
          preventDefault: () => {},
          ...clientXObj(49.9),
        });
        jasmine.clock().tick(1);

        const stringValue = mockAdapter.setMarkerValue.calls.mostRecent().args[0].toLocaleString();
        expect(stringValue).toEqual('0');
      });
```

Tester798:
Done. Also added 1 more space indent to match code above.

PR #4982, opened 2019-08-09
fix: Use head instead of body to detect edge pseudo var bug

aomarks:
If this script is invoked in the head, then the body won't exist yet, which causes the appendChild line to throw.

I confirmed that the bug is still detected in Edge 15.15063.

### DIFF
--- a/packages/mdc-ripple/util.ts
+++ b/packages/mdc-ripple/util.ts
@@ -34,7 +34,9 @@ function detectEdgePseudoVarBug(windowObj: Window): boolean {
   const document = windowObj.document;
   const node = document.createElement('div');
   node.className = 'mdc-ripple-surface--test-edge-var-bug';
-  document.body.appendChild(node);
+  // Append to head instead of body because this script might be invoked in the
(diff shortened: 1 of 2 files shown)


packages/mdc-ripple/util.ts
```diff
@@ -34,7 +34,9 @@ function detectEdgePseudoVarBug(windowObj: Window): boolean {
   const document = windowObj.document;
   const node = document.createElement('div');
   node.className = 'mdc-ripple-surface--test-edge-var-bug';
-  document.body.appendChild(node);
+  // Append to head instead of body because this script might be invoked in the
+  // head, in which case the body doesn't exist yet. The probe works either way.
+  document.head.appendChild(node);
```
abhiomkar: Is this detection still required in latest version of Edge?

aomarks: Nope, the test returns false in Edge 18.

aomarks:
@abhiomkar Should be mergeable now. Fixed the failing test (needed to add `head` to the mock window), and added a semantic prefix to the PR title.

Issue #4984, opened 2019-08-12
[mdc-list] include all icons themes in mdc-list-item__meta to stylelint-disable selector-class-pattern


ItangSanjana:
## Bug report
It seems that currently only `material-icons` is supported by `mdc-list-item__meta`. Thus, the icons presented with other themes will not appear.

### Steps to reproduce
1. add `<i class="mdc-list-item__meta material-icons-outlined" aria-hidden="true">chevron_right</i>`
2. See the icon not showing.

### Actual behavior
For icons with themes other than default, icons do not appear.

### Expected behavior
Icons appear for all themes.

### Possible solution
Include these themes to the CSS rule:
```
material-icons-outlined
material-icons-two-tone
material-icons-round
material-icons-sharp
```

PR #4985, opened 2019-08-12
Fix list list-item__meta stylelint-disable

ItangSanjana:
Fix issue #4984

### DIFF
--- a/packages/mdc-list/_mixins.scss
+++ b/packages/mdc-list/_mixins.scss
@@ -113,7 +113,7 @@
 
   .mdc-list-item__meta {
     // stylelint-disable selector-class-pattern
-    &:not(.material-icons) {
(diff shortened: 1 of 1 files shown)


packages/mdc-list/_mixins.scss
```diff
@@ -113,7 +113,7 @@
 
   .mdc-list-item__meta {
     // stylelint-disable selector-class-pattern
-    &:not(.material-icons) {
+    &:not(.material-icons):not(.material-icons-outlined):not(.material-icons-round):not(.material-icons-two-tone):not(.material-icons-sharp) {
```
abhiomkar: This increases specificity, we should create a modifier for meta text and meta icon that applies specific styles.

Example: `.mdc-list-item__meta--text`

ItangSanjana: Yes, that's would be great.

abhiomkar:
Closing since inactive.

PR #4989, opened 2019-08-13
feat(button): use [disabled] CSS selector instead of :disabled

### LINKED ISSUE #1339: MDC-Web Button - Disabled Attribute Fails On Anchor Flavored Buttons
https://codepen.io/rhroyston/pen/xXOvOz

andrewseguin:
This adds support for disabling buttons using the anchor tag. Currently, anchor tags are not "enabled" or "disabled" via the pseudo-selectors (`:enabled`, `:disabled`). Using the `[disabled]` attribute selector captures the disabled intent for both `button` and `a` elements.

This change will help usage in Angular Material since anchor buttons can be disabled, and it is currently non-trivial to override the styling since it actively applies the "enabled" styles rather than applying a specific set of "disabled" styles.

Fixes #1339

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -151,9 +151,7 @@
 @mixin mdc-button-container-fill-color($color, $query: mdc-feature-all()) {
   $feat-color: mdc-feature-create-target($query, color);
 
-  // :not(:disabled) is used to support link styled as button
-  // as link does not support :enabled style
-  &:not(:disabled) {
+  &:not([disabled]) {
     @include mdc-feature-targets($feat-color) {
       @include mdc-theme-prop(background-color, $color, $edgeOptOut: true);
     }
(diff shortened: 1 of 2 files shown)


andrewseguin:
Right, this is to help support the case in Angular Material where anchors using `mat-button` do have a `disabled` input/attribute. This maintains the same look-and-feel between `button` and `a` tags.

The negative aspect of this change seems to be unexpected styling from users who have (for some reason) a `disabled` attribute on their anchor tags outside of Angular Material. I'm not sure this would actually happen in the world.

Alternatively, this change could modify the selector to also match on an optional `mdc-button-disabled` class. Or the styles could be inverted so that if a button is disabled, a mixin is applied. That way we can also call this mixin from Angular Material

abhiomkar:
Separate mixin that sets disabled styles for any button SGTM. It is counter intuitive to use attribute selector for button that supports `:disabled` pseudo selector.

WDYT?

andrewseguin:
Sounds good, a separate mixin would be great. To support that, I suspect that the styling would need a bigger adjustment since color is applied when enabled rather than disabled styles applied when disabled.

PR #4990, opened 2019-08-13
feat(text-field): Added mixin to set custom height and center aligned inner elements

### LINKED ISSUE #4967: [Density] Center align inner elements of Text Field to support dynamic height

abhiomkar:
### Description

* Added `mdc-text-field-height` and `mdc-text-field-outlined-height` mixins to set custom height for default variant and outlined text field variant.
* Floating label
    * Vertically aligned to center, previously positioned `18px` from top for default variant.
* Notched outline
    * Added `mdc-notched-outline-floating-label-float-position-absolute` to float label in `px` unit which is based on the component height. Previously in percentage unit which is based on float label height.
    * Notch element now has transparent top border when label is center aligned to prevent jump on focused state.
    * Added `mdc-notched-outline-notched-reset` to reset border for labels where it is not center aligned.

Tested on IE11, iOS browser, Firefox & Chrome.

Text field screenshot with custom height

Fixes #4967 

BREAKING CHANGE: Removed sass variable in notched outline - `$mdc-notched-outline-transition-duration`.

### DIFF
--- a/packages/mdc-notched-outline/_variables.scss
+++ b/packages/mdc-notched-outline/_variables.scss
@@ -20,13 +20,21 @@
 // THE SOFTWARE.
 //
 
-$mdc-notched-outline-transition-duration: 150ms !default;
 // Keep this in sync with constants.numbers.MIN_LEADING_STROKE_EDGE_POSITION
 $mdc-notched-outline-min-leading-stroke-edge-position: 12px !default;
 // The gap between the stroke end and floating label
 // Keep this in sync with constants.numbers.NOTCH_GUTTER_SIZE
 $mdc-notched-outline-notch-gutter-size: 4px !default;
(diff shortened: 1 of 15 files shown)


packages/mdc-notched-outline/_variables.scss
```diff
@@ -30,3 +30,5 @@ $mdc-notched-outline-leading-width: 12px !default;
 $mdc-notched-outline-padding: 4px !default;
 // This variable keeps the before/after JS label centered in the notch when the font-size is changed.
 $mdc-notched-outline-label-adjust: 14% !default;
+$mdc-notched-outline-label-box-height: 18px !default;
+$mdc-notched-outline-label-adjust-absolute: 2.5px !default;
```
joyzhong: Add comment describing this - it's not clear from the var name.

PR #4996, opened 2019-08-14
feat(rtl): Added a flag to turn-off mdc-rtl CSS

abhiomkar:


test/screenshot/spec/mdc-rtl/variables/include.html
```diff
@@ -0,0 +1,82 @@
...
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'UA-118996389-2');
+    </script>
+  </head>
+
+  <body class="test-container">
+    <main class="test-viewport test-viewport--mobile" dir="rtl">
+      <p class="mdc-typography--caption">NOTE: Expected these component pixels to be broken when RTL is turned off</p>
```
joyzhong: nit: This wording would be more clearer to me: "NOTE: These components are expected to be broken since `$mdc-rtl-include` is false."

PR #4998, opened 2019-08-14
feat(snackbar): Add option for indefinite timeout

### LINKED ISSUE #153: Add indefinite snackbar duration
## Feature Request

Currently, snackbars time out after a set period of time, which fits with the _transient_ principle in the [material design guidelines](https://material.io/guidelines/components/snackbars-toasts.html#snackbars-toasts-usage). The android spec, however, allows the user to display a snackbar for an  [indefinite amount of time](https://developer.android.com/reference/android/support/design/widget/Snackbar.html#LENGTH_INDEFINITE). 

I think this feature could be useful to have in the web components. It could still be transient by default, but the user could optionally specify an indefinite length of time. It could be a separate property like:

```js
const snackbar_options = {
    indefinite: true
}
```

or another value for timeout, like:

```js
const snackbar_options = {
    timeout: null
}
```

cdonovan:
Resolves #153.  I was not sure if we should add a new discrete `indefinite` option or provide an "invalid" value to denote the indefinite case.  Currently, I have it set to look for -1 as a timeoutMs value, but if another approach is more desirable, please advise.


packages/mdc-snackbar/foundation.ts
```diff
@@ -88,12 +88,15 @@ export class MDCSnackbarFoundation extends MDCFoundation<MDCSnackbarAdapter> {
...
 
       this.animationTimer_ = setTimeout(() => {
+        const timeoutMs = this.getTimeoutMs();
         this.handleAnimationTimerEnd_();
         this.adapter_.notifyOpened();
-        this.autoDismissTimer_ = setTimeout(() => {
-          this.close(REASON_DISMISS);
-        }, this.getTimeoutMs());
-      }, numbers.SNACKBAR_ANIMATION_OPEN_TIME_MS);
+        if (timeoutMs !== -1) {
```
abhiomkar: Create a constant `INDETERMINATE` for this value.

packages/mdc-snackbar/foundation.ts
```diff
@@ -138,7 +141,7 @@ export class MDCSnackbarFoundation extends MDCFoundation<MDCSnackbarAdapter> {
     const minValue = numbers.MIN_AUTO_DISMISS_TIMEOUT_MS;
     const maxValue = numbers.MAX_AUTO_DISMISS_TIMEOUT_MS;
 
-    if (timeoutMs <= maxValue && timeoutMs >= minValue) {
+    if (timeoutMs === -1 || (timeoutMs <= maxValue && timeoutMs >= minValue)) {
       this.autoDismissTimeoutMs_ = timeoutMs;
     } else {
       throw new Error(`timeoutMs must be an integer in the range ${minValue}–${maxValue}, but got '${timeoutMs}'`);
```
abhiomkar: Update Error message accordingly, also update README file.

packages/mdc-snackbar/foundation.ts
cdonovan: Added `INDETERMINATE` as a const to the numbers enum and replaced existing usages.

packages/mdc-snackbar/foundation.ts
cdonovan: Updated the error message, along with the README to denote the new usage.

cdonovan:
Thanks for the feedback @abhiomkar!  I've addressed the noted issues and have pushed a few updates.  I wasn't sure if the subsequent commits required the same message format (since it seems the PRs are squashed), so apologies in advance if that is incorrect.

PR #5007, opened 2019-08-16
fix(top-app-bar): reopen short top-app-bar

### LINKED ISSUE #4928: TopAppBar short never reopens after initial scroll

TroyTae:
moved from #4995
fixes #4928

### DIFF
--- a/packages/mdc-top-app-bar/mdc-top-app-bar.scss
+++ b/packages/mdc-top-app-bar/mdc-top-app-bar.scss
@@ -34,10 +34,13 @@
 
   display: flex;
   position: fixed;
+  top: 0;
+  left: 0;
   flex-direction: column;
   justify-content: space-between;
   box-sizing: border-box;
   width: 100%;
+  will-change: transform;
   z-index: 4;
 
   &__row {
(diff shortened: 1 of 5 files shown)


azakus:
This will break the always-collapsed variant after the first scroll, as the state changes are:
```
init:
isCollapsed?: no
remove class?: no

scroll down:
isCollapsed?: yes
add class?: yes

scroll up to top:
isCollapsed?: yes
remove class?: yes
```

The foundation needs a setter to skip the scroll handling if the bar should always be collapsed.

TroyTae:
@azkkus
Thank you!
And it seems you already create another PR.
I’ll close this :)

azakus:
@TroyTae I think you should take the `top` => `translate` change from this PR and make a new PR. Probably still a good idea :)

TroyTae:
@azakus 
Cool!
I'll create a new PR, after your code is merged :)

Issue #4993, opened 2019-08-14
[Select][Foundation Delegation] Refactor for foundation delegation


abhiomkar:
- [x] Remove programmatically adding `WITH_LEADING_ICON` classes, update docs to recommend adding these classes in HTML instead.
- [x] Remove programmatically adding `aria-required` attr and `REQUIRED` class, update docs to recommend adding these in HTML instead.
- [x] Remove mutation observers from component and replace it with `setRequired` API that consumer can use to set required field which delegates to foundation.
- [x] Move `setAnchorElement` and `setAnchorCorner` to foundation from `menuSetup_`.
- [x] Split and delegate `getValue` adapter logic to foundation.
- [x] Move `isMenuOpen_` state logic to menu foundation / component that distinguishes state between `opening` and `opened`.
- [x] Move `handleMenuOpened_` & `handleMenuClosed_` callback logic to foundation.
- [x] Rename `setupMenu` to `init` to make it clear that this foundation API has to be called during initialization. `setAnchorElement` and `setAnchorCorner` can also be moved here as mentioned above.
- [x] Remove programmatically setting the selected value (i.e., `this.foundation_.setValue` in `initialSyncWithDOM`), update docs to recommend having selected item inside selected text element.
- [x] Remove menu item index validation in adapter methods and move it to foundation. `const menuItem = this.menu_.items[index]; if (menuItem) {...}`

See epic bug #4814 for more details explaining this refactor.

PR #5010, opened 2019-08-20
refactor(select): Move rest of menu logic into foundation

bonniezhou:
Fixes the rest of #4993.
- Rename `setupMenu` to `init` to make it clear that this foundation API has to be called during initialization
- Move `setAnchorElement` and `setAnchorCorner` to `init`
- Remove menu item index validation in adapter methods, the index is already validated in foundation in all cases
- Move `isMenuOpen_` state logic to menu foundation / component that distinguishes state between `opening` and `opened`
- Move `handleMenuOpened_` & `handleMenuClosed_` callback logic to foundation


packages/mdc-select/README.md
```diff
@@ -407,7 +411,7 @@ If you are using a JavaScript framework, such as React or Angular, you can creat
 | `isValid() => boolean` | Gets the valid state through the adapter's `checkValidity` API. |
 | `setRequired(isRequired: boolean) => void` | Sets the required state through the adapter. |
 | `getRequired() => boolean` | Gets the required state through the adapter. |
-| `setupMenu() => void` | Handles menu setup. |
+| `init() => void` | Handles menu setup. |
```
abhiomkar: nit: Update method description.

packages/mdc-select/adapter.ts
```diff
@@ -148,7 +170,7 @@ export interface MDCSelectAdapter {
   /**
    * Gets the text content of the menu item element at the given index.
    */
-  getMenuItemTextAtIndex(index: number): string;
+  getMenuItemTextAtIndex(index: number): string | null;
```
abhiomkar: `getMenuItemTextAtIndex` should always return `string` type.

packages/mdc-select/foundation.ts
```diff
@@ -111,9 +121,11 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
     const previouslySelectedIndex = this.selectedIndex_;
     this.selectedIndex_ = index;
 
-    this.adapter_.setSelectedText(
-        this.selectedIndex_ === numbers.UNSET_INDEX ?
-            '' : this.adapter_.getMenuItemTextAtIndex(this.selectedIndex_).trim());
+    if (this.selectedIndex_ === numbers.UNSET_INDEX || !this.adapter_.getMenuItemTextAtIndex(this.selectedIndex_)) {
```
abhiomkar: When'll this return falsy value?

packages/mdc-select/foundation.ts
```diff
@@ -111,9 +121,11 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+      this.adapter_.setSelectedText('');
+    } else {
+      this.adapter_.setSelectedText(this.adapter_.getMenuItemTextAtIndex(this.selectedIndex_)!.trim());
```
abhiomkar: nit: Save `this.adapter_.getMenuItemTextAtIndex(this.selectedIndex_)` return value to variable and reuse here and above.

packages/mdc-select/foundation.ts
```diff
@@ -185,10 +197,24 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
 
   handleMenuOpened() {
     this.adapter_.addClass(cssClasses.ACTIVATED);
+
+    if (this.adapter_.getMenuItemValues().length === 0) {
```
abhiomkar: Move this return statement before adding ACTIVATED class name.

packages/mdc-select/foundation.ts
```diff
@@ -185,10 +197,24 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
+    this.adapter_.focusMenuItemAtIndex(focusItemIndex);
   }
 
   handleMenuClosed() {
     this.adapter_.removeClass(cssClasses.ACTIVATED);
+    this.isMenuOpen_ = false;
+
+    this.adapter_.setSelectedTextAttr('aria-expanded', 'false');
+    if (!this.adapter_.isSelectedTextFocused()) {
+      this.handleBlur();
```
abhiomkar: We may not call handle* inside another handle* method in foundation. This might introduce problem in future if we add additional logic to calling handler method. I've seen this pattern in other places but this'll create issues in future.

Abstract required logic to separate method if required.

packages/mdc-select/component.ts
```diff
@@ -132,30 +130,8 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
     };
     this.handleKeydown_ = (evt) => this.foundation_.handleKeydown(evt);
     this.handleMenuSelected_ = (evtData) => this.selectedIndex = evtData.detail.index;
```
abhiomkar: Delegate this to foundation? Also, rename this to `handleMenuItemAction` to be consistent with menu component.

packages/mdc-select/foundation.ts
bonniezhou: If `index` is out of range. Alternatively we could add an adapter method `getNumberOfMenuItems` to make sure that it is in range?

packages/mdc-select/foundation.ts
bonniezhou: I should have asked this earlier, but do you know this is even here? Is there any case in which the menu closed but the selected text is not focused?

abhiomkar:
> Sorry for the delay, didin't realize this PR was still in review!
> 
> What do you mean that it doesn't seem to update select text?

No problem! 

Regarding the select text issue, open this [screenshot test page](https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2019/08/22/23_30_26_077/spec/mdc-select/classes/baseline.html) (this PR's snapshot) and select an option in 1st and 2nd example. The issue is that the selected text is not getting updated.

packages/mdc-select/foundation.ts
bonniezhou: I'm gonna keep it like this so it's executed after `this.selectedIndex_ === numbers.UNSET_INDEX` is confirmed false

packages/mdc-select/foundation.ts
abhiomkar: Adapter to get menu count sounds good. Foundation can throw error when index is not valid.

packages/mdc-select/foundation.ts
abhiomkar: I think this is handling a case where menu is closed without a selection (By clicking outside the select or TAB'ing out). It removes the focused state from the component.

abhiomkar:
@bonniezhou To resolve screenshot diffs, revert your golden.json file with the version of `origin/develop` branch after merging with this branch. Then re-run screenshot tests.

packages/mdc-select/foundation.ts
bonniezhou: Moved blur logic into private helper to be called in both handlers.

packages/mdc-select/adapter.ts
abhiomkar: I think we can typecast it with `... as string` operator in adapter implementation.

packages/mdc-select/foundation.ts
```diff
@@ -346,14 +373,36 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
...
       this.setValue(value);
     }
   }
+
+  /**
+   * Unfocuses the select component.
+   */
+  private blur_() {
+    this.adapter_.removeClass(cssClasses.FOCUSED);
+    this.handleChange(false);
```
abhiomkar: We also need to remove calls to `handleChange` foundation method. 

`handleChange()` currently seem to have a didChange boolean flag that is used to trigger notify event, instead handleChange method should always trigger an event on change.

My understanding is that didChange was required to notch the outline on initial DOM sync which can be abstracted out into `init` or `layout` methods.

Sorry for going back & forth on this, but this refactor puts select component implementation in much better state and makes it much easier to wrap.

bonniezhou: Good catch! I moved that logic out into its own method `updateLabel`, so `handleChange` no longer needs any parameters.

packages/mdc-select/component.ts
```diff
@@ -165,11 +141,11 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
...
     this.menu_!.listen(menuSurfaceConstants.strings.CLOSED_EVENT, this.handleMenuClosed_);
     this.menu_!.listen(menuSurfaceConstants.strings.OPENED_EVENT, this.handleMenuOpened_);
-    this.menu_!.listen(menuConstants.strings.SELECTED_EVENT, this.handleMenuSelected_);
-    this.foundation_.setupMenu();
+    this.menu_!.listen(menuConstants.strings.SELECTED_EVENT, this.handleMenuItemAction_);
+    this.foundation_.init();
 
     // Initially sync floating label
-    this.foundation_.handleChange(/* didChange */ false);
+    this.foundation_.updateLabel();
```
abhiomkar: Can we call this inside foundation's init method and make it a private method of foundation? This foundation method is never called except on init.

abhiomkar:
Also list all breaking changes in PR description and when you squash & merge. :)

We need to merge `feature/select` to `develop` at the same time when you sync this internally.

PR #5013, opened 2019-08-21
feat(button): support disabled design in <a> element

### LINKED ISSUE #5012: [Button] Support disabled design same with a element
## Feature Request
Now MCW is allowed to use `<a>` element in Button.
But `<a>` element cannot apply disabled attribute and disabled design isn't applied.

### Proposed solution
1. Add style rule `[href=""]`
2. Add style rule `[disabled]`

TroyTae:
closes #5012

I want to write just `[disabled]` or `[href=""]`.
But [&:not(:disabled)](https://github.com/material-components/material-components-web/blob/master/packages/mdc-button/_mixins.scss#L156) rule has higher priority, so I write like this...
If you have some great idea, please tell me :)

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -276,7 +276,8 @@
     }
   }
 
-  &:disabled {
+  &:disabled,
+  &[disabled][href=""] {
     @include mdc-feature-targets($feat-color) {
       @include mdc-theme-prop(background-color, transparent);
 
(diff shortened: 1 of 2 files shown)


TroyTae:
@abhiomkar 
I leave this comment link because I cannot reopen the issue.

TroyTae:
@abhiomkar
I leave this comment link because I cannot reopen the issue.
(#5012 (comment))

abhiomkar:
Please see #5012 for updates.

Thanks!

PR #5015, opened 2019-08-21
fix(button): Replace density sets with min & max scales

abhiomkar:


abhiomkar:
> Out of curiosity, what's the reasoning behind this change?

We're removing labeling these in MVP version of density. But we might add these back in future once we're confident about these density values in these sets.

PR #5041, opened 2019-08-29
fix(tabs): Fix tab img icon styling.

joyzhong:
Styles were accidentally changed in 
https://github.com/material-components/material-components-web/commit/c7efc10afea37f724c7c258e80dfd95a455f31d8#diff-d1669061bc1d2998fb197e25382d0f6c

### DIFF
--- a/packages/mdc-tab/_mixins.scss
+++ b/packages/mdc-tab/_mixins.scss
@@ -81,9 +81,11 @@
   }
 
   .mdc-tab__icon {
-    width: $mdc-tab-icon-size;
-    height: $mdc-tab-icon-size;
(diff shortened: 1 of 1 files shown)


packages/mdc-tab/_mixins.scss
```diff
@@ -59,7 +60,32 @@
 
   .mdc-tab__text-label,
   .mdc-tab__icon {
-    @include mdc-tab__content-child_($query);
+    @include mdc-feature-targets($feat-animation) {
```
abhiomkar: Move these styles to respective selector blocks below.

PR #5047, opened 2019-09-03
fix(menu): recompute index before marking selection

lasalvavida:
Fixed issues with changing menu contents in menus with selectable items. Marking selected items in a menu is deferred until the menu closes, so if the menu contents change before the callback is executed, the wrong item can be marked in the DOM.


packages/mdc-menu/foundation.ts
```diff
@@ -97,8 +97,10 @@ export class MDCMenuFoundation extends MDCFoundation<MDCMenuAdapter> {
 
     // Wait for the menu to close before adding/removing classes that affect styles.
     this.closeAnimationEndTimerId_ = setTimeout(() => {
-      if (this.adapter_.isSelectableItemAtIndex(index)) {
-        this.setSelectedIndex(index);
+      // Recompute the index in case the menu contents have changed.
```
abhiomkar: I think, the list foundation need to have a separate method (For example, `.layout()`) to recompute indexes which can be called on demand.

abhiomkar: Looking at your changes again, looks good to me! is it possible to add a unit test for this?

lasalvavida: Added

PR #5049, opened 2019-09-04
feat(list): Add setEnabled to foundation

EstebanG23:


packages/mdc-list/foundation.ts
```diff
@@ -304,6 +304,20 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
     return lastIndex;
   }
 
+  /**
+   * @param itemIndex Index of the list item
+   * @param isEnabled Sets the list item to enabled or disabled.
+   */
+  setEnabled(itemIndex: number, isEnabled: boolean): void {
+    if (isEnabled) {
```
abhiomkar: Can you also add validation for itemIndex?

See `setSelectedIndex` method for example.

test/unit/mdc-list/foundation.test.js
```diff
@@ -996,3 +996,35 @@ test('#getSelectedIndex should be in-sync with setter method for multi-select ch
   foundation.setSelectedIndex([0, 2, 3]);
   assert.deepEqual([0, 2, 3], foundation.getSelectedIndex());
 });
+
+test('#setEnabled should remove class name mdc-list-item--disabled', () => {
```
abhiomkar: Can you merge this test with `#setEnabled should set aria-disabled to false when enabled is true`.

Same for below test. I think we can add just 2 unit tests that checks for class name and attribute changes at the same time.

packages/mdc-list/foundation.ts
```diff
@@ -304,6 +304,20 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
     return lastIndex;
   }
 
+  /**
+   * @param itemIndex Index of the list item
+   * @param isEnabled Sets the list item to enabled or disabled.
+   */
+  setEnabled(itemIndex: number, isEnabled: boolean): void {
```
abhiomkar: Can you also add API to list component.ts to set this state?

packages/mdc-list/foundation.ts
EstebanG23: Done

test/unit/mdc-list/foundation.test.js
EstebanG23: Done!

packages/mdc-list/foundation.ts
EstebanG23: Done!

packages/mdc-list/component.ts
```diff
@@ -129,6 +129,10 @@ export class MDCList extends MDCComponent<MDCListFoundation> {
     }
   }
 
+  setEnabled(itemIndex: number, isEnabled: boolean) {
```
abhiomkar: Add TSDoc for method description & params.

packages/mdc-list/component.ts
```diff
@@ -129,6 +129,10 @@ export class MDCList extends MDCComponent<MDCListFoundation> {
...
+    this.foundation_.setEnabled(itemIndex, isEnabled);
```
abhiomkar: You also need tests for this. :) You wouldn't need to mock since you'll be testing with real DOM for component.

packages/mdc-list/component.ts
abhiomkar: Update README file for this method.

EstebanG23: Got it!

packages/mdc-list/component.ts
EstebanG23: sounds good! I followed the same testing format as the initializeListType method in the component to continue the pattern but if let me know if you think there is a better way to test it?

packages/mdc-list/component.ts
EstebanG23: Done!

test/unit/mdc-list/mdc-list.test.js
```diff
@@ -156,6 +156,12 @@ test('#initializeListType does not populate selectedIndex when no item is select
   td.verify(mockFoundation.setSelectedIndex(td.matchers.anything()), {times: 0});
 });
 
+test('#setEnabled when true adds disabled class and aria-disabled to true', () => {
+  const {component, mockFoundation} = setupTest();
+  component.setEnabled(1, true);
```
abhiomkar: did you mean false here? This list item is already enabled.

EstebanG23: You're right, I changed it to a better testing tittle.

EstebanG23:
I was the only author in the code, first couple commits did not have my exact email address but I can confirm that CLA's look good.

PR #5054, opened 2019-09-06
feat(menu): add setEnabled to allow dynamic enabling or disabling menu item

EstebanG23:


packages/mdc-menu/component.ts
```diff
@@ -156,6 +156,15 @@ export class MDCMenu extends MDCComponent<MDCMenuFoundation> {
     this.foundation_.setSelectedIndex(index);
   }
 
+  /**
+   * Sets the disabled state to isDisabled for the menu item at the given index.
+   * @param index Index of the menu item
+   * @param isDisabled The desired disabled state of the menu item.
+   */
+  setDisabled(index: number, isDisabled: boolean): void {
```
abhiomkar: Can we keep the same API name as MDC List to be consistent?

EstebanG23: Done!

PR #5064, opened 2019-09-10
fix(form-field): Fix radio RTL alignment bug.

joyzhong:
Also add form-field radio screenshot tests.
Fixes b/140750923 internally.


test/screenshot/spec/mdc-form-field/fixture.scss
```diff
@@ -0,0 +1,24 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+@import "../../../../packages/mdc-form-field/mixins";
+@import "../../../../packages/mdc-radio/mixins";
```
abhiomkar: is this required? this does not emit any CSS and we're not using any mixins from these packages.

PR #5070, opened 2019-09-11
feat(tab-bar): Add density mixin to tab-bar

### LINKED ISSUE #4884: [Density] Implement dense mixins for Tab Bar

abhiomkar:
#### Description

* Adds `mdc-tab-bar-density` and `mdc-tab-bar-stacked-density` mixins per each variant.
* Fixed the inner content layout to support dynamic height.
* Screenshot test with density mixins.

#### Screenshots

screenshot of tab with custom density
screenshot of tab stacked variant with custom density

Fixes #4884


packages/mdc-tab/_mixins.scss
```diff
@@ -266,20 +278,15 @@
...
-    }
-
-    .mdc-tab__icon {
-      padding-top: 12px;
+      justify-content: center;
     }
 
     .mdc-tab__text-label {
-      padding-bottom: 16px;
+      padding-top: 6px;
```
patrickrodee: These lines look like they have potential to introduce screenshot changes in Google.

patrickrodee: Possibly unlikely, though. I don't believe the stacked variant is in the GM spec.

packages/mdc-tab-bar/_variables.scss
```diff
@@ -0,0 +1,47 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+@import "@material/density/variables";
+@import "@material/tab/variables";
+
+$mdc-tab-bar-height: $mdc-tab-height !default;
+$mdc-tab-bar-minimum-height: 30px !default;
+$mdc-tab-bar-maximum-height: $mdc-tab-bar-height !default;
```
patrickrodee: Just to make sure, it looks like the maximum possible height for tabs is the default 48px. Is that correct?

abhiomkar: That's correct. 48px is the max height that is supported.

packages/mdc-tab/_mixins.scss
abhiomkar: Just ran tests. This is only affecting IE11 (Actually fixing the alignment on IE11 compared to previous screenshots).

PR #5103, opened 2019-09-17
chore: New GitHub Action for MDC Lint

abhiomkar:
Moved MDC Lint CI from Travis to shiny new GitHub Action. All other tests will eventually be moved to GitHub Action.


abhiomkar:
> Nice! Will this block submission like Travis if the lint fails?

Yes that is correct. PR will be blocked when either of the test fails.

PR #5112, opened 2019-09-19
fix(button): Adjust touch target size when density is applied

abhiomkar:
Resets styles related to touch target when density applied since dense component don't necessarily need additional touch target.

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -243,6 +243,31 @@ $mdc-button-ripple-target: ".mdc-button__ripple";
   );
 
   @include mdc-button-height($height, $query: $query);
+
+  @if $density-scale != 0 {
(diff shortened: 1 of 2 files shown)


packages/mdc-button/_mixins.scss
```diff
@@ -243,6 +243,25 @@ $mdc-button-ripple-target: ".mdc-button__ripple";
...
+
+  @if $density-scale != 0 {
+    @include mdc-button-touch-target-reset_;
+  }
+}
+
+///
+/// Resets touch target-related styles. This is called from the density mixin to
+/// automatically remove the increased touch target, since dense components
+/// don't have the same default a11y requirements.
```
joyzhong: Reminder to update this comment to reflect internal edits.

abhiomkar: This was updated :)

PR #5116, opened 2019-09-20
fix(chips): Reset touch target when chip density mixin is applied.

joyzhong:


test/scss/_feature-targeting-test.scss
```diff
@@ -90,6 +90,7 @@
     @include mdc-chip-trailing-icon-color(red, $query: $query);
     @include mdc-chip-trailing-icon-margin($query: $query);
     @include mdc-chip-trailing-icon-size(0, $query: $query);
+    @include mdc-chip-touch-target-reset_($query: $query);
```
abhiomkar: I think `mdc-chip-density(-1, $query: $query)` would cover this private mixin.

joyzhong: Good point :) Removed.

PR #5126, opened 2019-09-25
feat(switch): add ripple opacity customization mixins

allan-chen:
Allow customization of ripple opacity in MDC switch.

### DIFF
--- a/packages/mdc-switch/README.md
+++ b/packages/mdc-switch/README.md
@@ -132,6 +132,9 @@ Mixin | Description
 `mdc-switch-toggled-off-thumb-color($color)` | Sets color of the thumb when the switch is toggled off.
 `mdc-switch-toggled-on-ripple-color($color)` | Sets the color of the ripple surrounding the thumb when the switch is toggled on.
 `mdc-switch-toggled-off-ripple-color($color)` | Sets the color of the ripple surrounding the thumb when the switch is toggled off.
+`mdc-switch-ripple-hover-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when hovered.
+`mdc-switch-ripple-focus-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when focused.
+`mdc-switch-ripple-press-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when pressed.
 
 ## `MDCSwitch` Properties and Methods
 
--- a/packages/mdc-switch/README.md
+++ b/packages/mdc-switch/README.md
@@ -132,9 +132,7 @@ Mixin | Description
 `mdc-switch-toggled-off-thumb-color($color)` | Sets color of the thumb when the switch is toggled off.
 `mdc-switch-toggled-on-ripple-color($color)` | Sets the color of the ripple surrounding the thumb when the switch is toggled on.
 `mdc-switch-toggled-off-ripple-color($color)` | Sets the color of the ripple surrounding the thumb when the switch is toggled off.
-`mdc-switch-ripple-hover-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when hovered.
-`mdc-switch-ripple-focus-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when focused.
-`mdc-switch-ripple-press-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when pressed.
+`mdc-switch-ripple-states-opacity($opacity-map)` | Sets the opacity of the ripple surrounding the thumb in any of the `hover`, `focus`, or `press` states. The `opacity-map` can specify any of these states as keys. States not specified in the map resort to default opacity values.
 
 ## `MDCSwitch` Properties and Methods
 
--- a/packages/mdc-switch/_mixins.scss
+++ b/packages/mdc-switch/_mixins.scss
@@ -158,9 +158,9 @@
   }
 }
 
-@mixin mdc-switch-toggled-on-ripple-color($color, $opacity-map: (), $query: mdc-feature-all()) {
+@mixin mdc-switch-toggled-on-ripple-color($color, $query: mdc-feature-all()) {
   &.mdc-switch--checked .mdc-switch__thumb-underlay {
-    @include mdc-states($color, false, $opacity-map, $query);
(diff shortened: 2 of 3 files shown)


packages/mdc-switch/README.md
```diff
@@ -132,6 +132,9 @@ Mixin | Description
 `mdc-switch-toggled-off-thumb-color($color)` | Sets color of the thumb when the switch is toggled off.
 `mdc-switch-toggled-on-ripple-color($color)` | Sets the color of the ripple surrounding the thumb when the switch is toggled on.
 `mdc-switch-toggled-off-ripple-color($color)` | Sets the color of the ripple surrounding the thumb when the switch is toggled off.
+`mdc-switch-ripple-hover-opacity($opacity)` | Sets the opacity of the ripple surrounding the thumb when hovered.
```
abhiomkar: We usually do not allow customization based on state. This is usually handled by mdc-states() mixin of mdc-ripple.

What's the context of this change?

packages/mdc-switch/_mixins.scss
```diff
@@ -192,6 +192,29 @@
   }
 }
 
+// Opacity customizations apply to both on and off states to ensure symmetry.
+// Both checked and unchecked selectors are included to preserve specificity.
+@mixin mdc-switch-ripple-hover-opacity($opacity, $query: mdc-feature-all()) {
+  &:not(.mdc-switch--checked) .mdc-switch__thumb-underlay,
+  &.mdc-switch--checked .mdc-switch__thumb-underlay {
+    @include mdc-states-hover-opacity($opacity, $query);
```
abhiomkar: I think we need to modify `mdc-states()` mixin to accept opacities map as optional parameter. This'll allow customizing `$mdc-ripple-*-ink-opacities` by mdc components.

Clients would send opacity map of each state instead of individually calling these mixins.

For example,

```scss
mdc-switch-states($color: primary, $opacity-map: (hover: .06, focus: .3));

// or 

mdc-switch-states-opacity($opacity-map: (hover: .06, focus: .3));
```

These mixins call mdc-state mixins with new opacity map argument.

/cc @patrickrodee

packages/mdc-switch/_mixins.scss
```diff
@@ -192,6 +192,29 @@
   }
 }
 
+// Opacity customizations apply to both on and off states to ensure symmetry.
+// Both checked and unchecked selectors are included to preserve specificity.
+@mixin mdc-switch-ripple-hover-opacity($opacity, $query: mdc-feature-all()) {
+  &:not(.mdc-switch--checked) .mdc-switch__thumb-underlay,
```
abhiomkar: Since this selectors covers both checked and unchecked state, we can just use `.mdc-switch__thumb-underlay` as a selector?

packages/mdc-switch/_mixins.scss
allan-chen: The principle of `mdc-states()` seems to be to apply the basic default opacities only, while any customizations are the responsibilities of individual `mdc-states-hover-opacity($opacity, $query)` mix-ins, so would modifying `mdc-states()` to include custom opacity map introduce problems with separation of concern? I could modify the interface at the MDC-switch mix-ins layer to take an opacity map instead but still use the underlying `mdc-states-*-opacity()` mix-ins in ripple?

packages/mdc-switch/_mixins.scss
allan-chen: This addresses specificity concerns - the base mixins for switch customize toggled-on and toggled-off colors/opacities for ripple separately, so to override them the same level of specificity is maintained

abhiomkar: Can we just use `&.mdc-switch .mdc-switch__thumb-underlay` in that case replace two selectors? Also adding a code comment would be helpful.

packages/mdc-switch/_mixins.scss
abhiomkar: If more than one component start allowing state opacity customizations we would then have to move this logic to `mdc-states()`. I think it is very uncommon that client would only modify a single state opacity, all state opacities would've to be applies at the same time to meet the contrast ratio. I do not see value in having separate mixin for each state (i.e., hover, focus, pressed, etc).

Having a single opacity mixin for switch that calls individual mdc-states opacities mixin SGTM ands looks reasonable for now.

patrickrodee: We have an internal client team who is asking to modify a single opacity (hover). We already have mixins for individual opacity levels (hover, focus, and press). This is exposing those in a more discrete method so clients don't need to reference MDC classes and can instead directly call mixins.

Adding an opacity map mixin is a good idea but feels orthogonal to the initial request.

We could modify the hover/focus/press opacity mixins to accept a prefix selector. That would accomplish the same goal and meet the client's needs.

packages/mdc-switch/_mixins.scss
allan-chen: Yes, that's a better solution! Also added comment.

packages/mdc-switch/_mixins.scss
patrickrodee: Hmmm the prefix selector has the same problem of requiring knowledge and access to the MDC classnames 😞 [See this demo](https://www.sassmeister.com/gist/495fca8b4a58f0ba75cfb903e577b2bc) for an example.

abhiomkar: mdc-states() already has `$ripple-target` to support this feature but I think that would require clients to use mdc-* class names which is not intended.

I think we've two options:

1) Have the map merge with default opacity map to allow granular customization.
2) Keep individual mixins for each state (hover, focus, etc) like what Allan did.

allan-chen: @abhiomkar pushed a commit that drafts strategy 1 above. If this implementation is preferred I can update the docs accordingly.

packages/mdc-switch/_mixins.scss
```diff
@@ -192,6 +192,24 @@
   }
 }
 
+// Opacity customizations apply to both on and off states to ensure symmetry.
```
abhiomkar: Follow SassDoc convention. Also add docs for param using `@param` annotation.

See http://sassdoc.com/annotations/

PR #5130, opened 2019-09-26
fix(button): Set ripple target box-sizing to content-box

robert-j-webb:
Often users of the web apply box-sizing: border-box; to all of their components.
If this is done with a page that has a mdc-button, the ripple hover state will be off by a few pixels.
Since this practice is so common and it is a very difficult bug to track down and fix, we can take a preventitve measure and enforce the default style on the ripple target.
This prevents the problem from happening and has no effect on users who do not use the universal border-box rule.


patrickrodee:
Looks like lint is failing on the CSS property order. Here's the order we follow: https://github.com/material-components/material-components-web/blob/4bde8b4a90f721543ecce5be77aa5ca15944557c/.stylelintrc.yaml#L32

robert-j-webb:
Updated

PR #5131, opened 2019-09-26
fix(slider): Applies secondary color to slider track container

svencowart:
Overcomes css limitation of not being able to use a hex value inside of an rgba value by using the after pseudo class and applying opacity on it instead.  In this way, the container and track are left alone and only the track container has the opacity applied to it.


svencowart:
@googlebot I fixed it.

PR #5132, opened 2019-09-26
fix(slider): use secondary custom property color for slider container

svencowart:
Overcomes css limitation of not being able to use a hex value inside of an rgba value by using the after pseudo class and applying opacity on it instead.  In this way, the container and track are left alone and only the track container has the opacity applied to it.

### DIFF
--- a/packages/mdc-slider/_mixins.scss
+++ b/packages/mdc-slider/_mixins.scss
@@ -435,8 +435,16 @@
   $feat-color: feature-targeting-functions.create-target($query, color);
 
   .mdc-slider__track-container {
-    @include feature-targeting-mixins.targets($feat-color) {
-      @include theme-mixins.prop(background-color, rgba(theme-variables.prop-value($color), $opacity));
+    &::after {
+      display: block;
+      width: 100%;
+      height: 100%;
+      opacity: $opacity;
+      content: "";
+
+	  @include feature-targeting-mixins.targets($feat-color) {
+        @include theme-mixins.prop(background-color, rgba(theme-variables.prop-value($color), $opacity));
+      }
     }
   }
 }
--- a/packages/mdc-slider/_mixins.scss
+++ b/packages/mdc-slider/_mixins.scss
@@ -435,9 +435,9 @@
   $feat-color: feature-targeting-functions.create-target($query, color);
 
   .mdc-slider__track-container {
-	@include feature-targeting-mixins.targets($feat-color) {
+	  @include feature-targeting-mixins.targets($feat-color) {
       &::after {
-        @include theme-mixins.prop(background-color, rgba(theme-variables.prop-value($color), $opacity));
+        @include theme-mixins.prop(background-color, $color);
 
         display: block;
         width: 100%;
--- a/packages/mdc-slider/_mixins.scss
+++ b/packages/mdc-slider/_mixins.scss
@@ -435,15 +435,15 @@
   $feat-color: feature-targeting-functions.create-target($query, color);
 
   .mdc-slider__track-container {
-    &::after {
-      display: block;
-      width: 100%;
-      height: 100%;
-      opacity: $opacity;
-      content: "";
-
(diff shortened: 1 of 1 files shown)


packages/mdc-slider/_mixins.scss
```diff
@@ -435,8 +435,16 @@
   $feat-color: mdc-feature-create-target($query, color);
 
   .mdc-slider__track-container {
-    @include mdc-feature-targets($feat-color) {
-      @include mdc-theme-prop(background-color, rgba(mdc-theme-prop-value($color), $opacity));
+    &::after {
```
abhiomkar: This PR does not seem to change the color of slider to secondary color?

abhiomkar: How is this using pseudo selector different from directly applying on the element?

svencowart: Add the following to the test/screenshot/spec/mdc-slider/clases/baseline_continuous.html:

Inside of the head tag:
```
<style>
  .test-secondary-color {
    --mdc-theme-secondary: orange;
  }
</style>
```

Inside of the body, add a second slider with the test-secondary-color class:
```
<div class="mdc-slider test-slider test-secondary-color" tabindex="0" role="slider"
     aria-valuemin="0" aria-valuemax="100" aria-valuenow="0"
     aria-label="Select Value">
  <div class="mdc-slider__track-container">
    <div class="mdc-slider__track"></div>
  </div>
  <div class="mdc-slider__thumb-container">
    <svg class="mdc-slider__thumb" width="21" height="21">
      <circle cx="10.5" cy="10.5" r="7.875"></circle>
    </svg>
    <div class="mdc-slider__focus-ring"></div>
  </div>
</div>
```

Then run `npm start` and navigate to that page.  You will see the added slider use the secondary color.

svencowart: This comment on the issue over in material-components-web-components has more context:  https://github.com/material-components/material-components-web-components/issues/80#issuecomment-535065517

Without the pseudo selector, the color of the slider track defaults to the primary color because it uses the mdc-theme-prop-value which cannot handle secondary css variable definitions.  Additionally, because rgba is used here previously, css variables will not function properly because passing in a css variable that is a hex value to an rgba function will invalid.  The only way to make this work with css variables is by using a pseudo-class that fills the parent element and applying that css variable to the background-color and setting an opacity value.  Applying an opacity value directly to the track container will also not work because that will then also apply to its children which means the track itself will be opaque which is not the desired state either.

TitanNano:
@svencowart @abhiomkar is there any change this will get merged in the near future? It would fix a quite serious bug https://github.com/material-components/material-components-web-components/issues/80

asyncLiz:
@svencowart can you rebase master onto this branch?

```sh
# If you don't already have upstream set to this repo
git remote set-url upstream git@github.com:material-components/material-components-web.git

git pull --rebase upstream master
git push --force
```

svencowart:
@asyncLiz I have rebased master onto this branch, but it seems the unit tests are not working because of some issue with starting Chrome.  My permissions do not allow me to look into it further.

asyncLiz:
Yea we're having some build issues, I'll look into it while I run internal tests

packages/mdc-slider/_mixins.scss
```diff
@@ -435,8 +435,16 @@
   $feat-color: feature-targeting-functions.create-target($query, color);
 
   .mdc-slider__track-container {
-    @include feature-targeting-mixins.targets($feat-color) {
-      @include theme-mixins.prop(background-color, rgba(theme-variables.prop-value($color), $opacity));
+	  @include feature-targeting-mixins.targets($feat-color) {
```
abhiomkar: nit: fix indentation.

abhiomkar:
@asyncLiz Unit tests does not work on forked repo. Let's merge this once the CL is submitted. Thanks!

asyncLiz:
See [cl/290088778](http://cl/290088778)

PR #5136, opened 2019-09-27
fix(text-field): Fixes input text alignment on IE11 for densed textfield

abhiomkar:


packages/mdc-textfield/_mixins.scss
```diff
@@ -91,6 +91,21 @@
 ///
 @mixin mdc-text-field-height($height) {
   height: $height;
+
+  @if $height <= 40px {
+    // Target only IE 11
```
joyzhong: Add a comment indicating that this fixes text alignment in IE 11 for dense textfields?

PR #5141, opened 2019-10-01
fix(ripple): Always set even num when initial ripple size is calculated

abhiomkar:


abhiomkar:
> Is there a screenshot test we could add for this?

I think we should try to update our internal screenshot infra to support this, the current test for FOCUS does not really cover this these values are programmatically generated on the fly.

packages/mdc-ripple/foundation.ts
```diff
@@ -502,7 +502,9 @@ export class MDCRippleFoundation extends MDCFoundation<MDCRippleAdapter> {
     this.maxRadius_ = this.adapter_.isUnbounded() ? maxDim : getBoundedRadius();
 
     // Ripple is sized as a fraction of the largest dimension of the surface, then scales up using a CSS scale transform
-    this.initialSize_ = Math.floor(maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE);
+    const initialSize = Math.floor(maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE);
+    // Ripple size should always be even number to equally center align when unbounded.
+    this.initialSize_ = initialSize % 2 === 0 ? initialSize : initialSize - 1;
```
joyzhong: Add the unbounded check that you added internally?

abhiomkar: Good catch! synced the code with internal.

packages/mdc-ripple/foundation.ts
```diff
@@ -502,7 +502,15 @@ export class MDCRippleFoundation extends MDCFoundation<MDCRippleAdapter> {
...
-    this.initialSize_ = Math.floor(maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE);
+    const initialSize = Math.floor(maxDim * MDCRippleFoundation.numbers.INITIAL_ORIGIN_SCALE);
+    // Unbounded ripple size should always be even number to equally center align.
+    if (this.adapter_.isUnbounded() && initialSize % 2 !== 0) {
+      this.initialSize_ = initialSize - 1;
+    } else {
+      this.initialSize_ = initialSize;
+    }
+    // Ripple size should always be even number to equally center align when unbounded.
+    this.initialSize_ = initialSize % 2 === 0 ? initialSize : initialSize - 1;
```
joyzhong: Remove these 2 lines.

PR #5144, opened 2019-10-02
feat(tab): Add text transform mixin

patrickrodee:
Used internally to set the text transform property


packages/mdc-tab/_mixins.scss
```diff
@@ -304,3 +304,11 @@
     }
   }
 }
+
+@mixin mdc-tab-text-transform($transform, $query: mdc-feature-all()) {
```
abhiomkar: Add sassdoc (http://sassdoc.com/annotations/)

Also update README file.

PR #5147, opened 2019-10-02
fix(text-field): Fixes input text alignment on IE11 for densed textfield 2

abhiomkar:
Other changes:

- notched-outline: Remove padding calculation from notch offset
- Removed applying color to asterisk since it is inherited from parent (floating label).

This PR keeps text field in sync with internal version.

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -100,7 +100,6 @@
         // Set line-height to the height of input element excluding padding & border.
         line-height:
           $height
-          - $mdc-text-field-input-padding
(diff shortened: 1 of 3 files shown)


packages/mdc-textfield/_mixins.scss
```diff
@@ -92,15 +92,13 @@
 @mixin mdc-text-field-height($height) {
   height: $height;
 
-  // Target only dense text field.
```
patrickrodee: It this comment no longer accurate?

abhiomkar: Added this comment back.

PR #5154, opened 2019-10-04
feat(tab-scroller): Mixin for scroll transition

### LINKED ISSUE #5123: [MDCTabScroller] needs a mixin to disable animations and/or alter CSS transition time
I'm using TabBar for a scrollable set of tabs, but since Shift+MouseWheel is a little onerous for horizontal scrolling (for non-Mac, non-touch users), I also need to be able to scroll the TabBar programmatically (when the user hovers over divs at the TabBar edges).

Previously I've done this in JS via setTimeout & scrollLeft. To avoid reaching into TabScroller's internals I'm prepared to use its incrementScroll method, but that uses an animation with a 250ms transition duration, preventing me from, say, manually moving 2 pixels every frame. To go this route I'd need a Sass mixin to disable the CSS transition.

I could instead allow the transition to happen and stop it when the user mouses out of the edge divs. For this I'd need a mixin to *increase* the transition duration to a few seconds (otherwise the user can't stop it when they see the tab they want) AND a method in the controller to stop the animation (I'm prepared to just call handleInteraction, but it would probably be more semantic to just make stopScrollAnimation public.)

https://github.com/material-components/material-components-web/blob/ef99808cba809294cb2d90903d12cdf930071f4d/packages/mdc-tab-scroller/foundation.ts#L293

cletusw:
Fixes #5123

### DIFF
--- a/packages/mdc-tab-scroller/_mixins.scss
+++ b/packages/mdc-tab-scroller/_mixins.scss
@@ -26,12 +26,36 @@
 @import "@material/feature-targeting/mixins";
 @import "./variables";
 
+///
+/// Sets the CSS transition for the tab scrolling animation.
+///
+/// @param {Number | String} $duration-ms - Duration (in ms) of the animation.
+/// @param {String} $timing-function - Optionally overrides the default animation timing function.
+///
+@mixin mdc-tab-scroller-transition(
+  $duration-ms,
+  $timing-function: $mdc-animation-standard-curve-timing-function,
+  $query: mdc-feature-all()
+) {
+  $feat-animation: mdc-feature-create-target($query, animation);
+
+  &.mdc-tab-scroller--animating .mdc-tab-scroller__scroll-content {
+    @include mdc-feature-targets($feat-animation) {
+      transition: $duration-ms transform $timing-function;
+    }
+  }
+}
+
 @mixin mdc-tab-scroller-core-styles($query: mdc-feature-all()) {
   $feat-structure: mdc-feature-create-target($query, structure);
   $feat-animation: mdc-feature-create-target($query, animation);
 
   // postcss-bem-linter: define tab-scroller
   .mdc-tab-scroller {
+    @include mdc-tab-scroller-transition(
+      $duration-ms: $mdc-tab-scroller-transition-duration,
+      $query: $query);
+
     @include mdc-feature-targets($feat-structure) {
       overflow-y: hidden;
     }
@@ -98,12 +122,6 @@
     .mdc-tab-scroller__scroll-area {
       @include mdc-feature-targets($feat-animation) {
         -webkit-overflow-scrolling: auto;
-      }
-    }
-
-    .mdc-tab-scroller__scroll-content {
-      @include mdc-feature-targets($feat-animation) {
-        transition: $mdc-tab-scroller-transition-duration transform $mdc-animation-standard-curve-timing-function;
       }
     }
   }
--- a/packages/mdc-tab-scroller/README.md
+++ b/packages/mdc-tab-scroller/README.md
@@ -71,6 +71,12 @@ CSS Class | Description
 
 > _NOTE_: The `align` modifier classes are only applicable when the contents do not meet or exceed the width of the Tab Scroller and Tab Bar (i.e., most commonly, when `mdc-tab--min-width` is used on each tab).
 
+### Sass Mixins
+
+Mixin | Description
+--- | ---
+`mdc-tab-scroller-transition($duration-ms, $timing-function)` | Sets the scroll animation duration and, optionally, the timing function.
+
 ## `MDCTabScroller` Methods
 
 Method Signature | Description


packages/mdc-tab-scroller/_mixins.scss
```diff
@@ -26,10 +26,28 @@
...
 @import "./variables";
 
+@mixin mdc-tab-scroller-transition(
+  $duration-ms,
+  $timing-function: $mdc-animation-standard-curve-timing-function,
+  $query: mdc-feature-all()
+) {
+  $feat-animation: mdc-feature-create-target($query, animation);
+
+  .mdc-tab-scroller--animating {
```
abhiomkar: Add `&` here.

`&.mdc-tab-scroller--animating` 

Mixins are usually applied at root of a component. `.mdc-tab-scroller` in this case.

packages/mdc-tab-scroller/_mixins.scss
```diff
@@ -26,10 +26,28 @@
 @import "@material/feature-targeting/mixins";
 @import "./variables";
 
+@mixin mdc-tab-scroller-transition(
```
abhiomkar: Please add SassDoc (http://sassdoc.com/annotations/) for this mixin.

See https://github.com/material-components/material-components-web/blob/develop/packages/mdc-button/_mixins.scss#L237 for example.

You would also need to update README file in this folder.

abhiomkar: Can you merge this to `develop` branch. This is the branch which gets synced to google3/.

packages/mdc-tab-scroller/_mixins.scss
abhiomkar: You would also need to make sure that tabscroller accepts custom class names to apply this mixin.

packages/mdc-tab-scroller/_mixins.scss
cletusw: It looks like develop is currently behind? It's missing my previous PR #5142 which is causing merge conflicts.

packages/mdc-tab-scroller/_mixins.scss
cletusw: Not sure what you mean by custom class names -- is there an example elsewhere I can imitate?

cletusw: Also, I'm a little confused about the &. Currently I've applied this mixin at the root of `@mixin mdc-tab-scroller-core-styles`, *outside* its `.mdc-tab-scroller` rule. If I'm understanding correctly, I should use the & but move the application *inside* `.mdc-tab-scroller { ... }`?

packages/mdc-tab-scroller/_mixins.scss
abhiomkar: Just synced your changes to develop.

`develop` is always ahead of `master`. Can you please rebase your branch with origin/develop? Thanks!

packages/mdc-tab-scroller/_mixins.scss
abhiomkar: The tab scroller template (Google-internal) should accept your class name to apply this mixin.

For example,

```
{call .tabScroller}
  {param classes: css('%CustomTabScroller') /}
{/call}
```

packages/mdc-tab-scroller/_mixins.scss
```diff
@@ -24,11 +24,36 @@
...
+      }
+    }
+  }
+}
 
 @mixin mdc-tab-scroller-core-styles($query: mdc-feature-all()) {
   $feat-structure: mdc-feature-create-target($query, structure);
   $feat-animation: mdc-feature-create-target($query, animation);
 
+  @include mdc-tab-scroller-transition($mdc-tab-scroller-transition-duration, $query)
```
abhiomkar: This should be applied inside `.mdc-tab-scroller` selector.

packages/mdc-tab-scroller/_mixins.scss
abhiomkar: Please see my comment here: https://github.com/material-components/material-components-web/pull/5154/files#r332697955

Theme mixins are always applied at root of that component.

cletusw: Oh, that makes sense. Thank you for both clarifications! Done.

packages/mdc-tab-scroller/_mixins.scss
cletusw: Done!

packages/mdc-tab-scroller/_mixins.scss
cletusw: Done, thanks!

packages/mdc-tab-scroller/_mixins.scss
```diff
@@ -26,6 +26,28 @@
...
+///
+@mixin mdc-tab-scroller-transition(
+  $duration-ms,
+  $timing-function: $mdc-animation-standard-curve-timing-function,
+  $query: mdc-feature-all()
+) {
+  $feat-animation: mdc-feature-create-target($query, animation);
+
+  &.mdc-tab-scroller--animating {
+    .mdc-tab-scroller__scroll-content {
```
abhiomkar: nit: `&.mdc-tab-scroller--animating .mdc-tab-scroller__scroll-content` {

packages/mdc-tab-scroller/_mixins.scss
```diff
@@ -35,6 +57,8 @@
     @include mdc-feature-targets($feat-structure) {
       overflow-y: hidden;
     }
+
+    @include mdc-tab-scroller-transition($mdc-tab-scroller-transition-duration, $query)
```
abhiomkar: Move this line before above include statement.

cletusw: Note this is a change in the generated CSS. Before:

```css
.mdc-tab-scroller--animating .mdc-tab-scroller__scroll-content {...}
```

whereas now it will be:

```css
.mdc-tab-scroller .mdc-tab-scroller--animating .mdc-tab-scroller__scroll-content {...}
```

The added specificity shouldn't be a problem since MDC doesn't support style overrides except via mixin/variable, correct?

packages/mdc-tab-scroller/_mixins.scss
cletusw: Done.

packages/mdc-tab-scroller/_mixins.scss
cletusw: Done.

PR #5160, opened 2019-10-07
refactor(switch): move core ripple opacity logic from switch to ripple

allan-chen:
This addresses the issue of easy re-usability of ripple opacity mix-ins across all components, not just switch, as raised in #5126. It introduces a single interface for customizing one or more ripple states. Changes are backwards compatible, since old mixins for customizing individual ripple states are kept and simply wrapped in the new mixin.

### DIFF
--- a/packages/mdc-ripple/README.md
+++ b/packages/mdc-ripple/README.md
@@ -113,7 +113,7 @@ Mixin | Description
 
 > _NOTE_: `$has-nested-focusable-element` defaults to `false` but should be set to `true` if the component contains a focusable element (e.g. an input) inside the root element.
 
-> _DEPRECATED_: The individual mixins `mdc-states-hover-opacity($opacity)`, `mdc-states-focus-opacity($opacity, $has-nested-focusable-element)`, and `mdc-states-press-opacity($opacity)` are are deprecated in favor of the unified `mdc-states-opacities($opacity-map, $has-nested-focusable-element)` mixin above.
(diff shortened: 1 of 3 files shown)


packages/mdc-ripple/README.md
```diff
@@ -67,9 +67,7 @@ In order to fully style the ripple effect for different states (hover/focus/pres
   @include mdc-ripple-surface;
   @include mdc-ripple-radius-bounded;
   @include mdc-states-base-color(black);
-  @include mdc-states-hover-opacity(.1);
-  @include mdc-states-focus-opacity(.3);
-  @include mdc-states-press-opacity(.4);
+  @include mdc-states-opacities(("hover": .1, "focus": .3, "press": .4));
```
abhiomkar: You wouldn't need double quotes.

`(hover: .1, focus: .3, press: .4)`

packages/mdc-ripple/_mixins.scss
```diff
@@ -201,6 +201,24 @@
   }
 }
 
+/// Customizes ripple opacities in `hover`, `focus`, or `press` states
+/// @param {map} $opacity-map - map specifying custom opacity of zero or more states
+/// @param {bool} $has-nested-focusable-element - whether the component contains a focusable element in the root
```
abhiomkar: nit: Leave empty comment line in first and last line of this doc.

PR #5168, opened 2019-10-10
feat(radio): Add disabled state color mixins

allan-chen:
Adds the `mdc-radio-disabled-unchecked-stroke-color`, `mdc-radio-disabled-checked-stroke-color`, and the `mdc-disabled-radio-ink-color` mix-ins for radio buttons in disabled state.

### DIFF
--- a/packages/mdc-radio/_mixins.scss
+++ b/packages/mdc-radio/_mixins.scss
@@ -50,14 +50,14 @@ $mdc-radio-ripple-target: ".mdc-radio__ripple";
   $feat-structure: mdc-feature-create-target($query, structure);
 
   .mdc-radio {
-    @include mdc-radio-unchecked-stroke-color($mdc-radio-unchecked-color, $query);
-    @include mdc-radio-checked-stroke-color($mdc-radio-baseline-theme-color, $query);
-    @include mdc-radio-ink-color($mdc-radio-baseline-theme-color, $query);
-    @include mdc-radio-disabled-unchecked-stroke-color($mdc-radio-circle-color, $query);
-    @include mdc-radio-disabled-checked-stroke-color($mdc-radio-circle-color, $query);
-    @include mdc-radio-disabled-ink-color($mdc-radio-circle-color, $query);
-    @include mdc-radio-focus-indicator-color($mdc-radio-baseline-theme-color, $query);
-    @include mdc-radio-density($mdc-radio-density-scale, $query);
+    @include mdc-radio-unchecked-stroke-color($mdc-radio-unchecked-color, $query: $query);
+    @include mdc-radio-checked-stroke-color($mdc-radio-baseline-theme-color, $query: $query);
+    @include mdc-radio-ink-color($mdc-radio-baseline-theme-color, $query: $query);
+    @include mdc-radio-disabled-unchecked-stroke-color($mdc-radio-disabled-circle-color, $query: $query);
+    @include mdc-radio-disabled-checked-stroke-color($mdc-radio-disabled-circle-color, $query: $query);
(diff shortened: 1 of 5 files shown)


packages/mdc-radio/_mixins.scss
```diff
@@ -53,6 +53,9 @@ $mdc-radio-ripple-target: ".mdc-radio__ripple";
     @include mdc-radio-unchecked-stroke-color($mdc-radio-unchecked-color, $query);
     @include mdc-radio-checked-stroke-color($mdc-radio-baseline-theme-color, $query);
     @include mdc-radio-ink-color($mdc-radio-baseline-theme-color, $query);
+    @include mdc-radio-disabled-unchecked-stroke-color($mdc-radio-circle-color, $query);
```
abhiomkar: nit: `$query: $query` It is a good practice to name the optional parameters so it still works when we add a new parameter to this mixin.

packages/mdc-radio/_mixins.scss
```diff
@@ -393,3 +411,108 @@ $mdc-radio-ripple-target: ".mdc-radio__ripple";
...
+  }
+}
+
+///
+/// Helps select the radio background only when its native control is in the
+/// disabled state.
+/// @access private
+///
+@mixin mdc-radio-if-disabled_ {
+  [aria-disabled="true"] .mdc-radio__native-control,
```
abhiomkar: is aria selector necessary?

test/screenshot/golden.json
```diff
@@ -952,11 +952,11 @@
     }
   },
   "spec/mdc-radio/mixins/mixins.html": {
-    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2019/09/20/22_06_30_583/spec/mdc-radio/mixins/mixins.html?utm_source=golden_json",
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/allanchen/2019/10/10/22_20_16_416/spec/mdc-radio/mixins/mixins.html?utm_source=golden_json",
```
abhiomkar: Are there no screenshot changes to Radio's baseline.html?

abhiomkar: Radio has grey colors for disabled in [current screenshots](https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2019/09/20/22_06_30_583/spec/mdc-radio/classes/baseline.html.windows_chrome_76.png).

packages/mdc-radio/_mixins.scss
abhiomkar: Can we rename this variable to have `*-disabled-*` name?

abhiomkar: the current color for `$mdc-radio-circle-color` is on-surface, is this correct?

test/screenshot/golden.json
allan-chen: Per discussion, this is intended to be left as is

packages/mdc-radio/_mixins.scss
allan-chen: `on-surface` is correct. However, what opacity should be is more ambiguous.

packages/mdc-radio/_mixins.scss
allan-chen: I assumed it was there for a11y reasons. Any reason it may potentially be obsolete?

packages/mdc-radio/_mixins.scss
allan-chen: The plan is to make this only about adding mixins, then correcting incorrect colors (mostly opacities) in separate PR so the changes are clear.

PR #5174, opened 2019-10-14
fix(touch-target): Add class to touch target wrapper.

joyzhong:
This will make it more transparent to clients that the wrapper element is part of the touch target configuration, as well as make future changes easier.

### DIFF
--- a/packages/mdc-touch-target/_mixins.scss
+++ b/packages/mdc-touch-target/_mixins.scss
@@ -25,7 +25,7 @@
 @import "@material/feature-targeting/mixins";
 @import "./variables";
 
-// Styles applied to the component's touch target wrapper element.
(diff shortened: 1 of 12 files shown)


packages/mdc-touch-target/_mixins.scss
```diff
@@ -20,11 +20,29 @@
 // THE SOFTWARE.
 //
 
+@import "@material/base/mixins";
 @import "@material/feature-targeting/functions";
 @import "@material/feature-targeting/mixins";
 @import "./variables";
 
-// Styles applied to the component's inner touch target element.
+// Styles applied to the component's touch target wrapper element.
```
abhiomkar: nit: Use three slashes for mixin description.

Issue #4802, opened 2019-06-09
[MDCDrawer] Add build-in support for change drawer type at runtime


m-alzam:
## Feature Request

Hello. I'am working on a app that should be adapted depending of the user's screen size. It uses a Drawer to show the navigation menu. The problem is the following: if I use a Modal Drawer, the user needs to click on the menu button to open it when he's in a large screen. Mobile users don't have this problem. To solve the problem, I can use a Dismissible Drawer, but it's not designed to mobile devices.

Is it possible to add a new element (like Responsive Drawer) that changes depending on the screen size? Or there's a easy way to change the type of Drawer using css `@media`? The Material Design specification says that "[Modal Drawer] can be replaced by standard drawers on tablet and desktop", but it is not easy to do in web (or I don't know how).

### Proposed solution

IMHO, adding a new type of Drawer is the easier solution. It's a good idea to add a section in the [Drawer Web Component](https://material.io/develop/web/components/drawers/) to help developers, showing a way to use a "responsive" drawer that adapts itself depending of the screen.

The #590 issue is similar to this, but it doesn't show a clear solution to the problem.

Thanks in advance.

abhiomkar:
Thanks for the suggestion Manuel! 

Please see example on Glitch for responsive drawer:
https://glitch.com/~material-responsive-drawer 

We can link this example to our drawer README file, since this is requested frequently. 

PR Welcome :)

samstride:
That was a useful example. Thanks.

Is there an example where the dismissible drawer is below the top app bar (ex: https://github.com/material-components/material-components-web/tree/master/packages/mdc-drawer#dismissible-drawer-below-top-app-bar) for non-mobile screens and switches to modal drawer for mobile screens?

youbet1980:
Yes like @samstride says I am looking for something that works exactly the same as news.google.com
Eg:
The Dismissible Drawer appears Non Modal when in desktop. ( But can still be hidden )
and as soon as it goes to tablet or mobile view it becomes modal.

m-alzam:
Hello.
If I have time this summer, I will try to create   a MDCResponsibleDrawer or something similar, @abhiomkar.
Thanks

m-alzam:
Hello
I have styling the MDCDrawer component to be responsive, but I don't know who to create or modify a component, so I decided to put the code here and ask if someone can help me.

CSS:
```scss
.mdc-drawer {
  position: inherit;
}

.mdc-drawer-scrim {
  display: none;
[...]

PR #5178, opened 2019-10-15
Add link to suggested implementation for responsive drawer

dbollinger:
In response to the comment suggesting this solution for responsive drawers in https://github.com/material-components/material-components-web/issues/4802#issuecomment-501823698


dbollinger:
@googlebot I fixed it.

PR #5180, opened 2019-10-15
fix(linear-progress): Fix indeterminate animation bug

allan-chen:
This fixes the bug where if the bar is in indeterminate state,
setting it to reverse after the page has loaded will cause the animation timers to go out of
sync and make the bar look funny.

BREAKING CHANGE: adapter has new forceLayout() method

### DIFF
--- a/packages/mdc-linear-progress/foundation.ts
+++ b/packages/mdc-linear-progress/foundation.ts
@@ -72,9 +72,11 @@ export class MDCLinearProgressFoundation extends MDCFoundation<MDCLinearProgress
       this.setScale_(this.adapter_.getBuffer(), this.buffer_);
     } else {
       if (this.isReversed_) {
-        // reset translation animation timer that was started when reverse
-        // class was introduced, as to keep it in sync with the new scaling
-        // animation timer that's about to start from adding indeterminate class
+        // Adding/removing REVERSED_CLASS starts a translate animation, while
+        // adding INDETERMINATE_CLASS starts a scale animation. Here, we reset
+        // the translate animation in order to keep it in sync with the new
+        // scale animation that will start from adding INDETERMINATE_CLASS
+        // below.
         this.adapter_.removeClass(cssClasses.REVERSED_CLASS);
         this.adapter_.forceLayout();
         this.adapter_.addClass(cssClasses.REVERSED_CLASS);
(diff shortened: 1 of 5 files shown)


packages/mdc-linear-progress/foundation.ts
```diff
@@ -64,11 +65,21 @@ export class MDCLinearProgressFoundation extends MDCFoundation<MDCLinearProgress
...
   setDeterminate(isDeterminate: boolean) {
     this.isDeterminate_ = isDeterminate;
+
     if (this.isDeterminate_) {
       this.adapter_.removeClass(cssClasses.INDETERMINATE_CLASS);
       this.setScale_(this.adapter_.getPrimaryBar(), this.progress_);
       this.setScale_(this.adapter_.getBuffer(), this.buffer_);
     } else {
+      if (this.isReversed_) {
+        // reset translation animation timer that was started when reverse
```
joyzhong: nit: Capitalize/add period at end of comment (and below)

packages/mdc-linear-progress/foundation.ts
```diff
@@ -64,11 +65,21 @@ export class MDCLinearProgressFoundation extends MDCFoundation<MDCLinearProgress
...
+        // class was introduced, as to keep it in sync with the new scaling
+        // animation timer that's about to start from adding indeterminate class
```
joyzhong: I'd reword to something like this for clarity:

Adding REVERSED_CLASS starts a `translate` animation, while adding INDETERMINATE_CLASS starts a `scale` animation. 
Here, we reset the `translate` animation in order to keep it in sync with the new `scale` animation that will start from adding INDETERMINATE_CLASS below.

packages/mdc-linear-progress/foundation.ts
```diff
@@ -91,6 +102,17 @@ export class MDCLinearProgressFoundation extends MDCFoundation<MDCLinearProgress
 
   setReverse(isReversed: boolean) {
     this.isReversed_ = isReversed;
+
+    if (!this.isDeterminate_) {
+      // reset scaling animation timer that was started when indeterminate
+      // class was introduced, as to keep it in sync with the reversed
+      // translation timer that's about to start from adding/removing reversed
+      // class
```
joyzhong: Also would reword to something like:

Adding/removing REVERSED_CLASS starts a `translate` animation, while adding INDETERMINATE_CLASS starts a `scale` animation. 
Here, we reset the `scale` animation in order to keep it in sync with the new `translate` animation that will start from adding/removing REVERSED_CLASS below.

packages/mdc-linear-progress/foundation.ts
allan-chen: Reworded. Thanks!!

Issue #5198, opened 2019-10-23
[MDCMenuSurface] duplicate export


dgoogles:
## Bug report
index.ts exports Corner and CornerBit twice which causes errors when building with rollup.

### Steps to reproduce
1. Import MDCMenuSurface into a polymer project and try to compile with polymer-cli

### Actual behavior
Rollup fails to compile with the duplicate export error
 
### Expected behavior
This module should be buildable using rollup

### Your Environment:

| Software         | Version(s) |
| ---------------- | ---------- |
| MDC Web          |3.2.0
| Browser          |N/A
| Operating System | OSX

### Possible solution
There is no need to export Corner and CornerBit since they are already exported when you do export * from ./constants

dgoogles:
https://github.com/material-components/material-components-web/blob/80a7ae4300ac94c9998a598c47be04ca828ea14a/packages/mdc-menu-surface/index.ts#L26

bonniezhou:
Hi, are you importing both `constants` and `index`? There is no need to import from both, please follow [these instructions](https://github.com/material-components/material-components-web/blob/80a7ae4300ac94c9998a598c47be04ca828ea14a/docs/importing-js.md) for importing JS, specifically if you're using Webpack/Rollup.

abhiomkar:
I think the fix looks correct, we accidentally exported Corner & CornerBit twice.

Thanks for the PR! It's merged now.

PR #5200, opened 2019-10-23
fix(menu-surface): remove duplicate export from menu-surface

dgoogles:
This addresses issue: https://github.com/material-components/material-components-web/issues/5198

The removed variables are already exported below when all constants are exported.


dgoogles:
@googlebot I signed it!

PR #5201, opened 2019-10-24
refactor(menu-surface): Allow any type during migration

bonniezhou:
Allow any type during TS migration.


bonniezhou:
@googlebot I consent.

PR #5208, opened 2019-10-30
feat(text-field): Add disabled state color mixins

allan-chen:
BREAKING CHANGE: Redundant mixins `mdc-text-field-textarea-fill-color`, `mdc-text-field-textarea-stroke-color`, `mdc-text-field-fullwidth-bottom-line-color` removed. Instead, use `mdc-text-field-fill-color`, `mdc-text-field-outline-color`, and `mdc-text-field-bottom-line-color` respectively to achieve the same effect.


packages/mdc-textfield/_mixins.scss
```diff
@@ -201,78 +201,150 @@
...
+/// @param {Color} $color - The desired bottom line color.
+///
+@mixin mdc-text-field-disabled-bottom-line-color($color) {
+  &.mdc-text-field--disabled {
     @include mdc-text-field-bottom-line-color_($color);
   }
 }
 
+///
+/// Customizes the hover text field bottom line color except the outlined and textarea variants.
```
abhiomkar: Instead of indicating that this mixin does not work for certain text field variants, I think we can say that it only works for filled text field variant.

abhiomkar: Same for above.

test/screenshot/spec/mdc-textfield/classes/disabled-character-counter.html
```diff
@@ -0,0 +1,101 @@
+<!DOCTYPE html>
```
abhiomkar: Can we use single screenshot test for all disabled text fields? We can adjust the size of viewport since it doesn't fit in `test-viewport--mobile`.

We want to reduce the number of screenshot counts. Also makes the migration easier.

packages/mdc-textfield/_mixins.scss
```diff
@@ -201,78 +201,150 @@
...
 }
 
-@mixin mdc-text-field-bottom-line-color($color) {
-  &:not(.mdc-text-field--disabled):not(.mdc-text-field--outlined):not(.mdc-text-field--textarea) {
+///
+/// Customizes the disabled text field bottom line color except the outlined and textarea variants.
+/// @param {Color} $color - The desired bottom line color.
+///
+@mixin mdc-text-field-disabled-bottom-line-color($color) {
+  &.mdc-text-field--disabled {
```
abhiomkar: Can we create `if-text-field-disabled` & `if-text-field-enabled` mixins similar to checkbox?

test/screenshot/spec/mdc-textfield/classes/disabled-character-counter.html
allan-chen: Could we defer this as a separate effort? We can work out a plan to thin down screenshot counts as a whole across components in a unified way.

abhiomkar: I think any new screenshots from now on should be limited and clubbed into one big screenshot when possible. Similar to how we test text field internally.

Otherwise, this'll get only piled up and makes our tests even slower. We do not want to refactor or unify our existing screenshots since we're migrating testing infra internally.

packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -28,6 +28,12 @@
   }
 }
 
+@mixin mdc-text-field-disabled-icon-color($color, $styleSecondIcon: false) {
```
abhiomkar: Add SassDoc for this mixin.

test/screenshot/golden.json
```diff
@@ -1399,6 +1399,14 @@
       "desktop_windows_ie@11": "https://storage.googleapis.com/mdc-web-screenshot-tests/abhiomkar/2019/11/01/18_25_15_143/spec/mdc-textfield/classes/baseline.html.windows_ie_11.png"
     }
   },
+  "spec/mdc-textfield/classes/disabled-character-counter.html": {
```
abhiomkar: Let's delete this file too since this is already covered in below screenshot.

packages/mdc-textfield/icon/_mixins.scss
allan-chen: On it

test/screenshot/golden.json
allan-chen: It's a baseline test that was missing before and failed to catch that the character counter color was previously incorrect when disabled. It's not one of the mixin tests

PR #5213, opened 2019-10-31
chore(button): remove whitespace

allan-chen:
part of cleanup for internal sync


allan-chen:
Incorporating into #5215 instead

PR #4997, opened 2019-08-14
refactor(mdc-fab): Move Ripple to inner Element.

bwobrien:

PR #5214, opened 2019-10-31
fix(fab): Add overflow: hidden; to ripple target to fix bounded ripple.

bwobrien:
Fix for bug introduced in PR #4997


abhiomkar:
Updated your branch to latest goldens. Screenshots should pass now.

PR #5231, opened 2019-11-05
feat(fab): Add support for increased touch target to mini FAB.

joyzhong:


packages/mdc-fab/_mixins.scss
```diff
@@ -305,6 +312,17 @@ $mdc-fab-icon-enter-duration_: 180ms;
   }
 }
 
+@mixin mdc-fab--touch_($query: mdc-feature-all()) {
```
abhiomkar: Can we move styles from this mixin and apply directly to `.mdc-fab--touch` css selector?

packages/mdc-fab/_mixins.scss
```diff
@@ -305,6 +312,17 @@ $mdc-fab-icon-enter-duration_: 180ms;
...
+  @include mdc-touch-target-component(
```
abhiomkar: Not in the scope of this PR but would it make sense to rename `mdc-touch-target-component()` to `mdc-touch-target-margin()`? I think this mixin is only applying margin to increase its touch size.

WDYT?

joyzhong: Makes sense! I'll do this in a follow-up CL/PR.

packages/mdc-fab/_mixins.scss
```diff
@@ -25,11 +25,12 @@
...
 @import "@material/ripple/variables";
-@import "@material/theme/functions";
-@import "@material/theme/mixins";
+@import "@material/rtl/mixins";
 @import "@material/shape/mixins";
 @import "@material/shape/functions";
-@import "@material/rtl/mixins";
+@import "@material/theme/functions";
+@import "@material/theme/mixins";
+@import "@material/touch-target/mixins";
```
abhiomkar: FAB's package.json missing dependency for touch target which might be causing failures in [canary release](https://github.com/material-components/material-components-web/commit/0c4d8f3923f9a089132ed8dca4062b72d3576aca/checks?check_suite_id=299135665)?

joyzhong: Sent https://github.com/material-components/material-components-web/pull/5236 to fix.

PR #5246, opened 2019-11-09
feat(icon-button): Add disabled state color mixins

allan-chen:


packages/mdc-icon-button/_mixins.scss
```diff
@@ -144,18 +144,26 @@
   }
 }
 
+///
+/// Sets the font color and the ripple color to the provided color value for an enabled icon button.
+/// @param {Color} $color - The desired font and ripple color.
+///
 @mixin mdc-icon-button-ink-color($color, $query: mdc-feature-all()) {
-  $feat-color: mdc-feature-create-target($query, color);
+  @include mdc-icon-button-ink-color_($color, $query: $query);
```
abhiomkar: Do we not need enabled selector for this?

packages/mdc-icon-button/_mixins.scss
```diff
@@ -182,14 +190,35 @@
     }
   }
 
-  &:disabled {
-    @include mdc-feature-targets($feat-color) {
-      @include mdc-theme-prop(color, text-disabled-on-light);
-    }
+  @include mdc-icon-button-if-disabled_ {
```
abhiomkar: We can use `mdc-icon-button-disabled-ink-color` mixin here.

packages/mdc-icon-button/_mixins.scss
allan-chen: It was never there beforehand, I assume people didn't run into problems with this?

packages/mdc-icon-button/_mixins.scss
allan-chen: Sure - I can extract the color part out and leave the pointer events stuff under `mdc-icon-button-if-disabled_`

packages/mdc-icon-button/_mixins.scss
abhiomkar: Ok. I think we can remove "for an enabled icon button." from docs.

allan-chen: Sounds good

PR #5249, opened 2019-11-12
feat(elevation): Add elevation overlay mixins

patrickrodee:
Add mixins to set the overlay fill color, opacity, size, parent position, and shadow.

BREAKING CHANGE: Functions moved into the `_functions.scss` file

### DIFF
--- a/packages/mdc-elevation/README.md
+++ b/packages/mdc-elevation/README.md
@@ -76,16 +76,26 @@ CSS Class | Description
 Mixin | Description
 --- | ---
 `mdc-elevation($z-value, $color, $opacity-boost)` | Sets the elevation to the z-space for that given elevation, and optionally sets the color and/or boosts the opacity of the shadow
+`mdc-elevation-overlay-common` | Called once per application to setup the universal elevation overlay styles
+`mdc-elevation-shadow($box-shadow)` | Sets the `box-shadow` of the closest parent selector
+`mdc-elevation-overlay-parent` | Sets the positioning of the overlay's parent element so that the overlay can be appropriately centered
+`mdc-elevation-overlay-size($width, $height: $width)` | Sets the width and height of the elevation overlay
+`mdc-elevation-overlay-fill-color($color)` | Sets the color of the elevation overlay
+`mdc-elevation-overlay-opacity($opacity)` | Sets the opacity of the elevation overlay
+
 
 Function | Description
 --- | ---
 `mdc-elevation-transition-value($duration, $easing)` | Returns a value for the `transition` property to transition an element between elevations
+`mdc-elevation-overlay-transition-value($duration, $easing)` | Returns a value for the `transition` property to transition the elevation overlay between elevations
 
 Variable | Description
 --- | ---
 `$mdc-elevation-property` | Default property for elevation transitions
 `$mdc-elevation-transition-duration` | Default duration value for elevation transitions
 `$mdc-elevation-transition-timing-function` | Default easing value for elevation transitions
+`$mdc-elevation-overlay-color` | Default color for the elevation overlay
+`$mdc-elevation-overlay-property` | Default property for the elevation overlay transitions
 
 If you need more configurability over your transitions, use the `mdc-elevation-transition-value` function in conjunction with the exported sass variables.
 
--- a/packages/mdc-elevation/README.md
+++ b/packages/mdc-elevation/README.md
@@ -53,7 +53,7 @@ However, you can also apply elevation to specific components using `mdc-elevatio
 
 #### Elevation Overlay
 
-The elevation overlay should appear *above* the component container but *below* the ripple. To accomplish this, the `.mdc-elevation-overlay` element should appear *before* the `.mdc-<component>__ripple` element in the DOM.  Here's sample markup for a button with a touch target.
+The elevation overlay should appear *above* the component container in the stacking context but *below* the ripple. To accomplish this, the `.mdc-elevation-overlay` element should appear *before* the `.mdc-<component>__ripple` element in the DOM context.  Here's sample markup for a button with a touch target.
 
 ```html
 <button class="mdc-button mdc-button--raised">
--- a/packages/mdc-elevation/README.md
+++ b/packages/mdc-elevation/README.md
@@ -51,6 +51,21 @@ However, you can also apply elevation to specific components using `mdc-elevatio
   </div>
 ```
 
+#### Elevation Overlay
+
+The elevation overlay should appear *above* the component container but *below* the ripple. To accomplish this, the `.mdc-elevation-overlay` element should appear *before* the `.mdc-<component>__ripple` element in the DOM.  Here's sample markup for a button with a touch target.
+
+```html
+<button class="mdc-button mdc-button--raised">
+  <div class="mdc-elevation-overlay"></div>
+  <div class="mdc-button__ripple"></div>
+  <i class="material-icons mdc-button__icon" aria-hidden="true">favorite</i>
+  <span class="mdc-button__label">Font Icon</span>
+</button>
+```
+
+This ensures the ripple parts are rendered *above* the overlay.
+
 ### Styles
 
 ```scss
--- a/packages/mdc-elevation/_mixins.scss
+++ b/packages/mdc-elevation/_mixins.scss
@@ -23,6 +23,7 @@
 @import "@material/base/mixins";
 @import "@material/feature-targeting/functions";
 @import "@material/feature-targeting/mixins";
+@import "@material/theme/mixins";
 @import "@material/theme/variables";
 @import "./functions";
 @import "./variables";
@@ -134,7 +135,7 @@
 
   @include mdc-elevation-overlay-selector_ {
     @include mdc-feature-targets($feat-color) {
-      background-color: $color;
+      @include mdc-theme-prop(background-color, $color);
     }
   }
 }
--- a/packages/mdc-elevation/_mixins.scss
+++ b/packages/mdc-elevation/_mixins.scss
@@ -89,6 +89,7 @@
   $feat-color: mdc-feature-create-target($query, color);
 
   @include mdc-feature-targets($feat-color) {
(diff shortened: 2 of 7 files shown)


packages/mdc-elevation/_mixins.scss
```diff
@@ -46,6 +47,113 @@
...
+
+    @include mdc-feature-targets($feat-animation) {
+      @include mdc-base-emit-once("mdc-elevation/common/animation") {
+        transition: mdc-elevation-overlay-transition-value();
+      }
+    }
+  }
+
+  @include mdc-base-emit-once("mdc-elevation/common/color") {
+    @include mdc-elevation-overlay-fill-color($mdc-elevation-overlay-color, $query);
```
abhiomkar: It is a good practice to always name the optional sass parameters, so it doesn't break when we introduce new param to a mixin. `$query: $query`

packages/mdc-elevation/_mixins.scss
```diff
@@ -46,6 +47,113 @@
...
+  $feat-animation: mdc-feature-create-target($query, animation);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-elevation-overlay-selector_ {
+    @include mdc-feature-targets($feat-structure) {
+      @include mdc-base-emit-once("mdc-elevation/common/structure") {
+        position: absolute;
+        top: 50%;
+        right: 50%;
+        bottom: 50%;
```
joyzhong: Are right, bottom properties necessary?

packages/mdc-elevation/_variables.scss
```diff
@@ -116,6 +116,14 @@ $mdc-elevation-ambient-map: (
 // example in a `will-change` rule.
 $mdc-elevation-property: box-shadow !default;
 
+// The default color for the elevation overlay.
+$mdc-elevation-overlay-color: #fff;
```
abhiomkar: I think this should be `surface`.

packages/mdc-elevation/_mixins.scss
```diff
@@ -95,3 +220,16 @@
...
 }
+
+///
+/// Sets the elevation overlay transition value.
+///
+/// @param {String} $duration - The duration of the transition.
+/// @param {String} $easing - The easing function for the transition.
+/// @return {String}
+///
+@function mdc-elevation-overlay-transition-value(
```
abhiomkar: Can we move this into `_functions.scss`?

packages/mdc-elevation/_mixins.scss
```diff
@@ -71,11 +179,28 @@
...
+///
+/// Sets the elevation overlay transition value.
+///
+/// @param {String} $duration - The duration of the transition.
+/// @param {String} $easing - The easing function for the transition.
+/// @return {String}
+/// @access private
+///
+@mixin mdc-elevation-overlay-selector_ {
+  .mdc-elevation-overlay {
```
abhiomkar: Optional: would it be easier to read if we directly use `.mdc-elevation-overlay` selector?

packages/mdc-elevation/_mixins.scss
```diff
@@ -46,6 +47,113 @@
...
+    /* @alternate */
+    position: relative;
+  }
+}
+
+///
+/// Sets the elevation overlay size.
+///
+/// @param {Number} $width - The width of the elevation overlay.
+/// @param {Number} $height [$width] - The height of the elevation overlay (defaults to the width)
```
joyzhong: nit: No need for "defaults to the width" comment, since it's already implied from [$width] docs.

packages/mdc-elevation/README.md
```diff
@@ -76,16 +76,26 @@ CSS Class | Description
 Mixin | Description
 --- | ---
 `mdc-elevation($z-value, $color, $opacity-boost)` | Sets the elevation to the z-space for that given elevation, and optionally sets the color and/or boosts the opacity of the shadow
+`mdc-elevation-overlay-common` | Called once per application to setup the universal elevation overlay styles
```
abhiomkar: I think it is useful if you can add sample usage section to this doc, for example, how would you apply this elevation on a button.

packages/mdc-elevation/_mixins.scss
```diff
@@ -46,6 +47,113 @@
...
+/// Sets the elevation overlay fill color.
+///
+/// @param {Color} $color - The color of the elevation overlay.
+///
+@mixin mdc-elevation-overlay-fill-color($color, $query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+
+  @include mdc-elevation-overlay-selector_ {
+    @include mdc-feature-targets($feat-color) {
+      background: $color;
```
joyzhong: background-color?

packages/mdc-elevation/_mixins.scss
```diff
@@ -46,6 +47,113 @@
...
 }
 
+///
+/// Called once per application to set up the global default elevation styles.
+///
+@mixin mdc-elevation-overlay-common($query: mdc-feature-all()) {
+  $feat-animation: mdc-feature-create-target($query, animation);
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-elevation-overlay-selector_ {
```
abhiomkar: mdc elevation layer doesn't seem to have z-index value. how is this stacked in relation to ripple and other inner elements of a component?

packages/mdc-elevation/_mixins.scss
patrickrodee: Likely not. GSS does some funky calculations with left/right swaps if only one property is specified. I'll remove for now and we can add them back if necessary.

packages/mdc-elevation/_mixins.scss
patrickrodee: SGTM!

packages/mdc-elevation/_mixins.scss
joyzhong: Ah yeah, I think you'll need the /* @noflip */ annotations above `left`, or the GSS compiler will flip it to `right` in RTL mode (which isn't necessary since we're just centering here).

packages/mdc-elevation/_variables.scss
patrickrodee: It's going to be the inverse of `surface`. In dark theme, surface becomes `#121212` but the overlay should be `white`.

packages/mdc-elevation/_mixins.scss
patrickrodee: Sure thing! I'll move all the functions in there.

packages/mdc-elevation/_mixins.scss
patrickrodee: Could very well be. I prefer the mixin simply for the added indirection. Should the classname need to change, we'd only need to update the markup and this mixin.

I'm all ears though. If you'd prefer using the selector directly, I'll happily update it.

packages/mdc-elevation/README.md
patrickrodee: Great idea, I'll add that.

packages/mdc-elevation/_mixins.scss
patrickrodee: From discussion with design, the overlay appears directly above the surface but below the ripple.

patrickrodee: Will include this information in the docs demo

packages/mdc-elevation/README.md
patrickrodee: Docs update

packages/mdc-elevation/_mixins.scss
patrickrodee: Docs update

packages/mdc-elevation/README.md
```diff
@@ -51,6 +51,21 @@ However, you can also apply elevation to specific components using `mdc-elevatio
   </div>
 ```
 
+#### Elevation Overlay
+
+The elevation overlay should appear *above* the component container but *below* the ripple. To accomplish this, the `.mdc-elevation-overlay` element should appear *before* the `.mdc-<component>__ripple` element in the DOM.  Here's sample markup for a button with a touch target.
```
abhiomkar: Explicitly mention above / below are referring to stacking the layer (z-index) since we're referring DOM order in following section.

packages/mdc-elevation/_mixins.scss
```diff
@@ -46,6 +48,112 @@
...
+/// Sets the elevation overlay fill color.
+///
+/// @param {Color} $color - The color of the elevation overlay.
+///
+@mixin mdc-elevation-overlay-fill-color($color, $query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+
+  @include mdc-elevation-overlay-selector_ {
+    @include mdc-feature-targets($feat-color) {
+      background-color: $color;
```
abhiomkar: Use `mdc-theme-prop()` to apply color.

packages/mdc-elevation/README.md
patrickrodee: Done

packages/mdc-elevation/_mixins.scss
patrickrodee: Done

packages/mdc-elevation/_mixins.scss
```diff
@@ -71,27 +180,27 @@
...
-// Calling this function directly is useful in situations where a component needs to transition
-// more than one property.
-//
-// ```scss
-// .foo {
-//   transition: mdc-elevation-transition-value(), opacity 100ms ease;
-//   will-change: $mdc-elevation-property, opacity;
-// }
-// ```
-@function mdc-elevation-transition-value(
```
abhiomkar: Since this PR is a breaking change for `mdc-elevation-transition-value` function, let's add a BREAKING CHANGE line in your commit message when you squash & merge.

patrickrodee: Done!

PR #5254, opened 2019-11-14
fix(fab): Adjust fab line-height

EstebanG23:


packages/mdc-fab/_mixins.scss
```diff
@@ -333,6 +333,8 @@ $mdc-fab-icon-enter-duration_: 180ms;
     width: auto;
     max-width: 100%;
     height: $mdc-fab-extended-height;
+    /* @alternate */
```
abhiomkar: Can you add a code comment about why we do this?

EstebanG23: Is is mostly an issue with the CSS compilation because there are multiple identical, non-alternate declarations in the same ruleset. line-height is being set twice within `.mdc-fab--extended`, once in `mdc-typography(button, $query: $query)` and then again here. Similar code is done within button/mixins.scss.

Added comment!

abhiomkar: Sorry, I meant comment for `line-height: inherit` :)

EstebanG23: No worries, updated the comment 👍

PR #5255, opened 2019-11-14
fix(select): Do not fire change event on programmatic change

### LINKED ISSUE #5234: MDCSelect 4.0.0 fires change event twice
## Bug report
MDCSelectbox 4.0.0 fires change event twice.

When setting a value via javascript the change event is fired twice. I think it should not fire an event at all when changed programmatically.
There was already an bug report which confirms this: https://github.com/material-components/material-components-web/issues/3577

Here is a example of the issue:
Codepen: https://codepen.io/Jochen99/pen/VwwXRzm

bonniezhou:
Fixes #5234 

Changes to `layout-grid` and `textfield` are purely to pass the linter.


packages/mdc-select/component.ts
```diff
@@ -181,7 +181,9 @@ export class MDCSelect extends MDCComponent<MDCSelectFoundation> {
   }
 
   set value(value: string) {
-    this.foundation_.setValue(value);
+    // Replicate behavior of native select, which does not emit change events
+    // when the value is changed programmatically.
+    this.foundation_.setValue(value, false /* notifyChange */);
```
abhiomkar: Since `setValue` is used for programmatic change of select's value and I think we do not need notifyChange boolean. 

Only `handleChange()` triggers event, where as `setValue()` does not.

WDYT?

bonniezhou: Good point. Removed the boolean from `setValue` so that it's only in `handleChange`.

packages/mdc-select/foundation.ts
```diff
@@ -227,9 +230,8 @@ export class MDCSelectFoundation extends MDCFoundation<MDCSelectAdapter> {
   /**
    * Handles value changes, via change event or programmatic updates.
    */
-  handleChange() {
+  handleChange(notifyChange = true) {
```
abhiomkar: We can remove boolean from here as well and make handleChange always notify the change. 

We'll need to abstract out the code from `handleChange` to private method so it can be reused in other places (`setValue`, `setSelectedIndex`, etc). It is a good practice to bind `handle*` methods to only event callbacks in component implementation so we won't break other calls when we introduce any changes to `handleChange` callback in future. 

Currently `handleChange` is called in component implementation and also used has helper function within foundation.

abhiomkar:
Please let me know when this PR is ready for review.

packages/mdc-select/foundation.ts
bonniezhou: Updated. `setValue` and `setSelectedIndex` now both won't fire a change event since they are both programmatic changes.

PR #5256, opened 2019-11-14
feat(button): Setup elevation overlay

patrickrodee:


packages/mdc-button/_mixins.scss
```diff
@@ -386,9 +387,12 @@ $query: mdc-feature-all()) {
 
   @include mdc-typography(button, $query);
   @include mdc-button-horizontal-padding($mdc-button-horizontal-padding, $query);
+  @include mdc-elevation-overlay-parent($query: $query);
```
abhiomkar: Can we rename this to `mdc-elevation-overlay-position()`? seems more appropriate.

WDYT?

packages/mdc-button/_mixins.scss
```diff
@@ -386,9 +387,12 @@ $query: mdc-feature-all()) {
...
+  @include mdc-elevation-overlay-size(100%, $query: $query);
 
   @include mdc-feature-targets($feat-structure) {
     display: inline-flex;
+    /* @alternate */
     position: relative;
```
abhiomkar: Can we remove `position: relative` from base if elevation mixin is already setting it?

packages/mdc-button/_mixins.scss
patrickrodee: I named it `parent` because it should be applied to the parent element, not the overlay itself. What about `mdc-elevation-overlay-parent-position`?

packages/mdc-button/_mixins.scss
patrickrodee: We could. My only concern is it's not obvious from the mixin that the positioning is `relative`. I'm open to suggestions though and happy to remove the default `position: relative` here.

packages/mdc-button/_mixins.scss
abhiomkar: Gotcha. `mdc-elevation-overlay-parent-position` makes more sense in that case.

packages/mdc-button/_mixins.scss
abhiomkar: position relative is needed by both ripple & elevation. Do we need a separate mixin for parent position? We can document this instead.

Most of our components already have set to position relative.

WDYT?

patrickrodee: The separate mixin makes it clear which element will contain the elevation. Some elements, like the switch thumb or card, don't have position relative.

I'll remove the existing `position: relative;` from here and keep the mixin. Let me know if you think otherwise.

PR #5265, opened 2019-11-21
feat(chips): Remove event propagation stoppage

patrickrodee:


abhiomkar:
Can we close this if no longer required?

PR #5268, opened 2019-11-21
fix(button): Fix outline & ink color according to spec guidance

### LINKED ISSUE #5170: [Button] Outlined outline color should not be set to primary but 12% on-surface
## Bug report

The ~[Material spec for button theming](https://material.io/components/buttons/#theming)~ Material spec notes that the border color of an outlined button should be set to ~secondary theme~ on-surface at 12%, but [we seem to set it as primary](https://github.com/material-components/material-components-web/blob/master/packages/mdc-button/_mixins.scss#L109)

### Steps to reproduce
1. Create an mdc-button that is outlined
2. Set secondary to be a different color, like red

### Actual behavior

The outline of the button is the same, primary color

### Expected behavior

The outline of the button should be the secondary color (red in this case)

### Your Environment:

| MDC Web          | canary |
| Browser          | chrome |
| Operating System | macos  |

allan-chen:
Closes #5170 

Enabled outline color: primary -----> 12% on-surface.
Disabled outline color: 38% on-surface -----> 12% on-surface.
Disabled ink color: 37% ---> 38% on-surface.

BREAKING CHANGE: Variable `$mdc-button-disabled-container-fill-color`
renamed to `$mdc-button-disabled-container-color`.

### DIFF
--- a/packages/mdc-button/_mixins.scss
+++ b/packages/mdc-button/_mixins.scss
@@ -467,8 +467,8 @@ $query: mdc-feature-all()) {
   $feat-structure: mdc-feature-create-target($query, structure);
 
   @include mdc-button-outline-width($mdc-button-outlined-border-width, $query: $query);
-  @include mdc-button-outline-color(primary, $query);
-  @include mdc-button-disabled-outline-color($mdc-button-disabled-ink-color, $query);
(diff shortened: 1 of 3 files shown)


abhiomkar:
Assigning this to on-call for review.

PR #5285, opened 2019-11-27
Remove "demo-width-class"

una:
Demo width class was causing issues with the dropdown:

Kapture 2019-11-27 at 13 48 20


bonniezhou:
Where is this demo? `demo-width-class` is meant to be an optional class for users to add to set the width of the select.

asyncLiz:
Closing due to inactivity

PR #5287, opened 2019-11-28
feat(floating-label): add feature targeting for styles

crisbeto:
Adds support for feature targeting to the `mdc-floating-label` styles.

Relates to #4227.

### DIFF
--- a/packages/mdc-floating-label/_mixins.scss
+++ b/packages/mdc-floating-label/_mixins.scss
@@ -21,60 +21,149 @@
 //
 
 @import "@material/rtl/mixins";
+@import "@material/theme/variables";
 @import "@material/theme/mixins";
+@import "@material/feature-targeting/functions";
+@import "@material/feature-targeting/mixins";
+@import "@material/animation/variables";
+@import "@material/typography/mixins";
 @import "./variables";
 
-@mixin mdc-floating-label-ink-color($color) {
-  @include mdc-theme-prop(color, $color);
+@mixin mdc-floating-label-core-styles($query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+  $feat-animation: mdc-feature-create-target($query, animation);
+
+  // postcss-bem-linter: define floating-label
+  .mdc-floating-label {
+    @include mdc-typography(subtitle1, $query: $query, $exclude-props: (line-height));
+
+    @include mdc-feature-targets($feat-structure) {
+      position: absolute;
+      /* @noflip */
+      left: 0;
+      /* @noflip */
+      transform-origin: left top;
+      line-height: 1.15rem;
+      text-align: left;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+      cursor: text;
+      overflow: hidden;
+
+      /* @alternate */
+      // Force the label into its own layer to prevent visible layer promotion adjustments
+      // when the ripple is activated behind it.
+      will-change: transform;
+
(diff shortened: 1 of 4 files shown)


test/scss/_feature-targeting-test.scss
```diff
@@ -159,6 +160,15 @@
     @include mdc-elevation-overlay-fill-color(red, $query: $query);
     @include mdc-elevation-overlay-opacity(99%, $query: $query);
 
+    // Floating label
+    @include mdc-floating-label-core-styles($query: $query);
```
crisbeto: Note that the `test:feature-targeting` check doesn't seem to be running on the CI because there are failures in the files that's in master. I've added my changes here for completeness.

packages/mdc-floating-label/_mixins.scss
```diff
@@ -21,60 +21,148 @@
...
+@mixin mdc-floating-label-fill-color($color, $query: mdc-feature-all()) {
+  $feat-color: mdc-feature-create-target($query, color);
+
+  @include mdc-feature-targets($feat-color) {
+    @include mdc-theme-prop(background-color, $color);
+  }
+}
+
+@mixin mdc-floating-label-shake-keyframes($modifier, $positionY, $positionX: 0%, $scale: .75,
+$query: mdc-feature-all()) {
```
crisbeto: The indentation here is weird because the top line was over the limit so I had to break it up, but stylelint starts complaining if I add any indentation to it.

packages/mdc-floating-label/mdc-floating-label.scss
```diff
@@ -20,60 +20,9 @@
...
-    cursor: auto;
-  }
-
-  @at-root {
-    @include mdc-floating-label-float-position($mdc-floating-label-position-y);
-    @include mdc-floating-label-shake-animation(standard);
-  }
-
-  @include mdc-floating-label-shake-keyframes(standard, $mdc-floating-label-position-y);
+  @include mdc-floating-label-core-styles($query: mdc-feature-all());
```
mmalerba: nit: this is the default `$query`, can just not pass any params

crisbeto:
I've addressed the feedback @mmalerba and I've attached the before/after files. There are a few lines of difference, but they're within the same rule and shouldn't have an effect at runtime. I moved these lines down so that they could fit into an existing feature target.

[before.css.txt](https://github.com/material-components/material-components-web/files/3933489/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/3933491/after.css.txt)

packages/mdc-floating-label/_mixins.scss
```diff
@@ -21,60 +21,149 @@
...
+  $feat-structure: mdc-feature-create-target($query, structure);
+  $feat-animation: mdc-feature-create-target($query, animation);
 
-// Used for textarea in case of scrolling
-@mixin mdc-floating-label-fill-color($color) {
-  @include mdc-theme-prop(background-color, $color);
-}
+  // postcss-bem-linter: define floating-label
+  .mdc-floating-label {
+    @include mdc-typography(subtitle1, $query: $query);
```
abhiomkar: nit: Please use `$exclude-props` to exclude `line-height` in typography to avoid duplication with below CSS override.

BUILD will throw an error otherwise. :)

`/* @alternative */` comment before CSS override would also fix this.

crisbeto: Fixed.

PR #5293, opened 2019-11-30
chore: fix and run feature tageting checks

crisbeto:
The `test:feature-targeting` check wasn't running on the CI and was failing as a result. Furthermore, since the check wasn't running, there were some `mdc-fab` styles that weren't under a feature target which will cause them to always be emitted.

### DIFF
--- a/test/scss/_feature-targeting-test.scss
+++ b/test/scss/_feature-targeting-test.scss
@@ -154,8 +154,8 @@
     @include mdc-elevation-overlay-common($query: $query);
     @include mdc-elevation(0, $query: $query);
     @include mdc-elevation-shadow(none, $query: $query);
-    @include mdc-elevation-overlay-parent($query: $query);
     @include mdc-elevation-overlay-dimensions(100%, $query: $query);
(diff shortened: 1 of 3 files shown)


crisbeto:
Rebased @abhiomkar, although I didn't get any conflicts.

test/scss/_feature-targeting-test.scss
```diff
@@ -154,7 +154,7 @@
     @include mdc-elevation-overlay-common($query: $query);
     @include mdc-elevation(0, $query: $query);
     @include mdc-elevation-shadow(none, $query: $query);
-    @include mdc-elevation-overlay-parent($query: $query);
```
abhiomkar: This change seem unexpected.

crisbeto: From what I can tell these mixins were removed. The error I was getting is that the mixin doesn't exist and I couldn't find it in a code search.

crisbeto:
Rebased again @abhiomkar.

PR #5300, opened 2019-12-03
revert: fix(chips): Do not throw error if chip set becomes empty

abhiomkar:
Reverts material-components/material-components-web#5290


teddywing:
@abhiomkar @patrickrodee I just experienced this issue (opened in https://github.com/material-components/material-components-web/issues/5275) (example: https://jsfiddle.net/e872rdm6/). Would either of you be able to clarify why the fix was reverted?

Is there an alternative that we should be using to avoid this error?

Issue #4293, opened 2019-01-24
Double ripple on zoomed touch devices


patrickrodee:
When zoomed in on a touch device, tapping a ripple surface triggers two ripples.

[Here's a video showing it on iOS on Safari](https://drive.google.com/file/d/1sZ_QLrezzW2rK7VnbZnLpxEr_IyrSgbz/view)

acdvorak:
![double ripple omg](https://user-images.githubusercontent.com/409245/51867242-e249c100-22ff-11e9-985b-0d1270869bb0.gif)

PR #5305, opened 2019-12-04
fix(ripple): eliminate double ripple effect on iOS

peterminhk:
This PR fix the issue that ripple effect is activated twice on iOS when zoom in page.
see issue details: #4293

When tap a ripple surface on iOS Safari/Chrome below events are fired sequentially:
* pointerdown (activate ripple)
* touchstart (skipped)
* pointerup (deactivate ripple)
* mousedown (activate ripple again)

To solve this issue, we could find a solution to prevent mousedown after pointerup, but I had a doubt that why we had to handle mousedown event. There might be some sort of history, such as pointerdown was not supported on major browsers, but I couldn't figure out the actual reason (since this project was migrated from [material-design-lite](https://github.com/google/material-design-lite) and there are no history remains about this) and, at this time, it looks like mousedown can be replaced with pointerdown.

Therefore, I removed mousedown/mouseup from ACTIVATION/DEACTIVATION event types and it worked well on various devices, including:
* Safari / Chrome on iOS (13.1.3)
* Safari / Chrome on macOS (10.15.1)
* Chrome on Android (9)
* Edge / Chrome on Windows 10

### DIFF
--- a/packages/mdc-ripple/foundation.ts
+++ b/packages/mdc-ripple/foundation.ts
@@ -46,17 +46,17 @@ interface Coordinates {
   top: number;
 }
 
-type ActivationEventType = 'touchstart' | 'pointerdown' | 'mousedown' | 'keydown';
-type DeactivationEventType = 'touchend' | 'pointerup' | 'mouseup' | 'contextmenu';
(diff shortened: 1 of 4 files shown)


anlexN:
can we solve PR?

PR #5313, opened 2019-12-09
fix(text-field): change background colour opacity from 4% to 12%

### LINKED ISSUE #5035: [MDCTextField] Filled variant background color is incorrect according to spec
https://github.com/material-components/material-components-web/blob/805d098aa73810659ecf9fb4d2d701181955eb54/packages/mdc-textfield/_variables.scss#L40

Listed as 4%, but should be 12% according to design.

jlnjkb:
Resolves #5035


jlnjkb:
@googlebot I fixed it.

PR #5316, opened 2019-12-09
fix(checkbox): change checkbox event type from change to click and add some logic for IE browser

### LINKED ISSUE #4893: [Checkbox] Checkbox does not trigger change event when on indeterminate
### Steps to reproduce
1. Go to https://codepen.io/abhiomkar/pen/NZoJmo
2. Click on Checkbox which is in indeterminate state.

### Actual behavior

Notice that checkbox does not trigger change event (Which is indented behaviour of native checkbox input as per IE).

### Expected behavior

Checkbox should trigger change event when changed for indeterminate to checked / unchecked to be consistent with other browsers.

### Your Environment:

| MDC Web          | v3.0.0
| Browser          | IE11
| Operating System | Windows

### Additional context

See this response on https://stackoverflow.com/a/33529024 for more details on native input checkbox behaviour on IE.

candysonya:
Resolve #4893
Run some local test.


packages/mdc-checkbox/foundation.ts
```diff
@@ -48,12 +48,14 @@ export class MDCCheckboxFoundation extends MDCFoundation<MDCCheckboxAdapter> {
...
       removeClass: () => undefined,
       removeNativeControlAttr: () => undefined,
+      setChecked: () => undefined,
       setNativeControlAttr: () => undefined,
       setNativeControlDisabled: () => undefined,
     };
   }
 
   private currentCheckState_ = strings.TRANSITION_STATE_INIT;
+  private previousIsChecked_ = false;
```
bonniezhou: I think we should rename this to `realCheckedState_` or `inputCheckedState_`, since that will be a more accurate description of what this variable indicates

packages/mdc-checkbox/foundation.ts
```diff
@@ -98,7 +101,19 @@ export class MDCCheckboxFoundation extends MDCFoundation<MDCCheckboxAdapter> {
   }
 
   /**
-   * Handles the change event for the checkbox
+   * Handles the click event for the checkbox
+   * added for IE browser to fix compatibility issue b/144588184
```
bonniezhou: Remove this comment about IE and add it inside the handler (see my comment below). Also include the link to the bug on GitHub

packages/mdc-checkbox/foundation.ts
```diff
@@ -98,7 +101,19 @@ export class MDCCheckboxFoundation extends MDCFoundation<MDCCheckboxAdapter> {
...
+   */
+  handleClick() {
+    const {TRANSITION_STATE_INDETERMINATE} = strings;
```
bonniezhou: You can add the comment about the IE issue above this line

packages/mdc-checkbox/foundation.ts
```diff
@@ -98,7 +101,19 @@ export class MDCCheckboxFoundation extends MDCFoundation<MDCCheckboxAdapter> {
...
+    if (this.currentCheckState_ === TRANSITION_STATE_INDETERMINATE) {
+      this.adapter_.setChecked(!this.previousIsChecked_);
+    }
+    this.handleChange();
```
bonniezhou: It is best not to call other event handlers within event handlers. So here you can call `transitionCheckState_` directly

packages/mdc-checkbox/foundation.ts
```diff
@@ -48,12 +48,14 @@ export class MDCCheckboxFoundation extends MDCFoundation<MDCCheckboxAdapter> {
...
       removeClass: () => undefined,
       removeNativeControlAttr: () => undefined,
+      setChecked: () => undefined,
       setNativeControlAttr: () => undefined,
       setNativeControlDisabled: () => undefined,
     };
   }
 
   private currentCheckState_ = strings.TRANSITION_STATE_INIT;
+  private realCheckState_ = false;
```
bonniezhou: Can you add a comment above this line that provides more info, something like:
```
// Native checkboxes can only have two real states: checked/true or unchecked/false
// The indeterminate state is visual only.
```

bonniezhou: And add link to the stackoverflow question that explains this too: https://stackoverflow.com/a/33529024

abhiomkar:
@bonniezhou @candysonya Hi, is there a CL for this PR? Merging a PR without a CL would create discrepancies for other PRs until we automate this stuff.

If CL is not yet submitted please consider reverting this PR.

allan-chen:
Reverting this since it breaks MWC, should sync with internal first.

PR #5319, opened 2019-12-10
fix(select): Fix notch outline width when floating

e111077:
Need to call notch first to get width of label before it floats (and is consequentially shrunk). Same behavior as outlined textfield


allan-chen:
Next time please make sure to sync to internal first before merging the PR.

PR #5320, opened 2019-12-10
fix(select): make select focus themeable

### LINKED ISSUE #2360: MDC Select focus color is not themeable
0.32

Firefox/Chrome

Create a custom theme and click inside of a select.  While the focus is on the select, the underline will be themed correctly, but the select text will still be using `color:rgba(98,0,238,.87)` instead of the primary colors you selected.  It shouldn't matter but here is the theme I am applying in case it does:

```
#app {
	--mdc-theme-primary: #03A9F4;
	--mdc-theme-primary-light: #90CAF9;
	--mdc-theme-primary-dark: #1976D2;
	--mdc-theme-secondary: #5C6BC0;
	--mdc-theme-secondary-light: #9FA8DA;
	--mdc-theme-secondary-dark: #1A237E;
}
```

I am using mdc-vue-adapter 0.11.2, but I was able to trace the problem back to `material-components-web.min.css` so I don't think that should matter.  The offending line is: `.mdc-select:not(.mdc-select--disabled) .mdc-select__surface:focus .mdc-select__label{color:rgba(98,0,238,.87)}`

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #2360
- [x] bug
- [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
use `$mdc-select-focused-label-color: primary !default;` instead of `$mdc-select-focused-label-color: rgba(mdc-theme-prop-value(primary), .87) !default;`

##### Is there anything you'd like reviewers to focus on?
Replaced dropdown `svg` image with `css` to draw a triangle as its color need to match with custom primary color

### DIFF
--- a/packages/mdc-select/_mixins.scss
+++ b/packages/mdc-select/_mixins.scss
@@ -191,6 +191,7 @@
 
 @mixin mdc-select-dd-arrow-bg_($color, $opacity: 1) {
   @include mdc-theme-prop(border-top-color, $color);
+
   opacity: $opacity;
 }
 
--- a/packages/mdc-select/_mixins.scss
+++ b/packages/mdc-select/_mixins.scss
@@ -47,7 +47,7 @@
 @mixin mdc-select-focused-label-color($color) {
   &:not(.mdc-select--disabled) {
     &.mdc-select--focused .mdc-floating-label {
-      @include mdc-floating-label-ink-color(mdc-theme-prop-value($color));
+      @include mdc-floating-label-ink-color($color);
     }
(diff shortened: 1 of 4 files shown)


tanmoyopenroot:
@googlebot I signed it!

asyncLiz:
Thanks for the PR! We're going to do a bit of investigating to see how we want to tackle theming as a whole for all components.

asyncLiz:
Closing since select's focus state is now themable

PR #5321, opened 2019-12-10
test(checkbox): Remove checkbox Mocha foundation test.

joyzhong:


karma.conf.js
```diff
@@ -69,6 +69,7 @@ const istanbulInstrumenterLoader = {
     /node_modules/,
     /adapter.[jt]s$/,
     /constants.[jt]s$/,
+    /checkbox\/foundation.ts$/,
```
patrickrodee: Why is this excluded from Istanbul? Are we testing coverage via Jasmine?

karma.conf.js
```diff
@@ -135,6 +136,8 @@ const jasmineConfig = {
           functions: 50,
           lines: 80,
           excludes: [
+            'adapter.ts',
```
patrickrodee: Just to verify: this excludes the adapter and constants from coverage checks? (which seems good to me since they're not effectively testable)

karma.conf.js
joyzhong: Yep, since we now have the Jasmine foundation test for checkbox (with coverage for Jasmine tests: https://github.com/material-components/material-components-web/blob/50f110a6cf8100e594bdbd6c02ee278c39924008/karma.conf.js#L132), we can remove test coverage here.

karma.conf.js
joyzhong: Yes, this mirrors the original coverage checks for Mocha tests: https://github.com/material-components/material-components-web/blob/50f110a6cf8100e594bdbd6c02ee278c39924008/karma.conf.js#L70

karma.conf.js
patrickrodee: Dooooope

karma.conf.js
patrickrodee: 💯

PR #5330, opened 2019-12-12
test(infra): Fix `npm run test:unit` to exit with code 1 if Jasmine or Mocha tests fail.

joyzhong:
Currently, `npm run test:unit` will incorrectly exit with success code 0 even if the Jasmine tests fail. This fixes to exit with code 1 if either Jasmine or Mocha tests fail, while still running the Mocha tests even if Jasmine tests fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "test:dependency": "./scripts/dependency-test.sh",
     "test:feature-targeting": "node test/scss/verify-feature-targeting.js",
     "test:site": "npm run clean:site && ./scripts/site-generator-test.sh",
-    "test:unit": "npm run test:jasmineunit; npm run test:mochaunit",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "test:dependency": "./scripts/dependency-test.sh",
     "test:feature-targeting": "node test/scss/verify-feature-targeting.js",
     "test:site": "npm run clean:site && ./scripts/site-generator-test.sh",
-    "test:unit": "npm run test:jasmineunit; npm run test:mochaunit",
+    "test:unit": "npm run test:jasmineunit; jasmine_exit_code=$?; npm run test:mochaunit && exit \"$jasmine_exit_code\"",
```
patrickrodee: What will happen if the Mocha tests fail? Will it exit with code 1?

joyzhong: Yep, since exit "$jasmine_exit_code" will only execute if `npm run test:mochaunit` succeeds.

patrickrodee: Cool cool, thanks!

Issue #5220, opened 2019-11-01
Migrate to use the Sass module system


jathak:
As discussed offline with @abhiomkar and @aprigogin, we'd like to migrate the stylesheets in this repo to use the [Sass module system](http://sass.logdown.com/posts/7858341-the-module-system-is-launched).

I'm working on preparing a script that uses the [Sass migrator](https://sass-lang.com/documentation/cli/migrator) to handle most of this migration, but I have a question about how to handle member prefixes.

Right now, most variables, mixins, and functions are manually namespaced with `mdc-<component>-`.  Since the module system adds automatic namespacing, it makes sense to remove these manual prefixes (which the migrator can handle automatically).

However, there are two cases where this pattern doesn't quite hold:

- The colors in [`mdc-theme/_color-palette.scss`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-theme/_color-palette.scss) are instead prefixed with `material-color-`. For these variables, should we just leave their names as-is?
- Some components have a mixin just named `mdc-<component>` (so removing the prefix would leave an empty name). How should these be handled?
Examples (there might be more though):
  - [`mdc-elevation`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-elevation/_mixins.scss#L53)
  - [`mdc-layout-grid`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-layout-grid/_mixins.scss#L96)
  - [`mdc-rtl`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-rtl/_mixins.scss#L71)
  - [`mdc-typography`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-typography/_mixins.scss#L49)

Note that however we choose to handle namespacing during this migration, we'll add [import-only files](https://sass-lang.com/documentation/at-rules/import#import-only-files) that forward the members with their original, manually-namespaced names. When downstream users migrate, the migrator will be able to detect this and automatically change these names when necessary.

abhiomkar:
Thanks @jathak for the update!

* We can leave color palette names as-is for import path by forwarding with `material-color-*` prefix.
* This is interesting case, what options do we have? I wish `@forward` rule supports forwarding per mixin or variable name instead of file name.

Here is what I'm thinking:

#### Option 1

```sass
// rtl/_base.scss (Sass module)
@mixin rtl() {...}

// rtl/_mixins.scss (Sass module)
@mixin reflexive-box() {...}
@forward "base"

[...]

PR #5337, opened 2019-12-13
fix: add missing dependencies

jathak:
Add missing dependencies to five files that previously didn't depend on all the stylesheets they reference (even indirectly). This caused issues with the module system migrator (#5220), so this fix is necessary for that.


jathak:
Done

PR #5352, opened 2019-12-18
feat(button): Add overflow ellipsis mixin

patrickrodee:


packages/mdc-button/_mixins.scss
```diff
@@ -381,6 +381,15 @@ $query: mdc-feature-all()) {
   }
 }
 
+///
+/// Sets the button label to overflow as ellipsis
+///
+@mixin mdc-button-label-overlow-ellipsis($query: mdc-feature-all()) {
```
abhiomkar: When do we use this mixin? are there any special cases? Can we apply this mixin to MDC button by default?

patrickrodee: This was requested by an internal team.

PR #5355, opened 2019-12-18
fix(textfield): add placeholder mixins and disabled placeholder color

asyncLiz:


packages/mdc-textfield/_mixins.scss
```diff
@@ -233,7 +253,7 @@
 
 ///
 /// Customizes the background color of the text field or textarea when disabled.
-/// @param {Color} $color - The desired label text color.
+/// @param {Color} $color - The desired background color.
```
allan-chen: good catch

packages/mdc-textfield/_mixins.scss
```diff
@@ -816,6 +837,18 @@
   }
 }
 
+@mixin mdc-text-field-placeholder-color_($color) {
+  .mdc-text-field__input::placeholder {
+    color: $color;
```
allan-chen: Can we use `mdc-theme-prop(color, $color)`?

packages/mdc-textfield/_mixins.scss
```diff
@@ -816,6 +837,18 @@
...
+  }
+
+  // Override the placeholder styles in IE with important rule to improve specificity.
+  // stylelint-disable-next-line selector-no-vendor-prefix
+  .mdc-text-field__input:-ms-input-placeholder {
+    color: $color !important;
```
allan-chen: Almost same as above, except can we use `mdc-theme-prop(color, $color, $important: true)`?

test/screenshot/spec/mdc-textfield/mixins/disabled-3.html
```diff
@@ -0,0 +1,79 @@
+<!DOCTYPE html>
```
allan-chen: Can this fit inside `disabled-2.html`? Would like to reduce the number of screenshots if possible

asyncLiz: I think it might be able to, but I know we'll have to have another one for all the disabled color customization tests with prefix/suffix, so I thought it might be good to go ahead and start a new one.

Want me to try to fit into 2 still?

allan-chen: Let's try to squeeze it in for now, prefix and suffix can be in their own thing when the time comes

asyncLiz: Hm they kinda fit, but they cause scrollbars to appear. Is this acceptable?

https://storage.googleapis.com/mdc-web-screenshot-tests/lizmitchell/2019/12/18/21_04_39_522/spec/mdc-textfield/mixins/disabled-2.html.windows_chrome_77.png

allan-chen: Yeah, should be fine

Issue #5202, opened 2019-10-25
[mdc-switch] add required aria-checked 


kvdheuvel:
## Feature Request
The current mdc-switch component uses `role='switch'` but not the required aria attribute `aria-checked` as described by MDN:

> The aria-checked attribute is required when using the switch role, as it represents the current state of the widget that the switch role is applied to. A value of true represents the "on" state; false represents the "off" state; a value of mixed is not supported by the switch role, and is treated as false. The default value is false.

### Proposed solution
add `aria-checked` to the mdc-switch component

### Alternatives considered
/

### Additional context
Noticed this issue while searching for accessibility issues on our current projects. Everything else works as expected

allan-chen:
Hi,

Thank you for your suggestion. I'm assuming you are referring to updating the README for switch to include `aria-checked` in the listed templates, since the aria-checked attribute is already present in the demo? (https://material-components.github.io/material-components-web-catalog/#/component/switch)

Please feel free to submit a pull request to update the docs accordingly!

kvdheuvel:
Hey,

Yes, i'm talking about the README files. I (and i guess most people) use the codes snippets presented in the README docs.

I'll go read the pull-request documentation and give it a try.

allan-chen:
Looks like the logic is missing from the component itself - the catalog's react component wrapping hides this issue.

PR #5357, opened 2019-12-18
fix(switch): handle aria-checked correctly. (#5202)

Yavanosta:
Handle aria-checked attribute on MDC Web level, so developers should not implement this logic.

Aria checked is [required](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role) for `role="switch"` elements.

In addition after this fix [MDC Web catalog website](https://github.com/material-components/material-components-web-catalog/blob/master/src/SwitchCatalog.js#L53) should be updated.

### DIFF
--- a/packages/mdc-switch/README.md
+++ b/packages/mdc-switch/README.md
@@ -43,7 +43,7 @@ npm install @material/switch
   <div class="mdc-switch__track"></div>
   <div class="mdc-switch__thumb-underlay">
     <div class="mdc-switch__thumb"></div>
-    <input type="checkbox" id="basic-switch" class="mdc-switch__native-control" role="switch">
+    <input type="checkbox" id="basic-switch" class="mdc-switch__native-control" role="switch" aria-checked="false">
   </div>
 </div>
 <label for="basic-switch">off/on</label>
@@ -78,7 +78,7 @@ Add the `mdc-switch--disabled` class to the `mdc-switch` element, and the `disab
   <div class="mdc-switch__track"></div>
   <div class="mdc-switch__thumb-underlay">
     <div class="mdc-switch__thumb"></div>
-    <input type="checkbox" id="another-basic-switch" class="mdc-switch__native-control" role="switch" disabled>
+    <input type="checkbox" id="another-basic-switch" class="mdc-switch__native-control" role="switch" aria-checked="false" disabled>
   </div>
 </div>
 <label for="another-basic-switch">off/on</label>
@@ -93,7 +93,7 @@ Add the `mdc-switch--checked` class to the `mdc-switch` element, and the `checke
   <div class="mdc-switch__track"></div>
   <div class="mdc-switch__thumb-underlay">
     <div class="mdc-switch__thumb"></div>
-    <input type="checkbox" id="another-basic-switch" class="mdc-switch__native-control" role="switch" checked>
+    <input type="checkbox" id="another-basic-switch" class="mdc-switch__native-control" role="switch" aria-checked="false" checked>
   </div>
 </div>
 <label for="another-basic-switch">off/on</label>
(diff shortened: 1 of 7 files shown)


packages/mdc-switch/README.md
```diff
@@ -93,7 +93,7 @@ Add the `mdc-switch--checked` class to the `mdc-switch` element, and the `checke
   <div class="mdc-switch__track"></div>
   <div class="mdc-switch__thumb-underlay">
     <div class="mdc-switch__thumb"></div>
-    <input type="checkbox" id="another-basic-switch" class="mdc-switch__native-control" role="switch" checked>
+    <input type="checkbox" id="another-basic-switch" class="mdc-switch__native-control" role="switch" aria-checked="false" checked>
```
allan-chen: Should be aria-checked="true" since switch is initially checked

packages/mdc-switch/foundation.ts
```diff
@@ -80,6 +83,14 @@ export class MDCSwitchFoundation extends MDCFoundation<MDCSwitchAdapter> {
       this.adapter_.removeClass(cssClasses.CHECKED);
     }
   }
+
+  private updateAriaChecked_(checked: boolean) {
+    this.adapter_.setNativeControlAttr(
+        strings.ARIA_CHECKED_ATTR,
+        checked ?
```
allan-chen: Can we simply use `` `${checked}` `` here and remove the constants?

test/unit/mdc-switch/foundation.test.js
```diff
@@ -106,3 +128,21 @@ test('#handleChange removes mdc-switch--checked from the switch when it is an un
   foundation.handleChange({target: {checked: false}});
   td.verify(mockAdapter.removeClass(MDCSwitchFoundation.cssClasses.CHECKED));
 });
+
+test('#handleChange sets aria-checked to true when the swith is a checked state', () => {
```
allan-chen: nit: typo

test/unit/mdc-switch/foundation.test.js
```diff
@@ -106,3 +128,21 @@ test('#handleChange removes mdc-switch--checked from the switch when it is an un
...
+  const {foundation, mockAdapter} = setupTest();
+
+  foundation.handleChange({target: {checked: true}});
+  td.verify(mockAdapter.setNativeControlAttr(
+    MDCSwitchFoundation.strings.ARIA_CHECKED_ATTR,
+    MDCSwitchFoundation.strings.ARIA_CHECKED_CHECKED_VALUE));
+});
+
+test('#handleChange sets aria-checked to false when the swith is a checked state', () => {
```
allan-chen: nit: typo, also unchecked

packages/mdc-switch/README.md
Yavanosta: Thank you Alan, sorry for inadvertency.

Fixed.

test/unit/mdc-switch/foundation.test.js
Yavanosta: Fixed `swith` -> `switch`, `checked` -> `unchecked`.

test/unit/mdc-switch/foundation.test.js
Yavanosta: Thank you!

Fixed `swith` -> `switch`.

packages/mdc-switch/foundation.ts
Yavanosta: I just thought that in case if `checked` will be some false-like value (if used from JavaScript for example) that `aria-checked` can become invalid. But you are right - using constants looks like overkill too.

Maybe it will be better to use `${!!checked}`? Or just `${checked}` is enough?

allan-chen: Good point! `${!!checked}` looks good

packages/mdc-switch/adapter.ts
```diff
@@ -48,4 +48,9 @@ export interface MDCSwitchAdapter {
    * Sets the disabled state of the native HTML control underlying the switch.
    */
   setNativeControlDisabled(disabled: boolean): void;
+
+  /**
+   * Set an attribute value of the native HTML control underlying the switch.
```
abhiomkar: Make sure you sync-in all the changes from your CL. :)

Yavanosta: Yes, I remember that MDC Web is not synced yet, so added commit with typo fix. Thank you!

Yavanosta: Added to PR description and to [commit](https://github.com/material-components/material-components-web/pull/5357/commits/57100507c6372332ed8417daaae30925e160e97c).

PR #5366, opened 2019-12-19
feat(line-ripple): add active/inactive states to line-ripple

asyncLiz:
BREAKING CHANGE: `mdc-line-ripple-color()` mixin has been renamed to `mdc-line-ripple-active-color()`

PR #5367, opened 2019-12-19
fix(textfield): add separate classes for leading/trailing icons

asyncLiz:
BREAKING_CHANGE: icons must use `.mdc-text-field__leading-icon` or `.mdc-text-field__trailing-icon` classes. `mdc-text-field-icon-color()` mixin has been split into `mdc-text-field-leading-icon-color()` and `mdc-text-field-trailing-icon-color()`.

### DIFF
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -602,10 +602,13 @@
 }
 
 @mixin mdc-text-field-hover-outline-color_($color) {
-  &:not(.mdc-text-field--focused):hover {
-    .mdc-notched-outline {
-      @include mdc-notched-outline-color($color);
-    }
+  &:not(.mdc-text-field--focused) {
+    // stylelint-disable-next-line selector-combinator-space-after
+    .mdc-text-field__input:hover ~,
+    .mdc-text-field__icon:hover ~ {
+      .mdc-notched-outline {
+        @include mdc-notched-outline-color($color);
+      }
   }
 }
 
--- a/packages/mdc-textfield/_mixins.scss
+++ b/packages/mdc-textfield/_mixins.scss
@@ -565,8 +565,7 @@
     top: 14px;
   }
 
-  .mdc-text-field__leading-icon,
-  .mdc-text-field__trailing-icon {
+  .mdc-text-field__icon {
     top: 12px;
   }
 }
@@ -593,8 +592,7 @@
     z-index: 1;
   }
 
-  .mdc-text-field__leading-icon,
-  .mdc-text-field__trailing-icon {
+  .mdc-text-field__icon {
     z-index: 2;
   }
 
@@ -621,12 +619,13 @@
 
 @mixin mdc-text-field-with-leading-icon_ {
   @include mdc-text-field-icon-horizontal-position_(
-    leading,
     left,
-    $mdc-text-field-icon-position,
-    $mdc-text-field-icon-padding,
-    $mdc-text-field-input-padding
+    $mdc-text-field-icon-position
   );
+
+  .mdc-text-field__input {
+    @include mdc-rtl-reflexive-property(padding, $mdc-text-field-icon-padding, $mdc-text-field-input-padding);
+  }
 
   .mdc-floating-label {
     @include mdc-rtl-reflexive-position(left, $mdc-text-field-icon-padding);
@@ -635,12 +634,13 @@
 
 @mixin mdc-text-field-dense-with-leading-icon_ {
   @include mdc-text-field-icon-horizontal-position_(
-    leading,
     left,
-    $mdc-text-field-dense-icon-position,
-    $mdc-text-field-dense-icon-padding,
-    $mdc-text-field-input-padding
+    $mdc-text-field-dense-icon-position
   );
+
+  .mdc-text-field__input {
+    @include mdc-rtl-reflexive-property(padding, $mdc-text-field-dense-icon-padding, $mdc-text-field-input-padding);
+  }
 
   .mdc-floating-label {
     @include mdc-rtl-reflexive-position(left, $mdc-text-field-dense-icon-padding);
@@ -649,11 +649,8 @@
 
 @mixin mdc-text-field-outlined-with-leading-icon_ {
   @include mdc-text-field-icon-horizontal-position_(
-    leading,
     left,
-    $mdc-text-field-icon-position,
-    $mdc-text-field-icon-padding,
-    $mdc-text-field-input-padding
+    $mdc-text-field-icon-position
   );
   @include mdc-notched-outline-floating-label-float-position-absolute($mdc-text-field-outlined-label-position-y, 32px);
   @include mdc-floating-label-shake-animation(text-field-outlined-leading-icon);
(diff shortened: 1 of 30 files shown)


packages/mdc-textfield/component.ts
```diff
@@ -113,21 +113,13 @@ export class MDCTextField extends MDCComponent<MDCTextFieldFoundation> implement
...
-        this.trailingIcon_ = iconFactory(iconElements[1]);
-      } else {
-        if (this.root_.classList.contains(cssClasses.WITH_LEADING_ICON)) {
-          this.leadingIcon_ = iconFactory(iconElements[0]);
-        } else {
-          this.trailingIcon_ = iconFactory(iconElements[0]);
-        }
-      }
-    }
+    // Leading icon
```
patrickrodee: Wow... This is a stark improvement!

packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -42,14 +52,21 @@
...
-@mixin mdc-text-field-icon-horizontal-position_($position-property, $position, $padding, $input-padding) {
-  .mdc-text-field__icon {
+@mixin mdc-text-field-icon-horizontal-position_(
+  $icon-position,
+  $position-property,
+  $position,
+  $padding,
+  $input-padding
+) {
+  .mdc-text-field__#{$icon-position}-icon {
```
patrickrodee: Let's break the contents of this into two separate mixins: one for the icon positioning and one for the input positioning. We can drop the selectors inside each and then call them with the relevant selectors above.

For example, this mixin: https://github.com/material-components/material-components-web/blob/2167ee095b23315e90b4640f5edeb2c7936b438c/packages/mdc-textfield/_mixins.scss#L601-L613

would instead become

```
@mixin mdc-text-field-with-leading-icon_ {
  .mdc-text-field__icon--leading {
    @include mdc-text-field-icon-position-or-other-name_(...args go here);
  }

  .mdc-text-field__input {
    @include mdc-text-field-input-position-or-other-name_(...args go here);
  }

  .mdc-floating-label {
    @include mdc-rtl-reflexive-position(left, $mdc-text-field-icon-padding); 
  }
}
```

It's now much clearer what's being affected and we can make additional changes here without having to complicate the child mixins.

asyncLiz:
> Overall comment: let's leave the `mdc-text-field__icon` element class on the icons but add the `mdc-text-field__icon--trailing` or `mdc-text-field__icon--leading` modifier classes. It'll allow us to keep generalized behavior but also apply variant-specific behavior where necessary.

@patrickrodee does that make sense for the long term though? A lot of the shared styles right now have to deal with absolute positioning, which will go away. [In the end](https://github.com/material-components/material-components-web/compare/feat/textfield/prefix-suffix) the only shared styles will be for the [cursor](https://github.com/material-components/material-components-web/compare/feat/textfield/prefix-suffix#diff-f7c44d2746e4c925cd75a8940cae639dR71-R79) and [scaling with density](https://github.com/material-components/material-components-web/compare/feat/textfield/prefix-suffix#diff-4fb72464ee61cdee22d068c1791c191bR169-R173).

It's still a breaking change since we need the leading/trailing modifiers, and it feels like an unnecessary class to me at that point.

packages/mdc-textfield/icon/_mixins.scss
asyncLiz: I could do that work, but the end goal is that this mixin [will be completely removed](https://github.com/material-components/material-components-web/compare/feat/textfield/prefix-suffix#diff-f7c44d2746e4c925cd75a8940cae639dL49-L63) once we get rid of absolute positioning. Is it worth the effort to organize this mixin if we're just going to delete it shortly?

patrickrodee:
@asyncLiz Yeah, let's keep the existing `mdc-text-field__icon` class and add the `mdc-text-field__icon--leading/trailing` classes. Having an element class on both is good future proofing and better follows [BEM](http://getbem.com/).

packages/mdc-textfield/icon/_mixins.scss
patrickrodee: Yeah, let's get it cleaned up. If for some reason we aren't able to get the `<label>` root element change in, we'll at least have improvements to the existing Sass.

asyncLiz:
Would we want to do something similar for prefix/suffix?

```html
<span class="mdc-text-field__affix mdc-text-field__affix-prefix">$</span>
```

patrickrodee:
@asyncLiz Yeah, something similar with prefix/suffix would be a good idea.

packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -42,58 +52,56 @@
...
 
 // Private mixins
 
-@mixin mdc-text-field-icon-horizontal-position_($position-property, $position, $padding, $input-padding) {
-  .mdc-text-field__icon {
-    @include mdc-rtl-reflexive-position($position-property, $position);
-  }
-
-  // Move the input's position, to allow room for the icon
+@mixin mdc-text-field-icon-horizontal-position_($position-property, $position) {
```
patrickrodee: This looks great! Let's simplify it more by creating two separate mxins, `mdc-text-field-leading-icon-horizontal-position_` and `mdc-text-field-trailing-icon-horizontal-position_` and using those mixins directly. It'll make it clearer what each mixin is doing.

packages/mdc-textfield/_mixins.scss
```diff
@@ -589,13 +581,9 @@
...
-  &:not(.mdc-text-field--focused) {
-    // stylelint-disable-next-line selector-combinator-space-after
-    .mdc-text-field__input:hover ~,
-    .mdc-text-field__icon:hover ~ {
-      .mdc-notched-outline {
-        @include mdc-notched-outline-color($color);
-      }
+  &:not(.mdc-text-field--focused):hover {
+    .mdc-notched-outline {
+      @include mdc-notched-outline-color($color);
```
patrickrodee: Why did these lines change?

packages/mdc-textfield/_mixins.scss
```diff
@@ -379,16 +380,7 @@
...
-  &.mdc-text-field--with-trailing-icon {
-    &:not(.mdc-text-field--with-leading-icon) {
-      @include mdc-text-field-icon-color($mdc-text-field-error);
-    }
-
-    &.mdc-text-field--with-leading-icon {
-      @include mdc-text-field-icon-color($mdc-text-field-error, /* styleSecondIcon */ true);
-    }
-  }
+  @include mdc-text-field-trailing-icon-color($mdc-text-field-error);
```
patrickrodee: Drastic improvement!

packages/mdc-textfield/icon/mdc-text-field-icon.scss
```diff
@@ -20,19 +20,8 @@
...
-  top: 50%;
-  transform: translateY(-50%);
-  cursor: pointer;
-}
-
-.mdc-text-field__icon:not([tabindex]),
-.mdc-text-field__icon[tabindex="-1"] {
-  cursor: default;
-  pointer-events: none;
+.mdc-text-field__icon {
```
patrickrodee: Another stark improvement!

packages/mdc-textfield/_mixins.scss
asyncLiz: I can revert it, it's being fixed in the [line ripple changes](https://github.com/material-components/material-components-web/pull/5366/files#diff-602886a88f66102525d0b2eb6eed3479L591). I think I originally fixed it here too because my first draft removed the `__icon` class entirely.

packages/mdc-textfield/icon/_mixins.scss
patrickrodee: @asyncLiz This should be the last outstanding fix.

asyncLiz: Woops, missed this, thanks for the reminder

packages/mdc-textfield/icon/_mixins.scss
```diff
@@ -59,25 +59,15 @@
...
-      @include mdc-rtl-reflexive-position($position-property, $position);
-    }
-  } @else {
-    .mdc-text-field__icon--trailing {
-      @include mdc-rtl-reflexive-position($position-property, $position);
-    }
-  }
-}
-
-@mixin mdc-text-field-icon-horizontal-position-two-icons_($position-left, $position-right) {
```
patrickrodee: Why was this removed?

asyncLiz: It's doing the exact same thing as the other two mixins that set the leading/trailing positions.

It was necessary before because we didn't have separate classes for each one. It was juggling setting left vs right on `.mdc-text-field__icon` depending on if there was `with-leading`, `with-trailing`, or both on the host element.

asyncLiz: Looks like I might need to add some logic back in to fix the screenshot tests to the correct state. It's using the wrong position for the trailing icon that we have b/146057700 for

abhiomkar:
@asyncLiz nit: `BREAKING_CHANGE` => `BREAKING CHANGE` when your squash & merge in commit description to show up on CHANGELOG.

Issue #5349, opened 2019-12-17
[Custom Properties] Allow for more custom property-based theming


una:
### Steps to reproduce
1. Go to https://material-theme-builder.glitch.me
2. Try to theme with custom properties

### Actual behavior
The following components do not get accurate color styles:

- [ ] Data tables (selected backgrounds)
- [ ] Active list items
- [ ] Selected List items
- [ ] Menu bar header
- [ ] Active Input labels

None of the components are getting the correct shape styling or typography styling via custom properties.

### Additional context
See also this bandaids file I made when building a custom-property based Material demo: https://glitch.com/edit/#!/material-theme-generator?path=public/bandaids.css:1:18

### Possible solution

The following components fo not get on-color styles. A potential fix:

```
.mdc-card .mdc-typography,
.mdc-menu .mdc-list,
.mdc-dialog .mdc-dialog__title,
.mdc-dialog .mdc-dialog__content,
.mdc-chip__text,
.mdc-chip--selected .mdc-chip__checkmark-path {
  color: var(--mdc-theme-on-surface);
}

.mdc-top-app-bar {
  color: var(--mdc-theme-on-primary);
}
```

The following components do not get shape styles. A potential fix:

```
.mdc-button,
.mdc-text-field {
  border-radius: var(--mdc-shape-small);
}  

.mdc-text-field--outlined .mdc-notched-outline .mdc-notched-outline__trailing,
.mdc-select--outlined .mdc-notched-outline .mdc-notched-outline__trailing {
  border-radius: 0 var(--mdc-shape-small) var(--mdc-shape-small) 0;
}

.mdc-text-field--outlined .mdc-notched-outline .mdc-notched-outline__leading,
.mdc-select--outlined .mdc-notched-outline .mdc-notched-outline__leading {
  border-radius: var(--mdc-shape-small) 0 0 var(--mdc-shape-small);
}

.mdc-select, .mdc-select__native-control {
  border-radius: var(--mdc-shape-small) var(--mdc-shape-small) 0 0;
}

.mdc-card,
.mdc-dialog .mdc-dialog__surface {
  border-radius: var(--mdc-shape-medium);
}

.mdc-drawer {
  border-radius: 0 var(--mdc-shape-large) var(--mdc-shape-large) 0;
}

.mdc-menu {
  border-radius: var(--mdc-shape-medium);
}
```

The following components do not get typography styles. A potential fix:

```
.mdc-typography--headline1,
.mdc-typography--headline2,
[...]

PR #5375, opened 2019-12-20
feat: Enable custom property styling for base typography

una:
See #5349

This would result in an output of: 

`font-family: var(--mdc-typography-font-family), Roboto, sans-serif;` instead of `font-family: Roboto, sans-serif;` for each base instance, enabling overrides with custom properties. Unsupported browsers will ignore the first font declaration for the second.

This would enable base styling overrides with custom properties:

```
root: {
  --mdc-typography-font-family: "fontName", serif;
}
```

And per-class overrides with custom properties:

```
.mdc-typography--headline2 {
  --mdc-typography-font-family: "myOtherFont", serif;
}
```

### DIFF
--- a/packages/mdc-typography/_variables.scss
+++ b/packages/mdc-typography/_variables.scss
@@ -22,7 +22,11 @@
 
 @import "./functions";
 
-$mdc-typography-font-family: unquote("Roboto, sans-serif") !default;
+:root {
+  --mdc-typography-font-family: 'Roboto', sans-serif;
+}
+
+$mdc-typography-font-family: var(--mdc-typography-font-family), unquote("Roboto, sans-serif") !default;
 
 $mdc-typography-base: (
   font-family: $mdc-typography-font-family,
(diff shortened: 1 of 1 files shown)


packages/mdc-typography/_variables.scss
```diff
@@ -22,7 +22,11 @@
 
 @import "./functions";
 
-$mdc-typography-font-family: unquote("Roboto, sans-serif") !default;
+:root {
+  --mdc-typography-font-family: "Roboto", sans-serif;
+}
+
+$mdc-typography-font-family: var(--mdc-typography-font-family), unquote("Roboto, sans-serif") !default;
```
abhiomkar: Would it make sense to apply this CSS custom property in [mdc-typography-base()](https://github.com/material-components/material-components-web/blob/d6ff712eaa4be043125c34083153c4c73e65cd4c/packages/mdc-typography/_mixins.scss#L39) mixin?

abhiomkar:
Assigning this to @asyncLiz since she is working on custom property support for Typography.

asyncLiz:
Closing in favor of https://github.com/material-components/material-components-web/pull/5618

PR #5378, opened 2019-12-22
feat(text-field): add feature targeting for styles

crisbeto:
Sets up feature targeting for the text field and its related styles.

Relates to #4227.

cc @mmalerba @devversion


crisbeto:
Here's the before/after output:
[before.css.txt](https://github.com/material-components/material-components-web/files/3992401/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/3992402/after.css.txt)

You'll notice that there are some differences. They come from the fact that I had to split the ripple styles into a separate mixin so that we can opt out of them for Angular Material. A side-effect of the split is that I also had to split away the outlined styles, because they need to be included after the ripple styles, because they have some overrides that hide the ripple. I've gone through all the examples in the screenshot test app and these changes don't seem to have an effect on the appearance.

packages/mdc-textfield/_mixins.scss
```diff
@@ -20,35 +20,411 @@
...
+
+  // This needs to be included separately from `mdc-text-field-without-ripple` and it needs to come
+  // after `mdc-text-field-ripple` so that it can override the ripple styles and remove it.
+  @include mdc-text-field-outlined($query: $query);
+}
+
+@mixin mdc-text-field-without-ripple($query: mdc-feature-all()) {
+  $feat-structure: mdc-feature-create-target($query, structure);
+
+  @include mdc-floating-label-core-styles($query: $query);
```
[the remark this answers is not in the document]
crisbeto: Sure, I can switch those styles back to be added through `@import`. The only downside is that consumers of the mixin would have to know to include the styles themselves.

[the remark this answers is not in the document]
abhiomkar: Most of MDC Web style dependencies are structured this way. Changing style dependency is debatable. I prefer not to make these changes in same PR.

We can add documentation about style dependency with above sub-components for now and come back to this later.

WDYT?

abhiomkar:
PR #5367 is merged. Can you please rebase with master, I'll test this PR internally today.

Thanks!

packages/mdc-textfield/_mixins.scss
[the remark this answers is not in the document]
crisbeto: Done.

crisbeto:
@abhiomkar I've rebased it and reworked it based on the feedback. Here are the updated before/after files.

[before.css.txt](https://github.com/material-components/material-components-web/files/4047928/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/4047929/after.css.txt)

packages/mdc-textfield/_mixins.scss
```diff
@@ -528,350 +950,468 @@
...
-@mixin mdc-text-field-fullwidth-invalid_ {
-  @include mdc-text-field-bottom-line-color($mdc-text-field-error);
+@mixin mdc-text-field-fullwidth-invalid_($query: mdc-feature-all()) {
+  @include mdc-text-field-bottom-line-color($mdc-text-field-error, $query: $query);
 }
 
 // Textarea
-@mixin mdc-text-field-textarea-disabled_ {
-  @include mdc-text-field-outlined-disabled_;
-  /* @alternate */
```
abhiomkar: Revert `/* @alternate */`. This is required to prevent failure with duplicate CSS property check.

packages/mdc-textfield/_mixins.scss
```diff
@@ -20,35 +20,398 @@
...
+    $mdc-text-field-textarea-label-position-y,
+    0%,
+    $query: $query
+  );
+}
+
+// This API is intended for use by frameworks that may want to separate the ripple-related styles
+// from the other text field styles. It is recommended that most users use `mdc-text-field-core-styles` instead.
+@mixin mdc-text-field-ripple($query: mdc-feature-all()) {
+  @include mdc-ripple-common($query);
```
abhiomkar: This mixin wasn't included before. Remove this.

packages/mdc-textfield/_mixins.scss
```diff
@@ -20,35 +20,398 @@
...
 @import "helper-text/mixins";
 @import "character-counter/mixins";
 @import "icon/mixins";
 @import "icon/variables";
 @import "./variables";
 @import "./functions";
 
+@mixin mdc-text-field-core-styles($query: mdc-feature-all()) {
+  @include mdc-text-field-without-ripple($query);
+  @include mdc-text-field-ripple($query);
```
abhiomkar: Include `mdc-text-field-ripple` before `mdc-text-field-without-ripple`. 

This order breaks textarea's ripple otherwise.

image

packages/mdc-textfield/_mixins.scss
crisbeto: Wasn't the ripple previously included via an import? I think that this is what we did for the other components where we separated the ripple styles from everything else.

abhiomkar: Can we `@import` ripple/common in mdc-text-field file instead? (Similar to floating label, line ripple etc.)

It is always safer to retain the CSS ordering to avoid any further breaking changes.

crisbeto: It goes against the pattern we've set until now though. Here are a few examples:

https://github.com/material-components/material-components-web/blob/master/packages/mdc-button/_mixins.scss#L138
https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/_mixins.scss
https://github.com/material-components/material-components-web/blob/master/packages/mdc-radio/_mixins.scss

If you still think that it's okay to diverge in this case it's fine by me.

packages/mdc-textfield/_mixins.scss
```diff
@@ -20,35 +20,398 @@
...
+
+  // This needs to be included separately from `mdc-text-field-without-ripple` and it needs to come
+  // after `mdc-text-field-ripple` so that it can override the ripple styles and remove it.
+  @include mdc-text-field-outlined($query: $query);
```
abhiomkar: After re-ordering `-ripple()` and `-without-ripple()` mixins, I moved `mdc-text-field-outlined()` mixin to its original location. It fixes the issue that I described in [this comment](https://github.com/material-components/material-components-web/pull/5378#pullrequestreview-341461733).

Can we move this mixin after `.mdc-text-field__input:-webkit-autofill + .mdc-floating-label` selector block? This'll retain the CSS ordering for `__input` padding.

packages/mdc-textfield/_mixins.scss
abhiomkar: hmm.. Can you add `// COPYBARA_COMMENT_THIS_LINE` comment next to the -common mixin then? It'll create duplicate styles internally otherwise.

abhiomkar:
Apart from above review comments, rest of the changes looks good!

This PR is conflicting with incoming change related to sass module refactor PR #5453

I'm not able to run global test since it is conflicting with above PR.

@jathak Let me know if you are able to get the sass module change in today or next week first-half. If you think you'll need some more time we can merge this first and re-run migrator script on top of this.

jathak:
I think we can get the module system migration in by EOD Monday. I'm waiting on a TGP at 6 PM just to confirm that everything's working (a previous TGP with mostly similar code passed yesterday), but other than that I just need review/approval.

crisbeto:
Thank you for looking into it @abhiomkar, I've pushed all the changes. It also cut the number of changes in the final changes in half 👍.

abhiomkar:
Thanks @crisbeto for prompt changes! As per above comment, we would wait till PR #5453 is merged. Once it is merged, you'll need to rebase your branch and resolve merge conflicts.

@jathak Sounds good.. We'll target to merge the Sass module system changes first.

abhiomkar:
@asyncLiz Are we expecting any big changes to text field sass files? Can this PR go in first?

asyncLiz:
Nothing coming up. The only change we have soon for textfield is #5439, and that's just DOM structure changes.

abhiomkar:
Thanks Liz!

@crisbeto It's time to rebase your PR with master. I'll run tests internally again once your PR is ready. Thanks!

crisbeto:
I've resolved all of the conflicts and I have the feature targeting test passing, but the production build throws this error and I'm not quite sure how to deal with it:

```
 ERROR in ./packages/material-components-web/material-components-web.scss
    Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
    ModuleBuildError: Module build failed (from ./node_modules/sass-loader/lib/loader.js):
    undefined
    ^
          Module packages\material-components-web\node_modules\@material\animation\_variables.scss and the new module both forward a variable named $mdc-animation-deceleration-curve-timing-function.
       ╷
    42 │ @forward "mdc-text-field";
       │ ^^^^^^^^^^^^^^^^^^^^^^^^^
       ╵
      packages\material-components-web\node_modules\@material\textfield\mdc-text-field.import.scss 42:1  @import
      stdin 54:9                                                                                         root stylesheet
          in C:\projects\material-components-web\packages\material-components-web\node_modules\@material\textfield\mdc-text-field.import.scss (line 42, column 1)
        at runLoaders (C:\projects\material-components-web\node_modules\webpack\lib\NormalModule.js:302:20)
        at C:\projects\material-components-web\node_modules\loader-runner\lib\LoaderRunner.js:367:11
        at C:\projects\material-components-web\node_modules\loader-runner\lib\LoaderRunner.js:233:18
        at context.callback (C:\projects\material-components-web\node_modules\loader-runner\lib\LoaderRunner.js:111:13)
        at render (C:\projects\material-components-web\node_modules\sass-loader\lib\loader.js:52:13)
        at Function.call$2 (C:\projects\material-components-web\node_modules\dart-sass\sass.dart.js:55025:16)
        at _render_closure.call$0 (C:\projects\material-components-web\node_modules\dart-sass\sass.dart.js:33874:23)
        at Object.Primitives_applyFunction (C:\projects\material-components-web\node_modules\dart-sass\sass.dart.js:1057:30)
        at Object.Function_apply (C:\projects\material-components-web\node_modules\dart-sass\sass.dart.js:4894:16)
        at _callDartFunctionFast (C:\projects\material-components-web\node_modules\dart-sass\sass.dart.js:6574:16)
        at C:\projects\material-components-web\node_modules\dart-sass\sass.dart.js:6552:18
    Child mini-css-extract-plugin node_modules/css-loader/dist/cjs.js??ref--4-1!node_modules/postcss-loader/src/index.js??ref--4-2!node_modules/sass-loader/lib/loader.js??ref--4-3!packages/material-components-web/material-components-web.scss:
           1 module
        ERROR in ./packages/material-components-web/material-components-web.scss (./node_modules/css-loader/dist/cjs.js??ref--4-1!./node_modules/postcss-loader/src??ref--4-2!./node_modules/sass-loader/lib/loader.js??ref--4-3!./packages/material-components-web/material-components-web.scss)
        Module build failed (from ./node_modules/sass-loader/lib/loader.js):
        undefined
        ^
              Module packages\material-components-web\node_modules\@material\animation\_variables.scss and the new module both forward a variable named $mdc-animation-deceleration-curve-timing-function.
           ╷
        42 │ @forward "mdc-text-field";
           │ ^^^^^^^^^^^^^^^^^^^^^^^^^
           ╵
          packages\material-components-web\node_modules\@material\textfield\mdc-text-field.import.scss 42:1  @import
          stdin 54:9                                                                                         root stylesheet
              in C:\projects\material-components-web\packages\material-components-web\node_modules\@material\textfield\mdc-text-field.import.scss (line 42, column 1)
```

abhiomkar:
Make sure you're testing with Sass version `v1.24.3` or above. Looking at the above error message, it seems like it is using previous Sass version which had this known issue.

crisbeto:
It was because I was mixing `@use` and `@import` in one of the files. It should be good to go now. Here are the before/after files:

[before.css.txt](https://github.com/material-components/material-components-web/files/4075439/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/4075440/after.css.txt)

abhiomkar:
Seems like there are still few minor CSS ordering changes that are unexpected.

For example, `.mdc-text-field--with-leading-icon {}` selector is rendered after `.mdc-text-field--with-leading-icon.mdc-text-field--outlined`.

Since text field has lots of subtle animations not all screenshot tests would capture it.

As discussed on chat, let's try to keep the CSS ordering changes to minimal. Also it would be great if we can avoid introducing new mixins (Such as `input()` & `outlined()`) in the same PR.

crisbeto:
@abhiomkar I managed to reduce the differences even more, but there are still some which are difficult to work around. Here's the breakdown:

1. The order of the animation keyframes and the helper text/icon/character counter is swapped. There's no way to get around it if we want to keep everything under the `core-styles` mixin. I don't think this will be a problem since the styles don't affect each other.
2. The ripple-related styles are moved out after all the other form field styles. We can't get around it since the ripple styles and the rest of the styles are in different mixins.
3. In two places the `transition` property was moved to the end so that things fit neatly inside feature targets. Technically we can work around it, but the formatting would look weird and I'd expect someone to eventually refactor it.

Here are the updated before/after files.
[before.css.txt](https://github.com/material-components/material-components-web/files/4082601/before.css.txt)
[after.css.txt](https://github.com/material-components/material-components-web/files/4082602/after.css.txt)

PR #5382, opened 2019-12-23
test(chips): Add Jasmine component test to chip

allan-chen:


allan-chen:
Thanks! Should be good - number of Jasmine tests increased from 171 -> 217

PR #5408, opened 2020-01-06
refactor: avoid BEM naming for mixins

jathak:
This replaces `__` and `--` in certain mixin names with a single dash to
allow migrating to the module system without breaking downstream users
that use these pseudoprivate mixins.


abhiomkar:
Update: waiting for cl/288369347

PR #5414, opened 2020-01-07
test(list): Add Jasmine component test.

joyzhong:


joyzhong:
> Do we need to update karma config to include these new test paths?

Already done in https://github.com/material-components/material-components-web/pull/5406.

Issue #5348, opened 2019-12-17
[All] Secondary Variant is Used Instead of Secondary Color


una:
### Steps to reproduce
See any component that uses secondary color

### Actual behavior
Color seen is secondary variant
<img width="419" alt="Screen Shot 2019-12-17 at 12 32 56 PM" src="https://user-images.githubusercontent.com/1693164/71033556-83db2c00-20e5-11ea-9b80-47d114bd1d0b.png">

### Expected behavior
Secondary color
<img width="1042" alt="Screen Shot 2019-12-17 at 12 33 29 PM" src="https://user-images.githubusercontent.com/1693164/71033569-8b9ad080-20e5-11ea-92e6-35487ba5394f.png">

Note: other platforms use the secondary color instead of variant for these components. I.e. on Flutter:

<img width="1081" alt="Screen Shot 2019-12-17 at 3 56 11 PM" src="https://user-images.githubusercontent.com/1693164/71033680-d0bf0280-20e5-11ea-84d7-7cd764359066.png">

judearasu:
@una  on mdc-theme package secondary variant color  **#018786** configured for $mdc-theme-accent which in turn used for secondary.

```
$mdc-theme-accent: #018786 !default; // baseline teal, 600 tone
$mdc-theme-secondary: $mdc-theme-accent !default;
```

PR #5415, opened 2020-01-07
fix(theme): set correct value for secondary color. (#5348)

yavanosta:
Update secondary color to correct value.

Baseline colors for material design can be found [here](https://material.io/design/color/the-color-system.html#color-theme-creation).

### DIFF
--- a/packages/mdc-theme/_variables.scss
+++ b/packages/mdc-theme/_variables.scss
@@ -33,7 +33,7 @@ $mdc-theme-on-primary: if(mdc-theme-contrast-tone($mdc-theme-primary) == "dark",
 
 // The $mdc-theme-accent variable is DEPRECATED - it exists purely for backward compatibility.
 // The $mdc-theme-secondary* variables should be used for all new projects.
-$mdc-theme-accent: #018786 !default; // baseline teal, 600 tone
+$mdc-theme-accent: #03dac6 !default; // baseline teal, 200 tone
 $mdc-theme-secondary: $mdc-theme-accent !default;
 $mdc-theme-on-secondary: if(mdc-theme-contrast-tone($mdc-theme-secondary) == "dark", #000, #fff) !default;
 $mdc-theme-background: #fff !default; // White
--- a/test/screenshot/golden.json
+++ b/test/screenshot/golden.json
@@ -56,11 +56,11 @@
     }
   },
   "spec/mdc-dialog/classes/baseline-confirmation.html": {
-    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/abhiomkar/2019/08/27/16_53_20_055/spec/mdc-dialog/classes/baseline-confirmation.html?utm_source=golden_json",
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2020/01/10/11_56_34_227/spec/mdc-dialog/classes/baseline-confirmation.html?utm_source=golden_json",
     "screenshots": {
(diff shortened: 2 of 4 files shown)


packages/mdc-theme/_variables.scss
```diff
@@ -33,7 +33,7 @@ $mdc-theme-on-primary: if(mdc-theme-contrast-tone($mdc-theme-primary) == "dark",
 
 // The $mdc-theme-accent variable is DEPRECATED - it exists purely for backward compatibility.
 // The $mdc-theme-secondary* variables should be used for all new projects.
-$mdc-theme-accent: #018786 !default; // baseline teal, 600 tone
+$mdc-theme-accent: #03dac6 !default; // baseline teal, 600 tone
```
asyncLiz: This is a bit nit picky, but is the comment "600 tone" still accurate?

Yavanosta: Oh, yes that's good question.

I used [color tool](https://material.io/design/color/the-color-system.html#tools-for-picking-colors) to find tone number.

Updated to 200.

asyncliz (approved):
Looks good :)

test/screenshot/golden.json
```diff
@@ -432,11 +432,11 @@
     }
   },
   "spec/mdc-list/classes/checkbox.html": {
-    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2019/05/10/14_25_41_111/spec/mdc-list/classes/checkbox.html?utm_source=golden_json",
+    "public_url": "https://storage.googleapis.com/mdc-web-screenshot-tests/travis/2020/01/09/13_35_33_056/spec/mdc-list/classes/checkbox.html?utm_source=golden_json",
```
abhiomkar: is this white-tick color expected on new secondary color? Can we use the same ink color on new secondary for consistency? Fab renders black on new secondary as per above screenshots.

test/screenshot/golden.json
```diff
@@ -432,11 +432,11 @@
     }
   },
   "spec/mdc-list/classes/checkbox.html": {
```
abhiomkar: Remove these screenshot updates. These are deleted in PR #5442

abhiomkar (changes requested):

test/screenshot/golden.json
Yavanosta: That's because checkbox [use secondary](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/_variables.scss#L30) as a background but use [on primary](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/_variables.scss#L27) for checkmark.

Should I update it?

abhiomkar: I think we need to fix that first. Can you change it to `on-secondary`? Also to other components wherever applicable.

Yavanosta: In this PR or create another Issue an another PR?

abhiomkar: Preferably in the same PR since this PR affects the ink color in components.

Yavanosta: Done!

test/screenshot/golden.json
Yavanosta: Done

test/screenshot/golden.json
Yavanosta: Also added separate style for disabled checkbox mark. Now it is the same for enabled and disabled and depends on secondary color. So change of secondary color affects disabled checkbox checkmark which is on the same grey background.

anlexN:
can we solve PR?

PR #5423, opened 2020-01-07
test(textfield): Migrate component unit test for icon to Jasmine

allan-chen:


packages/mdc-textfield/icon/test/component.test.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+
+import {emitEvent} from '../../../../testing/dom/events';
+import {MDCTextFieldIcon, MDCTextFieldIconFoundation} from '../../../mdc-textfield/icon/index';
+
+const getFixture = () => {
+  const wrapper = document.createElement('div');
+  wrapper.innerHTML = `
+    <div class="mdc-text-field__icon"></div>
```
asyncLiz: Add `mdc-text-field__icon--leading` and this will be good to go with my other PR :)

allan-chen: Done - I'll run another sync once your #5367 is merged

packages/mdc-textfield/icon/test/foundation.test.ts
```diff
@@ -148,7 +148,6 @@ describe('MDCTextFieldIconFoundation', () => {
     const evt = {
       target: {},
       type: 'click',
-      preventDefault: () => {}
```
allan-chen: @asyncLiz is this intended?

asyncLiz: Yes, it needs to be there because the icon foundation calls preventDefault() to ensure that clicks don't bubble up to the host label and focus the input.

We're not actually providing a real event, so we have to mock that function.

allan-chen: Okay, I'll add that in internal and resync

PR #5425, opened 2020-01-08
test(data-table): Add data table foundation jasmine test

abhiomkar:
Project import generated by Copybara.

PR #5426, opened 2020-01-08
test(data-table): Add data table component Jasmine test.

abhiomkar:


packages/mdc-data-table/test/foundation.test.ts
```diff
@@ -0,0 +1,318 @@
+/**
```
abhiomkar: This file is synced in PR #5425

PR #5442, opened 2020-01-09
test(list): Remove list screenshot tests.

joyzhong:
These are being migrated to use internal screenshot testing infra.


joyzhong:
> LGTM, Can you update golden.json file.
> 
> You can manually delete all entries for spec/mdc-list

Done.

PR #5448, opened 2020-01-10
fix(floatinglabel): Estimate hidden scroll width

patrickrodee:
In some cases, the floating label needs to immediately know its width. This creates problems if the floating label is instantiated inside a `display: none;` parent element, like a hidden dialog. To resolve that, we provide a helper method that estimates the width of an element if hidden. If visible, it computes the true width.

### DIFF
--- a/packages/mdc-dom/README.md
+++ b/packages/mdc-dom/README.md
@@ -34,7 +34,7 @@ Function Signature | Description
 --- | ---
 `closest(element: Element, selector: string) => ?Element` | Returns the ancestor of the given element matching the given selector (which may be the element itself if it matches), or `null` if no matching ancestor is found.
 `matches(element: Element, selector: string) => boolean` | Returns true if the given element matches the given CSS selector.
-`estimateScrollWidth(element: Element) => number`  | 
+`estimateScrollWidth(element: Element) => number`  | Returns the true optical width of the element if visible or an estimation if hidden by a parent element with `display: none;`.
 
 ### Event Functions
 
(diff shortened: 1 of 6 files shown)


packages/mdc-dom/ponyfill.ts
```diff
@@ -47,3 +47,29 @@ export function matches(element: Element, selector: string): boolean {
...
+  // parent, the offsetParent property will be null (see
+  // https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent).
+  // This check ensures we only clone the node when necessary.
+  if (element.offsetParent !== null) {
+    return element.scrollWidth;
+  }
+
+  const clone = element.cloneNode(true) as HTMLElement;
+  clone.style.setProperty('position', 'absolute');
+  clone.style.setProperty('transform', '-9999px, -9999px');
```
asyncLiz: Does this need `translate(...)` in the string around the value?

patrickrodee: Sure does! Thanks Liz

packages/mdc-dom/README.md
```diff
@@ -34,6 +34,7 @@ Function Signature | Description
 --- | ---
 `closest(element: Element, selector: string) => ?Element` | Returns the ancestor of the given element matching the given selector (which may be the element itself if it matches), or `null` if no matching ancestor is found.
 `matches(element: Element, selector: string) => boolean` | Returns true if the given element matches the given CSS selector.
+`estimateScrollWidth(element: Element) => number`  |
```
asyncLiz: Do we need a short description for this?

packages/mdc-floating-label/component.ts
```diff
@@ -22,6 +22,7 @@
  */
 
 import {MDCComponent} from '@material/base/component';
+import {ponyfill} from '@material/dom/index';
```
asyncLiz: What if you did `import {estimateScrollWidth} from '@material/dom/ponyfill';`? That seems to be how the other packages import ponyfill functions.

packages/mdc-dom/README.md
patrickrodee: Done

packages/mdc-floating-label/component.ts
patrickrodee: Done

PR #5453, opened 2020-01-10
refactor: migrate to the Sass module system

jathak:
This PR was generated by running [this script](https://gist.github.com/jathak/dc9559e03e39ab5e784e4965c54757c8) based on the [Sass migrator](https://github.com/sass/migrator). It migrates all components to the Sass module system, removing manual `mdc-$component` prefixes and implicit dependencies between files (though still leaving both intact for users still using `@import`).

High-level overview of the changes made here:

- Prefixes (e.g. `mdc-button-`) for each component have been removed from all variables, functions, and mixins within the code. Prefixes on CSS selectors are unaffected.
- When depending on one of a component's files via `@use`, only members directly declared in that file are available. The `mdc-$component.scss` files should only expose CSS, not Sass members.
- Import-only files have been generated that add back all of the prefixes and explicitly forward all members that were implicitly available via `@import` previously, so existing users that still use `@import` shouldn't break as a result of these changes. These files will then be used by the migrator when those users later migrate to the module system themselves.
- **Minor breaking change**: Four variables and a mixin in mdc-textfield were renamed to use a `mdc-text-field-` prefix when depended on via `@import`. None of the variables are used within Google and the mixin (formerly `mdc-required-text-field-label-asterisk_`, now `required-label-asterisk_`) is used only once by MWC.


jathak:
One other note about the script used here: while the underlying Sass migrator is fairly robust, the script itself is fragile, as preserving the proper prefixes for some of the more complex components (e.g. mdc-ripple, mdc-select, mdc-textfield) required a bunch of `sed` transformations, some of which reference specific lines to change. We can reuse the script if there are enough changes to master before this is submitted to make a merge difficult, but future runs may require tweaks to it.

jathak:
Not sure why the linter is reporting an error here. It doesn't seem to be related to the Sass code

jathak:
The GitHub Actions linter seems to be using an old config

jathak:
Okay, the lint issue was that stylelint-scss was out of date. The unit tests seem to be failing because of an issue with Chrome

jathak:
[cl/289132951](http://cl/289132951)

PR #5458, opened 2020-01-11
fix(textfield): prevent placeholder styles from collapsing with minifiers

asyncLiz:


asyncLiz:
@patrickrodee my CL is synced to head, but it looks like what it generated is missing some of the latest master commits. Any ideas why?

asyncLiz:
Need to wait for #5453 first

Issue #5441, opened 2020-01-09
[MDC Tab-Bar] in package.json is @material/tab-indicator dependency missing


damian-mariachi-team:
### Your Environment:

| Software         | Version(s) |
| ---------------- | ---------- |
| MDC Web          | 4.0.0
| Browser          | Chrome
| Operating System | Windows 10

### Possible solution
**package.json now:**
```
"dependencies": {
    "@material/animation": "^4.0.0",
    "@material/base": "^4.0.0",
    "@material/density": "^4.0.0",
    "@material/elevation": "^4.0.0",
    "@material/feature-targeting": "^4.0.0",
    "@material/tab": "^4.0.0",
    "@material/tab-scroller": "^4.0.0",
    "tslib": "^1.9.3"
  },
```
**package.json solution:**
```
"dependencies": {
    "@material/animation": "^4.0.0",
    "@material/base": "^4.0.0",
    "@material/density": "^4.0.0",
    "@material/elevation": "^4.0.0",
    "@material/feature-targeting": "^4.0.0",
    "@material/tab": "^4.0.0",
    "@material/tab-indicator" : "^4.0.0",
    "@material/tab-scroller": "^4.0.0",
    "tslib": "^1.9.3"
  }, 
```

asyncLiz:
While it does not _technically_ depend on `@material/tab-indicator`, I agree that it should be part of its dependencies. A tab indicator cannot be used on its own and a tab-bar _should not_ be used without an indicator.

Would you be willing to open a PR for this?

damian-mariachi-team:
Hi asyncLiz. 

What is PR and how can i it open?

asyncLiz:
It is a Pull Request, a way for you to contribute code changes to the repository. You can [fork this repo](https://help.github.com/en/github/getting-started-with-github/fork-a-repo), make the changes, and then [open a PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork) to submit them to us for review.

damian-mariachi-team:
Thanks.
I have a Pull Request opened...

allan-chen:
I notice that tab depends on tab-indicator. Does the transitive dependency not work here?

asyncLiz:
Good find @allan-chen. I created a new directory and ran `npm install @material/tab-bar` and that tab-indicator is installed as well.

@damian-mariachi-team can you confirm?

asyncLiz:
[...]

PR #5461, opened 2020-01-13
test: Ensure package.json dependencies match with actual dependencies

allan-chen:
Can test on current master state by running the following on master:
```
git checkout exp/checkscript scripts/check-pkg-for-release.js
npm run test:dependency
```
Sample output if dependencies don't match up:

```
AssertionError [ERR_ASSERTION]: FAILURE: The following MDC dependencies were used in @material/checkbox but were not declared in its package.json:
@material/density@4.0.0
@material/dom@4.0.0

Please add the missing dependencies to package.json manually, or by running the following command(s) on the root of the repository:
npx lerna add @material/density@4.0.0 packages/mdc-checkbox
npx lerna add @material/dom@4.0.0 packages/mdc-checkbox

    at ...
```


scripts/check-pkg-for-release.js
```diff
@@ -278,3 +319,19 @@ function getPkgName() {
   }
   return name;
 }
+
+function getMissingDependencyRemedy(missingDeps) {
+  let remedyStr = 'Please add the missing dependencies using the following command(s):\n';
+  missingDeps.forEach((dep) => {
+    remedyStr += `npx lerna add ${dep} packages/${PACKAGE_RELATIVE_PATH.split('/')[1]}\n`;
```
abhiomkar: This suggestion might mislead someone who exported a CL, executing this command would only the PR. Can we also list out missing dependency (with version number) so that we can manually copy & paste it to package.json internally?

WDYT?

allan-chen: How about checking if this is an exported CL by seeing if branch is `cl_*`, then conditionally printing a warning message following this to patch all `package.json` changes to internal? This saves us from having to dig up the version number using something like `npm show {pkg} version`, which learna would have done anyways

abhiomkar: Can we show both suggestions? Author will figure out how to fix it.

For example,

```
Failure: Missing dependencies in mdc-checkbox
  @material/feature-targeting@4.0.0, @material/density@4.0.0

Manually add these dependencies to mdc-checkbox's package.json file
or run following command on root of GitHub repository for auto fix.
  npx lerna add <missing_deps> ...
```

WDYT?

scripts/check-pkg-for-release.js
```diff
@@ -269,6 +272,44 @@ function checkComponentExportedAddedInMDCPackage(ast) {
...
+  const declaredButNotUsed = [...declaredDeps].filter((x) => !usedDeps.has(x));
+
+  assert.equal(usedButNotDeclared.length, 0,
+    'FAILURE: Component ' + CLI_PACKAGE_JSON.name +
+    ' uses one or more MDC dependencies not declared in package.json.\n' +
+    getMissingDependencyRemedy(usedButNotDeclared));
+
+  assert.equal(declaredButNotUsed.length, 0,
+    'FAILURE: Component ' + CLI_PACKAGE_JSON.name +
+    ' declares one or more MDC dependencies in package.json that are unused.\n' +
```
asyncLiz: This would break #5441, which I think is a valid use case. While `@material/tab-indicator` is not directly used in code, but it is required in the DOM markup.

I would expect `npm install @material/tab-bar` to install all required dependencies for both code and DOM.

Thoughts?

scripts/check-pkg-for-release.js
allan-chen: For versioning, is it safe to assume master `material-components-web` version will always be the same as the sub-module versions? 4.0.0

abhiomkar: Version of material-components-web and sub-modules are not always same.

But we can get a current version of a package by executing this command:

```shell
npm show @material/checkbox version
```

scripts/check-pkg-for-release.js
abhiomkar: What do you mean by referenced in DOM?

If a subcomponent is used, clients would've to explicitly install all the packages required to render that component. I think we should update installation section of [tab-bar README](https://github.com/material-components/material-components-web/tree/master/packages/mdc-tab-bar#installation) file to include `@material/tab-indicator`, `@material/tab-scroller` & `@material/tab`

asyncLiz: I think [Allan just found](https://github.com/material-components/material-components-web/issues/5441#issuecomment-574772803) that tab-indicator is depended on by tab, and that does install it correctly. We may not need to do anything there or update the readme since all the subcomponents are installed alongside it.

scripts/check-pkg-for-release.js
allan-chen: I revised the script to include these details in the message and accommodate the new sass module changes

scripts/check-pkg-for-release.js
```diff
@@ -269,6 +273,37 @@ function checkComponentExportedAddedInMDCPackage(ast) {
   return isExported;
 }
 
+function checkUsedDependenciesMatchDeclaredDependencies() {
```
abhiomkar: Add JSDoc

PR #5468, opened 2020-01-15
fix(tab-bar): add dependency for tab-indicator

damian-mariachi-team:


packages/mdc-tab-bar/package.json
```diff
@@ -25,6 +25,7 @@
     "@material/elevation": "^4.0.0",
     "@material/feature-targeting": "^4.0.0",
     "@material/tab": "^4.0.0",
+	"@material/tab-indicator": "^4.0.0",
```
asyncLiz: Can you make sure this is indented with 4 spaces?

asyncLiz:
Going to go ahead and close this one, seeing as `@material/tab` [depends on tab-indicator](https://github.com/material-components/material-components-web/blob/master/packages/mdc-tab/package.json#L25) and it is installed correctly when running `npm install @material/tab-bar`.

PR #5474, opened 2020-01-16
fix(chips): fix missing constants of chips

TroyTae:


asyncLiz:
@TroyTae what is the motivation for this change? These constants are not being used anywhere by the foundation or component.

TroyTae:
@asyncLiz 
Now I'm making a ui module that inherit sass of material-components.
(foundations/adapters are hard to inherit...)
I worried about changing className.
If I use string literal, I need to track breaking changes(about className changes).
But If you provide us constants, I don't need any migration work, just update version.
So I create this PR :)

asyncLiz:
Thanks for the PR, but I spoke with the team and we don't think this would be a viable solution to mitigate breaking changes.

Typically when class names change, the DOM structure will also change. This PR would not prevent breaking changes in the DOM or class name usage in html templates, so you'll still have to update libraries that extend MDC.

PR #5491, opened 2020-01-18
refactor: Add index stylesheets to each package

nex3:
As discussed with @abhiomkar, this adds an `_index.scss` file to each package so that each package can be loaded with a single `@use` rule. As a bonus, that `@use` rule will automatically use the package's name as its namespace.

[Here's an example](https://gist.github.com/nex3/53cfb8ad8e6717e3e90a61be7aeccc23/revisions?diff=unified) of the difference between an API using each separate files versus index files.


packages/mdc-checkbox/_variables.import.scss
```diff
@@ -1,7 +1,5 @@
...
...ut-ripple, mdc-checkbox-ripple, mdc-checkbox-density, mdc-checkbox-ripple-size, mdc-checkbox-container-colors, mdc-checkbox-disabled-container-colors, mdc-checkbox-ink-color, mdc-checkbox-disabled-ink-color, mdc-checkbox-focus-indicator-color, mdc-checkbox-touch-target, mdc-checkbox-base-, mdc-checkbox-disabled-, mdc-checkbox-child--upgraded-, mdc-checkbox-anim-, mdc-checkbox-if-enabled-, mdc-checkbox-if-disabled-, mdc-checkbox-if-unmarked-enabled-, mdc-checkbox-if-unmarked-disabled-, mdc-checkbox-if-marked-enabled-, mdc-checkbox-if-marked-disabled-, mdc-checkbox-background-, mdc-checkbox-background--marked-, mdc-checkbox-container-colors-, mdc-checkbox-focus-indicator-, mdc-checkbox-focus-indicator--focused-, mdc-checkbox-native-control-, mdc-checkbox-checkmark-, mdc-checkbox-checkmark--checked-, mdc-checkbox-checkmark--indeterminate-, mdc-checkbox-ink-color-, mdc-checkbox-checkmark-path-, mdc-checkbox-checkmark-path--marked-, mdc-checkbox-mixedmark-, mdc-checkbox-mixedmark--checked-, mdc-checkbox-mixedmark--indeterminate-, mdc-checkbox-touch-target-reset-, mdc-checkbox-container-keyframes-, mdc-checkbox-mark-keyframes-, mdc-checkbox-transition-enter, mdc-checkbox-transition-exit;
```
abhiomkar: Why is it hiding some of the mixins starting with mdc-checkbox- prefix? Example: `mdc-checkbox-ripple` etc.

Same for other components.

nex3: The goal is to forward exactly the same set of names that were forwarded previously. In the case of this file, only the `variables` file was forwarded previously, so all the mixins and functions (as well as `$mdc-checkbox-ripple-target` which is defined in `_mixins.scss`) are hidden to preserve the existing behavior.

abhiomkar:
> 
> I tried running `mdc import_pr 5491` but I got an error:
> 
> ```
> ERROR: Cannot find last imported revision. Use --force if you really want to proceed with the migration
>   CAUSED BY: Cannot find reference '6fb7ccb68afea5ad782e79948764ca16f1b40468'
> ```

Please use `--force` option to force snapshot the PR.

> My main goal is to unblock downstream migrations as soon as possible. I'll try to make time to circle back to this later on, but it may need to wait until @jathak is back in the office and the team has more bandwidth again.

SGTM. Thanks!

packages/mdc-checkbox/_variables.import.scss
abhiomkar: Can we not use index paths in import files in that case? I do not see any benefit of using index files here. This import path is very specific to variables might be a good idea to scope it to variables only. WDYT?

Can migrator script handle this without making a big change?

jathak:
> * Can we also update actual _mixin, _variables files to use these index files, using index files greatly improves readability of our Sass files.

I don't think there's currently a simple way to do this automatically, but writing one shouldn't be too challenging (since it's essentially just merging the `@use` rules and changing namespaces), but since it's a change that's local to the file, it makes sense to do that separately, once the main `index` change is done.

packages/mdc-checkbox/_variables.import.scss
nex3: When the migrator is generating new `@use` rules for stylesheets that depend on MDC Web, it figures out which URLs to use based on the existing `@forward` rules in the `.import.scss` files. Forwarding index files here means that downstream stylesheets migrated using the migrator will automatically use index files—otherwise they'd default to using the same old mixins, variables, and functions files.

We could potentially special-case this for the migration within Google in particular, but in the external world these `@forward` rules are the only way to communicate the correct pattern to downstream invocations of the migrator.

abhiomkar: I do not have any strong preference on how import file `@forward`s things from.

IMO, Since we intend to only forward variables here, using the direct path (Eg: `@material/checkbox/variables`) seems more idiomatic and avoids hiding lot of things.

IIUC, downstream users will not notice whether it is `@forward`-ed from index file or variables file.

nex3: > IIUC, downstream users will not notice whether it is `@forward`-ed from index file or variables file.

They will notice when they run the migrator. If this file forwards from a variable file, the migrator will generate `@use "@material/checkbox/variables"`, but if this file forwards from an index file, the migrator will generate `@use "material/checkbox"`.

abhiomkar: Thanks for clarifying! If the migrator dependents on this import path then we can use index paths like you did.

Resolving this comment.

nex3:
Obsoleted by https://github.com/material-components/material-components-web/pull/5539

PR #5500, opened 2020-01-22
chore: Migrate to yarn, replacing npm

abhiomkar:


asyncLiz:
I'm curious, what do we gain by switching to yarn?

abhiomkar:
I’m just experimenting with yarn to see if we can fix recurring NPM lock failure on CI.

Created this PR for CI tests.

abhiomkar:
NPM lock failure has a fix. See https://github.com/isaacs/chownr/pull/22

We no longer need this change. Need to bump NPM version when it fix is ready to use.

PR #5510, opened 2020-01-23
docs(icon-button): remove stray aria-hidden

### LINKED ISSUE #5509: [IconButton] documentation adds aria-hidden
## Bug report
Some of the HTML examples have an `aria-hidden="true"` attribute on the `<button>` element. I think that probably shouldn't be there, since it would hide the button from the accessibility tree.

### Steps to reproduce
1. Go to https://material.io/develop/web/components/buttons/icon-buttons/#icon-button-toggle
2. Note the `aria-hidden="true"` attribute in the HTML example
3. See the same thing at https://material.io/develop/web/components/buttons/icon-buttons/#icon-button-toggle-with-svg and https://material.io/develop/web/components/buttons/icon-buttons/#icon-button-toggle-with-an-image

### Actual behavior
The `aria-hidden="true"` attribute is present on the `<button>` element in each example.

### Expected behavior
The examples show `<button>` elements without the `aria-hidden="true"` attribute, as seen for example on the demos at https://material-components.github.io/material-components-web-catalog/#/component/icon-button

jmorahan:
Remove an aria-hidden="true" attribute from the <button> element in some of the HTML examples.
The attribute would hide the element from the accessibility tree, making it inaccessible.

Fix #5509.

### DIFF
--- a/packages/mdc-icon-button/README.md
+++ b/packages/mdc-icon-button/README.md
@@ -79,7 +79,6 @@ Then instantiate an `MDCIconButtonToggle` on the root element.
 <button id="add-to-favorites"
    class="mdc-icon-button"
    aria-label="Add to favorites"
-   aria-hidden="true"
    aria-pressed="false">
    <i class="material-icons mdc-icon-button__icon mdc-icon-button__icon--on">favorite</i>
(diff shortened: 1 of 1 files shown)


bonniezhou:
Thanks for submitting!

PR #5512, opened 2020-01-23
fix(slider): slider track not visible

crisbeto:
These changes fix the following issues which were introduced by #5132:
1. The background of track is now provided by `.mdc-slider__track-container::after`, however `::after` wasn't positioned correctly which made it invisible.
2. The `::after` structural styles were set inside the `color` feature target. This means that they'll end up being included multiple times and in the wrong place.

Screenshot for reference:
Screenshot at Jan 23 21-36-34

### DIFF
--- a/packages/mdc-slider/_mixins.scss
+++ b/packages/mdc-slider/_mixins.scss
@@ -449,6 +449,7 @@
   .mdc-slider__track-container::after {
     @include feature-targeting-mixins.targets($feat-color) {
       @include theme-mixins.prop(background-color, $color);
+
       opacity: $opacity;
     }
   }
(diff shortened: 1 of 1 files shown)


packages/mdc-slider/_mixins.scss
```diff
@@ -262,6 +262,19 @@
...
+
+    &::after {
+      @include feature-targeting-mixins.targets($feat-structure) {
+        position: absolute;
+        top: 0;
+        left: 0;
+        display: block;
+        width: 100%;
+        height: 100%;
+        opacity: $opacity;
```
abhiomkar: Seems like build is failing because of undefined `$opacity`.

Should this property moved back to rail-color_ mixin?

crisbeto: ops, yeah I was being too quick there. It's fixed now.

PR #5539, opened 2020-01-29
feat: Add index stylesheets to each MDC Web package

abhiomkar:

Author: @nex3


testing/dom/index.ts
```diff
@@ -33,14 +33,3 @@ export const html =
       result += strings[strings.length - 1];
       return result;
     };
-
-/**
- * @returns HTMLElement for given HTML content in string format.
- */
-export function getFixture(content: string): HTMLElement {
```
abhiomkar: Should revert this before merge.

abhiomkar:
@nex3 mdc-image-list & mdc-layout-grid are missing in this PR. This is expected since these 2 packages are not synced internally.

Can we run the migrator on these 2 packages in a separate PR? We don't need to import these.

Thanks!

nex3:
Done: https://github.com/material-components/material-components-web/pull/5546

joyzhong:
LGTM based on cl/292403730 review

PR #5554, opened 2020-02-01
fix(textfield): incorrect mixin forward path

crisbeto:
Fixes an error that was due to a `@forward` which is pointing to an incorrect file path. Seems to have been introduced in https://github.com/material-components/material-components-web/commit/181486643532e2166dced95daff9da786af3bdd1.


mmalerba:
@abhiomkar I'm not sure but it looks like the kind of thing that could point to a bug in your copybara script. You may want to look into whether that's the root cause

abhiomkar:
> @abhiomkar I'm not sure but it looks like the kind of thing that could point to a bug in your copybara script. You may want to look into whether that's the root cause

Yes, this definitely looks like incorrect Copybara transform related to `@forward` pattern.

PR #5556, opened 2020-02-03
docs(select): fix aria attribute on select example

mikemonteith:
The `aria-selected="true"` attribute was given to the wrong element.
The "Vegetables" option is an example of a disabled item.
The "" option should be the example selected item and have the aria-selected attribute.


mikemonteith:
@googlebot I signed it!

mikemonteith:
I don't know why the automated check is failing. My change is only documentation.

PR #5559, opened 2020-02-03
fix(textfield): move ripple to separate element

asyncLiz:
BREAKING CHANGE: filled text fields must include a `<div class="mdc-text-field__ripple"></div>`


docs/open_source/rotation.md
```diff
@@ -0,0 +1,111 @@
+# Rotation
```
patrickrodee: Out of sync?

asyncLiz: Synced up!

PR #5562, opened 2020-02-04
feat(progress-indicator): Add common progress indicator interface

allan-chen:

### DIFF
--- a/scripts/check-pkg-for-release.js
+++ b/scripts/check-pkg-for-release.js
@@ -73,15 +73,22 @@ const CSS_WHITELIST = [
   'density',
   'dom',
   'feature-targeting',
+  'progress-indicator',
   'rtl',
   'shape',
   'touch-target',
+];
+
+const JS_WHITELIST = [
+  'animation',
+  'progress-indicator',
 ];
 
 const NOT_AUTOINIT = [
   'auto-init',
   'base',
   'dom',
+  'progress-indicator',
   'tab', // Only makes sense in context of tab-bar
   'tab-indicator', // Only makes sense in context of tab-bar
   'tab-scroller', // Only makes sense in context of tab-bar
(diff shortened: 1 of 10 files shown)


packages/material-components-web/index.ts
```diff
@@ -38,6 +38,7 @@ import * as list from '@material/list/index';
 import * as menuSurface from '@material/menu-surface/index';
 import * as menu from '@material/menu/index';
 import * as notchedOutline from '@material/notched-outline/index';
+import * as progressIndicator from '@material/progress-indicator/index';
```
patrickrodee: I don't think this is necessary. We only use this internally.

scripts/webpack/js-bundle-factory.js
```diff
@@ -184,6 +184,7 @@ class JsBundleFactory {
         menu: getAbsolutePath('/packages/mdc-menu/index.ts'),
         menuSurface: getAbsolutePath('/packages/mdc-menu-surface/index.ts'),
         notchedOutline: getAbsolutePath('/packages/mdc-notched-outline/index.ts'),
+        progressIndicator: getAbsolutePath('/packages/mdc-progress-indicator/index.ts'),
```
patrickrodee: I also don't think this is necessary.

packages/material-components-web/index.ts
allan-chen: Should we follow convention of exporting every package, even ones that may not be used directly (i.e. base). Leaving it out would require adding some whitelists to the dependency test just for this package

scripts/webpack/js-bundle-factory.js
allan-chen: I agree - webpack doesn't need this

packages/material-components-web/index.ts
patrickrodee: Updating the dependency test would be good. I'd prefer that over unnecessarily exporting the progress indicator interface.

scripts/check-pkg-for-release.js
```diff
@@ -126,14 +133,17 @@ function checkDependencyAddedInWebpackConfig() {
...
-  const jsconfig = WEBPACK_CONFIG.find((value) => {
-    return value.name === 'main-js-a-la-carte';
-  });
-  const nameCamel = camelCase(CLI_PACKAGE_JSON.name.replace('@material/', ''));
-  assert.notEqual(typeof jsconfig.entry[nameCamel], 'undefined',
-    'FAILURE: Component ' + CLI_PACKAGE_JSON.name + ' javascript dependency is not added to webpack ' +
-    'configuration. Please add ' + nameCamel + ' to ' + WEBPACK_CONFIG_RELATIVE_PATH + '\'s js-components ' +
-    'entry before commit.');
+  const name = getPkgName();
+  if (JS_WHITELIST.indexOf(name) === -1) {
```
patrickrodee: Let's restructure this to early exit via `return` if the name is included in the whitelist.

scripts/check-pkg-for-release.js
```diff
@@ -162,10 +172,13 @@ function checkDependencyAddedInMDCPackage() {
 }
 
 function checkPkgDependencyAddedInMDCPackage() {
-  assert.notEqual(typeof MASTER_PACKAGE_JSON.dependencies[CLI_PACKAGE_JSON.name], 'undefined',
-    'FAILURE: Component ' + CLI_PACKAGE_JSON.name + ' is not a dependency for MDC Web. ' +
-    'Please add ' + CLI_PACKAGE_JSON.name +' to ' + MASTER_PACKAGE_JSON_RELATIVE_PATH +
-    '\' dependencies before commit.');
+  const name = getPkgName();
+  if (!CSS_WHITELIST.filter((x) => JS_WHITELIST.includes(x)).includes(name)) {
```
patrickrodee: Same comment as above: favor early exits

packages/material-components-web/index.ts
allan-chen: Done - let's continue further discussions in #5564

scripts/check-pkg-for-release.js
allan-chen: Just saw this - will make the changes in the new PR

scripts/check-pkg-for-release.js
allan-chen: Done

allan-chen:
Continuing in #5564

PR #5573, opened 2020-02-07
feat: Use @use syntax in material-components-web Sass file and add index file.

abhiomkar:


abhiomkar:
Sent #5574 to fix dependency test to support `@use` syntax.

PR #5577, opened 2020-02-07
fix(chips): .d.ts file generated with syntax error

crisbeto:
Due to the way a comment was formatted in the chip foundation the .d.ts bundler ended up generating a file with a syntax error.

For reference, here's what was being generated:
●_mdc chips d ts_-_material-components-web_-_Visua_2020-02-07_19-24-23

### DIFF
--- a/packages/mdc-chips/chip/foundation.ts
+++ b/packages/mdc-chips/chip/foundation.ts
@@ -71,9 +71,7 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
     };
   }
 
-  /**
-   * Whether a trailing icon click should immediately trigger exit/removal of the chip.
-   */
+  /** Whether a trailing icon click should immediately trigger exit/removal of the chip. */
   private shouldRemoveOnTrailingIconClick_ = true;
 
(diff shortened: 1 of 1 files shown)


packages/mdc-chips/chip/foundation.ts
```diff
@@ -71,9 +71,7 @@ export class MDCChipFoundation extends MDCFoundation<MDCChipAdapter> {
     };
   }
 
-  /**
```
abhiomkar: is there any bug created on `dts-bundler`?

The comment syntax looks correct to me. We might come across this issue in future if we add back this comment style.

crisbeto: Agreed that there's nothing wrong with the comment. I couldn't find anything in the `dts-bundler` issue tracker and the project seems mostly dead. I think that the correct approach here would be to use the compiler to generate the files during compilation, rather than depending on a third party tool.

abhiomkar: Thanks @crisbeto!

Created Issue #5580 to address this issue.

PR #5587, opened 2020-02-07
fix(chips): Move touch target inside primary action

patrickrodee:
BREAKING CHANGE: The touch target and text now appear inside the primary action element. Please see the readme for markup changes.


abhiomkar:
Can you use auto generated PR ? Services should be up now.

PR #5634, opened 2020-02-20
fix(typography): Separate @material/feature-targeting

EstebanG23:
Was causing sass test to fail and blocking PR merging.

### DIFF
--- a/packages/mdc-typography/_mixins.scss
+++ b/packages/mdc-typography/_mixins.scss
@@ -23,11 +23,12 @@
 @use "sass:list";
 @use "sass:map";
 @use "sass:string";
-@use "@material/feature-targeting" as feature-targeting;
-@use "@material/theme";
+@use "@material/feature-targeting/functions" as feature-targeting-functions;
+@use "@material/feature-targeting/mixins" as feature-targeting-mixins;
+@use "@material/theme/mixins" as theme-mixins;
 @use "./variables";
 
-@mixin core-styles($query: feature-targeting.all()) {
+@mixin core-styles($query: feature-targeting-functions.all()) {
   .mdc-typography {
     @include base($query: $query);
   }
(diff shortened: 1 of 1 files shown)


asyncLiz:
@abhiomkar suggested the different syntax, any idea why this would cause tests to fail?

packages/mdc-typography/_mixins.scss
```diff
@@ -23,11 +23,12 @@
 @use "sass:list";
 @use "sass:map";
 @use "sass:string";
-@use "@material/feature-targeting" as feature-targeting;
```
patrickrodee: Have we tried dropping the "as feature-targeting"? I don't think it's necessary.

EstebanG23: I didn't I was following the way we currently have it implemented in the rest of the mixin files.

asyncLiz: It was because copybara changes it to "featuretargeting" internally, which breaks the usage of it in the file without an explicit `as`

patrickrodee: What about `as ft` to its clearly different than `feature-targeting`?

I'm trying to find ways we can take advantage of the `_index.scss` files.

asyncLiz: If we're just looking for another name, I'd recommend `feature` as a name that is concise and makes sense.

```scss
@use "@material/feature-targeting" as feature;

@mixin my-component-core-styles($query: feature.all()) {
  $feat-structure: feature.create-target($query, structure);

  @include feature.targets($feat-structure) {
    // ...
  }
}
```

EstebanG23: I tried both ft and without the `as feature-targeting` results are as follow: 
---------
Error: Can't find stylesheet to import.
26 │ @use "@material/feature-targeting" as ft;
---------

Also do want to add this is blocking other PRs from being merged but I would encourage to file a bug/issue to investigate ways to take advantage of the _index.scss files.

patrickrodee:
Could be related to https://github.com/webpack-contrib/sass-loader/issues/804

packages/mdc-typography/_mixins.scss
abhiomkar: Thanks @EstebanG23 for sending a fix! Wondering why Copybara didn't block CL submit when github action failed in first place. 🤔 

Sass couldn't compile Sass test files with `@import` statements since the custom importer doesn't handle new _index files (See `testing/featuretargeting/sass-test-compile.helper.ts`). 

Next steps:

- Fix `sass-test-compile.helper.ts` to handle `@import` syntax. (Created #5641)
- We should migrate all our test files to use `@use` syntax.

PR #5637, opened 2020-02-20
chore: Exclude WIP circular progress from dependency test

allan-chen:
This will unblock incremental PRs for the component from merging (i.e. #5636).

### DIFF
--- a/scripts/check-pkg-for-release.js
+++ b/scripts/check-pkg-for-release.js
@@ -69,6 +69,7 @@ const CSS_EXCLUDES = new Set([
   'base',
   'animation',
   'auto-init',
+  'circular-progress', // TODO: remove once component is complete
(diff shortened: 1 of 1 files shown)


scripts/check-pkg-for-release.js
```diff
@@ -69,6 +69,7 @@ const CSS_EXCLUDES = new Set([
   'base',
   'animation',
   'auto-init',
+  'circular-progress', // TODO: remove once component is complete
```
patrickrodee: Let's file a real bug for this.

patrickrodee: Or link to an existing bug.

allan-chen: Done

PR #5640, opened 2020-02-20
fix(list): ensure sass mixin works if leading is provided in px

mmalerba:
This switches the subheader margin calculation from static to dynamic so
that it will work if the `$mdc-list-subheader-leading` is provided in
`px` instead of `rem`. Angular Material's current typography settings
specify it in `px`. We do plan to move away from that and use `rem` like
MDC, but this fix will make it work in the mean time.


abhiomkar:
@mmalerba Can you send a CL instead? See go/mdc-web-g3sync for reference.

mmalerba:
yep, closing in favor of CL

PR #5674, opened 2020-03-01
build: add missing & update package-lock.json on packages

borisdamevin:
- Add missing `package-lock.json` files on most packages.
- Update existing `package-lock.json` files.

### DIFF
--- a/packages/material-components-web/package-lock.json
+++ b/packages/material-components-web/package-lock.json
@@ -1,0 +1,4327 @@
+{
+  "name": "material-components-web",
+  "version": "5.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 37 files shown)


asyncLiz:
Thanks for the PR! At this moment we are not using package-locks per package folder since we use `lerna` to manage them. We do not `npm install` individual packages without bootstrapping everything.

borisdamevin:
Thank your @asyncLiz for this precision. But why some package have `packge-lock.json` and others don't?

asyncLiz:
Good catch! We shouldn't have those there. Would you be willing to open a PR deleting them?

borisdamevin:
@asyncLiz Yes of course, you can find the PR [here](https://github.com/material-components/material-components-web/pull/5697).

PR #5681, opened 2020-03-04
updated

Prajwal-Koirala:


Prajwal-Koirala:
@googlebot I signed it!

asyncLiz:
Thank you for the PR! However, we intentionally want to pull the `latest` tag from unpkg so that we don't have to continually update the readme with each release.

PR #5692, opened 2020-03-08
build: add @sass-collective functions

borisdamevin:
Use external dependencies for easily generate REM a EM value.


asyncLiz:
Thanks for the PR!

Unfortunately we do not want to bring in an external dependency to address this problem. Additionally, we want to discuss with design how the Material on the web should react to font-size changes before we're ready to implement a strategy.

PR #5731, opened 2020-03-19
fix(checkbox): Use secondary and on-secondary as default colors

### LINKED ISSUE #5730: [checkbox] Default $mark-color does not align with default $baseline-theme-color
## Bug report

Currently, in [`packages/mdc-checkbox/_variables.scss`](https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/_variables.scss) these two variables are set:

```scss
$baseline-theme-color: secondary !default;
$mark-color: theme-variables.prop-value(on-primary) !default;
```

My understanding is that the first sets the colors to use for the background of the checkbox and the second sets the color of the checkmark.

The defaults of `secondary` and `on-primary` respectively seem incongruous, by default creating a checkbox with a background of the primary color and a foreground of a color that shows up well on the _secondary_ color.

### Actual behavior

Default is a checkbox with a background of the primary color and a foreground of a color that shows up well on the _secondary_ color.

### Expected behavior

The default checkmark color should be a color that is intended to show up well on the default checkbox background color.

### Possible solution

I believe that the material designs specify that the checkbox should use the secondary color.  So to fix that `$mark-color` should be set:

```scss
$mark-color: theme-variables.prop-value(on-secondary)
```

 I'm happy to raise a PR to address this if you approve!

zposten:
Previously, the secondary color was by default used for the background
of the checkbox and on-primary was used for the color of the checkmark.

Fixes #5730

### DIFF
--- a/packages/mdc-checkbox/_variables.scss
+++ b/packages/mdc-checkbox/_variables.scss
@@ -25,10 +25,10 @@
 @use "@material/theme/variables" as theme-variables;
 @use "@material/density/variables" as density-variables;
 
-$mark-color: theme-variables.prop-value(on-primary) !default;
+$baseline-theme-color: secondary !default;
+$mark-color: theme-variables.prop-value(on-secondary) !default;
 $border-color: rgba(theme-variables.prop-value(on-surface), .54) !default;
 $disabled-color: rgba(theme-variables.prop-value(on-surface), .38) !default;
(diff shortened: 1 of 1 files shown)


joyzhong:
Thanks for the fix! This LGTM, but I'll need to run this change against internal tests before it can be merged in.

PR #5751, opened 2020-03-26
fix(data-table): unable to redefine colors in class-based theming

crisbeto:
In Angular Material component themes are implemented using classes (e.g. `.dark-theme{ @include mdc-core-styles() }`), however the data table color styles were inside an `@at-root` which causes them to always be generated at the root, circumventing any parent classes. These changes move it outside the `@at-root` since the mixins already have their selectors baked in.

### DIFF
--- a/packages/mdc-data-table/_mixins.scss
+++ b/packages/mdc-data-table/_mixins.scss
@@ -52,25 +52,6 @@
     @include stroke-size(variables.$stroke-size, $query: $query);
     @include stroke-color(variables.$stroke-color, $query: $query);
 
-    @at-root {
-      @include row-fill-color(variables.$row-fill-color, $query: $query);
-      @include header-row-fill-color(variables.$header-row-fill-color, $query: $query);
-      @include selected-row-fill-color(variables.$selected-row-fill-color, $query: $query);
-      @include divider-color(variables.$divider-color, $query: $query);
-      @include divider-size(variables.$divider-size, $query: $query);
-      @include row-hover-fill-color(variables.$row-hover-fill-color, $query: $query);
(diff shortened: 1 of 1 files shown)


packages/mdc-data-table/_mixins.scss
```diff
@@ -53,22 +53,12 @@
...
-      @include divider-size(variables.$divider-size, $query: $query);
-      @include row-hover-fill-color(variables.$row-hover-fill-color, $query: $query);
-      @include header-row-text-color(variables.$header-row-text-color, $query: $query);
-      @include row-text-color(variables.$row-text-color, $query: $query);
       @include density(variables.$default-density-scale, $query: $query);
       @include cell-padding(
         $leading-padding: variables.$cell-leading-padding,
         $trailing-padding: variables.$cell-trailing-padding,
         $query: $query
       );
```
abhiomkar: Can we not also move these mixins outside root selector?

crisbeto: When I was testing it yesterday I tried moving everything out and it messed up some of the heights/paddings so I decided to play it safe.

abhiomkar: I suspect this PR changes the CSS output and possibility that it might be a breaking change. Moving all the mixins under `@at-root` outside root selector at once should produce 0 diffs.

Let me know if that is not the case. Thanks!

crisbeto: I tested it again and it seems to work. I think the first time around it broke, beause I hadn't moved it all the way outside of `.mdc-data-table`. Can you take another look?

abhiomkar: Thanks @crisbeto !

PR #5776, opened 2020-04-03
fix(data-table): update `get-header-cell-height` function body for allowing incompatible units

dutchenkoOleg:
The internal function `get-header-cell-height` does not allow me to change the unit not pixels but rem or em for `$header-cell-height` variable. 

Even if I override the variable before use it, the function still executes. As far as I understand this is feature of the preprocessor and encapsulations of `@use/@forward` directives.

Therefore, I propose to simply remove this function.

### DIFF
--- a/packages/mdc-data-table/_variables.scss
+++ b/packages/mdc-data-table/_variables.scss
@@ -45,7 +45,7 @@ $stroke-size: 1px !default;
 $stroke-color: rgba(theme-variables.prop-value(on-surface), .12) !default;
 
 $cell-height: 52px !default;
-$header-cell-height: get-header-cell-height($cell-height) !default;
+$header-cell-height: 56px !default; // $cell-height + 4
 $cell-leading-padding: 16px !default;
 $cell-trailing-padding: 16px !default;
 
@@ -59,7 +59,3 @@ $density-config: (
     minimum: $minimum-cell-height,
   )
 );
-
-@function get-header-cell-height($height) {
-  @return $height + 4px;
-}
--- a/packages/mdc-data-table/_variables.scss
+++ b/packages/mdc-data-table/_variables.scss
@@ -45,7 +45,7 @@ $stroke-size: 1px !default;
 $stroke-color: rgba(theme-variables.prop-value(on-surface), .12) !default;
 
 $cell-height: 52px !default;
-$header-cell-height: 56px !default; // $cell-height + 4
+$header-cell-height: get-header-cell-height($cell-height) !default;
 $cell-leading-padding: 16px !default;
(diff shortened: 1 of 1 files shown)


dutchenkoOleg:
Ok, I am understood the problem.
So I set condition to determinet when it need to add `+4px`.

dutchenkoOleg:
Your action `MDC Test / sass-test` is failing on `npm install` ...
But I do not touched any dependencies ))

It is the some external problems

EstebanG23:
Hi there! 

Thank you for your pull request. Can I ask for more background on the problem? I see you are trying to set the height of the header cell with units other than `px`s. 

If you are trying to set the header height you can use the `header-cell-height()` [mixin](https://github.com/material-components/material-components-web/blob/master/packages/mdc-data-table/_mixins.scss#L409) provided in the mixins.scss file of this package. 
Similarly, if you would like to set the height of the rest of the cells, you can use the `cell-height()` [mixin](https://github.com/material-components/material-components-web/blob/master/packages/mdc-data-table/_mixins.scss#L426). 

Would these mixins provide a solution for you to use? If not please let me know and we can continue to explore the solution.

dutchenkoOleg:
> I see you are trying to set the height of the header cell with units other than `px`s.

Yes, I am  tryin to rewrite all values to `rem`s.  
That is necassary for most of my projects.

I want to redeclarete Sass variables of "material"

```scss
@import "my-variables" // redeclaration of !default vars
@import "~@matetial/data-table/mdc-data-table"
```

---

@EstebanG23 
In solution you offer, I will receive redundant CSS code.

```scss
@import "~@matetial/data-table/mdc-data-table" // reproducing original data-table code
@import "mixins-to-override"                   // reproducing my code to override orignal :(
```

EstebanG23:
Understandable on the redundant code. I want to better understand the problem, could you reproduce the problem in a Codepen or Glitch project? These are some demos for easy repro settings: 
* https://glitch.com/edit/#!/remix/new-web
* https://codepen.io/abhiomkar/pen/gQWarJ

dutchenkoOleg:
I have my reasons for using `rem`. It is convenient for adaptive and accessible interfaces.
I understand the mixin approach. But the problem is not what units I use.
I just need to add a condition to the function.
If you do this, it will be very cool.

abhiomkar:
@dutchenkoOleg As mentioned in my above comment Density subsystem does not support use of units other than `px`.

Adding support for `rem` unit requires changes to Density subsystem level (But not by component level). You can use `height() / size()` Sass mixins instead as workaround.

I'm closing this PR. Please create a feature request explaining your use case to support `rem` units for Density.

PR #5805, opened 2020-04-10
feat(switch): support customizable track border color

### LINKED ISSUE #5804: [Switch] [Angular] Not clear (no pun intended) how to obtain transparent track border color
Angular Material wants to use the Switch with different colors. The mixins support this by using this set of Sass includes:

```scss
.my-rainbow-switch {
  @include mdc-switch-without-ripple($query: (structure));

  .mdc-switch {
    @include mdc-switch-toggled-on-track-color(red);
    @include mdc-switch-toggled-on-thumb-color(purple);
    @include mdc-switch-toggled-off-track-color(green);
    @include mdc-switch-toggled-off-thumb-color(blue);
  }
}
```

However, this misses one important style found in the switch's `thumb_` mixin:
```scss
@mixin track_($query) {
  ..  

  @include feature-targeting-mixins.targets($feat-color) {
    @include theme-mixins.prop(border-color, transparent);
  }

  ..
}
```

The workaround would be to call `mdc-switch-without-ripple((structure, color)` but this includes a lot of color styles that would be unnecessary and overwritten.

Perhaps we can have a separate mixin that applies the border-color like `mdc-switch-track-border-color($color)`?

andrewseguin:
Fixes #5804

### DIFF
--- a/packages/mdc-switch/_mixins.scss
+++ b/packages/mdc-switch/_mixins.scss
@@ -272,13 +272,12 @@
 @mixin track_($query: feature-targeting-functions.all()) {
   $feat-animation: feature-targeting-functions.create-target($query, animation);
   $feat-structure: feature-targeting-functions.create-target($query, structure);
-  $feat-color: feature-targeting-functions.create-target($query, color);
 
   @include feature-targeting-mixins.targets($feat-structure) {
     box-sizing: border-box;
     width: variables.$track-width;
     height: variables.$track-height;
-    border: 1px solid;
+    border: 1px solid transparent;
     border-radius: variables.$track-height / 2;
     opacity: .38;
   }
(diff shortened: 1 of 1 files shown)


packages/mdc-switch/_mixins.scss
```diff
@@ -289,10 +300,6 @@
       functions.transition(background-color),
       functions.transition(border-color);
   }
-
-  @include feature-targeting-mixins.targets($feat-color) {
```
patrickrodee, line 293: Hey Andrew, would a possible fix for this be to remove the color feature-target here? Color is an inaccurate feature for this anyway, it's more-so structure.

patrickrodee: My main thinking is the border for the track isn't designed to be themed in Material and the transparent border-color is purely for high-contrast mode, not any actual theming purposes.

andrewseguin: SGTM - I agree with the notion that it's structure (at least when compared to color or the other features)

PR #5841, opened 2020-04-20
docs(select): mdc-select a11y documentation

dbollinger:
With the removal of support for native select elements, it is now essential to provide authors with the appropriate ARIA markup to use when implementing the mdc-select component.

A few minor issues in the existing documentation:

1. Screen readers do not announce the number of items in the list.  In order to indicate how many items are in the listbox, `role="listbox"` must be on applied to the `<ul>` element, not the menu surface element.
2. The list itself is not labelled.  The element with `role="listbox"` needs to specify a label with `aria-labelledby`, because it is possible for users with assistive technology to navigate outside of the list (and focus on the listbox element itself).
3. This PR also proposes emphasizing the accessibility requirements by moving the section directly below the standard HTML example.  Without these aria attributes, the component does not work with assistive technology like screen readers, and it may be worth considering these as a requirement (would it be possible to make them a requirement, just like `mdc-select__selected-text`?


dbollinger:
Sounds good, I'll make that change 👍

dbollinger:
Oh I misread your message -- you still want an explicit accessibility section that links to the HTML section?

allan-chen:
Yes, just containing the blurb and instead of referring to a whole new template, just refer to the baseline one. Thanks again!

packages/mdc-select/README.md
```diff
@@ -78,52 +78,24 @@ The select requires that you set the `width` of the `mdc-select__anchor` element
...
-@use "@material/menu-surface/mdc-menu-surface";
-@use "@material/menu/mdc-menu";
-@use "@material/select/mdc-select";
-
-.demo-width-class {
-  width: 400px;
-}
-```
-
-### JavaScript Instantiation
```
jsalzman89: Yes

packages/mdc-select/README.md
```diff
@@ -78,52 +78,24 @@ The select requires that you set the `width` of the `mdc-select__anchor` element
...
 [Collapsible Dropdown Listbox](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html).
-The following is an example of the select component with all of the necessary aria attributes.
+
+The following example applies ARIA attributes that provide the semantic structure required for assistive technology:
 
 ```html
 <div class="mdc-select">
   <div class="mdc-select__anchor">
     <i class="mdc-select__dropdown-icon"></i>
-    <div id="demo-selected-text" class="mdc-select__selected-text" role="button" aria-haspopup="listbox" aria-labelledby="demo-label demo-selected-text">Vegetables</div>
```
jsalzman89, line 120: Role

packages/mdc-select/README.md
jsalzman89: Closed

packages/mdc-select/README.md
```diff
@@ -89,7 +89,7 @@ The following example applies ARIA attributes that provide the semantic structur
 <div class="mdc-select">
   <div class="mdc-select__anchor">
     <i class="mdc-select__dropdown-icon"></i>
-    <div id="demo-select-text" class="mdc-select__selected-text" role="button" aria-haspopup="listbox" aria-labelledby="demo-label demo-select-text">Vegetables</div>
```
jsalzman89: Selected

packages/mdc-select/README.md
```diff
@@ -122,8 +94,8 @@ The following is an example of the select component with all of the necessary ar
     <span class="mdc-line-ripple"></span>
   </div>
 
-  <div class="mdc-select__menu mdc-menu mdc-menu-surface" role="listbox">
```
jsalzman89, line 125: Option

PR #5854, opened 2020-04-23
fix(mdc-slider): avoid server side rendering error

devversion:
Recently with d10412cb24150639acc617caef1c7fac4fb6e4bd, support for pointer events has been
added to the `MdcSlider`. The feature detection for pointer events currently fails
on the server as it accesses `window` directly. We should not error if we are on the
server as these events are not having any effect anyway.

### DIFF
--- a/packages/mdc-slider/foundation.ts
+++ b/packages/mdc-slider/foundation.ts
@@ -36,11 +36,14 @@ type MoveEventMap = {
   readonly[K in DownEventType]: MoveEventType;
 };
 
-const hasPointer = !!window.PointerEvent;
+// Accessing `window` without a `typeof` check will throw on Node environments.
+const hasWindow = typeof window !== 'undefined';
+const hasPointerEventSupport = hasWindow && !!window.PointerEvent;
+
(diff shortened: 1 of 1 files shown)


packages/mdc-slider/foundation.ts
```diff
@@ -36,11 +36,14 @@ type MoveEventMap = {
   readonly[K in DownEventType]: MoveEventType;
 };
 
-const hasPointer = !!window.PointerEvent;
+// Accessing `window` without a `typeof` check will throw on Node environments.
+const hasWindow = typeof window !== 'undefined';
+const hasPointerEventSupport = hasWindow && !!window.PointerEvent;
```
abhiomkar, line 41: This looks good to me.

I think the right fix would be to add adapter method to check if window object has `PointerEvent`. Ideally foundation business logic shouldn't directly access DOM related objects.

WDYT? /cc @e111077

devversion: I would agree. I didn't know whether there was a reason to have it set up there. i.e. not having to construct a new array on every foundation creation.

PR #5863, opened 2020-04-27
docs(line-ripple): update README for active/inactive color change

mroskamp:
Change Sass mixin listing to include both active-color() and inactive-color(), instead of the deprecated color() mixin

### DIFF
--- a/packages/mdc-line-ripple/README.md
+++ b/packages/mdc-line-ripple/README.md
@@ -63,7 +63,8 @@ CSS Class | Description
 
 Mixin | Description
 --- | ---
-`color($color)` | Customizes the color of the line ripple when active.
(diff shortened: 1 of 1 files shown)


mroskamp:
@googlebot I signed it!

packages/mdc-line-ripple/README.md
```diff
@@ -63,7 +63,8 @@ CSS Class | Description
 
 Mixin | Description
 --- | ---
-`color($color)` | Customizes the color of the line ripple when active.
+`active-color($color)` | Customizes the color of the line ripple when active.\
```
asyncLiz, line 66: Small nit: remove the trailing slash here, I don't believe it's necessary

```suggestion
`active-color($color)` | Customizes the color of the line ripple when active.
```

PR #6004, opened 2020-05-29
chore: Added support to run demos for local development

abhiomkar:
See [local development document](https://github.com/material-components/material-components-web/blob/891b5c03bd05586618c41768385ceb8b9bae4d9a/docs/local-development.md) for preview.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ node_modules/
 
 # Build output
 /build/
+/dist/
 /coverage/
 packages/*/dist
 packages/**/*.js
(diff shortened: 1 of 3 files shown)


asyncLiz:
Is the idea that we'll actually be creating these demo folders and checking them in? Or is this just a recommendation for 3P/local dev?

abhiomkar:
We'll not create demos folder (We can leverage Web Components implementation for demos in future).

This document is for third-party contributors (created this document for intern).

PR #6032, opened 2020-06-04
chore: Removed number-leading-zero from Sass lint rule

abhiomkar:
Revert this change again.


allan-chen:
Closing since bfe957ca7dc7c0584ead080dbbf8494cbc8dd47e fixes. Hopefully we don't have to open again.

Issue #2270, opened 2018-02-18
Possibility for empty cells in grid to take up no space


PeteCrighton:
In order to center elements in a grid on all devices, it would be good to have the ability to declare something like `<div class="mdc-layout-grid__cell--span-0-<TYPE_OF_DEVICE>"></div>` that takes up no space at all.
See example below where I have several cards in a layout grid. If there are not enough cards to fill a row, I want to center them on the page and have all that flexible on every type of device. Here it works for desktop and mobile (sort of – I assume I will be able to correctly adjust the gutter around empty grid cells). But I can’t get tablet to work – `mdc-layout-grid__cell--span-12-tablet` helps for an empty row (again, gutter needs adjusting), but I see no way to make empty cells that are needed for desktop disappear for tablet.

```php
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <link rel="stylesheet"
        href="https://unpkg.com/material-components-web@latest/dist/material-components-web.min.css">
</head>
<body class="mdc-layout-grid">
  <div class="mdc-layout-grid__inner">
    <?php $array = array(1); ?>
    <?php /* $array = array(1, 2); */ ?>
    <?php /* $array = array(1, 2, 3); */ ?>
    <?php /* $array = array(1, 2, 3, 4); */ ?>
    <?php /* $array = array(1, 2, 3, 4, 5); */ ?>
    <?php /* $array = array(1, 2, 3, 4, 5, 6); */ ?>
    <?php $index = 0; ?>
    <?php foreach($array as $a): ?>

      <?php if(count($array)==1 && $index==0): ?>
        <div class="mdc-layout-grid__cell
                    mdc-layout-grid__cell--span-4-desktop
                    mdc-layout-grid__cell--span-2-tablet"></div>
      <?php endif ?>
      <?php if(count($array)==2 && $index==0): ?>
        <div class="mdc-layout-grid__cell
                    mdc-layout-grid__cell--span-2-desktop
                    mdc-layout-grid__cell--span-12-tablet"></div>
      <?php endif ?>
      <?php if(count($array)==3 && $index==0): ?>

      <?php endif ?>
      <?php if(count($array)==4 && $index==3): ?>
        <div class="mdc-layout-grid__cell
                    mdc-layout-grid__cell--span-4-desktop"></div>
      <?php endif ?>
      <?php if(count($array)==5 && $index==3): ?>
[...]

PR #6049, opened 2020-06-09
fix(layout-grid): "base" span class specificity

### LINKED ISSUE #2228: Grid cell styles for tablet don't take precedence over default
0.29.0
Chrome
### What OS are you using?
MacOS Sierra
https://codepen.io/anon/pen/eVWYvX
When resizing the window to a tablet size the spans should drop from 4:4:4 to 2:4:2 maintaining a single row (tablet has an 8 grid).
Because the span-4 style takes precedence over the span-2-tablet style the tablet styling is not respected and the row wraps

SampsonCrowley:
"base" span classes should not be more specific than device-specific spans.
fixes #2228 

also added `span-0` classes to implement #2270

### DIFF
--- a/packages/mdc-layout-grid/_mixins.scss
+++ b/packages/mdc-layout-grid/_mixins.scss
@@ -80,19 +80,24 @@
     @error "Invalid style specified! Choose one of #{map.keys(variables.$columns)}";
   }
 
-  $percent: math.percentage($span / map.get(variables.$columns, $size));
+  @if $span == 0 {
+    display: none;
+  } @else {
+    $percent: math.percentage($span / map.get(variables.$columns, $size));
 
-  @if $percent > 100% {
-    $percent: 100%;
-  }
+    @if $percent > 100% {
+      $percent: 100%;
+    }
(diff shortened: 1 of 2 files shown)


SampsonCrowley:
@googlebot I signed it!

patrickrodee:
Hey Sampson, thanks for your PR! There are more changes in here than just layout-grid fixes (comment changes, decimal changes, etc). Let's revert those and keep this scoped to just the layout-grid fixes.

Let me know if you have any questions!

SampsonCrowley:
@patrickrodee without those "changes", the system doesn't build because they violate the linting rules

chasingduende:
@SampsonCrowley has this effort been abandoned?  It seems like the only fix having grid spacing...

SampsonCrowley:
@chasingduende it wasn't responded to last time I jumped through the hoops to get things merged in, even though it was definitely something that was needed and all tests were passing.

So yeah I haven't looked at it again. Feel free to steal it and claim the credit though

PR #6067, opened 2020-06-14
fix(data-table): not inverting text alignment in rtl

crisbeto:
The table cell has `text-align: left` which means that it won't invert automatically in RTL. These changes add the inversion, similar to the one on the table headers.


asyncLiz:
@abhiomkar did this get imported in?

abhiomkar:
Seems like a glitch with Copybara sync. This was [imported](cl/317104384) & [merged](https://github.com/material-components/material-components-web/commit/bd8d1aafab5c6da614135702f5814447de5ea448) to master.

PR #6151, opened 2020-06-26
fix(snackbar): remove use of @at-root

andrewseguin:
Same type of changes as https://github.com/material-components/material-components-web/pull/5751

The Angular Material styles add a dark-themed snackbar since the snackbar styles use at-root.


joyzhong:
LGTM, importing changes via Copybara.

PR #6172, opened 2020-07-01
fix(circular-progress): Default all variables

zposten:
Every other package !default's all of their variables, giving consumers the ability to customize them.

Circular Progress for some reason only currently !default's one, not giving full flexibility to consumers.


zposten:
@allan-chen Sorry about that, should be fixed now

allan-chen:
@zposten Thanks! Syncing to internal

PR #6179, opened 2020-07-02
fix(circular-progress): Correctly position circle-right

zposten:
The circular-progress animation does not currently work without this change.  Being relatively positioned from the circle-clipper class, the circle-right goes off to the side instead of being on top of the circle-left.

Additionally, the README for the circular-progress component looks very different, and is missing some of the information provided by all other MDC READMEs. This PR tries to fix that too.

### DIFF
--- a/packages/mdc-circular-progress/_mixins.scss
+++ b/packages/mdc-circular-progress/_mixins.scss
@@ -146,8 +146,10 @@
   }
 
   .mdc-circular-progress__circle-right {
-    .mdc-circular-progress__indeterminate-circle-graphic {
-      @include feature-targeting-mixins.targets($feat-structure) {
+    @include feature-targeting-mixins.targets($feat-structure) {
+      position: absolute;
+
+      .mdc-circular-progress__indeterminate-circle-graphic {
         /* @noflip */
         left: -100%;
       }


zposten:
I split out my documentation changes into a different PR.  Those could be seen as somewhat subjective, whereas this fix is needed immediately.

allan-chen:
Hi Zach,

Thank you for the PR. What is the issue? Feel free to fork the codepen below to reproduce:

https://codepen.io/allanchenmdc/pen/MWKEqrK

allan-chen:
Taking a guess here, but I believe any layout issue you may be having is caused by introducing space between the ending and starting `</div><div>` tags in the markup. See this [link](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) for why that markup is intentional. Please feel free to let me know the issue you originally intended to bring up is different.

zposten:
Well shoot.  You're completely right.  Putting gap-patch on its own line caused the effect that I saw, and looking back more closely at the README, that was already called out.  

This is just my opinion at this point, but I feel that especially for something meant to be used as broadly as MDC is, relying on a markup quirk to position things correctly is a bad idea when the problem can be so easily fixed with CSS.

PR #6185, opened 2020-07-06
revert: "Include tooltip directory for future copybara syncs."

abhiomkar:
Reverts material-components/material-components-web#6182


sayris:
should css-bundle-factory.js also get reverted?

PR #6198, opened 2020-07-08
feat(segmented-button): Added foundation business logic

gschrag:


packages/mdc-segmented-button/segmented-button/test/foundation.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  it('if single select, #handleSelected reselects segment if it was unselected', () => {
+
+  });
+
+  it('if multi select, #handleSelected changes nothing if segment is selected', () => {
+
+  });
+
+  //? tests ability to unselect everything...this is not an obscure nor problematic edge case
+  it('if multi select, #handleSelected changes nothing if segment is unselected', () => {
```
gschrag, line 162: Are these tests too specific? I don't want to miss any test cases but am worried that I'm including unnecessary tests.

packages/mdc-segmented-button/segment/foundation.ts
```diff
@@ -42,22 +42,37 @@ export class MDCSegmentedButtonSegmentFoundation extends MDCFoundation<MDCSegmen
   }
 
   isSelected(): boolean {
-    return false;
+    return this.adapter.hasClass('mdc-segmented-button__segment--selected');
   }
 
   setSelected() {
```
abhiomkar, line 48: Can this accept a boolean (we can replace `setUnselected`). WDYT?

packages/mdc-segmented-button/segment/foundation.ts
```diff
@@ -42,22 +42,37 @@ export class MDCSegmentedButtonSegmentFoundation extends MDCFoundation<MDCSegmen
...
 
   setUnselected() {
-    return;
+    this.adapter.removeClass('mdc-segmented-button__segment--selected');
+    this.setAriaAttr('false');
   }
 
   getSegmentId(): string {
-    return '';
+    return this.adapter.getAttr('id');
```
abhiomkar, line 59: Are we not using data attribute instead of HTML `id` attribute for segment id?

packages/mdc-segmented-button/segment/foundation.ts
```diff
@@ -42,22 +42,37 @@ export class MDCSegmentedButtonSegmentFoundation extends MDCFoundation<MDCSegmen
...
   }
 
-  handleClick() {
-    return;
+  handleClick(): void {
+    if (this.isSelected()) {
```
abhiomkar, line 63: Are we adding logic only for multi-select in this PR?

packages/mdc-segmented-button/segmented-button/test/foundation.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    let selectedSegment = mockAdapter.getSegments()[SELECTED];
+    let selectedSegments = foundation.getSelectedSegments();
+    
+    expect(selectedSegments.length).toEqual(1);
+    expect(selectedSegments[0].index).toEqual(selectedSegment.index);
+    expect(selectedSegments[0].selected).toBeTruthy();
+    expect(selectedSegments[0].segmentId).toEqual(selectedSegment.segmentId);
+  });
+
+  /*** #selectSegment ***/
```
abhiomkar, line 133: Use `describe()` function to group tests. 

See data table's foundation test as an example.

packages/mdc-segmented-button/segment/foundation.ts
gschrag: That's how I had it originally, but I changed it per Jeremy Elbourn's comment on the design doc:
```
I generally discourage APIs with boolean flag arguments:
https://martinfowler.com/bliki/FlagArgument.html

Might be worth considering separate methods for `select` and `deselect`
```

packages/mdc-segmented-button/segment/foundation.ts
gschrag: Oh yeah, I meant to put `segmentId` instead of `id`. But I wasn't familiar with the difference between a data attribute and a regular attribute when I wrote this. Do you think it's ok to have the Component handle the difference when implementing the Adapter, or should there be a Adapter.getCustomAttr or Adapter.getDataAttr method?

packages/mdc-segmented-button/segment/foundation.ts
gschrag: Yes. The child segment changes status automatically when it's clicked and notifies the parent segment-button. Then the parent-segment button handles the single-/multi-select logic and modifies the child segments as needed.

packages/mdc-segmented-button/segment/foundation.ts
abhiomkar: I think foundation should be opinionated about which attribute name to fetch. It can use `data-segment-id` directly instead of `id`.

packages/mdc-segmented-button/segment/foundation.ts
abhiomkar: Single select segmented button can't be toggled. It should behave similar to radio group. For example, click on segment button that is already selected should not unselect it (if it is single select).

I think segment foundation can be of two types (one that toggles & another that does not toggle)

packages/mdc-segmented-button/segmented-button/test/foundation.test.ts
gschrag: Thanks for the example!

packages/mdc-segmented-button/segment/foundation.ts
gschrag: I had it to unselecting a segment in a single select segmented button would unselect it, then reselect it to be functionally the same as not being unselected, but I can see why this could have drawbacks. Since the Adapter already has a isSingleSelect method I can use that to keep the segment from unselecting.

packages/mdc-segmented-button/segment/foundation.ts
abhiomkar: Got it! Separate method for select / unselect sounds good..

packages/mdc-segmented-button/segment/foundation.ts
e111077: This can be handled by the parent; it fires an event upward and the parent can do the logic of whether it should unselect or not by calling the adapter to select this one. That would obviate the need for `isSingleSelect`

packages/mdc-segmented-button/segmented-button/foundation.ts
```diff
@@ -40,27 +40,38 @@ export class MDCSegmentedButtonFoundation extends MDCFoundation<MDCSegmentedButt
...
-    return false;
+    return this.adapter.hasClass('mdc-segmented-button--single-select');
   }
 
-  handleSelected(_detail: SegmentDetail) {
-    return;
+  handleSelected(detail: SegmentDetail) {
+    if (this.isSingleSelect()) {
+      let selectedSegments = this.getSelectedSegments();
+      if (detail.selected) {
```
e111077, line 67: can you pull this out into its own private method to prevent nested conditionals?

packages/mdc-segmented-button/segmented-button/foundation.ts
```diff
@@ -40,27 +40,38 @@ export class MDCSegmentedButtonFoundation extends MDCFoundation<MDCSegmentedButt
...
+        selectedSegments.filter(segmentDetail => segmentDetail.index !== detail.index)
```
e111077, line 68: please use `for of` loop as `filter(...).forEach(...)` creates an unnecessary array

packages/mdc-segmented-button/segment/foundation.ts
abhiomkar: Selecting a segment which is already selected in Single select mode does should not change color and also should not unselect it. But it can trigger an event.

I would keep the save behaviour as [radio group](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html).

A segment can be toggled only for multi-select. I think it makes sense to set this boolean from parent depending on single or multi-select.

gschrag: `isSingleSelect` is also used to determine which aria-pattern to use, which seemed for appropriate for the child to control but could be handled by the parent, thus removing the need for `isSingleSelect`.
If we keep `isSingleSelect` and the child doesn't allow itself to be unselected when the parent is single-select then the parent doesn't have to check whether to re-select the child. The parent only has to unselect the previously selected child.

packages/mdc-segmented-button/segmented-button/foundation.ts
gschrag: If we have the child segment keep itself from being unselected then I can delete the `if (detail.selected)` logic. If we leave it up to the parent to take care of this, I'll be sure to pull this logic into a helper method!

packages/mdc-segmented-button/segment/foundation.ts
```diff
@@ -42,22 +42,37 @@ export class MDCSegmentedButtonSegmentFoundation extends MDCFoundation<MDCSegmen
...
 
   getSegmentId(): string {
-    return '';
+    return this.adapter.getAttr('data-segment-id');
   }
 
-  handleClick() {
-    return;
+  handleClick(): void {
+    if (!this.isSelected()) {
```
abhiomkar, line 63: Can we simplify this?

```ts
if (this.adapter.isSingleSelect()) {
  this.setSelected();  
} else {
  this.toggleSelection();
}

...

toggleSelection() {
  if (this.isSelected()) {
    this.setUnselected();
  } else {
    this.setSelected();
  }
}
```

packages/mdc-segmented-button/segment/test/foundation.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      'addClass',
+      'removeClass',
+      'hasClass',
+      'notifySelectedChange'
+    ]);
+  });
+
+  // This value is arbitrary
+  const SEGMENT_ID = 'segment0';
+  const SELECTED_CLASS = 'mdc-segmented-button__segment--selected';
```
abhiomkar, line 43: nit: define this in `constants.ts` file in `cssClasses` object.

packages/mdc-segmented-button/segment/test/foundation.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    it('#setUnselected removes the `selected` css class and if not singleSelect then sets aria-pressed to false', () => {
+      const {foundation, mockAdapter} = setupSelectedTest();
+      foundation.setUnselected();
+
+      expect(mockAdapter.removeClass).toHaveBeenCalledWith(SELECTED_CLASS);
+      expect(mockAdapter.setAttr).toHaveBeenCalledWith('aria-pressed', 'false');
+    });
+  });
+
+  describe('Click handling', () => {
```
abhiomkar, line 119: It would be useful to group this by feature instead of method names.

For example, `single selection`, `multi selection`, etc.

packages/mdc-segmented-button/segmented-button/foundation.ts
```diff
@@ -40,27 +40,39 @@ export class MDCSegmentedButtonFoundation extends MDCFoundation<MDCSegmentedButt
...
 
   getSelectedSegments(): readonly SegmentDetail[] {
-    return [];
+    return this.adapter.getSegments().filter(segmentDetail => segmentDetail.selected);
   }
 
-  isSegmentSelected(_indexOrSegmentId: number | string): boolean {
-    return false;
+  isSegmentSelected(indexOrSegmentId: number | string): boolean {
+    let segment = this.adapter.getSegments().filter(segmentDetail => segmentDetail.index === indexOrSegmentId || segmentDetail.segmentId === indexOrSegmentId);
```
abhiomkar, line 56: You can also use `.some()` API (this'll exit as soon as segment index or id is found).

```ts
isSegmentSelected(indexOrSegmentId: number | string): boolean {
  return this.adapter.getSegments().some(
     segmentDetail => segmentDetail.index === indexOrSegmentId || segmentDetail.segmentId === indexOrSegmentId);
}
```

gschrag:
This PR should be ready to merge.

abhiomkar:
@e111077 Can you approve this PR?

packages/mdc-segmented-button/segmented-button/constants.ts
```diff
@@ -0,0 +1,38 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+export const indices = {
```
abhiomkar, line 24: This should use `enum`.

Also, exported entities should have [TSdoc](https://github.com/microsoft/tsdoc) (code documentation).

packages/mdc-segmented-button/segmented-button/constants.ts
```diff
@@ -0,0 +1,38 @@
...
+  NOT_PRESENT: -1,
+  SELECTED: 0,
```
abhiomkar, line 26: nit: Can we set 1 (truthy value) to `SELECTED`?

packages/mdc-segmented-button/segmented-button/constants.ts
```diff
@@ -0,0 +1,38 @@
...
+  UNSELECTED: 1
+}
+
+export const strings = {
+  NOT_PRESENT_SEGMENT_ID: 'segment-1',
```
abhiomkar, line 31: Shouldn't the segment id be set by client instead?

gschrag: I only plan on using these strings for the unit tests. I can rename it to `test_strings` to be more clear.

abhiomkar: Got it! This should be moved to respective test files if it is only intended for unit tests.

PR #6222, opened 2020-07-14
feat(segmented-button): Add component outlines

gschrag:
Add method outlines for segment and segmented-button components


kseamon:
LGTM

PR #6224, opened 2020-07-15
fix(list): removed unnecessary setTimeout method

akaSourav:
You are using the `setTimeout` method to wait and check that next focused element is inside the list or not.
But for `focusout` event [FocusEvent.relatedTarget](https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget) is the `EventTarget` which is receiving focus. So you can use `FocusEvent.relatedTarget` instead of the `setTimeout` method to check the focus is inside the list.


Sourav-21:
@googlebot I signed it!

PR #6271, opened 2020-07-27
fix(segmented-button): Fixed unit test in IE11

abhiomkar:


packages/mdc-segmented-button/segmented-button/test/component.test.ts
```diff
@@ -58,13 +58,13 @@ const setupTest = () => {
 };
 
 const setAllSelected = (els: NodeListOf<Element>) => {
-  els.forEach((el) => {
+  [].slice.call(els).forEach((el: Element) => {
```
joyzhong, line 61: We should be able to use Array#from now, jfyi!

abhiomkar: That's much better! Thanks..

Done.

PR #6277, opened 2020-07-27
feat(segmented-button): Add ripple and touch-target support

gschrag:
Add ripple effects and support touch-target wrapper div


packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -54,16 +75,44 @@
...
   // Style for when element is selected
   .mdc-segmented-button__segment--selected {
     @include _selected($query);
+  }
+
+  // maintains visual design when inside a touch-target-wrapper
+  // This could also be accomplished with --first/--last classes
+  .mdc-touch-target-wrapper {
+    .mdc-segmented-button__segment {
+      border-radius: 0px;
```
kseamon, line 100: omit unit for 0 value

packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -54,16 +75,44 @@
...
+      border-right-width: 0;
+    }
 
-    // Avoids double borders
-    + .mdc-segmented-button__segment {
+    &:first-child .mdc-segmented-button__segment {
       @include feature-targeting.targets($feat-structure) {
-        border-left: hidden;
+        border-radius: segment-theme.$border-radius 0px 0px
```
kseamon, line 106: also here

kseamon: and more places below - not going to comment again

packages/mdc-segmented-button/segment/component.ts
```diff
@@ -37,22 +40,41 @@ export type MDCSegmentedButtonSegmentFactory =
...
+  private rippleComponent!: MDCRipple; // assigned in initialize
   private handleClick!:
       SpecificEventListener<'click'>; // assigned in initialSyncWithDOM
 
+  initialize(
+    rippleFactory: MDCRippleFactory =
+      (el, foundation) => new MDCRipple(el, foundation)
+  ) {
+    const rippleAdapter: MDCRippleAdapter = {
+      ...MDCRipple.createAdapter(this),
```
kseamon, line 62: @ngwattcos is this going to be an issue with your adaptor changes?

gschrag: Just fyi, I used mdc-chips/chip as a reference for the ripple TS/SCSS code.

packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -54,16 +75,44 @@
...
       }
     }
+
+    .mdc-segmented-button__touch {
+      @include touch-target-mixins.touch-target($query: $query);
+    }
+
+    .mdc-segmented-button--touch {
+      @include touch-target-mixins.margin(
+        $component-height: 32px,
```
gschrag, line 85: Should the touch target height be the same or larger than the segment's height?

packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -40,30 +54,65 @@
...
+        $query: $query
+      );
+    }
```
gschrag, line 88: These should probably be renamed to `.mdc-segmented-button__segment__touch` and `.mdc-segmented-button__segment--touch`.

gschrag:
I've noticed an issue where the background color of a selected segment causes the border color to be darker (because it's mostly transparent). Does anyone know how to keep the colors from overlapping?

packages/mdc-segmented-button/segment/component.ts
ngwattcos: @kseamon @gschrag Not that I'm aware of. I did remove the default adapter implementation because I assumed it would receive a class or object that fully implements MDCRippleAdapter.

packages/mdc-segmented-button/segment/_segment.scss
darrellgaspar: By "touch target" would it actually affect the "touchable" area that the button has to register a click/touch? If that's the case, I would make the touch target be the same height as the segment so there are no "accidental" clicks/touches.

gschrag: I don't fully understand how the touch target is supposed to work, but I think it's meant to expand the touchable area for mobile or users who have trouble clicking accurately.

packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -77,28 +126,57 @@
   }
 }
 
+@mixin ripple($query: feature-targeting.all()) {
```
abhiomkar, line 129: Move this to `_segment-ripple.scss` file and include with and without ripple styles in `/segment/styles.scss`

packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -24,12 +24,26 @@
 
 @use '@material/feature-targeting';
 @use '@material/typography/mixins' as typography-mixins;
+@use "@material/touch-target/mixins" as touch-target-mixins;
```
abhiomkar, line 27: Importing from `_mixins.scss` files are now deprecated. Most of the MDC base components (typography, touch-target, etc) have core sass modules that you can use them directly.

e.g., `@material/typography/typography`

packages/mdc-segmented-button/segment/_segment.scss
abhiomkar: `.mdc-segmented-button__segment--touch` (Consistent with `mdc-button--touch`)

packages/mdc-segmented-button/segment/_segment.scss
abhiomkar: Touch target height should be same as MDC button's touch target height (i.e., 48px)

abhiomkar: See https://material.io/develop/web/components/touchtarget

abhiomkar:
Let me know when this PR is ready for re-review. Thanks!

packages/mdc-segmented-button/segment/_segment.scss
gschrag: A couple, like elevation and ripple, have `elevation/elevation` and `elevation/elevation-theme`. Is it better to use both or just use `elevation`?

packages/mdc-segmented-button/segment/_segment-ripple.scss
```diff
@@ -0,0 +1,55 @@
...
+@use '@material/ripple/ripple';
+@use '@material/ripple/ripple-theme';
+@use './segment-theme';
+
+$ripple-target: '.mdc-segmented-button__ripple';
+
+@mixin ripple($query: feature-targeting.all()) {
+  $feat-structure: feature-targeting.create-target($query, structure);
+
+  @include ripple.common($query);
```
abhiomkar, line 33: We can let clients import common ripple styles manually when using this library. We should update docs to mandate including this styles before using segmented button.

Let's remove this line.

packages/mdc-segmented-button/segment/_segment-theme.scss
```diff
@@ -45,16 +44,21 @@ $border-radius: 4px;
...
+  );
+
   .mdc-segmented-button__segment {
-    $property: custom-properties.create(
-      --mdc-segmented-button-outline-color,
-      $color
-    );
-    @include _outline-color($property, $query);
+    @include feature-targeting.targets($feat-color) {
+      @include custom-properties.apply('border-color', $property);
```
abhiomkar, line 60: replace all `.apply()` call with `theme.property()`. I recently learnt that .apply() does not resolve tokens.

packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -22,14 +22,21 @@
...
-@use '@material/typography/mixins' as typography-mixins;
+@use '@material/feature-targeting/feature-targeting';
+@use '@material/typography/typography';
+@use '@material/touch-target/touch-target';
+@use '@material/elevation/elevation';
+@use '@material/elevation/elevation-theme';
 @use './segment-theme';
 
-@mixin core-styles($query: feature-targeting.all()) {
+@mixin without-ripple($query: feature-targeting.all()) {
```
abhiomkar, line 32: Since we've moved with-ripple() styles to separate module let's rename this back to `core-styles()`

gschrag: Then in `segment/styles.scss` should I keep the mixin as `core-styles`? Or should I remove `segment/styles.scss` and just put those 2 includes in `./styles.scss`?

packages/mdc-segmented-button/segment/styles.scss
```diff
@@ -0,0 +1,30 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+@use '@material/feature-targeting/feature-targeting';
+@use './segment';
+@use './segment-ripple';
+
+@mixin core-styles($query: feature-targeting.all()) {
```
abhiomkar, line 27: This Sass module should emit CSS.

Directly include below 2 mixins outside this mixin.

gschrag: So should `segment/` and `segmented-button/` each have a `styles.scss` file emitting css, then `./styles.scss` just includes `@use` for these `styles.scss` files?

abhiomkar: Child component should not have `styles.scss` file. This component should have a single entry point.

gschrag: Ok, I think I misunderstood one of your previous comments:

> Move this to `_segment-ripple.scss` file and include with and without ripple styles in `/segment/styles.scss`


So there should only be `./styles.scss` that includes the `core-styles()` and `ripple()` mixins from `segmented-button/_segmented-button.scss`, `segment/_segment.scss`, and `segment/_segment-ripple.scss`?

abhiomkar: Sorry, I meant `segmented-button/styles.scss` in my above comment.

yes, that's correct.

packages/mdc-segmented-button/segment/_segment-ripple.scss
gschrag: I've updated the README

PR #6380, opened 2020-08-13
fix(segmented-button): Move include statements to avoid nested classes

gschrag:
Remove use of segment-theme mixins with references to css classes inside of references to css classes.

### DIFF
--- a/packages/mdc-segmented-button/segment/_segment.scss
+++ b/packages/mdc-segmented-button/segment/_segment.scss
@@ -37,8 +37,8 @@
   @include touch-target.wrapper($query);
   @include elevation.overlay-common($query);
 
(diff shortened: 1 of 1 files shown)


packages/mdc-segmented-button/segment/_segment.scss
```diff
@@ -139,7 +135,8 @@
   }
 }
 
-@mixin _unselected($query: feature-targeting.all()) {
+@mixin _unselected-themes($query: feature-targeting.all()) {
```
abhiomkar, line 138: Remove `-themes` suffix. Same for below mixin.

gschrag:
I don't know why the unit tests are failing. I'm not getting useful error messages through Github nor locally.

PR #6381, opened 2020-08-13
feat(segmented-button): Add select validations for singleSelect

gschrag:
Unselect extra initial selected segments if singleSelect and select first segment if none already selected.


packages/mdc-segmented-button/segmented-button/component.ts
```diff
@@ -69,6 +69,14 @@ export class MDCSegmentedButton extends MDCComponent<MDCSegmentedButtonFoundatio
       segment.setIndex(index);
       segment.setIsSingleSelect(isSingleSelect);
     });
+
+    const selectedSegments = this.segments_.filter((segment) => segment.isSelected());
```
abhiomkar, line 73: I think it is not required modify the DOM to force select single segmented on initialization.

It might be helpful if component throws error when a component has multi selected segments for single select variant.

WDYT?

gschrag: Throwing an error makes sense so that we don't surprise the user with potentially unintuitive behavior.

PR #6390, opened 2020-08-15
fix(data-table): Check if data table has checkboxes on destroy

### LINKED ISSUE #6385: [DataTable] destroy() exception
## Bug report
The destroy method of MDCDataTable doesn't check if headerRowCheckbox is set before calling destroy on it, this causes an exception when not using row selection:

```
MDCDataTable.prototype.destroy = function () {
    this.headerRow.removeEventListener('change', this.handleHeaderRowCheckboxChange);
    this.headerRow.removeEventListener('click', this.headerRowClickListener);
    this.content.removeEventListener('change', this.handleRowCheckboxChange);
    this.headerRowCheckbox.destroy();
    this.rowCheckboxList.forEach(function (checkbox) {
        checkbox.destroy();
    });
};
```

Docment says "Instantiating MDCDataTable component is only required to add interactions for example, row selection.", but client needs to instantiate JS to use data table with column sorting.

### Steps to reproduce
1. Create a Data Table that doesn't use row selection, but it uses column sorting.
2. Call destroy on the data table
3. Exception is thrown

### Actual behavior
Exception is thrown

### Expected behavior
No exception should be thrown

### Possible solution
Check if it exists before calling destroy.

YuoMamoru:
fix(data-table): Check if data table has checkboxes on destroy

Fix #6385

### DIFF
--- a/packages/mdc-data-table/component.ts
+++ b/packages/mdc-data-table/component.ts
@@ -58,21 +58,27 @@ export class MDCDataTable extends MDCComponent<MDCDataTableFoundation> {
   initialSyncWithDOM() {
     this.headerRow =
         this.root.querySelector(`.${cssClasses.HEADER_ROW}`) as HTMLElement;
-    this.handleHeaderRowCheckboxChange = () =>
-        this.foundation.handleHeaderRowCheckboxChange();
-    this.headerRow.addEventListener(
-        'change', this.handleHeaderRowCheckboxChange);
+    if (this.headerRow.querySelector(selectors.HEADER_ROW_CHECKBOX)) {
+      this.handleHeaderRowCheckboxChange = () =>
+          this.foundation.handleHeaderRowCheckboxChange();
+      this.headerRow.addEventListener(
+          'change', this.handleHeaderRowCheckboxChange);
+    }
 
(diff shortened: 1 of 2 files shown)


YuoMamoru:
@googlebot I signed it!

packages/mdc-data-table/component.ts
```diff
@@ -135,10 +135,14 @@ export class MDCDataTable extends MDCComponent<MDCDataTableFoundation> {
     this.headerRow.removeEventListener('click', this.headerRowClickListener);
     this.content.removeEventListener('change', this.handleRowCheckboxChange);
 
-    this.headerRowCheckbox.destroy();
-    this.rowCheckboxList.forEach((checkbox) => {
-      checkbox.destroy();
-    });
+    if (this.headerRowCheckbox) {
```
abhiomkar, line 138: Thanks for sending the pull request! Can you also update `initialSyncWithDOM()` method to add event listener only when row checkboxes are present?

YuoMamoru:
I updated `initialSyncWithDOM()` method to add an event listener only when target elements are present.
How about this?

YuoMamoru:
I added a test for a case where an exception was thrown.

PR #6430, opened 2020-08-25
fix(dialog): support deep textareas

### LINKED ISSUE #6429: [dialog] ignoring enter key on deep elements
In the dialog foundation we have this:
https://github.com/material-components/material-components-web/blob/58eaa9f027bb7ced126d3fe97cab5a0f627eb098/packages/mdc-dialog/foundation.ts#L208-L212

Essentially to prevent the enter key from triggering the primary action if you're in a "special" element ([here](https://github.com/material-components/material-components-web/blob/58eaa9f027bb7ced126d3fe97cab5a0f627eb098/packages/mdc-dialog/constants.ts#L47-L50) defined as a textarea or a list-item).

This works in most cases but doesn't account for event retargeting.. so if we have a textarea in a shadow root, inside the dialog, it will not be detected as one of these "special cases" and will trigger the primary action still when you press enter.

Am i ok to get a PR together which changes this to use the composed path instead of the event target?

43081j:
Fixes #6429.

If this isn't the right solution, please do let me know and ill update or abandon.

### DIFF
--- a/packages/mdc-dialog/foundation.ts
+++ b/packages/mdc-dialog/foundation.ts
@@ -69,6 +69,7 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
   private scrimClickAction_ = strings.CLOSE_ACTION;
   private autoStackButtons_ = true;
   private areButtonsStacked_ = false;
+  private suppressedPressSelector: string = strings.SUPPRESS_DEFAULT_PRESS_SELECTOR;
 
   constructor(adapter?: Partial<MDCDialogAdapter>) {
     super({...MDCDialogFoundation.defaultAdapter, ...adapter});
@@ -167,6 +168,14 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
     this.autoStackButtons_ = autoStack;
   }
 
+  getSuppressedPressSelector(): string {
+    return this.suppressedPressSelector;
+  }
+
+  setSuppressedPressSelector(selector: string): void {
+    this.suppressedPressSelector = selector;
+  }
+
   layout() {
     if (this.layoutFrame_) {
       cancelAnimationFrame(this.layoutFrame_);
@@ -207,7 +216,7 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
 
     const target = evt.composedPath ? evt.composedPath()[0] : evt.target;
     const isDefault = !this.adapter.eventTargetMatches(
-        target, strings.SUPPRESS_DEFAULT_PRESS_SELECTOR);
+        target, this.suppressedPressSelector);
     if (isEnter && isDefault) {
       this.adapter.clickDefaultButton();
     }
--- a/packages/mdc-dialog/foundation.ts
+++ b/packages/mdc-dialog/foundation.ts
@@ -205,8 +205,9 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
       return;
     }
 
+    const target = evt.composedPath()[0];
     const isDefault = !this.adapter.eventTargetMatches(
-        evt.target, strings.SUPPRESS_DEFAULT_PRESS_SELECTOR);
+        target, strings.SUPPRESS_DEFAULT_PRESS_SELECTOR);
     if (isEnter && isDefault) {
       this.adapter.clickDefaultButton();
     }
--- a/packages/mdc-dialog/foundation.ts
+++ b/packages/mdc-dialog/foundation.ts
@@ -69,7 +69,7 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
   private scrimClickAction_ = strings.CLOSE_ACTION;
   private autoStackButtons_ = true;
   private areButtonsStacked_ = false;
-  private suppressedPressSelector: string = strings.SUPPRESS_DEFAULT_PRESS_SELECTOR;
+  private suppressDefaultPressSelector = strings.SUPPRESS_DEFAULT_PRESS_SELECTOR;
 
   constructor(adapter?: Partial<MDCDialogAdapter>) {
     super({...MDCDialogFoundation.defaultAdapter, ...adapter});
@@ -168,12 +168,12 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
     this.autoStackButtons_ = autoStack;
   }
 
-  getSuppressedPressSelector(): string {
-    return this.suppressedPressSelector;
+  getSuppressDefaultPressSelector(): string {
+    return this.suppressDefaultPressSelector;
   }
 
-  setSuppressedPressSelector(selector: string): void {
-    this.suppressedPressSelector = selector;
+  setSuppressDefaultPressSelector(selector: string) {
+    this.suppressDefaultPressSelector = selector;
   }
 
   layout() {
@@ -216,7 +216,7 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
 
     const target = evt.composedPath ? evt.composedPath()[0] : evt.target;
     const isDefault = !this.adapter.eventTargetMatches(
-        target, this.suppressedPressSelector);
+        target, this.suppressDefaultPressSelector);
     if (isEnter && isDefault) {
       this.adapter.clickDefaultButton();
     }
--- a/packages/mdc-dialog/foundation.ts
+++ b/packages/mdc-dialog/foundation.ts
@@ -214,6 +214,19 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
       return;
     }
 
+    // `composedPath` is used here, when available, to account for use cases
+    // where a target meant to suppress the default press behaviour
+    // may exist in a shadow root.
+    // For example, a textarea inside a web component:
+    // <mwc-dialog>
+    //   <horizontal-layout>
+    //     #shadow-root (open)
+    //       <mwc-textarea>
+    //         #shadow-root (open)
+    //           <textarea></textarea>
+    //       </mwc-textarea>
+    //   </horizontal-layout>
+    // </mwc-dialog>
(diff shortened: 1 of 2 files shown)


asyncLiz:
I have three primary concerns with this approach:

1. `event.composedPath()` is not available on IE11, so we'd need to check if it exists and fall back to `event.target`
2. This doesn't solve the problem for elements where the shadow root is closed (a future possibility with MWC for production builds)
3. The selectors are fairly brittle. `textarea` works, but `.mdc-menu .mdc-list-item` will not match MWC's menu list item

Instead, what if we approach this by keeping `event.target` for IE11 and assuming shadow roots are closed (so we'd only know the event came from `<mwc-textarea>` for example), but change the selectors to be configurable?

- `foundation.setSuppressDefaultPressSelectors(selectors: string[]): void`
- `foundation.getSuppressDefaultPressSelectors(): string[]` 
- default value of `strings.SUPPRESS_DEFAULT_PRESS_SELECTOR`

This would allow MWC's menu to configure the selectors as `['textarea', 'mwc-textarea', 'mwc-menu mwc-list-item']`, which should work properly. It would also allow components to configure and add their own selectors and elements as enter-key-exempt targets.

WDYT?

43081j:
we shouldn't worry about closed roots. composedPath already excludes these in the path, so it isn't our issue to deal with (it will just result in a higher up non-closed parent being the target).

i like your suggestion. i pushed those if you want to take a look. it should be non-breaking since it isn't on the adapter interface too

packages/mdc-dialog/foundation.ts
```diff
@@ -69,6 +69,7 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
   private scrimClickAction_ = strings.CLOSE_ACTION;
   private autoStackButtons_ = true;
   private areButtonsStacked_ = false;
+  private suppressedPressSelector: string = strings.SUPPRESS_DEFAULT_PRESS_SELECTOR;
```
asyncLiz, line 72: For consistency with the constant, let's rename this and the functions to 

- `suppressDefaultPressSelector`
- `getSupressDefaultPressSelector()`
- `setSupressDefaultPressSelector()`

Since we're suppressing the _default_ press and not all presses.

asyncLiz, line 72: nit: you can remove `: string`, typescript will infer the correct type since you assign it a string type

packages/mdc-dialog/foundation.ts
```diff
@@ -167,6 +168,14 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
     this.autoStackButtons_ = autoStack;
   }
 
+  getSuppressedPressSelector(): string {
+    return this.suppressedPressSelector;
+  }
+
+  setSuppressedPressSelector(selector: string): void {
```
asyncLiz, line 175: nit: go ahead and drop the `: void` here

packages/mdc-dialog/foundation.ts
```diff
@@ -205,8 +214,9 @@ export class MDCDialogFoundation extends MDCFoundation<MDCDialogAdapter> {
       return;
     }
 
+    const target = evt.composedPath ? evt.composedPath()[0] : evt.target;
```
asyncLiz, line 217: We still need to address closed shadow roots because the dialog will behave differently if the shadow root is open or closed.

https://jsbin.com/zakopahove/edit?html,output

For these two textarea elements, the target on keypress is `<closed-textarea>` for the first (because composedPath cannot open the closed shadow root), and `<textarea>` for the second (because the shadow root) is open. If we're only looking for a `textarea` selector, we'll match the second, but not the first.

If this was the _same_ element, but the shadow root mode was toggled based on a production vs debug mode flag, that would lead to a bug.

The dialog shouldn't assume that all of its contents will be open shadow roots. Instead, by just viewing `evt.target`, it places the responsibility on the content (aka the ones who _know_ if their selectors need to be updated) to inform the dialog of what the new selectors are.

In the above jsbin example, the content is responsible for informing the dialog that the selector `"closed-textarea, open-textarea"` should suppress the default press.

packages/mdc-dialog/foundation.ts
43081j: ah i had misread it, i assumed DEFAULT meant its the _default selector_. rather than a selector for the default... 🤦

packages/mdc-dialog/foundation.ts
43081j: This behaviour is correct. It is closed for a reason, so you can't look into it... as far as we're concerned the correct target we're interested in is the first _visible_ element, i.e. the first entry of the composed path (as it already excludes nodes from closed roots).

im pretty sure this is right, as closed shadow roots are purposely designed so you won't penetrate their boundaries. we definitely should not be aware there is a textarea inside one. it is up to the consumer to instead provide a higher up selector outside the closed root.

asyncLiz: It's not that the behavior is incorrect on the browser, the semantics of what we're doing is unsafe for the dialog.

`event.composedPath()` has the same penetrative capabilities as `element.shadowRoot.querySelector()`. For open shadow roots it allows external observers to break a shadow root's encapsulation. That's great if you _know_ that all your shadow roots are open and you allow that, but that's something we strictly do not allow with our web components internally.

For MDC's dialog, we'd prefer not to use the `event.composedPath()` API for the same reason that we wouldn't use `element.shadowRoot.querySelector()`. We have use cases where the same element will present with either an open or a closed shadow root depending on the build. If the dialog does not behave the same way for both situations, that can lead to a difficult-to-debug problem (i.e. `<mwc-textarea>` works in debug mode because we're able to penetrate the open shadow root and find its `textarea`, but breaks in production mode because the shadow root is closed).

_We_ could compensate for that with our elements, but then we place that burden on other developers to understand that subtle pattern. It can easily be missed if documentation isn't read carefully. 

The safer option is to avoid shadow root penetrating APIs and require opaque selectors for all web components, regardless of whether or not their shadow root is open or closed.

43081j: i still disagree. look at this from a functional point of view, we want to detect _visible_ targets key events have originated from, where "visible" is in terms of DOM tree exposure. we want to then "do nothing" when one of those origins is a `textarea`. The way to do that is exactly what we have here.

Everything you've described is because you're trying to account for implementation detail in how you've chosen to structure your tests, completely unrelated to the change right here.

Closed shadow roots are purposely impossible to traverse from outside (AFAIK), there is no way to know if one exists and no way to penetrate the boundary of one. So it really isn't any concern of ours to care for the purposes of the code here.

> The safer option is to avoid shadow root penetrating APIs and require opaque selectors for all web components, regardless of whether or not their shadow root is open or closed.

This would re-introduce the problem we are currently trying to fix. We **want** to detect elements in shadow roots here because it is the correct thing to do and the reason this fix and its related issue exists at all. 

Again, this sounds to me like you're trying to apply restrictions to what would otherwise be a valid fix because of how your tests are currently executed/structured around it. If that's the case, the tests are wrong, not the fix. This is specifically a feature to penetrate shadow roots, on purpose. Keep in mind the usage of closed shadow roots is incredibly low, and anyone who chooses to use one im sure is very capable of accounting for the fact that _the platform_ will hide their root's children from _everything_.

43081j: Let's take an example. I have a dialog with a textarea in it, 4 shadow roots deep.

With this, i can safely know that the primary action will not be triggered because i am in a textarea. If i have other elements i want to be able to press "enter" in, i can change the selector to also match those.

Let's now assume this doesn't exist. Let's assume you came up with some other solution where we are more explicit (maybe we pass a handler in, and deal with textarea detection ourselves as the consumer). You still will never be able to detect or enter a closed shadow root (as defined by the browser), so neither can a consumer. So how exactly do you expect them to ever be capable of preventing a textarea in their closed root from triggering the primary action?

It simply should not be possible and is not possible as far as i know, because thats what a closed shadow root is meant to behave like. Given that, the only possible solution for someone with a closed root seems to be to use a selector/element ABOVE the closed root. Which is exactly where we're at here.

asyncLiz: Thanks for the explanation!

I discussed this with the team, and while they agree that we generally want to avoid `composedPath()`, the argument for nested shadow root content is a good one considering that using web components as a "layout" element is not uncommon.

```
<mwc-dialog>
  <horizontal-layout>
    #shadow-root
      <mwc-textarea></mwc-textarea>
      <mwc-checkbox></mwc-checkbox>
  </horizontal-layout>
</mwc-dialog>
```

In the above scenario, only pressing enter while focused in the textarea should prevent the dialog from closing, which would not be possible unless the dialog penetrates the shadow root (the user would have to instruct the dialog that any enter event from `<horizontal-layout>` would prevent the dialog from closing).

To wrap up this discussion, let's add a comment so we can capture why we're using `composedPath()` here:

```
// Prefer composedPath() (if available) to retrieve the original target from any open shadow roots.
// This avoids problems when using web components in the dialog's content
// <div class="mdc-dialog__content">
//   <horizontal-layout>
//     #shadow-root
//       <textarea></textarea>
//   </horizontal-layout>
// </div>
const target = evt.composedPath ? evt.composedPath()[0] : evt.target;
```

packages/mdc-dialog/test/foundation.test.ts
```diff
@@ -27,8 +27,8 @@ import {cssClasses, numbers, strings} from '../constants';
 import {MDCDialogFoundation} from '../foundation';
 
 const ENTER_EVENTS = [
-  {type: 'keydown', key: 'Enter', target: {}},
-  {type: 'keydown', keyCode: 13, target: {}},
+  {type: 'keydown', key: 'Enter', target: {}, composedPath: () => [{}]},
```
asyncLiz, line 30: You should be able to revert these unit test changes now that the foundation handles both composedPath and event target. Our internal test on Chrome and IE11 will ensure that both paths are properly tested.

43081j:
i rebased and threw away the unit test changes. from what i understood, you meant your internal tests will cover those cases so these ones are fine only following the `target` path. it may be worth having test cases in here though at some point to cover both paths maybe?

PR #6432, opened 2020-08-25
refactor(list): expose structural style mixins publicly

crisbeto:
On Angular Material we only need the MDC list base styles for a few components like autocomplete, menu and select. Currently we depend on the private `base_` and `item-base_` in order to avoid bringing in extra styles that we don't need, but this isn't very sustainable.

These changes make the mixins public so that they can be used safely.

### DIFF
--- a/packages/mdc-list/_mixins.scss
+++ b/packages/mdc-list/_mixins.scss
@@ -56,7 +56,7 @@
   );
 
   .mdc-list {
-    @include base_($query);
+    @include base($query);
   }
 
   @include single-line-density(
@@ -82,7 +82,7 @@
 
   .mdc-list-item {
     @include feature-targeting.targets($feat-structure) {
-      @include item-base_;
+      @include item-base;
     }
 
     // For components using aria-activedescendant, the focus pseudoclass is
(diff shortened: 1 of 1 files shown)


packages/mdc-list/_mixins.scss
```diff
@@ -56,7 +56,7 @@
   );
 
   .mdc-list {
-    @include base_($query);
+    @include base($query);
```
abhiomkar, line 59: As per Material Sass Style guide we're not allowing private mixins to exposed as public API. This breaks the encapsulation and makes it hard to refactor later.

But, MDC components can have separate mixins for static (structural) and theme styles. See MDC checkbox as an example.

**Static style mixin:** Styles that don't change when customizing a component should go here.
**Theme style mixin:** Styles that change when customizing should go here (e.g., typography, color, density, etc).

Please feel free to update this PR to add these mixins if applicable.

Let me know if you've any questions!

crisbeto: After looking through it, I think I understand how the checkbox is set up, but I'm not completely sure about the file organization. Should I try to mimic the organization from the checkbox in `mdc-list`? Here's what I have in mind:

* Add a new file called `_list.scss` with all the existing mixins.
* Change `_mixins.scss` to only re-export public mixins from `_list.scss`.
* Split the `base` mixin into `static-styles` and `theme-styles`. The `item-base` mixin can be renamed to just `item-static-styles` since it doesn't have any theme styles.

abhiomkar: crisbeto@ Just discussed this with the team. We're experimenting with theming proposals for MDC and it is not completely fleshed-out. We're expecting the proposal to be out in 1-2 weeks.

In addition to this, MDC list is under great revamp which involves refactoring it to match updated design spec and adds many new features/variants.

Can this change wait until then?

crisbeto: Sure, it isn't urgent since we've been depending on these private mixins for months now. Is there a GitHub issue that I can keep an eye on?

PR #6461, opened 2020-09-08
refactor(segmented-button): Add import only files for segmented button styles for backwards compatibility

annieyw:
Angular Material's [MDC based components](https://github.com/angular/components/tree/master/src/material-experimental) uses `@import`rule  to import styles so these files are needed for backwards compatibility.


abhiomkar (commented):
Can you not use `@use` syntax to import Segmented button Sass? Import-only files are added during Sass module migration for backward compatibility in other places. Not applicable for new Sass files.

Using `@import` is deprecated and will not be supported by MDC in future.

PR #6456, opened 2020-09-07
docs: Add Aurelia wrappers link

MaximBalaganskiy:

PR #6606, opened 2020-11-12
docs: added link to MDC Web bridge for Aurelia.io

mreiche:


asyncLiz:
Thanks for the PR! We already have an open issue for this at https://github.com/material-components/material-components-web/pull/6456

PR #6688, opened 2020-12-09
test: Run conventional-commit-check only on PR's, not pushes.

joyzhong:
Our current build is failing on pushes with `conventional-commit-check` error: "Error: This action can only be invoked in `pull_request` events. Otherwise the pull request can't be inferred."

Workflow syntax: https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/#if-at-the-job-level


abhiomkar (approved):
Thanks Joy!

PR #6704, opened 2020-12-18
Changed `query` to `options` in getting-started.md

facorread:
Changed `query` to `options` according to https://webpack.js.org/loaders/babel-loader/

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -312,7 +312,7 @@ Then configure webpack to convert `app.js` into `bundle.js` by modifying the fol
    {
      test: /\.js$/,
      loader: 'babel-loader',
-     query: {
+     options: {
        presets: ['@babel/preset-env'],
      },
    }
@@ -369,7 +369,7 @@ module.exports = {
       {
         test: /\.js$/,
         loader: 'babel-loader',
-        query: {
+        options: {
           presets: ['@babel/preset-env'],
         },
       }


docs/getting-started.md
```diff
@@ -312,7 +312,7 @@ Then configure webpack to convert `app.js` into `bundle.js` by modifying the fol
    {
      test: /\.js$/,
      loader: 'babel-loader',
-     query: {
+     options: {
        presets: ['@babel/preset-env'],
      },
```
asyncLiz, line 317: Looks like this section is missing the `use: { ... }` block for webpack rules

```suggestion
     use: {
       loader: 'babel-loader',
       options: {
         presets: ['@babel/preset-env'],
       },
     },
```

docs/getting-started.md
```diff
@@ -369,7 +369,7 @@ module.exports = {
       {
         test: /\.js$/,
         loader: 'babel-loader',
-        query: {
+        options: {
           presets: ['@babel/preset-env'],
         },
```
asyncLiz, line 374:
```suggestion
        use: {
          loader: 'babel-loader',
          options: {
            presets: ['@babel/preset-env'],
          },
        },
```

asyncliz (changes requested):
Thanks for fixing this! Just two adjustments to make to add the `use: { ... }` block for webpack rules that I think we're missing.

Additionally, can you go ahead and sign the CLA so that we can accept this PR?

anlexN:
what is CLA? so many duplicated webpack problems. can we solve PR?

PR #6705, opened 2020-12-18
fix: Execute getListItemCount() only once

insys-dev:
Better for performance


asyncLiz:
Thanks for the PR!

This change makes sense to me, can you go ahead and sign the CLA so we can accept it?

insys-dev:
@googlebot I signed it!

asyncLiz (approved):

PR #6728, opened 2020-12-29
Update README.md

SalimBensiali:
Change docs to correct text-transform example


SalimBensiali:
@googlebot I signed it!

asyncLiz (approved):
Thanks for updating our docs!

PR #6759, opened 2021-01-13
Update getting-started.md

jv-k:
According to https://webpack.js.org/concepts/loaders/#example:
"Loaders can be configured with an `options` object (using `query` parameters to set options is still supported but has been deprecated)."

However with babel-loader@8.2.2, when using `query` within rules, you get:

```
 - configuration[0].module.rules[1] has an unknown property 'query'.
```

Replacing `query` with `options` fixes this.


jv-k:
@googlebot I signed it!

PR #6927, opened 2021-03-12
Update getting-started.md

chynenovsemen:
'query' in not valid in Webpack anymore. We need to use 'options' instead.


chynenovsemen:
@googlebot I signed it!

PR #6982, opened 2021-04-02
refactor(list): expose evolution base mixins

crisbeto:
In Angular Material we have several components that are based on simple, single-line lists. Currently if we were to go through the `without-ripple` mixin, we would end up with a lot of unused styles, because all we care about are the base structural styles, colors and typography.

These changes make the `list-base` and `item-base` mixins public so that we have more granular control over the included styles.


brandondiamond (approved):
LGTM!

crisbeto:
Do I have to set a label for the PR to be merged?

PR #7020, opened 2021-04-17
chore: add publishing for subdirectories

asyncLiz:


patrickrodee (approved):

asyncLiz:
Going to close this for now in favor of exporting the deprecated switch directly for now. I'll keep the branch alive though in case we want to revisit this approach

PR #7021, opened 2021-04-17
fix(list): density configuration mixins do not account for leading avatars

devversion:
**fix(list): density configuration mixins do not account for leading avatars**

Currently when someone uses the MDC list w/ evolution styles, it's possible to generate secondary
styles with a different density scale. This can be achieved by using the following mixins: 
`one-line-item-density`, `two-line-item-density` and `three-line-item-density`.

These mixins currently do not account for leading avatars, unlike they do for icons, checkboxes,
switches, radio's, images and videos. This commit adds code fo leading avatars.

**refactor(list): expose evolution base mixins**
In Angular Material we have several components that are based on simple, single-line lists.
Currently if we were to go through the `without-ripple` mixin, we would end up with a lot of
unused styles, because all we care about are the base structural styles, colors and typography.

These changes make the `list-base` and `item-base` mixins public so that we have more granular
control over the included styles.

### DIFF
--- a/packages/mdc-list/_evolution-mixins.scss
+++ b/packages/mdc-list/_evolution-mixins.scss
@@ -38,11 +38,11 @@
   @include _high-contrast-mode($query);
 
   .mdc-list {
-    @include _list-base($query: $query);
+    @include list-base($query: $query);
   }
 
   .mdc-list-item {
-    @include _item-base($query: $query);
+    @include item-base($query: $query);
 
     @include _one-line-item-density(
       variables.$one-line-item-density-scale,
@@ -1003,7 +1003,7 @@
   @include three-line-item-height($height, $query: $query);
 }
 
-@mixin _list-base($query: feature-targeting.all()) {
+@mixin list-base($query: feature-targeting.all()) {
   $feat-color: feature-targeting.create-target($query, color);
   $feat-structure: feature-targeting.create-target($query, structure);
   $feat-typography: feature-targeting.create-target($query, typography);
@@ -1027,7 +1027,7 @@
   }
 }
 
-@mixin _item-base($query: feature-targeting.all()) {
+@mixin item-base($query: feature-targeting.all()) {
   $feat-structure: feature-targeting.create-target($query, structure);
   @include feature-targeting.targets($feat-structure) {
     display: flex;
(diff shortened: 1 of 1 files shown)


devversion:
I've also incorporated @crisbeto's changes to expose the evolution base mixins. The changes he made have been accidentally reverted with https://github.com/material-components/material-components-web/commit/ab99b8064a9b3cc1311c43245b3f5a0b84ef965d as it seems.

packages/mdc-list/_evolution-mixins.scss
```diff
@@ -696,8 +696,7 @@
   @include feature-targeting.targets($feat-structure) {
     // Extra specificity needed to override default customizations.
     &.mdc-list-item {
-      // We rely on auto to avoid clobbering the opposite padding.
-      @include rtl-mixins.reflexive-property(padding, 0, auto);
+      @include rtl-mixins.reflexive-property(padding, 0, 16px);
```
brandondiamond, line 699: Thank you so much for investigating this issue -- we definitely want to get RTL padding straightened out. Unfortunately, setting an explicit padding here won't work with how we've implemented leading and trailing variants.

The use of "auto" is an intentional workaround to get fall-through behavior so that we can set leading and trailing padding independently. By passing auto, we're able to clear the leading padding without influencing the trailing padding ("auto" is ignored and the previous value is retained). 

This allows us to have each leading and trailing style to be independent; if we didn't use this technique, we'd need to style every possible combination of leading/trailing variant independently (since the only cross-browser method of specifying RTL/LTR styles requires both left/right values to be specified simultaneously).

That said, we should definitely figure out a fix for the leading padding being lost in RTL. I'll investigate this on our end and update #7024. For now, it may be best to remove this commit from the PR.

brandondiamond (commented):
Everything looks great -- let's remove the trailing-padding commit from this PR for now. After that, LGTM to merge!

packages/mdc-list/_evolution-mixins.scss
devversion: Oh I see. I started thinking about the possible combinations of each leading/trailing variant, but then just naively though using `16px` as elsewhere, would be sufficient. The problem with `auto` is that it will fall through, but only to the LTR value.. so it ends up with the `0px`. I can't think of a simple solution to this, but the ideal one would be to only apply the LTR value if the browser is not setting `dir=rtl` (though that has other downsides)

devversion: @brandondiamond Removed the commit. thanks for reviewing.

devversion:
@brandondiamond I've rebased the PR. Should be ready for merge.

PR #7028, opened 2021-04-20
fix: broken links in README.md

JamesBLewis:
# Problem
the following [important links](https://github.com/material-components/material-components-web#important-links) result in a 404:
- Getting Started Guide
- Material on other frameworks
- Examples using Material Web
- Supported browsers

These links were broken by #6928. 

# solution
rename g3docs -> docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,13 @@ Material Components for the web is the successor to [Material Design Lite](https
 
 ## Important links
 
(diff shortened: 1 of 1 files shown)


JamesBLewis:
@googlebot I signed it!

allan-chen:
Thanks for the PR - we're working on a fix and they should be automatically rolled back soon

PR #7039, opened 2021-04-23
chore: support ts parsing for dependency test

asyncLiz:
Adds support for using TS syntax in https://github.com/material-components/material-components-web/blob/master/packages/material-components-web/index.ts

### DIFF
--- a/scripts/check-pkg-for-release.js
+++ b/scripts/check-pkg-for-release.js
@@ -37,6 +37,7 @@ const childProcess = require('child_process');
 const {default: traverse} = require('babel-traverse');
 const camelCase = require('camel-case');
 const recast = require('recast');
+const typescriptParser = require('recast/parsers/typescript');
 
 const CLI_PACKAGE_JSON_RELATIVE_PATH = process.argv[2];
 if (!CLI_PACKAGE_JSON_RELATIVE_PATH) {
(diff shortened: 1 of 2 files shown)


scripts/check-pkg-for-release.js
```diff
@@ -256,11 +254,15 @@ function checkAutoInitAddedInMDCPackage(ast) {
       const callee = node.expression.callee;
       const args = node.expression.arguments;
       if (callee.object.name === 'autoInit' && callee.property.name === 'register') {
-        const expression = args.find((value) => {
-          return value.type === 'MemberExpression';
-        });
-        if (expression.object.name === nameCamel) {
-          autoInitedCount++;
+        for (let value of args) {
+          while (value.type === 'TSAsExpression') {
```
patrickrodee, line 257: Can you add a comment explaining why we check for the string "TSAsExpression"?

scripts/check-pkg-for-release.js
```diff
@@ -256,11 +254,15 @@ function checkAutoInitAddedInMDCPackage(ast) {
...
+            value = value.expression;
+          }
+
+          if (value.type === 'MemberExpression' && value.object.name === nameCamel) {
```
patrickrodee, line 261: Can you add a comment explaining why we check for the string "MemberExpression"?

patrickrodee (approved):
Approved with some clarifying comment requests

scripts/check-pkg-for-release.js
```diff
@@ -210,9 +209,7 @@ function checkJSDependencyAddedInMDCPackage() {
   const nameCamel = camelCase(CLI_PACKAGE_JSON.name.replace('@material/', ''));
   const src = fs.readFileSync(path.join(process.env.PWD, MASTER_TS_RELATIVE_PATH), 'utf8');
   const ast = recast.parse(src, {
-    parser: {
-      parse: (code) => parser.parse(code, {sourceType: 'module'}),
-    },
+    parser: require('recast/parsers/typescript'),
```
patrickrodee, line 212: Could we move this `require` to the top of the file and reference it below?

```
const typescriptParser = require('recast/parsers/typescript');
...
const ast = recast.parse(src, {parser: typescriptParser});
```

asyncLiz: Done

scripts/check-pkg-for-release.js
asyncLiz: Done

scripts/check-pkg-for-release.js
asyncLiz: Done

patrickrodee (approved):
🔥

PR #7141, opened 2021-05-21
fix(select): Remove @forward "../functions"

KajdeMunter:
https://github.com/material-components/material-components-web/commit/2fcee40cba11ac8e62541b48ef77e85c74c4a675 removed the function file but it is still used in `packages/mdc-select/helper-text/mdc-select-helper-text.import.scss`.

That means:
```
@import "@material/select/helper-text/mdc-select-helper-text";
```
Is broken.


KajdeMunter:
@googlebot I signed it!

asyncLiz:
Thanks for the PR! These import files were automatically generated to support the move from `@import` to `@use`. MDC does not support `@import` in its current versions anymore, and these files will eventually be removed once internal support has been migrated.

Breaking changes such as these are not updated in the legacy import files.

Issue #7133, opened 2021-05-18
TopAppBar is always fixed


mcookdev:
## Bug report
TopAppBar (by default) is always using `position: fixed`, even when **not** using `mdc-top-app-bar--fixed`.

### Steps to reproduce
- Go to https://material-components.github.io/material-components-web-catalog/#/component/top-app-bar
- Inspect the "Standard" demo
  - `position: fixed` is set

### Actual behavior
The "Standard" demo appears to be _faking_ a non-fixed behavior by using inline styles to set `top: 0px` - though this does not work correctly as things will still slide under the TopAppBar (see screenshot below).

### Expected behavior
TopAppBar should only be `fixed` when using the `mdc-top-app-bar--fixed` class.

### Screenshots
![image](https://user-images.githubusercontent.com/18746762/118690785-03cf5f00-b7ce-11eb-9598-2c9ae04cfe04.png)
![image](https://user-images.githubusercontent.com/18746762/118690801-08941300-b7ce-11eb-831c-e2f1adf29242.png)

### Your Environment:

| Software         | Version(s) |
| ---------------- | ---------- |
| MDC Web          | 11
| Browser          | Chrome 90
| Operating System | MacOSX

### Possible solution
`position: fixed` should be moved to the `mdc-top-app-bar--fixed` class.

asyncLiz:
Thanks for the bug!

Even though the linked demo is very old (v4), the behavior of the standard app bar is accurate. Here's an up-to-date demo: https://jsbin.com/pomazuzaxi/edit?html,output

The standard app bar should slide away when scrolling down, then peek back as the user scrolls up. The app bar will cover some content (which can result in content being scrolled underneath the app bar) as it peeks.

I'll be closing this bug as working as intended, but please feel free to comment and we'll reopen if needed!

PR #7147, opened 2021-05-26
fix: position of mdc-top-app-bar

adityatheoctocatdev:
### solved #7133


adityatheoctocatdev:
> @googlebot I signed it!

@googlebot I signed it!

asyncLiz:
The issue in question (#7133) is actually working as intended, so we won't need this change anymore, but thanks for the PR!

PR #7151, opened 2021-05-29
docs: Changed Query to Options for webpack babel config in getting-started.md

MateusAmin:
Changed Query to Options for webpack babel config after receiving this error: 
```
```
I followed the guide precisely and after receiving the error looked at the babel documentation. I only see mentions of the query syntax in this use case from 2016-17... 

I do not have a good understanding of npm and webpacks. Just dabbling for a landing page to a flutter app.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -312,7 +312,7 @@ Then configure webpack to convert `app.js` into `bundle.js` by modifying the fol
    {
      test: /\.js$/,
      loader: 'babel-loader',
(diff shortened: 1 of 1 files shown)


MateusAmin:
@googlebot I signed it!

mateus.amin@gmail.com

MateusAmin:
***sigh***

MateusAmin:
https://github.com/material-components/material-components-web/pull/7152

PR #7152, opened 2021-05-29
docs: Updated webpack babel config

MateusAmin:
Updated webpack babel config after receiving this error: 
```
```
I followed the guide precisely and after receiving the error looked at the babel documentation. I only see mentions of the query syntax in this use case from 2016-17...  I do not have a good understanding of npm and webpacks. Just dabbling for a landing page to a flutter app.

babel config refs:

https://webpack.js.org/loaders/babel-loader/
https://babeljs.io/setup#installation

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -312,7 +312,7 @@ Then configure webpack to convert `app.js` into `bundle.js` by modifying the fol
    {
      test: /\.js$/,
      loader: 'babel-loader',
-     query: {
+     options: {
        presets: ['@babel/preset-env'],
      },
    }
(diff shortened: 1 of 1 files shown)


MateusAmin:
Might as well do it properly. One second.

MateusAmin:
@asyncLiz This should work as requested and close the two outstanding PRs on this topic.

https://github.com/material-components/material-components-web/pull/6704#pullrequestreview-559706802

https://github.com/material-components/material-components-web/pull/6759

MateusAmin:
I take you are going all in on Flutter with respect to allocating resources to material design implementation on the web.

PR #7285, opened 2021-08-04
docs: mdc-card__media used twice

mhitza:


mhitza:
Just noticed that the first class usage has a `my-` prefix, my bad :)

PR #7287, opened 2021-08-04
chore: remove d.ts bundles

asyncLiz:
This is blocking mwc-switch (https://github.com/material-components/material-components-web-components/pull/2591/checks?check_run_id=3246042094) due to a TypeScript error where our files will resolve to two different types:

One (`observer-foundation.ts`) resolves to `node_modules/@material/base/foundation.d.ts` (correct) since it uses a relative import path. The other (MWC's `base-element.ts`) resolves to `node_modules/@material/base/dist/mdc.base.d.ts` (incorrect) since it uses an absolute import path which finds the `"types"` package.json param and uses that.

This leads to confusing error messages ("MDCFoundation<AdapterType> is not assignable to MDCFoundation<AdapterType>"). With the exception of the all-in-one package, our types can use source file d.ts type files instead of a bundled type file.

### DIFF
--- a/packages/mdc-animation/package.json
+++ b/packages/mdc-animation/package.json
@@ -11,7 +11,6 @@
   "main": "dist/mdc.animation.js",
   "module": "index.js",
   "sideEffects": false,
-  "types": "dist/mdc.animation.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/material-components/material-components-web.git",
--- a/packages/mdc-auto-init/package.json
+++ b/packages/mdc-auto-init/package.json
@@ -5,7 +5,6 @@
   "main": "dist/mdc.autoInit.js",
   "module": "index.js",
   "sideEffects": false,
-  "types": "dist/mdc.autoInit.d.ts",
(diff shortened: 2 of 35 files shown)


allan-chen (approved):

abhiomkar (approved):

asyncLiz:
> Just curious how does the TS types work for independant packages in absence of `types` parameter in package.json file.

When importing a file, TS will resolve to the path of the file. For example:

```ts
import {MDCFoundation} from '@material/base';
```

TS (if using node resolution), will find the package in `node_modules/@material/base/package.json` and first look for a `"types"` key. If it doesn't find that key/file to use, it will next look for `node_modules/@materal/base/index.d.ts`, which it will find and use.

For this import:

```ts
```

TS does something similar. It finds the package in `node_modules/@material/base/package.json` and looks for `"types"` first. If it can't find that, then the first thing it looks for next is `node_modules/@material/base/foundation/index.d.ts`, followed by `node_modules/@material/base/foundation.d.ts` (it'll find the last one).

PR #7297, opened 2021-08-06
chore: clean .gitattributes

TKDev7:
Enforce Unix newlines and remove redundant attributes. I believe they should be handled automatically by GitHub now.


MrHBS:
@googlebot I signed it!

anlexN:
can we solve PR?

PR #7321, opened 2021-08-15
fix(chips): expose all symbols through main entry point

crisbeto:
Fixes the following issues related to the chips entry point:
* The chip and chip action symbols weren't exported through the primary entry point, making them difficult to consume in a Bazel environment.
* The various `types.ts` files weren't being exported.
* Some of the symbols had to be re-exported through different names, because they were conflicting with other symbols. E.g. both the chip and chip action had enums called `Events` and `CssClasses`.

### DIFF
--- a/packages/mdc-chips/action/index.ts
+++ b/packages/mdc-chips/action/index.ts
@@ -23,7 +23,13 @@
 
 export * from './adapter';
 export * from './component';
-export * from './constants';
 export * from './foundation';
 export * from './primary-foundation';
 export * from './trailing-foundation';
+export * from './types';
+export {
+  CssClasses as ChipActionCssClasses,
+  Events as ChipActionEvents,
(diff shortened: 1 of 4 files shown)


packages/mdc-chips/action/index.ts
```diff
@@ -23,7 +23,13 @@
 
 export * from './adapter';
 export * from './component';
-export * from './constants';
 export * from './foundation';
 export * from './primary-foundation';
 export * from './trailing-foundation';
+export * from './types';
+export {
+  CssClasses as ChipActionCssClasses,
```
crisbeto, line 31: I'm not attached to this naming scheme, but I went with it because I saw something similar in the other entry points.

patrickrodee (approved):

patrickrodee:
Closing since this is now obsolete after #7378

PR #7354, opened 2021-08-25
fix(list): Removed unnecessary setTimeout method for focusOut event.

akaSourav:
`FocusEvent.relatedTarget` is the secondary target that is receiving focus in `focusOut` event. So we can use this `EventTarget` to find out whether the focus is inside the root.


anlexN:
microsoft gives up IE, so we can give up it too. can we solve PR?

Issue #5784, opened 2020-04-06
[Theme] Update to M2 and clean up text styles


asyncLiz:
## Feature Request
MDC's theme needs to be cleaned up. There are references to the old M1 spec and some missing features that are blocking theming for MWC.

### Proposed solution
**1. Text style names**
In M1, text styles were referred to as "primary" and "secondary". This caused confusion with "primary" and "secondary" theme colors. In M2 text opacity is now referred to as ["high" and "medium" emphasis](https://material.io/design/color/text-legibility.html#text-backgrounds).

Our [text styles](https://github.com/material-components/material-components-web/tree/master/packages/mdc-theme#text-styles) should replace `primary` with `high` and `secondary` with `medium`.

**2. Provide on-surface text styles**
We provide keys for [`text-<STYLE>-on-background`](https://github.com/material-components/material-components-web/blob/6a40ef217f597138ee2920d2160364649dbf5620/packages/mdc-theme/_variables.scss#L101-L106), but none for `on-surface`. These should be added, and components that use `on-surface` with a custom opacity should use the new keys.

For example, textfield's [`$ink-color`](https://github.com/material-components/material-components-web/blob/776291ef03205e4063b4040eb66f9648e16b4af6/packages/mdc-textfield/_variables.scss#L47) should use `text-high-on-surface` instead of `rgba(on-surface, 0.87)`.

**3. Use $text-emphasis values**

[`$text-emphasis`](https://github.com/material-components/material-components-web/blob/6a40ef217f597138ee2920d2160364649dbf5620/packages/mdc-theme/_variables.scss#L70-L74) defines M2 text opacity values, but they are not currently being used in the [`$text-colors`](https://github.com/material-components/material-components-web/blob/6a40ef217f597138ee2920d2160364649dbf5620/packages/mdc-theme/_variables.scss#L53-L68). This should change.

[...]

PR #7397, opened 2021-09-16
refactor: update to M2 text style names

ttalbot:
Update text style names so that there is no more confusion with "primary" and "secondary".
Names replaced by `high` and `medium` to match M2 high and medium emphasis test styles.
as referenced in #5784 first solution.


asyncLiz:
Thanks for the PR! As mentioned in #5784, we're focused on implementing [Material You](https://github.com/material-components/material-web/issues/2441) right now, so refactoring the existing M2 theme keys is extra work that will quickly become obsolete with the new tokens and theming support provided by Material You :)

Issue #7391, opened 2021-09-14
[checkbox] sass compilation error with last sass "sass@1.40.0"


ftaiolivista:
```
sass --style=expanded --load-path=../node_modules --load-path=/home/vista/dev/iter-mono/iter-web/app/styles --load-path=/home/vista/dev/iter-mono/iter-web --load-path=/home/vista/dev/iter-mono/iter-web/node_modules app/styles/index.scss
Error: Expected "(" or ".".
    ╷
329 │   $checkbox-padding: calc((_ripple-size - _icon-size) / 2);
    │                                        ^
    ╵
  ../node_modules/@material/checkbox/_checkbox-theme.scss 329:40  @use
  ../node_modules/@material/checkbox/_checkbox.scss 42:1          @forward
  ../node_modules/@material/checkbox/_variables.scss 24:1         @forward
  @material/_index.scss 1:1                                       @use
  @vista/iter-theme/app/styles/main.scss 16:1                     @import

```

SlickNicky10:
The same issue is occurring with dialog.

```
Error: Expected "(" or ".".
    ╷
834 │       calc(title - 2 * close),
    │                 ^
    ╵
  node_modules\@material\dialog\_mixins.scss 834:17  @forward
  node_modules\@material\dialog\_index.scss 2:1      @use
  node_modules\@smui\dialog\_style.scss 1:1          @use
  node_modules\smui-theme\_style.scss 9:1            @use
  node_modules\smui-theme\_index.scss 2:1            root stylesheet
```

shvlv:
Look like this is caused by https://github.com/sass/dart-sass/releases/tag/1.40.0.

ftaiolivista:
> This may break your stylesheets, but only if they were already producing broken CSS.

dfreedm:
Looks like sass is parsing templated `calc()` expressions used by `@theme.property()` now.
https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/_checkbox-theme.scss#L329-L336

aomarks:
Found a few other spots, but they'll be fixed in https://github.com/material-components/material-components-web/pull/7402

melknooHansNatur:
Is there a fix for this now? Do I need to update anything? Still getting the error :(

PR #7402, opened 2021-09-17
chore: Bump Node and Sass versions, and some misc housekeeping

aomarks:
These changes make this repo compile cleanly for a user running a modern version of Node, and fixes a Sass syntax error that occurs after a minor version bump.

- Remove `fibers` dependency because it is not compatible with Node 16+. It causes `npm run build` to hard-crash Node. See https://github.com/webpack-contrib/sass-loader#string and https://github.com/laverdet/node-fibers#readme.
- Bump all workflows from Node 10/12 to 14. Node 10 is no longer supported. Node 12 is maintenance LTS. Node 14 is active LTS. See https://nodejs.org/en/about/releases/
- Bump Sass version so that the next fix is caught in CI...
- Use a quoted syntax for some more syntax-invalid `calc()` rules for compatibility with Sass 1.41, which checks them more thoroughly (see https://github.com/material-components/material-components-web/issues/7391 and b/199808917)
- Fix some eslint errors in a script.
- Fix sass lint errors by allowing `@provide` and `@require` rules in the sass linter config.
- Prevent hand-written `packages/mdc-base/externs.d.t`s file from being deleted by `npm run clean`.

Fixes https://github.com/material-components/material-components-web/issues/7391

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: 14
           registry-url: https://registry.npmjs.org/
       - name: Cache dependencies
         uses: actions/cache@v2
--- a/.github/workflows/release-nightly.yml
+++ b/.github/workflows/release-nightly.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: 14
           registry-url: https://registry.npmjs.org/
       - name: Cache dependencies
         uses: actions/cache@v2
--- a/.github/workflows/release-pull-request.yml
+++ b/.github/workflows/release-pull-request.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: 14
           registry-url: https://registry.npmjs.org/
       - name: Cache dependencies
         uses: actions/cache@v2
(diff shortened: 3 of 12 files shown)


joyzhong (changes requested):
I think `package-lock.json` can be reverted?

aomarks:
> I think `package-lock.json` can be reverted?

Is that because you prefer to stay on package-lock v1? If the team is mostly still on npm 6 instead of npm 7, then I think that makes sense yeah.

FWIU the v2 package lock format is backwards compatible with 6. But if an npm 6 user runs `npm install`, then it will be replaced with a v1 package-lock (just tested it, and that's what happens). If you're mostly on npm 6, then maybe that churn would be annoying.

aomarks:
> > I think `package-lock.json` can be reverted?
> 
> Is that because you prefer to stay on package-lock v1? If the team is mostly still on npm 6 instead of npm 7, then I think that makes sense yeah.
> 
> FWIU the v2 package lock format is backwards compatible with 6. But if an npm 6 user runs `npm install`, then it will be replaced with a v1 package-lock (just tested it, and that's what happens). If you're mostly on npm 6, then maybe that churn would be annoying. Reverse happens if you're on npm 7. So I think it depends on which version most of the team is on!

Ok, I reverted that changed (rebased so the commits remain easier to read). Package lock remains on v1.

joyzhong:
Hi Al, sorry for the churn - I'm still learning about npm / package-lock myself. Discussed with Abhinay and the original `package-lock.json` that you checked in (compatible with npm 7) would be preferable actually.

joyzhong (approved):

joyzhong:
LGTM, importing to google3 now!

aomarks:
> (For the package-lock issue, I discussed with Abhinay and we'd like to keep the package-lock that is compatible with npm 6, to match our GitHub actions Node version which uses node 14 ([which uses npm 6](https://nodejs.org/en/download/releases/)).

SGTM, thanks!

PR #7412, opened 2021-09-24
chore: Replace deprecated uglifyjs-webpack-plugin with terser-webpack-plugin

jkevingutierrez:


anlexN:
can we solve PR?

anlexN:
can we solve PR?

PR #7443, opened 2021-11-10
feat(mdc-list): introduce selection change event

devversion:
Introduces a new custom event called `MDCList:selectionChange` that
fires whenever the selection of the list has changed through user
interaction. This matches with the semantics of the native `change`
event of input controls.

This event is useful as not every type of list has an underlying
native control, like in the checkbox-type list, where the `change`
event from the control may bubble up. Even for those cases though
the native event is not guaranteed to fire in all cases. e.g. it
does not fire when a user leverages the `CTRL + A` keybinding we
support in the list foundation.

This commit adds a new type of event/notification that can be
leveraged in the foundation through an adapter method (needed by
Angular), or consumers could use it through `MDCList:selectionChange`.

Note that the custom action event being emitted here is not semantically
equivalent as this one fires regardless of whether the selection has actually changed
or not. This seems like another use-cases, as seen in the MDC menu. 

**Note**: also fixes that the unit tests did not run properly as there was a `fit` leftover pushed to `master`. This commit fixes that and sets up a lint to prevent this in the future.

### DIFF
--- a/packages/mdc-list/foundation.ts
+++ b/packages/mdc-list/foundation.ts
@@ -34,6 +34,17 @@ function isNumberArray(selectedIndex: MDCListIndex): selectedIndex is number[] {
   return selectedIndex instanceof Array;
 }
 
+type SelectionUpdateOptions = {
+  /** Whether the update was triggered by an user interaction. */
+  isUserInteraction?: boolean;
+  /**
+   * Whether the UI should be updated regardless of whether the
+   * selection would be a noop according to the foundation state.
+   * https://github.com/material-components/material-components-web/commit/5d060518804437aa1ae3152562f1bb78b1af4aa6.
+   */
+  forceUpdate?: boolean;
+}
+
 export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
   static override get strings() {
     return strings;
@@ -61,6 +72,7 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
       isRootFocused: () => false,
       listItemAtIndexHasClass: () => false,
       notifyAction: () => undefined,
+      notifySelectionChange: () => {},
       removeClassForElementIndex: () => undefined,
       setAttributeForElementIndex: () => undefined,
       setCheckedCheckboxOrRadioAtIndex: () => undefined,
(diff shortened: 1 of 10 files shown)


packages/mdc-list/foundation.ts
```diff
@@ -657,33 +712,41 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
    * toggleCheckbox value. User interaction should not toggle list item(s) when
```
brandondiamond, line 712: Looks like we also need to update "toggleCheckbox" name in this comment.

packages/mdc-list/foundation.ts
```diff
@@ -657,33 +712,41 @@ export class MDCListFoundation extends MDCFoundation<MDCListAdapter> {
...
-    let isChecked = this.adapter.isCheckboxCheckedAtIndex(index);
-
-    if (toggleCheckbox) {
-      isChecked = !isChecked;
-      this.adapter.setCheckedCheckboxOrRadioAtIndex(index, isChecked);
+    const adapterIsChecked = this.adapter.isCheckboxCheckedAtIndex(index);
+
+    // By default the checked value from the adapter is toggled unless the
+    // checked state in the adapter has already been updated beforehand.
+    // This can be happen when the underlying native checkbox has already
```
brandondiamond, line 730: Hm, I'm not sure this is supposed to be possible --  I believe we disable pointer-events on all contained inputs (in css). Is there a situation in which the underlying control might still be toggled directly?

brandondiamond (commented):
Lots of nice improvements to names and code organization. Thanks for all the effort that went into this PR!

Added a general comment about circumstances in which the underlying control would have received a click event -- I may be misunderstanding the comment, but the intention is for these controls not to be tappable/clickable (via pointer-events: none).

packages/mdc-list/foundation.ts
devversion: Yeah, good point. In Angular Material IIRC we disabled this so that we would not need this special logic, but based on the existance of this property (and the way it is used in the component adapter), it seemed like this was still needed, especially since any arbitrary `input[type="checkbox/radio]` would be picked up by the `MDCList` component.

If this is no longer needed (which would be nice), then we'd need to double-check/enforce that whether pointer events are actually prevented + we need to enforce that the native control is not part of the tab order (as one could still tab to the input and press space right?). maybe that should be a follow-up? I just cleaned this up a little.

devversion: In general I think the goal was to allow for the native `change` event to work, but that seems to not work correclty anyway wtihin an list item.

packages/mdc-list/foundation.ts
devversion: fixed 👍

packages/mdc-list/foundation.ts
brandondiamond: All excellent points. I think most of this is legacy, unfortunately -- the pointer-events stuff came with List Evolution. Before that, the controls were clickable (though list would manipulate tabindex IIRC). 

The latest guidance is for list items to be the sole focus of interaction; any embedded control is, at least in spirit if not implementation, a visual indicator. 

All that being said, I think addressing this is out of scope. So, as long as we're able to preserve existing behavior (while adding the new event), we should be in good shape. Is this consistent with the current state of the PR?

devversion: @brandondiamond thanks for the confirmation. I agree that the embedded control is really just a visual indicator. Let me know if you'd want some help on addressing this in the future (happy to help!)

For this PR: correct. I really just renamed the property from `toggleCheckbox` to something a little more accurate w/ some additional comments. Mostly because I found it confusing and very inaccurate. Everything should behave the same, just the `notifySelectionChange` event is added on top of the existing logic.

devversion:
merged with https://github.com/material-components/material-components-web/commit/7d8ea4624e7dcdc5ce69edf1e747c77354457f42

Issue #7404, opened 2021-09-17
MDCSlider rounding error


simonziegler:
## Bug report
A rounding error exists in [slider foundation.ts at line 1031](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts#L1031) when:

- value is 33.3
- min is 0
- step is 0.1

The correct calculated value would be 333, however a value of 332.99999999999994 is yielded, thus throwing an exception on line 1033.

### Your Environment:

| Software         | Version(s) |
| ---------------- | ---------- |
| MDC Web          |   v12.0.0  |
| Browser          |    Edge 93.0.961.47 |
| Operating System |   Windows 10 (fully updated) |

simonziegler:
I found this for reference

https://www.google.co.uk/amp/s/www.avioconsulting.com/blog/overcoming-javascript-numeric-precision-issues%3fhs_amp=true

simonziegler:
I just took a look at the code now that v13.0.0 is released and believe that the issue remains.

simonziegler:
Sorry to push @joyzhong, but can you please give an indication of when you will be able to look at this issue?

joyzhong:
Hi, sorry I likely won't get around to this for at least a few weeks. Feel free to investigate in the meantime and send us a PR if you find a fix. Thank you!

vkirilyuk:
@simonziegler, I propose a fix like

\- const numStepsValueFromMin = (value - min) / step;

\+ let numStepsValueFromMin = (value - min) / step;
\+ // 3.3 / 0.1 = 32.99999999999999 in Chrome 95
\+ numStepsValueFromMin = Number(numStepsValueFromMin.toFixed(6));

but it will take me time to prepare a PR (as I have no experience). Could you do that please?

simonziegler:
Let me see what I can do. I'm more familiar with C# than JavaScript so will make sure I'm certain of my stuff before doing the PR.

simonziegler:
@joyzhong and @vkirilyuk I have submitted a PR. I'm waiting for signature on the Contributor Licence Agreement for the PR to pass tests prior to somebody at your end reviewing the changes. The PR does fix this bug (following a similar pattern to that suggested by @vkirilyuk above), however I still believe a more resilient approach is proper decimal math, using for instance something like [decimal.js](https://mikemcl.github.io/decimal.js/). Have you included any similar libraries in material components web?

ThomasOrlita:
[...]

PR #7458, opened 2021-11-16
fix(slider): value validation and integer comparison

simonziegler:
Fixes issue #7404. PR submitted upon a request from @vkirilyuk in the issue discussion.  It may still be a better quality solution to use a full precision decimal representation for these two integer checks. See [this article for ideas](https://www.avioconsulting.com/blog/overcoming-javascript-numeric-precision-issues)

### DIFF
--- a/packages/mdc-focus-ring/_focus-ring.scss
+++ b/packages/mdc-focus-ring/_focus-ring.scss
@@ -21,6 +21,7 @@
 //
 
 @use '@material/feature-targeting/feature-targeting';
+@use '@material/rtl/rtl';
 
 $inner-ring-radius: 6px !default;
 $inner-ring-width: 2px !default;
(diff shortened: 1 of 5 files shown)


joyzhong:
Thanks for the fix! Looks good, but could you also please add a test?
https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/test/foundation.test.ts

simonziegler:
@joyzhong I'm setting myself a challenge to see if I can but this is way outside my comfort zone. Would it be something similar to [line 42](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/test/foundation.test.ts#L42)?

joyzhong:
That's correct, it could be something like https://github.com/material-components/material-components-web/blob/978a3b5bbf8c55fd1c81e09cef6bd3912aed92fe/packages/mdc-slider/test/foundation.test.ts#L186 where you verify that we don't throw an error for certain initial values that would previously have thrown.

simonziegler:
> That's correct, it could be something like
> 
> https://github.com/material-components/material-components-web/blob/978a3b5bbf8c55fd1c81e09cef6bd3912aed92fe/packages/mdc-slider/test/foundation.test.ts#L186
> 
> where you verify that we don't throw an error for certain initial values that would previously have thrown.

Thanks - I hope I got this right in the test code.

Issue #7388, opened 2021-09-11
[mdc-ripple] Inconsistency of tag name when the ripple class is applied


mhitza:
## Bug report
In the various component README's when the component's associated ripple class is applied it's either on a `div` or `span` element. Eg: some button examples use span, but button examples within other components might be with div tags.

Unless the block/inline display properties of div/span matter for the underlying ripple class (which I'd guess it doesn't) I think it would be preferred to stick to a single tag in all the examples.

joyzhong:
Button examples should use <span> rather than <div> for the ripple element - see [reasoning](https://stackoverflow.com/questions/12982269/is-it-semantically-incorrect-to-put-a-div-or-span-inside-of-a-button/48383389).

Please feel free to send us a PR to fix any inconsistencies you see, thanks!

PR #7490, opened 2021-12-15
docs: use span tags within inline elements

mhitza:
Related to #7388 #7396


asyncLiz (approved):
Thanks for the fix! Sorry for the long delay 😅

anlexN:
can we solve PR?

Issue #5638, opened 2020-02-20
Investigate adding sourcemaps and removing Babel


cdock1029:
### Steps to reproduce
1. Use packages that depend on `@material/...` components
2. Build
3. Observe bundler not finding source files listed in source maps.

### Actual behavior
Parcel bundler showing many errors of `Could not load source file.... in source map..`
because only javascript files are uploaded to npm, but source map is looking for typescript files.

### Expected behavior
No error messages polluting the console. The source map is directed to the javascript files or the typescript files are uploaded to npm.

### Screenshots

![error message](https://doc-00-28-docs.googleusercontent.com/docs/securesc/ha0ro937gcuc7l7deffksulhg5h7mbp1/707c8r0uht05dv6igocvgdsefbrbbbte/1582235100000/09745027216445744095/*/1d3W9frBFrq5nBjTxVZxLeiYZ6pqb5Vnz)

### Your Environment:

| Software         | Version(s) |
| ---------------- | ---------- |
| MDC Web          |3.2.0
| Operating System |Pop!_OS 19.10

### Additional context

https://github.com/parcel-bundler/parcel/issues/2185

https://github.com/material-components/material-components-web/pull/4206

### Possible solution

Upload typescript files to npm.
OR, remove sourcemaps

EstebanG23:
Thank you for reporting. Moving to backlog to add ts files.

mattmspie:
I recently updated `react-scripts` in a React project that includes `@material/circular-progress`. I'm now seeing the following warnings during build, that look the same as this issue:

```
Failed to parse source map from '<path-sanitized>/node_modules/@material/base/component.ts' file : error : ENOENT: no such file or directory, open '<path-sanitized>/node_modules/@material/base/component.ts'

Failed to parse source map from '<path-sanitized>/node_modules/@material/base/foundation.ts' file : error : ENOENT: no such file or directory, open '<path-sanitized>/node_modules/@material/base/foundation.ts'

Failed to parse source map from '<path-sanitized>/node_modules/@material/circular-progress/adapter.ts' file : error : ENOENT: no such file or directory, open '<path-sanitized>/node_modules/@material/circular-progress/adapter.ts'  

Failed to parse source map from '<path-sanitized>/node_modules/@material/circular-progress/component.ts' file : error : ENOENT: no such file or directory, open '<path-sanitized>/node_modules/@material/circular-progress/component.ts'

[...]

PR #7506, opened 2022-01-11
chore: Disable tsc sitemap generation

mattmspie:
The sitemaps included with npm packages are broken because they point to
.ts source files, which are not distributed. Until a final decision is
made as to whether .ts source should be distributed, disable tsc's
sitemap generation.

### DIFF
--- a/tsconfig-base.json
+++ b/tsconfig-base.json
@@ -17,7 +17,7 @@
     "preserveConstEnums": true,
     "pretty": true,
     "skipLibCheck": true,
-    "sourceMap": true,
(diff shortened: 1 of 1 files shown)


mattmspie:
Google CLA signed.

This is meant to be a stop-gap for #5638 until the team decides whether sourcemaps + `.ts` source should be included in npm packages. Hence, not adding `Fixes #...` line to end of commit message.

asyncliz (approved):
This seems like a reasonable solution for now!

anlexN:
can we solve PR?

PR #7591, opened 2022-04-16
fix(list): add command+A for MacOs "select all" functionality.

volvachev:
Fixes a bug in Material `list` when using MacOs `command`+`a` does not work for 'select all functionality'. (see [angular/components#24788](https://github.com/angular/components/pull/24788))


anlexN:
can we solve PR?

Issue #7179, opened 2021-06-12
[webpack-cli] Invalid configuration object: Unknown property 'query'


ChakshuGupta13:
In the [getting-started.md](https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md), while proceeding with [Step 3](https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md#step-3-webpack-with-es2015), I received the following error:
```
[webpack-cli] Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration[0].module.rules[1] has an unknown property 'query'. These properties are valid:
   object { compiler?, dependency?, descriptionData?, enforce?, exclude?, generator?, include?, issuer?, issuerLayer?, layer?, loader?, mimetype?, oneOf?, options?, parser?, realResource?, resolve?, resource?, resourceFragment?, resourceQuery?, rules?, scheme?, sideEffects?, test?, type?, use? }
   -> A rule description with conditions and effects for modules.
```
With the help the documentation of webpack ([babel-loader](https://webpack.js.org/loaders/babel-loader/#root)), I changed the code mentioned in the Step 3 with the following lines:
```
{
    test: /\.js$/,
    use: {
        loader: 'babel-loader',
        options: {
            presets: ['@babel/preset-env']
        }
    },
}
```
Though I am not sure whether it is correct patch for this error but it should be resolved as soon as possible.

steffdimitrov:
Yes, it is correct.

milindgoel15:
I'm also facing similar issue while using babel-loader 
```
Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration has an unknown property 'modules'. 
```

```
const path = require('path');

module.exports = {
    entry: {
        barba: './js/barba.js',
        kursor: './js/kursor.js',
        snowflakes: './js/snowflakes.js',
    },
    modules: {
        rules: [
            {
                test: /\.(js)$/,
                use: "babel-loader",
            },
        ],
    },
    output: {
        path: path.resolve(__dirname, "dist"),
        filename: 'bundle.js',
    },

    mode: 'production'
}
```

apparently, removing babel-loader does fix this error

NSanchay:
yes I am also facing this problem
configuration.module.rules[9].use has an unknown property 'exclude'. These properties are valid:
   object { ident?, loader?, options? }

PR #7654, opened 2022-07-03
docs(getting-started): Update `babel-loader` object (#7179)

ChakshuGupta13:
**Why**
- While proceeding with [Step 3], the following error is received:
```
```

**What**
- Update `babel-loader` object.

**References**
- [`babel-loader` Usage](https://webpack.js.org/loaders/babel-loader/#usage)

[Step 3]: https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md#step-3-webpack-with-es2015

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -311,9 +311,11 @@ Then configure webpack to convert `app.js` into `bundle.js` by modifying the fol
    ```js
    {
      test: /\.js$/,
-     loader: 'babel-loader',
-     query: {
-       presets: ['@babel/preset-env'],
+     use: {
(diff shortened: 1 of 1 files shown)


anlexN:
can we solve PR?

ChakshuGupta13:
Need approval to run workflow on PR - until then merging is blocked.

anlexN:
hello, my friend. I have another question: do you upgrade all webpackjs configurations?

ChakshuGupta13:
Note: Merging is blocked for me.
> The base branch restricts merging to authorized users.

PR #7752, opened 2022-09-28
fix: Makes material component to depend on https://github.com/google/safevalues and be Trusted Type compatible.

neuracr:
This is another attempt for #7718, authored externally.


neuracr:
I fixed safevalues to have the original PR to pass the tests: https://github.com/material-components/material-components-web/actions/runs/3150898483

